Issue #381, opened 2018-05-07
Environment variables file doesn't work with start-api


pcx:
My `template.yaml` looks like this:

```
AWSTemplateFormatVersion : '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Resources:
  TestFunction:
    Type: AWS::Serverless::Function
    Properties:
      Handler: test.handler
      Runtime: nodejs8.10
      Events:
        GetSyncObject:
          Type: Api
          Properties:
            Path: /test
            Method: get
```

The `env.json` looks like this:
```
{
    "TestFunction": {
        "ENV_NAME": "development",
        "DATABASE_URL": "http://dynamodb:8000/"
    }
}
```

The `test.js` file looks like this:

```
module.exports.handler = (event, context, callback) => {
  console.log(process.env);
};
```

I am running the command as:
`sam local start-api --env-vars env.json`

The out put in the terminal does not show the env vars from the file specified above. I am using OSX and `aws-sam-local@0.2.11`.

sanathkr:
You need to define the `Environment` section in the Function for the envars to show up during execution. Only the variables in the Environment section will be read from the file and show up in the execution

pcx:
@sanathkr This was not immediately apparent from the README. Thanks for clearing that up!

PR #538, opened 2018-07-05
Update README for --env-vars option

rishabhkrishna:
Only the variables in the Environment section will be read from the file and show up in the execution

*Issue #381 , Updating readme for this issue

*Description of changes:*

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -641,6 +641,22 @@ defined in your function. The file should be structured as follows:
 .. code:: bash
 
    $ sam local start-api --env-vars env.json
+   
+Only the variables in the Environment section will be read from the file and show up in the execution. In the example below, only TABLE_NAME and BUCKET_NAME will be read if defined in the env.json file
+
+.. code:: yaml
+
+
+   Resources:
+     MyFunction1:
+       Type: AWS::Serverless::Function
+       Properties:
+         Handler: index.handler
+         Runtime: nodejs4.3
+         Environment:
+           Variables:
+             TABLE_NAME: prodtable
+             BUCKET_NAME: prodbucket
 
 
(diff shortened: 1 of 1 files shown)


jfuss:
@rishabhkrishna Thanks for the PR but not sure if this is actually needed. The Environments section already explains this. "If your Lambda function uses environment variables, you can provide values for them will passed to the Docker container. Here is how you would do it:". 

If the values are not in the template, you can't expect them to show up. The file is only to override the values: "Use --env-vars argument of invoke or start-api commands to provide a JSON file that contains values for environment variables defined in your function." 

Maybe the wording of what we already have needs to be more clear?

bdunbar22:
Hi @jfuss, I ran into this as well recently. After rereading the docs I think your current wording is valid, but for me a more clear wording would be:

` Use the ``--env-vars`` argument of ``invoke`` or ``start-api`` commands to
 provide a JSON file that contains values to override the environment variables
 already defined in your function template.`

jfuss:
@bdunbar22 I think that is totally fair and defiantly clearer. 

@rishabhkrishna Would you mind updating to @bdunbar22 suggestion so we can get this merged?

jfuss:
@bdunbar22 Would you be update this or push another PR with the updates you suggested?

bdunbar22:
Hi @jfuss, Yes I should have some time tonight or tomorrow to create one (I don't think I'd have push access to this one). I will mention you in a comment when it is ready.

jfuss:
@bdunbar22 Awesome! Thanks taking this on. 

Given that and no response from @rishabhkrishna I will close this PR out in favor of the one from @bdunbar22

Issue #531, opened 2018-07-05
Getting an invalid response with new line symbol at the response on SAM CLI 0.4.0


makskant:
**Description:**

**Steps to reproduce the issue:**
1. Function handler
```
def handler(event, context):
    return {'body': json.dumps([]), 'statusCode': 200, 'headers': {"Cache-Control":"max-age=1200"}}
```
2. Resource
```
    TestFunction:
        Type: AWS::Serverless::Function
        Properties:
            Handler: test.handler
            Role: !GetAtt FunctionRole.Arn
            CodeUri: functions/
            AutoPublishAlias: live
            Events:
                Api:
                    Type: Api
                    Properties:
                        Path: /test
                        Method: get

```

3. Running `sam local start-api` and trying to requests a http://127.0.0.1:3000/test

**Observed result:**
```
Function returned an invalid response (must include one of: body, headers or statusCode in the response object). Response received: b'\n{"body": "[]", "statusCode": 200, "headers": {"Cache-Control": "max-age=1200"}}'
```
**Expected result:**
Expecting to have an empty array as a response with status code 200

**OS**: Ubuntu 16.04

**Output of `sam --version`**: 0.4.0

**Python version**: 3.6.1.

**Optional Debug logs**:

```
START RequestId: 90197d0b-c013-4575-bbe0-2eb8c3d67714 Version: $LATEST
END RequestId: 90197d0b-c013-4575-bbe0-2eb8c3d67714
REPORT RequestId: 90197d0b-c013-4575-bbe0-2eb8c3d67714 Duration: 339 ms Billed Duration: 400 ms Memory Size: 128 MB Max Memory Used: 29 MB
2018-07-05 16:52:11 http://localhost:None "GET /v1.35/containers/db47404069e7220f7ee521898dad6f27131a5e8f8b6d2b0a6d035aa3ffd9097e/json HTTP/1.1" 200 None
2018-07-05 16:52:11 http://localhost:None "DELETE /v1.35/containers/db47404069e7220f7ee521898dad6f27131a5e8f8b6d2b0a6d035aa3ffd9097e?force=True&v=False&link=False HTTP/1.1" 204 0
2018-07-05 16:52:11 Function returned an invalid response (must include one of: body, headers or statusCode in the response object). Response received: b'\n{"body": "[]", "statusCode": 200, "headers": {"Cache-Control": "max-age=1200"}}'
[...]

PR #541, opened 2018-07-07
fix: invalid response with new line symbol

zsc347:
*Issue #531*

*Description of changes: fix corner case that only one '\n' in response*


jfuss:
@zsc347 Would you be able to fix the test? When I added the decode the last parameter in the `param()` needs to be a str not bytes.

zsc347:
I'll fix it

zsc347:
@jfuss Fix done. Could you pls check?

PR #544, opened 2018-07-09
feat(LocalLambdaInvokeService): Command for start-lambda

jfuss:
Command help text:

```
Usage: samdev local start-lambda [OPTIONS]

  Allows you to run a Local Lambda Service that will service the invoke path
  to your functions for quick development &   testing through the AWS CLI or
  SDKs. When run in a directory that contains your Serverless functions and
  your AWS   SAM template, it will create a local HTTP server that wil
  response to the invoke call to your functions.   When accessed (via
  browser, cli etc), it will launch a Docker container locally to invoke the
  function. It will read   the CodeUri property of AWS::Serverless::Function
  resource to find the path in your file system containing the Lambda
  Function code. This could be the project's root directory for interpreted
  languages like Node & Python, or a build   directory that stores your
  compiled artifacts or a JAR file. If you are using a interpreted language,
  local changes   will be available immediately in Docker container on every
  invoke. For more compiled languages or projects requiring   complex
  packing support, we recommended you run your own building solution and
  point SAM to the directory or file   containing build artifacts.

Options:
  --host TEXT                     Local hostname or IP address to bind to
                                  (default: '127.0.0.1')
  -p, --port INTEGER              Local port number to listen on (default:
                                  '3000')
  -t, --template PATH             AWS SAM template file  [default:
                                  template.[yaml|yml]]
  -n, --env-vars PATH             JSON file containing values for Lambda
                                  function's environment variables.
  -d, --debug-port TEXT           When specified, Lambda function container
                                  will start in debug mode and will expose
                                  this port on localhost.
  --debug-args TEXT               Additional arguments to be passed to the
                                  debugger
  -v, --docker-volume-basedir TEXT
                                  Specifies the location basedir where the SAM
                                  file exists. If the Docker is running on a
                                  remote machine, you must mount the path
                                  where the SAM file exists on the docker
                                  machine and modify this value to match the
                                  remote machine.
  --docker-network TEXT           Specifies the name or id of an existing
                                  docker network to lambda dockercontainers
                                  should connect to, along with the default
                                  bridge network. If not specified,the Lambda
                                  containers will only connect to the default
                                  bridge docker network.
  -l, --log-file TEXT             logfile to send runtime logs to
  --skip-pull-image               Specify whether CLI should skip pulling down
                                  the latest Docker image for Lambda runtime
  --profile TEXT                  Specify which AWS credentials profile to
                                  use.
  --debug                         Turn on debug logging
  --help                          Show this message and exit.
```

*Description of changes:*
Added the command line interface for 'start-lambda'

What is left:
- [x] Unit test
- [ ] Integ tests
- [ ] Update 'help' message

### DIFF
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -55,6 +55,36 @@ def template_click_option():
                         help="AWS SAM template file")
 
 
+def service_common_options(f):
+    """
+    Common CLI Options that are shared for service related commands ('start-api' and 'start-lambda')
+
+    Parameters
+    ----------
+    f function
+        Callback passed by Click
+
+    Returns
+    -------
+    function
+        The callback function
+    """
+    service_options = [
+        click.option('--host',
+                     default="127.0.0.1",
+                     help="Local hostname or IP address to bind to (default: '127.0.0.1')"),
+        click.option("--port", "-p",
+                     default=3000,
+                     help="Local port number to listen on (default: '3000')")
+    ]
+
+    # Reverse the list to maintain ordering of options in help text printed with --help
+    for option in reversed(service_options):
+        option(f)
+
+    return f
+
+
 def invoke_common_options(f):
     """
     Common CLI options shared by "local invoke" and "local start-api" commands
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -57,7 +57,7 @@ def template_click_option():
 
 def service_common_options(f):
     """
-    Common CLI Options that are shared for service related commands ('start-api' and 'start-lambda')
+    Common CLI Options that are shared for service related commands ('start-api' and 'start_lambda')
 
     Parameters
     ----------
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -55,34 +55,38 @@ def template_click_option():
                         help="AWS SAM template file")
 
 
-def service_common_options(f):
-    """
-    Common CLI Options that are shared for service related commands ('start-api' and 'start_lambda')
+def service_common_options(port):
+    def construct_options(f):
+        """
+        Common CLI Options that are shared for service related commands ('start-api' and 'start_lambda')
 
-    Parameters
-    ----------
-    f function
-        Callback passed by Click
+        Parameters
+        ----------
+        f function
+            Callback passed by Click
+        port int
+            port number to use
 
-    Returns
-    -------
-    function
-        The callback function
-    """
-    service_options = [
-        click.option('--host',
-                     default="127.0.0.1",
-                     help="Local hostname or IP address to bind to (default: '127.0.0.1')"),
-        click.option("--port", "-p",
-                     default=3000,
-                     help="Local port number to listen on (default: '3000')")
-    ]
+        Returns
+        -------
+        function
+            The callback function
+        """
+        service_options = [
+            click.option('--host',
+                         default="127.0.0.1",
+                         help="Local hostname or IP address to bind to (default: '127.0.0.1')"),
+            click.option("--port", "-p",
+                         default=port,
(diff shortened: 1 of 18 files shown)


samcli/commands/local/cli_common/options.py
```diff
@@ -55,6 +55,36 @@ def template_click_option():
...
+    -------
+    function
+        The callback function
+    """
+    service_options = [
+        click.option('--host',
+                     default="127.0.0.1",
+                     help="Local hostname or IP address to bind to (default: '127.0.0.1')"),
+        click.option("--port", "-p",
+                     default=3000,
```
sanathkr: `start-lambda` should start in a different port by default. So you can use it with `start-api` without extra configuration. What do you think?

samcli/commands/local/lib/local_lambda_service.py
```diff
@@ -0,0 +1,57 @@
...
+        # to the console or a log file. stderr from Docker container contains runtime logs and output of print
+        # statements from the Lambda function
+        service = LocalLambdaInvokeService(lambda_runner=self.lambda_runner,
+                                           port=self.port,
+                                           host=self.host,
+                                           stderr=self.stderr_stream)
+
+        service.create()
+
+        LOG.info("Starting the Locally Lambda Service. You can now invoke you Lambda Functions through the endpoint.")
```
sanathkr: `Locally` - typo. This message is also unclear when you say "you can invoke Lambda functions through the endpoint?" it is not clear

samcli/commands/local/start_lambda/cli.py
```diff
@@ -0,0 +1,79 @@
+"""
+CLI command for "local start-api" command
```
sanathkr: it is not start-api

samcli/commands/local/start_lambda/cli.py
```diff
@@ -0,0 +1,79 @@
...
+from samcli.cli.main import pass_context, common_options as cli_framework_options
+from samcli.commands.local.cli_common.options import invoke_common_options, service_common_options
+from samcli.commands.local.cli_common.invoke_context import InvokeContext
+from samcli.commands.local.cli_common.user_exceptions import UserException
+from samcli.commands.local.lib.local_lambda_service import LocalLambdaService
+from samcli.commands.validate.lib.exceptions import InvalidSamDocumentException
+
+LOG = logging.getLogger(__name__)
+
+HELP_TEXT = """
```
sanathkr: This help text needs more work. It doesn't well explain how this command works.

samcli/commands/local/start_lambda/cli.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+
+
+@click.command("start-lambda", help=HELP_TEXT, short_help="Runs a Local Lambda Service (for the Invoke path only)")
+@service_common_options
+@invoke_common_options
+@cli_framework_options
+@pass_context
+def cli(ctx,
+        # start-api Specific Options
```
sanathkr: not start-api

samcli/commands/local/cli_common/options.py
jfuss: Good call. I think just doing a +1 is enough. Not sure if the delta is a big deal. Let me know if you think something greater is better.

samcli/commands/local/lib/local_lambda_service.py
jfuss: Updated but still don't think it describes what I want very well. Will think on this message more.

samcli/commands/local/start_lambda/cli.py
jfuss: gah... I didn't copy start-api at all... ;)

samcli/commands/local/start_lambda/cli.py
jfuss: :(

jfuss:
@sanathkr I am ok with that.

PR #563, opened 2018-07-19
docs: Fix supported Python version number

sanathkr:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,7 +100,7 @@ Prerequisites
 ~~~~~~~~~~~~~
 
 - Docker
-- Python2.7 or Python3.6
+- Python2.7 or Python3.6+
 
 Running Serverless projects and functions locally with SAM CLI requires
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -100,7 +100,7 @@ Prerequisites
 ~~~~~~~~~~~~~
 
 - Docker
-- Python2.7 or Python3.6
+- Python2.7 or Python3.6+
```
jfuss: This is not true. We only run tests in Py2.7 and 3.6. What this communicates is that anything above 3.6 is supported. So if Python comes out with a new version (even if it is a minor version), we claim to support it.

I want us to be careful of what version we support. Supporting multiple versions has (non-trivial) additional cost to maintenance.

sanathkr: I agree with you. Until we have tox testing in multiple Py versions this can't be true. I'll discard the PR

PR #526, opened 2018-07-03
Revert "feat(debugging): Support debugging Golang functions. (#495)"

jfuss:
This reverts commit 9ae4b80313d094d3eded4c5ca61731481b36c950.
The commit breaks functional and integ tests. Reverting to make develop buildable

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #495, opened 2018-06-19
Support debugging Golang functions.

austinlparker:
*Issue #, if available:*
#281 
*Description of changes:*
- Modify container options to allow for tracing inside a container (required by Delve)
- Pass debug port into new entrypoint

Note, this will fail until https://github.com/lambci/docker-lambda/pull/97 is merged and a new image is published. In the meantime, if you want to get going with this, you can build that PR's Dockerfile locally and tag it appropriately then `--skip-pull-image` on SAM CLI to use the local version.

Please don't merge this right away, I'd like to make it more robust and have some questions back in #281.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #281, opened 2018-01-27
Debugging Golang code in SAM CLI


tfnaylor:
I am able run my SAM Local Golang project just fine (get expected results, etc.) but cannot get my breakpoints to be hit in Visual Studio Code. I am using a Mac and the project is a simple REST API. 

The VS Code debugger starts but the break icons never get filled in and breaks don't happen. My launch.json file configuration also gives me a warning that it doesn't recognize the debug type even though I can debug pure Golang applications on this machine without issue (Delve installed, etc.).

My launch.json file configurations section looks like this:

```json
{
            "name": "Attach to SAM Local",
            "type": "go",
            "request": "attach",
            "address": "localhost",
            "port": 5858,
            "localRoot": "${workspaceRoot}/main.go",
            "remoteRoot": "/var/task"
}
```
I use the configuration to start the debugger, which seems to start normally (the debug control widget appears) and no errors are reported.

I suspect that the standard Go extension for VS Code (lukehoban.go) isn't compatible with this configuration. Is anyone else having this problem?

afdabro:
I've ran into the same issue. It is a pain not being able to debug from SAM/VsCode. I keep the main func light and use a combination of DI and unit testing for the handle request func. Context and ProxyRequest are mockable making unit testing straight forward.

`
func init() {
// DI arguments init
}

func main() {
	lambda.Start(handleRequest)
}

func handleRequest(ctx context.Context, request events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
	return myDIHandleAPIRequest(ctx, request, ... any DI arguments ...)
}
`

sanathkr:
SAM CLI doesn't have support for debugging Golang support yet. We need an entry point for Golang similar to https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/docker/lambda_container.py#L131.

This should be fairly easy I would think. How do you usually start Go in debugging mode?

austinlparker:
@sanathkr I started looking into this because we're using a lot of golang + lambda and I've made some progress. Go, out of the box, doesn't really have a debugger (other than `gdb` or whatever), `delve` (https://github.com/derekparker/delve) is the thing everyone uses afaik.

[...]

PR #565, opened 2018-07-19
feat(debugging): Support debugging Golang functions.

jfuss:
Golang Debugging was initally implemented in #495, but reverted in #526
due to functional and integration tests failures. This commit reverts #526
and adds in the testing gaps. The specific differences are:

* _get_debug_context method was replaces with the __bool__/__nonzero (py2)
* Added additional unit tests for lambda_container to cover added methods
* Updated Functional tests
* Added debug_context support in start-lambda

*Issue #, if available:*
#281

*Description of changes:*

- [x] Run debugger locally with a go function

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -504,21 +504,21 @@ When debugging, you must compile your function in debug mode:
 You must compile `delve` to run in the container and provide its local path
 via the `--debugger-path` argument. Build delve locally as follows:
 
-`GOARCH=amd64 GOOS=linux go build -o <output path>/dlv github.com/derekparker/delve/cmd/dlv`
+`GOARCH=amd64 GOOS=linux go build -o <delve folder path>/dlv github.com/derekparker/delve/cmd/dlv`
 
 NOTE: The output path needs to end in `/dlv`. The docker container will expect the dlv binary to be in the <output path>
 and will cause mounting issue otherwise.
 
 Then invoke `sam` similar to the following:
 
-`sam local start-api -d 5986 --debugger-path <output path>`
+`sam local start-api -d 5986 --debugger-path <delve folder path>`
 
 NOTE: The `--debugger-path` is the path to the directory that contains the `dlv` binary compiled from the above.
 
 The following is an example launch configuration for Visual Studio Code to
 attach to a debug session.
 
-.. code:: bash
+.. code:: json
 
   {
     "version": "0.2.0",
--- a/README.rst
+++ b/README.rst
@@ -497,14 +497,23 @@ Java, and Python. We require `delve <https://github.com/derekparker/delve>`__
 as the debugger, and wrap your function with it at runtime. The debugger
 is run in headless mode, listening on the debug port.
 
+When debugging, you must compile you function in debug mode:
+
+`GOARCH=amd64 GOOS=linux go build -gcflags='-N -l' -o <output path> <path to code directory>
+
 You must compile `delve` to run in the container and provide its local path
 via the `--debugger-path` argument. Build delve locally as follows:
 
-`GOARCH=amd64 GOOS=linux go build -o <output path> github.com/derekparker/delve/cmd/dlv`
+`GOARCH=amd64 GOOS=linux go build -o <output path>/dlv github.com/derekparker/delve/cmd/dlv`
+
+NOTE: The output path needs to end in `/dlv`. The docker container will expect the dlv binary to be in the <output path>
+and will cause mounting issue otherwise.
 
 Then invoke `sam` similar to the following:
 
 `sam local start-api -d 5986 --debugger-path <output path>`
+
+NOTE: The `--debugger-path` is the path to the directory that contains the `dlv` binary compiled from the above.
 
 The following is an example launch configuration for Visual Studio Code to
 attach to a debug session.
--- a/README.rst
+++ b/README.rst
@@ -506,7 +506,7 @@ via the `--debugger-path` argument. Build delve locally as follows:
 
 `GOARCH=amd64 GOOS=linux go build -o <delve folder path>/dlv github.com/derekparker/delve/cmd/dlv`
 
-NOTE: The output path needs to end in `/dlv`. The docker container will expect the dlv binary to be in the <output path>
+NOTE: The output path needs to end in `/dlv`. The docker container will expect the dlv binary to be in the <delve folder path>
 and will cause mounting issue otherwise.
 
 Then invoke `sam` similar to the following:
--- a/README.rst
+++ b/README.rst
@@ -489,6 +489,47 @@ port to your host machine.
    issue <https://github.com/Microsoft/vscode-python/issues/71>`__ for
    updates.
 
+Debugging Golang functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Golang function debugging is slightly different when compared to Node.JS,
+Java, and Python. We require `delve <https://github.com/derekparker/delve>`__
+as the debugger, and wrap your function with it at runtime. The debugger
+is run in headless mode, listening on the debug port.
+
+You must compile `delve` to run in the container and provide its local path
+via the `--debugger-path` argument. Build delve locally as follows:
+
+`GOARCH=amd64 GOOS=linux go build -o <output path> github.com/derekparker/delve/cmd/dlv`
+
+Then invoke `sam` similar to the following:
+
+`sam local start-api -d 5986 --debugger-path <output path>`
+
+The following is an example launch configuration for Visual Studio Code to
+attach to a debug session.
+
+.. code:: bash
+
+  {
+    "version": "0.2.0",
+    "configurations": [
+    {
+        "name": "Connect to Lambda container",
+        "type": "go",
+        "request": "launch",
+        "mode": "remote",
+        "remotePath": "",
+        "port": <debug port>,
+        "host": "127.0.0.1",
+        "program": "${workspaceRoot}",
+        "env": {},
+        "args": [],
+      },
+    ]
+  }
+
+
 Passing Additional Runtime Debug Arguments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/README.rst
+++ b/README.rst
@@ -497,7 +497,7 @@ Java, and Python. We require `delve <https://github.com/derekparker/delve>`__
 as the debugger, and wrap your function with it at runtime. The debugger
 is run in headless mode, listening on the debug port.
 
-When debugging, you must compile you function in debug mode:
+When debugging, you must compile your function in debug mode:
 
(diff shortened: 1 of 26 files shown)


jfuss:
I have successfully debugged my first Go application. I update the README.rst for the Golang Debugging section to include extra steps I needed to do to get debugging working in VSCode. 

I am considering updating the README in the Golang example we generate from `sam init --runtime go`, to include some of this information on debugging. What is your thought on that @sanathkr and @austinlparker?

Also, shout out to @austinlparker for helping me get debugging working and test everything :)

README.rst
```diff
@@ -489,6 +489,56 @@ port to your host machine.
...
 
+Debugging Golang functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Golang function debugging is slightly different when compared to Node.JS,
+Java, and Python. We require `delve <https://github.com/derekparker/delve>`__
+as the debugger, and wrap your function with it at runtime. The debugger
+is run in headless mode, listening on the debug port.
+
+When debugging, you must compile you function in debug mode:
```
austinlparker: Small typo; `you` should be `your`.

jfuss: :)

austinlparker:
I think some updates to the Golang example might be useful, yeah. I wouldn't say it's blocking a release though.

README.rst
```diff
@@ -489,6 +489,56 @@ port to your host machine.
...
+is run in headless mode, listening on the debug port.
+
+When debugging, you must compile your function in debug mode:
+
+`GOARCH=amd64 GOOS=linux go build -gcflags='-N -l' -o <output path> <path to code directory>
+
+You must compile `delve` to run in the container and provide its local path
+via the `--debugger-path` argument. Build delve locally as follows:
+
+`GOARCH=amd64 GOOS=linux go build -o <output path>/dlv github.com/derekparker/delve/cmd/dlv`
```
sanathkr: Can you change `<output path>` to something like `<delve folder path>` to be more descriptive?

README.rst
```diff
@@ -489,6 +489,56 @@ port to your host machine.
...
+Then invoke `sam` similar to the following:
+
+`sam local start-api -d 5986 --debugger-path <output path>`
+
+NOTE: The `--debugger-path` is the path to the directory that contains the `dlv` binary compiled from the above.
+
+The following is an example launch configuration for Visual Studio Code to
+attach to a debug session.
+
+.. code:: bash
```
sanathkr: `json` language not bash

requirements/dev.txt
```diff
@@ -8,7 +8,7 @@ pylint==1.7.2
 pytest==3.0.7
 py==1.4.33
 mock==2.0.0
-requests>=2.19.0
+requests==2.19.0
```
sanathkr: No! why!

samcli/commands/local/invoke/cli.py
```diff
@@ -100,3 +100,16 @@ def _get_event(event_file_name):
...
+
+def _get_debug_context(debug_port, debug_args, debugger_path):
+    """
+    Returns a debug context from debug options; Separated out for unit testing.
+    :param int debug_port: Container debug port
+    :param string debug_args: Extra debug arguments for process
+    :param string debugger_path: Path to debugger on host
+    :return DebugContext:
+    """
+    if not debug_port or debug_args or debugger_path:
```
sanathkr: you don't need this method anymore, do you?

samcli/local/docker/lambda_container.py
```diff
@@ -87,12 +90,51 @@ def _get_exposed_ports(debug_port):
...
+        debugger executable into the container.
+        :param DebugContext debug_options: DebugContext for the runtime of the container.
+        :return dict: Dictionary containing volume map passed to container creation.
+        """
+        if not debug_options or not debug_options.debugger_path:
+            return None
+
+        return {
+            debug_options.debugger_path: {
+                "bind": "/tmp/lambci_debug_files",
```
sanathkr: Can you move this static string to a class member? Also add comments explaining if this is the path in local file system or inside docker container

tests/unit/commands/local/invoke/test_cli.py
```diff
@@ -29,10 +31,15 @@ def setUp(self):
...
     @patch("samcli.commands.local.invoke.cli._get_event")
-    def test_cli_must_setup_context_and_invoke(self, get_event_mock, InvokeContextMock):
+    @patch("samcli.commands.local.invoke.cli._get_debug_context")
+    def test_cli_must_setup_context_and_invoke(self, get_debug_context_mock, get_event_mock, InvokeContextMock):
         event_data = "data"
         get_event_mock.return_value = event_data
 
+        debug_context_data = DebugContext(debug_port=self.debug_port, debug_args=self.debug_args,
+                                          debugger_path=self.debugger_path)
+        get_debug_context_mock.return_value = debug_context_data
```
sanathkr: This needs to change because `get_debug_context` method won't exist

tests/unit/commands/local/lib/test_debug_context.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_init(self):
+        context = DebugContext('port', 'debuggerpath', 'debug_args')
+
+        self.assertEquals(context.debug_port, 'port')
+        self.assertEquals(context.debugger_path, 'debuggerpath')
+        self.assertEquals(context.debug_args, 'debug_args')
+
+    @parameterized.expand([
+        ('1000', 'debuggerpath', 'debug_args'),
```
sanathkr: good thought to test with port number as strings :)

tests/unit/local/docker/test_lambda_container.py
```diff
@@ -130,14 +141,62 @@ def test_debug_arg_must_be_split_by_spaces_and_appended_to_entrypoint(self, runt
...
     def test_must_provide_entrypoint_even_without_debug_args(self, runtime):
-
-        result = LambdaContainer._get_entry_point(runtime, self.debug_port)
+        debug_options = DebugContext(debug_port=1235, debug_args=None)
+        result = LambdaContainer._get_entry_point(runtime, debug_options)
 
         self.assertIsNotNone(result)
+
+
+class TestLambdaContainer_get_additional_options(TestCase):
```
sanathkr: thanks!

requirements/dev.txt
jfuss: These are test requirements and without pinning we fail some tests because the request version is in the user-agent headers: https://github.com/awslabs/aws-sam-cli/blob/develop/tests/functional/local/apigw/test_local_apigw_service.py#L629

```
"headers": {"Host": "0.0.0.0:33651", "User-Agent": "python-requests/2.19.0", ...
```

Requests library adds these headers. I had different versions of requests in my local Py3 vs Py2.

samcli/commands/local/invoke/cli.py
jfuss: I do not.

README.rst
jfuss: I can. I still want to keep the /dlv at the end since it needs to be that name (right now).

README.rst
jfuss: done

tests/unit/commands/local/invoke/test_cli.py
jfuss: will update the test where needed.

samcli/local/docker/lambda_container.py
jfuss: Good call!

samcli/local/docker/lambda_container.py
```diff
@@ -87,12 +90,51 @@ def _get_exposed_ports(debug_port):
...
+        Return additional volumes to be mounted in the Docker container. Used by container debug for mapping
+        debugger executable into the container.
+        :param DebugContext debug_options: DebugContext for the runtime of the container.
+        :return dict: Dictionary containing volume map passed to container creation.
+        """
+        if not debug_options or not debug_options.debugger_path:
+            return None
+
+        return {
+            debug_options.debugger_path: {
```
jfuss: One other thing I ran into is if the debugger_path is a relative path, this breaks.

Command:
`sam local start-api -d 5858 --debugger-path ./sam-app`

Failure (start-api works but when invoking I get):
`docker.errors.APIError: 400 Client Error: Bad Request ("create ./sam-app: "./sam-app" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path")`

The right approach here is to use `pathlib.Path(debug_options.debugger_path).resolve()` (encapsulate this inside the DebugContext of course). This will resolve the path to the absolute path. My only concern here is windows and docker integrations. We are already running into path issues with Windows and Docker Toolbox and adding this Path resolving will only make that issue more complex. Short term resolving the path is right but we should really bring all these path resolving through some common library to make our lives a little easier long term.

What are your thoughts here @sanathkr?

requirements/dev.txt
jfuss: I would rather handle this outside of this PR. Pip should not break for pinning (we already do this for pytest, py, mock, etc) in the dev.txt requirements. For developing with the cli, we recommend pyenv to allow multiple python versions for testing. Devs should not be installing this into an existing pyenv. I think the concern here is valid but very low for 'break if we pin'.

sanathkr: Ahh my bad, I thought this was the base requirements.txt. Dev requirements are okay to be pinned

jfuss:
Adding some additional validation on the debugger_path to make for a better experience and fail fast. Hoping to have a new revision here soon

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -268,6 +301,46 @@ def _setup_log_file(log_file):
...
+        ------
+        samcli.commands.local.cli_common.user_exceptions.DebugContext
+            When the debugger_path is not valid
+        """
+        if debug_port and debugger_path:
+            try:
+                debugger = Path(debugger_path).resolve(strict=True)
+            except OSError as error:
+                if error.errno == errno.ENOENT:
+                    raise DebugContextException("'debugger_path' could not be found.")
```
sanathkr: Minor comment: Instead of the string 'debugger_path', can you add the actual value of variable debugger_path  to the message? like '/usr/loca/ is not found'

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -268,6 +301,46 @@ def _setup_log_file(log_file):
...
+                else:
+                    raise error
+
+            # We turn off pylint here due to https://github.com/PyCQA/pylint/issues/1660
+            if not debugger.is_dir():  # pylint: disable=no-member
+                raise DebugContextException("'debugger_path' should be a directory with the debugger in it.")
```
sanathkr: have the actual value of debugger path in error message

jfuss:
@austinlparker Did all the heavy lifting :)

austinlparker:
Y’all too kind. 😛

PR #566, opened 2018-07-19
adds anchor & alias support for yaml parser

shortjared:
*Issue #, if available:*

No issue number specifically, but this has been mentioned before in several cases and a feature I wanted natively in CloudFormation but we aren't going to get anytime soon. (ex: https://github.com/awslabs/serverless-application-model/issues/228)

*Description of changes:*

This simply enhances the yaml parser in exactly the same way I propose the aws-cli does (https://github.com/aws/aws-cli/pull/3462). Mostly just making sure we maintain consistency across the platform.

Important to note, this is doing _nothing_ that nearly all other yaml libs can't do. Take the below template for example and plug it in to http://yaml-online-parser.appspot.com and you'll get perfectly valid json outputs with no additional logic. Of course, you would still need `aws cloudformation package` to populate the `CodeUri` but that is a different process.

```yaml
Transform: 'AWS::Serverless-2016-10-31'
Description: A starter AWS Lambda function.

Metadata:
  Templates:
    Resources:
      Properties: &Props
        Handler: lambda_function.lambda_handler
        Runtime: python3.6
        CodeUri: .
        Description: A starter AWS Lambda function.
        MemorySize: 128
        Timeout: 3

Resources:
  helloworldpython3:
    Type: 'AWS::Serverless::Function'
    Properties:
      <<: *Props

  helloworldnode:
    Type: 'AWS::Serverless::Function'
    Properties:
      <<: *Props
      Runtime: nodejs8.10
```

### DIFF
--- a/samcli/yamlhelper.py
+++ b/samcli/yamlhelper.py
@@ -46,6 +46,17 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
     return {cfntag: value}
 
 
+def _resolve_aliases(dict_to_resolve):
+    """
+    Resolves all anchors in a yaml template
+    :param dict_to_resolve
+    :return:
+    """
+    yaml.SafeDumper.ignore_aliases = lambda *args: True
+    return yaml.safe_load(
+        yaml.safe_dump(dict_to_resolve, default_flow_style=False))
+
+
 def yaml_dump(dict_to_dump):
     """
     Dumps the dictionary as a YAML document
@@ -64,4 +75,4 @@ def yaml_parse(yamlstr):
         return json.loads(yamlstr)
     except ValueError:
         yaml.SafeLoader.add_multi_constructor("!", intrinsics_multi_constructor)
-        return yaml.safe_load(yamlstr)
+        return _resolve_aliases(yaml.safe_load(yamlstr))
--- a/tests/unit/test_yamlhelper.py
+++ b/tests/unit/test_yamlhelper.py
@@ -46,6 +46,29 @@ class TestYaml(TestCase):
         }
     }
 
+    yaml_with_aliases = """
+    Resource:
+        Foobar: &Foobar
+            Qux: Quux
+            Overridable: unknown
+        Baz:
+            <<: *Foobar
+            Overridable: True
+    """
+
+    parsed_yaml_with_aliases = {
+        "Resource": {
+            "Foobar": {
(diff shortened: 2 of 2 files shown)


jfuss:
@shortjared Thanks for the PR but not sure if this should be added. 

Templates that use this would not be deployable directly through CloudFormation and must require this yaml conversion. A SAM template should be deployable through by using `aws cloudformation create-change-set` and `aws cloudformation execute-change-set`, which this will break. I do not think we should deviate from this right now, as it is not clear why one would pass or fail other than using this tool over that.

The example you provided looks reasonable but we already provide a `Globals` section in SAM for this reason. They look like they are solving the same problem and would encourage the use of Globals for this instead.

timoschilling:
@jfuss I see your point but @shortjared has opened the same PR on https://github.com/aws/aws-cli/pull/3462. His example is solvable with globals but there are even use cases which are not.
For example I have a CF Distribution with many Behaviors which are have mostly all the same setting.

jfuss:
@timoschilling Yes the same PR is opened there but my point still holds true. These templates are now not deployable through regular CloudFormation commands. To this point, templates can be used either way and this to SAM and the AWS CLI `cloudformation package` and `cloudformation deploy` commands adds a confusing layer to the already confusing abstraction.

My view is not to accept this change until we work on getting our own package and deploy commands (not shell out to the aws cli).

jfuss:
After discussing with the team, we will not be accepting this change in. The requirement we have is that a template with the Transform of `AWS::Serverless-2016-10-31` needs to be deployable through CloudFormation. This change will break that requirement. 

We do hear the community on the need for this in general. Adding this directly to the [Serverless Application Model](https://github.com/awslabs/serverless-application-model) or using [AWS CloudFormation Macros](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html) might provide a better generic solution for this.

shortjared:
https://github.com/aws/aws-cli/pull/3827

shortjared:
So it looks like AWS CLI has merged YAML anchor unrolling to the CLI in the develop branch. It would be good to make sure that sam cli reflects a similar change. https://github.com/aws/aws-cli/pull/3827

cscetbon:
@shortjared I don't think it's working correctly. We had many issues with anchors. Also it doesn't seem it's supported in samconfig.yaml

PR #579, opened 2018-07-26
Fix stringifying λ environment variables when using Python2

Cliffzz:
*Issue #, if available:*
https://github.com/awslabs/aws-sam-cli/issues/517

*Description of changes:*
Fix stringifying  λ environment variables that are passed when using Python2.

### DIFF
--- a/samcli/local/lambdafn/env_vars.py
+++ b/samcli/local/lambdafn/env_vars.py
@@ -193,6 +193,7 @@ class EnvironmentVariables(object):
             result = "false"
 
         # value is a scalar type like int, str which can be stringified
+        # do not stringify unicode in Py2, Py3 str supports unicode
         elif sys.version_info.major > 2:
             result = str(value)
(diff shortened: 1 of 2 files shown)


jfuss:
@Cliffzz If you add the test case you have in this PR only, the tests pass already in Py2 and Py3. Unless I am missing something else, this is not the solution to that issue.

Cliffzz:
@jfuss Sorry my bad, forgot to actually add a unicode char to the string.
Updated the test so that it fails before the changes.

samcli/local/lambdafn/env_vars.py
```diff
@@ -191,7 +193,11 @@ def _stringify_value(self, value):
             result = "false"
 
         # value is a scalar type like int, str which can be stringified
-        else:
+        elif sys.version_info.major > 2:
```
jfuss: Could you add a comment here on why this is needed. It is not super clear on the reason for this by just looking at the code.

Issue #555, opened 2018-07-15
Can't debug using vscode v1.25, Unable to open 'index.js': File not found 


estyh:
**Description:**

I'm trying to debug a lambda with sam cli. It starts up the docker container, and then waits for me to attach a debugger. When I try to attach, vscode pops up a warning:
```Unable to open 'index.js': File not found (file:///c:/Users/esty.heller/Documents/Dev/benmanagesoftware/lambda/runtime/node_modules/awslambda/index.js).```
And then it just stops and waits, can't do anything except disconnect and let the function run.

Just running the function with sam local invoke works fine.

**Additional environment details (Ex: Windows, Mac, Amazon Linux etc)**
Windows 10, Docker version 18.03.1-ce, build 9ee9f40 (using linux containers), latest vscode

**Output of `sam --version`**:
SAM CLI, version 0.4.0

**Optional Debug logs**:
```
$ sam local invoke -e event-create.json -d 5858 CreateUrl
2018-07-15 11:17:47 Invoking create-url.handler (nodejs8.10)
2018-07-15 11:17:47 Found credentials in shared credentials file: ~/.aws/credentials

Fetching lambci/lambda:nodejs8.10 Docker container image......
2018-07-15 11:17:49 Mounting C:\Users\esty.heller\Documents\Dev\benmanagesoftware\lambda\url-shortener as /var/task:ro inside runtime container
Debugger listening on ws://0.0.0.0:5858/3e08373e-2480-470f-9eed-d7fe2651cae0
For help see https://nodejs.org/en/docs/inspector
Debugger attached.
```

At this point, I try to attach using my launch config:
```
{
			"name": "Attach to Sam Local",
			"type": "node",
			"request": "attach",
			"port": 5858,
			"protocol": "inspector",
			"address": "localhost",
			"localRoot": "${workspaceRoot}/",
			"remoteRoot": "/var/task/"
		}
```
And I get a popup warning ```Unable to open 'index.js': File not found (file:///c:/Users/esty.heller/Documents/Dev/benmanagesoftware/lambda/runtime/node_modules/awslambda/index.js).``` and it won't attach.

Here is my template.yaml
```
AWSTemplateFormatVersion: '2010-09-09' 
Transform: AWS::Serverless-2016-10-31  
 
Globals: 
  Function: 
    Timeout: 300 
    Runtime: nodejs8.10  
    MemorySize: 512 
    
Resources: 
  CreateUrl: 
[...]

PR #598, opened 2018-08-03
Issue #555: Invoke Function Without Parameters (skip STDIN)

agrawroh:
Issue #555 
Invoke Function Without Parameters (skip STDIN)


jfuss:
Thank you for putting the time into this. We appreciate the PR. #604 is the same as this but has done some additional checks. Instead of asking for changes here, I am going to close this in favor of the other.

Issue #556, opened 2018-07-16
Invoke Function Without Parameters (skip STDIN)


adeitcher:
**Description:**

It does not seem that there is a way to invoke a function without passing in some kind of event, even if the lambda Handler takes no arguments, as the lack of params defaults to waiting for input from STDIN. No arg handler functions are valid, at least in Golang (haven't looked at other programming languages)

https://docs.aws.amazon.com/lambda/latest/dg/go-programming-model-handler-types.html

**Steps to reproduce the issue:**
1. invoke function normally `sam local invoke MyFunc`

**Observed result:**
1. With no event, function execution waits for input from stdin
2. With a dummy event e.g. `echo '{}' | sam local invoke MyFunc`, function executes immediately despite not taking any argument params `func handler() error`

**Expected result:**
Possibly pass a command line flag like `--no-args` so that the lambda executes immediately rather than waiting for STDIN

**Additional environment details (Ex: Windows, Mac, Amazon Linux etc)**
OS X 10.13.5 High Sierra

**Output of `sam --version`**:
SAM CLI, version 0.3.0

tebruno99:
This actually would be super useful when developing against SQS.

akkaash:
Looking into this.

jfuss:
Closing this as it was released in [0.6.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.6.0)

PR #604, opened 2018-08-03
Invoke Function Without Parameters

akkaash:
*Issue #556 *

Invoke Function Without Parameters

Adds support to `sam local invoke` to accept no event argument. 

### DIFF
--- a/samcli/commands/local/invoke/cli.py
+++ b/samcli/commands/local/invoke/cli.py
@@ -26,40 +26,47 @@ $ sam local invoke "HelloWorldFunction" -e event.json\n
 Invoking a Lambda function using input from stdin
 $ echo '{"message": "Hey, are you there?" }' | sam local invoke "HelloWorldFunction" \n
 """
+STDIN_FILE_NAME = "-"
 
 
 @click.command("invoke", help=HELP_TEXT, short_help="Invokes a local Lambda function once.")
 @click.option("--event", '-e',
               type=click.Path(),
-              default="-",  # Defaults to stdin
+              default=STDIN_FILE_NAME,  # Defaults to stdin
               help="JSON file containing event data passed to the Lambda function during invoke. If this option "
                    "is not specified, we will default to reading JSON from stdin")
+@click.option("--no-args", is_flag=True, default=False, help="Invoke Function without parameters")
 @invoke_common_options
 @cli_framework_options
 @click.argument('function_identifier', required=False)
 @pass_context
-def cli(ctx, function_identifier, template, event, env_vars, debug_port, debug_args, debugger_path,
+def cli(ctx, function_identifier, template, event, no_args, env_vars, debug_port, debug_args, debugger_path,
         docker_volume_basedir, docker_network, log_file, skip_pull_image, profile):
 
     # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
 
-    do_cli(ctx, function_identifier, template, event, env_vars, debug_port, debug_args, debugger_path,
+    do_cli(ctx, function_identifier, template, event, no_args, env_vars, debug_port, debug_args, debugger_path,
            docker_volume_basedir, docker_network, log_file, skip_pull_image, profile)  # pragma: no cover
 
 
-def do_cli(ctx, function_identifier, template, event, env_vars, debug_port, debug_args,  # pylint: disable=R0914
-           debugger_path, docker_volume_basedir, docker_network, log_file, skip_pull_image, profile):
+def do_cli(ctx, function_identifier, template, event, no_args, env_vars, debug_port,  # pylint: disable=R0914
+           debug_args, debugger_path, docker_volume_basedir, docker_network, log_file, skip_pull_image, profile):
     """
     Implementation of the ``cli`` method, just separated out for unit testing purposes
     """
 
     LOG.debug("local invoke command is called")
 
-    event_data = _get_event(event)
-
     # Pass all inputs to setup necessary context to invoke function locally.
     # Handler exception raised by the processor for invalid args and print errors
     try:
+        if no_args and event == STDIN_FILE_NAME:
+            event_data = "{}"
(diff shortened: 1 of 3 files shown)


samcli/commands/local/invoke/cli.py
```diff
@@ -26,40 +26,47 @@
...
-    event_data = _get_event(event)
-
     # Pass all inputs to setup necessary context to invoke function locally.
     # Handler exception raised by the processor for invalid args and print errors
     try:
+        if no_args and event == STDIN_FILE_NAME:
+            event_data = "{}"
+        elif no_args and event != STDIN_FILE_NAME:
+            # Do not know what the user wants. no_args and event both passed in.
+            raise UserException("no_args and event both given. Do not know what to do")
```
jfuss: Can you update this to some thing of the effect: "no_args and event cannot be used together. Please provide only one."

samcli/commands/local/invoke/cli.py
```diff
@@ -26,40 +26,47 @@
...
 
 
 @click.command("invoke", help=HELP_TEXT, short_help="Invokes a local Lambda function once.")
 @click.option("--event", '-e',
               type=click.Path(),
-              default="-",  # Defaults to stdin
+              default=STDIN_FILE_NAME,  # Defaults to stdin
               help="JSON file containing event data passed to the Lambda function during invoke. If this option "
                    "is not specified, we will default to reading JSON from stdin")
+@click.option("--no-args", is_flag=True, default=False, help="Invoke Function without parameters")
```
jfuss: Thinking about this more, I don't think `--no-args` describes what this does. This is really giving a default value to the event. Maybe something like `--no-event` (so it matches with the rest of the command) and then update the help text to be `Invoke the Function with an empty event dictionary`?

@SoodDhruv Do you have any thoughts here?

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -94,3 +94,14 @@ def test_invoke_when_function_writes_stderr(self):
         process_stderr = b"".join(process.stderr.readlines()).strip()
 
         self.assertIn("Docker Lambda is writing to stderr", process_stderr.decode('utf-8'))
+
+    def test_invoke_raises_exception_with_noargs_and_event(self):
```
jfuss: Can you add a small integ test for providing `--no-agrs` and making sure the function invoked and returns as expected?

samcli/commands/local/invoke/cli.py
SoodDhruv: How about not requiring the --no-args/--no-event option? 'sam local invoke "FunctionName"' could just invoke the function with a default payload.

tests/integration/local/invoke/test_integrations_cli.py
akkaash: Will do.

samcli/commands/local/invoke/cli.py
akkaash: @SoodDhruv `sam local invoke "FunctionName"` currently allows the user to use STDIN to provide the event payload. This would cause a conflict with current behavior. What do you think?
@jfuss I agree. I like `--no-event` better for the option name.

samcli/commands/local/invoke/cli.py
akkaash: Will do.

samcli/commands/local/invoke/cli.py
jfuss: @akkaash That makes sense. I can see an argument for changing this behavior but something maybe we shouldn't do now. Let's just leave this as `--no-event` and then we can see if there is a path/reason to change this behavior to what @SoodDhruv suggested.

akkaash:
@jfuss I've updated the pr. Multiple merge conflicts had to be fixed.

tests/unit/commands/local/invoke/test_cli.py
```diff
@@ -73,6 +77,74 @@ def test_cli_must_setup_context_and_invoke(self, get_event_mock, InvokeContextMo
...
+                       debugger_path=self.debugger_path,
+                       docker_volume_basedir=self.docker_volume_basedir,
+                       docker_network=self.docker_network,
+                       log_file=self.log_file,
+                       skip_pull_image=self.skip_pull_image,
+                       profile=self.profile,
+                       region=self.region)
+
+        msg = str(ex_ctx.exception)
+        print(msg)
```
jfuss: Can we remove this print statement from the tests? Assuming you where using this during testing.

samcli/commands/local/invoke/cli.py
```diff
@@ -26,40 +26,48 @@
...
     """
 
     LOG.debug("local invoke command is called")
 
-    event_data = _get_event(event)
-
     # Pass all inputs to setup necessary context to invoke function locally.
     # Handler exception raised by the processor for invalid args and print errors
     try:
+        if no_event and event == STDIN_FILE_NAME:
```
jfuss: Can we move this out of the try except block like it was before. Having a tight try except is a good practice to follow.

What do you think about update the checks to something like this? I think it is a little easier to reason and read (less mental checks). This is optional as I am ok with what you have as well. Just trying to see if we could simplify these checks in some manner.
```
if no_event and event != STDIN_FILE_NAME:
  raise UserException()

if no_event:
  event_data = "{}"
else:
  event_data = _get_event(event)
```

akkaash: I can see how this is less cognitive overload for a reader. Made the change.

tests/unit/commands/local/invoke/test_cli.py
akkaash: Woops. My bad. I've fixed that.

akkaash:
@jfuss is there anything else I need to do in order to merge this in? 😃

jfuss:
@akkaash Nothing is needed. I had to update the branch and never got back to it after the build passed.

PR #619, opened 2018-08-17
UX: Improves event names to reflect Lambda Event Sources

heitorlessa:
*Issue #, if available:*

*Description of changes:*

This PR changes the local event generation for SQS, CodeCommit, Kinesis, Cloudwatch and SNS to reflect what the event and the event source means to Lambda:

Before | After
------------------------------------------------- | ---------------------------------------------------------------------------------
sam local generate-event sqs sqs | sam local generate-event sqs receive-message
sam local generate-event sns sns | sam local generate-event sns notifications
sam local generate-event kinesis kinesis | sam local generate-event kinsis get-records
sam local generate-event codecommit codecommit | sam local generate-event codecommit repository
sam local generate-event logs cloudwatch-logs | sam local generate-event cloudwatch logs
sam local generate-event events cloudwatch | sam local generate-event cloudwatch scheduled-event


jfuss:
Can we expand the scope here while we are touching this. I just went through all of the events again. Here are some things that could be confusing to customers.

Other duplicates:
* codecommit
* sns
* kinesis

We have two CloudWatch related events (logs and events). Shouldn't this follow the service to type event model we have? So `sam local generate-event event events` and sam local generate-event logs` would be combined into `sam local generate-event cloudwatch` with two sub commands of `scheduled-event` and `logs`?

@SoodDhruv What are your thoughts? We should find better names to the duplicate like @heitorlessa has done with SQS.

heitorlessa:
I agree and that was the confusion when I was presenting this feature in
Turkey Meetups.

I’ll wait for Dhruv’s comment and am happy to expand this to others to
improve UX

SoodDhruv:
Thanks for fixing this, @heitorlessa. I agree with @jfuss. Please increase the scope of this change to cover other services with duplicate commands.

heitorlessa:
@SoodDhruv  and @jfuss  - Done!

PR description updated and table with changes to reflect the new scope

SoodDhruv:
👍

jfuss:
@heitorlessa Should `kinesis get-records` be `kinesis get-record` and `sns notifications` be `sns notification` (singular). There are only one record in each list. This is really nit picky but want to make sure we communicate the right thing here.

heitorlessa:
Good catch - For Kinesis it should be get-records as that’s the API and it
could have multiple records — let me add another record in there

For SNS itself it’s called notifications but I agree that the SNS mechanism
will only send one event at a time (1:1 vs Many:1 in Kinesis).

Let me change that and send a push later tonight

heitorlessa:
@jfuss pushed SNS changes and had a second link at Kinesis and by making this change (2 records) it'd have to increase the scope of parameters to change data + sequence + partition which it'd make it overly complex.

Given that the API Call is [GetRecords](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetRecords.html) and a Kinesis Data Stream can have a single or multiple records I'd vote to keep as-is otherwise we could likely diverge from its essence - Happy to hear otherwise

jfuss:
@heitorlessa Sounds reasonable. Thanks for updating.

PR #633, opened 2018-08-28
fix(path): change windows style path to unix

TheSriram:
- On supplying path for docker in windows environments, change path to unix style.
- This allows sam-cli to work with docker toolkit.

*Issue #, if available:*

- https://github.com/awslabs/aws-sam-cli/issues/461

*Description of changes:*

- change windows style paths to unix style paths, to allow operability of the sam cli under windows 8.
- Did testing on windows environment with docker toolbox and docker for windows.

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -6,6 +6,7 @@ import logging
 import docker
 
 from samcli.local.docker.attach_api import attach
+from .utils import to_posix_path
 
 LOG = logging.getLogger(__name__)
 
@@ -103,6 +104,9 @@ class Container(object):
 
         if self._additional_volumes:
             kwargs["volumes"].update(self._additional_volumes)
+
+        # Make sure all mounts are of posix path style.
+        kwargs["volumes"] = {to_posix_path(host_dir): mount for host_dir, mount in kwargs["volumes"].items()}
 
         if self._env_vars:
             kwargs["environment"] = self._env_vars
--- a/samcli/local/docker/utils.py
+++ b/samcli/local/docker/utils.py
@@ -1,0 +1,36 @@
+"""
+Helper methods that aid with changing the mount path to unix style.
+"""
+
+import os
+import posixpath
+import re
+try:
+    import pathlib
+except ImportError:
+    import pathlib2 as pathlib
+
+
+def to_posix_path(code_path):
+    """
+    Change the code_path to be of unix-style if running on windows when supplied with an absolute windows path.
+
+    Parameters
+    ----------
+    code_path : str
+        Directory in the host operating system that should be mounted within the container.
+    Returns
+    -------
+    str
+        Posix equivalent of absolute windows style path.
+    Examples
+    --------
+    >>> to_posix_path('/Users/UserName/sam-app')
+    /Users/UserName/sam-app
+    >>> to_posix_path('C:\\\\Users\\\\UserName\\\\AppData\\\\Local\\\\Temp\\\\mydir')
+    /c/Users/UserName/AppData/Local/Temp/mydir
+    """
+
+    return re.sub("^([A-Za-z])+:",
+                  lambda match: posixpath.sep + match.group().replace(":", "").lower(),
+                  pathlib.PureWindowsPath(code_path).as_posix()) if os.name == "nt" else code_path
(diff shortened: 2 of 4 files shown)


samcli/local/lambdafn/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    import pathlib
+except ImportError:
+    import pathlib2 as pathlib
+
+from functools import wraps
+
+
+def path_modifier(func):
+    """
+    This method decorator allows conversions of windows style path to unix style path
```
jfuss: Can you update this to use numpy doc strings instead? We are moving off this format onto the numpy format with new docs.

tests/unit/commands/local/lib/test_local_lambda.py
```diff
@@ -260,6 +263,28 @@ def test_must_resolve_absolute_codeuri(self, codeuri):
         self.assertEquals(expected, actual)
         self.assertTrue(os.path.isabs(actual), "Result must be an absolute path")
 
+    @patch("samcli.local.lambdafn.utils.os")
+    def test_must_resolve_windows_path_to_unix_paths(self, os_mock):
+
+        os_mock.name = "nt"
+        with mock.patch("os.path", ntpath):
+            ntpath.isabs = lambda path: False
```
jfuss: ooo... this is interesting, never considered this a way to mock functions out before!

tests/unit/commands/local/lib/test_local_lambda.py
```diff
@@ -260,6 +263,28 @@ def test_must_resolve_absolute_codeuri(self, codeuri):
         self.assertEquals(expected, actual)
         self.assertTrue(os.path.isabs(actual), "Result must be an absolute path")
 
+    @patch("samcli.local.lambdafn.utils.os")
+    def test_must_resolve_windows_path_to_unix_paths(self, os_mock):
+
+        os_mock.name = "nt"
+        with mock.patch("os.path", ntpath):
```
jfuss: This could just be a style difference but why would you do this over mocking out `os.path` through `os_mock.path = mock()`?

samcli/local/docker/container.py
```diff
@@ -53,7 +54,7 @@ def __init__(self,
         self._image = image
         self._cmd = cmd
         self._working_dir = working_dir
-        self._host_dir = host_dir
+        self._host_dir = posix_path(host_dir)
```
sanathkr: Can we set the posix_path on all the volumes that get mounted? Apply to the `kwargs` on the `create()` method. 

When we mount other volumes in the future, this will just transparently work.

samcli/local/docker/utils.py
```diff
@@ -0,0 +1,25 @@
...
+import os
+import posixpath
+import re
+try:
+    import pathlib
+except ImportError:
+    import pathlib2 as pathlib
+
+
+def posix_path(code_path):
```
sanathkr: nit: `posix_path` or `to_posix_path`?

samcli/local/docker/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Change the code_path to be of unix-style if running on windows.
+
+    Parameters
+    ----------
+    code_path : str
+        Directory in the host operating system that should be mounted within the container.
```
sanathkr: Returns?

tests/unit/local/docker/test_container.py
```diff
@@ -46,6 +46,32 @@ def test_init_must_store_all_values(self):
         self.assertEquals(None, container.id)
         self.assertEquals(self.mock_docker_client, container.docker_client)
 
+    @patch("samcli.local.docker.utils.os")
+    def test_init_change_host_dir_unix_style(self, os_mock):
+        os_mock.name = "nt"
```
sanathkr: Can you have a test for the non-nt case?

TheSriram:
Addressed the comments.

samcli/local/docker/utils.py
TheSriram: Added Returns and Examples.

samcli/local/docker/container.py
```diff
@@ -104,6 +105,9 @@ def create(self):
         if self._additional_volumes:
             kwargs["volumes"].update(self._additional_volumes)
 
+        # Make sure all mounts are of posix path style.
```
sanathkr: nice! Thakns

PR #644, opened 2018-09-01
add sam.yml as another default template name

AlJohri:
I expect `sam.yml` to work by default. Countless examples online use this naming convention. for example: https://github.com/awslabs/aws-serverless-java-container/tree/master/samples/spring/pet-store

this PR maintains backward compatibility and allows sam.yml as an additional default

### DIFF
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -5,7 +5,7 @@ Common CLI options for invoke command
 import os
 import click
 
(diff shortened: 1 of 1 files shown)


jfuss:
@AlJohri Thank you for submitting the PR. Unfortunately, we will not be accepting this. Templates can be named however the builder/writer of the template wants to and therefore we can't default to everyone's liking. We have defined the default to be template.[yaml|yml] and any other naming will require the `-t or --template` parameter to be provided by the caller of the command. 

Closing this PR for the reason above.

Issue #629, opened 2018-08-23
AWS CLI not working with SAM CLI 0.5.0


ejhayes:
**Description:**

It appears that the latest version of the AWS CLI is not working with SAM CLI. I had to install AWS CLI version 1.14.46 to get it working with SAM CLI.

**Steps to reproduce the issue:**
When I install the latest version of awscli:
```
$ pip install awscli --user --upgrade
Collecting awscli
  Downloading https://files.pythonhosted.org/packages/4e/cf/0c313db4b8e3b231447d3807657db4f5e7fad26d5eaeb294b3cfa1388a6c/awscli-1.15.84-py2.py3-none-any.whl (1.3MB)
    100% |████████████████████████████████| 1.3MB 4.0MB/s
Requirement already satisfied, skipping upgrade: docutils>=0.10 in /Library/Python/2.7/site-packages (from awscli) (0.14)
Requirement already satisfied, skipping upgrade: rsa<=3.5.0,>=3.1.2 in /Library/Python/2.7/site-packages (from awscli) (3.4.2)
Requirement already satisfied, skipping upgrade: colorama<=0.3.9,>=0.2.5 in /Users/****/Library/Python/2.7/lib/python/site-packages (from awscli) (0.3.7)
Collecting botocore==1.10.83 (from awscli)
  Downloading https://files.pythonhosted.org/packages/b4/fa/b0b08db2c935ae27fb0c88bad73d9a04fee95984a5a616836ae49fc8c445/botocore-1.10.83-py2.py3-none-any.whl (4.5MB)
    100% |████████████████████████████████| 4.5MB 4.0MB/s
Requirement already satisfied, skipping upgrade: PyYAML<=3.13,>=3.10 in /Users/****/Library/Python/2.7/lib/python/site-packages (from awscli) (3.12)
Requirement already satisfied, skipping upgrade: s3transfer<0.2.0,>=0.1.12 in /Users/****/Library/Python/2.7/lib/python/site-packages (from awscli) (0.1.13)
Requirement already satisfied, skipping upgrade: pyasn1>=0.1.3 in /Users/****/Library/Python/2.7/lib/python/site-packages (from rsa<=3.5.0,>=3.1.2->awscli) (0.4.4)
Requirement already satisfied, skipping upgrade: jmespath<1.0.0,>=0.7.1 in /Library/Python/2.7/site-packages (from botocore==1.10.83->awscli) (0.9.3)
[...]

PR #649, opened 2018-09-04
Fix(installation): Bump Boto3 requirement

jfuss:
If you were to upgrade AWS CLI to a version greater than 1.14.46,
then SAM CLI would fail execution of commands due to a botocore
version conflict. This was due to the newer versions of the AWS
CLI bumping boto3 (which in turn bumps botocore). Boto3 has a min
and max version of botocore defined. Since AWS CLI and SAM CLI
had different Boto3 versions, their installations  overrode the
required version for each other causing commands to fail with
DistributionNotFound.

*Issue #, if available:*
#629

*Description of changes:*

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -222,6 +222,18 @@ homebrew and try again:
 
 Once installed then repeat the `Installation process <#windows-linux-macos-with-pip>`_
 
+DistributionNotFound with botocore/boto3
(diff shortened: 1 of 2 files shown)


jfuss:
Closing PR. After discussing with the team, this will end up breaking customers who have not upgraded AWS CLI and moves the bar for entry on installing the SAM CLI higher with a newer version of Boto3, since we are installed globally.

PR #652, opened 2018-09-04
Fix(generate-event): SubCommand help text to be short-text.

jfuss:
The help text for the SubCommands in generate-event were getting truncated
in the terminal. The help text being displayed was the short_help text and not
the full text. If a click command does not have short_text assoicated, click
will truncate the help text to 45 characters.  We will now set the short_text
to be the full text to allow the contents to be displayed on the screen.


TheSriram:
👍

PR #665, opened 2018-09-13
Add dotnetcore 2.1 support for init command

heitorlessa:
*Issue #, if available:* https://github.com/awslabs/aws-sam-cli/issues/609

*Description of changes:*

* Adds support for dotnetcore 2.1 
* Defaults to dotnetcore 2.1 when `dotnet` is selected
* Addresses a regression noted in https://github.com/awslabs/aws-sam-cli/issues/673

### DIFF
--- a/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/src/HelloWorld/aws-lambda-tools-defaults.json
+++ b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/src/HelloWorld/aws-lambda-tools-defaults.json
@@ -8,8 +8,13 @@
   "profile": "default",
   "region": "us-east-2",
   "configuration": "Release",
+  {%- if cookiecutter.runtime == 'dotnetcore2.0' %}
   "framework": "netcoreapp2.0",
   "function-runtime": "dotnetcore2.0",
+  {%- else %}
+  "function-runtime": "dotnetcore2.1",
+  "framework": "netcoreapp2.1",
+  {%- endif %}
   "function-memory-size": 256,
   "function-timeout": 30,
   "function-handler": "HelloWorld::HelloWorld.Function::FunctionHandler"
--- a/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/src/HelloWorld/aws-lambda-tools-defaults.json
+++ b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/src/HelloWorld/aws-lambda-tools-defaults.json
@@ -10,8 +10,8 @@
   "configuration": "Release",
   {%- if cookiecutter.runtime == 'dotnetcore2.0' %}
   "framework": "netcoreapp2.0",
-  "function-runtime": "dotnetcore2.0",
-  {%- else %}
+  "function-runtime": "{{ cookiecutter.runtime }}",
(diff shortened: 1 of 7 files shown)


heitorlessa:
cc @sanathkr @jfuss

gitsno:
If you're making dotnetcore 2.1 the default then I think you also need to change the version in the aws-lambda-tools-defaults.json file in the src/HelloWorld folder.

heitorlessa:
Good catch!! Didn’t know about that.

Will make that change tomorrow

heitorlessa:
@gitsno Change made and I also found another file that needed changing so Visual Studio would pick up the changes too (not only SAM). 

Followed guidelines here: https://docs.microsoft.com/en-us/aspnet/core/migration/20_21?view=aspnetcore-2.1

Installed latest Visual Studio and it recognized the latest and the target framework to 2.1 too.

samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,7 +16,11 @@ Resources:
         Properties:
             CodeUri: src/HelloWorld/bin/Debug/netcoreapp2.0/publish
             Handler: HelloWorld::HelloWorld.Function::FunctionHandler
+            {%- if cookiecutter.runtime == 'dotnetcore2.0' %}
             Runtime: dotnetcore2.0
```
jfuss: Can runtime just be the value of `cookiecutter.runtime`?

samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/src/HelloWorld/aws-lambda-tools-defaults.json
```diff
@@ -8,9 +8,14 @@
   "profile": "default",
   "region": "us-east-2",
   "configuration": "Release",
+  {%- if cookiecutter.runtime == 'dotnetcore2.0' %}
   "framework": "netcoreapp2.0",
   "function-runtime": "dotnetcore2.0",
+  {%- else %}
+  "function-runtime": "dotnetcore2.1",
```
jfuss: If there becomes a dotnetcore2.2, we will default to dotnetcore2.1.

What if we had defaults to set here and then explicitly checked the runtime for each dotnetcore version:

```
if runtime == dotnetcore2.0:
  # stuff for 2.0 version 
elif runtime == dotnetcore2.1:
  # stuff for 2.1 version
else:
  # default

heitorlessa: As of now I don't know of any VS defaults and purely following the migration guidelines. Now that you mentioned, I think it's best to explicit check runtime and remove `else` altogether until there are known defaults

heitorlessa: done

samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/template.yaml
heitorlessa: Done for explicit and valid runtime value

heitorlessa:
@jfuss @sanathkr now all feedbacks are addressed and this also fixes https://github.com/awslabs/aws-sam-cli/issues/673

Issue #626, opened 2018-08-21
Raise error when env vars file is in the wrong format


sthulb:
**Description:**
Currently, the environment variables file expects the format to be in either:
```json
{
"Parameters": { ... }
}
```

or 

```json
{
"FuncName": {...}
}
```

When an env var file is specified in the wrong format, SAM silently ignores the file.

Ideally, SAM should throw an error to make the user aware that the format is incorrect.

**Observed result:**
Provide a badly formatted env var json file, SAM silently ignores the file

**Expected result:**
Provide a badly formatted env var json file, SAM should error stating the file is in the wrong format

Amazon Alias: [redacted]@

sthulb:
Code responsible: https://github.com/awslabs/aws-sam-cli/blob/9359e4994003cb862561caa811eff366e4c7eedf/samcli/commands/local/lib/local_lambda.py#L144

AlJohri:
👍 I was using a format like `{"myvar": "blahblah"}` before checking the docs and couldn't tell why it wasn't working

sanathkr:
@sthulb  Can you provide a PR to raise an error? This should be pretty simple

sthulb:
I’ve got a PR ready to push. I’ll do it a bit later today.

sthulb:
this was fixed in #675

PR #675, opened 2018-09-22
fix: raise error for invalid environment variables file

sthulb:
*Issue #, if available:* #626 

*Description of changes:*
Introduces an exception to be thrown when an environment variable file is invalid. 

SAM still swallows regular malformed env files, but will throw an error if the json structure doesn't match as expected.

### DIFF
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -21,7 +21,7 @@ class LocalLambdaRunner(object):
     """
     PRESENT_DIR = "."
     MAX_DEBUG_TIMEOUT = 36000  # 10 hours in seconds
-
+ 
     def __init__(self,
                  local_runtime,
                  function_provider,
@@ -156,8 +156,12 @@ class LocalLambdaRunner(object):
 
         for env_var_value in self.env_vars_values.values():
             if not isinstance(env_var_value, dict):
-                LOG.debug("Environment variables overrides data is not in a recognized format")
-                raise OverridesNotWellDefinedError("Environment variables overrides data is not in a recognized format")
+                reason = """
+                            Environment variables must be in either CloudFormation parameter file 
+                            format or in {FunctionName: {key:value}} JSON pairs
+                            """
+                LOG.debug(reason)
+                raise OverridesNotWellDefinedError(reason)
 
         if "Parameters" in self.env_vars_values:
             LOG.debug("Environment variables overrides data is in CloudFormation parameter file format")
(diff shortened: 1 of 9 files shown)


samcli/commands/local/lib/local_lambda.py
```diff
@@ -141,6 +142,12 @@ def _make_env_vars(self, function):
         #
         # Standard format is {FunctionName: {key:value}, FunctionName: {key:value}}
         # CloudFormation parameter file is {"Parameters": {key:value}}
+
+        for fn in self.env_vars_values:
+            if not isinstance(self.env_vars_values.get(fn), dict):
```
jfuss: This can be simplified further. You only care about the values of the dictionary so you can iterate over the values of the dictionary and check types on that.

```
for env_var_value in self.env_vars_values.values():
    if not isinstance(env_var_value, dict):
        raise
```

You can also loop over keys and values at the same time by if you really need/want the key for something:

```
for key, value in self.env_vars_values.items():
    pass
```

samcli/commands/local/lib/local_lambda.py
```diff
@@ -141,6 +142,12 @@ def _make_env_vars(self, function):
...
+                LOG.debug("Environment variables overrides data is not in a recognized format")
```
jfuss: Add context on what failed here. Customers can then use `--debug` to help them in understanding the issue. This will also be useful for us when issues come in with `--debug` data to understand where the problem is.

jfuss: @sthulb Can you please address this?

samcli/commands/local/lib/local_lambda.py
```diff
@@ -141,6 +142,12 @@ def _make_env_vars(self, function):
...
+                raise OverridesNotWellDefinedError("Environment variables overrides data is not in a recognized format")
```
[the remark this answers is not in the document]
jfuss: @sthulb Can you please address this: "The message here should say what the recognized format is. Customers should be able to get this error and correct the issue without searching further for what is wrong."

sthulb:
@jfuss I think this is about done now, unless you have other changes

tests/unit/commands/local/start_api/test_cli.py
```diff
@@ -96,6 +97,17 @@ def test_must_raise_user_exception_on_invalid_sam_template(self, invoke_context_
         expected = "bad template"
         self.assertEquals(msg, expected)
 
+    @patch("samcli.commands.local.start_api.cli.InvokeContext")
+    def test_must_raise_user_exception_on_invalid_env_vars(self, invoke_context_mock):
```
jfuss: It's almost like we should have a Test Suite for these kinds of common errors across similar commands. Will make these kinds of tests parametrizable instead of copying across tests. This is out-of-scope for this PR but something I noticed so capturing here.

Issue #676, opened 2018-09-23
Mount options for SELinux


medinarrior:
To expand upon this, if docker is set to run with the --selinux-enabled flag in /etc/sysconfig/docker, the (lack of) SELinux labeling on the host directory being passed to the container as a volume will cause the underlying docker container to be unable to read the files.

You can address this without disabling SELinux protection in Docker, however! To work around this issue on SELinux-enabled systems, tag your lambda's working directory on the docker host with the svirt_sandbox_file_t tag:
`chcon -Rt svirt_sandbox_file_t .` This should make `aws-sam-local start-api` work.

_Originally posted by @jmthomas-mitre in https://github.com/awslabs/aws-sam-cli/issues/208#issuecomment-374715877_

m1keil:
I just spent an hour sort of bashing my head with `sam local invoke` issue that wasn't working. The underlying issue was of course selinux.

I stumbled across [this blog post](https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/) from project atomic and it seems like by using z or Z mounting options might be able to help. Note that I'm not selinux expert. I'm just using it because it comes pre-installed on Fedora.

Before:
```
$ docker run --rm -it -v /home/m1ke/src/sam-app/hello-world:/var/task --entrypoint ls lambci/lambda:python3.7 
ls: cannot open directory .: Permission denied
```

After:
```
$ docker run --rm -it -v /home/m1ke/src/sam-app/hello-world:/var/task:z --entrypoint ls lambci/lambda:python3.7 
app.js	package.json  tests


$ docker run --rm -it -v /home/m1ke/src/sam-app/hello-world:/var/task:Z --entrypoint ls lambci/lambda:python3.7 
app.js	package.json  tests

```

morefaey:
@medinarrior this should not be relevant to the current version of sam cli. if you think it is not the case, please reopen.

PR #677, opened 2018-09-23
the bind mount content is private and unshared

medinarrior:
Issue #676 

*Description of changes:*

The Z option indicates that the bind mount content is private and unshared.


jfuss:
These seems pretty risky to me. I don't fully understand this change either and how it impacts all OS's.

From [here](https://stackoverflow.com/a/35222815):
" Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand."

sanathkr:
This is a risky change. The risk is not worth the benefits. A better solution will be to actually copy files over instead of bind mounting. #976 is implementing a copy functionality to support Golang builds.

PR #679, opened 2018-09-24
feat: travis CI support for Python 3.7

TheSriram:
* Referencing https://github.com/travis-ci/travis-ci/issues/9815
* listed as a current workaround until travis CI comes up with official support for 3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,13 @@ python:
   - "2.7"
   - "3.6"
 
+# Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
+matrix:
+  include:
+    - python: 3.7
+      dist: xenial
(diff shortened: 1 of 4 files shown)


jfuss:
@TheSriram Do we need to be updating `tox.ini` or `DEVELOPMENT_GUIDE.rst` with Py37 things. 

If we are supporting 3.7, we should add the classifier in the `classifiers` dict of setup.py.

sanathkr:
@jfuss  Yes I think we should. They are small enough that we can bundle it in this PR

TheSriram:
we could, I had to do some wrangling with symlinks to get `tox -e py36` working after adding py37 to tox.ini, but that could be because of brew's python getting updated to 3.7.0 by default.

PR #687, opened 2018-10-01
fix: fix validate command error reporting for python 2

brettstack:
*Issue #, if available:*

*Description of changes:*
The errors were not showing when running in Python 2. Changed it to explicitly print the exception's `message` property.

### DIFF
--- a/samcli/commands/validate/lib/sam_template_validator.py
+++ b/samcli/commands/validate/lib/sam_template_validator.py
@@ -84,7 +84,7 @@ class SamTemplateValidator(object):
                 LOG.debug("Translated template is:\n%s", yaml_dump(template))
         except InvalidDocumentException as e:
             raise InvalidSamDocumentException(
-                functools.reduce(lambda message, error: message + ' ' + str(error), e.causes, str(e)))
+                functools.reduce(lambda message, error: message + ' ' + str(error.message), e.causes, str(e.message)))
 
     def _replace_local_codeuri(self):
         """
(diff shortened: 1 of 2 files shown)


samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -84,7 +84,7 @@ def warning(self, message):
                 LOG.debug("Translated template is:\n%s", yaml_dump(template))
         except InvalidDocumentException as e:
             raise InvalidSamDocumentException(
-                functools.reduce(lambda message, error: message + ' ' + str(error), e.causes, str(e)))
+                functools.reduce(lambda message, error: message + ' ' + str(error.message), e.causes, str(e.message)))
```
jfuss: This is kind of a gross way to surface errors to customers. This is not your fault but rather just noticing it now. 

I feel that InvalidDocumentException and/or InvalidResourceException should implement the `__str__` method so anyone using SAM doesn't have to do this weird reducing and would standardize the errors customers see from SAM. 

Thoughts?

brettstack: I agree for a long term solution. Think we should merge this in now though to fix Py2 and change after we get it fixed and released in SAM.

jfuss: Can we get a issue tracking that update then? 

Also do we need updates anywhere else? We are doing something similar here: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/samlib/wrapper.py#L70 We should make all these consistent until we get the 'long term solution' out. There might be others spread in the code base, a quick grep should surface them.

jfuss: @brettstack Can you link the long term solution issue here? Also, please rebase.

sanathkr:
SAM uses a non-standard exception structure (with `.message` property). Its better if we don't implement the change here and instead implement in SAM library and when Py2 is deprecated we don't have to worry about this

PR #692, opened 2018-10-04
Change cookiecutter URL to point at correct github repo

stealthycoin:
Removed the `hello` in the github reference since that repo does not exist.

### DIFF
--- a/samcli/local/init/templates/cookiecutter-aws-sam-hello-python/README.md
+++ b/samcli/local/init/templates/cookiecutter-aws-sam-hello-python/README.md
@@ -6,7 +6,7 @@ It is important to note that you should not try to `git clone` this project but 
 
 ## Requirements
 
-Install `cookiecutter` command line: 
+Install `cookiecutter` command line:
 
 **Pip users**:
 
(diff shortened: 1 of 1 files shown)


stealthycoin:
In any case the one I was changing it from is a 404: https://github.com/aws-samples/cookiecutter-aws-sam-hello-python. Someone else had reported that the SAM CLI  when used to run the template didn't set up the testing requirements properly (pytest etc) so I was trying out the cookiecutter template directly to see if it generated them. If it was actually a different template that I wound up running, it would make sense why it had the test dependencies declared and the other one didn't.

jfuss:
That github doesn't exist and we should fix this doc but we can't change that reference the one you have since it is completely different. 

My point is that I think we should just remove those lines you changed but wanted to understand why it is the way it is now before moving forward.

heitorlessa:
@jfuss This is an unpaid tech debt. README comes from unpublished repo before sam init was created; we're creating dedicated repo for each runtime as initially intended and moving these outside this repo. 

What customers should be reading today is not this README but the README that is generated as part of the template as these are raw files.

jfuss:
@heitorlessa That doesn't mean we should keep it either though. If it does not provide any value, we should just remove it. Thoughts?

heitorlessa:
I agree on removing it as it only confuses those who look up at the source
thinking this would provide any value

jfuss:
@heitorlessa Are you talking these types of things in with your rework of all the init examples? Do we still need this if that is the case?

jfuss:
@stealthycoin Thanks for the PR. We addressed this in a different PR (#849) while redoing some of the init examples. Closing this as it is no longer needed.

PR #694, opened 2018-10-05
Explain how to debug Python functions using VS Code

heitorlessa:
*Issue #, if available:* https://github.com/awslabs/aws-sam-cli/issues/54 https://github.com/awslabs/aws-sam-cli/issues/172 

*Description of changes:*

* [x] Reformats **Debugging Python functions** to explain the 2-step process required by SAM CLI users. 
* [x] Includes **Visual Studio Code** (VS Code) Debugger configuration

**Special Thanks** to [Sriram](https://twitter.com/TheSrirama) for discovering that

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -339,8 +339,22 @@ Python debugging requires you to enable remote debugging in your Lambda function
 2. Configure your IDE to connect to the debugger you configured for your function
   - SAM CLI debug port option ``--debug-port`` or `-d` will map that port to the local Lambda container execution your IDE needs to connect to
 
+As this may be your first time using SAM CLI, let's start with a boilerplate Python app and install both app's dependencies and ptvsd:
+
+.. code:: bash
+
+    sam init --runtime python3.6 --name python-debugging
+    cd python-debugging/
+    # Install dependencies of our boilerplate app
+    pip install -r requirements.txt -t hello_world/build/
+    # Install ptvsd library for step through debugging
+    pip install ptvsd -t hello_world/build/
+    cp hello_world/app.py hello_world/build/
+
 Ptvsd configuration
 ^^^^^^^^^^^^^^^^^^^
+
+As we installed ptvsd library in the previous step, we need to enable ptvsd within our code, therefore open up `hello_world/build/app.py` and add the following ptvsd specifics:
 
 Provided that your source code has ptvsd available for your Lambda function you can do the following to enable the debugger:
 
@@ -351,6 +365,7 @@ Provided that your source code has ptvsd available for your Lambda function you 
 
   def lambda_handler(event, context):
     ...
+    # Enable ptvsd on 0.0.0.0 address and on port 5890 that we'll connect later with our IDE
     ptvsd.enable_attach(address=('0.0.0.0', 5890), redirect_output=True)
     ptvsd.wait_for_attach()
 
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -333,25 +333,74 @@ configuration.
...
+              "name": "SAM CLI Python Hello World",
+              "type": "python",
+              "request": "attach",
+              "port": 5890,
+              "host": "localhost",
+              "pathMappings": [
+                  {
+                      "localRoot": "${workspaceFolder}/hello_world/build",
+                      "remoteRoot": "/var/task"
+                      // "stopOnEntry": true,
```
jfuss: This is not a valid key here right? Can this just be removed?

For python, I haven't seen this stop on entry at all.

TheSriram: Yeah this can be removed, its used in node.

heitorlessa: Great catch and it was in fact commented out. Removed it

docs/usage.rst
```diff
@@ -333,25 +333,73 @@ configuration.
...
+Provided that your source code has ptvsd available for your Lambda function you can do the following to enable the debugger:
+
+.. code:: python
+
+  import ptvsd
+  ...
+
+  def lambda_handler(event, context):
+    ...
+    ptvsd.enable_attach(address=('0.0.0.0', 5890), redirect_output=True)
```
jfuss: From VSCode's Remote Python debugging [docs](https://code.visualstudio.com/docs/python/debugging): 
```
Local computer: set a breakpoint in the code where you want to start debugging.

Tip: setting a single breakpoint on the statement immediately following the ptvsd.wait_for_attach() line may not work. Set at least one other breakpoint on another statement.
```

Putting this into the function code risks some customer confusion on debugging not working correctly. Can we move this up to where ptvsd is imported?

docs/usage.rst
```diff
@@ -333,25 +333,73 @@ configuration.
...
+       "configurations": [
+           {
+              "name": "SAM CLI Python Hello World",
+              "type": "python",
+              "request": "attach",
+              "port": 5890,
+              "host": "localhost",
+              "pathMappings": [
+                  {
+                      "localRoot": "${workspaceFolder}/hello_world/build",
```
jfuss: You state to use this launch config to get debugging working but this localRoot is specific to our init example. Without understanding how to set this, customers that are not using the hello world init example may have difficulty.

My suggestion is to make explicit steps to help customers. Something along the lines:

There are X steps to get Python Debugging working with VSCode.

1. Install ptvsd into your project. This should be part of your dependencies that are mounted into the container.
`pip install ptvsd -t <path to project>` You can also add it as a 'dev dependency' and install them into the project folder like you would normally.
2. Open your app in vscode.
3. Copy 'this mostly complete' launch config to your launch config for your project.
4. Update 'configurations.pathMappings.localRoot' value to be ${workspaceFolder}/<path to the function code>
Note: the <path to the function code> should be the path to the folder that will be mounted to the docker container. If you have opened vscode in your project that has your `template.yml` then <path to the function code> should match your `CodeUri` on your function
5. Run sam local:
6. Launch the debugger

I think this gives a better way to understand the steps and make them very reproducible.

If you think the current structure is better, I ask to make it clear what needs to change and how to change it to get your function debuggable instead of saying this config will just make it work.

heitorlessa: Fair point. Reworded as a walkthrough style and used `sam init --runtime python3.6..` as a starting point and provided explicit step-by-step from there.

docs/usage.rst
heitorlessa: Just tested and it also works in the global space. Fixed:

```python
import ptvsd

# Enable ptvsd on 0.0.0.0 address and on port 5890 that we'll connect later with our IDE
ptvsd.enable_attach(address=('0.0.0.0', 5890), redirect_output=True)
ptvsd.wait_for_attach()
```

docs/usage.rst
heitorlessa: * Pushed extended wording explaining step by step
* Moved generic wording under Debugging functions section as they're the same for other runtimes (explanation about `-d, --debug-port`). 

This line and onwards expand on the additional ask: https://github.com/awslabs/aws-sam-cli/pull/694/files#diff-6b8768d928ef9ab4f13707da19516369R339

PR #699, opened 2018-10-08
Updated isBase64Encoded value to bool as defined in documentation

sliedig:
*Issue #, if available:* N/A

*Description of changes:* Documentation states that the isBase64Encoded key is a bool value (see [https://amzn.to/2QAWECb](https://amzn.to/2QAWECb)). The aws-proxy generated sample event defines it as a string. This is causing type casting issues in languages like Go.
 
Also, fixed integration test assertion for "test_lambda_function_raised_error"

### DIFF
--- a/samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
+++ b/samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
@@ -3,7 +3,7 @@
   "resource": "{{resource}}",
   "path": "/{{path}}",
   "httpMethod": "{{method}}",
-  "isBase64Encoded": false,
+  "isBase64Encoded": true,
   "queryStringParameters": {
     "foo": "bar"
   },
--- a/samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
+++ b/samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
@@ -3,7 +3,7 @@
   "resource": "{{resource}}",
   "path": "/{{path}}",
   "httpMethod": "{{method}}",
-  "isBase64Encoded": "false",
+  "isBase64Encoded": false,
   "queryStringParameters": {
     "foo": "bar"
   },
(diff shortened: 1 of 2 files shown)


samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/README.md
```diff
@@ -6,8 +6,9 @@ This is a sample template for {{ cookiecutter.project_name }}
 
 * AWS CLI already configured with at least PowerUser permission
 * [Docker installed](https://www.docker.com/community-edition)
-* [SAM Local installed](https://github.com/awslabs/aws-sam-cli)
+* [SAM CLI installed](https://github.com/awslabs/aws-sam-cli)
 * [DotNet Core installed](https://www.microsoft.com/net/download)
+* [.NET Core Global Tools for AWS](https://aws.amazon.com/blogs/developer/net-core-global-tools-for-aws/) See appendix for setup instructions.
```
jfuss: Is this being used by this init project? 

Also this is an increase in scope to this PR. Please remove the dotnet changes and add them in a separate PR.

samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
```diff
@@ -3,7 +3,7 @@
   "resource": "{{resource}}",
   "path": "/{{path}}",
   "httpMethod": "{{method}}",
-  "isBase64Encoded": "false",
+  "isBase64Encoded": false,
```
jfuss: This should actually be `true` since SAM CLI does convert the data to Base64

samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/README.md
sliedig: Opened #705

samcli/commands/local/lib/generated_sample_events/events/apigateway/AwsProxy.json
jfuss: The value should reflect the actual event. And in this case, that means this field should be a bool and `true` since the event data is base64 encoded.

PR #704, opened 2018-10-10
fix: correct launch.json for node js debugging through vs code

TheSriram:
*Issue #, if available:*

https://github.com/awslabs/aws-sam-cli/issues/555

*Description of changes:*

stopOnEntry flag allows breaking directly into user code, instead of the runtime bootstrap file.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -304,33 +304,39 @@ Visual Studio Code:
    SAM Local debugging example
 
 In order to setup Visual Studio Code for debugging with AWS SAM CLI, use
-the following launch configuration:
+the following launch configuration after setting directory where the template.yaml is present
+as workspace root in Visual Studio Code:
 
 .. code:: json
 
-	{
-		"version": "0.2.0",
-		"configurations": [{
-	        "name": "Attach to SAM CLI",
-	        "type": "node",
-	        "request": "attach",
-	        "address": "localhost",
-	        "port": 5858,
-	        "localRoot": "${workspaceRoot}",
-	        "sourceMapPathOverrides": {
-	            "${workspaceRoot}/runtime": "${remoteRoot}"
-			    },
-	        "remoteRoot": "/var/task",
-	        "protocol": "inspector",
-	        "stopOnEntry": false
-		}]
-	}
+  {
+       "version": "0.2.0",
+       "configurations": [
+           {
+               "name": "Attach to SAM CLI",
+               "type": "node",
+               "request": "attach",
+               "address": "localhost",
+               "port": 5858,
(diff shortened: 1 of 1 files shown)


heitorlessa:
AWESOME @TheSriram ;) 

Would you be able to update the PR to point out that this will work if the source code is in the root (where template.yaml is). However, if they have functions inside folders (i.e. function1/app.js, function2/app.js) then they'll need to specify that in `localroot` too. 

Here's an example to debug NodeJS when using the boilerplate provided by SAM init:

```bash
sam init
cd sam-app
# Brings Visual Studio code for this app
code . 
```

Creates a new debugger configuration using the following:

```json
```

docs/usage.rst
```diff
@@ -308,21 +308,23 @@ the following launch configuration:
...
+		"configurations": [{
+	        "name": "Attach to SAM CLI",
+	        "type": "node",
+	        "request": "attach",
+	        "address": "localhost",
+	        "port": 5858,
+	        "localRoot": "${workspaceRoot}",
+	        "sourceMapPathOverrides": {
+	            "${workspaceRoot}/runtime": "${remoteRoot}"
+			    },
```
jfuss: Can you format the JSON here?

docs/usage.rst
```diff
@@ -308,21 +308,23 @@ the following launch configuration:
...
+	        "remoteRoot": "/var/task",
+	        "protocol": "inspector",
```
jfuss: The previous protocol was legacy. What does updating this to inspector indicate?

docs/usage.rst
```diff
@@ -308,21 +308,23 @@ the following launch configuration:
...
-   }
+	{
+		"version": "0.2.0",
+		"configurations": [{
+	        "name": "Attach to SAM CLI",
+	        "type": "node",
+	        "request": "attach",
+	        "address": "localhost",
+	        "port": 5858,
+	        "localRoot": "${workspaceRoot}",
```
jfuss: This should be the ${workspaceRoot}/<path to the function code dir> right? This localRoot only works if you open the directory of the function code and not the whole project.

TheSriram:
``
At what point does the previous config we had not work? Can we clearly document which version(s) of VSCode the config you added supports? We should probably keep the one one around too and say use that config if your VSCode version is lower than X?
``

It's a change in the way the first file the debugger tries to break into. we break into the code_uri directly with the stopOnEntry flag. Also this flag seems to exist from 2016, so it should definitely be backwards compatible. Yes this is VS code v1.27.2, I can add the version this was tested on.

PR #705, opened 2018-10-10
Updates to .NET Core init template

sliedig:
*Description of changes:* 

1. Removed Item Group to include DotNetCliToolReference 
in project file. This package provided integration with the .NET Core CLI tools for Lambda which is now being made available in the .NET Core Global Tools for AWS. This has new approach does not require the project to take a dependency on the Amazon.Lambda.Tools nuget package.

2. Update README.md for init project to include instructions on how to install the Global CLI tools.
 
3. Updated packages Amazon.Lambda.APIGatewayEvents and Amazon.Lambda.Serialization.Json to latest versions.


samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/{{cookiecutter.project_name}}/README.md
```diff
@@ -136,6 +137,16 @@ aws cloudformation describe-stacks \
     --stack-name {{ cookiecutter.project_name.lower().replace(' ', '-') }} --query 'Stacks[].Outputs'
 ```
 
+## Installing .NET Core Global Tools for AWS
+
+To install a Global Tool, use the dotnet command line. Here is an example of installing the Lambda Global Tool.
```
jfuss: The blog you link to states this is new for .NET Core 2.1. These examples also include .Net Core 2.0 and .Net Core 1.0. Is the Global Tools supported for those versions as well?

sliedig: Yes, you do require version 2.1 to install the tools, but they can be used to deploy .NET Core 1.0 and 2.0 projects. The `DotNetCliToolReference` in the csproj should be using version 2.2.0

Having said that, if you look at the README file in the root directory, all the instructions for building and packaging the project using Cake (https://cakebuild.net).  The AWS Lambda tools are actually not being used, so removing the `DotNetCliToolReference` won’t adversely affect the application. Also means there is no need to install the Global tools so we could probably remove that.

The README file in the function directory has references to Visual Studio and build, test and deploy commands that reference the dotnet core and lambda cli tools and probably should be removed, along with the `aws-lambda-tools-defaults.json` file which is only used by Visual Studio.

jfuss: We shouldn't be removing the README or `aws-lambda-tools-defaults.json`. They are both used to help customers in the next steps after they create the project. This is how all of our init examples are formatted and we shouldn't break that here.

sliedig: @jfuss  here is my reasoning for removing the files.

The README file was removed primarily because its straight from the Visual Studio template generated project and contains commands to build the project that contract the README in the root directory by making references to CLI tools that are not being used by the project. The primary build mechanism is Cake Build not the AWS CLI extensions for .NET Core.

If you want to maintain the README file at the project/function level to maintain consistency with the rest of the projects, then that's cool, I think we need to come up with something different that what is there at the moment.

I think the changes to the root README actually explain the .NET Extensions option reasonably well, making people aware that if they want to import the SAM generated project into Visual Studio, then they have native tool support there too.

The `aws-lambda-tools-defaults.json` file was removed, because it is only used by Visual Studio, to remember/pre-populate the dialog boxes that come up when you "right-click > Publish to AWS Lambda". If you deleted this file in Visual Studio, it would only recreate it the next time you deployed it from the IDE, so the risk to SAM init project is minimal. Also it has a dependency on the AWS Extensions, which the SAM init project isn't using.

So this is just an attempt to simplify the solution and make it easier to use the project, rather than sending mixed messages about the build automation tools they should be using.

jfuss:
@sliedig Could you rebase with the HEAD of develop. I have approved the PR. Once that is done, I can merge this PR.

Thanks for going through the iterations.

sliedig:
Merged. Thanks @jfuss

PR #710, opened 2018-10-16
WIP: Adding AppVeyor config 

sanathkr:
*Issue #, if available:*

*Description of changes:*
Allows us to run integration tests that use Docker because AppVeyor supports running Docker, which Travis doesn't. We should also consider porting all Travis steps to AppVeyor so we can maintain one system. 

Also, AppVeyor supports Windows but doesn't support running linux containers inside Windows. So even though we can't run all integ tests on Windows, we should be able to run a subset of them.

AppVeyor comes built-in with multiple Python versions (incl 3.7) so we can test on a matrix of Py versions.

As I am working through the config, see https://ci.appveyor.com/project/sanathkr/aws-sam-cli for latest builds.

Remaining items:
- [ ] Make all integ tests work
- [ ] Run on a matrix of Python versions (2.7, 3.6 and 3.7)
- [ ] Run unit tests on Windows

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,7 @@
 version: 1.0.{build}
 image: Visual Studio 2017
+install:
+- docker version
 build_script:
-- docker run hello-world:linux
+- docker run hello-world
+- docker image hello-world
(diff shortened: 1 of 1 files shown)


sanathkr:
Not needed because Travis was updated to run integ tests

PR #722, opened 2018-10-22
chore: Enable Travis to run integ tests

jfuss:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #729, opened 2018-10-25
Flush streams after lambda function invocation

gyfis:
*Description of changes:*
The output from a single function invocation doesn't get necessarily written out due to buffering, and to get the output, you need to invoke the function again. 

I added conditional `stdout.flush()` and `stderr.flush()` to the `finally` block of the `invoke` function of the `LambdaRuntime`.

### DIFF
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -108,6 +108,12 @@ class LambdaRuntime(object):
                     timer.cancel()
                 self._container_manager.stop(container)
 
+                # Flush stdout and stderr streams so that logs from the invocation are written out
+                if stdout:
+                    stdout.flush()
+                if stderr:
+                    stderr.flush()
+
     def _configure_interrupt(self, function_name, timeout, container, is_debugging):
         """
         When a Lambda function is executing, we setup certain interrupt handlers to stop the execution.
--- a/tests/unit/local/lambdafn/test_runtime.py
+++ b/tests/unit/local/lambdafn/test_runtime.py
@@ -34,8 +34,8 @@ class LambdaRuntime_invoke(TestCase):
     def test_must_run_container_and_wait_for_logs(self, LambdaContainerMock):
         event = "event"
         code_dir = "some code dir"
-        stdout = "stdout"
-        stderr = "stderr"
+        stdout = Mock()
+        stderr = Mock()
         container = Mock()
         timer = Mock()
         debug_options = Mock()
@@ -81,12 +81,16 @@ class LambdaRuntime_invoke(TestCase):
         timer.cancel.assert_called_with()
         self.manager_mock.stop.assert_called_with(container)
 
+        # Flush streams
+        stdout.flush.assert_called()
+        stderr.flush.assert_called()
+
     @patch("samcli.local.lambdafn.runtime.LambdaContainer")
     def test_exception_from_run_must_trigger_cleanup(self, LambdaContainerMock):
         event = "event"
         code_dir = "some code dir"
-        stdout = "stdout"
-        stderr = "stderr"
+        stdout = Mock()
+        stderr = Mock()
         container = Mock()
         timer = Mock()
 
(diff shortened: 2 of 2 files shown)


samcli/local/lambdafn/runtime.py
```diff
@@ -108,6 +108,12 @@ def invoke(self,
                     timer.cancel()
                 self._container_manager.stop(container)
 
+                # Flush stdout and stderr streams so that logs from the invocation are written out
+                if stdout:
```
jfuss: A better place for this would be when we write the logs to stdout and stderr. This can be found here: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/docker/container.py#L203

Placing the flushing here makes the code a little hard to maintain and reason about, since the `.flush` happens in a different place then the `.write`.

gyfis:
Hi, I’ll try to come up with a test case to test this; it happened to me when I had a lot of long logs (e.g. base64 images), some just didn’t appear in the file until I invoked the function again. I like the proposed place better, thanks for the tip!

sanathkr:
@gyfis Are you using a Python Lambda function? Is this happening only during debugging or in normal cases too? If so, you can invoke with `PYTHONBUFFERED=0` so the function doesn't buffer.  

`flush()` is super expensive and we don't wan to to be doing them on every write. So trying to understand if there we can reduce the scope of the flush

sanathkr:
(sorry closed by accident)

jfuss:
@sanathkr my thought was to move this into the container.py and only do the ‘flush’ after we exhaust the output iterator. But having a test that we can reproduce this with would make understanding the problem easier.

gyfis:
I'm trying to modify the functional test `test_function_result_is_available_in_stdout_and_logs_in_stderr` in `tests/functional/local/docker/test_lambda_container.py`, however the functional tests (`make func-test`) don't seem to be passing at all. Also, I don't see them included in the `.travis.yml` file, CI builds, or mentioned in the Contributing guide; is this intentional? Do you have any tips for running those tests locally?

I also don't see the `make integ-test` in the `.travis.yml` file, but that one _is_ included when listing the `develop` travis config file on site, so I guess that's some config on travis-ci.org.

---

On an unrelated note, I see you're calling `make init` in the `install` phase and `make pr` in the `script` phase; since `make pr` also does `make init`, this could be improved (e.g. by replacing `make pr` by `make dev` in the `.travis.yml` file). If you want, I can create an issue for this for separate tracking.

jfuss:
@gyfis `make func-test` should be passing. I know they were run in #740. What is the failure?

CI builds is intentional at the moment. I attempted to add running functional tests in #722 but ran into problems. We generate the function code in our functional tests and in travis I was seeing the code not being mounted into the container. This resulted in majority of the tests failing. It was hard to narrow down the cause of this and opted to get the integration tests running in travis as a first step.

The Contributing guide does state: "5. Does the PR have updated/added unit, functional, and integration tests?". While having the commands to run would be better, we do say to make sure all tests are updated. Happy to take an update to the Guide to make things clearer.

`make integ-test` does run in Travis as of #722. Updating your branch will pick up the new Travis config.

TheSriram:
@gyfis just curious are you seeing this on a normal invoke, or while debugging, we had a fix to make sure our streams were unbuffered if we are in debugging mode.

https://github.com/awslabs/aws-sam-cli/commit/34c7fb06cbcc5c669e1702f3b9db5d6e3fbb27f5

ndobryanskyy:
I had came across this issue again [here](https://github.com/awslabs/aws-sam-cli/pull/759). In python 3.7 this environment variable just stopped working. According to this [link](https://docs.python.org/3/using/cmdline.html#cmdoption-u) something has changed for this version of python. 

My suggestion is to `flush()` after every write in debug configuration. Or better yet, use some abstraction above the `stderr` and `stdout`, which will facade write implementation and adopt to the configuration.

gyfis:
Hi everyone, sorry I didn't get to this issue just yet!

ndobryanskyy:
@gyfis Hello, I found that this issue majorly affects debugging and I have some thoughts of how to fix it. Do you mind me if I do take over this issue?

gyfis:
@ndobryanskyy Oh my! By all means, go for it! 👏

ndobryanskyy:
Okay, I've got it fixed in my local branch already, I will file up a PR tomorrow as I fix and extend tests

ndobryanskyy:
@gyfis , I've opened #843 I'd appreciate if you take a look, thanks in advance

gyfis:
With the work @ndobryanskyy is doing in #834, let’s close this.

Issue #735, opened 2018-10-29
[Feature Request] Multi-Value Headers and Query String Parameters


charsleysa:
**Description:**
API Gateway has recently (Oct 4th post in AWS What's New) released support for multi-value headers and query string parameters. It would be awesome if SAM CLI could support this too.

Here's the relevant doc https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#apigateway-multivalue-headers-and-parameters

jfuss:
@charsleysa Thanks for the request. @medinarrior has already got a PR out for it :)

Updating labels.

lkonya:
@jfuss Hi! Any update on this one?

jasonrberk:
Status?  I need query params in my Sam template

jfuss:
Status: Waiting for release

You can install from head (brew install aws-sam-cli --head), build from source, or wait for the next release.

jfuss:
This was released in [v0.18.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.18.0)

Closing

PR #741, opened 2018-10-30
Support for Multi-Value Headers and Query String Parameters

medinarrior:
*Issue #735*

*Description of changes:* Adding support for multi-value query string parameters.


sanathkr:
@jfuss already started taking a peek at this PR. I will assign to him for review

samcli/local/apigw/local_apigw_service.py
```diff
@@ -287,7 +287,9 @@ def _construct_event(flask_request, port, binary_types):
                                       resource=endpoint,
                                       request_context=context,
                                       query_string_params=query_string_dict,
+                                      multi_value_query_string_params=flask_request.args.to_dict(False),
```
jfuss: Can you add the key to the arg here? It will make this line easier to ready/understand.

Just thinking (typing) out load here. We currently have a function to get the Query strings (`LocalApigwService._query_string_params()`). This function is already iterating over the `request.args`. We can adjust this slight to either:
* Pass in the `flask_request.args.to_dict(flat=False)` and iterate over that.
* Keep the parameter the same (`flask_request`) and return two values (`multi_value_query_string_params` and `query_string_params`.

I am ok with either. I think this might help keep all the query string related things into one place, is my thought. 

Also, remove the comment on like 280-282. This is no longer relevant. - Github won't let me comment on those lines for whatever reason :(.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -327,6 +329,29 @@ def _query_string_params(flask_request):
...
+            Request Headers
+
+        Returns dict (str: list of str)
+        -------
+            Returns a dictionary of key to list of strings
+
+        """
+        header_dict = {}
+
+        # Multi-value request headers is not really supported by Flask.
```
jfuss: [This](https://github.com/pallets/werkzeug/blob/master/werkzeug/datastructures.py#L910) indicates otherwise. Are we sure it is not supported?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -327,6 +329,29 @@ def _query_string_params(flask_request):
...
+        # See https://github.com/pallets/flask/issues/850
+        for header_key, header_value in event_headers.items():
```
jfuss: Can we iterate over the keys and then get the list values for them using https://github.com/pallets/werkzeug/blob/master/werkzeug/datastructures.py#L1017 or a different method on the Headers class?

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -386,14 +388,14 @@ def setUp(self):
         self.request_mock.method = "GET"
         self.request_mock.remote_addr = "190.0.0.0"
         self.request_mock.get_data.return_value = b"DATA!!!!"
-        query_param_args_mock = Mock()
-        query_param_args_mock.lists.return_value = {"query": ["params"]}.items()
+        query_param_args_mock = ImmutableMultiDict(MultiDict({"query": ["params"]}))
```
jfuss: Let's not change this. Let's keep the mocks.

samcli/local/apigw/local_apigw_service.py
medinarrior: Sure :+1:

tests/unit/local/apigw/test_local_apigw_service.py
medinarrior: Reverting it.

medinarrior:
Still working with integ-test and func-tests.

jfuss:
@medinarrior Sounds good. Let me know if you run into trouble with them.

samcli/local/apigw/local_apigw_service.py
medinarrior: Returning query_string_dict, multi_value_query_string_dict from _query_string_params.

medinarrior:
Made changes, let me know if there is more to do :)

jfuss:
@medinarrior Taking another pass through this now. Could you rebase when you get a chance? I don't think there should be any conflicts looking at the files.

medinarrior:
@jfuss Not able to see CodeBuild results.

jfuss:
@medinarrior Ignore the codebuild. We need to remove it.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -251,13 +251,10 @@ def _construct_event(flask_request, port, binary_types):
         identity = ContextIdentity(source_ip=flask_request.remote_addr)
 
         endpoint = PathConverter.convert_path_to_api_gateway(flask_request.endpoint)
-        method = flask_request.method
```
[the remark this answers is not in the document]
medinarrior: Removed those variables because of:

> R:243, 4: Too many local variables (17/15) (too-many-locals)

Updating maximum number of locals for function / method body to:

> max-locals=17

samcli/local/apigw/local_apigw_service.py
```diff
@@ -268,26 +265,23 @@ def _construct_event(flask_request, port, binary_types):
...
-                                 http_method=method,
+                                 http_method=flask_request.method,
                                  stage="prod",
                                  identity=identity,
                                  path=endpoint)
 
-        event_headers = dict(flask_request.headers)
-        event_headers["X-Forwarded-Proto"] = flask_request.scheme
-        event_headers["X-Forwarded-Port"] = str(port)
+        headers_dict, multi_value_headers_dict = LocalApigwService._event_headers(flask_request, port)
```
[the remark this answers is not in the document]
medinarrior: :)

medinarrior:
@jfuss Let me know if there is something I need to change :)

jfuss:
@medinarrior Appologizes on the time it has taken to look at your update. I am reviewing now.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -321,11 +319,46 @@ def _query_string_params(flask_request):
...
+        -------
+            Returns a dictionary of key to list of strings
+
+        """
+        headers_dict = {}
+        multi_value_headers_dict = {}
+
+        # Multi-value request headers is not really supported by Flask.
+        # See https://github.com/pallets/flask/issues/850
+        for header_key in flask_request.headers.keys():
```
jfuss: What do you think about logging some information about not fully supporting multiple headers? We would have to print this each time, since we have no idea if the customer is impacted but would result in two things:

1) It would give the customer some information when invoking (ideally linking out to a Github issue). This could result in less of a "Wait is this me or is this SAM CLI not giving the function the headers". 
2) It would give us (SAM CLI) a way to get feedback on how many customers are impacted by this through the Github Issue +1s. 

I know this is current state and I think the only alternative is to rewrite the local service or possibly patch Flask in our implementation so we can support this. I just want to try and avoid the confusion and frustration that may come about from this.

medinarrior: Good idea, what message should be displayed? Agree, it's time to consider patching Flask or rewriting the local service.

jfuss: Maybe something like:
"WARNING: Multi-value request headers are not fully supported. See issue: <link to sam cli github issue> for more details."

We would just need to create the issue with some details on how Flask doesn't support this and therefore we currently can't.

Thoughts?

jfuss:
There was a conversation earlier that suggested Flask does not support multiple headers but a recent experiment does show this will work.

```
from flask import Flask, request
app = Flask(__name__)

@app.route("/")
def hello():
     return "get: " + json.dumps(request.headers.get('Accept')) + "\n" + \
     "get_all: " + json.dumps(request.headers.get_all('Accept')) + "\n" + \
     "getlist: " + json.dumps(request.headers.getlist('Accept')) + "\n" + \
     "environ: " + json.dumps(request.environ['HTTP_ACCEPT']) + "\n"

flask run
``` 

Produces:
```
curl -s http://127.0.0.1:5000 -H'Accept: Working' -H'Accept: Not Working'
get: "Working,Not Working"
get_all: ["Working,Not Working"]
getlist: ["Working,Not Working"]
environ: "Working,Not Working"
```

samcli/local/events/api_event.py
```diff
@@ -191,7 +204,10 @@ def to_dict(self):
                      "resource": self.resource,
                      "requestContext": request_context_dict,
                      "queryStringParameters": dict(self.query_string_params) if self.query_string_params else None,
+                     "multiValueQueryStringParameters": dict(self.multi_value_query_string_params)
```
TheSriram: do we need ```dict(self.multi_value_query_string_params)``` again? we do validation above to make sure its a dict.

samcli/local/events/api_event.py
```diff
@@ -191,7 +204,10 @@ def to_dict(self):
...
+                     if self.multi_value_query_string_params else None,
                      "headers": dict(self.headers) if self.headers else None,
+                     "multiValueHeaders": dict(self.multi_value_headers) if self.multi_value_headers else None,
```
TheSriram: same here

samcli/local/events/api_event.py
sanathkr: Probably because someone could mutate the public variable on the object after initializing?

PR #742, opened 2018-10-30
Adding a Design doc template

sanathkr:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #743, opened 2018-10-30
Request for Comments: Design of new `sam build` command

sanathkr:
💻 [**Rendered Design Doc is here**](https://github.com/sanathkr/aws-sam-cli/blob/build-cmd-design/designs/sam_build_cmd.rst)

🗓 **RFC Open for feedback until**: 5 Nov, 2018

*Description of changes:*

Adding a design document for discussion. In this document, I explain the motivation and technical design behind a new command to create deployable artifacts for Lambda functions from source - `sam build`. 

Requesting comments on every aspect of the design, from user experience to implementation details. Once the discussion is closed, I will update the Task Breakdown section with actual list of tasks to implement this feature and create corresponding Github Issues.

This document is based on the design document template I added in #742  


designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+
+To build, package and deploy this app, customers would do the following:
+
+**1. Build:** Run the following command to build all functions in the template and output a SAM template that can be run through
+the package command:
+
+.. code-block:: bash
+
+    # Build the code and write artifacts to ./build folder
+    $ sam build -t template.yaml -b ./build -o built-template.yaml
```
heitorlessa: For a better user experience, I'd like to suggest adding them as default values and have the user to simply run `sam build`. 

Flags would be then used to for custom scenarios of which they can also use `samrc` config file

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+            CodeUri: ./build/MyFunction2
+            ...
+
+**2. Package and Deploy:** Package the built artifacts by running the *package* command on the template output
+by *build* command
+
+.. code-block:: bash
+
+    # Package the code
+    $ sam package --template-file built-template.yaml --s3-bucket mybucket --output-template-file packaged-template.yaml
```
heitorlessa: Similar to previous comment on UX, we could share the vision as to how defaults would be used here:

``sam package --s3-bucket mybucket``. ``--output-template-file`` and ``template-file`` should have these as defaults

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+
+Implementation
+==============
+
+CLI Changes
+-----------
+*Explain the changes to command line interface, including adding new commands, modifying arguments etc*
+
+#. Adding a new top-level command called ``sam build``.
+#. Add ``built-template.yaml`` to list of default template names searched by ``sam local`` commands
```
heitorlessa: As we're adding defaults to `sam local` commands, would it be possible to add that as part of `sam package` and `sam deploy` too?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+By running ``sam build`` on a build system that creates a new environment from scratch
+(ex: Travis/CircleCI/CodeBuild/Jenkins etc), you can achieve truly stable builds. For more information on why this
+is important, refer to Debian's guide on `reproducible builds <https://reproducible-builds.org/>`_.
+
+
+SAM CLI does the following to produce stable builds:
+
+#. Clean build folder on every run
+#. Include metadata when coping files and folders
+#. Run build actions with minimal information passed from the environment
```
heitorlessa: For those that don't have `buildspec.yaml`, are we running shell commands as per the Build Command table suggests? If so, it'd be great to make this explicit as the user would need to have package managers available via their prompt

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+arbitrary program that is accessible (via ``PATH`` resolution). The program will be invoked by setting three
+environment variables:
+
+#. ``__SOURCE_DIR`` - Folder where the source files are located
+#. ``__BUILD_DIR`` - Folder where the built artifacts should be written to
+#. ``__RUNTIME`` - AWS Lambda function runtime
+#. ``__TEMPLATE`` - Path to SAM Template
+#. Current Working Directory - Same directory where ``sam build`` command was invoked from
+
+This command is supposed to return an exit code of ``zero`` to indicate success. Non-zero exit codes indicate failure.
```
heitorlessa: s/supposed/expected

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+
+Documentation Changes
+---------------------
+TBD
+
+Open Questions
+--------------
+
+#. Should we support ``artifacts.json`` now to be future-proof?
+#. Should we create the default ``build`` folder within a ``.sam`` folder inside the project to provide a home for
```
heitorlessa: +1 to `.sam` folder as this provides room for growth (i.e. transform, build, dev plugins, etc)

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+   other scratch files if necessary?
+
+
+Task Breakdown
```
heitorlessa: Perhaps we should also have `sam init` as a task to update existing samples with `.samrc` and `sam build` as part of the README generated

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+**Manifest Name**: ``*.csproj``
+
++-------------+----------------------------------------------------------------------------------------+
+| Action      | Command                                                                                |
++=============+========================================================================================+
+| Resolve     | No Op                                                                                  |
++-------------+----------------------------------------------------------------------------------------+
+| Compile     | ``dotnet lambda package --configuration release --output-package $BUILD/package.zip``  |
```
heitorlessa: cc @sliedig, @blomskog and @pcolazurdo to share their DotNet experience. 

I remember having a chat with @sliedig and @pcolazurdo once that `dotnet lambda package`  includes additional files that are not necessarily needed for Lambda to function -- Perhaps there's a flag or something well known in that space that we can include

designs/sam_build_cmd.rst
sanathkr: Agreed. I am showing them here for explanation, but all of these will have defaults so you type nothing but `sam build`

designs/sam_build_cmd.rst
sanathkr: I don't want to change `sam package` functionality now. We should do that eventually when we migrate package completely into SAM CLI

designs/sam_build_cmd.rst
sanathkr: Yes we can, but I want to do it separately

designs/sam_build_cmd.rst
sanathkr: Actually, if you have a buildspec, `sam build` won't execute it locally. This feature is out-of-scope in this revision (disagree?)

We are instead running the package managers directly. Yes, we can make this upfront in the docs and also in error messages that you need the package manager installed.

designs/sam_build_cmd.rst
sanathkr: updated

designs/sam_build_cmd.rst
sanathkr: Agreed. My concern around `.sam` folder for this revision is inventing too many features at once and ending up with a lot of debt. Can we live without `.sam` folder for this release and add it later?

designs/sam_build_cmd.rst
sanathkr: `init` samples shouldn't need a custom `.samrc`. Whatever comes built-in with SAM CLI by default should just work with init samples. 

+1 on adding `sam build` command as part of readme. Will add that task to the task list

designs/sam_build_cmd.rst
sanathkr: feedback here would be super helpful!

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. Run build actions with minimal information passed from the environment
+
+
+Built-in Build Actions
+~~~~~~~~~~~~~~~~~~~~~~
+Build actions natively supported by SAM CLI follow a standard workflow:
+
+#. Search for a supported dependency manifest file. If a known manifest is not present, we will abort the build.
+#. Setup: Create build folder
+#. Resolve: Install dependencies
```
jfuss: nit: Why do you call this a Resolve step if we are installing?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+**Manifest Name**: ``package.json``
+
++-------------+--------------------------------------+
+| Action      | Command                              |
++=============+======================================+
+| Resolve     | ``npm install``                      |
++-------------+--------------------------------------+
+| Compile     | No Op                                |
++-------------+--------------------------------------+
+| Copy Source | Copy files and exclude node_modules  |
```
jfuss: exclude node_modules? Not sure I understand the context here. Is this a guard to not copy node_modules if they installed them into their source directory? If so, why don't we have the same 'excludes' elsewhere?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+Go's CLI will build the binary.
+
+**Manifest Name**: ``Gopkg.toml``
+
++-------------+---------------------------------------------------+
+| Action      | Command                                           |
++=============+===================================================+
+| Resolve     | ``dep ensure -v``                                 |
++-------------+---------------------------------------------------+
+| Compile     | ``GOOS=linux go build  -ldflags="-s -w" main.go`` |
```
jfuss: As a Go Lambda customer, how can I use build to compile my app for debugging locally?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+**Manifest Name**: ``*.csproj``
+
++-------------+----------------------------------------------------------------------------------------+
+| Action      | Command                                                                                |
++=============+========================================================================================+
+| Resolve     | No Op                                                                                  |
++-------------+----------------------------------------------------------------------------------------+
+| Compile     | ``dotnet lambda package --configuration release --output-package $BUILD/package.zip``  |
```
jfuss: Same statement as the go but for .Net. Can I build for debugging locally, so I do not have to change my flow?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+| Copy Source | Copy all files                                                                        |
++-------------+---------------------------------------------------------------------------------------+
+
+
+Custom Build Commands
+~~~~~~~~~~~~~~~~~~~~~
+
+Users can provide custom build command for each runtime via ``.samrc`` that will be run instead of the default build
+action provided by SAM CLI. SAM CLI will invoke this command for every resource that uses this runtime. This can be any
+arbitrary program that is accessible (via ``PATH`` resolution). The program will be invoked by setting three
```
jfuss: arbitrary screams scary...

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. ``__TEMPLATE`` - Path to SAM Template
+#. Current Working Directory - Same directory where ``sam build`` command was invoked from
+
+This command is expected to return an exit code of ``zero`` to indicate success. Non-zero exit codes indicate failure.
+
+Building Native Binaries
+~~~~~~~~~~~~~~~~~~~~~~~~
+To build native binaries, we need to run on an architecture and operating system that is similar to AWS Lambda. We use
+the Docker containers provided by `Docker Lambda <https://github.com/lambci/docker-lambda>`_ project to run the same
+set of commands described above on this container. We will mount source code folder and build folder into the container
```
jfuss: Do we need to install extra things for each runtime to make this work on Docker Lambda (donet cli or go cli for example)? If so, how can we do this in a way that doesn't require downloading of the same things per build?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+so the commands have access to necessary files.
+
+If the user specified a custom build command, we will attempt to run the command as-is within the Docker Container.
+But it is possible that the command might not be available inside the container.
```
jfuss: Called this out above, but how do we ensure that doesn't happen (for at least the default cases)? It's a pretty poor experience if I can build locally w/o a docker container and then can't inside the container. This breaks the experience of build, imo.

designs/sam_build_cmd.rst
jfuss: A way to weigh the now vs later:
Is this `build` folder something we will encourage customers to edit or mutate? If so, moving this later becomes more difficult, as we have to be concerned with breaking/change customer behavior.

On the same note: Should `build` be a hidden folder?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+    # Package the code
+    $ sam package --template-file built-template.yaml --s3-bucket mybucket --output-template-file packaged-template.yaml
+
+    # Deploy the app
+    $ sam deploy --template-file packaged-template.yaml --stack-name mystack
+
+Other Usecases
+~~~~~~~~~~~~~~~
+
+#. **Build Native Dependencies**: Pass the ``--native`` flag to the *build* command
```
TheSriram: can you expand on this a little bit?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. **Out-of-Source Builds**: In this scenario, Lambda function code is present in a folder outside the folder containing
+   the SAM template. Absolute path to these folders are determined at runtime in a build machine. Set the
+   ``--root=/my/folder`` flag to absolute path to the folder relative to which we will resolve relative *CodeUri* paths.
+#. **Inherited dependency manifest**: By default, we will look for a dependency manifest (ex: package.json) at same
+   folder containing SAM template. If a ``--root`` flag is set, we will look for manifest at this folder. If neither
+   locations have a manifest, we will look for a manifest within the folder containing function code. Manifest present
+   within the code folder always overrides manifest at the root.
+#. **Arbitrary build commands**: Override build commands per-runtime by specifying full path to the command in
+   ``.samrc``.
```
TheSriram: `samrc` ftw.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+(ex: timestamp, ownership) on source files changes. This is an important attribute of a build system. Since SAM CLI
+relies on 3rd party package managers like NPM to do the heavy lifting, we can only provide a "best effort" service here.
+By running ``sam build`` on a build system that creates a new environment from scratch
+(ex: Travis/CircleCI/CodeBuild/Jenkins etc), you can achieve truly stable builds. For more information on why this
+is important, refer to Debian's guide on `reproducible builds <https://reproducible-builds.org/>`_.
+
+
+SAM CLI does the following to produce stable builds:
+
+#. Clean build folder on every run
```
TheSriram: should we clean it, only if there is change in source?

designs/sam_build_cmd.rst
heitorlessa: If not too much work running CodeBuild locally would be great but understand MVP and time constraints - And yes, adding an extra clarification here to avoid any possible confusion:

https://github.com/sanathkr/aws-sam-cli/blob/build-cmd-design/designs/sam_build_cmd.rst#stable-builds

designs/sam_build_cmd.rst
heitorlessa: I vote for having `.sam` but no `artifacts.json`. `.sam` folder is easier to relate to and can easily be added to `.gitignore` so you don't have to add both `built-template.yaml` + `build` folder and whatever else comes afterward.

For now, I see that as a way to keep things tidy that will avoid polluting their workspace. 

Hope that makes sense

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+   * Nodejs with NPM
+   * Java with Maven
+   * Python with PIP
+   * Golang with Go CLI
+   * Dotnetcore with DotNet CLI
+
+
+#. Each Lambda function in SAM template gets built
+
+#. Produce stable builds (best effort): If the source files did not change, built artifacts should not change.
```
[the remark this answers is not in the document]
heitorlessa: +1. Taking a step further is how are we going to decide on that? Hashing? Hashing stored locally?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+^^^^^^^^^^^^^^^^^^^^
+
+Install dependencies specified by ``package.json`` and copy source files
+
+**Manifest Name**: ``package.json``
+
++-------------+--------------------------------------+
+| Action      | Command                              |
++=============+======================================+
+| Resolve     | ``npm install``                      |
```
simalexan: Appreciate the way you're going with this, so we (from Claudia.js) would like to ask if you'd like us to share the code from our `claudia pack` command? 

As just an `npm install` with excluding `node_modules` is not going to be enough. You will have other files, such as `.eslintrc`, still present everywhere.

sanathkr: YES YES YES 🔥Thank you so much for offering. This is exciting to get deeper Nodejs build support into the command. The current technical design does not support such integrations. I will rework and send an updated PR.

Broadly, here are some questions I am trying to answer:
- Can we support build actions written natively in other languages (JS/Golang/C#) etc?
- How can we auto-detect certain frameworks and invoke appropriate build actions for that framework because SAM CLI's built-in functionality is never going to be as deep and opinionated as the framework's? (ex: call `claudia pack` for ClaudiaJS apps, or `zappa package` for Zappa apps) 
- How can I let other non-SAM CLI developers write custom build actions?
- How can we build all the custom build actions into SAM CLI so customers don't have to install a plugin?
- If the build command is a collection of 20-30 different build actions for various frameworks, does it make sense to create a separate repository and vend it as a library that SAM CLI and other tools can consume? I know we are never going to ever support all frameworks/opinions applying the 80:20 rule, what would a good 80% look like?

I might be digressing from the original intent of the build command, but seeing the community interest and going through other build/package implementations, I feel like we can all benefit from each other if we collect our efforts.

Thoughts?

gojko: @sanathkr `claudia pack` will do the proper packaging for any nodejs runtime (dependency installation & rewriting for local deps and removing test resources), so it's not just for projects made with claudia. eg for video puppet, I'm using cloudformation to deploy lambdas with nodejs (not even sam, because of nested templates not working), but I'm using claudia pack to create deployable zips.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. Produce stable builds (best effort): If the source files did not change, built artifacts should not change.
+
+#. Built artifacts should "just work" with ``sam local`` and ``sam package`` suite of commands
+
+#. Ability to provide arbitrary build commands for each Lambda Function runtime. This will either override the built-in
+   default or add build support for languages/tools that SAM CLI does not support (ex: java+gradle).
+
+#. Opt-in to building native dependencies that can run on AWS Lambda using Docker.
+
+#. Support one dependency manifest (ex: package.json) entire app or one per each Lambda function.
```
txase: I'm unaware of a good use case for having a global dependency manifest instead of per-function manifests. Per-function manifests is a best practice:

* Reduces package sizes (which has lots of benefits like reduced cold-start latencies)
* Reduces security attack space (more deps means more potential vulnerabilities in each function)
* Reduces confusion (why is *this* module in *that* function)
* Reduces dependency resolution issues (in case you mix global and per-function manifests)
* and more I haven't thought of off the top of my head...

The only use case for global manifests that I can see is that you only need to update one manifest to update dependencies that all functions use. But the burden of updating each manifest individually is trivial. Simple tooling, if necessary, can remove this burden.

gojko: The 50MB limit per package isn't much anyway, so by sharing dependencies you're limiting what a single SAM template can do unnecessarily. 

For node.js projects at least, I'd prefer separate package.json per lambda if there are multiple lambdas in a project. in fact, it's actually a separate package.json per project (directory), so a single directory can host multiple lambdas. that way, if someone wants to speed up build time by sharing dependencies, great, they can do it. if they want to keep it minimal per lambda (makes more sense performance-wise), they can.

designs/sam_build_cmd.rst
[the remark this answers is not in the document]
ikben: > Identical sources file will produce identical build artifacts
Can be confusing for dependency updates, which you may want to include even of your source code stays the same.

designs/sam_build_cmd.rst
heitorlessa: adding Mark @Ubiguchi as he also has great experience in this area for comments

designs/sam_build_cmd.rst
sanathkr: Running CodeBuild locally is possible, I felt like it could be a post-MVP thing.

designs/sam_build_cmd.rst
sanathkr: @jfuss  No customers are not expected to mutate the build folder, but it is not something we can control.

@heitorlessa I agree with your reasoning. It does standardize a few things. But it also adds new surface area for things we need to maintain. We need to think thru questions like where does the folder exist, can it exist in multiple places (project root & user home), can other tools consume the contents of this folder, how does sam cli operate in absence of this folder, how do CI/CD systems use this folder to extra artifacts etc. I am just not sure if this proposal is the right place to address these questions. It certainly adds another reason why we need this folder.

Can you create a Github Issue instead so we can track this work for future? It will also help get feedback from community

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
+``sam build`` command
+=====================
+This is the design for a command to **build** a Lambda function. **Build** is the operation of converting the function's
+source code to an artifact that can be executed on AWS Lambda.
```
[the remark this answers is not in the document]
sanathkr: Artifact is the contents of the zip file but within a folder. This allows customers or other tools to make more modifications on the folder before they run "sam package" command to create the zip file.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,440 @@
...
+the source code location, creating temporary folders to store built artifacts, running the build command and
+move artifacts to right location.
+
+
+Success criteria for the change
+-------------------------------
+#. Support all programming languages supported by AWS Lambda
+
+   * Nodejs with NPM
+   * Java with Maven
```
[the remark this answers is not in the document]
sanathkr: Coin toss - just kidding. "sam init" example uses Maven. We choose that for a reason. So sticking with Maven.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. Support out-of-source builds: ie. source code of Lambda function is outside the directory containing SAM template.
+
+#. Integrate build action with ``sam local/package/deploy`` commands so the Lambda functions will be automatically
+   built as part of the command without explicitly running the build command.
+
+
+Out-of-Scope
+------------
+#. Supports adding data files ie. files that are not referenced by the package manager (ex: images, css etc)
+#. Support to exclude certain files from the built artifact (ex: using .gitignore or using regex)
```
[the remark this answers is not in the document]
sanathkr: We will exclude certain files & folders by default like `*.pyc` that makes sense, but we won't be able to support parsing an arbitrary `.gitignore` file. So it is possible that we don't create the slimmest zip file. But this could be a fast follow feature.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. Support in AWS CLI, ``aws cloudformation``, suite of commands
+#. Support for fine-grained hooks (ex: hooks that run pre-build, post-build, etc)
+
+
+User Experience Walkthrough
+---------------------------
+Let's assume customers has the following SAM template:
+
+    **NOTE**: Currently we advice customers to set *CodeUri* to a folder containing built artifacts that can be readily
+    packaged by ``sam package`` command. But to use with *build* command, customers need to set *CodeUri* to the folder
```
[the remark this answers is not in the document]
sanathkr: Customers do need to modify their SAM template to make use of build command (CodeUri points to source folder instead of built folder). Unfortunately there is no way around it. It is not a breaking change in the sense that your existing workflows won't break. Customers need to explicitly make this change to opt-into the build feature.

Also, the `built-template.yaml` is not the best UX. But besides introducing a `.sam` folder and storing these files there, I don't know of a good solution. Another option is to combine build & package so customers go straight to the `packaged-template.yaml` without the additional step.

Thoughts?

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+   within the code folder always overrides manifest at the root.
+#. **Arbitrary build commands**: Override build commands per-runtime by specifying full path to the command in
+   ``.samrc``.
+#. **Build & Run Locally**: Use the ``--template`` property of ``sam local`` suite of commands to specify the
+   template produced by *build* command (ex: ``build-template.yaml``)
+
+Implementation
+==============
+
+CLI Changes
```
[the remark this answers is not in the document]
sanathkr: Good point. Will do

designs/sam_build_cmd.rst
sanathkr: Resolving dependencies - it could be installing or downloading or just linking files correctly

designs/sam_build_cmd.rst
sanathkr: We should have similar excludes elsewhere. I will update the doc to reflect that

designs/sam_build_cmd.rst
sanathkr: Interesting use case. `sam build` won't help you with this because it doesn't add debugging symbols. I will explicitly call this out as out-of-scope so we can iterate on this in future

designs/sam_build_cmd.rst
sanathkr: Yes! I want to skip this feature for MVP because it opens a pandoras box. Running arbitrary commands without a real RPC-like protocol and controls seems scary. At that point customers can just write a full build script themselves. I want to think this through before we go support it.

designs/sam_build_cmd.rst
sanathkr: No, `docker-lambda` publishes a build container that has these preinstalled

designs/sam_build_cmd.rst
sanathkr: Agreed. And hence another reason to skip support for running arbitrary build commands in MVP.

designs/sam_build_cmd.rst
```diff
@@ -0,0 +1,442 @@
...
+#. Built artifacts should "just work" with ``sam local`` and ``sam package`` suite of commands
+
+#. Ability to provide arbitrary build commands for each Lambda Function runtime. This will either override the built-in
+   default or add build support for languages/tools that SAM CLI does not support (ex: java+gradle).
+
+#. Opt-in to building native dependencies that can run on AWS Lambda using Docker.
+
+#. Support one dependency manifest (ex: package.json) entire app or one per each Lambda function.
+
+#. Support out-of-source builds: ie. source code of Lambda function is outside the directory containing SAM template.
```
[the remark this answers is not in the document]
sanathkr: Could be a different github repo. But for the scope of this command, it is just a separate folder that was already downloaded from remote.

designs/sam_build_cmd.rst
sanathkr: Sure, will do.

designs/sam_build_cmd.rst
sanathkr: It will give better performance. But it also makes the build implementation complicated. We can start here and then improve this in future

designs/sam_build_cmd.rst
sanathkr: Ah I didn't know `claudia pack` is generic. Yes, it would be fantastic to be able to adopt that functionality into SAM CLI. 

We have two options:

**Option 1: Include a JS module in SAM CLI's Python source**
To build Node projects, SAM CLI just invoke the JS module `node pack.js`.

Pros: 
- Easy to lift & shift
- Easy to use language specific libraries that can support deeper integrations in future like webpack build or running gulp scripts
- Sets precedence for other runtimes like Java which might need reflexion to create the package
- Easier to get help from JS community who is more familiar with building JS packages.

Cons:
- Vending JS files in Python package
- Might take dependency on certain version of Node. We can't enforce that customers have this version of Node on their system. 
- Might have to webpack all dependencies, minify and vend one file that we just run using `node pack.js`.
- Need to spawn & manage processes
- Could become a tech debt if this approach doesn't scale.

**Option 2: Rewrite the `claudia pack` functionality in Python**

Pros:
- Native implementation. 
- Implementation might create general features that can be used by build actions for other runtimes
- Faster

Cons:
- Takes a lot more work. Can't lift & shift
- Won't get to use language native features.


What do you guys think?

designs/sam_build_cmd.rst
sanathkr: @txase @gojko  Thanks for the detailed comments. This makes a lot of sense. I will remove the app-level manifest support for now. We can always add it later if ppl have a real good use-case for it. May be make it opt-in at that point.

designs/sam_build_cmd.rst
stojanovic: I guess we can start with Node.js (option 1), and rewrite in Python if needed later.

For Node.js version, with Claudia.js we try to fully support versions supported by Lambda, but it works in all major versions of Node. Also it’s easy to have multiple versions of Node.js with tools like nvm.

gojko: I think it's safe to assume that people deploying node.js lambda projects have node.js installed locally. If not, then neither of the two solutions will work, unless you want to fully rewrite NPM in python as well. 
 
Regarding your option 2 and second con, I don't think that is relevant. if you're just thinking of packaging up lambdas, there are no language-native features required for it (apart from loading JSON, which I assume you can do with Python as well). It's just a matter of moving files around, zipping them up, and invoking NPM a few times with the right options. 

How about having an `install` and `package` steps for language-dependent builders? `install` could prepare the builder from python, by effectively creating an local executable in a runtime-specific way (for example, in case of node.js, this would run `npm install` on a subdirectory where you have the packaging script, to pull the right dependencies; in case of Java it could grab the maven shade plugin...). If that step fails, SAM can complain and tell the user that the local environment needs aren't met, so it can't build the project, and potentially direct them to a URL documenting what local tools are required for building that runtime. If that step succeeds, you know that it's possible to build it correctly, and SAM would just need to invoke a system executable produced/configured by the first step.

designs/sam_build_cmd.rst
pcolazurdo: The latest version of the Amazon.Lambda.Tools creates a very slim version of the package so I would trust this. The tool creates a zip with all the required dependencies. If project has static files and they are included in the csproj definition, they will be automatically included in the zip

pcolazurdo: I would use the standard microsoft/dotnet docker image to compile. It would need to add dependencies to work with Lambda so a Dockerfile can be included that would add these dependencies on package creation. Not sure what is the model with other languages, but I think this would be a good solution.

designs/sam_build_cmd.rst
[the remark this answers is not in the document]
pcolazurdo: Stable build should be based only on source changes. The user could run clean/build to force dependencies update

designs/sam_build_cmd.rst
heitorlessa: Happy to create a new issue for that but I think we should discuss here - Here's the answers to these questions for a MVP:

#### Can it exist in multiple places (project root & user home)?

No, just project root. `.sam` folder will contain build artifacts and it's project specific; moving to user home brings unwanted side effects (disk space and logic complexity).

#### Q: Can other tools consume the contents of this folder?

`.sam` folder contains read-only build artifacts so any tool built to work with sam can read from that.

#### Q: How does sam cli operate in absence of this folder?

We should create it in the absence of it - That provides a `Just Works(tm)` experience. Here's a sample scenario for this:

##### No explicit sam build from the user

**UX**

This is the ideal scenario assuming `.sam` folder contains artifacts, packaged-template and other commands are aware of `.sam` and `.samrc`:

* Initialize boilerplate: `sam init --name my-project && cd my-project`
* Deploy: `sam deploy`

SAM application will be deployed seamless and the entire process logged. One could introduce `sam local` before `sam deploy` and `sam build` will run seamlessly if `.sam` folder is absent.
Power users can then modify the convention and opinions if they wanted to.

**Logic**

* `sam deploy` checks `.sam` absence and run `sam build` for the runtime chosen in `template.yaml`
* `sam deploy` then run `sam package`
* `sam package` creates a S3 bucket based on project name (`.samrc`) if one doesn't exist
* `sam deploy`  ultimately deploys the application
* `.samrc` is instrumental in allowing the user to add his/her opinions to change this flow (e.g. custom build, custom stack name, custom s3 bucket name)

**Pros**

1. Seamless experience and just works without having to know `built-template.yaml`, `packaged-template.yaml`, etc.
2. New users don't have to learn upfront how Lambda packaging works and embed `sam local` within their package managers
    - `npm run build -> sam build`
    - `npm run deploy -> sam deploy`
    - `make build -> sam build`
    - `make deploy -> sam deploy`
3. Power users can rely on `.sam` and build wrappers/tools until `sam package/deploy` changes its logic

**Cons**

1. This can only happen post-MVP as `sam local`, `sam package`, and `sam deploy` would need to check for `.sam`

#### Q: How do CI/CD systems use this folder to extra artifacts etc.

Extra artifacts is a responsibility of the CI system and SAM only takes into account the application artifact so CloudFormation can deploy it afterwards - But perhaps I didn't understand the question.

Here's an example using CodeBuild where CodePipeline would pass that as an Input Artifact to CloudFormation:

```yaml
phases:
  install:
    commands:
      - echo "[Security phase]"
      - pipenv check # run some CVE checks on deps
  pre_build:
    commands:
      - echo "[Pre-Build phase] Nothing to be done as SAM build now handles that for me ;)"
  build:
    commands:
      - echo "[Build phase] Starting SAM build `date` in `pwd`"
      - sam build
      - echo "[Build phase] SAM build completed on `date`"
  post_build:
    commands:
      - echo "[Post-build] Uploading artifacts to S3 `date`"
      - sam package

artifacts:
  files:
    - .sam/template-packaged.yaml
  discard-paths: yes
```

By supporting `.sam` folder and only storing `artifacts` this can set the path for a fast follow up on `sam package` and `sam deploy` afterwards.

designs/sam_build_cmd.rst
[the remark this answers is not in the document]
gojko: another benefit of using NPM for node, is that this is automatically handled by NPM.

designs/sam_build_cmd.rst
sanathkr: @heitorlessa  thanks for the detailed thoughts. It is very encouraging to see this mature. I created https://github.com/awslabs/aws-sam-cli/issues/748 to follow up on a separate thread. This needs proper design to make sure we are making the right decisions. I don't want to haste into it.

designs/sam_build_cmd.rst
sanathkr: @pcolazurdo Thanks for the review. Its great to know we are in the right direction here. 

Can you elaborate on your comment about using standard `microsoft/dotnet` image? With native dependencies (like C libraries), we need to compile against a close version of Amazon Linux that Lambda runs. Doesn't dotnet have a similar requirement?

Also, for dependencies, wouldn't you be adding them to your .csproj file or something? Why do you need the Dockerfile?

designs/sam_build_cmd.rst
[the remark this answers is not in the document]
sanathkr: Agreed.. This is the intention too.

designs/sam_build_cmd.rst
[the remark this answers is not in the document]
sanathkr: This is interesting. It also creates a problem where we don't have a consistent experience across different programming languages (ie. Python won't respect .gitignore but Nodejs builds will). May be that is acceptable?

designs/sam_build_cmd.rst
sanathkr: @stojanovic Yes, let's do that. The one thing I underestimated is the level of help existing Javascript libraries will provide to package JS apps. Ex: You could just import a gulpfile as module and run actions without treating it like a CLI. 

This creates an interesting state where SAM CLI's Python package will contain JS libraries, but I think that is fine.

sanathkr: @gojko I like your idea of splitting the `install` & `package` phases. I would even split the install phase into `prepare`, `resolve-dependencies`, `copy-source` phases where we need to call NPM to simply `resolve-dependencies` and for everything else we will have default Python implementations.

When we consume Claudia's `pack` functionality, we can work through the details of where to draw the line for each phase. But I think it is worth keeping the `pack` code in Javascript so we can add support for advanced building like running Gulpscripts or Webpack etc without re-implementing all these libraries in Python.

sanathkr: I added information about how this will be implemented to the design doc: https://github.com/sanathkr/aws-sam-cli/blob/build-cmd-design/designs/sam_build_cmd.rst#implementation-of-build-actions

sanathkr:
I have addressed all comments and updated design document to reflect the latest discussion. Can you guys approve the PR so I can start implementation?

@jfuss @TheSriram @simalexan @gojko @stojanovic

willyg302:
See below the notes from our security-focused review.

### Background notes

- .samrc and custom build actions are out of scope. This removes a large portion of the attack surface.
- The library of build actions will start with the defaults documented. In the future, the hope is to grow to 20-30 builders covering 90% of use cases (for example, a webpack builder). Tentatively, SAM CLI will use a combination of runtime and manifest to decide which builder takes precedence. Because the library is curated, we can guarantee precedence behavior.
- For now, data files and ignoring can be delegated to the "native" command such as the npm install script.
- Consider moving `built-template.yaml` to `build/template.yaml`.

### Discussion 1: protecting customers

We want to prevent customers from shooting themselves in the foot. Consider the following scenario:

1. A malicious actor publishes an open-source Node.js SAM app to GitHub, complete with a package.json that contains a malicious command in the install script (for example, `rm -rf /`).
2. A customer hears about SAM and the app from 1. They install SAM CLI, clone the app, and read that to deploy the app to their own AWS account they must `sam build/package/deploy`.
3. Upon running `sam build`, something bad happens.

Note that 3 would have also happened if the customer ran just `npm install`, and is a general problem with package managers that support arbitrary logic. However, there is a layer of indirection and it may not be immediately obvious what `sam build` will do without reading documentation beforehand.

Mitigations:

- Print/log commands that SAM CLI runs as it runs them.
- Check for sudo. Only run as root if `--force` or `--native` (since that's in a container already) are present.
- Recommend that customers sandbox builds by using `--native`.

Other ideas:

- Sanitizing commands before running them. This is hard and brittle, and could lead to other unintended consequences.
- Prompting customers before running commands and supporting e.g. `sam build --yes` for automation. This is annoying. Customers will learn to pass `--yes`, giving them no added protection.
- Make `--native` the default. This is a breaking change since it assumes Docker is installed, which currently isn't necessary for the package/deploy lifecycle.

### Discussion 2: exec logic

The library of builders will bundle all dependencies, so that there is no post-install needed. For initial launch there should be no bundled dependencies. In the future the tenet is to assume library-specific dependencies (such as webpack) are installed and available on SAM CLI's path -- this path will be forwarded.

There are two points of execution:

- **SAM CLI --> builder script**. Using Python's `subprocess` module. Already used in SAM CLI [here](https://github.com/awslabs/aws-sam-cli/blob/1d276ea37c3f446605ef6570b7e3be77195c26aa/samcli/lib/samlib/cloudformation_command.py), will follow similar approach (no `shell=True`, etc.).
- **builder script --> native (`npm`, etc.)**. Logic here is builder-dependent. The strategy going forward is likely to treat the builder library package as a true dependency, with separate reviews. SAM CLI will pin to every major release of that package and pull in added builders as a batch each time.

These are reasonable approaches. If a future pen test of SAM CLI is conducted, remember to call out these integration points to focus on.

sanathkr:
Created a separate Github Repository to implement each build action: https://github.com/awslabs/aws-lambda-builders. This helps evolve the builders independent of the CLI. Contributors writing build actions need not know/use SAM CLI.

sanathkr:
@gojko Actually hang on for a few days. I want to prototype Python first so we have the right scaffolding before we take the JS lib.

sanathkr:
### Security Review Notes

**Threats:**
1. Abusing sam build to escalate privileges 
2. Sam build relies on third party dependencies with known security vulnerabilities. 
3. External customers having a bad experience with sam build
	1. Locally (Unexpected behavior)
	2. The generated lambda artifacts are invalid or don’t work

**Recommendations:**
1. Implement sudo check when sam build is invoked. When sudo rights are detected sam does a hard fail. The hard fail behavior can be overwritten by an environment variable. 
2. Writing temp folders with 700
3. Provide mechanism to inform customers about security vulnerabilities in sam cli (probably thru a message when CLI is used)
4. When using sam build to compile native dependencies in amazon linux we use docker. In order to move the build artifacts out of docker we recommend to perform an explicit copy from inside the container to the workstation using `docker cp`. 
5. Proper unit & integration tests on various platforms to ensure customers don't get a bad experience w.r.t the generated artifacts.

edtbl76:
Any update on this?

TheSriram:
@edtbl76 sam build is available today for Python Runtimes, try giving the latest version of aws-sam-cli a whirl.

edtbl76:
Yes, I need node runtimes for this to be useful. Is there an ETA?

jfuss:
@edtbl76 You can find other language support in our issues. Those issues will be closed after we release support. You can track progress there. 

Locking this issue to reduce noise and asks about language support. This was a design for sam build and will track further language supports through issues.

Issue #299, opened 2018-02-09
Support request validators and request/response mappings


kadrach:
It would be great if `aws-sam-local` could support the request validation features of the APIGateway and request/response mappings with Velocity templates! Alternatively a log message on local calls that these parts of the definition are being ignored :)

brunofitas:
Hi guys. Any progress on this? This is a really important feature.
Thanks

ejhayes:
@brunofitas @kadrach I have implemented this functionality locally https://github.com/ejhayes/aws-sam-cli. 

The velocity template mapping and JSONPath looks to be working fine, but java specific string functions like `replaceAll` will not work. For example `$util.escapeJavaScript($input.json('$.errorMessage'))` would work, but `$util.escapeJavaScript($input.json('$.errorMessage')).replaceAll("\\'","'")` would not.

Hoping to have a PR opened in the next couple days.

brunofitas:
Thanks Eric!

heitorlessa:
Awesome one Eric!!!

Look forward to the PR and shout out if you need a hand to us

hanstf:
Hi, will this going to be implemented ?
also @ejhayes, is your pull request only cater the x-amazon-apigateway-integration ?
As my template is came from the CDK and something like this

```
      Integration:
        ContentHandling: CONVERT_TO_TEXT
        IntegrationHttpMethod: POST
        IntegrationResponses:
          - ResponseParameters:
              method.response.header.Access-Control-Allow-Origin: "'*'"
            ResponseTemplates:
              application/json: "{\"data\": $input.json('$')}"
            StatusCode: "200"
          - ResponseParameters:
              method.response.header.Access-Control-Allow-Origin: "'*'"
            ResponseTemplates:
              application/json: "{\"message\":\"$input.path('$.message')\"}"
            SelectionPattern: .*Exception.*|.*error.*
            StatusCode: "400"
        PassthroughBehavior: WHEN_NO_TEMPLATES
        RequestTemplates:
          image/jpeg: '{"base64Image":"$input.body","token":"$input.header.Authorization"}'
```

rv2673:
@jfuss or others? Is this in the aws sam cli roadmap at all? This issue is more then 2 years old, had contributing PR that was rejected, so is it being worked on at all?

hirosato:
@jfuss  I would like to restart [this PR](https://github.com/aws/aws-sam-cli/pull/746). I understand the concerns about adding new dependencies.
[...]

PR #746, opened 2018-11-01
feat(request-templates): API Request Data Mapping

ejhayes:
*Issue #, if available:*
#299 

*Description of changes:*
- Add support for VTL request template mapping (matches [API Gateway Mapping Template Reference Documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html)). The following are supported:
  - `$context`
  - `$input`
  - `$stageVariables`
  - `$util`
- Request templates must be defined in `x-amazon-apigateway-integration` in the swagger definition, for example:
```
...
x-amazon-apigateway-integration:
    uri:
      Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${MyFunction.Arn}/invocations
    requestTemplates:
      application/json: "...VTL TEMPLATE HERE..."
...
```

This is a WIP. Remaining:
- [ ] Tests
- [ ] Documentation
- [ ] Examples

Known caveats/limitations:
- [`$method`](https://docs.aws.amazon.com/apigateway/latest/developerguide/request-response-data-mappings.html) is not supported.
- Java string functions (e.g. `replaceAll`) are not supported. This means that things like `$util.escapeJavaScript($input.json('$.errorMessage')).replaceAll("\\'","'")` don't work, but `$util.escapeJavaScript($input.json('$.errorMessage'))` will.
- Kinda unrelated, but something I noticed. `Fn::Sub` isn't fully implemented so the following will result in the wrong function name being returned:
```
# parsed function name is CORRECT
Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${MyFunction.Arn}/invocations

# parsed function name is INCORRECT (returns "FunctionArn" instead of MyFunction.Arn)
Fn::Sub:
  - arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${FunctionArn}/invocations
  - { FunctionArn: !GetAtt MyFunction.Arn }
```

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,4 +11,5 @@ docker>=3.3.0
 dateparser~=0.7
 python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
-
+airspeed~=0.5.10
+jsonpath-rw~=1.4.0
--- a/samcli/commands/local/lib/local_api_service.py
+++ b/samcli/commands/local/lib/local_api_service.py
@@ -100,7 +100,7 @@ class LocalApiService(object):
         routes = []
         for api in api_provider.get_all():
             route = Route(methods=[api.method], function_name=api.function_name, path=api.path,
-                          binary_types=api.binary_media_types)
+                          binary_types=api.binary_media_types, request_template=api.request_template)
             routes.append(route)
 
         return routes
--- a/samcli/commands/local/lib/provider.py
+++ b/samcli/commands/local/lib/provider.py
@@ -75,10 +75,16 @@ _ApiTuple = namedtuple("Api", [
     "cors",
 
     # List(Str). List of the binary media types the API
-    "binary_media_types"
+    "binary_media_types",
+
+    # String. Optional velocity template.
+    # If this configuration is set, then the event passed to the function
+    # will be transformed using Velocity templating language.
+    "request_template"
 ])
 _ApiTuple.__new__.__defaults__ = (None,  # Cors is optional and defaults to None
-                                  []     # binary_media_types is optional and defaults to empty
+                                  [],    # binary_media_types is optional and defaults to empty
+                                  None   # request_template is optional and defaults to None
                                   )
 
 
--- a/samcli/commands/local/lib/swagger/parser.py
+++ b/samcli/commands/local/lib/swagger/parser.py
@@ -85,8 +85,10 @@ class SwaggerParser(object):
                     # Convert to a more commonly used method notation
                     method = self._ANY_METHOD
 
-                api = Api(path=full_path, method=method, function_name=function_name, cors=None,
-                          binary_media_types=binary_media_types)
+                request_template = self._get_integration_req_template(method_config)
+
+                api = Api(path=full_path, method=method, function_name=function_name, request_template=request_template,
+                          cors=None, binary_media_types=binary_media_types)
                 result.append(api)
 
         return result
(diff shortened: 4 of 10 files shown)


IRoninCoder:
This pull request will enable VTL in `requestTemplate`s and we have a crucial migration project that's pending this feature. Is it possible to have this in the next release?

requirements/base.txt
```diff
@@ -11,4 +11,5 @@ docker>=3.3.0
 dateparser~=0.7
 python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
-
+airspeed~=0.5.10
```
jfuss: This project doesn't appear to be maintained and isn't being tested on Python 3.5 or later. Considering we support 3.6 and 3.7, this is concerning.

requirements/base.txt
```diff
@@ -11,4 +11,5 @@ docker>=3.3.0
...
+jsonpath-rw~=1.4.0
```
jfuss: Doesn't look like 3.7 is supported and notes that there are segfaults in the README which is concerning: "It is tested with Python 2.6, 2.7, 3.2, 3.3, 3.4, 3.5, and 3.6. (On travis-ci there is a segfault when running the tests with pypy; I don't think the problem lies with this library)."

IRoninCoder:
@jfuss I find it contradicting with `start-api` feature since without the VTL that may not work at all due to breakage of APIs which expect a certain `event` schema. As far as the cloud services, i'm not very familiar but i assume that they do not modify the swagger's hierarchy that often, if at all (so assumption is that `requestTemplate` mostly stays where it is). That leaves the VTL syntax support, which has been mostly the same for a very long time and this branch only adds a small subset of VTL support as mentioned in the description. Although to add to your concerns, someone could ask for more VTL support and that does drag into the maintenance situation, but the basic can stay as is with minimal maintenance.

requirements/base.txt
ejhayes: This is a valid concern @jfuss --the underlying library looked fairly lightweight and I didn't see anything else that implemented velocity templates in Python. It seems that we left left with several options here:

1. Close this PR / not support this functionality in SAM Local
2. Tie into java version of velocity templates (this seems unwieldy and like a total PITA)
3. Roll our own velocity templating engine as part of SAM Local
4. Fork/maintain airspeed codebase. Ensure compatibility with python 3.6/3.7. Limit scope of functionality provided by library to be as minimal as possible.

Considering that SAM CLI is on track to becoming the official supported way to develop serverless locally I think having the ability to mimic certain core functionality locally is very important for adoption of SAM Local. Velocity templates can feel a bit heavy handed and having to push to AWS to test out different request templates definitely feels tedious.

How do you think we should proceed?

requirements/base.txt
ejhayes: Same comment as above with airspeed. The specific error occurring looks to be for pypy vs. the regular install of python. Unless a large number of users of SAM Local are using Pypy this seems like less of a concern. As for support for python 3.7, I have a PR in for that right now (see https://github.com/kennknowles/python-jsonpath-rw/pull/65)

jfuss:
Closing due to be inactive and out of date.

hirosato:
@jfuss I would like to restart this PR. I understand the concerns about adding new dependencies.
How about implementing `$input.json('$')` and `$input.params().path['paramname']`, for example, instead of trying to support as much expressions as possible? I mean we can choose a few and often used template mappings and implement them without introducing new dependencies. The implementation should look like simple replacements.

PR #753, opened 2018-11-07
fix(docs): Update installation instructions to link to docs.aws.amazon.com

TheSriram:


docs/installation.rst
```diff
@@ -1,244 +1,15 @@
...
--  `Usage <usage.rst>`__
--  `Packaging and deploying your application <deploying_serverless_applications.rst>`__
--  `Advanced <advanced_usage.rst>`__
+==============
+Installation
+==============
+
+Installation Instructions for AWS SAM CLI are linked below.
+
+-  `Pip`_
```
jfuss: Can we put Pip last? I think we want customers to use Pip as a last effort and use the other options first.

TheSriram: Yes, thats a good suggestion!

Issue #568, opened 2018-07-21
Debugging dotnet core apps with the SAM CLI


burck1:
Currently the sam cli does not yet support debugging the dotnet core 2.0/2.1 runtimes with vscode. After  #565 / #281 has been merged (thanks @jfuss and @austinlparker), I think it should be pretty straightforward to add in debugging support. I've been using vscode for dotnet core + lambda development a lot recently and this feature would be be very useful.

The basic requirements to attach a remote debugger to dotnet core from the C# extension for vscode is outlined pretty well here: https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes. The main requirement is that [VSDBG](https://github.com/Microsoft/MIEngine/wiki/Offroad-Debugging-of-.NET-Core-on-Linux---OSX-from-Visual-Studio) is setup within the container. It handles sending and receiving the _debug adapter protocol_ messages to the remote debugger.

> Visual Studio can now communicate using the debug adapter protocol which is used by Visual Studio, Visual Studio Code and Visual Studio for Mac. This protocol is now used for debugging cross-platform .NET Core applications.

After that is setup, it's just a matter of [configuring vscode's launch.json](https://github.com/OmniSharp/omnisharp-vscode/wiki/Attaching-to-remote-processes#configuring-launchjson) to use pipeTransport.

My plan would be to wait until #565 is merged before starting work on this. Thoughts?

(also related to #500)

sanathkr:
Yaay! This is awesome! Sounds like a plan. 

I have a few questions: 

- will this method of using vsdbg work across all Visual studio products?

- Does debugging require a change in the way function code is authored or compiled? 

- does this work for both dotnetcore 2.0 and 2.1?

- Do you know if this requires a way in which the dotnetcore runtime is started? ie. Would it require a change to the Docker container?

Looking forward to working with you on this!

austinlparker:
Good to see someone else taking this up, here's some notes I've got from looking into this already.

1. Ought to be possible to map VSDBG from the host into the container in a similar way to how Delve is being mapped in for Golang debugging.
2. VS Code and VS both use VSDBG as the remote C# debugger.
3. The user's function code will need to be compiled in Debug mode (with embedded PDB's) for the best debugging experience, but this is where things get tricky...

[...]

PR #759, opened 2018-11-12
Request for Comments: Debugging dotnetcore Lambda Functions

ndobryanskyy:
### Original Issue
#568   - Feature request

### Design document
⚡ Rendered document is available [here](https://github.com/ndobryanskyy/aws-sam-cli/blob/dotnetcore-debugging-design/designs/dotnetcore-debugging.md)

Doc is updated and utilizes new `docker ps` filter approach. @jfuss, @TheSriram, please take a look on it and my updated [POC](https://github.com/ndobryanskyy/dotnetcore-aws-local-debugging-poc) -as always comments and questions are welcome! 

If you are okay with the new approach, the only thing left is merging PR to `lamci/lambda`.

As requested I provide design document and example which uses modified runner from [PR](https://github.com/lambci/docker-lambda/pull/130) which I would also appreciate reviewing


designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+And let's break it down to the checklist
+
+* Put the `vsdbg` on the Docker container;
+* Add _"wait for the debugger"_ flag in runner program and implement corresponding functionality;
+* Implement corresponding container entry point override in SAM to support invoking the runner with this flag **on**.
+* Provide instructions for the customer of how to configure VS Code and VS to attach.
+
+### What will be changed
+
+Runner program for .NET core [2.0](https://github.com/lambci/docker-lambda/blob/master/dotnetcore2.0/run/MockBootstraps/Program.cs) and [2.1](https://github.com/lambci/docker-lambda/blob/master/dotnetcore2.1/run/MockBootstraps/Program.cs) in another [repo](https://github.com/lambci/docker-lambda) to support waiting for the debugger to attach. Already have [PR](https://github.com/lambci/docker-lambda/pull/127) (guys and @mhart - reviewers wanted :)). I think we should provide new `sam local` flag  `--container-name` which will assign desired name to the container with the Lambda runtime to simplify `docker exec` experience from VS Code later.
```
TheSriram: Providing a container name is definitely something interesting to do, instead of auto generated names like `quirky_turtle` etc. I fully agree that docker exec becomes easier to do.

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+
+### User Experience Walkthrough
+
+##### 1. Getting the debugger locally
+
+At first, user should be able to get and install `vsdbg` on their machine easily. For this we should provide the script for them to invoke, which will effectively run the .NET Lambda runtime Docker container with mounted via `bind` path to get the debugger on the **host** machine. In this container following: `curl -sSL https://aka.ms/getvsdbgsh | bash /dev/stdin -v latest -l /vsdbg`  should be run to get and install required debugger (specifically built for this container). Later SAM will mount this same folder with the debugger to the target running container. 
+
+Instead of script we can certainly provide manual instructions of how to do the above stuff step-by-step as it is now done for golang debugger. _Feedback appreciated_.
+
+Customer should go through this step **only once** to set up the debugger.
```
TheSriram: is my understanding correct, that we just need to be able to download the debugger to the host machine. I dont think there is any setup required, just a mount into the docker container right?

Yeah in that case, just downloading the ```vsdbg``` debugger onto one path., only once is enough.

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+##### 3. Attaching
+
+Customer will use ready-to-go `launch.json` debug configuration for VS Code, fill in the container name for the `docker exec` and supply this name to SAM.
+
+_For example:_
+
+```sh
+sam local invoke --event event.json -d 6000 --debugger-path ~/vsdbg --container-name debugger "MyFunction"
```
TheSriram: will the codeuri in the template.yaml be any different, if the app is published under debug mode?

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+##### Container name caveat
+
+With this `launch.json` in place the only caveat left is the container name to step into. As for now, SAM will not assign consistent names for started containers, and neither it provides the way for the user to supply that name. So my proposal is to create _new flag_ for `sam local` command -  `--container-name` which will provide the user a way to specify the name for the container when it runs. This is a required configuration since user needs consistent naming to have usable attaching experience. SAM will use provided name and add it to `container.start()` configuration.
+
+### Open questions
+
+1. `--container-name` - does this option make sense to you, guys, maybe you have some thoughts of how to it better without exposing the ability to assign the name to the lambda container. As for me, I think this option is more than handy, and does not uncover additional implementation details, as the user is now already well aware, that his function is invoked inside Docker container. Maybe we want to have some environment variable too? Will we include that in `.samrc`?
```
TheSriram: the one thing to be aware about if we are including this in samrc is, how do we differentiate which container will be used when. Lets say, I'm debugging 3 functions at once, (3 different languages), doing a local invoke in another. All of these should work seamlessly.

If I were to assign a container name in .samrc, how would sam know which invoke maps to that container name?

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+### Open questions
+
+1. `--container-name` - does this option make sense to you, guys, maybe you have some thoughts of how to it better without exposing the ability to assign the name to the lambda container. As for me, I think this option is more than handy, and does not uncover additional implementation details, as the user is now already well aware, that his function is invoked inside Docker container. Maybe we want to have some environment variable too? Will we include that in `.samrc`? 
+
+2. A bit off-topic, but still, has someone else encountered the problem with python 3.7.1 ? As it now does not flush the `stdout` and `stderr` from the running container immediately but rather after SAM ends the invocation, and user does not see _"waiting for the debugger to attach..."_ message. 
+
+   _From SAM code - container.py:_
+
+   ```python
+    for frame_type, data in output_itr:
```
TheSriram: Yeah, I believe this : https://github.com/awslabs/aws-sam-cli/pull/729

We do have unbuffered streams, whenever a debug port is set explicitly. but in this case there is no debugger port. because of the .net debugging works here, so we may need to slightly rethink that.

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,196 @@
...
+                   # Frame type 2 is stderr data.
+                   stderr.write(data)
+                   # with this in place, everything works fine
+                   stderr.flush()
+   ```
+
+   As you see above with flush, everything works. Without it all is buffered, help needed, maybe this [link](https://docs.python.org/3/using/cmdline.html#cmdoption-u) will give something (see note for python 3.7),
+
+3. Help needed in investigation of how to adopt that approach for VS 2017. See this [link](https://github.com/Microsoft/MIEngine/wiki/Offroad-Debugging-of-.NET-Core-on-Linux---OSX-from-Visual-Studio) for some information. Maybe we can try to use Docker as a pipe there too.
+4. Docker in VS Code reports _"The pipe program 'docker' exited unexpectedly with code 137."_ after stopping or hitting continue which leads to quitting the program. It seems, that 137 is **killed** exit code, which seems a bit strange, maybe you guys, have some ideas on that? Haven't investigated yet.
```
TheSriram: This is strange and definitely needs to be investigated.

designs/dotnetcore-debugging.md
ndobryanskyy: Yes, I've meant the path of getting and installing it via Docker. User would have to go along it only once. Just wanted to point that out.

designs/dotnetcore-debugging.md
ndobryanskyy: Yes, I think is one can come very handy

designs/dotnetcore-debugging.md
ndobryanskyy: Yes, nice point, seems like this is a not an option for the `.samrc` then?

designs/dotnetcore-debugging.md
ndobryanskyy: And, actually, setup is required, because we are just downaloding `getvsdbg.sh` which needs to be executed in order to have `vsdbg` built for specific system. That is why we are doing that through the Docker container of the actual  runtime, on which debugging will occur.

designs/dotnetcore-debugging.md
mikemorain: Perhaps we could have a slightly more complex configuration section in .samrc which specifies names for each runtime's container instance? This would ultimately be the most flexible option, and should be pretty easy to implement. Thoughts?

sanathkr: I think `--container-name` is a short term solution until we have a robust IDE integration which could query `docker ps` for the generated container name and connect the debugger to it. So I don't want to go too deep into this feature

ndobryanskyy: By the way, have you guys checked out [this](https://github.com/aws/aws-lambda-dotnet/tree/master/Tools/LambdaTestTool) solution? I really like how they utilized web server.

And for the auto-attach part, yes deep IDE integration could  solve the problem

ndobryanskyy: No container name flag is required any more. Resolving this.

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,233 @@
+# .NET Core 2.0 / 2.1 Debugging
+
+This is a design for the .NET Core 2.0 and 2.1 missing debugging feature.
+
+### What is the problem
+
+Currently SAM CLI does not provide debugging support for .NET Core 2.0 and 2.1 because `dotnet` command does **not** support _start and wait for the debugger_ configuration out of the box and neither does current AWS Lambda runner ([2.0](https://github.com/lambci/docker-lambda/tree/master/dotnetcore2.0) / [2.1](https://github.com/lambci/docker-lambda/tree/master/dotnetcore2.1)) has debugging support flag or anything.
```
TheSriram: There is some talk here that it might make it to 3.0

https://github.com/dotnet/core-setup/issues/198

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,233 @@
...
+2. Provide ready-to-go `launch.json` configuration for attaching to the Lambda function via VS Code.
+3. Customer should be able to easily debug .NET Core 2.0 and 2.1 apps on VS Code (minimal requirement).
+
+### Out-of-Scope
+
+Visual Studio 2017 (can be a target, but it has some other very different Docker container debugging support - needs additional investigation). But some trick with `launch.json` [might be possible](https://github.com/Microsoft/MIEngine/wiki/Offroad-Debugging-of-.NET-Core-on-Linux---OSX-from-Visual-Studio). Help on this area is highly appreciated.
+
+Rider support also needs separate investigation, as it does not support `vsdbg` by any means (licensing [issue](https://github.com/dotnet/core/issues/505)) and therefore they have [their own](https://blog.jetbrains.com/dotnet/2017/02/23/rider-eap-18-coreclr-debugging-back-windows/) debugger implementation and possibly UX around .NET Core remote debugging. If support is required - I think we should open another issue.
```
TheSriram: Rider may be using a ssh based transport, so we need to be extensible of adding that.

designs/dotnetcore-debugging.md
```diff
@@ -0,0 +1,233 @@
...
+Waiting for the debugger to attach...
+```
+
+After that user should just conveniently click the _start debugging_ button with our **".NET Core Docker Attach"** configuration. 
+
+## Implementation
+
+### CLI Changes
+
+None required. We would happily use `debugger_port` and `debugger_path` like other runtimes.
```
TheSriram: This is brilliant. 💯

designs/dotnetcore-debugging.md
ndobryanskyy: Yes, I've mentioned this issue in the document too. But I don't think it has some priority there now

designs/dotnetcore-debugging.md
ndobryanskyy: The worst thing, is that they are not using `vsdbg` 😞 because of the licensing

TheSriram: aaaah. :|

PR #766, opened 2018-11-14
Implementation of "sam build" CLI command

sanathkr:
*Description of changes:*
Implementation of #743 - `sam build` command.

Implementation is complete. It includes unit and integration tests.

sambuild


samcli/commands/build/command.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@click.command("build", help=HELP_TEXT, short_help="Build your Lambda function code")
+@click.option('--build-dir', '-b',
+              default="build",
+              type=click.Path(),
+              help="Path to a folder where the built artifacts will be stored")
+@click.option("--source-root", "-s",
+              default=".",
+              type=click.Path(),
```
jfuss: `click.Path()` by default allows files and directories. Do we want to constrain this or the `--build-dir` to a directory? Should we have click resolve the path as well? What about the writable or readable checks?

https://click.palletsprojects.com/en/6.x/api/#click.Path

samcli/commands/build/command.py
```diff
@@ -0,0 +1,73 @@
...
+@pass_context
+def cli(ctx,
+        build_dir,
+        source_root,
+        native,
+        clean,
+        template):
+    # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
+
+    do_cli(template, source_root, build_dir, clean, native)  # pragma: no cover
```
jfuss: nit: can we match param order between `cli` and `do_cli` methods

samcli/commands/build/command.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Implementation of the ``cli`` method
+    """
+
+    LOG.debug("'build' command is called")
+
+    with BuildContext(template,
+                      source_root,
+                      build_dir,
+                      clean=True,   # TODO: Forcing a clean build for testing. REmove this
```
jfuss: This could be a good default to make sure the builds are always 'correct' and allow customers to optimize the speed over 'clean'.

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,94 @@
...
+        return self._use_container
+
+    @property
+    def output_template_path(self):
+        return os.path.join(self._build_dir, "built-template.yaml")
+
+    def _setup_build_dir(self):
+
+        # Get absolute path
+        self._build_dir = os.path.abspath(self._build_dir)
```
jfuss: Still stuck on os.path I see... ;)

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if not os.path.exists(self._build_dir):
+            # TODO: What permissions should I apply to this directory?
```
jfuss: I was going to ask this question. Its possible that this will need execute permissions within the files and the current user should have read/write (I would say). I don't think group or global permissions are needed at all for the build directory. 700?

On the same note, how do we ensure that the files we write or produce in this directory have correctly scoped permissions? Is that handled in the aws_lambda_builders module?

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,94 @@
...
+            os.mkdir(self._build_dir)
+
+        if os.listdir(self._build_dir) and self._clean:
+            # Build folder contains something inside. Clear everything.
+            shutil.rmtree(self._build_dir)
```
jfuss: Failure cases that should be handled?

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,94 @@
...
+        self._build_dir = self._setup_build_dir()
+
+        return self
+
+    def __exit__(self, *args):
+        pass
+
+    @property
+    def function_provider(self):
+        return self._function_provider
```
jfuss: Should we go one level further and just provide a class that conforms to a provider instead of a specific one?

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,137 @@
...
+class ApplicationBuilder(object):
+    """
+    Class to build an entire application. Currently, this class builds Lambda functions only, but there is nothing that
+    is stopping this class from supporting other resource types. Building in context of Lambda functions refer to
+    converting source code into artifacts that can be run on AWS Lambda
+    """
+
+    _builder_capabilities = {
+        "python2.7": {
+            "language": "python",
```
jfuss: Why not make this content a namedtuple instead?

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,137 @@
...
+
+        Returns
+        -------
+        dict
+            Returns the path to where each resource was built as a map of resource's LogicalId to the path string
+        """
+
+        result = {}
+
+        for lambda_function in self.function_provider.get_all():
```
jfuss: Is there a way I can specify a specific function to build? I am thinking of a use case for fast dev cycles on one of the many functions I define.

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,137 @@
...
+            Map of LogicalId of a resource to the path where the the built artifacts for this resource lives
+
+        Returns
+        -------
+        dict
+            Updated template
+        """
+        # TODO: Move this method to a "build provider" or some other class. It doesn't really fit within here.
+
+        for logical_id, resource in template_dict.get("Resources", {}).items():
```
jfuss: Is it expected that any Transforms are run before this point (aka SAM)?

We also do this all over the place. I wonder if there is some abstract we can do which makes this iterating much easier to read/do. Might be out of scope here but something that popped in my head so writing it down to capture.

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,137 @@
...
+
+            if logical_id not in built_artifacts:
+                # this resource was not built. So skip it
+                continue
+
+            resource_type = resource.get("Type")
+            if resource_type == "AWS::Serverless::Function":
```
jfuss: We need a better way to do these things.

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,137 @@
...
+                template_dict["Resources"][logical_id]["Properties"]["Code"] = built_artifacts[logical_id]
+
+        return template_dict
+
+    def _build_function(self, function_name, codeuri, runtime):
+
+        capability = self._builder_capabilities[runtime]
+
+        # Create the arguments to pass to the builder
+        code_dir = os.path.normpath(os.path.join(self.source_root, codeuri))
```
jfuss: Pathlib?

samcli/commands/_utils/options.py
```diff
@@ -0,0 +1,81 @@
+
+import os
+import click
+
+_TEMPLATE_OPTION_DEFAULT_VALUE = "template.[yaml|yml]"
+_TEMPLATE_SEARCH_PATHS = [
```
jfuss: Instead. What if you keep the default and have commands pass in the extra resolving that should take priority? 

This will simplify resolving and will allow the commands to control the resolve order. For example, package should only care about build and deploy should only care about package (templates). If they don't exist then some other flow would be triggered anyways.

samcli/commands/_utils/options.py
```diff
@@ -0,0 +1,81 @@
...
+    os.path.join(".sam", "build", "template.yaml"),
```
jfuss: Consider making this directory bound instead of template. 

Also pathlib :)

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,113 @@
...
+        self._docker_network = docker_network
+        self._skip_pull_image = skip_pull_image
+
+        self._function_provider = None
+        self._template_dict = None
+        self._app_builder = None
+        self._container_manager = None
+
+    def __enter__(self):
+        self._template_dict = self._get_template_data(self._template_file)
```
jfuss: Be very cautious here. This template_dict is not run through any transforms or resolves parameters. This could lead to confusion if someone expects one other the other.

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,113 @@
...
+    @staticmethod
+    def _get_template_data(template_file):
+        """
+        Read the template file, parse it as JSON/YAML and return the template as a dictionary.
+
+        :param string template_file: Path to the template to read
+        :return dict: Template data as a dictionary
+        :raises InvokeContextException: If template file was not found or the data was not a JSON/YAML
+        """
+        # TODO: This method was copied from InvokeContext. Move it into a common folder
```
jfuss: Don't forget about this.

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,209 @@
...
+        # artifacts directory will be created by the builder
+        artifacts_dir = os.path.join(self.build_dir, function_name)
+
+        # TODO: Create this
+        scratch_dir = None
```
jfuss: Is this needed?

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def _build_function_on_container(self,
+                                     capability,
+                                     source_dir,
+                                     artifacts_dir,
+                                     manifest_path,
+                                     runtime):
+
+        # If we are printing debug logs in SAM CLI, the builder library should also print debug logs
+        log_level = LOG.getEffectiveLevel()
```
jfuss: Is it possible to set this higher in the flow? This is a common pattern we should be doing (for example SAM).

samcli/local/docker/lambda_build_container.py
```diff
@@ -0,0 +1,132 @@
...
+    def _make_request(language,
+                      dependency_manager,
+                      application_framework,
+                      container_dirs,
+                      manifest_file_name,
+                      runtime,
+                      optimizations,
+                      options):
+
+        return json.dumps({
```
jfuss: We should be using OrderedDicts instead. Probably ok but we have been burnt by py2 and py3 differences in the past.

samcli/commands/build/command.py
sanathkr: Done

samcli/commands/build/command.py
sanathkr: Perfect. Made the change

samcli/commands/build/command.py
sanathkr: We don't support a 'dirty' build yet. So I am removing this flag altogether from CLI interafce.

samcli/commands/build/build_context.py
sanathkr: changed to use Pathlib

samcli/commands/build/build_context.py
sanathkr: This directory is not zipped and passed to Lambda, instead it is only the artifacts that are passed. By default artifacts get written with 755 on directory and 644 on files. But in future we might apply a build action that just verifies this

samcli/commands/build/build_context.py
sanathkr: Failure can happen when we don't have permissions to write/read this dir. I'd just let the process raise an exception because this shouldn't happen usually

samcli/commands/build/build_context.py
sanathkr: Possibly. But this is a good easy place to start

samcli/lib/build/app_builder.py
sanathkr: No reason. Dict was the quickest. I am changing this to something more general.

samcli/lib/build/app_builder.py
sanathkr: Not at the moment. It might be a good future option

samcli/commands/_utils/options.py
```diff
@@ -0,0 +1,118 @@
...
+                        default=_TEMPLATE_OPTION_DEFAULT_VALUE,
+                        type=click.Path(),
+                        envvar="SAM_TEMPLATE_FILE",
+                        callback=partial(get_or_default_template_file_name, include_build=include_build),
+                        show_default=True,
+                        help="AWS SAM template file")
+
+
+def docker_common_options(f):
+    for option in reversed(docker_click_options()):
```
TheSriram: Q: What does reversed really give us?

samcli/commands/build/build_context.py
```diff
@@ -0,0 +1,123 @@
...
+
+from samcli.local.docker.manager import ContainerManager
+from samcli.commands.local.lib.sam_function_provider import SamFunctionProvider
+from samcli.commands._utils.template import get_template_data
+from samcli.commands.exceptions import UserException
+
+
+class BuildContext(object):
+
+    _BUILD_DIR_PERMISSIONS = 0o755
```
TheSriram: Maybe a comment on why these permissions?

samcli/commands/build/command.py
```diff
@@ -0,0 +1,151 @@
...
+LOG = logging.getLogger(__name__)
+
+
+HELP_TEXT = """
+Use this command to build your AWS Lambda Functions source code to generate artifacts that can run on AWS Lambda.\n
+\b
+Supported Resource Types
+------------------------
+1. AWS::Serverless::Function\n
+2.AWS::Lambda::Function\n
```
TheSriram: Nit: spacing

samcli/commands/build/command.py
```diff
@@ -0,0 +1,151 @@
...
+\b
+To build on your workstation, run this command in folder containing
+SAM template. Built artifacts will be written to .aws-sam/build folder
+$ sam build\n
+\b
+To build inside a AWS Lambda like Docker container
+$ sam build --use-container
+\b
+To build & run your functions locally
+$ sam build && sam local invoke
```
TheSriram: I love how simple this is going to make things. 🎆 🍾

samcli/commands/build/command.py
```diff
@@ -0,0 +1,151 @@
...
+Next Steps
+==========
+[*] Invoke Function: sam local invoke -t {template}
+[*] Package: sam package --template-file {template}
+            """.format(artifacts_dir=os.path.relpath(ctx.build_dir),
+                       template=os.path.relpath(ctx.output_template_path))
+
+            click.secho("\nBuild Succeeded", fg="green")
+
+            click.secho(msg, fg="yellow")
```
TheSriram: Colors 👍

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,305 @@
...
+from samcli.local.docker.lambda_build_container import LambdaBuildContainer
+from aws_lambda_builders.builder import LambdaBuilder
+from aws_lambda_builders.exceptions import LambdaBuilderError
+from aws_lambda_builders import RPC_PROTOCOL_VERSION as lambda_builders_protocol_version
+
+
+LOG = logging.getLogger(__name__)
+
+
+class UnsupportedRuntimeException(Exception):
```
TheSriram: nit: these exceptions could go into its own file

samcli/lib/build/app_builder.py
TheSriram: Good point, maybe we need a note here, thats transforms are done by now?

samcli/lib/build/app_builder.py
TheSriram: Dictionaries for code branching, makes code look cleaner as well :)
store functions as values to a dictionary key "AWS::Serverless::Function", "AWS::Lambda::Function"

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,305 @@
...
+        # Create the arguments to pass to the builder
+
+        # Code is always relative to the given base directory.
+        code_dir = str(pathlib.Path(self._base_dir, codeuri).resolve())
+
+        # artifacts directory will be created by the builder
+        artifacts_dir = str(pathlib.Path(self._build_dir, function_name))
+
+        with osutils.mkdir_temp() as scratch_dir:
+            manifest_path = self._manifest_path_override or os.path.join(code_dir, config.manifest_name)
```
TheSriram: the manifest would be requirements.txt here? for python ie

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,305 @@
...
+
+            # By default prefer to build in-process for speed
```
TheSriram: +1 for speedy builds!

samcli/lib/build/app_builder.py
```diff
@@ -0,0 +1,305 @@
...
+            # Report this as a crash
+            LOG.debug("Builder crashed")
+            raise
+
+        if "error" in response:
+            error = response.get("error", {})
+            err_code = error.get("code")
+            msg = error.get("message")
+
+            if 400 <= err_code < 500:
```
TheSriram: errorcode to exception mapping, can be encapsulated in a dict as well.

```
{ 
"505':UnspoortedError,
"-400": BuildError
}
```

Helps avoid code branching logic using just if statements.

samcli/lib/utils/osutils.py
```diff
@@ -0,0 +1,82 @@
+"""
```
TheSriram: I like this osutils.py separation!

samcli/lib/utils/osutils.py
```diff
@@ -0,0 +1,82 @@
...
+import sys
+import os
+import shutil
+import tempfile
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def mkdir_temp(mode=0o755):
```
TheSriram: do we want to make a comment, on why we chose these permissions?

samcli/local/docker/lambda_build_container.py
```diff
@@ -0,0 +1,164 @@
...
+                    "language": language,
+                    "dependency_manager": dependency_manager,
+                    "application_framework": application_framework
+                },
+                "source_dir": container_dirs["source_dir"],
+                "artifacts_dir": container_dirs["artifacts_dir"],
+                "scratch_dir": container_dirs["scratch_dir"],
+
+                # Path is always inside a Linux container. So '/' is valid
+                "manifest_path": "{}/{}".format(container_dirs["manifest_dir"], manifest_file_name),
```
TheSriram: we can use posixpath.sep or something like that too, if we dont want to use '/' explicitly.

samcli/local/docker/lambda_build_container.py
TheSriram: I think this should be fine, in this particular case.

samcli/local/docker/lambda_build_container.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        base = "/tmp/samcli"
+        result = {
+            "source_dir": "{}/source".format(base),
+            "artifacts_dir": "{}/artifacts".format(base),
+            "scratch_dir": "{}/scratch".format(base),
+            "manifest_dir": "{}/manifest".format(base)
+        }
+
+        if pathlib.PurePath(source_dir) == pathlib.PurePath(manifest_dir):
```
TheSriram: Nice check!

samcli/local/init/templates/cookiecutter-aws-sam-hello-python/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,7 +16,7 @@ Resources:
     HelloWorldFunction:
         Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
         Properties:
-            CodeUri: hello_world/build/
+            CodeUri: hello_world/
```
TheSriram: Yaasss!

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -0,0 +1,118 @@
...
+        LOG.info("Invoking built function '{}'", function_logical_id)
+
+        cmdlist = [self.cmd, "local", "invoke", function_logical_id, "-t", str(template_path), "--no-event",
+                   "--parameter-overrides", overrides]
+
+        process = subprocess.Popen(cmdlist, stdout=subprocess.PIPE)
+        process.wait()
+
+        process_stdout = b"".join(process.stdout.readlines()).strip().decode('utf-8')
+        print(process_stdout)
```
TheSriram: we can probably remove the prints?

tests/integration/testdata/buildcmd/Python/requirements.txt
```diff
@@ -0,0 +1,5 @@
+# These are some hard packages to build. Using them here helps us verify that building works on various platforms
+
+numpy~=1.15
+cryptography~=2.4
+Jinja2~=2.10
```
TheSriram: didnt know Jinja2 was hard to build. #TIL

tests/unit/local/docker/test_lambda_build_container.py
```diff
@@ -0,0 +1,157 @@
...
+                                                    "language",
+                                                    "dependency",
+                                                    "application",
+                                                    container_dirs,
+                                                    "manifest_file_name",
+                                                    "runtime",
+                                                    "optimizations",
+                                                    "options")
+
+        self.maxDiff = None  # Print whole json diff
```
TheSriram: prolly dont need this?

samcli/lib/build/app_builder.py
sanathkr: It is possible to do something more abstract, but I don't see much value in it at the moment.

samcli/commands/build/build_context.py
sanathkr: Agreed, that's why we access only thru providers.

samcli/lib/build/app_builder.py
sanathkr: No, this is specifically passed to the builder library running inside a container. So think of this as an equivalent of `--debug` flag passed to this library.

samcli/local/docker/lambda_build_container.py
sanathkr: Probably, but that actually sets the wrong expectation. JSON dicts are expected to be unordered. So I think it is better not using ordered dict here.

samcli/commands/_utils/options.py
sanathkr: It allows us to list options in the code in the order it will appear in help text. Just for our convenience.

samcli/commands/build/build_context.py
sanathkr: Done

samcli/commands/build/command.py
sanathkr: good catch

samcli/commands/build/command.py
sanathkr: I know! I am super excited :)

samcli/commands/build/command.py
sanathkr: Yes!!

samcli/local/init/templates/cookiecutter-aws-sam-hello-python/{{cookiecutter.project_name}}/template.yaml
sanathkr: Finally :)

samcli/lib/build/app_builder.py
sanathkr: Agreed. I started with something small and expanded in scope. We should move them out later

samcli/lib/build/app_builder.py
sanathkr: yup, exactly

samcli/lib/build/app_builder.py
sanathkr: oh yeah! this will be really neat. I am going to stick with if/else for now because it offers more flexibility

samcli/lib/utils/osutils.py
sanathkr: Will do.

samcli/local/docker/lambda_build_container.py
sanathkr: ah right! I wanted to be explicit here so someone doesn't accidentaly "fix" it later

samcli/local/docker/lambda_build_container.py
sanathkr: I spent 3 hours debugging an issue that stemmed from not having this check ;)

tests/integration/buildcmd/test_build_cmd.py
sanathkr: I do want to have prints for debugging. pytest prints them only if your test fails

tests/unit/local/docker/test_lambda_build_container.py
sanathkr: No, its helpful in debugging when the diff doesn't match up

mvanbaak:
THANK YOU THANK YOU THANK YOU!

As we just started to explore SAM, we quickly bumped into the python packaging crap.
You saved my day with this PR!

PR #767, opened 2018-11-14
API Gateway path validator

matteofigus:
*Description of changes:*

By using SAM local, I discovered that while everything was working locally, the `deploy` was failing because of an invalid route path (it contained an invalid char `~`). I found the experience confusing and decided to create a Pull Request for alerting the user early in the development rather on the deploy phase.

### DIFF
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -6,6 +6,7 @@ import base64
 
 from flask import Flask, request
 
+from samcli.commands.exceptions import UserException
 from samcli.local.services.base_local_service import BaseLocalService, LambdaOutputParser, CaseInsensitiveDict
 from samcli.local.lambdafn.exceptions import FunctionNotFound
 from samcli.local.events.api_event import ContextIdentity, RequestContext, ApiGatewayLambdaEvent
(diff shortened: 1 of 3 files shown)


jfuss:
@matteofigus Thanks for taking the time to add this validation but the CLI isn't the correct place for this. Instead, this should be apart of [SAM](https://github.com/awslabs/serverless-application-model) instead. While I agree that this is something that should be captured before you deploy, adding these checks strictly within the CLI and not the Spec leads this validation only happening within the CLI and not any other tools like [cfn-lint](https://github.com/awslabs/cfn-python-lint).

Can port this over to SAM instead?

matteofigus:
> Can port this over to SAM instead?
Sounds like a good idea! Let me have a go, feel free to close here in the meanwhile.

jfuss:
> Do you think this should be only in SAM or both?

With adding it into SAM, SAM CLI will automatically get this when we upgrade the version we depend on. Before SAM CLI does template parsing to find the resources we care about (functions, apis, etc), we run the SAM Translator to make sure it is valid in terms of SAM.

matteofigus:
> With adding it into SAM, SAM CLI will automatically get this when we upgrade the version we depend on. 

Amazing, thanks for taking the time to guide me on that. I'll link the SAM PR here when ready. In the meanwhile I'm closing.

PR #774, opened 2018-11-19
Container name feature implementation

ndobryanskyy:
#759 - WIP: PR, which describes the need for new `--container-name` flag. It is arequred step for enabling .NET Core debugging. For detailed explanation see the [rendered doc](https://github.com/ndobryanskyy/aws-sam-cli/blob/dotnetcore-debugging-design/designs/dotnetcore-debugging.md).

Adds new `--container-name` flag, which enables you to assign specific name to the container with AWS Lambda runtime when it runs.

_I do still need help of where to put some help to let  the user know,  that this flag is here. Any comments on this is appreciated._

### DIFF
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -8,9 +8,6 @@ import sys
 import os
 import yaml
 
-import docker
-import requests
-
 from samcli.yamlhelper import yaml_parse
 from samcli.commands.local.lib.local_lambda import LocalLambdaRunner
 from samcli.commands.local.lib.debug_context import DebugContext
@@ -44,7 +41,7 @@ class InvokeContext(object):
     This class sets up some resources that need to be cleaned up after the context object is used.
     """
 
-    def __init__(self,
+    def __init__(self,  # pylint: disable=R0914
                  template_file,
                  function_identifier=None,
                  env_vars_file=None,
@@ -57,7 +54,8 @@ class InvokeContext(object):
                  debug_args=None,
                  debugger_path=None,
                  aws_region=None,
-                 parameter_overrides=None):
+                 parameter_overrides=None,
+                 container_name=None):
         """
         Initialize the context
 
@@ -92,6 +90,8 @@ class InvokeContext(object):
             AWS region to use
         parameter_overrides dict
             Values for the template parameters
+        container_name str
+            Docker container name
 
         """
         self._template_file = template_file
@@ -107,12 +107,14 @@ class InvokeContext(object):
         self._debug_args = debug_args
         self._debugger_path = debugger_path
         self._parameter_overrides = parameter_overrides or {}
+        self._container_name = container_name
 
         self._template_dict = None
         self._function_provider = None
         self._env_vars_value = None
         self._log_file_handle = None
         self._debug_context = None
+        self._container_manager = None
 
     def __enter__(self):
         """
@@ -132,7 +134,20 @@ class InvokeContext(object):
                                                       self._debug_args,
                                                       self._debugger_path)
 
-        self._check_docker_connectivity()
+        container_manager = self._get_container_manager(self._docker_network, self._skip_pull_image)
+
+        if not container_manager.is_docker_reachable():
+            raise InvokeContextException("Running AWS SAM projects locally requires Docker. Have you got it installed?")
+
+        if self._container_name:
+            if not container_manager.is_valid_container_name(self._container_name):
+                raise InvokeContextException("'{}' is not a valid Docker container name"
+                                             "only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.".format(self._container_name))
+
+            if container_manager.is_container_name_taken(self._container_name):
+                raise InvokeContextException("'{}' Docker container name is already taken".format(self._container_name))
+
+        self._container_manager = container_manager
 
         return self
 
@@ -180,13 +195,11 @@ class InvokeContext(object):
             locally
         """
 
-        container_manager = ContainerManager(docker_network_id=self._docker_network,
-                                             skip_pull_image=self._skip_pull_image)
-
-        lambda_runtime = LambdaRuntime(container_manager)
+        lambda_runtime = LambdaRuntime(self._container_manager)
         return LocalLambdaRunner(local_runtime=lambda_runtime,
                                  function_provider=self._function_provider,
                                  cwd=self.get_cwd(),
+                                 container_name=self._container_name,
                                  env_vars_values=self._env_vars_value,
                                  debug_context=self._debug_context,
                                  aws_profile=self._aws_profile,
@@ -322,6 +335,25 @@ class InvokeContext(object):
         return open(log_file, 'wb')
 
     @staticmethod
+    def _get_container_manager(docker_network, skip_pull_image):
+        """
+        Creates a ContainerManager
+
+        Parameters
+        ----------
+        docker_network str
+            Docker network identifier
+        skip_pull_image bool
+            Should the manager skip pulling the image
+
+        Returns
+        -------
+        samcli.local.docker.manager.ContainerManager
+            Object representing Docker container manager
+        """
+        return ContainerManager(docker_network_id=docker_network, skip_pull_image=skip_pull_image)
+
+    @staticmethod
     def _get_debug_context(debug_port, debug_args, debugger_path):
         """
         Creates a DebugContext if the InvokeContext is in a debugging mode
@@ -360,21 +392,3 @@ class InvokeContext(object):
             debugger_path = str(debugger)
 
         return DebugContext(debug_port=debug_port, debug_args=debug_args, debugger_path=debugger_path)
-
-    @staticmethod
-    def _check_docker_connectivity(docker_client=None):
-        """
-        Checks if Docker daemon is running. This is required for us to invoke the function locally
-
-        :param docker_client: Instance of Docker client
-        :return bool: True, if Docker is available
-        :raises InvokeContextException: If Docker is not available
-        """
-
-        docker_client = docker_client or docker.from_env()
-
-        try:
-            docker_client.ping()
-        # When Docker is not installed, a request.exceptions.ConnectionError is thrown.
-        except (docker.errors.APIError, requests.exceptions.ConnectionError):
-            raise InvokeContextException("Running AWS SAM projects locally requires Docker. Have you got it installed?")
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -110,6 +110,10 @@ def invoke_common_options(f):
                           "pairs. Use the same format as the AWS CLI, e.g. 'ParameterKey=KeyPairName,"
                           "ParameterValue=MyKey ParameterKey=InstanceType,ParameterValue=t1.micro'"),
 
+        click.option('--container-name',
+                     help="When specified, Lambda function container will start with this name.",
+                     envvar="SAM_DOCKER_CONTAINER_NAME"),
+
         click.option('--debug-port', '-d',
                      help="When specified, Lambda function container will start in debug mode and will expose this "
                           "port on localhost.",
--- a/samcli/commands/local/invoke/cli.py
+++ b/samcli/commands/local/invoke/cli.py
@@ -44,18 +44,18 @@ STDIN_FILE_NAME = "-"
 @pass_context  # pylint: disable=R0914
 def cli(ctx, function_identifier, template, event, no_event, env_vars, debug_port,
         debug_args, debugger_path, docker_volume_basedir, docker_network, log_file, skip_pull_image, profile, region,
-        parameter_overrides):
+        parameter_overrides, container_name):
 
     # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
 
     do_cli(ctx, function_identifier, template, event, no_event, env_vars, debug_port, debug_args, debugger_path,
            docker_volume_basedir, docker_network, log_file, skip_pull_image, profile, region,
-           parameter_overrides)  # pragma: no cover
+           parameter_overrides, container_name)  # pragma: no cover
 
 
 def do_cli(ctx, function_identifier, template, event, no_event, env_vars, debug_port,  # pylint: disable=R0914
            debug_args, debugger_path, docker_volume_basedir, docker_network, log_file, skip_pull_image, profile,
-           region, parameter_overrides):
+           region, parameter_overrides, container_name):
     """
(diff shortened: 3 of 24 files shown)


sthulb:
I looked at this issue yesterday. I was thinking of going further than just names for containers — I was also thinking of adding labels too (perhaps ones for another PR)

ndobryanskyy:
@sthulb yes we can add this functionality, but I don't see practical use-case for this yet.
`--container-name` is specifically added to enable .NET Core debugging, see - #568 .

And what do you think about the changes, do you have any remarks, comments?

sthulb:
from a personal point of view, I’d rather see the container name be automatic, so every container gets a name, probably based on the resource name. 

It would still give you the ability to have an explicit name to allow for debugging.

ndobryanskyy:
@sthulb, you mean, that SAM should assign explicit name for each container it runs? And `--container-name` option will override this behavior?

samcli/local/docker/manager.py
```diff
@@ -33,6 +38,62 @@ def __init__(self,
...
+        bool
+            True, if Docker is available, False otherwise
+        """
+
+        try:
+            self.docker_client.ping()
+            return True
+
+        # When Docker is not installed, a request.exceptions.ConnectionError is thrown.
+        except (docker.errors.APIError, requests.exceptions.ConnectionError):
```
sanathkr: Can you add a `LOG.debug(exc_info=ex)` here so we know why it failed when `--debug` is enabled?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -131,7 +133,20 @@ def __enter__(self):
...
+
+        if not container_manager.is_docker_reachable:
+            raise InvokeContextException("Running AWS SAM projects locally requires Docker. Have you got it installed?")
+
+        if self._container_name:
+            if not container_manager.is_valid_container_name(self._container_name):
+                raise InvokeContextException("'{}' is not a valid Docker container name"
+                                             "only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.".format(self._container_name))
+
+            if container_manager.is_container_name_taken(self._container_name):
```
sanathkr: This is going to add extra overhead to the invoke call. Instead of checking here, can we let the Docker create API throw an exception, catch it there, and surface the error to customers? 

I would expect the common case to be the one where container name is available. So this will optimize for the common case.

samcli/local/docker/manager.py
```diff
@@ -64,13 +125,20 @@ def run(self, container, input_data=None, warm=False):
...
+                # Create the container in appropriate Docker network
+                container.network_id = self.docker_network_id
+                container.create()
+
+            container.start(input_data=input_data)
+        except docker.errors.APIError as api_error:
+            if api_error.status_code == 409:
+                raise DockerContainerException("'{}' Docker container name is already taken".format(container.name))
+            if api_error.is_server_error():
+                raise DockerContainerException("Something went wrong on the Docker server")
```
sanathkr: Instead of this custom error message, I would pass the actual exception message returned by Docker API - `str(ex)`. This will add more information.

samcli/local/docker/manager.py
```diff
@@ -124,3 +192,10 @@ def has_image(self, image_name):
 
 class DockerImagePullFailedException(Exception):
     pass
+
+
+class DockerContainerException(UserException):
```
sanathkr: We have a convention of not using the `UserException` inside the library classes. `UserException` is strictly a CLI interface exception. Can you instead catch the `DockerContainerException` inside the `do_cli` methods and re-raise as a UserException?

This will allow us to move the library into a separate package away from the CLI interface.

samcli/local/docker/manager.py
```diff
@@ -64,13 +125,20 @@ def run(self, container, input_data=None, warm=False):
...
+            if not container.is_created():
+                # Create the container first before running.
+                # Create the container in appropriate Docker network
+                container.network_id = self.docker_network_id
+                container.create()
+
+            container.start(input_data=input_data)
+        except docker.errors.APIError as api_error:
+            if api_error.status_code == 409:
+                raise DockerContainerException("'{}' Docker container name is already taken".format(container.name))
```
sanathkr: nit: Can we replace 'taken' with 'in use' ? Conveys the intent better

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -131,7 +133,20 @@ def __enter__(self):
...
                                                       self._debugger_path)
 
-        self._check_docker_connectivity()
+        container_manager = self._get_container_manager(self._docker_network, self._skip_pull_image)
+
+        if not container_manager.is_docker_reachable:
+            raise InvokeContextException("Running AWS SAM projects locally requires Docker. Have you got it installed?")
+
+        if self._container_name:
+            if not container_manager.is_valid_container_name(self._container_name):
```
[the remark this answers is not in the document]
ndobryanskyy: Very good idea, thanks! I'll certainly do it.

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -131,7 +133,20 @@ def __enter__(self):
                                                       self._debug_args,
                                                       self._debugger_path)
 
-        self._check_docker_connectivity()
+        container_manager = self._get_container_manager(self._docker_network, self._skip_pull_image)
```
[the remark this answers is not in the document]
ndobryanskyy: By making it local, I was  trying to simplify reading, because it would avoid `self._` on every access, but if you think that is a negligible enhancement I will assign to self here. What do you think?

samcli/local/docker/manager.py
ndobryanskyy: No problem

samcli/local/docker/manager.py
ndobryanskyy: Maybe let's LOG.debug the actual exception? I don't think custom needs to know the exact bad thing, that happened, or he does?

samcli/local/docker/manager.py
ndobryanskyy: Makes sense, I will fix that

samcli/local/docker/manager.py
ndobryanskyy: Your version aligns more closely with actual docker error message. So why not, I will do it.
image

samcli/commands/local/cli_common/invoke_context.py
ndobryanskyy: There is no problem eliminating all of the check s  upfront, and that will boost performance  in common case, agreed. I'll  remove it

My thinking for that was: "If you can avoid exceptions, do that" :)

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -300,6 +313,25 @@ def _setup_log_file(log_file):
 
         return open(log_file, 'wb')
 
+    @staticmethod
+    def _get_container_manager(docker_network, skip_pull_image):
```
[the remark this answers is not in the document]
ndobryanskyy: I've chosen that name, because of the surrounding code. It has `_get_{something}` pattern. Like `_get_debug_context()` which also creates a new instance

samcli/commands/local/cli_common/invoke_context.py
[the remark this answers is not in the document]
sanathkr: Understood. It would be more clear if you assigned to self here

samcli/commands/local/cli_common/invoke_context.py
sanathkr: Makes sense.

samcli/local/docker/manager.py
```diff
@@ -33,6 +38,62 @@ def __init__(self,
...
+        Returns
+        -------
+        bool
+            True, if name is a valid Docker container name, False otherwise
+        """
+
+        return True if ContainerManager._container_name_regex.match(name) else False
+
+    @property
+    def is_docker_reachable(self):
```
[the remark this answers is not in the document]
ndobryanskyy: Agreed 💯

samcli/local/docker/manager.py
ndobryanskyy: I've removed `is_container_name_taken` method altogether. Now it does this
```python
except docker.errors.APIError as api_error:
            raise DockerContainerException("[From Docker]\n{}".format(api_error))
```

Preceding the error message with "[From Docker]", because Docker  API error messages are pretty and informative. 

If you want I can still log.debug(exc_info=True) here also to get traceback

samcli/commands/local/cli_common/invoke_context.py
ndobryanskyy: Eliminated all checks

samcli/commands/local/cli_common/invoke_context.py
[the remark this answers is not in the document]
ndobryanskyy: It seems, that there were common options already, so I've just added callback

https://github.com/awslabs/aws-sam-cli/blob/ae7aadff8e20cb141bd73937abcd92b98bff34d4/samcli/commands/local/cli_common/options.py#L67-L70

https://github.com/awslabs/aws-sam-cli/blob/ae7aadff8e20cb141bd73937abcd92b98bff34d4/samcli/commands/local/cli_common/options.py#L52-L56

Will that do?

ndobryanskyy:
Fixed review comments and implemented integration and functional tests. @sanathkr, please, take a look, when you have time.

jfuss:
@ndobryanskyy Could you please rebase and address the merge conflicts. Our recent release brought in a bunch of changes.

I am going to review now, as well.

samcli/commands/local/cli_common/options.py
```diff
@@ -5,6 +5,8 @@
 import click
 from samcli.commands._utils.options import template_click_option, docker_click_options, parameter_override_click_option
 
+from samcli.local.docker.manager import ContainerManager
```
jfuss: I find this import strange. I understand why but the Options shouldn't need to interact with a ContainerManger. Maybe we should place the `is_valid_container_name` somewhere else?

ndobryanskyy: It can go to utils or something, but for me, ContainerManager should be the only one responsible for the whole Docker management cycle. It will semantically be better, what do you think?

jfuss: To me validation isn't part of that management cycle, it more of a attribute of the container itself not the class that manages. 
@sanathkr or @TheSriram what are your thoughts here. I just want to make sure we keep the right abstractions is all.

ndobryanskyy: @jfuss, if we treat it like an attribute we can simply inline this method here.

jfuss: I more meant on the `Container` class instead of the `ContainerManger`. This isn't a huge deal for me, just something I noticed while reading through.

jfuss: I am ok with this abstraction you have. I think your reasoning is good. I think we just have a slightly different model for what is in the Docker management cycle. I am more worried about the use of `container_name` outside of debugging than this :)

ndobryanskyy: @jfuss, okay, then 😅 As for the container_name I do agree, that it is more about debugging, then something else. I will get back to PC, consider options and get back to you really nice catch.

Also, as related question, what do you think about determining whether `DebugContext` is true with `debug_port or debugger_path` as debugging port does not make sense for don't core debugging and customer would be forced to supply unused argument to get it working

samcli/commands/local/invoke/cli.py
```diff
@@ -101,6 +102,8 @@ def do_cli(ctx, function_identifier, template, event, no_event, env_vars, debug_
         raise UserException(str(ex))
     except DockerImagePullFailedException as ex:
         raise UserException(str(ex))
+    except DockerContainerException as ex:
```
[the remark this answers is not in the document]
ndobryanskyy: Okay, I thought of that initially, but decided to stick to existing separation, as it seemed to me as intentional

samcli/commands/local/cli_common/options.py
TheSriram: It might make sense to add one more abstraction to debugContext, with the transport. Since technically dotnet debugging can be done over ssh or docker. I'm not sure about other languages.

We still want to be able support ssh based debugging too. Thoughts?

ndobryanskyy:
Guys, yesterday, after looking through your comments I've actually got hit by brilliant idea. What if we just eliminate the container name feature altogether and use some cool trick with Docker, namely `docker ps --filter`. 

### The Trick
Docker supports listing containers by using `docker ps` [command](https://docs.docker.com/engine/reference/commandline/ps/),  but this particular command not only can list all containers, but it could also filter them based on certain attributes. But what is most interesting for us is that it could filter running containers based on **published** port (and as you know,  when `-d <debugger_port>` flag is supplied to SAM CLI it will publish `debugger_port` for the running container), so using `docker ps --filter publish=<debugger_port> -q` (`-q` for returning only container ID) we could easily find id of the Lambda container started by SAM, without knowing or specifying the exact name. Pretty neat, right ?

### Real world example
Yesterday, I've also tried to use this approach for .NET Core debugging by checking out SAM `develop` branch, adding debugging support entry point there, which uses [my modified version of .NET Core lambda runner](https://github.com/lambci/docker-lambda/pull/130) with _wait for debugger_ flag support. And it worked! 🎉 

This how my `pipeTransport` section of `launch.json` look like:
```json
"pipeTransport": {
   "pipeProgram": "powershell",
   "pipeArgs": [ 
     "-c",
     "docker exec -i $(docker ps -q -f publish=6000)"
   ],
   "debuggerPath": "/tmp/lambci_debug_files/vsdbg",
   "pipeCwd": "${workspaceRoot}"
}
```
Lambda was invoked with this command: `samdev local invoke -e event.json -d 6000 --debugger-path "C:/users/ndobryanskyy/vsdbg" --skip-pull-image "SuperFunction" --region us-east`

I like that it requires nearly no changes to the SAM CLI in order to work,the only thing we need to is provide customers with this `launch.json` file. I will have time tomorrow to investigate whether it  is possible to use the same technique for VS 2017, but I'm [almost sure](https://github.com/Microsoft/MIEngine/wiki/Offroad-Debugging-of-.NET-Core-on-Linux---OSX-from-Visual-Studio),  that  the answer is **yes**.


@sanathkr @TheSriram @jfuss what do you guys think of it?
I can prepare the PR with required changes by Monday. And file another one with `ContainerManager`  improvements from this one.

sanathkr:
This is incredibly slick! I wish we had found this before you did the whole container name PR. 

+1 to take this route as it leads to a good UX

PR #775, opened 2018-11-19
Documentation updates + "sam package" defaults to built template

sanathkr:
*Issue #, if available:*

*Description of changes:*
- `sam init` Python example README changes to include `sam build`
- `sam package` command now just defaults to a built template. So you can do `sam build && sam package --s3-bucket foo` to build and then package. Don't need to manually wire the intermediate template files through.

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -18,6 +18,8 @@ from samcli.commands.exceptions import UserException
 
 class BuildContext(object):
 
+    # Build directories need not be world writable.
+    # This is usually a optimal permission for directories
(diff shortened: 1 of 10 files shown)


heitorlessa:
Wooot!!!!!!!

mvanbaak:
It just keeps getting better!

Thanks again!

mvanbaak:
```sam package``` output should still be sent to a packaged template that can be given to ```sam deploy``` right?
Or did I miss something?

sanathkr:
You are right. We need to change that as well. One step at a time :-)

PR #780, opened 2018-11-19
Getting missed commits to develop from master

TheSriram:
* Get missed commits from diverged master back over to develop
* After this commit, master and develop should be level.


sanathkr:
Funny that the diff has zero files.

Issue #790, opened 2018-11-21
[Feature Request] InlineCode support for sam local


j0lly:
Hello there,

I was thinking on taking some work on the topic as InlineCode is a great feature but a few iteration of sam local invoke would be much appreciated during code development.

thus I was planning something along the line of creating a tempfile with InlineCode string content and passing along the context instead the CodeUri path..

Not sure if you see any major issue with the approach (platform, filename (using handler)) otherwise I would happily start implement. 

The hook for the switch case seems reasonably here:
https://github.com/awslabs/aws-sam-cli/blob/1d7df8fce701711963efc32507305c2fdd84a3aa/samcli/commands/local/lib/local_lambda.py#L104

sanathkr:
Hey, thanks for starting this. Definitely a good feature to have. I have commented on the design PR

mihaa1:
Was there any progress with this?

rainabba:
Wow. Could we PLEASE at least get a useful error message for this then? I've wasted most of my day only to come to find the most obvious of all "Hello World" scenarios isn't supported. I just needed an /alive url so I could detect when `sam local start` is responsive. Is there a more appropriate way to do this? Below is my solution for anyone else on a similar journey.

```bash
#!/bin/sh
set -aeux

wait-for-url() {
    echo ":: Testing $1"
    timeout -s TERM 45 bash -c \
    'while [[ "$(curl -s -o /dev/null -L -w ''%{http_code}'' ${0})" != "200" ]];\
    do echo "Waiting for ${0}" && sleep 2;\
    done' ${1}
    echo "OK!"
    curl -I $1
}

trap "exit" INT TERM ERR
trap 'kill $(jobs -p)' EXIT

npm run startApiGateway &

printf "\n:: Waiting for local API Gateway to come up. Initial build/startup takes at least 10 seconds.\n"
sleep 10s
wait-for-url http://127.0.0.1:3000/alive

# Environment ready. Run tests
test/test_local.sh &

wait
```

rainabba:
[...]

PR #792, opened 2018-11-22
[WIP] sam local support for InlineCode key

j0lly:
*Issue #, if available:*

#790 

*Description of changes:*

Possibility to run `sam local invoke` for Functions authored via `InlineCode` or `Code` key

*Checklist:*

- [x] Write actual code
- [x] Write unit tests
- [x] `make pr` passes

### DIFF
--- a/designs/sam_local_inlinecode.rst
+++ b/designs/sam_local_inlinecode.rst
@@ -1,0 +1,82 @@
+.. contents:: **Table of Contents**
+   :depth: 2
+   :local:
+
+
+
+``InlineCode`` debug
+=====================
+This is the design for a feature to provide local testing capabilities to functions defined via **InlineCode** key schema.
+
+What is the problem?
+--------------------
+When new functions are started an absolute path of the folder/file containing AWS Lambda code package need to be passed
+to docker deamon in order to bind-mount the path inside the container running the AWs Lambda runtime.
+
+As of today you can only test locally your AWS Lambda Functions if defined via ``CodeUri`` key and not ``InlineCode``.
+From now on we will use ``InlineCode`` to refer to both ``Serverless::Funnction`` and ``Lambda::Function`` inline code implementation (Where ``Code`` is used as the property).
+
+What will be changed?
+---------------------
+In this proposal, we will change the code in a way that, if the tested AWS Lambda function is codified with
+``InlineCode`` key, we will create a temporary local file based on the ``InlineCode`` value, passing the temp
+file to bind-mount inside the container and make it possible to locally test ``InlineCode`` authored Functions.
+
+Success criteria for the change
+-------------------------------
+#. Support all programming languages supported by ``InlineCode``
+
+#. Local testing should just work for Functions authored with ``InlineCode``
+
+Out-of-Scope
+------------
+
+#. Anything not specifically related to ``InlineCode`` and sam local commands
+#. Provide support for Functions authored with ``InlineCode`` and leveraging ``!Sub`` pseudo function; in this case the local testing will fail
+#. Provide explicit support for ``build`` action; that means you cannot build with sam when an ``InlineCode`` function is present in the template
+
+User Experience Walkthrough
+---------------------------
+
+
+Implementation
+==============
+
+Design
+------
+*Explain how this feature will be implemented. Highlight the components of your implementation, relationships*
+*between components, constraints, etc.*
+
+Design should be pretty simple. Write a tempfile in ``./.aws-sam/build/<logical_id>`` and passing it to the Docker daemon as a bind-mount.
+File is handled inside a context-manager and cleaned out after invocation.
+
+
+Security
+--------
+
+*Tip: How does this change impact security? Answer the following questions to help answer this question better:*
+
+**What new dependencies (libraries/cli) does this change require?**
+None: InlineCode is handled transparently; no additional libraries are required beside python standard libraries
+
+**Are you reading/writing to a temporary folder? If so, what is this used for and when do you clean up?**
+Yes, I'm writing the ``InlineCode`` function string to a temporary file (inside the SAM project folder)
+If cleanup need to be performed, a context manager solution should be adopted
+
+Documentation Changes
+---------------------
+TBD
+
+Open Issues
+-----------
+
+Task Breakdown
+--------------
+- [x] Send a Pull Request with this design document
+- [ ] Build the required functions
+- [ ] Hook functions in actual code
+- [ ] Unit tests
+- [ ] Functional Tests
+- [ ] Integration tests
+- [ ] Run all tests also on Windows
+- [ ] Update documentation (if required)
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -86,7 +86,8 @@ class LocalLambdaRunner(object):
         config = self._get_invoke_config(function)
 
         # Invoke the function
-        self.local_runtime.invoke(config, event, debug_context=self.debug_context, stdout=stdout, stderr=stderr)
+        with function.codeuri:
+            self.local_runtime.invoke(config, event, debug_context=self.debug_context, stdout=stdout, stderr=stderr)
 
     def is_debugging(self):
         """
--- a/samcli/commands/local/lib/sam_function_code_provider.py
+++ b/samcli/commands/local/lib/sam_function_code_provider.py
@@ -1,0 +1,176 @@
+"""
+Class that provides codeuri from a given Function definition
+"""
+
+import logging
+import os
+import six
+
+LOG = logging.getLogger(__name__)
+
+
+class SamFunctionCodeProvider(object):
+    """
+    Lambda Function Code provider
+
+        Parameters
+        ----------
+        function_name str
+            FunctionId
+        resource_properties dict
+            Function dictionary
+        function_type str
+            Either Serverless::Function or Lambda::Function
+    """
+
+    _SERVERLESS_FUNCTION = "AWS::Serverless::Function"
+    _LAMBDA_FUNCTION = "AWS::Lambda::Function"
+    _DEFAULT_CODEURI = "."
+
+    def __init__(self, function_name, resource_properties, function_type):
+        """ Store Function info
+        This Class is a Provider for the codeuri property hold the path to an existing file/folder
+        or a pointer to a folder used to store InlineCode/ZipFile string property (Inline Lambdas)
+        It also provide a context manager function to dump the code inside a cache path and remove it
+        after invocation
+        """
+        self.function_name = function_name
+        self.handler = resource_properties.get('Handler')
+        self.runtime = resource_properties.get('Runtime')
+        self.index = self._setup_postfix(self.runtime)
+
+        self.inlinecode = self._extract_inline_code(resource_properties, function_type)
+        if self.inlinecode:
+            self.codeuri = './.aws-sam/inline/{}/'.format(function_name)
+        else:
+            self.codeuri = self._extract_code_uri(function_name, resource_properties, function_type)
+
+    def __enter__(self):
+        """ Dump Inline Code, if present for this Code Provider """
+        if self.inlinecode:
+            self._dump_code(self.inlinecode, self.codeuri, self.index)
+        return self
+
+    def __exit__(self, *exc):
+        """ CleanUp Resources created when using Inline code """
+        if self.inlinecode:
+            self._cleanup(self.codeuri, self.index)
+
+    def __repr__(self):
+        return repr(self.codeuri)
+
+    def __fspath__(self):
+        return self.codeuri
+
+    @staticmethod
+    def _setup_postfix(runtime):
+        index = 'index'
+        if runtime and runtime.startswith('python'):
+            return index + '.py'
+        return index
+
+    @staticmethod
+    def _extract_inline_code(resource_properties, function_type):
+        """ Parse InlineCode from and return the content or None """
+        code = None
+        if function_type == SamFunctionCodeProvider._SERVERLESS_FUNCTION:
+            code = resource_properties.get('InlineCode')
+        elif function_type == SamFunctionCodeProvider._LAMBDA_FUNCTION:
+            code = resource_properties.get('Code')
+            if isinstance(code, dict):
+                code = code.get('ZipFile')
+        if code:
+            return SamFunctionCodeProvider._sanitize_inlinecode(code)
+        return code
+
+    @staticmethod
+    def _extract_code_uri(function_name, resource_properties, function_type):
+        """ Based off Function Type, normalize the structure to return a path """
+        if function_type == SamFunctionCodeProvider._SERVERLESS_FUNCTION:
+            return SamFunctionCodeProvider.extract_codeuri(function_name, resource_properties, 'CodeUri')
+        elif function_type == SamFunctionCodeProvider._LAMBDA_FUNCTION:
+            return SamFunctionCodeProvider.extract_code(resource_properties, 'Code')
+
+    @staticmethod
+    def extract_codeuri(name, resource_properties, code_property_key):
+        """
+        Extracts the SAM Function CodeUri from the Resource Properties
+
+        Parameters
+        ----------
+        name str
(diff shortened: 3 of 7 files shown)


designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Implementation
+==============
+
+Design
+------
+*Explain how this feature will be implemented. Highlight the components of your implementation, relationships*
+*between components, constraints, etc.*
+
+Design should be pretty simple. Write a tempfile and passing it to the Doker daemon as a bind-mount.
```
sanathkr: You will run into the problem of Docker file sharing. If you create a temp folder using `tempfile.mkdtemp()`, chances are this creates within a folder that isn't shared in Docker. 

You have two options:
- Write the temp file somewhere within the project directory
- Or, use `docker cp` to copy the file into the container on startup. Since this file is likely to be very small, I am not concerned about the overhead added by this copy. The code already has a `copy` command that copies from within a container to host. You can follow this code to implement the reverse of it.

What do you think?

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,80 @@
...
+What will be changed?
+---------------------
+In this proposal, we will change the code in a way that, if the tested AWS Lambda function is codified with
+``InlineCode`` key, we will create a temporary local file based on the ``InlineCode`` value, passing the temp
+file to bind-mount inside the container and make it possible to locally test ``InlineCode`` authoer Functions.
+
+Success criteria for the change
+-------------------------------
+#. Support all programming languages supported by ``InlineCode``:
+   * Nodeds
```
jfuss: NodeJs

Should/do we need to add Ruby to this list?

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,80 @@
...
+Security
+--------
+
+*Tip: How does this change impact security? Answer the following questions to help answer this question better:*
+
+**What new dependencies (libraries/cli) does this change require?**
+TBD, most likely no external dependencies
+
+**Are you reading/writing to a temporary folder? If so, what is this used for and when do you clean up?**
+Yes, I'm writing the ``InlineCode`` function string to a temporary file (system tempdir)
```
jfuss: Are we going to restrict the file permissions at all for this tmp file and if so to what?

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,80 @@
...
+Design should be pretty simple. Write a tempfile and passing it to the Doker daemon as a bind-mount.
+
+
+Security
+--------
+
+*Tip: How does this change impact security? Answer the following questions to help answer this question better:*
+
+**What new dependencies (libraries/cli) does this change require?**
+TBD, most likely no external dependencies
```
jfuss: Update this with something concrete.

designs/sam_local_inlinecode.rst
jfuss: Should we place these generated files into `./.aws-sam/build/<logical_id>` for inline code?

samcli/commands/local/lib/local_lambda.py
```diff
@@ -86,6 +88,33 @@ def is_debugging(self):
         """
         return bool(self.debug_context)
 
+    def _sanitize_inlinecode(self, inlinecode):
```
jfuss: What if we did the InlineCode sanitizing while reading the template instead? We could create a provider that is responsible for understanding InlineCode and produces the file (lambda_handler) and updated template. Then everything will magically work.

This could help in the future if we wanted to load or 'pre-build' InlineCode during the `sam local [start-lambda|start-api]` commands.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -86,6 +88,33 @@ def is_debugging(self):
...
+        if isinstance(inlinecode, dict):
+            # We expect to have only 1 keypair..
+            inlinecode = list(inlinecode.values())[0]
+        # If ${something} found inside the code, raise a NotImplemented error
+        if re.search(r'\${[^!]\S+}', inlinecode):
+            LOG.debug('Not able to parse inline code: %s', inlinecode)
+            raise NotImplementedError('Not able to parse inline code because of !Sub presence')
+        return inlinecode
+
+    # FIXME: randommize dump path and make sure not to overwrite an existing file
```
jfuss: Does it need to be randomized? Capturing why would be a good addition to the design doc above.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -86,6 +88,33 @@ def is_debugging(self):
...
+    def _dump_inlinecode(self, inlinecode, path):
+        """
+        Dumps InlineCode's Function string to a file and register a function to remove the file when processes finish
+        """
+
+        inlinecode = self._sanitize_inlinecode(inlinecode)
+        file_location = '{}/index'.format(path)
+        with open(file_location, 'w+') as f:
+            f.write(inlinecode)
+        atexit.register(lambda: os.remove(file_location))
```
jfuss: We want this to be cleaned up after the invoke not after the process exists. For `start-lambda` and `start-api`, these files will hang around for a longer period of time for no reason. We should clean them up after invoking. This can be done with a context manager instead.

samcli/commands/local/lib/local_lambda.py
j0lly: makes completely sense. I just tried to hack around a quick way to get the job done but creating an object holding the code is definitively more elegant.. let me wrap around when i find few free hours to dedicate to it

samcli/commands/local/lib/local_lambda.py
j0lly: The point of randomizing it is in case we're using the local directory to dump the file (instead tempdir) with the risk of overwriting existing files.. another issue i might see is having multiple Functions dumped for the same session (rare but i had the feeling could be a use case for that)

samcli/commands/local/lib/local_lambda.py
j0lly: got it, going to revise heavily if taking the provider path

designs/sam_local_inlinecode.rst
j0lly: I'm not really sure if makes any sense/bring any value.. I would prefer to have a suggestion on this point rather than stating anything about it

designs/sam_local_inlinecode.rst
j0lly: Yes, I'm putting the file in ./.aws-sam/inline/<logical_id>/

samcli/commands/local/lib/local_lambda.py
j0lly: I've just rewrote  the skel. Is this what you actually would like to see?
The CodeProvider is a context manager that should be used before `self.local_runtime.invoke`, in `samcli/commands/local/lib/local_lambda.py`

samcli/commands/local/lib/local_lambda.py
j0lly: If using logical ids to store inline code files and placing them in ***cache***, I think there's no need to..

samcli/commands/local/lib/sam_function_provider.py
```diff
@@ -258,12 +217,17 @@ def _parse_layer_info(list_of_layers, resources):
                 codeuri = None
 
                 if resource_type == SamFunctionProvider._LAMBDA_LAYER:
-                    codeuri = SamFunctionProvider._extract_lambda_function_code(layer_properties, "Content")
+                    codeuri = SamFunctionCodeProvider.extract_lambda_function_code(
```
jfuss: We should have more generic names. This function is centered around Layers but we are extracting function code.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -72,7 +72,8 @@ def invoke(self, function_name, event, stdout=None, stderr=None):
         config = self._get_invoke_config(function)
 
         # Invoke the function
-        self.local_runtime.invoke(config, event, debug_context=self.debug_context, stdout=stdout, stderr=stderr)
+        with function.codeuri:
```
jfuss: So the way this works is the `function.codeuri` is not set to a class which is a context manager?

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,82 @@
...
+   * Python
+   * Ruby
+
+#. Local testing should just work for Functions authored with ``InlineCode``
+
+Out-of-Scope
+------------
+
+#. Anything not specifically related to ``InlineCode`` and sam local commands
+#. Provide support for Functions authored with ``InlineCode`` and leveraging ``!Sub`` pseudo function;
```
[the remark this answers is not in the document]
j0lly: For now I think should trow a Non Implemented (today is not handled at all alltogether)

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,82 @@
...
+
+What will be changed?
+---------------------
+In this proposal, we will change the code in a way that, if the tested AWS Lambda function is codified with
+``InlineCode`` key, we will create a temporary local file based on the ``InlineCode`` value, passing the temp
+file to bind-mount inside the container and make it possible to locally test ``InlineCode`` authoer Functions.
+
+Success criteria for the change
+-------------------------------
+#. Support all programming languages supported by ``InlineCode``:
```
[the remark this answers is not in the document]
j0lly: well, I'd say with the advent of custom runtimes this might open the door to other interpreted languages but, Java and friends should not be supported.. unless i missed something?

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html#cfn-lambda-function-code-zipfile here only python and node are listed..

samcli/commands/local/lib/local_lambda.py
j0lly: yes, it is like that. I did this to preserve most of the existing code

samcli/commands/local/lib/sam_function_code_provider.py
```diff
@@ -0,0 +1,164 @@
...
+        """ Delete InlineCode file """
+        os.remove(path + 'index')
+
+    @staticmethod
+    def _sanitize_inlinecode(inlinecode):
+        """
+        Verify if InlineCode has runtime substitution
+        """
+        # InlineCode isn't type safe
+        if isinstance(inlinecode, dict):
```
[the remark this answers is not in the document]
j0lly: When you use !Sub before the string, the code is passed as a dict :)

samcli/commands/local/lib/sam_function_code_provider.py
```diff
@@ -0,0 +1,164 @@
...
+            # We expect to have only 1 keypair..
+            inlinecode = list(inlinecode.values())[0]
+        # If ${something} found inside the code, raise a NotImplemented error
```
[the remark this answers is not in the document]
j0lly: Argh, I tried to provide a very basic check in order to make it usable for situations where !sub is provided but no substitution were used (I'd say improbable) so I'm ditching this part of the code as for a 1st iteration I'd not support !Sub altogether

samcli/commands/local/lib/sam_function_code_provider.py
```diff
@@ -0,0 +1,164 @@
...
+        """ Dump InlineCode to specified cache file """
+        if not os.path.exists(path):
+            os.makedirs(path)
+        with open(path + 'index', 'w') as index:
+            index.write(inlinecode)
+
+    @staticmethod
+    def _cleanup(path):
+        """ Delete InlineCode file """
+        os.remove(path + 'index')
```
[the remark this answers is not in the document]
j0lly: sure, too minimal :)
I'd just pass over here as _cleanup is used only during context manager cleanup thus, if file has been removed for whatever reason, I don't really care (imho)

designs/sam_local_inlinecode.rst
```diff
@@ -0,0 +1,82 @@
...
+
+As of today you can only test locally your AWS Lambda Functions if defined via ``CodeUri`` key and not ``InlineCode``.
+
+What will be changed?
+---------------------
+In this proposal, we will change the code in a way that, if the tested AWS Lambda function is codified with
+``InlineCode`` key, we will create a temporary local file based on the ``InlineCode`` value, passing the temp
+file to bind-mount inside the container and make it possible to locally test ``InlineCode`` authoer Functions.
+
+Success criteria for the change
```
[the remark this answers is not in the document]
j0lly: Yes I've been lazy.. Added also mentions to Code.ZipFile

TheSriram:
@j0lly Could you handle the rebase back onto develop and we can take a look at this again

j0lly:
> @j0lly Could you handle the rebase back onto develop and we can take a look at this again

I've just rebased.. But looking at integration tests, I'm pretty sure will be a mess to work out a decent solution for python 2.7. I'm looking for a different approach that might have to change a bit more the interfaces (codeUri is used all around) but will not fiddle with trying to make my class look like a string AND a file object :(

samcli/commands/local/lib/sam_function_code_provider.py
```diff
@@ -0,0 +1,176 @@
...
+            if isinstance(code, dict):
+                code = code.get('ZipFile')
+        if code:
+            return SamFunctionCodeProvider._sanitize_inlinecode(code)
+        return code
+
+    @staticmethod
+    def _extract_code_uri(function_name, resource_properties, function_type):
+        """ Based off Function Type, normalize the structure to return a path """
+        if function_type == SamFunctionCodeProvider._SERVERLESS_FUNCTION:
```
TheSriram: can we encapsulate this in a dictionary rather than multiple if else statements

samcli/commands/local/lib/sam_function_code_provider.py
```diff
@@ -0,0 +1,176 @@
...
+        index = 'index'
+        if runtime and runtime.startswith('python'):
+            return index + '.py'
+        return index
+
+    @staticmethod
+    def _extract_inline_code(resource_properties, function_type):
+        """ Parse InlineCode from and return the content or None """
+        code = None
+        if function_type == SamFunctionCodeProvider._SERVERLESS_FUNCTION:
```
TheSriram: can we encapsulate this in a dictionary rather than multiple if else statements

j0lly: @TheSriram  mind you elaborate? I've basically kept consistent behavior based on `_extract_functions` function in samcli/commands/local/lib/sam_function_provider.py
```suggestion
            def _extract_functions(resources):
        """
        Extracts and returns function information from the given dictionary of SAM/CloudFormation resources. This
        method supports functions defined with AWS::Serverless::Function and AWS::Lambda::Function

        :param dict resources: Dictionary of SAM/CloudFormation resources
        :return dict(string : samcli.commands.local.lib.provider.Function): Dictionary of function LogicalId to the
            Function configuration object
        """

        result = {}

        for name, resource in resources.items():

            resource_type = resource.get("Type")
            resource_properties = resource.get("Properties", {})

            if resource_type == SamFunctionProvider._SERVERLESS_FUNCTION:
                layers = SamFunctionProvider._parse_layer_info(resource_properties.get("Layers", []), resources)
                codeuri = SamFunctionProvider._parse_function_code(name, resource_properties, resource_type)
                result[name] = SamFunctionProvider._convert_sam_function_resource(
                    name, resource_properties, layers, codeuri
                )

            elif resource_type == SamFunctionProvider._LAMBDA_FUNCTION:
                layers = SamFunctionProvider._parse_layer_info(resource_properties.get("Layers", []), resources)
                codeuri = SamFunctionProvider._parse_function_code(name, resource_properties, resource_type)
                result[name] = SamFunctionProvider._convert_lambda_function_resource(
                    name, resource_properties, layers, codeuri
                )

            # We don't care about other resource types. Just ignore them

        return result
```

j0lly:
So, the implementation is working but I should look into how to make it work with python 2.7 implementation. There's anything else you see as a blocker to bring this PR forward?

ispyinternet:
Is this still happening?

jfuss:
Closing this PR.

If this is still something valuable we should add, please 👍 #790

PR #820, opened 2018-11-30
added support for powershell lambda functions

BrianFarnhill:
*Description of changes:*

Added support for PowerShell lambda functions. This is my first PR here so let me know if I missed anything!

*Checklist:*

- [X] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/Pipfile
```diff
@@ -0,0 +1,13 @@
+[[source]]
```
heitorlessa: You can delete this file as we will not run integ tests for these templates as of now :)

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/.gitignore
```diff
@@ -0,0 +1,168 @@
+
```
heitorlessa: Could you please update this .gitignore file to include this additional lines related to powershell/.NET?

```
# Exclude packaged modules
*.zip

# Exclude .NET assemblies from source
*.dll
```

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/README.md
```diff
@@ -0,0 +1,40 @@
+# Cookiecutter SAM for PowerShell based Lambda functions
```
heitorlessa: In parallel, we're currently simplifying the root README for each template to we don't confuse customers when they go to contribute and end up reading the wrong file (cookiecutter-aws-sam-hello-powershell/README vs cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README).

Could you update this file based on this reference for .NET?

https://github.com/awslabs/aws-sam-cli/blob/0622b9bb9b601a6c8c1e4054c2a0c93c2e70c96b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/README.md

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/tests/test_cookiecutter.py
```diff
@@ -0,0 +1,40 @@
+"""
```
heitorlessa: Similar to the other comment, you can remove this file as we're not running integ tests and they're now being cleaned up in another PR :)

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/.gitignore
```diff
@@ -0,0 +1,405 @@
+
```
heitorlessa: Could you please update this .gitignore to use .dll?

Btw, these were created via gitignore.io website :)

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,149 @@
+# {{ cookiecutter.project_name }}
```
heitorlessa: Could you please update this doc based on the new README we're working on?

https://github.com/awslabs/aws-sam-cli/blob/0622b9bb9b601a6c8c1e4054c2a0c93c2e70c96b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/%7B%7Bcookiecutter.project_name%7D%7D/README.md

It basically adds a new Appendix, SAM Logs section, brings an `event.json` file _and_ subsequently shares how to execute the function without API Gateway too, etc..

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,149 @@
...
+```bash
+dotnet test test/HelloWorld.Test
+```
+
+Alternatively, you can use Cake. It discovers and executes all the tests.
+
+### Linux & macOS
+
+```bash
+sh build.sh --target=Test
```
heitorlessa: I haven't tried running Powershell or this example yet, but is there any special configuration to be made other than selecting dotnetcore2.1 as a runtime and running Cake for restoring deps?

I guess I'm trying to think if there are other areas we need to cover if those customers are running powershell in Lambda for the first time

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -0,0 +1,33 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
heitorlessa: Could you update this template.yaml following this new one for dotnet?

https://github.com/awslabs/aws-sam-cli/blob/0622b9bb9b601a6c8c1e4054c2a0c93c2e70c96b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/%7B%7Bcookiecutter.project_name%7D%7D/template.yaml

Essentially, it removes Env Variables (not being used) and adds additional comments in the Outputs

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+
+Resources:
+
+    PowerShellFunction:
+        Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+        Properties:
+            CodeUri: ./artifacts/Function.zip
+            Handler: Function::Function.Bootstrap::ExecuteFunction
+            Runtime: dotnetcore2.1
```
heitorlessa: Is powershell only supported in dotnetcore2.1? Do we want to hardcode that or have some logic like:

```jinja
            {%- if cookiecutter.runtime == 'dotnetcore2.0' %}
            Runtime: {{ cookiecutter.runtime }}
            {%- elif cookiecutter.runtime == 'dotnetcore2.1' or cookiecutter.runtime == 'dotnet' %}
            Runtime: dotnetcore2.1
            {%- endif %}
```

That will be parsed at runtime when the customer pass in the flag ``--runtime <value>``

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+            MemorySize: 512
+            Environment: # More info about Env Vars: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#environment-object
+                Variables:
+                    PARAM1: VALUE
+
+Outputs:
```
heitorlessa: Is PowerShell not support API Gateway Proxy Integration mode? 

If it does, would you be able to update the template.yaml to so ``sam local start-api` would work? It's missing the Events and the Outputs section.

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/src/Function.ps1
```diff
@@ -0,0 +1,17 @@
...
+#
+# The last item in the PowerShell pipeline will be returned as the result of the Lambda function.
+#
+# To include PowerShell modules with your Lambda function, like the AWSPowerShell.NetCore module, add a "#Requires" statement 
+# indicating the module and version.
+
+#Requires -Modules @{ModuleName='AWSPowerShell.NetCore';ModuleVersion='3.3.221.0'}
+
+# This line publishes the input to CloudWatch logs so you know what triggered the function
+Write-Host (ConvertTo-Json -InputObject $LambdaInput -Compress -Depth 5)
```
heitorlessa: Could you change this to return a `hello world` or something like Python/Node/Ruby does: 

```json
{"message": "hello world"}
```

;)

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/test/Function.tests.ps1
```diff
@@ -0,0 +1,14 @@
+Import-Module -Name Pester
```
heitorlessa: Once the function is updated to return `hello world` then simply update the tests to reflect that so it can capture the body _and_ the statusCode ;)

Here's an example from DotNet core Tests: https://github.com/awslabs/aws-sam-cli/blob/0622b9bb9b601a6c8c1e4054c2a0c93c2e70c96b/samcli/local/init/templates/cookiecutter-aws-sam-hello-dotnet/%7B%7Bcookiecutter.project_name%7D%7D/test/HelloWorld.Test/FunctionTest.cs

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
BrianFarnhill: Looking at the new file, I don't think I need to change this as I don't include the API in this example (the 50 second warm up time based on module loading makes that prohibitive), let me know if you tihnk anything specific is missing from my file though and ill change it.

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README.md
BrianFarnhill: I've update the readme in the example which now spells out how to run it - have a look over that and see if that helps answer your concerns

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
BrianFarnhill: It only works on 2.1 so I hard coded that in, it won't work on 2.0

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/template.yaml
BrianFarnhill: It does - but as mentioned above the warm up time on the functions means it usually wouldn't be a use case that most people adopt, so I took it out. Happy to put it in if needed.

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,146 @@
...
+* [PowerShell Core installed](https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-powershell?view=powershell-6)
+* [AWSLambdaPSCore PowerShell module](https://www.powershellgallery.com/packages/AWSLambdaPSCore/1.1.0.0)
+* [Pester PowerShell module](https://github.com/pester/Pester)
+
+## Setup process
+
+### PowerShell (all Operating Systems)
+
+```powershell
+build.ps1
```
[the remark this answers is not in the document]
TheSriram: There is a current .NET builder in the works at aws-lambda-builders, do take a look, if this can take advantage of that. 

https://github.com/awslabs/aws-lambda-builders/pull/80

samcli/local/init/templates/cookiecutter-aws-sam-hello-powershell/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Requirements
+
+* AWS CLI already configured with Administrator permission
+* [Docker installed](https://www.docker.com/community-edition)
+* [SAM CLI installed](https://github.com/awslabs/aws-sam-cli)
+* [.NET Core installed](https://www.microsoft.com/net/download)
+* [PowerShell Core installed](https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-powershell?view=powershell-6)
+* [AWSLambdaPSCore PowerShell module](https://www.powershellgallery.com/packages/AWSLambdaPSCore/1.1.0.0)
+* [Pester PowerShell module](https://github.com/pester/Pester)
```
TheSriram: Is Pester the most used test framework? We need to start customers off on the right path.

jfuss:
We still see value in adding powershell in `sam init`. Instead of adding them directly in here, we should add them into https://github.com/awslabs/aws-sam-cli-app-templates/ instead. This will allow us to define a powershell example within the .NetCore runtime. 

Closing this PR but we would love a PR for this to our app templates repo that SAM CLI uses for init.

PR #822, opened 2018-12-03
list ruby as supported runtime

timoschilling:


jfuss:
@timoschilling While you are here, can you also add `Python3.7 and Provided` runtimes to the list. There is no init for `Provided`, though.

TheSriram:
This is a great catch! thanks for adding this, could you add the other supported runtimes such as the ones @jfuss mentioned.

timoschilling:
@jfuss applied

jfuss:
@timoschilling Thank you! Just waiting for the build. Will merge shortly after that finishes.

Issue #816, opened 2018-11-30
Local invoke fails when trying to use Layers locally


heitorlessa:
**Description:**

When trying to use the new Layers feature locally it fails due to not finding a Dockerfile which seems to be created at runtime -- Haven't dig too much into it yet as I just bumped into that.

**Steps to reproduce the issue:**
1. **Initialize a boilerplate app**: `sam init -r python; cd sam-app`
2. **Install dependencies into a lib folder**: `pip install -r hello_world/requirements.txt -t lib/`
3. **Create a Layer** in `template.yaml`:

```yaml
    RequestsLayer:
        Type: 'AWS::Serverless::LayerVersion'
        Properties:
            ContentUri: lib/
```

4. **Reference that Layer** in the `HelloWorldFunction` resource:

```yaml
    HelloWorldFunction:
        Type: AWS::Serverless::Function
        Properties:
            CodeUri: hello_world/
            Handler: app.lambda_handler
            Runtime: python3.6 # also tried python3.7 (default)
            Layers:
                - !Ref RequestsLayer
```

5. Run function locally: ``sam local invoke HelloWorldFunction --no-event``

**Observed result:**

Local invocation fails quickly and returning the trace data at the bottom.

**Expected result:**

Build Layer locally and local invocation to succeed.

**Additional environment details (Ex: Windows, Mac, Amazon Linux etc)** OSX Sierra 10.12.6 (16G1618)

**Output of `sam --version`**: SAM CLI, version 0.8.0

**Optional Debug logs**:

`Add --debug flag to command you are running`

```python
sam local invoke HelloWorldFunction --no-event
2018-11-30 08:42:46 Found credentials in shared credentials file: ~/.aws/credentials
2018-11-30 08:42:46 Invoking app.lambda_handler (python3.6)
2018-11-30 08:42:46 RequestsLayer is a local Layer in the template
2018-11-30 08:42:46 Image was not found.
2018-11-30 08:42:46 Building image...
Traceback (most recent call last):
  File "/Users/lessa/.pyenv/versions/work/bin/sam", line 11, in <module>
    sys.exit(cli())
  File "/Users/lessa/.pyenv/versions/3.6.2/envs/work/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/lessa/.pyenv/versions/3.6.2/envs/work/lib/python3.6/site-packages/click/core.py", line 697, in main
[...]

PR #823, opened 2018-12-03
fix: Create Layer Cache during building image

jfuss:
*Issue #, if available:*
#816

*Description of changes:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/layers/layer_downloader.py
+++ b/samcli/local/layers/layer_downloader.py
@@ -34,9 +34,22 @@ class LayerDownloader(object):
         lambda_client boto3.client('lambda')
             Boto3 Client for AWS Lambda
         """
-        self.layer_cache = layer_cache
+        self._layer_cache = layer_cache
         self.cwd = cwd
         self.lambda_client = lambda_client or boto3.client('lambda')
+
+    @property
+    def layer_cache(self):
+        """
+        Layer Cache property. This will always return a cache that exists on the system.
+
+        Returns
+        -------
+        str
+            Path to the Layer Cache
+        """
+        self._create_cache(self._layer_cache)
+        return self._layer_cache
 
     def download_all(self, layers, force=False):
         """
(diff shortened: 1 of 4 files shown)


tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -376,8 +375,23 @@ def test_caching_two_layers(self):
...
+    def setUp(self):
+        self.layer_cache = Path().home().joinpath("integ_layer_cache")
+
+    def tearDown(self):
+        docker_client = docker.from_env()
+        samcli_images = docker_client.images.list(name='samcli/lambda')
+        for image in samcli_images:
+            docker_client.images.remove(image.id)
+
+    def test_layer_does_not_exist(self):
```
TheSriram: So this test, ensures a cache is created for a non existent layer? Its not very explicit what it does.

samcli/local/layers/layer_downloader.py
```diff
@@ -34,10 +34,23 @@ def __init__(self, layer_cache, cwd, lambda_client=None):
...
+    def layer_cache(self):
+        """
+        Layer Cache property. This will always return a cache that exists on the system.
+
+        Returns
+        -------
+        str
+            Path to the Layer Cache
+        """
+        self._create_cache(self._layer_cache)
```
TheSriram: 👍

tests/integration/local/invoke/test_integrations_cli.py
jfuss: No. This is just moving existing tests that do not create the cache directory. These tests do not download or build an image and therefore the `layer_cache_dir` is not created. The class these were previously under, removed the layer_cache after every tests. Instead of making the tests ignore the delete errors (OSError when the cache does not exist), I moved this to a new class. That way we ensure we always have a clean cache and reduce the risk of a failure to one test not impacting another.

TheSriram: Ok got it, thanks for the explanation. Just trying to capture this information somewhere.

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -12,6 +13,10 @@
 from tests.integration.local.invoke.layer_utils import LayerUtils
 from .invoke_integ_base import InvokeIntegBase
 
+# Layers tests require credentials and Travis will only add credentials to the env if the PR is from the same repo.
+# This is to restrict layers tests to run outside of Travis and when the branch is not master.
+SKIP_LAYERS_TESTS = os.environ.get("TRAVIS", False) and os.environ.get("TRAVIS_BRANCH", "master") != "master"
```
TheSriram: Perfect.

PR #828, opened 2018-12-04
Move checking Docker connectivity to ContainerManager

ndobryanskyy:
*Description of changes:*

Eliminated `_check_docker_connectivity()` method and introduced new property in `ContainerManager` `is_docker_reachable`. Also now `ContainerManager` is created on `__enter__` by InvokeContext.

Updated tests and added new as required. This is the first part of porting enhancing changes from #774 .

I think we could close #774 now, right?

*Checklist:*

- [x] Implement required changes
- [x] Write unit tests
- [x] `make pr` passes


ndobryanskyy:
@jfuss more to come 😉

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -137,7 +135,11 @@ def __enter__(self):
                                                       self._debug_args,
                                                       self._debugger_path)
 
-        self._check_docker_connectivity()
+        self._container_manager = self._get_container_manager(self._docker_network,
+                                                              self._skip_pull_image)
+
+        if not self._container_manager.is_docker_reachable:
+            raise InvokeContextException("Running AWS SAM projects locally requires Docker. Have you got it installed?")
```
TheSriram: I wonder if we can say Docker > x version, I think that requires some investigation though.

ndobryanskyy: @TheSriram  SAM CLI is constrained to some Docker version?

TheSriram: Not really, just a way to ask people to upgrade to the latest. There have been some instances where users were using a really old version of docker and had to upgrade. example: to get mounting to work as expected.

PR #832, opened 2018-12-05
Init: Add Extra Context and name can now be used with location

heitorlessa:
*Issue #, if available:* https://github.com/awslabs/aws-sam-cli/issues/757

*Description of changes:*

* ``--name`` is now respected when using with ``--location``
    - We basically inject ``--name`` value as ``project_name`` extra context, so any custom template that may utilize this will work off-the-shelf too
* Introduces ``--extra-context`` so one can pass additional information to custom templates in combination with ``--no-input`` that is useful for IDEs and other integrations 

*Checklist:*

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation
- [X] Implement ``--extra-context` option
- [X] Write new tests to cover ``--extra-context``


heitorlessa:
Done - Pending CR

buggy:
I posted a comment on the original issue. Using `--runtime RUNTIME` with built-in templates but `-e runtime=RUNTIME` for custom templates is going to confuse people. Can we pass the `runtime` too?

heitorlessa:
Pushed last change before review: ``--runtime`` gets passed as extra context

In addition to tests, created a new boilerplate app using the following custom template:

```bash
samdev init --location https://github.com/aws-samples/cookiecutter-aws-sam-dynamodb-python --runtime python2.7 --no-input
```

samcli/commands/init/__init__.py
```diff
@@ -64,21 +65,31 @@ def cli(ctx, location, runtime, output_dir, name, no_input):
         \b
         $ sam init --location /path/to/template/folder
 
+        \b
+        Initializes a new SAM project using custom template with additional known context
+        \b
+        # Add multiple -e for multiple context to be added (-e some=Value -e other=Value)
+        $ sam init --location gh:aws-samples/cookiecutter-aws-sam-python -e project_short_description=Oi
```
TheSriram: This makes a lot of sense, for new cookiecutter samples, that have other content can be subbed.

samcli/commands/init/__init__.py
```diff
@@ -64,21 +65,31 @@ def cli(ctx, location, runtime, output_dir, name, no_input):
...
-def do_cli(ctx, location, runtime, output_dir, name, no_input):
+def do_cli(ctx, location, runtime, output_dir, name, no_input, extra_context):
     """
     Implementation of the ``cli`` method, just separated out for unit testing purposes
     """
     LOG.debug("Init command")
     click.secho("[+] Initializing project structure...", fg="green")
 
+    # Break up given extra context into dict(key:value); else use a sane default empty dict
+    options = (opts.split('=') for opts in extra_context)
```
TheSriram: question: will extra_context always be of form a=b and thats it?

will we ever do a=b,c=d. I think we need something similair to that in !Ref or env var subsitution. I fail to recall which.

heitorlessa: Great point. I took inspiration from other CLIs like docker-compose as that's easily readable when you have multiple params. 

I looked into the [SAM CLI code and the closest we do is for env-vars option](https://github.com/awslabs/aws-sam-cli/blob/f88fb592e072314a4902e1f00489329a58c79519/samcli/commands/local/lib/local_lambda.py#L119) and that expects a file in JSON format. ``sam deploy`` shells out to AWS CLI which in turn follows `a=b,c=d`. 

I'm happy either way:

A) Implement `a=b,c=d` format like AWS CLI does
B) Reuse what ``--env-vars`` does for JSON file like ``--env-vars`` and disallow input via arguments

Let me know your thoughts _and_ thanks for taking the time to review this!

jfuss:
@heitorlessa Can you rebase this and handle any conflicts that arise? Would love to get this over the finish line.

TheSriram:
@heitorlessa could you look at rebasing this, so that it can be reviewed! Thanks :)

heitorlessa:
Apologies for the delay - I'll go through this week and post updates

heitorlessa:
@TheSriram rebased in place, ran `make pr` and simply waiting for Travis at this stage. 

Not sure if we settled the comment on `-e context=value -e context2=value2` vs `-e context=value,context2=value`, but will leave at your discretion

samdengler:
Would it be possible to merge this PR?  I ran into the ignored --name parameter when using --location as well.

awood45:
This PR seems to be broken from age, though #1500 provides a lot of the intended functionality (it doesn't plumb `--name` directly, though you can pass it through as extra context). If we want to revive the `--name` pass-through for `--location` it should be easier post-1500.

Closing since #1500 will render this hard to merge, feel free to rework for `--name` if desired.

PR #840, opened 2018-12-07
[WIP] Request for Comments: Building Layer Resources

jfuss:
💻 [**Rendered Design Doc is here**](https://github.com/jfuss/aws-sam-cli/blob/layer-build-design/designs/build_layer_support.rst)
🗓 **RFC Open for feedback until**: TBD

*Description of changes:*

Capturing initially thoughts on how `sam build` can be expanded to Layers. Needs a more detailed design on the changes but a high level (what is in and out of scope) is captured.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/designs/build_layer_support.rst
+++ b/designs/build_layer_support.rst
@@ -1,0 +1,223 @@
+.. contents:: **Table of Contents**
+   :depth: 2
+   :local:
+
+``sam build`` support for Layers
+================================
+This is the design for add the capability to **build** a Lambda Layer. **Build** is the operation of converting the function's
+source code to an artifact that can be executed on AWS Lambda.
+
+
+What is the problem?
+--------------------
+``sam build`` was implemented with only the ability to build Lambda Functions. At re:Invent 2018, Lambda launched a new
+resource called Layer. This allows customers make code/files available during the invoke of a function without
+those files/code being packaged up directly with the function code itself. Just like with Functions, customers
+will need to build these layers for the Lambda Environment.
+
+
+What will be changed?
+---------------------
+``sam build`` will need to be able to parse and understand the Layer Resource (both SAM and vanilla CloudFormation) and
+then produce an artifact in the ``.aws-sam/build`` folder that can be packaged to allow execution on AWS Lambda.
+
+
+Success criteria for the change
+-------------------------------
+* Layers that define one runtime, will produce an artifact in ``.aws-sam/build``
+* The built template will point to the built artifact, if that resource was built
+* As more runtimes get added to ``sam build``, there will be no action to support layers for those runtimes
+
+
+Out-of-Scope
+------------
+* Building for multiple runtimes
+    * Building for many runtimes is a complex operation. You can't build for the highest version and guarantee it will
+      work for all runtimes specified. For example, in Python you can define different requirements to be used based
+      on the Python version being used. This means the artifacts generated from using Python 3.6 and Python 2.7 may not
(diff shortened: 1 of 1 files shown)


designs/build_layer_support.rst
```diff
@@ -0,0 +1,222 @@
...
+
+            Returns
+            -------
+            dict
+                Returns the path to where each resource was built as a map of resource's LogicalId to the path string
+            """
+            pass
+
+
+The ApplicationBuilder no longer understands a single provider. This requires the building logic to live at each
```
TheSriram: Will this result in any partial builds happening? as in we have 4 layers, if they are built independently and fail on the 4th, do we still have the other artifacts built?

designs/build_layer_support.rst
```diff
@@ -0,0 +1,222 @@
...
+resource.
+
+For each data model that represents a resource will implement a build method. This will allow each resource to define
```
TheSriram: can you expand a bit on this? Where will such a definition live?

jfuss: Sure. The definitions of a resources are defined [here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/local/lib/provider.py) (currently). This would add a build method in each one (we will have to expand the namedtuples slightly for this but the idea is the same).

Will update this to make it clearer and link to the current resource models.

designs/build_layer_support.rst
jfuss: I didn't think of this case. Great call out. 

My initial thought is: This should be no different than functions. If you have 4 functions and one fails what happens (I don't know the answer to this yet but will test it out). On the other hand, if you are embedding Layers in with your template/functions (Ref'ing a Layer resource in the Function definition). You will want your function invokable only if all the layers you are referencing have also built. 

I am going to ponder this a little more to see which one is better or maybe how we can support both.

patstrom:
Hello! What is the status on this? This would super useful.

adilosa:
Agreed. @patstrom, as a workaround, I found this: https://bryson3gps.wordpress.com/2018/12/06/trick-sam-into-building-your-lambda-layers/

rhbecker:
Would someone be willing to explain the use case for building layers? My probably incomplete understanding is that you'd either use layers or `sam build`.

I wrote up [this SO answer](https://stackoverflow.com/a/55109779/2755803) to describe how I've been using a layer to point at my locally installed dependencies. I'd love to hear more about how other people use these features.

patstrom:
Well, the layers are as much part of the application as the functions themselves, so it makes sense that they are built alongside everything else.

Right now I place shared python modules in layers so that all functions can reference them and if these layers have dependencies I have to `pip install -r requirements.txt -t .` (as well as placing everything in a "python/" subdirectory) manually before I can use `sam local`, which is perhaps not a huge deal but it does feel very lacking.

jfuss:
I am going to close this PR for now. We aren't actively working on this and cleaning up stale PRs. We can re-open and use this as a base for when we start the work on supporting layers.

If any community members are looking for something to contribute, this would be an awesome one to start. The design needs a little more fleshing out but I don't think its far off. Happy to help/be a point of contact for whoever takes it up, as well.

ijemmy:
Just in case any one run into this issue, it is now supported. You need to add a metadata into the layer and SAM will build it.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-layers.html

Issue #830, opened 2018-12-05
Support for multiValueHeaders locally


Sarke:
**Description:**
`sam local start-api` doesn't detect or pass along any of the `multiValueHeaders` from the lambda response.

This is most notable (for me) because it defaults to `application/json` content type even though another header is sent, so for example HTML will be displayed as plain text in the browser.

This is because it only checks the `headers` key in the lamba response, and not the `multiValueHeaders` key.

The code related to this issue is the `_parse_lambda_output` method, found here:
https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/apigw/local_apigw_service.py#L182-L214

Here's an example lambda response (using https://github.com/stackery/php-lambda-layer).

```json
{
  "body": "<html><body>test</body></html>",
  "multiValueHeaders": {
    "Connection": ["close"],
    "Content-type": ["text/html;charset=UTF-8"],
    "Date": ["Wed, 05 Dec 2018 00:51:05 +0000"],
    "Host": ["localhost:8000"],
    "X-Powered-By": ["PHP/7.1.7"]
    },
  "statusCode": 200
}
```

This works as expected once deployed, just not locally with `sam`.

See also:

https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#apigateway-multivalue-headers-and-parameters

**Steps to reproduce the issue:**
1. Create `template.yaml` like this:
```yaml
AWSTemplateFormatVersion: 2010-09-09
Description: PHP Lambda test
Transform: AWS::Serverless-2016-10-31
Resources:
  phpserver:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: phpserver
      Description: PHP Webserver
      CodeUri: src/php
      Runtime: provided
      Handler: index.php
      MemorySize: 256
      Timeout: 15
      Tracing: Active
      Layers:
        - 'arn:aws:lambda:us-west-2:887080169480:layer:php71:4'
      Events:
        api:
          Type: Api
          Properties:
            Path: /{proxy+}
            Method: ANY
```

2. Create `src/php/index.php` with any content, e.g.:
```html
<html><body>OK</body></html>
```
3. Run `sam local start-api` and go to `http://127.0.0.1:3000/test` (or whatever the address).

**Observed result:**
[...]

PR #842, opened 2018-12-07
Support for response multiValueHeaders locally

martysweet:
*Issue #, if available:*
Addresses is #830, supporting multiValueHeaders locally.

*Description of changes:*

Merges lamdba response headers and multiValueHeaders in accordance with guidance from here.
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

-  You can use the multiValueHeaders key to specify all of your extra headers, including any single-value ones.
- If you specify values for both headers and multiValueHeaders, API Gateway merges them into a single list.
- If the same key-value pair is specified in both, only the values from multiValueHeaders will appear in the merged list.

Within base64 logic, splits Content-Type by ;, to prevent issues with malformed Content types.
`content_type = lamba_response_headers['Content-Type'].split(";", 1)[0]`

- [ ] TODO: Add tests for the `_should_base64_decode_body` change - Guidance needed
 
*Checklist:*

- [x] Write unit tests (Tests for functionality of merger and update to parse_lambda tests for ensuring the component is called)
- [x] Write/update functional tests (Unsure what to add - Guidance needed)
- [x] Write/update integration tests (Unsure what to add - Guidance needed)
- [x] `make pr` passes
- [x ] Write documentation - No adjustment required (unless there is a supported features list somewhere?)

Please let me know what else is required/missing! :smile: This PR plays well with #741, which is for requests.

### DIFF
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -193,8 +193,8 @@ class LocalApigwService(BaseLocalService):
         if not isinstance(json_output, dict):
             raise TypeError("Lambda returned %{s} instead of dict", type(json_output))
 
-        headers = LocalApigwService._merge_response_headers(json_output.get("headers"),
-                                                            json_output.get("multiValueHeaders"))
+        headers = LocalApigwService._merge_response_headers(json_output.get("headers") or {},
+                                                            json_output.get("multiValueHeaders") or {})
         status_code = json_output.get("statusCode") or 200
         body = json_output.get("body") or "no data"
         is_base_64_encoded = json_output.get("isBase64Encoded") or False
@@ -265,16 +265,14 @@ class LocalApigwService(BaseLocalService):
 
         """
 
-        # Merge dictionary
-        h = (headers or {}).copy()
-        h.update(multi_headers or {})
+        processed_headers = CaseInsensitiveDict(headers)
 
-        # Now we have single headers map with Strings and Lists, convert Lists to CSV Strings for Flask
-        for _, s in enumerate(h):
-            if isinstance(h[s], list):
-                h[s] = ", ".join(h[s])
+        # Convert multi_header Lists to CSV Strings for Flask
+        # This gives multiValueHeaders precedence over the API Gateway headers property
+        for _, s in enumerate(multi_headers):
+            processed_headers[s] = ", ".join(multi_headers[s])
 
-        return CaseInsensitiveDict(h)
+        return processed_headers
 
     @staticmethod
     def _construct_event(flask_request, port, binary_types):
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -152,7 +152,7 @@ class LocalApigwService(BaseLocalService):
                                                                      route.binary_types,
                                                                      request)
         except (KeyError, TypeError, ValueError):
-            LOG.error("Function returned an invalid response (must include one of: body, headers or "
+            LOG.error("Function returned an invalid response (must include one of: body, headers, multiValueHeaders or "
                       "statusCode in the response object). Response received: %s", lambda_response)
             return ServiceErrorResponses.lambda_failure_response()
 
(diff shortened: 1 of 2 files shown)


martysweet:
Hi @jfuss, could you let me know what else is required for this? Thanks!

jfuss:
@martysweet Apologies on the slow response here. I am talking a look at this now.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -234,11 +235,47 @@ def _should_base64_decode_body(binary_types, flask_request, lamba_response_heade
         True if the body from the request should be converted to binary, otherwise false
 
         """
-        best_match_mimetype = flask_request.accept_mimetypes.best_match([lamba_response_headers["Content-Type"]])
+
+        # Get the first part of the content-type header, to allow for extras such as text/html; charset=utf-8
+        content_type = lamba_response_headers['Content-Type'].split(";", 1)[0]
+        best_match_mimetype = flask_request.accept_mimetypes.best_match([content_type])
```
jfuss: How did you validate this is the correct behavior?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -234,11 +235,47 @@ def _should_base64_decode_body(binary_types, flask_request, lamba_response_heade
...
+            multiValueHeaders map from the lambda_response_headers
+
+        Returns
+        -------
+        Merged list in accordance to the AWS documentation within a CaseInsensitiveDict
+
+        """
+
+        # Merge dictionary
+        h = (headers or {}).copy()
```
jfuss: Don't use one letter variable names. It makes reading and understanding the code more difficult.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -234,11 +235,47 @@ def _should_base64_decode_body(binary_types, flask_request, lamba_response_heade
...
+        h.update(multi_headers or {})
+
+        # Now we have single headers map with Strings and Lists, convert Lists to CSV Strings for Flask
+        for _, s in enumerate(h):
```
jfuss: You should just iterate over the `multi_headers`. The value of each header will be a list, so you won't need the `isinstance` check.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -234,11 +235,47 @@ def _should_base64_decode_body(binary_types, flask_request, lamba_response_heade
...
+            if isinstance(h[s], list):
+                h[s] = ", ".join(h[s])
+
+        return CaseInsensitiveDict(h)
```
jfuss: Why not create this from `headers`, iterate over the `multi_headers`, and insert in to the `case_insensitive_dict`.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -193,8 +193,9 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
         if not isinstance(json_output, dict):
             raise TypeError("Lambda returned %{s} instead of dict", type(json_output))
 
+        headers = LocalApigwService._merge_response_headers(json_output.get("headers"),
```
jfuss: You should set defaults when getting the keys. This will simplify the `_merge_response_headers` method.

samcli/local/apigw/local_apigw_service.py
martysweet: A MIME type should contain only `type/subtype`.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

However, the Content-Type header contains the MIME, then potentially extra charset information:
https://www.ietf.org/rfc/rfc2045.txt (page 12)
>  In the Augmented BNF notation of RFC 822, a Content-Type header field
   value is defined as follows:

     content := "Content-Type" ":" type "/" subtype
                *(";" parameter)
                ; Matching of media type and subtype
                ; is ALWAYS case-insensitive.

Passing this extra charset information into `flask_request.accept_mimetypes.best_match` results in Flask returning 500.

jfuss: @martysweet I should have been a little clearer. Does this behavior match API Gateway's? Meaning do they only look at the first Content-Type Header or do they compare with the list of all headers? Did we match documentation or what is observed from the service?

martysweet: @jfuss Ah understood. Looking into it, this might be more complex than just looking at `Content-Type`, though a bit outside of this PR.

So your question is, if multiple Content-Type Headers are returned to API Gateway, which one is used for the comparison against the binary types list?

I can't find anything in the docs, will have to run some tests.

Docs regarding `Accept` and `Content-Type` (which doesn't answer the question).
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-configure-with-console.html
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html

everyonce:
@martysweet I really would love for this to work.  when I'm testing it using both your branch here, and apigw, they do act different.  this code creates a comma-seperated list in a single header (which chrome, at least, doesnt like), and apigw creates multiple headers with the same key (in my tests I'm using set-cookie, but I'm sure it's for anything that it applies to)

dsanders11:
> and apigw creates multiple headers with the same key (in my tests I'm using set-cookie, but I'm sure it's for anything that it applies to)

@everyonce, this isn't true, actually. `Set-Cookie` is a special case, according to the HTTP spec headers aren't supposed to appear multiple times, but they allow `Set-Cookie` since it's different. API GW seems to follow this as well, only creating multiple headers for `Set-Cookie` and instead creating a list for other headers, like `X-Foo`. I've just tested it and using `X-Foo` definitely creates a comma-separated list.

martysweet:
Closing in favour of #1166

Issue #835, opened 2018-12-06
After upgrading SAM CLI - Docker StdOut appears to be buffered


singulli1:
### Description

Windows 10 64-bit
SAM CLI, version 0.7.0
Docker version 18.09.0, build 4d60db4
Runtime: node.js 6.10 

Before the upgrade, everything worked great.
After upgrade, it appears stdout from my application running in my docker image is buffering until the application completes, at which time stdout appears to be flushed and I can see it.

### Steps to reproduce

1. open windows cmd prompt.
2. navigate to node.js project directory
3. run sam local invoke cmd

example cmd:
sam local invoke  -e ./src/scheduledEvent.json  "ClaimMDIntegration" --template templateDev.yaml

In my node.js application, there are several console.log("Messages") commands.  I'd expect to see messages written to my command prompt window as they occur.

### Observed result
Instead, messages are written to the command prompt only after the program has completed execution.

### Expected result

Message should show up in command prompt as they occur, during program execution.  This was the behavior before I upgraded.

`Add --debug flag to command you are running`

C:\Users\scott\projectsSAM\psykdeskClaimsIntegration>sam local invoke --debug  -e ./src/scheduledEvent.json  "ClaimMDIntegration" --template templateDev.yaml
2018-12-06 14:06:43 Using SAM Template at C:\Users\scott\projectsSAM\psykdeskClaimsIntegration\templateDev.yaml
2018-12-06 14:06:43 local invoke command is called
2018-12-06 14:06:43 No Parameters detected in the template
2018-12-06 14:06:43 1 resources found in the template
2018-12-06 14:06:43 Found Serverless function with name='ClaimMDIntegration' and CodeUri='./build'
2018-12-06 14:06:43 Trying paths: ['C:\\Users\\scott\\.docker\\config.json', 'C:\\Users\\scott\\.dockercfg']
2018-12-06 14:06:43 Found file at path: C:\Users\scott\.docker\config.json
2018-12-06 14:06:43 Couldn't find auth-related section ; attempting to interpretas auth-only file
2018-12-06 14:06:43 Config entry for key stackOrchestrator is not auth config
2018-12-06 14:06:43 http://localhost:None "GET /v1.35/_ping HTTP/1.1" 200 2
2018-12-06 14:06:43 Trying paths: ['C:\\Users\\scott\\.docker\\config.json', 'C:\\Users\\scott\\.dockercfg']
[...]

PR #843, opened 2018-12-07
fix: Flush stdout and stderr on each write when debugging

ndobryanskyy:
*Issue #, if available:*
#835 
*Description of changes:*

Introduced `StreamWriter` which wraps `stdout` and `stderr` and is passed to the Lambda container. This class supports `auto_flushing` and could be further extended to support any flexibility in output as we want.

*Checklist:*

- [x] Write unit tests
- [x] Write/update functional tests. Do I need any?
- [x] Write/update integration tests. Do I need any?
- [x] `make pr` passes
- [x] Write documentation. Updated all existing doc strings referring to `stdout` and `stderr`


samcli/lib/utils/stream_writer.py
```diff
@@ -0,0 +1,37 @@
+"""
+This class acts like a wrapper around output streams to provide any flexibility with output we need
+"""
```
TheSriram: I'm curious to know what is the performance hit, if we just flush all the time.

gyfis:
While I don't have much context about the overall direction of the sam cli, I really like these changes and the fact that it's not just an if-clause somewhere, but a more complete approach to the problem. Thank you for taking the time to work on this :)

samcli/lib/utils/stream_writer.py
ndobryanskyy: I think it is completely negligible for the debugging, thoughts?

TheSriram: yeah totally we need it for debugging. I'm talking about non-debugging cases as well. just curious.

ndobryanskyy:
@TheSriram I wonder though why on python3.7 integration tests fail?

jfuss:
The Golang zip runtime test is being funky on 3.6 and 3.7 as of recently. I just restarted it. We aren't entirely sure why it fails sometimes at the moment.

ndobryanskyy:
@jfuss, understood, pretty strange 😅. Thanks for restarting tho, now it passed just fine :D

ndobryanskyy:
@TheSriram, thanks for the approve, the only thing that concerns me is that should we really think about "flushing all three time" as it is actually requested in #835 
Because my PR actually fixes only cases intended for debugging, thoughts?

TheSriram:
Correct, that is how I'm thinking of this as well. This PR adds mechanisms around how we flush. But flushing all the time should definitely be a separate PR, which validates performance and what are the impacts of such an operation.

ndobryanskyy:
Definitely, also I'm looking forward to investigate unbuffered streams in Python3 (in Python2 they just work), that should be a good thing (performancewise) if we want "auto flush" all the time behavior

PR #844, opened 2018-12-07
docs: Add sam publish command design doc

paoptu023:
*Description of changes:*
Design doc for adding a new command `sam publish`.

*Checklist:*

- [x] Write Design Document 
- [x] Build the command line interface
- [x] Build the underlying library
- [x] Unit tests
- [x] Integration tests
- [x] Run all tests on Windows
- [x] Update documentation


designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+the application code and SAM template to an Amazon S3 bucket, correctly set S3 bucket policy that grants the service read
+permissions for artifacts uploaded to S3, then open the AWS Serverless Application Repository console and provide information
+in a bunch of input boxes. If they use the AWS CLI, they need to pass all the information as parameters, and it's easy to make
+a mistake while typing in the command line.
+
+
+What will be changed?
+---------------------
+In this proposal, we will be providing a new command, ``sam publish app``, which takes a SAM template as input and publishes
+an application to AWS Serverless Application Repository using applicaiton metadata specified in the template. Customers just
```
jfuss: nit: remove just.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+need to provide application metadata information in the template, then ``sam package`` will handle uploading local files to S3,
+and ``sam publish app`` will create the app in Serverless Application Repository. We will also provide sharing options to set
+application permission policies. This command will greatly simplify the exsiting publishing experience.
```
jfuss: nit: remove "This command will greatly simplify the exiting publishing experience." While true, it is not needed to answer the "What will be changed?" question.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+
+Success criteria for the change
+-------------------------------
+#. Support all the following use cases:
+
+   * Create new application and its first version in SAR using ``sam publish app``
+   * Create new version of existing SAR application using ``sam publish app``
+   * Update application metadata of existing SAR application using ``sam publish app``
+   * Share the app publicly using the ``--make-public`` option
+   * Make the app private using the ``--make-private`` option
```
jfuss: Sorry if this is a silly questions: If I run `sam publish app --make-private`, is it automatically 'shared' with the account it exists in or do I need to provide that account in the `--account-ids` option?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+#. ``sam package`` command can upload local readme/license files to S3.
+
+
+Out-of-Scope
+------------
+#. Manage application permission separately without publishing/updating the app.
+
+#. Specify granular permission types as defined in `application permission`_ when sharing the application.
+
+#. Recursively publish nested apps in the template (SAR CreateApplication API doesn't support yet).
```
jfuss: How will customers do this? We should clearly document how customers can accomplish this. 

If I try to publish a nested app, will the command fail in any way?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+
+#. Run through CI/CD pipeline for the application before publishing.
+
+#. Publish to other repositories besides SAR.
+
+#. Recognize template changes and suggest version number.
+
+#. Parse metadata from template without an AWS::ServerlessRepo::Application section.
```
jfuss: I am not sure what this means exactly. Could you explain this further?

Is this saying other data in the `Metadata` section will not be parsed or touched?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+            SemanticVersion: 1.0.0
+            SourceCodeUrl: https://github.com/user1/my-app-project
+
+    Resources:
+        HelloWorldFunction:
+            Type: AWS::Lambda::Function
+            Properties:
+                CodeUri: s3://bucket/hello-world
+
+Package SAM template
```
jfuss: Let's go one set farther with including `sam build` in the workflow.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+
+
+#. ``sam package`` command can upload local readme/license files to S3.
+
+
+Out-of-Scope
+------------
+#. Manage application permission separately without publishing/updating the app.
+
+#. Specify granular permission types as defined in `application permission`_ when sharing the application.
```
jfuss: This is IAM Permissions right?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+            Properties:
+                CodeUri: s3://bucket/hello-world
+
+Package SAM template
+  Run ``sam package --template-file ./template.yaml --output-template-file packaged.yaml --s3-bucket my-bucket``
+  to upload code artifacts, readme and license files to S3 and generate the packaged template.
+
+Create new application in SAR
+  Run ``sam publish app -t ./packaged.yaml`` to publish a new application named my-app in SAR with the first version
+  created as 1.0.0. The app will be created as private by default. SAM CLI prints application created message and
```
jfuss: If `sam publish app` defaults to private, do we really need `--make-private` Option? Publish is not mutating any exiting application.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+  link to the console details page.
+
+Create new version of an existing SAR application
+  Modify the existing template, give a different SemanticVersion value, and run ``sam publish app -t ./packaged.yaml``.
+  SAM CLI prints application metadata updated message, application version created message, values of the current application
+  metadata and link to the console details page.
+
+Crete application/version and set application permission
+  Run ``sam publish app -t ./packaged.yaml --make-public`` to publish the app and share it publicly. If ``--make-private``
```
jfuss: Does this mean all permissions are allowed from  https://docs.aws.amazon.com/serverlessrepo/latest/devguide/access-control-resource-based.html#application-permissions ?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+  option is used, the app will only be visible to the owner. If ``--account-ids <account ids>`` is used, the app will be
+  shared with the provided AWS accounts.
```
jfuss: Shared how? What is the permissions granted for those accounts?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+
+Update the metadata of an exsiting application without creating new version
+  Keep SemanticVersion unchanged, then modify metadata fields like Description or ReadmeUrl, and run
+  ``sam publish app -t ./packaged.yaml``. SAM CLI prints application metadata updated message, values of the current
+  application metadata and link to the console details page.
+
+Output of the ``sam publish app`` command will be a link to the AWS Serverless Application Repository console details page
```
jfuss: Something I have been thinking about recently, is communicating next things customers can do after the command runs. You can use `sam build` as a guide/model for this.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+Implementation
+==============
+
+CLI Changes
+-----------
+*Explain the changes to command line interface, including adding new commands, modifying arguments etc*
+
+1. Add a new top-level command called ``sam publish app`` with the following options.
+
+  -t, --template PATH      AWS SAM template to publish.
```
jfuss: Can you update this to what `sam publish app --help` would print out? It helps get a wholistic view of the command.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,230 @@
...
+N/A
+
+**How do you validate new .samrc configuration?**
+
+N/A
+
+Documentation Changes
+---------------------
+
+We will document how to use the new ``sam publish app`` command for publishing SAR applications, and link to
```
jfuss: There are more docs that are needed for this. We need this command added here: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-command-reference.html
And docs on this page that explain the command, usage, troubleshooting, etc.

designs/sam_publish_app_cmd.rst
paoptu023: Actually `--make-private`, `--account-ids` are mutually exclusive options. `--make-private` is a boolean flag, it restricts the app to only be viewable by owner. `--account-ids TEXT` will share the app with provided accounts. You won't need to provide your own account id with `--make-private`. Maybe I should use a more descriptive name for the `--account-ids` option?

designs/sam_publish_app_cmd.rst
paoptu023: Customers can use aws cli to do this: https://docs.aws.amazon.com/serverlessrepo/latest/devguide/access-control-resource-based.html#application-permissions. We want to simplify the sam command interface since usually people want to give all permissions when they share app.

The command will fail if the referred nested app hasn't been created yet. Customers have to provide application id to nest the app in template, and publish doesn't touch those apps.

designs/sam_publish_app_cmd.rst
paoptu023: Yes they are.

designs/sam_publish_app_cmd.rst
paoptu023: If customers decide to update metadata w/ new version and flip the app from public or shared back to private. Probably won't be a popular command, but may still be useful.

designs/sam_publish_app_cmd.rst
paoptu023: Yes, it grants a `Deploy` permission to principal "*" under the hood, which enables all actions.

designs/sam_publish_app_cmd.rst
paoptu023: I'll add to the design doc that `Deploy` is granted to the principal when shared.

designs/sam_publish_app_cmd.rst
paoptu023: Nice suggestion, added what customers can do after the application is published.

designs/sam_publish_app_cmd.rst
paoptu023: Updated.

designs/sam_publish_app_cmd.rst
paoptu023: Thanks! I added a few, let me know if there are more.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,342 @@
...
+To publish an app to AWS Serverless Application Repository, customers need to go through the following steps: first upload
+the application code and SAM template to an Amazon S3 bucket, correctly set S3 bucket policy that grants the service read
+permissions for artifacts uploaded to S3, then open the AWS Serverless Application Repository console and provide information
+in a bunch of input boxes. If they use the AWS CLI, they need to pass all the information as parameters, and it's easy to make
+a mistake while typing in the command line.
+
+
+What will be changed?
+---------------------
+In this proposal, we will be providing a new command, ``sam publish app``, which takes a SAM template as input and publishes
```
TheSriram: Sorry if this is a silly question, does sam template need to have a resource or something that say this is a SAR app?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,342 @@
...
+need to provide application metadata information in the template, then ``sam package`` will handle uploading local files to S3,
+and ``sam publish app`` will create the app in Serverless Application Repository. We will also provide sharing options to set
+application permission policies.
+
+
+Success criteria for the change
+-------------------------------
+#. Support all the following use cases:
+
+   * Create new application and its first version in SAR using ``sam publish app``
```
TheSriram: so is sam publish app a constant upsert operation?

TheSriram: I dont know if we should have all this interwoven logic all in one command. Thoughts?

designs/sam_publish_app_cmd.rst
TheSriram: Never mind, I see this covered below.

```AWS::ServerlessRepo::Application```

File: designs/sam_publish_app_cmd.rst (hunk 3 above)
Comment (TheSriram) [reply designs/sam_publish_app_cmd.rst]: yeah, we need to make the distinction about it being mutually exclusive somewhere, maybe in the help text as well.

File: designs/sam_publish_app_cmd.rst [hunk 16]
```diff
@@ -0,0 +1,342 @@
...
+
+CLI Changes
+-----------
+*Explain the changes to command line interface, including adding new commands, modifying arguments etc*
+
+1. Add a new top-level command called ``sam publish app`` with the following help message.
+
+.. code-block:: text
+
+  Usage: samdev publish app [OPTIONS]
```
TheSriram: samdev -> sam

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,342 @@
...
+Build Lambda source code
+  Run ``sam build -t template.yaml -b ./build -o built-template.yaml`` to build all functions in the template and output
+  a SAM template that can be run through the package command.
+
+Package built artifacts and local file references
+  Run ``sam package --template-file built-template.yaml --output-template-file packaged.yaml --s3-bucket my-bucket``
+  to upload code artifacts, readme and license files to S3 and generate the packaged template.
+
+Create new application in SAR
+  Run ``sam publish app -t ./packaged.yaml`` to publish a new application named my-app in SAR with the first version
```
TheSriram: usually, the very first version is not 1.0.0? as in the first version is more like 0.x? before user decides to make a flip and call it thier finised product? Just a thought.

designs/sam_publish_app_cmd.rst
paoptu023: Yeah it's always a upsert operation. We want to simplify the user experience so that whenever customers have edited their apps locally, they can simply run `sam publish app -t template.yaml` to publish the changes. The command either creates a new app if it doesn't exist, or updates the app with local changes. The interwoven logic is actually hidden in implementation details so that customers don't need to be aware of.

designs/sam_publish_app_cmd.rst
paoptu023: This is probably a bad example, but SemanticVersion number defined the template will be used. ☺️

designs/sam_publish_app_cmd.rst
paoptu023: Thanks!

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,342 @@
...
+Success criteria for the change
+-------------------------------
+#. Support all the following use cases:
+
+   * Create new application and its first version in SAR using ``sam publish app``
+   * Create new version of existing SAR application using ``sam publish app``
+   * Update application metadata of existing SAR application using ``sam publish app``
+   * Share the app publicly using the ``--make-public`` option
+   * Make the app private using the ``--make-private`` option
+   * Share the app privately with other AWS accounts using the ``--account-ids`` option
```
TheSriram: Something I just thought of. can we get rid of --make-public, --make-private, by just allowing --accounts-ids = ["*"](public) and yanking it "*"(private)?

paoptu023: I removed `--make-public` and `--make-private`: https://github.com/awslabs/aws-sam-cli/pull/844/commits/3098f8309efcd03f4a6b9c99ebea8bc985ee9d1f

paoptu023:
Currently, when an app is first created by `sam publish app` it will be private by default. If no `--acount-ids` option is specified, the subsequent publish command runs don't change permissions. However, the following workflow may be confusing:

1. Run `sam publish app -t template.yaml` - app is created as private
2. Run `sam publish app -t template.yaml --acount-ids '*'` - app is updated and made public
3. Run `sam publish app -t template.yaml` - app is public

The permission state is different between 1 and 3.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,339 @@
+.. contents:: **Table of Contents**
+   :depth: 2
+   :local:
+
+``sam publish app`` command
+====================================
+
+This is the design for a command to publish an application to `AWS Serverless Application Repository (SAR)`_ with a SAM
+template. It can be used to create a new application and its first version, update exisitng application's metadata, create
```
jlhood:
```suggestion
template. It can be used to create a new application and its first version, update existing application's metadata, create
```

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,339 @@
...
+
+
+Success criteria for the change
+-------------------------------
+#. Support all the following use cases:
+
+   * Create new application and its first version in SAR using ``sam publish app``
+   * Create new version of existing SAR application using ``sam publish app``
+   * Update application metadata of existing SAR application using ``sam publish app``
+   * Share the app publicly using ``--acount-ids ['*']``
```
jlhood: I strongly disagree with consolidating `--make-public`, `--make-private`, and `--account-ids` into just `--account-ids`. It degrades the customer experience.

Also, spelling: `--acount-ids` -> `--account-ids`

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,339 @@
...
+    "Author": "qwang",
+    "Description": "description",
+    "ReadmeUrl": "s3://test/README.md"
+    ...
+  }
+  Shared Application Publicly
+  Click the link below to view your application in AWS console:
+  https://console.aws.amazon.com/serverlessrepo/home?region=<region>#/published-applications/<arn>
+
+Update the metadata of an exsiting application without creating new version
```
jlhood:
```suggestion
Update the metadata of an existing application without creating new version
```

designs/sam_publish_app_cmd.rst
jlhood: Yeah I'm not sure why people seem to fear 1.0.0 releases. 😝

designs/sam_publish_app_cmd.rst
paoptu023: Nice catch! I do this all the time 😅

designs/sam_publish_app_cmd.rst
jlhood: We could add a confirmation on `--make-private`. I agree it's generally bad practice as a publisher, but it seems odd to make people use an entirely different mechanism, e.g., AWS CLI. If I accidentally made an app that was meant to be privately shared public, I'd want to undo it as quickly as possible.

jfuss: > Since the CLI is provided as a convenient tool, we probably shouldn't make it so easy to revoke shared apps.

Taking this a step further, then we shouldn't be changing permissions or operating on permissions.

> If I accidentally made an app that was meant to be privately shared public, I'd want to undo it as quickly as possible.

If you did this through another means, why wouldn't you do it through that means. If SAM CLI doesn't do permissions and we just publish whatever is in the template, this isn't an issue.

I see value in the use-case but do not see how it connects to publish. Publish parses the metadata and publishes the app or template to SAR. I don't think publish is the right flow or command to be managing permissions. Especially when they are not granular. These options provide connivence during publish but a version of an app doesn't have Permissions right? It is defined on the application itself. To me, we are doing to many things in publish and the scope grows from I need to get this into SAR to I need to get this into SAR and can manage things. I don't think publish is the right action for this use-case.

paoptu023: An early design had `sam publish app` and `sam share app`, but permissions got merged into publish after an UX review. After discussing with James and Salman, we agreed that `sam publish app` is doing too many things, and we should take permissions out of scope. `sam share app` could be added as a follow-up.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,297 @@
...
+  Run ``sam build -t template.yaml -b ./build -o built-template.yaml`` to build all functions in the template and output
+  a SAM template that can be run through the package command.
+
+Package built artifacts and local file references
+  Run ``sam package --template-file built-template.yaml --output-template-file packaged.yaml --s3-bucket my-bucket``
+  to upload code artifacts, readme and license files to S3 and generate the packaged template.
+
+Create new application in SAR
+  Run ``sam publish app -t ./packaged.yaml`` to publish a new application named my-app in SAR with the first version
+  created as 0.0.1. If no permission option is passed, the app will be created as private by default.
```
jlhood: Missed a few places where it's still talking about permissions. Search the page for "permission" to find them.

paoptu023: Thanks!

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+  to upload code artifacts, readme and license files to S3 and generate the packaged template.
+
+Create new application in SAR
+  Run ``sam publish app -t ./packaged.yaml`` to publish a new application named my-app in SAR with the first version
+  created as 0.0.1. The app will be created as private by default. SAM CLI prints application created message, metadata
+  used to create application and link to the console details page.
+
+  >>> sam publish app -t ./packaged.yaml
+  Publish Succeeded
+  Created new application with the following metadata:
```
heitorlessa: Instead of giving metadata of the published app (e.g. sam app info instead) it'd be great to leave a ``AWS::Serverless::Application`` resource filled out with the information about this app newly published so one can go copy/paste into somewhere (slack, template, etc.).

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+
+    Resources:
+        HelloWorldFunction:
+            Type: AWS::Lambda::Function
+            Properties:
+              ...
+              CodeUri: ./source-code1
+              ...
+
+Build Lambda source code
```
sthulb: This could be something `sam app publish` could do out of the box. Detect the last build date or diff the prebuild app in `$PROJECT_DIR/.aws-sam/build` against the `$PROJECT_DIR`

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+
+    This command expects the template's Metadata section to contain an
+    AWS::ServerlessRepo::Application section with application metadata
+    for publishing. For more details on this metadata section, see
+    https://docs.aws.amazon.com/serverlessrepo/latest/devguide/serverless-app-publishing-applications.html
+
+    Examples
+    --------
+    To publish an application
+    $ sam publish app -t packaged.yaml --region <region>
```
heitorlessa: Maybe it's out of the scope but one of the great benefits of SAR is to share with other Accounts (or make it public), does it make sense to add a flag to facilitate that process so one doesn't have to go through the IAM process to get this done?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
+.. contents:: **Table of Contents**
+   :depth: 2
+   :local:
+
+``sam publish app`` command
+====================================
+
+This is the design for a command to publish an application to `AWS Serverless Application Repository (SAR)`_ with a SAM
+template. It can be used to create a new application w/ its first version, update existing application's metadata, and
```
sthulb: This could be something we could auto suggest, or implement flags for --bump-minor/patch

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+    AWS::ServerlessRepo::Application section with application metadata
+    for publishing. For more details on this metadata section, see
+    https://docs.aws.amazon.com/serverlessrepo/latest/devguide/serverless-app-publishing-applications.html
+
+    Examples
+    --------
+    To publish an application
+    $ sam publish app -t packaged.yaml --region <region>
+
+  Options:
```
sthulb: Could we add the ability to share account IDs or chose to make the app completely public?

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+in a bunch of input boxes. If they use the AWS CLI, they need to pass all the information as parameters, and it's easy to make
+a mistake while typing in the command line.
+
+
+What will be changed?
+---------------------
+In this proposal, we will be providing a new command, ``sam publish app``, which takes a SAM template as input and publishes
+an application to AWS Serverless Application Repository using applicaiton metadata specified in the template. Customers
+need to provide application metadata information in the template, then ``sam package`` will handle uploading local files to S3,
+and ``sam publish app`` will create the app in Serverless Application Repository.
```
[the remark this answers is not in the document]
jfuss: We are remaining with `sam publish`. commands should be actions not destinations.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+    Resources:
+        HelloWorldFunction:
+            Type: AWS::Lambda::Function
+            Properties:
+              ...
+              CodeUri: ./source-code1
+              ...
+
+Build Lambda source code
+  Run ``sam build -t template.yaml -b ./build -o built-template.yaml`` to build all functions in the template and output
```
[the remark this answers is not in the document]
jfuss: This is improvement outside of this command scope and something we are thinking about wider (invoke, package, etc) doing this build if not there.

designs/sam_publish_app_cmd.rst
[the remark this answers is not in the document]
sthulb: @jfuss what happens when we enhance the functionality? Do we end up with `sam delete app`?

designs/sam_publish_app_cmd.rst
paoptu023: Thanks for the suggestion! This can be added in a follow-up iteration to output CFN resource of the published app.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+Package built artifacts and local file references
+  Run ``sam package --template-file built-template.yaml --output-template-file packaged.yaml --s3-bucket my-bucket``
+  to upload code artifacts, readme and license files to S3 and generate the packaged template.
+
+Create new application in SAR
+  Run ``sam publish app -t ./packaged.yaml`` to publish a new application named my-app in SAR with the first version
+  created as 0.0.1. The app will be created as private by default. SAM CLI prints application created message, metadata
+  used to create application and link to the console details page.
+
+  >>> sam publish app -t ./packaged.yaml
```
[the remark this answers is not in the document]
paoptu023: Thanks, please see comment above: https://github.com/awslabs/aws-sam-cli/pull/844#discussion_r242308706

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
+.. contents:: **Table of Contents**
+   :depth: 2
+   :local:
+
+``sam publish app`` command
```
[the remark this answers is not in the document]
paoptu023: As Jacob has commented, the command itself should be an action instead of an entity to remain consistent with other sam commands.

designs/sam_publish_app_cmd.rst
```diff
@@ -0,0 +1,296 @@
...
+**How do you validate new .samrc configuration?**
+
+N/A
+
+Documentation Changes
+---------------------
+
+1. SAM specification updates:
+
+  - Add "AWS::ServerlessRepo::Application" sepc in `SAM specification`.
```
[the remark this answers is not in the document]
paoptu023: Nice catch!

designs/sam_publish_app_cmd.rst
paoptu023: We thought about doing this like [npm version](https://docs.npmjs.com/cli/version), but this is out of scope for publish currently.

paoptu023:
Removed the app subcommand from the design. Since we only support Serverless App Repo as the destination for now, `app` seems redundant and ambiguous. A destination option can be added in the future if we are going to support other destinations.

paoptu023:
Closing as released in [v0.10.0](https://github.com/awslabs/aws-sam-cli/releases/tag/) 🎉

jfuss:
@paoptu023 We want this merged. Designs are meant to be apart of the feature and exist to help understand past desicions and what is in or out of scope. Please resubmit the content that was in this PR. I can’t reopen since you deleted the branch.

PR #845, opened 2018-12-08
[WIP] White-list nodejs for aws-lambda-builders

gojko:
*Issue #, if available:* https://github.com/awslabs/aws-sam-cli/issues/798

*Description of changes:* White-list nodejs for aws-lambda-builders, required to use the `nodejs_npm` builder from SAM CLI. This should be merged only after https://github.com/awslabs/aws-lambda-builders/pull/44 is merged


jfuss:
@gojko Thanks for submitting this. I have just merged the NPM building into aws-lambda-builders. We still need to test this end-to-end from sam cli and publish a new version of the aws-lambda-builders. We will need to include that version into this PR otherwise, building node will not work.

I am doing some testing locally and using the `--use-container` flag. I can submit the integ tests and version bump here if you would like?

gojko:
any version of node is OK for this (it's executing the `npm` binary). I think even ancient versions such as 0.10 will still pack the files correctly. There might be problems executing custom scripts, but in that case NPM would explode with a reasonable error message and users would know how to fix it. 

I've not tried the `--use-container` option when testing locally, I guess that's a good note for future builder implementations to remember.

jfuss:
Testing locally with a dev build of the container is not trivial right now. :(

@gojko How should we go about validating? The build for python checks for [explicit versions](https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/validate.py#L14). By the sounds of it, we only care about npm? I am not super familiar with the nodejs world and how to go about this right now.

gojko:
@jfuss I think the python validation problem is specific because SAM is a python app building other python apps potentially needing different virtualenvs. My best guess is that for node we only care about `npm` existing. 

Until someone complains and actually has a problem that we could test against I suggest we don't do any additional validation apart from checking for NPM.

jfuss:
@gojko I updated the nodejs init examples and updated the `aws-lambda-builders` to 0.0.4 but can't push this to your remote so it shows up in this PR. Then the only thing left would be to write integ tests for building node.

jfuss:
@TheSriram When you get some time, can you please review. 

@gojko If you could look at the changes to the nodejs example I made, I would really appreciate it. I want to make sure the example aligns with what the community would expect. My one concern is that the `package.json` file in the function folder makes it hard to test the application through unit, functional, or integration tests. You have to `cd` into that directory in order to run `npm run test` and not sure if there is a practice to get around this.

samcli/lib/build/app_builder.py
```diff
@@ -51,6 +51,12 @@ def _get_workflow_config(runtime):
             dependency_manager="pip",
             application_framework=None,
             manifest_name="requirements.txt")
+    elif runtime.startswith("nodejs"):
```
TheSriram: Noo, lets not do if else's again, this might become like our entrypoint file. Given that there may be more runtimes that we support building.

samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/template.yaml
```diff
@@ -16,7 +16,7 @@ Resources:
     HelloWorldFunction:
         Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
         Properties:
-            CodeUri: hello_world/
+            CodeUri: hello-world/
```
TheSriram: just curious: any reason for the change in the name?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -124,3 +124,71 @@ def test_unsupported_runtime(self):
...
+
+class TestBuildCommand_NodeFunctions(BuildIntegBase):
+
+    EXPECTED_FILES_GLOBAL_MANIFEST = set()
+    EXPECTED_FILES_PROJECT_MANIFEST = {'node_modules', 'main.js'}
+    EXPECTED_NODE_MODULES = {'minimal-request-promise'}
+
+    FUNCTION_LOGICAL_ID = "Function"
+
+    @parameterized.expand([
```
TheSriram: Brilliant! thanks for adding all cases!

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -124,3 +124,71 @@ def test_unsupported_runtime(self):
...
+        process = subprocess.Popen(cmdlist, cwd=self.working_dir)
+        process.wait()
+
+        self._verify_built_artifact(self.default_build_dir, self.FUNCTION_LOGICAL_ID,
+                                    self.EXPECTED_FILES_PROJECT_MANIFEST, self.EXPECTED_NODE_MODULES)
+
+        self._verify_resource_property(str(self.built_template),
+                                       "OtherRelativePathResource",
+                                       "BodyS3Location",
+                                       os.path.relpath(
```
TheSriram: Thats a lotta path heh.

samcli/lib/build/app_builder.py
jfuss: What is your suggestion then?

samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/template.yaml
jfuss: Conversations with people familiar with Node.js said dashes are used not underscores. So I changed this to make it match what the community is doing/the Node.js style

tests/integration/buildcmd/test_build_cmd.py
jfuss: Mostly stolen from the python above.

samcli/lib/build/app_builder.py
TheSriram: Suggestion is to be encapsulate in a dictionary, and invoke corresponding actions based on a match. But I will not block on this. we can address this separately, I'll create an issue.

gojko:
> @gojko If you could look at the changes to the nodejs example I made, I would really appreciate it. I want to make sure the example aligns with what the community would expect. My one concern is that the `package.json` file in the function folder makes it hard to test the application through unit, functional, or integration tests. You have to `cd` into that directory in order to run `npm run test` and not sure if there is a practice to get around this.

It looks fine to me. My workflow is usually to whitelist files/directories by including them into `package.json` files section, so instead of excluding tests via `.npmignore`, I'd create something like 

```
files: ['app.js']
```

but both approaches are equally valid. Regarding test execution, I think this is OK. There are tools like [lerna](https://github.com/lerna/lerna) that help manage multi-directory projects, but there isn't anything that could be called standard, at least not that I know of it. I usually just script it using a makefile.

PR #852, opened 2018-12-12
Implement sam publish app command

paoptu023:
*Description of changes:*
* Include implementation, unit tests and integration tests for awslabs#844.
* `make dev` passed.
* Test coverage:
samcli/commands/publish/__init__.py                                 4      0      0      0   100%
samcli/commands/publish/app/__init__.py                             0      0      0      0   100%
samcli/commands/publish/app/cli.py                                 55      0      8      0   100%


paoptu023:
I'll expand this PR to include implementation and unit tests, just want to send out the interface first to get reviewed.

paoptu023:
I ran integration tests locally and they worked. According to [here](https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions), Travis doesn't allow encrypted variables to be used with forked repositories, so the Travis build failed.

tests/integration/testdata/publish/template_create_app.yaml
```diff
@@ -0,0 +1,38 @@
...
+  Function:
+    Timeout: 3
+Metadata:
+  AWS::ServerlessRepo::Application:
+    Author: user
+    Description: description
+    HomePageUrl: https://github.com/test/test
+    Labels:
+    - test-app
+    LicenseUrl: s3://test-sam-publish-app/LICENSE.txt
```
TheSriram: we want the integration tests to work without files being already present in S3, right? Anyone should be able to run the integration tests, can we Ref a local one or upload new files to s3 during integ test and then parameter over-ride them.

tests/integration/publish/app/test_cli_integ.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # Create application for each test
+        app_metadata_text = self.test_data_path.joinpath("metadata_create_app.json").read_text()
+        app_metadata = json.loads(app_metadata_text)
+        app_metadata['TemplateBody'] = self.test_data_path.joinpath("template_create_app.yaml").read_text()
+        response = self.sar_client.create_application(**app_metadata)
+        self.application_id = response['ApplicationId']
+
+        # Avoid race conditions
+        time.sleep(1)
```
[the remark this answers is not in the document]
paoptu023: I ran into intermittent errors that the application already exists because create_application is called immediately after delete_application, and another conflict exception would happen when I call update_application immediately after create_application.

tests/integration/testdata/publish/template_create_app.yaml
paoptu023: Thanks for the suggestion! I refactored the tests to run without assuming S3 file existence.

samcli/commands/local/cli_common/user_exceptions.py
```diff
@@ -66,3 +66,10 @@ class UnsupportedIntrinsic(UserException):
     Value from a template has an Intrinsic that is unsupported
     """
     pass
+
+
+class S3PermissionsRequired(UserException):
```
jlhood: Is this defined under the local command due to historical reasons?

samcli/commands/publish/app/cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if CREATE_APPLICATION in publish_output['actions']:
+        return "Created new application with the following metadata:\n{}".format(details)
+
+    return 'The following metadata of application "{}" has been updated:\n{}'.format(application_id, details)
+
+
+def _print_console_link(region, application_id):
+    """
+    Print link for the application in AWS Serverless Repository console.
```
jlhood:
```suggestion
    Print link for the application in AWS Serverless Application Repository console.
```

samcli/commands/publish/app/cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Parameters
+    ----------
+    region : str
+        AWS region name
+    application_id : str
+        The Amazon Resource Name (ARN) of the application
+
+    """
+    url = "https://console.aws.amazon.com/serverlessrepo/home?region={}#/published-applications/{}"
```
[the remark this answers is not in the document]
jlhood: A direct console link is the right dev experience. We are not going to change that link structure so this should be fine. SAR APIs are decoupled from the console so API output should never reference console links.

samcli/commands/publish/app/cli.py
```diff
@@ -0,0 +1,137 @@
...
+    Returns
+    -------
+    Exception
+        S3PermissionsRequired if S3 related or ClientError
+    """
+    error_code = ex.response['Error']['Code']
+    message = ex.response['Error']['Message']
+
+    if error_code == 'BadRequestException' and "Failed to copy S3 object" in message:
+        return S3PermissionsRequired(
```
jlhood: Nice. Glad to have clear instructions for this error!

samcli/commands/publish/app/cli.py
```diff
@@ -0,0 +1,137 @@
...
+
+def do_cli(ctx, template):
+    """Publish the application based on command line inputs."""
+    if not os.path.exists(template):
+        click.secho("Publish Failed", fg='red')
+        raise SamTemplateNotFoundException("Template at {} is not found".format(template))
+
+    with click.open_file(template, 'r') as template_file:
+        try:
+            output = publish_application(template_file.read())
```
jlhood: Wasn't there a change to the serverlessrepo module so you could pass in a client? How is the client getting initialized with the correct credentials?

tests/integration/publish/app/test_cli_integ.py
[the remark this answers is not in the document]
jlhood: Are you using the same application name over and over again? I'd suggest adding a UUID to the application name for each test run so the data for each test is isolated. Also, I'm not a fan of sleeps in integ tests, and an arbitrary sleep time won't close the race condition completely. If they're happening to the tests, can't they happen to our customers too? Can we make the logic of the command (or the test if it's test-specific) robust enough that it can handle conflict errors?

tests/integration/testdata/publish/main.py
```diff
@@ -0,0 +1,2 @@
+def lambda_handler(event, context):
```
jlhood: Is this necessary? Does the publish command validate and error if it doesn't find the actual file?

tests/integration/testdata/publish/template_create_app_version.yaml
```diff
@@ -0,0 +1,38 @@
...
+  Function:
+    Timeout: 3
+Metadata:
+  AWS::ServerlessRepo::Application:
+    Author: user
+    Description: description
+    HomePageUrl: https://github.com/test/test
+    Labels:
+    - test-app
+    LicenseUrl: s3://<bucket-name>/LICENSE
```
jlhood: So the integ tests assume the SAM template has been packaged? What if it hasn't? What does the command do? Should we have tests to verify the expected behavior?

samcli/commands/local/cli_common/user_exceptions.py
jfuss: No these are exceptions for the local suite of commands. The S3PermissionRequired should not live here.

tests/integration/testdata/publish/main.py
paoptu023: I think CreateApplication looks for Resource in the template, and will error out if the CodeUri under lambda resource is not a valid S3 path.

samcli/commands/publish/app/cli.py
paoptu023: The region/profile is set through sam cli [context](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/cli/context.py#L77) and passed to every command annotated with [aws_creds_options](https://github.com/awslabs/aws-sam-cli/pull/852/files/cc69a37f1e533b6321e588ec1f889eb2f9475c46#diff-2237f386debbe0f92032b8e0ceda2ae5R42). The boto3 client is initialized through `boto3. setup_default_session` therefore passing a sar_client is unnecessary.

tests/integration/testdata/publish/template_create_app_version.yaml
paoptu023: I added exception handling for using not packaged template and unit/integration tests.

tests/integration/publish/app/test_cli_integ.py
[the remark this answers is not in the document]
paoptu023: Thanks! The sleep logic is test-specific, it's not needed it anymore if I create a UUID application name each time.

paoptu023:
Tested on Mac with Python2.7, 3.6 and 3.7:
* `make dev` passed and integration tests for `publish` passed.

Tested on Windows with Python2.7, 3.6 and 3.7:
* All unit tests and integration tests for publish passed.
* Pylint passed.
* Flake8 passed for `/samcli`. Somehow flake8 doesn't recognize `.flake8` config for `/tests`, I couldn't figure out why.

jfuss:
@paoptu023 

What do you mean by:
"I couldn't run other integration tests, so I just confirmed that publish is working."? Does `make integ-test` not work? What is failing?

"Flake8 passed for /samcli. Somehow flake8 doesn't recognize .flake8 config for /tests, I couldn't figure out why." Flake8 passed on Travis, so not sure what the issue you are referring to here.

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
+"""CLI command for "publish" command."""
+
+import os
+import json
+import logging
+import click
+
+import boto3
+from botocore.exceptions import ClientError
+
```
jfuss: Keep imports to 3 sections, following PEP8

```
# Python standard library

# imports for dependencies

# imports for modules in the codebase
```

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+            click.secho("Publish Succeeded", fg="green")
+            click.secho(_gen_success_message(output), fg="yellow")
+        except ServerlessRepoError as ex:
+            click.secho("Publish Failed", fg='red')
+            raise UserException(str(ex))
+        except ClientError as ex:
+            click.secho("Publish Failed", fg='red')
+            raise _wrap_s3_uri_exception(ex)
+
+        application_id = output['application_id']
```
jfuss: Is `application_id` always in the response? Either way, we should be doing `output.get('application_id')` and have a safe default. Otherwise this throws a KeyError and blows up the command line with a stack trace.

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+
+def do_cli(ctx, template):
+    """Publish the application based on command line inputs."""
+    if not os.path.exists(template):
+        click.secho("Publish Failed", fg='red')
+        raise SamTemplateNotFoundException("Template at {} is not found".format(template))
+
+    with click.open_file(template, 'r') as template_file:
+        try:
+            output = publish_application(template_file.read())
```
jfuss: Can we name output to be more specific?

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+    do_cli(ctx, template)  # pragma: no cover
+
+
+def do_cli(ctx, template):
+    """Publish the application based on command line inputs."""
+    if not os.path.exists(template):
+        click.secho("Publish Failed", fg='red')
+        raise SamTemplateNotFoundException("Template at {} is not found".format(template))
+
+    with click.open_file(template, 'r') as template_file:
```
jfuss: What if the file doesn't exist?

We should reuse the template reading method we already have: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/_utils/template.py#L32

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    publish_output : dict
+        Output from serverlessrepo publish_application
+
+    Returns
+    -------
+    str
+        Detailed success message
+    """
+    application_id = publish_output['application_id']
```
jfuss: `.get()` this will throw a KeyError if it does not exist.

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+    details = json.dumps(publish_output['details'], indent=2)
```
jfuss: same

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if CREATE_APPLICATION in publish_output['actions']:
```
jfuss: `.get()`

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    region : str
+        AWS region name
+    application_id : str
+        The Amazon Resource Name (ARN) of the application
+
+    """
+    if not region:
+        region = boto3.Session().region_name
+    url = "https://console.aws.amazon.com/serverlessrepo/home?region={}#/published-applications/{}"
```
jfuss: Constant?

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Returns
+    -------
+    Exception
+        UserException if found invalid S3 URI or ClientError
+    """
+    error_code = ex.response['Error']['Code']
+    message = ex.response['Error']['Message']
+
+    if error_code == 'BadRequestException' and "Invalid S3 URI" in message:
```
jfuss: Is the only way of knowing this through the message? Seems kind of fragile

tests/integration/publish/publish_app_integ_base.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Grant serverlessrepo read access to the bucket
+        bucket_policy = cls.test_data_path.joinpath("s3_bucket_policy.json").read_text()
+        cls.s3_bucket.Policy().put(Policy=bucket_policy)
+
+        # Upload test files to S3
+        license_body = Path(__file__).resolve().parents[3].joinpath("LICENSE").read_text()
+        cls.s3_bucket.put_object(Key="LICENSE", Body=license_body)
+
+        readme_body = Path(__file__).resolve().parents[3].joinpath("README.rst").read_text()
```
jfuss: `Path(__file__).resolve().parents[3]` should be a variable.

tests/integration/publish/publish_app_integ_base.py
```diff
@@ -0,0 +1,92 @@
...
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
+
+class PublishAppIntegBase(TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.region_name = "us-east-1"
```
jfuss: Should we be reading this from the environment instead of explicitly setting?

tests/integration/publish/publish_app_integ_base.py
```diff
@@ -0,0 +1,92 @@
...
+        cls.bucket_name = str(uuid.uuid4())
+        cls.bucket_name_placeholder = "<bucket-name>"
```
jfuss: Is this only to reset the files? Can we use templates and generate temp files? Someone writing a new test class will know to do this and I rather follow a consistent pattern that is always reproducible instead of re-writing to the same file each time.

tests/integration/publish/test_command_integ.py
```diff
@@ -0,0 +1,101 @@
...
+        self.application_id = response['ApplicationId']
+
+    def tearDown(self):
+        # Delete application for each test
+        self.sar_client.delete_application(ApplicationId=self.application_id)
+        # Replace the application name with placeholder in test files
+        self.replace_text_in_file(self.application_name, self.application_name_placeholder)
+
+    def test_update_application(self):
+        template_path = str(self.test_data_path.joinpath("template_update_app.yaml"))
```
jfuss: Or you can cast inside of `get_command_list`

paoptu023:
> @paoptu023
> 
> What do you mean by:
> "I couldn't run other integration tests, so I just confirmed that publish is working."? Does `make integ-test` not work? What is failing?
> 
> "Flake8 passed for /samcli. Somehow flake8 doesn't recognize .flake8 config for /tests, I couldn't figure out why." Flake8 passed on Travis, so not sure what the issue you are referring to here.

I don't know if it's my config issue or what, but almost all the other integration tests either failed or erred. 

```
tests/integration/buildcmd/test_build_cmd.py FsFs.FFFFFF
tests/integration/init/test_init_command.py .
tests/integration/local/generate_event/test_cli_integ.py .
tests/integration/local/invoke/test_integrations_cli.py F.FFFFFFFFFFFFFEFEFEFEFEFEFEFEFEFEFEFEFEFE
tests/integration/local/invoke/runtimes/test_with_runtime_zips.py FFF
tests/integration/local/start_api/test_start_api.py FFFF.FFFFFFFF.FFFFFFFFFFFFFFFFFFFFFFFFFF
tests/integration/local/start_lambda/test_start_lambda.py F.FFFFFFFFF
tests/integration/publish/test_command_integ.py ....
```

Most were caused by `EndpointConnectionError: Could not connect to the endpoint URL: "http://127.0.0.1:30166/2015-03-31/functions/EchoEventFunction/invocations"`

Flake8 didn't pass for `/tests` on my **Windows** machine because the `.flake8` config file wasn't picked up.

tests/integration/publish/test_command_integ.py
```diff
@@ -0,0 +1,101 @@
+import re
+import uuid
+import json
+from subprocess import Popen, PIPE
+
+from .publish_app_integ_base import PublishAppIntegBase
+
+
+class TestPublishExistingApp(PublishAppIntegBase):
```
jfuss: All these test require credentials. Travis will fail with no credential error due to the PR not coming from the same repo. To combat this, we only run tests that have credentials when release (until we have a better solution). https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/local/invoke/test_integrations_cli.py#L18

tests/unit/commands/publish/test_command.py
```diff
@@ -0,0 +1,147 @@
+"""Test sam publish CLI."""
+import json
+from unittest import TestCase
+from mock import patch, call, Mock
+
+from botocore.exceptions import ClientError
+
```
jfuss: Same comment as other PEP8 imports I made above

tests/unit/commands/publish/test_command.py
```diff
@@ -0,0 +1,147 @@
...
+    def setUp(self):
+        self.template = "./template"
+        self.application_id = "arn:aws:serverlessrepo:us-east-1:123456789012:applications/hello"
+        self.ctx_mock = Mock()
+        self.ctx_mock.region = "us-east-1"
+        self.console_link = "Click the link below to view your application in AWS console:\n" \
+            "https://console.aws.amazon.com/serverlessrepo/home?region={}#/published-applications/{}"
+
+        path_patcher = patch('samcli.commands.publish.command.os.path')
```
jfuss: Can we follow what we have done in all of our testing to this point with adding annotations on the methods. Not a huge fan of the start and cleanup things that have to be done and it doesn't match the rest of the codebase.

tests/unit/commands/publish/test_command.py
```diff
@@ -0,0 +1,147 @@
...
+        self.path_mock = path_patcher.start()
+        self.path_mock.exists.return_value = True
```
jfuss: If this changes between tests, don't set the value in the `setUp`. It makes it more difficult to understand each test if the variables aren't assigned consistently.

samcli/commands/publish/command.py
paoptu023: Thanks, I fixed all the imports order.

samcli/commands/publish/command.py
paoptu023: Yeah, this is the only way.

tests/integration/publish/publish_app_integ_base.py
paoptu023: I ended up with generating a temp folder, writing to it based on the template files, then deleting the folder. I can't use tempfile because I need to pass the file path to the command, and [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) doesn't work so well on Windows.

tests/integration/publish/test_command_integ.py
```diff
@@ -90,7 +85,7 @@ def test_create_application(self):
         self.sar_client.delete_application(ApplicationId=application_id)
```
TheSriram: we could probably move deletion of applications if they exist onto the teardown.

tests/integration/publish/test_command_integ.py
```diff
@@ -99,3 +94,21 @@ def test_publish_not_packaged_template(self):
 
         expected_msg = "Please make sure that you have uploaded application artifacts to S3"
         self.assertIn(expected_msg, process_stderr.decode('utf-8'))
+
+    def test_create_application_infer_region_from_env(self):
```
TheSriram: Thanks for adding this test to check for region inference.

tests/integration/publish/test_command_integ.py
paoptu023: Thanks, I moved it to tearDown.

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,131 @@
...
+    # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
+
+    do_cli(ctx, template)  # pragma: no cover
+
+
+def do_cli(ctx, template):
+    """Publish the application based on command line inputs."""
+    template_data = get_template_data(template)
+    try:
+        publish_output = publish_application(yaml.safe_dump(template_data))
```
jfuss: since you are using `get_template_data` you no longer need to do a `yaml.safe_dump` here. This is take care of by `get_template_data`

samcli/commands/publish/command.py
jfuss: :(

tests/integration/publish/publish_app_integ_base.py
jfuss: You can use tempfile instead. Here is an example from our functional tests: https://github.com/awslabs/aws-sam-cli/blob/develop/tests/functional/function_code.py#L122

samcli/commands/publish/command.py
paoptu023: publish_application expects a string as input, so I have to serialize it back to yaml string.

paoptu023: I need to dump `template_data ` because `pubish_application` expects a YAML or JSON template as string. It doesn't accept template dictionary, which is why I didn't use `get_template_data` initially.

jfuss: The data is expected to be a python Dictionary here. So let's use json.dumps instead of yaml. We only need to use yaml if the data is yaml, which it is not.

paoptu023: I see, I'll use json then.

jlhood: So that means we have no way to preserve the order of the template? Will the template in SAR display as YAML or JSON? We prefer YAML and ideally maintaining the same order the customer gave it to us in. I realize there are several places where we're currently parsing and scrambling the template order, but I'd like to eventually clean all of them up so when a customer builds/packages/publishes, the YAML template looks very close to what it looks like when they edit it. Scrambling the order makes it very hard to read.

samcli/commands/publish/command.py
```diff
@@ -0,0 +1,135 @@
...
+    -------
+    Exception
+        UserException if found invalid S3 URI or ClientError
+    """
+    error_code = ex.response.get('Error').get('Code')
+    message = ex.response.get('Error').get('Message')
+
+    if error_code == 'BadRequestException' and "Invalid S3 URI" in message:
+        return UserException(
+            "You SAM template contains invalid S3 URIs. Please make sure that you have uploaded application "
```
jlhood: You -> Your

samcli/commands/publish/command.py
jfuss: @jlhood That is fair. We should fix this across then instead on one offing, especially when we have one place that does the reading. 

Currently `publish_application` takes a string instead of a dict. If SAR wants to truly control this, I think the `publish_application` should take an OrderedDict and do the serialization instead of relying on the caller to do this action (keeping the template the same).

PR #866, opened 2018-12-18
Add Ruby to Supported Runtime Enum

awood45:
This is required for the Ruby support added to aws-lambda-builders to work in SAM CLI.

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -57,6 +57,12 @@ def _get_workflow_config(runtime):
             dependency_manager="npm",
             application_framework=None,
             manifest_name="package.json")
+    elif runtime.startswith("ruby"):
+        return config(
+            language="ruby",
+            dependency_manager="bundler",
+            application_framework=None,
+            manifest_name="Gemfile")
     else:
         raise UnsupportedRuntimeException("'{}' runtime is not supported".format(runtime))
 


awood45:
I'll say that, right now, the example function is breaking when packaged and deployed. I'm investigating why this is.

samcli/lib/build/app_builder.py
```diff
@@ -57,6 +57,12 @@ def _get_workflow_config(runtime):
             dependency_manager="npm",
             application_framework=None,
             manifest_name="package.json")
+    elif runtime.startswith("ruby"):
```
TheSriram: Woot! sam build for ruby! :)

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
```diff
@@ -124,24 +114,22 @@ aws cloudformation describe-stacks \
...
 
 ## AWS CLI commands
 
-AWS CLI commands to package, deploy and describe outputs defined within the cloudformation stack:
+AWS CLI commands to package, deploy and describe outputs defined within the cloudformation stack after building:
 
 ```bash
 sam package \
-    --template-file template.yaml \
+    --template-file .aws-sam/build/template.yaml \
```
TheSriram: Need to investigate this, thought there was some code that automatically picked this up.

TheSriram:
This looks good, @awood45 do you want to add an integration test for sam build with ruby as well? and we can merge this.

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
jfuss: No need to add `--template-file` when using build anymore. We will auto pick up build if it exists otherwise default to `template.yaml`.

https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/package/__init__.py#L21
and 
https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/_utils/options.py#L35

TheSriram:
The integ tests worked locally, need to get a new release of the builders out to get the new docker images.

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
```diff
@@ -7,7 +7,8 @@ This is a sample template for {{ cookiecutter.project_name }} - Below is a brief
 ├── README.md                   <-- This instructions file
 ├── hello_world                 <-- Source code for a lambda function
 │   ├── app.rb                  <-- Lambda function code
-├── Gemfile                     <-- Ruby dependencies
+│   ├── Gemfile                 <-- Ruby function dependencies
+├── Gemfile                     <-- Ruby test/documentation dependencies
```
jfuss: Not a huge fan of the Double `Gemfile`. How do I know or remember which one to update? Is this absolutely needed? For Node, we had tests under the function directory and excluded them through `.npmignore`

tests/functional/local/apigw/test_local_apigw_service.py
```diff
@@ -630,7 +630,7 @@ def make_service(list_of_routes, function_provider, cwd):
...
+    response_str = '{"httpMethod": "GET", "body": null, "resource": "/something/{event}", "requestContext": {"resourceId": "123456", "apiId": "1234567890", "resourcePath": "/something/{event}", "httpMethod": "GET", "requestId": "c6af9ac6-7b61-11e6-9a41-93e8deadbeef", "accountId": "123456789012", "stage": "prod", "identity": {"apiKey": null, "userArn": null, "cognitoAuthenticationType": null, "caller": null, "userAgent": "Custom User Agent String", "user": null, "cognitoIdentityPoolId": null, "cognitoAuthenticationProvider": null, "sourceIp": "127.0.0.1", "accountId": null}, "extendedRequestId": null, "path": "/something/{event}"}, "queryStringParameters": null, "headers": {"Host": "0.0.0.0:33651", "User-Agent": "python-requests/2.20.1", "Accept-Encoding": "gzip, deflate", "Accept": "*/*", "Connection": "keep-alive"}, "pathParameters": {"event": "event1"}, "stageVariables": null, "path": "/something/event1", "isBase64Encoded": false}' # NOQA
```
jfuss: grrr.. let's just pin requests. We have installers now, so pinning isn't as dangerous. If customers run into conflicts, this gives them a reason to get onto our installers and out of pip (which is what we want/recommend anyways)

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
```diff
@@ -31,27 +32,16 @@ Setup Ruby Version Manager from [Ruby Version Manager](http://rvm.io/)
 Run following commands
 
 ```bash
-rvm install ruby-2.5.0
-rvm use ruby-2.5.0
-rvm --default use 2.5.0
+rvm install ruby-2.5.3
```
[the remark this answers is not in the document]
awood45: It's a best practice, yes. Ruby on AWS Lambda uses version 2.5.3 - while 2.5.0 will technically work for builds, there are security patches present on 2.5.3 that are not present on 2.5.0, so I would not intentionally direct users to use an older version in this case.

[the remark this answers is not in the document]
jfuss: Is there a way the newest just gets installed. I am worried this will be out of date and something we need to maintain.

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
awood45: So...I'm open to ways we can support this through `sam build`. If we had to pick one, we would not include the global test/docs Gemfile, but then the way the example code does testing wouldn't work easily. I don't know how to get SAM CLI to use a Gemfile in the root directory to build vendored dependencies into each function directory in a single command.

It's actually very possible that multiple Gemfiles may be a reality, and that we just need to teach the paradigm. Why?

1. Layers will eventually be a part of this build story, and you only want to specify those dependencies within the layer itself. In fact, we may want builds where we say "don't fetch all dependencies I need, expect some to be resolved in a layer" down the road.
2. I may not want all of my dependencies in every function.
3. I may want global dependencies in my build structure.

There is actually precedent for this, the Ruby SDK repo itself: https://github.com/aws/aws-sdk-ruby/

We have a global Gemfile for testing, building, and documentation generation. Within each modular package, we have its own gemspec for only the required dependencies. This may be the correct pattern for modular Lambda functions.

awood45: I can image an appropriate alternative approach, perhaps where we use "groups" per function to keep dependency versions in sync in a global Gemfile. In an ideal world, we could provide options to the user.

tests/functional/local/apigw/test_local_apigw_service.py
TheSriram: We could remove requests from test requirements (dev.txt) , and pin the requests in the base.txt.

But if we pin, we would periodically need to bump up requests as when there is a release of requests.

samcli/local/init/templates/cookiecutter-aws-sam-hello-ruby/{{cookiecutter.project_name}}/README.md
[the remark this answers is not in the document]
heitorlessa: do we need to provide instructions on how to use install Ruby for Ruby developers?  I wonder if Requirements (ruby 2.5) suffices as we do with Python and Node init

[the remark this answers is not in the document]
jfuss: This is how you install Ruby. Not sure what the question is @heitorlessa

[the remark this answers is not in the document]
heitorlessa: Sure, let me try once more and let me know if it's clearer: 

1. Do we need to tell Ruby developers how they should install Ruby?
2. Customers trying Ruby quickstart should have Ruby installed, shouldn't we assume that and cut the Installation part?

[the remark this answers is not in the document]
sthulb: I wouldn't add install instructions. People who use languages like ruby, go etc will have their tooling in place.

It would be like every pkg internally saying start off by installing the Amazon builder tools.

PR #891, opened 2018-12-27
fix: formatting to md from rst

TheSriram:
- DESIGN
- README
- DEVELOPMENT_GUIDE

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


README.md
```diff
@@ -0,0 +1,115 @@
...
+-   Validate SAM templates
+-   Get started with boilerplate Serverless Service in your chosen
+    Lambda Runtime `sam init`
+
+Get Started
+-----------
+
+Learn how to get started using the SAM CLI with these guides:
+
+-   [Installation](docs/installation.rst): Set up your macOS, Linux or
```
jfuss: Should we use this time to update these links to point to the aws docs we now have?

TheSriram: definitely. the installation docs already do, but need to move out the rest too.

jfuss: I wonder if we should just collapse the installation docs into the readme. Not sure if the extra hop is valuable. Or just link directly to the landing page which has the different installations: https://aws.amazon.com/serverless/sam/

Or link them here: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html but we need to update this one first to make pip an advanced install and default to installers first.

TheSriram: yeah moved it to the landing page, this way the first options seen to install are the official installation options.

PR #892, opened 2018-12-27
chore: Bump up serverlessrepo version

paoptu023:
*Issue #, if available:*
https://github.com/awslabs/aws-serverlessrepo-python/issues/21

*Description of changes:*
[serverlessrepo](https://pypi.org/project/serverlessrepo/) added support for Python 3.4 and 3.5 in version 0.1.6.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,4 +13,4 @@ python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
 requests==2.20.1
 aws_lambda_builders==0.0.5
-serverlessrepo==0.1.5
+serverlessrepo==0.1.6


jfuss:
@paoptu023 Is this the only change? https://github.com/awslabs/aws-sam-cli/pull/891 is fixing our support.

paoptu023:
Yeah this is the only change.

jfuss:
If that is the case, I am not inclined to add this version. SAM CLI does not support 3.5 or 3.4 and is being addressed in #891. It looks like we have had an issue with this not correctly working in PyPi for a while :( but isn't supported.

jfuss:
Closing since the only change is Python support and for Python version SAM CLI does not support.

Issue #893, opened 2018-12-27
Support Templates written through CDK through reading Resource Level Metadata


jfuss:
### Describe your idea/feature/enhancement

Currently, AWS SAM CLI only supports templates that are written in CloudFormation (`.json` or `.yaml` files). This leaves out other frameworks like CDK that produce a template. SAM CLI should support these use-cases. 

### Proposal

A design will need to be written for this support. CDK is adding some support to enable the bridging [1] of the tools. The design should capture how we will support templates written through CDK but should keep in mind how we can expand to other frameworks as well (Serverless Framework for example).

[1] https://github.com/awslabs/aws-cdk/pull/1433/

### Additional Details
Other related issues
* https://github.com/awslabs/aws-cdk/issues/716

tj:
Personally I find it pretty elegant that anything that wants to work with SAM can just generate some .json or .yaml

eladb:
@tj 👋

jfuss:
The design and implementation for supporting Resource Level Metadata is done and merged. Only thing left is to release. Updating labels and the issue title to be more accurate (instead of closing and opening another one to track release).

jfuss:
This will also allow people using Amplify to run and test locally with SAM CLI: https://github.com/aws-amplify/amplify-cli/pull/693

jfuss:
Released in [v0.11.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.11.0)

Closing

eladb:
Yey!

PR #907, opened 2019-01-03
feat(invoke): reading Metadata adding to Resources in a template

jfuss:
💻 [**Rendered Design Doc is here**](https://github.com/jfuss/aws-sam-cli/blob/experimental/cdk/designs/template_generation_through_frameworks.md)
🗓 **RFC Open for feedback until**: 1/10/2018

* Issue *
#893

*Description of changes:*

This design is meant to capture how we can directly support templates that are embedded with Metadata CDK appended into the template. This will allow customers to synthesize templates written in CDK and locally debug/invoke their functions through SAM CLI. While understanding CDK Metadata is the only thing in scope for this design, it is left open to allow further support and expansion (most likely through another command) that can generate the AWS CloudFormation templates directly within SAM CLI.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/designs/resource_metadata_overriding.md
+++ b/designs/resource_metadata_overriding.md
@@ -1,10 +1,11 @@
-Support template generation through other Frameworks
-====================================================
+Understand Resource Level Metadata
+==================================
 
 This is a design to capture how SAM CLI can support templates that are generated 
 from different frameworks, e.g. AWS Cloud Development Kit.
 
-Initially, the support will only be for processing the CDK metadata that is appended into a template.
+Initially, the support will only be for processing Resource Metadata within the template, which enables support for
+customers using AWS Cloud Development Kit (CDK).
 
 What is the problem?
 --------------------
@@ -20,9 +21,9 @@ applications.
 What will be changed?
 ---------------------
 
-To start, we will add support for processing metadata from CDK applications:
+To start, we will add support for processing Resource Metadata that is embedded into the template:
 SAM CLI will add a processing step on the templates it reads. This will consist of reading
-the template and for each resource reading the metadata and replacing values as specified.
+the template and for each resource reading the Metadata and replacing values as specified.
 
 In the future, we can support creating these templates from the different frameworks in a command directly within 
 SAM CLI but is out of scope in the initial implementation of support.
@@ -30,18 +31,18 @@ SAM CLI but is out of scope in the initial implementation of support.
 Success criteria for the change
 -------------------------------
 
-* Ability to invoke functions locally that was defined in AWS Cloud Development Kit (CDK).
+* Ability to invoke functions locally that contain Metadata on a Resource
 * Process a template with CDK Metadata on a resource.
 
 Out-of-Scope
 ------------
 
 * A command that will generate the template from the framework.
-* Handling multiple stacks.
-* Support for frameworks other than CDK.
 
 User Experience Walkthrough
 ---------------------------
+
+CDK is a framework that appends this Metadata to Resources within a template and will use this as an example.
 
 ### Customer using CDK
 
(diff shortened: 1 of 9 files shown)


jfuss:
As an FYI: @SoodDhruv @eladb @fulghum

designs/template_generation_through_frameworks.md
```diff
@@ -0,0 +1,145 @@
...
+All the providers, which are used to get resources out of the template provided to the command, call 
+`SamBaseProvider.get_template(template_dict, parameter_overrides)` to get a normalized template. This function call is 
+responsible for taking a SAM template dictionary and returning a cleaned copy of the template where SAM plugins have 
+been run and parameter values have been substituted. Given the current scope of this call, expanding it to also normalize 
+metadata, seems reasonable. We will expand `SamBaseProvider.get_tempalte()` to call a `CdkTemplateNormalizer` class
+that will be responsible for understanding the metadata and normalizing the template with respect to the metadata. 
+
+```python
+
+class CdkTemplateNormalizer(object):
```
jfuss: Thinking about this more. I think this should be a metadata normalized instead of tying it to CDK. CDK happens to be the first to add this type of metadata but nothing here forces it to be from CDK. 

I will work on removing references to that directly and center the design around the Metadata instead. I will also add a snippet of the Metadata on a Resource (slipped my mind when initially writing)

eladb: +1 we designed the metadata keys (aws:asset:*) to not be tied directly to cdk with the purpose of allowing other frameworks to utilize the same approach.

designs/resource_metadata_overriding.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+The two keys we will recognize are `aws:asset:path` and `aws:asset:property`. `aws:asset:path`'s value will be the path
+to the code, files, etc that are help on the machine, while `aws:asset:property` is the Property of the Resource that
+needs to be replaced. So in the example above, the `Code` Property will be replaced with `/path/to/function/code`.
+
+Below algorithm to do this Metadata Normalization on the template.
+
+```python
+class TemplateMetadataNormalizer(object):
```
TheSriram: Is this normalizer going to be injected? can we have more than one normalizer such that we can chain them?

designs/resource_metadata_overriding.md
```diff
@@ -0,0 +1,170 @@
...
+          S3Bucket: mybucket
+          S3Key: myKey
+        ...
+      Metadata:
+        aws:asset:path: '/path/to/function/code'
+        aws:asset:property: 'Code'
+```
+
+The two keys we will recognize are `aws:asset:path` and `aws:asset:property`. `aws:asset:path`'s value will be the path
+to the code, files, etc that are help on the machine, while `aws:asset:property` is the Property of the Resource that
```
eladb: ״help on the machine״?

jfuss: Not sure how 'help' got into that sentence.

designs/resource_metadata_overriding.md
jfuss: This will not be injected. The Method is static. I wanted to keep this out of the `SamBaseProvider` class, so it can make the decision on when/where to run this in the flow of getting the `template_dict`. It is possible to have more than one normalizer but they could happen at different times. For now, I elected to keep it simple and just make the method call to this class where we needed.

tests/unit/lib/samlib/test_resource_metadata_normalizer.py
```diff
@@ -0,0 +1,138 @@
...
+                        "Code": "some value"
+                    },
+                    "Metadata": {
+                        "aws:asset:path": "new path",
+                    }
+                }
+            }
+        }
+
+        ResourceMetadataNormalizer.normalize(template_data)
```
eladb: Shouldn’t that be an error? Or at least a warning ⚠️?

jfuss: This is done on the full template, which means a resource not being invoked could fail the invoke of another and don't think that is the correct UX. Maybe there are cases like `start-api` or `start-lambda` that could require this to be a failure but the CLI doesn't have any distinctions about which command is run. 

A warning seems reasonable.

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -0,0 +1,63 @@
...
+        resource dict
+            Dictionary representing the Resource to change
+        logical_id str
+            LogicalId of the Resource
+
+        """
+        if property_key and property_value:
+            resource.get(PROPERTIES_KEY, {})[property_key] = property_value
+        elif property_key or property_value:
+            LOG.info("WARNING: Ignoring Metadata for Resource %s. Metadata contains only aws:asset:path or "
```
eladb: `LOG.warning`?

jfuss: That will not get logged to console by default (due to how we currently have logging setup) and the pattern we have followed thus far is using `LOG.info` instead. There is a bigger story here on our logging (including adding colors to make reading easier) but that is for another day.

eladb: Sure no prob

TheSriram:
I dont see any major concerns, looks good to me. 👍

PR #924, opened 2019-01-10
chore: Use OrderedDict in yamlhelper to preserve template elements order

paoptu023:
*Description of changes:*
Currently elements in templates are scrambled after parsing and dumping, use OrderedDict to preserve elements order.

### DIFF
--- a/samcli/yamlhelper.py
+++ b/samcli/yamlhelper.py
@@ -3,6 +3,7 @@ Helper to be able to parse/dump YAML files
 """
 
 import json
+from collections import OrderedDict
 import six
 
 import yaml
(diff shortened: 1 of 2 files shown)


paoptu023:
The [Python3.7 Travis build](https://travis-ci.org/awslabs/aws-sam-cli/jobs/477614273) failed at `tests/integration/local/invoke/runtimes/test_with_runtime_zips.py:47`, but the test could run on my local.

samcli/yamlhelper.py
```diff
@@ -46,22 +47,34 @@ def intrinsics_multi_constructor(loader, tag_prefix, node):
...
     Dumps the dictionary as a YAML document
     :param dict_to_dump:
     :return:
     """
+    yaml.SafeDumper.add_representer(OrderedDict, _dict_representer)
     return yaml.safe_dump(dict_to_dump, default_flow_style=False)
 
 
+def _dict_constructor(loader, node):
+    # Necessary in order to make yaml merge tags work
```
paoptu023: According to [CloudFormation template formats](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-formats.html), aliases and hash merges are not supported, but I think it's still useful to be able to parse the syntax.

jfuss: We cannot support this at the moment. The package command (and a similar yamlhelper) lives within AWS CLI. Supporting this here means, customers can write this in template, test locally, but fail during deployments to AWS CloudFormation.

paoptu023: I have a similar [PR](https://github.com/aws/aws-cli/pull/3841) opened to aws-cli. The original yamlhelper is able to parse the merge syntax, and the `package` command could run without any issues, but the template would fail `cloudformation validate-template`. 

`loader.flatten_mapping(node)` is only needed for using customized dict constructor.    I don't feel strongly to keep it, but I want to make sure I don't break any existing functionalities.

jfuss: I see. This is to have support for what was perviously there. The initial comment made it sound like you were proposing we add this support.

paoptu023:
Close as duplicate of #983.

PR #942, opened 2019-01-16
fix: add --template-file into template common options

paoptu023:
*Description of changes:*
`package` and `deploy` commands use `--template-file` option to specify templates, but other commands like `validate` and `publish` use `--template|-t`. To make it consistent, we should add `--template-file` as an option for these commands.  

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -74,7 +74,7 @@ def template_click_option(include_build=True):
     """
     Click Option for template option
     """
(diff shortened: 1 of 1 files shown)


jfuss:
@paoptu023 Thanks for the PR but SAM CLI preference is to use `--template`. We have to support `--template-file` due to AWS CLI's naming. This will eventually be replaced with with `--template` when we remove the AWS CLI dependency. 

Closing this PR, since the current options are correct.

Issue #948, opened 2019-01-19
`sam local invoke` command does not support `!Sub` function keyword in `Layers` settings


goldeneggg:
### Description

There is a example how to use `Layers` in [serverless\-application\-model/2016\-10\-31\.md](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#referencing-lambda-version--alias-resources)

![2019-01-20 0 32 01](https://user-images.githubusercontent.com/994282/51428757-d51b3b00-1c4a-11e9-8144-15b9b21a28b0.png)

But it does not work at local.

### Steps to reproduce

The `template.yaml` below

- Use `!Sub` function and "${AWS::Region}" parameter in `Layers` section
- Layer function "arn:aws:lambda:${AWS::Region}:123456789012:layer:MyLayer:1" is already deployed

```yml
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello_world/
      Handler: app.lambdaHandler
      Runtime: nodejs8.10
      Events:
        HelloWorld:
          Type: Api
          Properties:
            Path: /hello
            Method: post
      Layers:
        - !Sub arn:aws:lambda:${AWS::Region}:123456789012:layer:MyLayer:1
```

Run `sam local invoke` and get an error as follows

```sh
$ sam local invoke -e event.json HelloWorldFunction

Unable to import module 'app': Error  # This is an error about module include of Node.js
```

On the other hand, the `template.yaml` below works correctly at local.

- __Don't__ use `!Sub` function
  - use literal string value. (e.g. "ap-northeast-1")

```yml
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello_world/
      Handler: app.lambdaHandler
      Runtime: nodejs8.10
      Events:
        HelloWorld:
          Type: Api
          Properties:
            Path: /hello
            Method: post
      Layers:
        - arn:aws:lambda:ap-northeast-1:123456789012:layer:MyLayer:1
```
```sh
$ sam local invoke -e event.json HelloWorldFunction

Downloading arn:aws:lambda:ap-northeast-1:123456789012:layer:MyLayer:1  [####################################]  7298732/7298732

{"statusCode":200,"body": ...
```

### Why?
[...]

PR #949, opened 2019-01-19
fix(layers): show warning if `!Sub` function is used in `Layers` section of template.yaml (#948)

goldeneggg:
https://github.com/awslabs/aws-sam-cli/issues/948

*Description of changes:*

If `!Sub` function is used in `Layers` section of template.yaml and run `sam local invoke` and get an error.

- got: Error reason is not clear
- expected: We want to know more information by log

```yml
Resources:
  HelloWorldFunction:
    Properties:
      CodeUri: hello_world/
      Handler: app.lambdaHandler
      Runtime: nodejs8.10
      Events:
        HelloWorld:
          Type: Api
          Properties:
            Path: /hello
            Method: post
      Layers:
        - !Sub arn:aws:lambda:${AWS::Region}:123456789012:layer:MyLayer:1
```
```sh
$ sam local invoke -e event.json HelloWorldFunction

# This warning log is added by this pull request
2019-01-20 00:56:43 '!Sub' keyword in Layer configuration does not work locally. Please replace all parameter strings(like '${AWS::Region}') with the original value(like 'ap-northeast-1') and remove '!Sub' keyword.

Unable to import module 'app': Error
```

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/sam_function_provider.py
+++ b/samcli/commands/local/lib/sam_function_provider.py
@@ -239,8 +239,9 @@ class SamFunctionProvider(FunctionProvider):
         layers = []
         for layer in list_of_layers:
             if isinstance(layer, dict) and layer.get("Fn::Sub"):
-                LOG.warning("'!Sub' keyword in Layer configuration does not work locally. "
-                            "Please replace all parameter strings(like '${AWS::Region}') with the original value(like 'ap-northeast-1') and remove '!Sub' keyword.")
+                LOG.warning("'!Sub' keyword in Layer configuration does not work at local. "
+                            "Please replace all parameter strings(like '${AWS::Region}') with the original value "
+                            "and remove '!Sub' keyword from template.yaml .")
 
             # If the layer is a string, assume it is the arn
             if isinstance(layer, six.string_types):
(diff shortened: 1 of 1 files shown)


mnapoli:
Stumbled on that as well, until such feature is supported it would be awesome to have a clear warning 👍

jfuss:
While this impacts Layers (probably disproportionally), I don't believe this is the correct place for adding this detail. SAM CLI (as a whole) does not have support for Sub (and many other intrinsics). Due to this, we shouldn't be trying to read these intrinsics within the higher provider classes. This means, for ever resource and property, we need to add this kind of logging. While it is contextually, it creates more work later on. Instead, I would be ok with adding this logging more broadly to communicate the known Intrinsic gaps. [Maybe this is a good place](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/local/lib/sam_base_provider.py#L41)?

jfuss:
@goldeneggg and @mnapoli We are addressing intrinsics in #1261. Once that is merged, integrated, and released, there shouldn't be any more intrinsic confusion on support. I am going to close this in favor of the full solution we are currently going through.

PR #953, opened 2019-01-22
Add support for prebuilt layer archives and function resources

### LINKED ISSUE #947: Layer not working locally due to not supporting zip file
**Description:**

I have a lambda layer which takes care of common code and node_modules, which works perfectly fine when deployed. But for local development its not working. Not sure if, I am doing something wrong or this feature is not supported yet. 

Below is `dir` structure

```
-src/
--test
---index.js
--shared
---nodejs
----package.json
----response.js       
-template.yaml
-.gitignore
```

template.yaml
```
Transform: 'AWS::Serverless-2016-10-31'

Globals:
  Function:
    Runtime: nodejs8.10

Parameters:
  StageName:
    Type: String

Resources:
  TestApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: dev

  #Lambda Layer for sharing code
  TestLambdaLayer:
    Properties:
      LayerName: test-layer
      Description: Lambda layer with all node module dependency and shared code between lambda
      ContentUri: src/shared/nodejs.zip
      CompatibleRuntimes:
        - nodejs8.10
      RetentionPolicy: Retain

  Test:
    Type: 'AWS::Serverless::Function'
    Properties:
      Layers:
        - !Ref TestLambdaLayer
      FunctionName: test
      CodeUri: src/test
      Handler: index.test
      Events:
        EndPoint:
          Type: Api
          Properties:
            RestApiId: !Ref TestApi
            Path: /test
            Method: get
```
test api
```
const sharedPath = process.env.LAYER_PATH || '/opt/nodejs/';

const boom = require('boom');

const { successResponse, errorResponse } = require(sharedPath + 'response');

async function test(event) {
  const data = {
    buildNumber: 1.0
  };
  try {
    return successResponse(data);
  } catch (error) {
    return errorResponse(boom.serverUnavailable());
  }
}

module.exports = {
  test
};

```

when I run `sam local start-api` i get `Unable to import module 'index': Error` . When I put node_module and shared code inside `test`  folder path everything works as expected locally. 

How can I make lambda layer work locally without changing the code? Thanks for the help. 

P.S: Also if you have some time I have another issue open on StackOverflow
https://stackoverflow.com/questions/54261625/api-gateway-options-method-throwing-403/54262046#54262046

brandond:
*Issue #, if available:*
Fixes #947

*Description of changes:*
Adds support for pre-archived Lambda Layers and Functions. The current code does not reliably support directories or zip archives - 'sam local invoke' supports function archives but not layer archives, while 'sam build' supports layer archives but not function archives.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -185,6 +185,10 @@ class ApplicationBuilder(object):
         # artifacts directory will be created by the builder
         artifacts_dir = str(pathlib.Path(self._build_dir, function_name))
 
+        if code_dir.lower().endswith(".zip"):
+           LOG.debug("Serverless function '%s' is already built", function_name)
+           return code_dir
+
         with osutils.mkdir_temp() as scratch_dir:
             manifest_path = self._manifest_path_override or os.path.join(code_dir, config.manifest_name)
 
--- a/samcli/local/layers/layer_downloader.py
+++ b/samcli/local/layers/layer_downloader.py
@@ -8,7 +8,7 @@ import boto3
 from botocore.exceptions import NoCredentialsError, ClientError
 
 from samcli.lib.utils.codeuri import resolve_code_path
-from samcli.local.lambdafn.zip import unzip_from_uri
+from samcli.local.lambdafn.zip import unzip, unzip_from_uri
 from samcli.commands.local.cli_common.user_exceptions import CredentialsRequired, ResourceNotFound
 
 try:
@@ -92,6 +92,13 @@ class LayerDownloader(object):
         if layer.is_defined_within_template:
             LOG.info("%s is a local Layer in the template", layer.name)
             layer.codeuri = resolve_code_path(self.cwd, layer.codeuri)
+
+            if layer.codeuri.lower().endswith('.zip'):
+                LOG.info("Decompressing %s", layer.codeuri)
+                layer_zip_path = layer.codeuri
+                layer.codeuri = layer.codeuri[:-4]
+                unzip(layer_zip_path, layer.codeuri)
+
             return layer
 
         # disabling no-member due to https://github.com/PyCQA/pylint/issues/1660
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -54,7 +54,7 @@ class BuildContext(object):
         except ValueError as ex:
             raise UserException(str(ex))
 
-        self._function_provider = SamFunctionProvider(self._template_dict, self._parameter_overrides)
+        self._function_provider = SamFunctionProvider(self._template_dict, self._parameter_overrides, self._template_file)
 
         if not self._base_dir:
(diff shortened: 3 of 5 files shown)


jfuss:
@brandond I can see pre-archived Lambda Layers being something to add but not sure what the purpose of building a function that is already zipped. Can you explain why you would be trying to build a code that is zipped? 

Note: `sam build` does not support building layers yet.

samcli/local/layers/layer_downloader.py
```diff
@@ -92,6 +92,13 @@ def download(self, layer, force=False):
         if layer.is_defined_within_template:
             LOG.info("%s is a local Layer in the template", layer.name)
             layer.codeuri = resolve_code_path(self.cwd, layer.codeuri)
+
+            if layer.codeuri.lower().endswith('.zip'):
```
jfuss: We should follow the same thing we do for functions here. Can you abstract [this method out](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/lambdafn/runtime.py#L152) to be usable here and in the runtime?

samcli/lib/build/app_builder.py
```diff
@@ -185,6 +185,10 @@ def _build_function(self, function_name, codeuri, runtime):
         # artifacts directory will be created by the builder
         artifacts_dir = str(pathlib.Path(self._build_dir, function_name))
 
+        if code_dir.lower().endswith(".zip"):
```
jfuss: Ahhh... I see. It's not building rather you want to skip functions that are zipped. 

This seems reasonable to me. The same comment I made on the layer_downloader file is also relevant here. We should skip any of the 'pre' built artifacts. I wonder if it is worth saying if `code_dir` is not a directory then skip? Thoughts?

brandond: Yeah, skip zipping on 'build', and extract out for 'local invoke'. There's zip handling scattered around about three different places; I can take a shot at consolidating this if it'd be well received.

jfuss:
@brandond Let's keep this PR scoped. It feels like this is starting to grow, which creates large PRs and harder to track the actual changes. 

Happy to discuss the S3 Uri use-case in the #samdev slack channel as well. This will help in understanding the is needed before doing the work and discussing in a PR

brandond:
Can do @jfuss . My ultimate goal is to be able to use the SAM CLI to test lambda functions and APIs that are prebuilt by an existing CI pipeline. That includes merging multiple template fragments, building layer and function archives, etc. I get that a lot of that may eventually be replaceable by functionality that SAM CLI will provide via build/package/deploy, but until those mature a good bit my projects are still pretty well tied to what's working now.

jfuss:
Closing this as it is out dated and hasn't been updated since Jan. Feel free to re-open or send another PR

PR #957, opened 2019-01-24
docs: Update `sam validate` command design doc

sthulb:
*Issue #, if available:*
#993

*Description of changes:*

*Checklist:*

- [x] Write Design Document
- [ ] Build the command line interface
- [ ] Build the underlying library
- [ ] Unit tests
- [ ] Integration tests
- [ ] Run all tests on Windows
- [ ] Update documentation

### DIFF
--- a/designs/sam_validate.md
+++ b/designs/sam_validate.md
@@ -33,13 +33,43 @@ The user will be able to find detailed information about their template's valida
 - Run `sam validate -t ./template.yaml`
 - Make changes
 - If errors:
+
   - Show problems associated with the template
+
 - If ok:
+
   - Awesome, carry on developing
+
+Any issues identified will be printed to the user's console for them to act upon. The issues will be printed in the same fashion as `cfn-python-lint` already prints them. Using the existing printing mechanism would allow the user to use `sam validate` in place of the `cfn-lint` command in editors.
 
 ## Implementation
 
 The improved functionality of `sam validate` will utilise the opensource `cfn-python-lint` package (https://github.com/awslabs/cfn-python-lint/) made by AWS. The changes will enhance the existing `sam validate` functionality.
+
+`cfn-python-lint` has an optional dotfile per directory (`.cfnlintrc`) which allows users to specify rules and config of the tool. This will be implmented as a part of the changes proposed by this document.
+
+The defaults for the tool will remain the same, whatever options `cfn-python-lint` sets by default will be used regardless of versional changes, it will be accepted that the default options are in everyone's best interest.
+
+### Example
+
+The following template has an error in the S3 Bucket resource, the `NotReal` key is erroneous:
+
+```yaml
+AWSTemplateFormatVersion: "2010-09-09"
+Transform: "AWS::Serverless-2016-10-31"
+Resources:
+  Bucket:
+    Type: "AWS::S3::Bucket"
+    Properties:
+      NotReal: "name"
+```
+
+Expected Output would be:
+
+```
+E3002 Invalid Property Resources/Bucket/Properties/NotReal
+workdir/invalid_template.yaml:7:7
+```
 
 ### CLI Changes
 
--- a/designs/sam_validate.md
+++ b/designs/sam_validate.md
@@ -1,0 +1,40 @@
+# `sam validate` command
+
+This is the design to improve the `sam validate` command to utilise the the `cfn-python-lint` package.
+
+It would be good to go beyond just implementing the CloudFormation package for validation of template files, there
+is the opportunity to create something better.
+
+## What's the current problem?
+
+The current iteration of `sam validate` only goes so far in validating the the SAM template, it does some cursorary checks:
+
+- Checks `CodeUri` to ensure an S3 URL exists
+- Checks if the template is valid but doesn't return errors
+  - If the template contains other resources outside of SAM it won't validate them
+
+## What will be changed?
+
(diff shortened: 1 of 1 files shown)


designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
+# `sam validate` command
+
+This is the design to improve the `sam validate` command to utilise the the `cfn-python-lint` package.
```
jfuss: Since we don't yet have a design doc for validate, what if we just say this is the design for the command? We started a pattern with creating these with `sam build` and did the same for `sam publish`.

designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
...
+
+The current iteration of `sam validate` only goes so far in validating the the SAM template, it does some cursorary checks:
+
+- Checks `CodeUri` to ensure an S3 URL exists
+- Checks if the template is valid but doesn't return errors
+  - If the template contains other resources outside of SAM it won't validate them
+
+## What will be changed?
+
+We will change how the `sam validate` command works, by integrating the [cfn-lint python package](https://github.com/awslabs/cfn-python-lint/) into the the validation workflow. This change would enhance the user's experience of working with SAM templates.
```
jfuss: Not only working with SAM Templates but any CloudFormation templates :)

designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
...
+The user will be able to find detailed information about their template's validation errors or warnings by running `sam validate`. The validate command will also find any issues in non-SAM resources, such as S3 buckets or SNS topics.
+
+## Out of Scope
+
+- Customisation of the rules through the use of flags
+- Ignoring checks
+- Appending rules
+- Specifying your own validation spec
+
+## User Experience Walkthrough
```
jfuss: In this section, it would be useful to capture how we will output data back to the customer. You can have two sections, one for a valid template and one for an invalid one. It will help in understanding the final UX we are targeting.

designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
...
+- Run `sam validate -t ./template.yaml`
+- Make changes
+- If errors:
+  - Show problems associated with the template
+- If ok:
+  - Awesome, carry on developing
+
+## Implementation
+
+The improved functionality of `sam validate` will utilise the opensource `cfn-python-lint` package (https://github.com/awslabs/cfn-python-lint/) made by AWS. The changes will enhance the existing `sam validate` functionality.
```
jfuss: We should have more details here. Snippets of code or classes could be helpful. Do we need to change any of the template to use [cfn-lint](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/validate/lib/sam_template_validator.py#L134)? Maybe a small example of how you will configure cfn-lint? Will we need to fetch any data (currently we get [managed policies](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/validate/lib/sam_template_validator.py#L54))? Should or do we need to configure a logger for the library to print correct debug or warning information?

designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
...
+
+We will change how the `sam validate` command works, by integrating the [cfn-lint python package](https://github.com/awslabs/cfn-python-lint/) into the the validation workflow. This change would enhance the user's experience of working with SAM templates.
+
+## Success criteria for the change
+
+The user will be able to find detailed information about their template's validation errors or warnings by running `sam validate`. The validate command will also find any issues in non-SAM resources, such as S3 buckets or SNS topics.
+
+## Out of Scope
+
+- Customisation of the rules through the use of flags
```
jfuss: In the implementation section, it would be good to call out any default flags we will configure. We will probably want documentation for what they are as well.

designs/sam_validate.md
```diff
@@ -0,0 +1,58 @@
...
+- Ignoring checks
```
jfuss: Does cfn-lint have any `.cnflintrc` or something similar? Would make this easier to be configurable without needing to expand the cli definition in the future.

Nr18: It does: https://github.com/awslabs/cfn-python-lint#config-file

designs/sam_validate.md
sthulb: That was the point of this PR, implementation will be in another PR :)

designs/sam_validate.md
sthulb: Sure, but I would have assumed they'd use cfn-lint directly if it wasn't a serverless project

designs/sam_validate.md
sthulb: As I mentioned on Slack, I'm not going out of my way to block this – if it comes along for free I'll leave it 😀

Judging by what I've seen by reading through the source, it should do

designs/sam_validate.md
jfuss: I know implementation will come later. I was more referring to making this the design for the command. We don't yet have one.

So it this wouldn't be a design to improve but would just be the design for.

designs/sam_validate.md
jfuss: Adding details here.

I still don't quiet understand the changes here. It would be really useful to understand with a snippet or example on how to actually integrate with cfn-lint. The [`sam publish` design](https://github.com/awslabs/aws-sam-cli/blob/develop/designs/sam_publish_cmd.rst#design) is a good example which will help us understand a little more in-depth of what the changes will be. This is especially useful what people reading the design later have little background on the library.

jfuss:
Cleaning up PRs that have stalled. Closing this PR but feel free to reopen or submit a new one.

Issue #968, opened 2019-01-29
SAM Build support for Go with modules


jfuss:
### Describe your idea/feature/enhancement

Support building go functions that use Go Modules.

eikeon:
I'm assuming this is relevant: awslabs/aws-lambda-builders#64

aripalo:
If I've understood correctly, the lambda builders project now has support for Go modules via this PR https://github.com/awslabs/aws-lambda-builders/pull/65 that is already merged.

There seems to be several different PRs around this project about adding Golang support, but they all seem to be closed. Am I missing something or has the golang build support been left out on purpose or is it still under development?

I'm currently in midst of writing a massive Makefile setup for my projects (they have Go 1.13, Node 10 and Python 3.7 Lambdas) that tackle those sort of "polyglot lambda monorepos" with Lambci Lambda dockers, but I'd rather use just `sam build --use-container`, but so far missing Golang support prevents going down that path. 

Wondering if there's any areas where I could help/contribute regarding the Golang (with Go modules support). I don't want to just shoot in the dark and open a random PR if there's already some active development going on somewhere 🙂

jfuss:
Hey @aripalo. I am not aware of anything in progress for Go Module support within SAM CLI and PRs are always welcome. There is a closed PR in the repo (from me) that tried to get this working some time ago but has feel off my plate for other things. That could be a good place to start.

One thing to note is, `--use-container` is a challenge to support. The main reason for that is Go likes to write to the host (and in this case container). We mount code in a readonly mode (for security reasons so things do not get written back to the host). The initially support we were planning was to disallow `--use-container` (like we do for dotnet for similar reasons). We can tar up the function directory and set that to the container but that is substantially more work. With Go being able to cross compile, most cases will never need a container.

Happy to talk this through further, if you are interested in picking this up.

yordis:
@jfuss any updates on adding support for this?

awood45:
#1731 adds this.

aripalo:
Yes yes! This is great! 🎉 I can finally start to teardown my weird hacks and use SAM CLI with Go! 

![awesome](https://media.tenor.com/images/f4c8059e75d21aa301174d4374ec4680/tenor.gif)

PR #976, opened 2019-02-04
[WIP] feat(build): Support for Go Modules

jfuss:
*Issue #, if available:*
#968

*Description of changes:*
This will add support for Go Modules.

*Issues need resolving*
[ ] Builder Workflow in docker does not work from SAM CLI. This is due to mounting with `ro` and go wanting to write a `go.sum` file.
```
(aws-sam-cli) ➜  sam-app git:(build/gomod) ✗ samdev build --use-container --skip-pull-image
2019-02-03 15:45:21 Starting Build inside a container
2019-02-03 15:45:22 Building resource 'HelloWorldFunction'
2019-02-03 15:45:22 Requested to skip pulling images ...

2019-02-03 15:45:22 Mounting /Users/jfuss/sam-github/aws-sam-cli/sam-app/hello-world as /tmp/samcli/source:ro inside runtime container
Build Failed
Running GoModulesBuilder:Build
Error: GoModulesBuilder:Build - Builder Failed: go: finding github.com/aws/aws-lambda-go v1.8.0
go: downloading github.com/aws/aws-lambda-go v1.8.0
go: writing go.sum: open /tmp/samcli/source/go.sum: read-only file system
```

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -193,20 +163,44 @@ def _build_function(self, function_name, codeuri, runtime):
...
+        handler str
+            Handler value of the Lambda Function Resource
+        Returns
+        -------
+        dict
+            Dictionary that represents the options to pass to the builder workflow or None if options are not needed
+        """
+        options = None
+        if language == 'go':
+            options = {'handler': handler}
```
sanathkr: hang on. Instead of handler, can we call this option the name of executable? As much as possible, I want to avoid builder specific options. Would this be generic enough? Thoughts?

samcli/lib/build/workflow_config.py
```diff
@@ -0,0 +1,67 @@
...
+    runtime str
+        The runtime of the config
+
+    Returns
+    -------
+    namedtuple(Capability)
+        namedtuple that represents the Builder Workflow Config
+    """
+
+    workflow_config_by_runtime = {
```
sanathkr: better!

samcli/lib/build/workflow_config.py
```diff
@@ -0,0 +1,67 @@
...
+        "go1.x": GO_MOD_CONFIG,
+        "python2.7": PYTHON_PIP_CONFIG,
+        "python3.6": PYTHON_PIP_CONFIG,
+        "python3.7": PYTHON_PIP_CONFIG,
+        "nodejs6.10": NODEJS_NPM_CONFIG,
```
sanathkr: node4.x?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -140,10 +133,8 @@ class TestBuildCommand_NodeFunctions(BuildIntegBase):
     FUNCTION_LOGICAL_ID = "Function"
 
     @parameterized.expand([
-        ("nodejs4.3", False),
```
sanathkr: Why are we removing these?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -266,8 +251,61 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files,
...
+
+    @parameterized.expand([
+        ("go1.x", False),
+        ("go1.x", "use_container")
+    ])
+    def test_with_default_go_mod(self, runtime, use_container):
+        overrides = {"Runtime": runtime, "CodeUri": "Go"}
+        cmdlist = self.get_command_list(use_container=use_container,
+                                        parameter_overrides=overrides,
+                                        # TODO remove this once aws-lambda-builders is released
```
sanathkr: So this is a problem I kept running into. Even if you skip pull images, this code won't build on Travis because there are no containers there. Instead, I would just turn off Container builds in the here until the Docker container is released

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -266,8 +251,61 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files,
...
+                                        skip_pull_image=True)
+
+        LOG.info("Running Command: {}".format(cmdlist))
+        process = subprocess.Popen(cmdlist, cwd=self.working_dir)
+        process.wait()
+
+        self._verify_built_artifact(self.default_build_dir, self.FUNCTION_LOGICAL_ID,
+                                    self.EXPECTED_FILES_PROJECT_MANIFEST)
+
+        self._verify_resource_property(str(self.built_template),
```
sanathkr: I don't know if we need to repeat this test for every language.

tests/integration/testdata/buildcmd/Go/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/aws/aws-lambda-go v1.8.0 h1:YMCzi9FP7MNVVj9AkGpYyaqh/mvFOjhqiDtnNlWtKTg=
```
sanathkr: shouldn't you .gitignore this file?

samcli/lib/build/app_builder.py
jfuss: We can make this more generic (will take a PR to aws-lambda-builders). What would you like to call this? `executable_name`?

samcli/lib/build/workflow_config.py
jfuss: [4.x is depreciated by Lambda](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html). I can add it back but we need to figure out how to depreciate these when Lambda no longer supports them.

tests/integration/buildcmd/test_build_cmd.py
jfuss: This was for me locally. I wanted to remove it only after we got closer to be able to merge this but didn't want to manage/keep track of this locally.

Not sure we want to turn off this test. We run the risk of needing to do a release and making the build unavailable with the `--use-container` flag. I would prefer we either leave it until we release `aws-lambda-builders` and containers get updated or we only ignore this test when running travis in the develop branch and all tests run when merging to master.

tests/integration/buildcmd/test_build_cmd.py
jfuss: My opinion is yes. This is to help us in not breaking this for a given language. As build grows, we may need to make language specific logic branches, which could lead to the wrong this happening. It's safer to just check it every time, it really doesn't hurt anything and is part of the expected outcome.

tests/integration/testdata/buildcmd/Go/go.sum
jfuss: Probably. Will remove.

tests/integration/buildcmd/test_build_cmd.py
jfuss: The other thing is until `aws-lambda-builders` is released `sam build` for a go function won't work because the released `aws-lambda-builders` does not have the Go Modules builder.

samcli/lib/build/app_builder.py
sanathkr: Can we call this something like "artifact_binary_name" or something that explicitly says this is built artifacts?

sanathkr:
Can you rebase on develop and submit the PR again?

samcli/local/docker/container.py
```diff
@@ -127,15 +125,47 @@ def create(self):
         if self.network_id == 'host':
             kwargs["network_mode"] = self.network_id
 
+        if not self.should_put_archive():
```
jfuss: I am not super crazy about having to do a check here and another check after the container is created (line 138). I have to do this because if we need to put the archive we shouldn't be telling the container to mount it on create. There has to be a better way to make this cleaner but not coming up with anything. @sanathkr or @TheSriram: Would love to hear if you have thoughts on how to clean this part up.

samcli/lib/build/app_builder.py
```diff
@@ -182,20 +183,44 @@ def _build_function(self, function_name, codeuri, runtime):
...
+        language str
+            Language of the runtime
+        handler str
+            Handler value of the Lambda Function Resource
+        Returns
+        -------
+        dict
+            Dictionary that represents the options to pass to the builder workflow or None if options are not needed
+        """
+        options = None
```
TheSriram: ``
{'artifact_executable_name': handler} if language=='go' else None
``

could be simpler

samcli/local/docker/container.py
TheSriram: copy is

```
self.docker_client.containers.create(self._image, **kwargs)
self._put_archives(docker_volumes, real_container)
```

mount is

```
kwargs["volumes"] = {to_posix_path(host_dir): mount for host_dir, mount in docker_volumes.items()}
self.docker_client.containers.create(self._image, **kwargs)
```

out of which below is common

```
self.docker_client.containers.create(self._image, **kwargs)
```

can we hide the order in which we call (mount,copy) through a wrapped decorator, where the decorator arguments determine which way to wrap the function? 

```
@files_relay(type={assigned based on some other logic eg: "copy" if go else "mount"}, volumes=docker_volumes)
def start_container(self._image, **kwargs):
   self.docker_client.containers.create(self._image, **kwargs)
```

```
def files_relay(type, volumes):
    def wrap(f):
        def wrapped_f(*args, **kwargs):
           if type == 'copy':
              real_container = f(*args,**kwargs)
              self._put_archives(volumes, real_container)
          else:
             kwargs["volumes"] = {to_posix_path(host_dir): mount for host_dir, mount in docker_volumes.items()}
             f(*args,**kwargs)
        return wrapped_f
    return wrap
```

then we just call 

```
start_container(self._image, **kwargs)
```

Notice how we are calling and changing the order of 'f' in the above, Does something like that work?

hgl:
Is this PR abandoned? would like to test lambda in go locally, seems not possible right now?

jfuss:
@hgl I am working on splitting this out into two prs, which will make getting Go modules build support easier to get through.

This PR is for build. You can already locally invoke Go functions, you just have to build the binary yourself.

Closing in favor of a new PR that I am working on

ragurney:
@jfuss Sorry for the noobish question, but what would building manually entail? Once I have the binary, do I move it to `.aws_sam/build` along with the template?

PR #983, opened 2019-02-06
chore: Use OrderedDict in yamlhelper to preserve template elements order

shwetaskatdare:
*Description of changes:*
Currently elements in templates are scrambled after parsing and dumping, use OrderedDict to preserve elements order.

Updated unit test case to match aws-cli

Note:
Original PR: https://github.com/awslabs/aws-sam-cli/pull/924
Wasn't able to update the original PR as I don't have permissions. Since the submitter of that PR is away on vacation, and we want to get this addressed as soon as possible, I am creating a new PR with the changes pulled in from the original PR. Comments have been addressed.

### DIFF
--- a/samcli/yamlhelper.py
+++ b/samcli/yamlhelper.py
@@ -3,6 +3,7 @@ Helper to be able to parse/dump YAML files
 """
 
 import json
+from collections import OrderedDict
 import six
 
 import yaml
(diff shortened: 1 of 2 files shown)


jfuss:
@sanathkr Can you take a look at this as well?

samcli/yamlhelper.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import json
+from collections import OrderedDict
```
sanathkr: I am not sure if this will be a drop-in replacement. There could be code somewhere that assumes this is a pure dict `val isinstance dict` that might break because of this change. 

Also, in Py3.6 dicts are already in insertion order (similar to OrderedDict). While this is a problem for Py2.7 users, Py27 is dying very very soon and we should probably be migrating to Py36+. I am not sure if this worth the change now.

Do you have specific use cases that won't work without this change?

shwetaskatdare: I cannot think of any special usecases.
@paoptu023  What are your thoughts? 

Also, @sanathkr when are you planning to migrate to Py36+?

paoptu023: I'm ok with not having this change if we will drop Py2.7 support soon. Otherwise, since we've already made the same change in SAR and aws cli, it might cause confusion due to the inconsistent behavior.

shwetaskatdare:
Closing this as no longer needed. The change will be taken care of when upgrade to Py3.6+ is done as per above conversation.

PR #987, opened 2019-02-07
feat: allow sam to be invoked as a module

TheSriram:
- `python -m samcli.cli.main` invokes sam cli

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/main.py
+++ b/samcli/cli/main.py
@@ -4,6 +4,7 @@ Entry point for the CLI
 
 import logging
 import json
+import sys
 import click
 
 from samcli import __version__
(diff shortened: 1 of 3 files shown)


samcli/cli/main.py
```diff
@@ -4,6 +4,7 @@
```
sanathkr: Can you instead add a __main__.py? `python -m ..` will call the __main__.py file - https://docs.python.org/3/library/__main__.html

TheSriram: Moved, this was a good call out.

TheSriram:
rebasing with develop, hence the force push.

PR #989, opened 2019-02-08
fix: Include dotnet build artifacts from possible subfolders.

apehkone:
*Description of changes:*

Pack task doesn't include possible runtime specific nuget packages which are located in the "runtimes" folder of the build directory. This will cause an error such as "An assembly specified in the application dependencies manifest (application.deps.json) was not found" when certain nuget packages are referenced. 


TheSriram:
There is a .NET builder in the works #1039 , does that look like it includes this use case?

apehkone:
I believe it does. It wasn't really a builder issue since it was related to the pack task and how it ignored subfolders from the publish directory.

sanathkr:
Closing this PR because `sam build` for dotnet is merged and should solve the problem

PR #997, opened 2019-02-09
feat: add sam local get-env-vars command

jlhood:
*Issue #, if available:*

*Description of changes:*

Proposing a new sam local subcommand. Not sure if anyone else has already proposed it, but wrote up a quick design doc for it to get the conversation started.

*Checklist:*

- [x] Write Design Document
- [x] Write Prototype
- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,10 @@ func-test:
 	@echo Telemetry Status: $(SAM_CLI_TELEMETRY)
 	pytest --cov samcli.local --cov samcli.commands.local --cov-report term-missing tests/functional
 
+smoke-test:
+	# Smoke tests run in parallel
+	pytest -n 4 tests/functional
+
 flake:
 	# Make sure code conforms to PEP8 standards
 	flake8 samcli
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,11 @@ environment:
       PYTHON_VERSION: '3.7.4'
       PYTHON_ARCH: '64'
 
+    - ONLY_SMOKE: 1
+      PYTHON_HOME: "C:\\Python37-x64"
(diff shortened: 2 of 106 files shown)


jasonmk:
Would this solve the issue of an ENV var that Refs another resource created by the template?

E.g., I ran into an issue today where I want to test a lambda function locally, but one of the things it needs is the arn of a secret stored in Secrets Manager.  My template has `SECRET_NAME: !Ref MySecret` as an environment variable on the function, but obviously it doesn't get populated in local mode.

If this addresses that issue, I would be a big fan of it.

jlhood:
@jasonmk Yes, since this command reads an already deployed stack, all of those references have been resolved so the JSON that the command outputs would be resolved values.

samcli/commands/local/get_env_vars/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    LOG.debug("'get-env-vars' command is called")
+
+    cfn_client = boto3.client('cloudformation')
+    lambda_client = boto3.client('lambda')
+
+    stack_resource_iterator = cfn_client.get_paginator('list_stack_resources').paginate(
+        StackName=stack_name
+    )
+
+    function_resource_summaries = [summary for response in stack_resource_iterator
```
TheSriram: do we need to consider AWS::Serverless::Function?

jlhood: No. This is a query on an actual deployed CloudFormation stack. `AWS::Serverless::Function` is a pseudo-resource of SAM, but the actual deployed stack will only contain `AWS::Lambda::Function` resources.

Issue #880, opened 2018-12-21
Java support for gradle in SAM build


abrooksv:
### Describe your idea/feature/enhancement
SAM build needs support for Java

### Additional Details
Feature request in toolkit: https://github.com/aws/aws-toolkit-jetbrains/issues/712

MichaelDz6:
Why is java8 checkbox checked on the main page if it isn't supported yet?

jfuss:
Scoping this issue to support for gradle. Will create one for maven explicitly.

sanathkr:
Blocked on https://github.com/awslabs/aws-sam-cli/pull/976/files because it refactors lots of the code paths necessary to support Gradle

jfuss:
This was released with [v0.12.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.12.0)

Closing

PR #1007, opened 2019-02-13
Support for gradle builder for Java

sanathkr:
*Issue #, if available:*
#880

*Description of changes:*
This won't build until we release aws-lambda-builders library. But sending the WIP PR to get early feedback.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


requirements/base.txt
```diff
@@ -12,5 +12,5 @@ dateparser~=0.7
 python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
 requests==2.20.1
-aws_lambda_builders==0.0.5
+aws_lambda_builders==0.1.0
```
jfuss: 0.1.0 🕺

jfuss:
Don't forget the java8 init templates will need updating from maven to gradle

sanathkr:
> Don't forget the java8 init templates will need updating from maven to gradle

It doesn't make sense. We will let the init templates use maven and eventually provide a maven builder. Yes, meanwhile sam init && sam build won't work, but that's a short term problem

TheSriram:
@sanathkr we need to change https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L206 to include `executable_search_paths`

samcli/lib/build/workflow_config.py
```diff
@@ -2,60 +2,158 @@
...
+        config = selector.get_config(code_dir, project_dir)
+        return config
+    except ValueError as ex:
+        raise UnsupportedRuntimeException("Unable to find a supported build workflow for runtime '{}'. Reason: {}"
+                                          .format(runtime, str(ex)))
+
+
+class BasicWorkflowSelector(object):
+    """
+    Basic workflow selector that returns the first available configuration in the given list of configurations
```
TheSriram: Is this going to be extended in the future? what is the value of this class, if its just returns the first one?

samcli/lib/build/workflow_config.py
```diff
@@ -2,60 +2,158 @@
...
+
+        # Search for manifest first in code directory and then in the project directory.
+        # Search order is important here because we want to prefer the manifest present within the code directory over
+        # a manifest present in project directory.
+        search_dirs = [code_dir, project_dir]
+        LOG.debug("Looking for a supported build workflow in following directories: %s", search_dirs)
+
+        for config in self.configs:
+
+            if any([self._has_manifest(config, directory) for directory in search_dirs]):
```
TheSriram: Like this 👍

samcli/commands/build/command.py
```diff
@@ -30,9 +30,10 @@
 \b
 Supported Runtimes
 ------------------
-1. Python2.7\n
-2. Python3.6\n
-3. Python3.7\n
+1. Python 2.7, 3.6, 3.7\n
+4. Nodejs 8.10, 6.10
+4. Ruby 2.5
+5. Java 8
```
jfuss: Do we need to call out which Dependency Managers we support?

samcli/local/docker/lambda_build_container.py
```diff
@@ -96,7 +98,8 @@ def _make_request(protocol_version,
                       manifest_file_name,
                       runtime,
                       optimizations,
-                      options):
+                      options,
```
TheSriram: Just making sure, the protocol version is passed in by importing the builders library right?

samcli/lib/build/app_builder.py
```diff
@@ -144,13 +144,14 @@ def update_template(self, template_dict, original_template_path, built_artifacts
         return template_dict
 
     def _build_function(self, function_name, codeuri, runtime):
-        config = get_workflow_config(runtime)
 
         # Create the arguments to pass to the builder
```
jfuss: This comment is floating. This seems like docstring worthy

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -252,3 +238,71 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files,
...
+    def test_with_gradle(self, runtime, code_path, use_container):
+        overrides = {"Runtime": runtime, "CodeUri": code_path, "Handler": "aws.example.Hello::myHandler"}
+        cmdlist = self.get_command_list(use_container=use_container,
+                                        parameter_overrides=overrides)
+
+        LOG.info("Running Command: {}".format(cmdlist))
+        process = subprocess.Popen(cmdlist, cwd=self.working_dir)
+        process.wait()
+
+        self._verify_built_artifact(self.default_build_dir, self.FUNCTION_LOGICAL_ID,
```
TheSriram: Can we look at built artifacts verify if we used `gradle` or `gradlew`?

samcli/lib/build/workflow_config.py
```diff
@@ -2,60 +2,158 @@
...
+        "java8": ManifestWorkflowSelector([
+            # Gradle builder needs custom executable paths to find `gradlew` binary
+            JAVA_GRADLE_CONFIG._replace(executable_search_paths=[code_dir, project_dir])
+        ]),
     }
 
-    try:
-        return workflow_config_by_runtime[runtime]
-    except KeyError:
+    if runtime not in selectors_by_runtime:
```
jfuss: It's much cleaner if you do

```
try:
  selector = selectors_by_runtime[runtime]
  config = selector.get_config(code_dir, project_dir)
except KeyError:
  #blah 
except ValueError:
  #another blah 
```

Also more pythonic :)

samcli/lib/build/workflow_config.py
```diff
@@ -2,60 +2,158 @@
...
+            if any([self._has_manifest(config, directory) for directory in search_dirs]):
+                return config
+
+        raise ValueError("None of the supported manifests '{}' were found in the following paths '{}'".format(
+                         [config.manifest_name for config in self.configs],
+                         search_dirs))
+
+    @staticmethod
+    def _has_manifest(config, directory):
+        return os.path.exists(os.path.join(directory, config.manifest_name))
```
jfuss: One day you will use pathlib.. one day..

samcli/lib/build/workflow_config.py
```diff
@@ -2,60 +2,158 @@
...
+        samcli.lib.build.workflow_config.CONFIG
+            A supported configuration if one is found
+
+        Raises
+        ------
+        ValueError
+            If none of the supported manifests files are found
+        """
+
+        # Search for manifest first in code directory and then in the project directory.
```
jfuss: How this works and is resolved needs to be clearly documented (somewhere)

samcli/commands/build/command.py
sanathkr: Hmm, yeah good idea. Will do

samcli/lib/build/app_builder.py
sanathkr: done

samcli/lib/build/workflow_config.py
sanathkr: For inheritance.. I didn't create an abstract base class because it didn't seem helpful. But in future we will create a new subclass that will look up Samrc to find the workflow config to use given a runtime. This can be useful for `provided` runtimes or if customer wants to override a default configuration

samcli/local/docker/lambda_build_container.py
sanathkr: Correct.

tests/integration/buildcmd/test_build_cmd.py
sanathkr: Ah good point. I don't know. I will check with Dongie

samcli/lib/build/workflow_config.py
sanathkr: Not really. try/catch blocks should be scoped very tightly to the code you expect to throw. So when you know two statemtents will throw for two different reasons, I would rather have them in separte try blocks or let them throw very specific exceptions - not KeyError & ValueError. That's why I made this way.

samcli/lib/build/workflow_config.py
sanathkr: when we move to py3. Pathlib in py2 is really confusing

samcli/lib/build/workflow_config.py
sanathkr: I will add to the help text for now

TheSriram:
I dont have any other concerns, will be nice if can determine which executable was used to build in the built artifacts. But I'm not blocking on that. will approve once build passes.

Issue #323, opened 2018-03-10
[Feature Request] CORS Support in start-api


walkerlangley:
Full disclosure, this is my first attempt at working with a SAM template so it's likely I'm doing something wrong, but I can't get CORS working for the life of me.  My functions work when I hit the endpoints using Insomnia, but when trying to hit it from my app, I get a 404 on the options request.  Also, when I run `sam local start-api`, I get a message saying `WARNING: Could not find function for [OPTIONS] to /login resource` right above a message saying `Mounting index.login (nodejs6.10) at http://127.0.0.1:3000/login [POST]`  
Below is my template.yaml.  I attempted to follow the example [here](https://github.com/awslabs/serverless-application-model/blob/master/examples/2016-10-31/api_swagger_cors/swagger.yaml), but still no luck. 

However, if I remove the "options" part of the swagger definition, I don't get the OPTIONS warning.

```
            options:
              consumes:
              - application/json
              produces:
              - application/json
              responses:
                '200':
                  description: 200 response
                  headers:
                    Access-Control-Allow-Origin:
                      type: string
                    Access-Control-Allow-Methods:
                      type: string
                    Access-Control-Allow-Headers:
                      type: string
              x-amazon-apigateway-integration:
                responses:
                  default:
                    statusCode: 200
                    responseParameters:
                      method.response.header.Access-Control-Allow-Methods: "'DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT'"
                      method.response.header.Access-Control-Allow-Headers: "'Content-Type,Authorization,X-Amz-Date,X-Api-Key,X-Amz-Security-Token'"
                      method.response.header.Access-Control-Allow-Origin: "'*'"
                passthroughBehavior: when_no_match
                requestTemplates:
                  application/json: "{\"statusCode\": 200}"
                type: mock
``` 

Any help / suggestions would be greatly appreciated

```
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'

Parameters:
  TableName:
    Default: <TableName>
    Type: String
  FunctionNamePrefix:
    Default: <TableName>
    Type: String
  Secret:
    Default: ""
    Type: String
  AppConfigReadCapacity:
    Default: 5
    Type: Number
  AppConfigWriteCapacity:
    Default: 5
    Type: Number
  AppConfigPrimaryKey:
    Default: id
    Type: String
  Env:
    Default: dev
    Type: String
  S3Bucket:
[...]

PR #1009, opened 2019-02-14
WIP: Implemented support for CORS in start-api command

devoto13:
*Issue #, if available:*

#323

*Description of changes:*

This adds a basic support for CORS in `start-api` command. Currently it only understands format specified in [SAM](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#cors-configuration) and doesn't understand explicit definition with `x-amazon-apigateway-integration` with `type=mock`. I believe that this should solve most of the use cases and support for the explicit specification can be added at the later stage as it is a more complex task.

At this stage I'm looking for the confirmation that the direction is right and if so I'll proceed with a cleanup, tests and documentation.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/local_api_service.py
+++ b/samcli/commands/local/lib/local_api_service.py
@@ -100,7 +100,7 @@ class LocalApiService(object):
         routes = []
         for api in api_provider.get_all():
             route = Route(methods=[api.method], function_name=api.function_name, path=api.path,
-                          binary_types=api.binary_media_types)
+                          binary_types=api.binary_media_types, cors=api.cors)
             routes.append(route)
 
         return routes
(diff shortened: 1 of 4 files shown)


devoto13:
Just a short updated on this. I've used `cors` property of the `Route` class, because it was already there, but after some thinking it seems that the better idea is to remove the property and instead introduce two subclasses for `Route`: one for Lambda integration and another one for the CORS response (which is essentially a quite limited `type=mock` integration, which can later be extended).

Then the implementation will add a new `OPTIONS` route for every unique path and if this route is hit, it will respond with the CORS headers specified in the SAM.

devoto13:
I switched to another project and won't have time to move this forward in the near future, so closing the PR. There is a decent workaround described in https://github.com/awslabs/aws-sam-cli/issues/323#issuecomment-483650280.

PR #1013, opened 2019-02-15
Accept non-integer statusCode JSON values

slank:
*Description of changes:*

The Lambda proxy integration for API Gateway does not specify that statusCode must be an integer, but SAM CLI requires it. This change fixes SAM CLI to accept non-integer status codes that contain integer values.

See https://github.com/slank/awsgi/pull/23 for additional background.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -341,6 +341,15 @@ def test_status_code_not_int(self):
                                                    binary_types=[],
                                                    flask_request=Mock())
 
+    def test_status_code_int_str(self):
```
jfuss: Can you add a test for when ValueError is raised when status_code is not a positive integer as a string?

slank: Sure thing. Added.

slank:
I've added an integration test in what I _think_ is the right place. I get S3 client errors trying to run those tests locally so we'll see what Travis says.

jfuss:
@slank Thanks for adding the tests. Some of our integ-tests require creds, so it's could be either you don't have the correct permissions or no credentials.

PR #1020, opened 2019-02-21
feat(publish): Add --semantic-version option to sam publish command

paoptu023:
*Description of changes:*
Add `--semantic-version` option to `sam publish` command. It overwrites the `SemanticVersion` value in the original template metadata, and publishes application with the provided version number. 

The use case is that customers can easily use the build number from their continuous deployment pipelines to publish new application versions, without manually updating the template.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Windows testing
- [ ] Write documentation


paoptu023:
Can we get an ETA for when this PR will get reviewed?

samcli/commands/publish/command.py
```diff
@@ -31,31 +33,40 @@
 """
 SHORT_HELP = "Publish a packaged AWS SAM template to the AWS Serverless Application Repository."
 SERVERLESSREPO_CONSOLE_URL = "https://console.aws.amazon.com/serverlessrepo/home?region={}#/published-applications/{}"
+SEMANTIC_VERSION_HELP = "Optional. The value provided here overwrites SemanticVersion in the template metadata."
```
TheSriram: are there any restrictions on what the version can be set to? or is it up to the user to setup any version?

paoptu023: It's up to the user, but if they pass in invalid version numbers, the API call would fail. I didn't include an explanation of valid version numbers here, because it's already in the [metadata doc](http://dev-dsk-mhirayam-2c-5d259f39.us-west-2.amazon.com/docs-preview/serverless-application-model/latest/developerguide/serverless-sam-template-publishing-applications-metadata-properties.html).

jfuss: Do we catch the exception correct and provide a useful message to customers within the CLI?

samcli/commands/publish/command.py
```diff
@@ -31,31 +33,40 @@
...
     try:
         template_data = get_template_data(template)
     except ValueError as ex:
         click.secho("Publish Failed", fg='red')
         raise UserException(str(ex))
 
+    # Overwrite SemanticVersion in template metadata when provided in command input
+    if semantic_version and SERVERLESS_REPO_APPLICATION in template_data.get(METADATA, {}):
+        template_data.get(METADATA).get(SERVERLESS_REPO_APPLICATION)[SEMANTIC_VERSION] = semantic_version
+        with open(template, 'w') as fp:
```
jfuss: What is the expectation here? What happens when this is a template that is generated and isn't the exact `source` copy customers edit themselves?

jfuss:
Please update the `sam publish` design doc on changes.

samcli/commands/publish/command.py
paoptu023: Will add code to catch the exception and print user-friendly error message.

samcli/commands/publish/command.py
paoptu023: The expectation is that the template used to publish will always be up to date, and the latest semantic version can be tracked in the template. In most cases it's not the source copy since the template needs to be packaged first, but modifying the source copy requires additional inputs (like path to the source template), even with that we can't ensure it's the "source". I think it's better to keep the command simple and just update the published template.

samcli/commands/publish/command.py
```diff
@@ -122,14 +133,19 @@ def _wrap_s3_uri_exception(ex):
...
+        UserException if found invalid S3 URI or SemVer, else ClientError
     """
     error_code = ex.response.get('Error').get('Code')
     message = ex.response.get('Error').get('Message')
 
-    if error_code == 'BadRequestException' and "Invalid S3 URI" in message:
-        return UserException(
-            "Your SAM template contains invalid S3 URIs. Please make sure that you have uploaded application "
-            "artifacts to S3 by packaging the template: 'sam package --template-file <file-path>'.")
+    if error_code == 'BadRequestException':
```
TheSriram: I'm not too fond of looking at the message of the exception to determine the exception we throw, I dont know if there is a better way to do this though.

paoptu023: The only way to tell whether this exception is about S3 or Version is from the message, because we throw general BadRequestException 😞: https://docs.aws.amazon.com/serverlessrepo/latest/devguide/applications-applicationid-versions-semanticversion.html.

samcli/commands/publish/command.py
jfuss: I am not crazy about writing this back to the template. Till now, none of the commands writes or updates the template that is being handled. Can we pass the semantic version to the `publish_application` instead? Customers using `build` will NEED to update the source template anyways, as running build will squash this change to the template. I would prefer a solution that matches the expectation of editing the source rather than possibly editing the generated sources.

samcli/commands/publish/command.py
jfuss: Are these the only ones we need to be catching? What is thrown when the semver already exists? Is throttle a concern?

To @TheSriram note, can this be handled within the `publish_application` call? Instead of consumers of the library having to deal with all these cases, you could expand the library to throw more friendly exceptions. This would result in cleaner code here and for any other consumers.

paoptu023: Makes sense, I'll handle the exceptions inside serverlessrepo, except the S3 URI one. I still want to keep it here since the error message would be specific to sam cli (telling people to run sam package), what do you think?

samcli/commands/publish/command.py
paoptu023: Made the change.

PR #1023, opened 2019-02-25
fix: Remove tests from .whl file (bdist_wheel)

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
Python wheels, `.whl` files, produced through `python setup.py bdist_wheel` do not read or understand `MANIFEST.in` files. Since we use this file to remove tests modules from the dists, tests end up in the `.whl` files that get uploaded to PyPi. This PR updates the `exclude` parameter in `find_packages` to exclude all `test.*` modules, following [setuptools documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages).

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     author_email='aws-sam-developers@amazon.com',
     url='https://github.com/awslabs/aws-sam-cli',
     license='Apache License 2.0',
-    packages=find_packages(exclude=('tests', 'docs')),
+    packages=find_packages(exclude=['tests.*', 'tests']),
     keywords="AWS SAM CLI",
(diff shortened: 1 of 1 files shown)


tbradley-sans:
Hi, I don't totally understand this change, so I'm wondering if it addresses the problem I'm having right now. I just realized that when I do `sam build -u` somehow some Python unit tests that don't belong to me are getting put into my project's `tests` directory. They don't have any copyright notices or even code comments (shame, shame) but since they're importing things like `serverlessrepo.parser` I'm guessing they belong to SAM. True?

PR #1030, opened 2019-02-28
fix(build): use container directories in executable_search_paths when building in container

### LINKED ISSUE #1029: `sam build` for Gradle using gradlew does not work with --use-container
### Description

Briefly describe the bug you are facing.
When a Gradle project uses `gradlew` scripts, this file does not get picked up with building inside a container.

### Steps to reproduce

Provide steps to replicate.
This integration test sets everything up to repro this issue - https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/buildcmd/test_build_cmd.py#L256

### Observed result

Please provide command output with `--debug` flag set.
`gradle` installation within the container is used instead of the `gradlew` script

### Expected result

Describe what you expected.
`gradlew` script is used to build the project 

1. OS:
2. `sam --version`:

`Add --debug flag to command you are running`

sanathkr:
Fixes #1029

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_build_container.py
+++ b/samcli/local/docker/lambda_build_container.py
@@ -46,6 +46,11 @@ class LambdaBuildContainer(Container):
 
         container_dirs = LambdaBuildContainer._get_container_dirs(source_dir, manifest_dir)
 
+        executable_search_paths = LambdaBuildContainer._convert_to_container_dirs(executable_search_paths, {
+            source_dir: container_dirs["source_dir"],
+            manifest_dir: container_dirs["manifest_dir"]
+        })
+
         request_json = self._make_request(protocol_version,
                                           language,
                                           dependency_manager,
@@ -160,6 +165,54 @@ class LambdaBuildContainer(Container):
             # It is possible that the manifest resides within the source. In that case, we won't mount the manifest
             # directory separately.
             result["manifest_dir"] = result["source_dir"]
+
+        return result
+
+    @staticmethod
+    def _convert_to_container_dirs(host_paths_to_convert, host_to_container_path_mapping):
+        """
+        Use this class to convert a list of host paths to a list of equivalent paths within the container
+        where the given host path is mounted. This is necessary when SAM CLI needs to pass path information to
+        the Lambda Builder running within the container.
+
+        If a host path is not mounted within the container, then this method simply passes the path to the result
+        without any changes.
+
+        Ex:
+            [ "/home/foo", "/home/bar", "/home/not/mounted"]  => ["/tmp/source", "/tmp/manifest", "/home/not/mounted"]
+
+        Parameters
+        ----------
+        host_paths_to_convert : list
+            List of paths in host that needs to be converted
+
+        host_to_container_path_mapping : dict
+            Mapping of paths in host to the equivalent paths within the container
+
+        Returns
+        -------
+        list
+            Equivalent paths within the container
+        """
+
+        if not host_paths_to_convert:
+            # Nothing to do
+            return host_paths_to_convert
+
+        # Make sure the key is absolute host path. Relative paths are tricky to work with because two different
+        # relative paths can point to the same directory ("../foo", "../../foo")
+        mapping = {str(pathlib.Path(p).resolve()): v for p, v in host_to_container_path_mapping.items()}
+
+        result = []
+        for original_path in host_paths_to_convert:
+            abspath = str(pathlib.Path(original_path).resolve())
+
+            if abspath in mapping:
+                result.append(mapping[abspath])
+            else:
+                result.append(original_path)
+                LOG.debug("Cannot convert host path '{}' to its equivalent path within the container. "
+                          "Host path is not mounted within the container", abspath)
 
         return result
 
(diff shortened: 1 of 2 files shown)


samcli/local/docker/lambda_build_container.py
```diff
@@ -46,6 +46,11 @@ def __init__(self,  # pylint: disable=too-many-locals
 
         container_dirs = LambdaBuildContainer._get_container_dirs(source_dir, manifest_dir)
 
+        executable_search_paths = LambdaBuildContainer._convert_to_container_dirs(executable_search_paths, {
```
TheSriram: Nit: can we add kwarg mapping.

samcli/local/docker/lambda_build_container.py
```diff
@@ -163,6 +168,54 @@ def _get_container_dirs(source_dir, manifest_dir):
...
+        # Make sure the key is absolute host path. Relative paths are tricky to work with because two different
+        # relative paths can point to the same directory ("../foo", "../../foo")
+        mapping = {str(pathlib.Path(p).resolve()): v for p, v in host_to_container_path_mapping.items()}
+
+        result = []
+        for original_path in host_paths_to_convert:
+            abspath = str(pathlib.Path(original_path).resolve())
+
+            if abspath in mapping:
+                result.append(mapping[abspath])
```
TheSriram: This is under the assumption that, the base path we are looking for, is already mounted within the container correct?

eg:
/a/b is already mounted within container, and we are just passing additional /a/b/c as additional path.

sanathkr: No, result is a list. I am just adding elements to the list without altering the container paths themselves.

samcli/local/docker/lambda_build_container.py
```diff
@@ -163,6 +168,54 @@ def _get_container_dirs(source_dir, manifest_dir):
 
         return result
 
+    @staticmethod
+    def _convert_to_container_dirs(host_paths_to_convert, host_to_container_path_mapping):
+        """
+        Use this class to convert a list of host paths to a list of equivalent paths within the container
```
TheSriram: class -> method

samcli/local/docker/lambda_build_container.py
```diff
@@ -163,6 +168,54 @@ def _get_container_dirs(source_dir, manifest_dir):
...
+        where the given host path is mounted. This is necessary when SAM CLI needs to pass path information to
+        the Lambda Builder running within the container.
+
+        If a host path is not mounted within the container, then this method simply passes the path to the result
```
TheSriram: ok this is a passthrough, but searching this path within the container will not turn up anything right?

sanathkr: Correct

samcli/local/docker/lambda_build_container.py
sanathkr: Done

samcli/local/docker/lambda_build_container.py
```diff
@@ -46,6 +46,18 @@ def __init__(self,  # pylint: disable=too-many-locals
 
         container_dirs = LambdaBuildContainer._get_container_dirs(source_dir, manifest_dir)
 
+        # `executable_search_paths` are provided as a list of paths on the host file system that needs to passed to
```
TheSriram: Perfect 🔥

samcli/local/docker/lambda_build_container.py
```diff
@@ -163,6 +175,54 @@ def _get_container_dirs(source_dir, manifest_dir):
...
+
+        # Make sure the key is absolute host path. Relative paths are tricky to work with because two different
+        # relative paths can point to the same directory ("../foo", "../../foo")
+        mapping = {str(pathlib.Path(p).resolve()): v for p, v in host_to_container_path_mapping.items()}
+
+        result = []
+        for original_path in host_paths_to_convert:
+            abspath = str(pathlib.Path(original_path).resolve())
+
+            if abspath in mapping:
```
TheSriram: In other words, if the path that is supplied to be added as executable_search_path within the container is already present in the mapping of paths to be mounted, we add the path to be mounted within the container to list of executable search paths.

Also this is very opinionated to adding just source_dir or manifest_dir within the container to search paths, and not any other sub paths.

eg: instead of just /tmp/source_dir and /tmp/manifest_dir , we dont yet support /tmp/source_dir/myexecutable_versions/

tests/unit/local/docker/test_lambda_build_container.py
```diff
@@ -3,6 +3,7 @@
 """
 
 import json
+import pathlib
```
TheSriram: This needs to be fixed for Python2

sanathkr: Fixed

samcli/local/docker/lambda_build_container.py
sanathkr: Correct, we don't support that. This is a good call out. We should be okay for the `gradlew` case now, but we definitely need to implement a generic path-translation logic that handles lot of these cases

PR #1036, opened 2019-03-02
Fix example in docs: `read_timeout` has invalid value (0) for urllib3.Timeout(read).

klichukb:
This should use if intention was to define "infinite timeout".

*Issue #, if available:*
I have couple of long-running lambdas over 1 minute. In order to not get timeouts I used `read_timeout=0` from example in docs, but that doesn't work:
```
ValueError: Attempted to set read timeout to 0, but the timeout cannot be set to a value less than or equal to 0.
```

*Description of changes:*
Use `None` instead as per 
https://urllib3.readthedocs.io/en/latest/reference/index.html#urllib3.Timeout
(Works in both py2 and py3)

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [x] Write documentation

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -114,7 +114,7 @@ if running_locally:
         verify=False,
         config=botocore.client.Config(
             signature_version=botocore.UNSIGNED,
-            read_timeout=0,
+            read_timeout=None,
             retries={'max_attempts': 0},
         )
     )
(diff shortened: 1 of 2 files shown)


TheSriram:
Does this require an integration test change as well?

jfuss:
We should update https://github.com/awsdocs/aws-sam-developer-guide/blob/master/doc_source/serverless-sam-cli-using-automated-tests.md instead of our `usage.md` file. 

I will make an issue to remove the duplicated `.md` files from our repo.

TheSriram:
https://github.com/awsdocs/aws-sam-developer-guide/ should be the authoritative docs page.

PR #1040, opened 2019-03-05
fix: gradle init support

TheSriram:
- Add ability in sam init to specify dependency manager and generate
  appropriate sam templates.

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/init/__init__.py
```diff
@@ -13,29 +13,53 @@
...
+    "ruby2.5_bundler": os.path.join(_templates, "cookiecutter-aws-sam-hello-ruby"),
+    "nodejs6.10_npm": os.path.join(_templates, "cookiecutter-aws-sam-hello-nodejs6"),
+    "nodejs8.10_npm": os.path.join(_templates, "cookiecutter-aws-sam-hello-nodejs"),
+    "nodejs_npm": os.path.join(_templates, "cookiecutter-aws-sam-hello-nodejs"),
+    "dotnetcore2.0_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore2.1_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore1.0_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore_cli_package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnet_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "go1.x_go-get": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
```
jfuss: go_get? Shouldn't this be dep vs modules to match our builders?

samcli/local/init/__init__.py
```diff
@@ -13,29 +13,53 @@
...
+    "go_go-get": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
+    "java8_maven": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-maven"),
+    "java_maven": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-maven"),
+    "java8_gradle": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-gradle"),
+    "java_gradle": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-gradle")
+}
+
+RUNTIME_DEP_MAPPING = {
```
jfuss: Why not encapsulate this into the [Runtimes enum](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/docker/lambda_image.py#L23) to keep all of this into one place?

samcli/local/init/__init__.py
```diff
@@ -13,29 +13,53 @@
...
+    "ruby2.5": ["bundler"],
+    "nodejs6.10": ["npm"],
+    "nodejs8.10": ["npm"],
+    "nodejs": ["npm"],
+    "dotnetcore2.0": ["cli-package"],
+    "dotnetcore2.1": ["cli-package"],
+    "dotnetcore1.0": ["cli-package"],
+    "dotnetcore": ["cli-package"],
+    "dotnet": ["cli-package"],
+    "go1.x": ["go-get"],
```
jfuss: dep or modules?

samcli/local/init/__init__.py
```diff
@@ -13,29 +13,53 @@
...
-    "dotnetcore2.0": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
-    "dotnetcore2.1": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
-    "dotnetcore1.0": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
-    "dotnetcore": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
-    "dotnet": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
-    "go1.x": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
-    "go": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
-    "java8": os.path.join(_templates, "cookiecutter-aws-sam-hello-java"),
-    "java": os.path.join(_templates, "cookiecutter-aws-sam-hello-java")
+RUNTIME_DEP_TEMPLATE_MAPPING = {
```
jfuss: There is a lot of mappings to keep updated all together. Can we condense these into one?

samcli/local/init/__init__.py
```diff
@@ -13,29 +13,53 @@
...
+    "nodejs_npm": os.path.join(_templates, "cookiecutter-aws-sam-hello-nodejs"),
+    "dotnetcore2.0_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore2.1_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore1.0_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnetcore_cli_package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "dotnet_cli-package": os.path.join(_templates, "cookiecutter-aws-sam-hello-dotnet"),
+    "go1.x_go-get": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
+    "go_go-get": os.path.join(_templates, "cookiecutter-aws-sam-hello-golang"),
+    "java8_maven": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-maven"),
+    "java_maven": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-maven"),
```
jfuss: Do we need a `java_maven` and a `java8_maven`? This questions is a general question as we have shortcuts for every language which makes `sam init --help` super long and is only going to get worse if we have to shortcut all language + dependency manager combinations.

The code for maven and gradle look the same. Are they? If so, do we need to maintain two copies directly? Can we combine them so our java examples are the same but can have different dependency managers?

tests/unit/local/init/test_init.py
```diff
@@ -4,26 +4,28 @@
...
-from samcli.local.init import RUNTIME_TEMPLATE_MAPPING
+from samcli.local.init import RUNTIME_DEP_TEMPLATE_MAPPING
 
 
 class TestInit(TestCase):
 
     def setUp(self):
         self.location = None
         self.runtime = "python3.6"
+        self.dependency_manager = "pip"
```
jfuss: What about the `None` case?

samcli/local/init/__init__.py
TheSriram: Yeah, you are right, I just went by the go inits.

samcli/local/init/__init__.py
TheSriram: Yes, will change it.

samcli/local/init/__init__.py
TheSriram: Maybe, I need a double mapping though. Language -> Dep manager and language, dep combo to template. will think about it.

samcli/local/init/__init__.py
TheSriram: Not really the same, slightly different source files, but the core java files are the same. We can look at de-duping that bit of source code.

tests/unit/local/init/test_init.py
TheSriram: Will add a test for this, if a dependency manager is not given, we default to our choice.

samcli/local/init/__init__.py
TheSriram: Will see how to condense this.

samcli/local/common/runtime_template.py
```diff
@@ -0,0 +1,98 @@
...
+    NODEJS_NPM_CONFIG, JAVA_GRADLE_CONFIG
+
+_init_path = str(pathlib.Path(os.path.dirname(__file__)).parent)
+_templates = os.path.join(_init_path, 'init', 'templates')
+
+# NOTE(TheSriram): Builder configs are only indicative of a builder config that is used for selecting a build workflow.
+# The config mutates based on the workflow selector, so they  are not to be imported and used from the template mapping.
+# They are an indication of metadata associated with a specific runtime.
+
+RUNTIME_DEP_TEMPLATE_MAPPING = {
```
sanathkr: This does not make sense. Why would you depend on the build workflow configs within init. They should be treated as isolated modules. Otherwise you will end up with a tangled mess of dependencies one year from now.

samcli/local/common/runtime_template.py
```diff
@@ -0,0 +1,98 @@
...
+            "runtimes": ["java8"],
+            "dependency_manager": "gradle",
+            "init_location": os.path.join(_templates, "cookiecutter-aws-sam-hello-java-gradle"),
+            "build": True,
+            "builder_config": JAVA_GRADLE_CONFIG
+        }
+    ]
+}
+
+SUPPORTED_DEP_MANAGERS = set([c['dependency_manager'] for c in list(
```
sanathkr: I understand what this line is doing, but boy there are so many brackets!

samcli/local/common/runtime_template.py
TheSriram: Will remove that. 👍

samcli/local/common/runtime_template.py
TheSriram: Yes, Having a single config, makes extraction a bit more dense.

PR #1046, opened 2019-03-10
Add Docker Delegated Consistency to Volume

metaskills:
ISSUE: Performance for Docker runs using `sam local start-api` on Mac is horrible. I benchmarked requiring a large Ruby application at ~45seconds.

FIX: Resolve Docker on Mac performance issues by allowing the project folder to be mounted as both "read only" (current behavior) and "delegated". The docs for Docker mention that a comma-separated list is an allowed value. Hence I maintained the `ro` mode and simply added `delegated` to the list. Doc: https://docs.docker.com/v17.09/engine/admin/volumes/volumes/

Please see this article for a description of the new option added. https://docs.docker.com/docker-for-mac/osxfs-caching/. If tests pass, I assume this option being added is a noop for non Mac platforms.

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -84,7 +84,7 @@ class Container(object):
         if self.is_created():
             raise RuntimeError("This container already exists. Cannot create again.")
 
-        LOG.info("Mounting %s as %s:ro inside runtime container", self._host_dir, self._working_dir)
+        LOG.info("Mounting %s as %s:ro,delegated inside runtime container", self._host_dir, self._working_dir)
 
         kwargs = {
             "command": self._cmd,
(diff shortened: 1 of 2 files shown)


metaskills:
OH! Performance for my application's require went down to ~5seconds after this change.

metaskills:
OK, just pushed up some test changes which should get things green.

metaskills:
Could you re-run the tests? The failure looks like a random one.

TheSriram:
What kind of guarantees does this affect?

metaskills:
None that I know of, especially given the mount was read-only to begin with. Did you have a specific guarantee in mind @TheSriram?

sanathkr:
After reading through https://docs.docker.com/docker-for-mac/osxfs-caching/, I am convinced that `delegated` is sufficient for a read-only bind mount where we need NO consistency between host and container file system (except at start up). 

Thanks for digging through Docker and making this change. I think many customers will find this valuable, especially on Mac.

PR #1049, opened 2019-03-10
enable python37 debugging and optional bootstrap file option

### LINKED ISSUE #1047: Enable python3.7 debugging
### Description

Currently python3.7 debugging is blocked but debugging does work as it is documented for python functions here - https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-debugging-python.html

I would also propose to allow the user to pass in a custom `--bootstrap` argument via the `--debug-options` argument. This would allow the following workflow...

1. User creates a `debug/` folder.
2. Install ptvsd into that folder - `pip install ptvsd -t debug/.`.
3. Create a file called `debug_bootstrap` with the following contents:

```
#!/var/lang/bin/python3.7

import ptvsd

ptvsd.enable_attach(address=('0.0.0.0', 5890), redirect_output=True)
ptvsd.wait_for_attach()

import sys
sys.path.insert(0,'/var/runtime')

from bootstrap import main


if __name__ == '__main__':
    main()
```

4. `chmod +x debug/debug_bootstrap`
5. Run `sam local start-api -d 5890 --debugger-path debug/ --debug-args "--bootstrap /tmp/lambci_debug_files/debug_bootstrap"`

billyshambrook:
Fixes #1047

Allow python3.7 debugging. Use the default entrypoint unless `--debug-options` are used to pass in the custom bootstrap file via the `--bootstrap` argument.

### DIFF
--- a/samcli/local/docker/lambda_container.py
+++ b/samcli/local/docker/lambda_container.py
@@ -157,7 +157,7 @@ class LambdaContainer(Container):
         return image_builder.build(runtime, layers)
 
     @staticmethod
-    def _get_entry_point(runtime, debug_options=None):
+    def _get_entry_point(runtime, debug_options=None):  # pylint: disable=R0912
         """
         Returns the entry point for the container. The default value for the entry point is already configured in the
         Dockerfile. We override this default specifically when enabling debugging. The overridden entry point includes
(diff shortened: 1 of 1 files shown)


TheSriram:
#1055 addresses this.

PR #1054, opened 2019-03-12
fix: Skip test_invoke_with_docker_network_of_host_in_env_var on windows

jfuss:
*Issue #, if available:*

*Description of changes:*

On Windows, `test_invoke_with_docker_network_of_host_in_env_var` test hangs indefinitely. This test checks for `SAM_DOCKER_NETWORK` is being read and attached to the container. Instead of creating a new network and inspecting the container, we try to attach to a network that does not exist and assert the command failed correctly. This causes docker to hang on windows (for whatever reason). This PR skips this test on Window platforms only.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -233,6 +234,8 @@ def test_invoke_with_docker_network_of_host(self):
 
         self.assertEquals(return_code, 0)
 
+    @skipIf(platform.system().lower() == 'windows',
```
sanathkr: Can you instead add a separate decorator to a testutils somewhere? I expect many such cases need exclusion for Windows. I'd rather have it defined in one place.

jfuss: Ahh. Good call. Will update this to make it reusable.

PR #1055, opened 2019-03-12
feat(debug): Enable debugging for Python37

sanathkr:
*Issue #, if available:*

*Description of changes:*
Dependent on https://github.com/lambci/docker-lambda/pull/172

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/docker/lambda_container.py
```diff
@@ -293,13 +294,21 @@ def _get_entry_point(runtime, debug_options=None):
                        "/var/runtime/awslambda/bootstrap.py"
                    ]
 
+        elif runtime == Runtime.python37.value:
```
TheSriram: at some point we need to clean up with the if else and make it cleaner. Maybe something like a dictionary key dispatch would work here.

https://github.com/awslabs/aws-sam-cli/issues/853

Issue #885, opened 2018-12-24
local invoke command causes botocore.exceptions.NoRegionError:


MariusVolkhart:
### Description

The `sam local invoke` command produces a `botocore.exceptions.NoRegionError` exception, which prevents its use. 

#583 Fixed this for the sample code it seems

### Steps to reproduce

On Mac OS:
1. Follow direction to install via Homebrew
2. Run ` echo '{"message": "Hey, are you there?" }' | sam local invoke "Signup"`. The function name doesn't matter; it can be anything.

### Observed result

```
 echo '{"message": "Hey, are you there?" }' | sam local invoke "Signup" --debug
2018-12-24 18:26:53 Using SAM Template at /Users/marius_v/Documents/personal/identityvault-serverless/template.yaml
2018-12-24 18:26:53 Changing event name from creating-client-class.iot-data to creating-client-class.iot-data-plane
2018-12-24 18:26:53 Changing event name from before-call.apigateway to before-call.api-gateway
2018-12-24 18:26:53 Changing event name from request-created.machinelearning.Predict to request-created.machine-learning.Predict
2018-12-24 18:26:53 Changing event name from before-parameter-build.autoscaling.CreateLaunchConfiguration to before-parameter-build.auto-scaling.CreateLaunchConfiguration
2018-12-24 18:26:53 Changing event name from before-parameter-build.route53 to before-parameter-build.route-53
2018-12-24 18:26:53 Changing event name from request-created.cloudsearchdomain.Search to request-created.cloudsearch-domain.Search
2018-12-24 18:26:53 Changing event name from docs.*.autoscaling.CreateLaunchConfiguration.complete-section to docs.*.auto-scaling.CreateLaunchConfiguration.complete-section
2018-12-24 18:26:53 Changing event name from before-parameter-build.logs.CreateExportTask to before-parameter-build.cloudwatch-logs.CreateExportTask
2018-12-24 18:26:53 Changing event name from docs.*.logs.CreateExportTask.complete-section to docs.*.cloudwatch-logs.CreateExportTask.complete-section
2018-12-24 18:26:53 Changing event name from before-parameter-build.cloudsearchdomain.Search to before-parameter-build.cloudsearch-domain.Search
2018-12-24 18:26:53 Changing event name from docs.*.cloudsearchdomain.Search.complete-section to docs.*.cloudsearch-domain.Search.complete-section
[...]

PR #1059, opened 2019-03-13
fix: Move boto3 client creation to a computed property

jfuss:
*Issue #, if available:*
#885

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -208,6 +208,8 @@ class LocalLambdaRunner(object):
 
         # to pass command line arguments for region & profile to setup boto3 default session
         LOG.debug("Loading AWS credentials from session with profile '%s'", self.aws_profile)
+        # boto3.session.Session is not thread safe. To ensure we do not run into a race condition with start-lambda
+        # or start-api, we create the session object here on every invoke.
         session = boto3.session.Session(profile_name=self.aws_profile, region_name=self.aws_region)
 
         if not session:
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -26,6 +26,8 @@ class LocalLambdaRunner(object):
                  local_runtime,
                  function_provider,
                  cwd,
+                 aws_profile=None,
+                 aws_region=None,
                  env_vars_values=None,
                  debug_context=None):
         """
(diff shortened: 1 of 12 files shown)


samcli/local/layers/layer_downloader.py
```diff
@@ -36,7 +36,11 @@ def __init__(self, layer_cache, cwd, lambda_client=None):
         """
         self._layer_cache = layer_cache
         self.cwd = cwd
-        self.lambda_client = lambda_client or boto3.client('lambda')
+        self._lambda_client = lambda_client
+
+    @property
+    def lambda_client(self):
+        return self._lambda_client or boto3.client('lambda')
```
TheSriram: I'm ok with the change, just curious if it will change our failure mode scenarios anywhere.

jfuss: No. This just means we create the default boto3 client when the property is accessed instead of on creation. This will mean we will be creating a new boto3 client on every access though. Maybe this is a even better way?

```
 self._lambda_client = self._lambda_client or boto3.client('lambda')
 return self._lambda_client
```

jfuss:
Finally had some time to dig into the build failures. Turns out, boto3 sessions are not thread safe.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -203,13 +207,10 @@ def get_aws_creds(self):
...
-            session = boto3.DEFAULT_SESSION
-        else:
-            session = boto3.session.Session()
-
-        profile_name = session.profile_name if session else None
-        LOG.debug("Loading AWS credentials from session with profile '%s'", profile_name)
+        LOG.debug("Loading AWS credentials from session with profile '%s'", self.aws_profile)
+        # boto3.session.Session is not thread safe. To ensure we do not run into a race condition with start-lambda
+        # or start-api, we create the session object here on every invoke.
+        session = boto3.session.Session(profile_name=self.aws_profile, region_name=self.aws_region)
```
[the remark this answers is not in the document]
jfuss: For the customer, it really doesn't matter. I don't think a debug statement for this makes sense. 

For further context, this is how we did it prior to the work we did on Layers. We got fancy in using the default session. Turns out that only really worked because we were calling `boto3.client`. Once we moved this call to be computed, I was seeing sporadic failures on a `KeyError` within `botocore`. So I reverted us back to our previous state, which ensures Sessions get created per invoke. Which means for `start-lambda` and `start-api` that could be per request.

PR #988, opened 2019-02-07
[WIP] design(build): metadata output

TheSriram:
- output metadata from `sam build` so that the results can be used for
  gleaning information to set up other tools. eg: IDEs, Debuggers.

- https://github.com/awslabs/aws-sam-cli/issues/930
*Checklist:*

- [X] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1095, opened 2019-04-02
Design and implementation for producing debug build artifacts

jfuss:
*Issue #, if available:*
[Rendered Design](https://github.com/jfuss/aws-sam-cli/blob/debuggable-build-artifacts/designs/build_debug_artifacts.md)

*Description of changes:*

Proposal for introducing a way to produce debug artifacts through `sam build`

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+
+Success criteria for the change
+-------------------------------
+
+1. Artifacts generated will be debuggable for DotNetCore2.1 Runtimes
+
+Out-of-Scope
+------------
+
+1. Other languages `sam build` supports will not be changed
```
TheSriram: its a no-op for other languages im assuming.

designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. A command line option (`--debug-artifact`)
+    * Pros
+        * Customer can control what kind of artifacts are produced.
+        * There is no guessing needed by the CLI
+    * Cons
+        * Yet another option for customers to learn or know about
+        * Makes the customer need to think about the artifacts they need to produce (though this can be hidden
+        behind the AWS Toolkit in IDEs)
+        * Customers running in the command line need to remember what artifacts to produce or what they previously produced
```
TheSriram: this is the case, even if you using are the native binary to do it. eg: dotnet publish -c Debug. although I'm not sure what artifacts are produced if no configurations are specified.

designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+2. An Environment Variable we read (`SAM_BUILD_DEBUG_ARTIFACT`)
+    * Pros
+        * Could force customers to use AWS Toolkit by default
+    * Cons
+        * Building debug artifacts becomes a hidden feature (silent contract)
```
TheSriram: we dont want it to be implicit definitely.

designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+        * Environment Variables tend to be more set and forget.
+        * Need to conditionally add Env Var instead of a more convenient flag
+3. `sam build` produces debug artifacts by default and `sam package` will build non debug artifacts by default
+    * Pros
+        * Customers should never have to think about 'when to produce debug artifacts' or forget to add a flag
```
TheSriram: They are completely different usecases right? so I would think would need to know about the modes required to produce 2 different sets of artifacts.

TheSriram:
In general, I'm all for us being very explicit on what the command would do, So i prefer the flag approach.

designs/build_debug_artifacts.md
jfuss: Yes, for now. We could produce metadata (debug specific) from #988 only under this case but that would be in the future, if we do follow that approach. For now, it will produce the same artifacts as build does today. We could add a log but the artifacts are already debuggable so I don't see a need to fail or do any special handling.

designs/build_debug_artifacts.md
jfuss: For DotNet, I believe it is Release by default and same with Go (since you have to add the compile flag at build time). 

I did consider following something like DotNet here. That is a `--artifacts Debug` but I don't see another build case, so flag seemed better to me.

designs/build_debug_artifacts.md
jfuss: Maybe? The framing I was thinking is, if one command does the building, packaging, and deploying then the action is defined for us. 

So let's just say we have two commands `invoke` (which only builds and invokes) and `deploy` (which builds, packages, and deploys). One could argue that `invoke` would always build debuggable artifacts and hide away the "do I need do pass different options to the command?". `Deploy` on the other hand, could build for production (leaving out debug headers, etc). In this model, customers never have to think about how they need to build. They would only have to care about what action do I need to accomplish. This can simplify customers interactions, of course we still need some way of controlling things (for say AWS Toolkit and any other tools) and there are other nuances I did not take into consideration. Strictly from a CLI point of view though, we may not need an option at all and that is what this was attempting to capture.

designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+
+What will be changed?
+---------------------
+
+We will introduce a way in `sam build` to produce these debuggable artifacts for those compiled languages.
+
+Success criteria for the change
+-------------------------------
+
+1. Artifacts generated will be debuggable for DotNetCore2.1 Runtimes
```
mpiroc: Is this 2.1 only, or will 2.0 and/or 1.0 be supported? [This doc](https://github.com/awslabs/aws-sam-cli/blob/develop/designs/dotnetcore-debugging.md) mentions both 2.0 and 2.1.

jfuss: Both 2.1 and 2.0 but not 1.0. I will update to make it more explicit.

samcli/commands/build/command.py
```diff
@@ -91,8 +92,10 @@ def cli(ctx,
         parameter_overrides):
     # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
 
+    # TODO fill this out on what env var
+    mode = BuildMode.DEBUG if os.environ.get("") else BuildMode.RELEASE
```
TheSriram: We need this now, if we are going with env var.

designs/build_debug_artifacts.md
```diff
@@ -0,0 +1,129 @@
...
+1. A command line option (`--debug-artifact`)
+    * Pros
+        * Customer can control what kind of artifacts are produced.
+        * There is no guessing needed by the CLI
+    * Cons
+        * Yet another option for customers to learn or know about
+        * Makes the customer need to think about the artifacts they need to produce (though this can be hidden
+        behind the AWS Toolkit in IDEs)
+        * Customers running in the command line need to remember what artifacts to produce or what they previously produced
+2. An Environment Variable we read (`SAM_BUILD_DEBUG_ARTIFACT`). IDE Toolkit will set the env var when calling `sam build`
```
TheSriram: this needs to be corrected as the same in code. `SAM_BUILD_MODE`

tests/unit/lib/build_module/test_app_builder.py
```diff
@@ -194,7 +195,7 @@ def test_must_use_lambda_builder(self, lambda_builder_mock):
                                                          "artifacts_dir",
                                                          "scratch_dir",
                                                          "manifest_path",
-                                                         "runtime")
+                                                         "runtime",)
```
TheSriram: mode here as well?

jfuss: `mode` isn't passed into the `_build_function_in_process` function because `mode` is an attribute on the class (accessed through `self.mode`).

TheSriram:
Integ tests wont pass, because they depend on lambda builders 0.3.0 which is not yet released.

Issue #1097, opened 2019-04-03
Allow compatibility with kotlin gradle scripts


philippeagra:
### Allow compatibility with kotlin gradle scripts

Since gradle 3, it's possible to use kotlin instead of groovy for the gradle build scripts. But this feature is not supported by aws-sam-cli.
An error occurs when we try to use aws-sam-cli on a project with a build.gradle.kts file:
```
Build Failed
Error: Unable to find a supported build workflow for runtime 'java8'.
Reason: None of the supported manifests '['build.gradle', 'pom.xml']' were found in the following paths ...
```

### Proposal

Add a specific configuration for kotlin based gradle projects

sanathkr:
PR merged! Thanks for the help.

hamadycisse:
For those stumbling on the same issue, the solution is now as per AWS SAW 1.18.1 to do:
`/usr/local/bin/sam build HelloWorldFunction  --template /Users/user/Source/Repos/token-payment/template.yaml --build-dir /Users/user/Source/Repos/token-payment/.aws-sam/build -m ./infrastructure/build.gradle.kts`

PR #1098, opened 2019-04-03
Adding specific configuration for gradle projects using kotlin build script

philippeagra:
* Issue #1097
* Allows to use `build.gradle.kts` instead of `build.gradle`


sanathkr:
@philippeagra Thanks for submitting this change. Can you also add an integration test so we know this works now and moving forward?

philippeagra:
@sanathkr yes, it's added.

PR #1111, opened 2019-04-09
Depend on development version of lambda-builders for dev builds

sanathkr:
*Issue #, if available:*
During development, we can't easily use development versions of other dependencies such as AWS Lambda Builders.

*Description of changes:*
In this change, I have separated out dev requirements from prod requirements. For Production, the dependencies are version pinned and pulled from PyPi. For dev requirements, the dependencies are pulled directly from Github source. 

I had to remote the `extras_requires` in setup.py because it did not support installing from Github. Instead I modified the Makefile to pip install these dependencies separately. This is acceptable for development builds.

For production, I concatenate `base.txt` with `prod.txt` and use it as the final list of requirements.

*Testing:*
Following commands worked as expected

- `SAM_CLI_DEV=1 pip install -e .`
- `pip install -e .`
- `make init`

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 init:
 	SAM_CLI_DEV=1 pip install -e .
 	SAM_CLI_DEV=1 pip install -r requirements/dev.txt
+	SAM_CLI_DEV=1 pip install -r requirements/tools.txt
 
 test:
 	# Run unit tests
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 init:
-	SAM_CLI_DEV=1 pip install -e '.[dev]'
+	SAM_CLI_DEV=1 pip install -e .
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,5 +1,7 @@
 init:
-	SAM_CLI_DEV=1 pip install -e '.[dev]'
+	SAM_CLI_DEV=1 pip install -e .
```
TheSriram: we have silently changed the contract of make init now, to always build the latest from lambda-builders in develop. this can issues when there is a breaking change in lambda builders, but the corresponding component to fix sam cli has not yet been merged.

we need to keep this in mind and fix it.

Issue #878, opened 2018-12-19
symlinks in zip are not retained on the docker image created when using Layers


innovia:
### Description
When adding a package that contains symlinks these symlinks are flattened on the docker image

Briefly describe the bug you are facing.
I am taking a folder with symlinks in it and zipping it using the --symlinks option of the zip
when I unpack the zip file locally the symlinks are there but when I run the `sam invoke` command there are failures to find these file.

when debugging the produced container using 
`docker run -it --rm --entrypoint=/bin/bash samcli/lambda:provided-11bcc8dc4810532c82a432b30 -i` the files are flatten into names with no symlinks.

### Steps to reproduce
create a folder with a symlink to a file
package the files using `zip -yr some.zip folder_name`
run `sam local invoke` and call the symlink file.
Provide steps to replicate.
```
git clone git@github.com:gkrizek/bash-lambda-layer.git
cd bash-lambda-layer
zip -yr layer.zip bootstrap bin/ lib/ libexec/ share/
aws lambda publish-layer-version --layer-name myLayer --zip-file fileb://layer.zip
```

template.yaml:
```yaml

AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  Sample SAM Template for sam-app
Globals:
  Function:
    Timeout: 300

Resources:
  Bash:
    Type: AWS::Serverless::Function
    Properties:
        CodeUri: .
        Handler: index.handler
        Runtime: provided
        Layers:
        - arn:aws:lambda:us-west-2:744348701589:layer:bash:4
```

index.sh
``` bash
handler () {
    set -e
    git clone git@github.com:hypnoglow/helm-s3.git
}
```
### Observed result
symlinks are flattened into files
```
ls -la
total 204
drwx------ 3 sbx_user1051  495   4096 Dec 18 20:33 .
drwxr-xr-x 1 root         root   4096 Dec 18 20:33 ..
drwx------ 2 sbx_user1051  495   4096 Dec 18 20:33 fipscheck
-rwx------ 1 sbx_user1051  495     17 Dec 18 20:33 libedit.so.0
-rwx------ 1 sbx_user1051  495 178816 Dec 18 20:33 libedit.so.0.0.27
-rwx------ 1 sbx_user1051  495     21 Dec 18 20:33 libfipscheck.so.1
-rwx------ 1 sbx_user1051  495   6960 Dec 18 20:33 libfipscheck.so.1.1.0
[...]

Issue #477, opened 2018-06-13
Compression packs symlinks as copies


garthk:
**Description:**

`sam package` compresses symlinks as an additional copy of the file to which they resolve. Given a traditional shared library layout with

```
libz.so → libz.so.1
libz.so.1 → libz.so.1.2.11
libz.so.1.2.11
```

… `sam package` will pack the library three times, not once with two symlinks.

**Steps to reproduce the issue:**

`make` in a directory with `.env`, `template.yml`, and `Makefile`:

`.env`: replace the values with your own profile, region, and S3 bucket name:
```
AWS_PROFILE=default
AWS_REGION=ap-southeast-2
BUCKET=your-bucket-name-I-do-not-want-to-know
```
 
`template.yml`:
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: Reproducing a SAM CLI compression wart

Globals:
  Function:
    Runtime: nodejs8.10

Resources:
  DummyFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: Dummy
      CodeUri: ./build
      Handler: lambda.dummy
  ```

`Makefile`:
```make
# Import AWS_PROFILE, AWS_REGION, BUCKET, and S3_PREFIX  from ./.env:
-include .env

# Set defaults, with none of which you'll likely be happy:
AWS_PROFILE ?= default
AWS_REGION ?= ap-southeast-2
BUCKET ?= your-bucket-name-here
S3_PREFIX ?= lambdas

# Targets
EXPECTED := expected.zip
OBSERVED := observed.zip

# Satisfy `aws` and `sam`:
export AWS_PROFILE
export AWS_REGION

# These targets don't produce files:
.PHONY: comparison clean

# This target is our default:
comparison: $(EXPECTED) $(OBSERVED)

# This cleans up:
clean:
	rm -rf build $(OBSERVED) $(EXPECTED)

# Construct the ZIP file we wish SAM had made:
$(EXPECTED): build
	( cd build ; zip -9y ../$(EXPECTED) * )
	unzip -l $(EXPECTED)

# Download and list the contents of the ZIP file `sam package` uploaded:
$(OBSERVED): packaged.yml
	aws s3 cp `grep CodeUri packaged.yml  | awk '{ print $$2 }'` $(OBSERVED)
	unzip -l $(OBSERVED)

[...]

PR #1140, opened 2019-04-20
Respect zipped symlinks

bubba-h57:
*Issue #878 :*

*Description of challenge:*
SAM CLI currently used the `Lib/zipfile.py` from the [Python Standard Library](https://docs.python.org/3/library/index.html). Unfortunately, that library has not been kept up to date with the [.ZIP File Format Specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) and while it does handle basic archival CRUD, there are a number of features that have become standard and which it does not support. 

The greatest interest to us is that of properly handling [symbolic links](https://en.wikipedia.org/wiki/Symbolic_link). It can neither archive them properly (see #477) nor can it extract them properly. This is particularly important now that we are compiling libraries for custom layers. Given the limited file space available to Lambda Functions, having `/libs/*.so` files all copied instead of symlinked exponentially increases the uncompressed layer size. 

Lambda itself handles the archived symlinks properly; however, attempting to test the same layers with `sam local` fails. This forces fat layers if the developer desires to use 'sam local'

*Description of changes:* 
The `unzip()` function in `samcli/local/lambdafn/zip.py` is modified to identify and properly handle any symlinks. The Standard `zipfile` Library continues to do all of the heavy work with a wrapper around it to handle the identification and extraction of the zipfiles. Regular files and directories continue to be extracted in precisely the same manner as before. All changes are documented with appropriate docblocks.

`tests/unit/local/lambdafn/test_zip.py` had to be modified in order to insert a symlink into the test archive as well as testing that the symlink was properly extracted. I did not see any `integration` nor `functional` tests that needed modification.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


sanathkr:
Actually, can you rebase on latest `develop`? It is hard for me to look at the real diff.

bubba-h57:
@sanathkr Rebased and ready. Thanks for reviewing it!

samcli/local/lambdafn/zip.py
```diff
@@ -19,6 +19,70 @@
...
+    file_info : zipfile.ZipInfo
+        The ZipInfo for a ZipFile
+
+    Returns
+    -------
+    bool
+        A response regarding whether the ZipInfo defines a symlink or not.
+    """
+
+    return (file_info.external_attr >> 28) == 0xA
```
jfuss: Should this be using the `S_IFLNK` constant from above?

Why not use `stat. S_IFLNK` here instead?

samcli/local/lambdafn/zip.py
```diff
@@ -19,6 +19,70 @@
...
+    link_name = os.path.normpath(os.path.join(output_dir, file_info.filename))
+
+    # make leading dirs if needed
+    leading_dirs = os.path.dirname(link_name)
+    if not os.path.exists(leading_dirs):
+        os.makedirs(leading_dirs)
+
+    # If the link already exists, delete it or symlink() fails
+    if os.path.lexists(link_name):
+        os.remove(link_name)
```
jfuss: Will this have any (negative) side effects? I can't think of any at the moment but asking the question to spark a discussion.

bubba-h57:
We did indeed test it to solve #878 ... that issue was the whole point of interest for us. :-)

jfuss:
> We did indeed test it to solve #878 ... that issue was the whole point of interest for us. :-)

WOOT! I just wanted to make sure. If you didn't, I was going to pull down and verify (mainly why I was asking).

samcli/local/lambdafn/zip.py
bubba-h57: I tested it fairly extensively and can not think of any negative side effects that might pop up. The unzip _intends_ to overwrite anything that is there regardless ... we will simply delete it prior to writing.

samcli/local/lambdafn/zip.py
bubba-h57: This is not looking at a file on the filesystem. It is looking at a number of bytes in the `file_info` entry, the External Attributes for the file. See [The zip format's external file attribute](https://unix.stackexchange.com/questions/14705/%20the-zip-formats-external-file-attribute) If you want to dig into it. But the short answer is because it isn't a file.

bubba-h57:
Well, I'm all for you testing it also. It has been a while since I looked it.

samcli/local/lambdafn/zip.py
jfuss: I see. You made a constant `S_IFLNK = 0xA` above. Can you either remove it the constant or replace the `0xA` here with `S_IFLNK`. Otherwise, it looks like an unused constant.

bubba-h57: ^ that was with regard to pythons `stat.S_ISLNK()` ... I have no good excuse for not comparing to the *S_IFLNK = 0xA* constant defined directly above the function. :-)

jfuss:
I ended up needing to revert this. Details can be found [here](https://github.com/awslabs/aws-sam-cli/pull/1315#issuecomment-518939441). We will need to revisit the implementation and either way for the [Py2.7 to be depreciated](https://github.com/awslabs/aws-sam-cli/issues/1273) or update this to not work on Py2.7 in windows.

Issue #917, opened 2019-01-07
Support to add SSL Certs to Container so network calls work


DevOpsChris:
I wish sam-cli would add the ability to add SSL certificates that are created for use within a proxy network.  This would enable downloading layers, etc. without tweaking the python code base to `verify = False` requests calls.

### Proposal

1.  Add a volume argument to mount the cert directory in the container to the cert.
-or-
2. Allow post copy of files into the container from aws-sam-cli as a flag.

DevOpsChris:
https://github.com/DevOpsChris/aws-sam-cli/tree/fix_proxy_issues  here's a fix in progress.

DevOpsChris:
Not needed.  Observed you can do this within ENV Variables and have the lambda use the variables within the code to connect to boto3 service calls.

DevOpsChris:
Made a mistake.  I've fixed the layer downloader from behind the proxy with ssl cert.  Here's the fix. #1143

jfuss:
Was released with [v0.16.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.16.0)

Closing

ryanpodonnell1:
This doesn't seem to be picking up the `AWS_CA_BUNDLE` environment variable

PR #1143, opened 2019-04-23
Layer Downloader now looks for AWS_CA_BUNDLE if it exists

DevOpsChris:
*Issue #, if available:* #917 

*Description of changes:*  This enables the layer downloader to us AWS_CA_BUNDLE environment variable if it exists.  It defaults to SSL verify True if it doesn't.

*Checklist:*

- [x] Write Design Document - No.  Doesn't meet reqs. ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/tests/unit/local/lambdafn/test_zip.py
+++ b/tests/unit/local/lambdafn/test_zip.py
@@ -5,7 +5,7 @@ import shutil
 from unittest import TestCase
 from contextlib import contextmanager
 from tempfile import NamedTemporaryFile, mkdtemp
-from mock import Mock, patch
+from mock import Mock, patch, mock
 
 from nose_parameterized import parameterized, param
 
@@ -86,6 +86,7 @@ class TestUnzipFromUri(TestCase):
     @patch('samcli.local.lambdafn.zip.progressbar')
     @patch('samcli.local.lambdafn.zip.requests')
     @patch('samcli.local.lambdafn.zip.open')
+    @mock.patch.dict(os.environ, {})
     def test_successfully_unzip_from_uri(self, open_patch, requests_patch, progressbar_patch, path_patch, unzip_patch):
         get_request_mock = Mock()
         get_request_mock.headers = {"Content-length": "200"}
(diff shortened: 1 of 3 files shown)


DevOpsChris:
I notice that tox runs the `make pr` command.  Is that sufficient, or do I run it alone on my dev box?

tests/unit/local/lambdafn/test_zip.py
```diff
@@ -104,7 +104,7 @@ def test_successfully_unzip_from_uri(self, open_patch, requests_patch, progressb
 
         unzip_from_uri('uri', 'layer_zip_path', 'output_zip_dir', 'layer_arn')
 
-        requests_patch.get.assert_called_with('uri', stream=True)
+        requests_patch.get.assert_called_with('uri', stream=True, verify=True)
```
jfuss: Can you mock out `os.environ` here to return an empty dict or a dict without the `AWS_CA_BUNDLE` key? This will keep the tests from failing on systems that have that env var already set.

DevOpsChris:
I plan on squashing once the testing is good.

tests/unit/local/lambdafn/test_zip.py
DevOpsChris: sounds good to me.  I'll get it done.

DevOpsChris:
Squashed & rebased on latest upstream/develop commit

DevOpsChris:
I'm ready once the test passes

jfuss:
Hey @DevOpsChris, I pushed a patch to your branch. The `os.environ` patch was not patching the environ correctly in the module. I spent some time trying to use the `@patch.dict` but I could not get it to work. So I ended up patching `os` in the module and mocking out the return value. I also added a tests that read the value and expected that `request.get` had the correct value for `verify`.

I will merge once the tests pass on Travis. Thanks for spending the time on this.

DevOpsChris:
Hey thank you for finding that quirk out.  It looks like Travis is temperamental, it failed on a Go language test.  I recommend re-testing it, it seems to make it work again.  I wonder if it's a virtual machine timeout issue.

jfuss:
@DevOpsChris this happens from time to time. It's unclear why but playing whack-a-mole (rerunning) is our current way around it.

PR #1146, opened 2019-04-23
feat(build): Build a single function

jfuss:
*Issue #, if available:*
N/A

### Description of changes:
This enables a customer to build a single function similar to how they would invoke a single function. To build a single function, you can pass the LogicalID to the command `sam build <MyLogicalId>`. If the LogicalID is omitted, the command will behave exactly how it does not and build the whole template.

### Concerns on approach:
This change enables a customer (or IDE) to quickly build a single function for faster test cycles. This could result in a deployment to of source code instead of built code, if the customer only builds one of the n number of functions in the template and then does a package and deploy. With this change, customers will need to still use `sam build` without the logicalId before package and deploying. This could create more customer confusion in the short term.

### Checklist:

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/buildcmd/test_build_cmd.py
+++ b/tests/integration/buildcmd/test_build_cmd.py
@@ -402,3 +402,83 @@ class TestBuildCommand_Dotnet_cli_package(BuildIntegBase):
         all_artifacts = set(os.listdir(str(resource_artifact_dir)))
         actual_files = all_artifacts.intersection(expected_files)
         self.assertEquals(actual_files, expected_files)
+
+
+class TestBuildCommand_SingleFunctionBuilds(BuildIntegBase):
+    template = "many-functions-template.yaml"
+
+    EXPECTED_FILES_GLOBAL_MANIFEST = set()
+    EXPECTED_FILES_PROJECT_MANIFEST = {'__init__.py', 'main.py', 'numpy',
+                                       # 'cryptography',
+                                       "jinja2",
+                                       'requirements.txt'}
+
+    def test_fucntion_not_found(self):
+        overrides = {"Runtime": 'python3.7', "CodeUri": "Python", "Handler": "main.handler"}
+        cmdlist = self.get_command_list(parameter_overrides=overrides,
+                                        function_identifier="FunctionNotInTemplate")
+
+        process = subprocess.Popen(cmdlist, cwd=self.working_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+
+        self.assertEquals(process.returncode, 1)
+        self.assertIn('FunctionNotInTemplate not found', str(stderr.decode('utf8')))
+
+    @parameterized.expand([
+        ("python3.7", False, "FunctionOne"),
+        ("python3.7", "use_container", "FunctionOne"),
+        ("python3.7", False, "FunctionTwo"),
+        ("python3.7", "use_container", "FunctionTwo")
+    ])
+    def test_build_single_function(self, runtime, use_container, function_identifier):
+        # Don't run test on wrong Python versions
+        py_version = self._get_python_version()
+        if py_version != runtime:
+            self.skipTest("Current Python version '{}' does not match Lambda runtime version '{}'".format(py_version,
+                                                                                                          runtime))
+
+        overrides = {"Runtime": runtime, "CodeUri": "Python", "Handler": "main.handler"}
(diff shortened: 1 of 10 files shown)


tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -402,3 +402,83 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
+        resource_artifact_dir = build_dir.joinpath(function_logical_id)
+
+        # Make sure the template has correct CodeUri for resource
+        self._verify_resource_property(str(template_path),
+                                       function_logical_id,
+                                       "CodeUri",
+                                       function_logical_id)
+
+        all_artifacts = set(os.listdir(str(resource_artifact_dir)))
+        print(all_artifacts)
```
TheSriram: nit: remove prints.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -402,3 +402,83 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
+class TestBuildCommand_SingleFunctionBuilds(BuildIntegBase):
+    template = "many-functions-template.yaml"
+
+    EXPECTED_FILES_GLOBAL_MANIFEST = set()
+    EXPECTED_FILES_PROJECT_MANIFEST = {'__init__.py', 'main.py', 'numpy',
+                                       # 'cryptography',
+                                       "jinja2",
+                                       'requirements.txt'}
+
+    def test_fucntion_not_found(self):
```
TheSriram: nit: typo

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -402,3 +402,83 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
         actual_files = all_artifacts.intersection(expected_files)
         self.assertEquals(actual_files, expected_files)
+
+
+class TestBuildCommand_SingleFunctionBuilds(BuildIntegBase):
+    template = "many-functions-template.yaml"
+
+    EXPECTED_FILES_GLOBAL_MANIFEST = set()
+    EXPECTED_FILES_PROJECT_MANIFEST = {'__init__.py', 'main.py', 'numpy',
+                                       # 'cryptography',
```
TheSriram: why is cryptography commented out?

tests/integration/buildcmd/test_build_cmd.py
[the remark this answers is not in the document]
jfuss: done

tests/integration/buildcmd/test_build_cmd.py
jfuss: Done

tests/integration/buildcmd/test_build_cmd.py
jfuss: This was copied from the above python tests.

This is the specific reason: https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/testdata/buildcmd/Python/requirements.txt#L4

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -402,3 +402,83 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
+        ("python3.7", "use_container", "FunctionTwo")
+    ])
+    def test_build_single_function(self, runtime, use_container, function_identifier):
+        # Don't run test on wrong Python versions
+        py_version = self._get_python_version()
+        if py_version != runtime:
+            self.skipTest("Current Python version '{}' does not match Lambda runtime version '{}'".format(py_version,
+                                                                                                          runtime))
+
+        overrides = {"Runtime": runtime, "CodeUri": "Python", "Handler": "main.handler"}
```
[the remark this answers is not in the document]
jfuss: This test is meant to test that we can build a single function in a template of multiple. I really only cared about the single function being built and not that is built the exact one. I am assuming that our other tests are covering this area. So instead of creating more scaffolding, I bootstrapped the these new tests to use the same scaffolding we already have.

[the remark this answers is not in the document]
TheSriram: Yeah i can understand that, what i was trying to get behind was that a customer will not create two resources named differently that reference the same handler. but given that the built artifacts are going to separate build folders, its okay in this case.

PR #1149, opened 2019-04-24
Removed broken snapcraft.io link

rothgar:
The link appears to be broken. I also searched snapcraft for aws-sam-cli and aws-sam and the app isn't found.

Removing the link so people aren't confused. Alternatively, the package could be added but that would probably take more time to fix.

*Description of changes:*
Made install docs clearer.


jfuss:
@rothgar Thanks for the PR. We ended up removing this when we revamped out README.

Closing as it is no longer needed.

Issue #522, opened 2018-06-29
Fix Local API Gateway doesn't catch exceptions and always return 200


danilop:
If you have an error/exception in the Lambda function or return a non-compliant response (without 'statusCode' or 'body'), the behavior of the Local API Gateway is different from the actual Amazon API Gateway: it is always returning a 200 OK.

ivanmartinvalle:
I'm seeing the same behavior. More specifically:

```sh
sam init -r nodejs -n no-error
cd no-error/hello_world
npm install
cd ..
# edit app.js
sam local start-api
curl 127.0.0.1:3000/hello
```

```js
// app.js
exports.lambda_handler = async (event, context, callback) => {
    callback(new Error('THIS IS AN ERROR'));
};
```

The error message appropriately is logged in the console, but the actual API returns a 200 OK with "no data" as the response body. API Gateway returns a 502 Bad Gateway Internal server error.

What's weird though, is that I was able to get the 502 locally ONCE...and then it went away and started returning a 200. I committed the 502 state and reset back to it...and the 502 didn't come back.

jfuss:
Yup this is a bug in the CLI. This was addressed for start-lambda command but it never got into start-api before the release. The explanation can be found [here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/services/base_local_service.py#L143)

```
When the Lambda Function returns an Error/Exception, the output is added to the stdout of the container. From our perspective, the container returned some value, which is not always true. Since the output is the only information we have, we need to inspect this to understand if the container returned a some data or raised an error
```

We need to [update this](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/apigw/local_apigw_service.py#L144). It should be as simply as taking the last value of the return from `LambdaOutputParser.get_lambda_output` and handling the error case correctly.

bernardobelchior:
When the function returns an empty body, `aws-sam-cli` returns the body as "no data". Is this intended behaviour? I'd think AWS Lambda does not work like this

tutumi:
> When the function returns an empty body, `aws-sam-cli` returns the body as "no data". Is this intended behaviour? I'd think AWS Lambda does not work like this

I ran into the same problem and it seems it's intended https://github.com/awslabs/aws-sam-cli/blob/ec5c62b513f560053ed99a77350cae82de528694/samcli/local/apigw/local_apigw_service.py#L198 but I'm not sure if it should be like this.

[...]

PR #1154, opened 2019-04-25
Validate API Gateway Response Keys

awood45:
*Issue #, if available: Resolves Issue #522

*Description of changes:* Per my testing, API Gateway has 4 allowable response keys: `statusCode`, `body`, `headers`, and `isBase64Encoded`. None of these keys are required, but any extra keys will result in a 502 error from the service. Currently, the local API Gateway implementation will ignore extra keys, this change will cause those keys to raise an error, matching service behavior.

*Checklist:*

- [N/A] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [Y] Write unit tests
- [N/A] Write/update functional tests
- [Y] Write/update integration tests
- [Y] `make pr` passes
- [N/A] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -230,6 +235,17 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
...
+    @staticmethod
+    def _invalid_apig_response(output):
+        allowable = [
+            "statusCode",
+            "body",
+            "headers",
+            "isBase64Encoded"
+        ]
+        invalid_keys = set(output) - set(allowable)
+        return invalid_keys != set()
```
sanathkr: `return bool(invalid_keys)` is sufficient! In Python empty list/set is considered Falsy. The `!=` syntax is not too Pythonic ;)

samcli/local/apigw/local_apigw_service.py
```diff
@@ -230,6 +235,17 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
...
 
+    @staticmethod
+    def _invalid_apig_response(output):
+        allowable = [
+            "statusCode",
+            "body",
+            "headers",
+            "isBase64Encoded"
+        ]
+        invalid_keys = set(output) - set(allowable)
```
sanathkr: nit: Is `output` dict? If so explicitly grab the keys using `output.keys()`.

awood45: Updated, and tests still passing locally.

samcli/local/apigw/local_apigw_service.py
awood45: Same story, tests still pass after this change, just merged.

tests/integration/local/start_api/test_start_api.py
```diff
@@ -86,6 +86,12 @@ def test_invalid_response_from_lambda(self):
         self.assertEquals(response.status_code, 502)
         self.assertEquals(response.json(), {"message": "Internal server error"})
 
+    def test_invalid_hash_response_from_lambda(self):
```
TheSriram: nit: json_response?

tests/integration/testdata/start_api/main.py
```diff
@@ -63,6 +63,10 @@ def invalid_response_returned(event, context):
     return "This is invalid"
 
 
+def invalid_hash_response(event, context):
```
TheSriram: nit: maybe just json response?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -230,6 +235,17 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
 
         return status_code, headers, body
 
+    @staticmethod
+    def _invalid_apig_response(output):
+        allowable = [
```
TheSriram: you could make allowable a set to begin with.
`
allowable = {...}
`

tests/integration/local/start_api/test_start_api.py
awood45: Sure

samcli/local/apigw/local_apigw_service.py
awood45: Changed, though I may have to change further as the set subtraction operator doesn't seem to be supported in Python 2.7

samcli/local/apigw/local_apigw_service.py
```diff
@@ -220,6 +220,11 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
             LOG.error(message)
             raise TypeError(message)
 
+        # API Gateway only accepts statusCode, body, headers, and isBase64Encoded in
+        # a response shape.
+        if LocalApigwService._invalid_apig_response(json_output):
+            raise ValueError
```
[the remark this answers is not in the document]
awood45: Done

samcli/local/apigw/local_apigw_service.py
awood45: I made this change, just had to add the `set()` casting explicitly to the keys.

awood45:
Python 2.7 incompatibility is fixed, so this is ready for review again.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -220,6 +220,12 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
             LOG.error(message)
             raise TypeError(message)
 
+        # API Gateway only accepts statusCode, body, headers, and isBase64Encoded in
+        # a response shape.
+        if LocalApigwService._invalid_apig_response(json_output):
+            msg = "Invalid Response: " + str(json_output)
```
TheSriram: do we need to be more explicit in the response about accepted keys?

TheSriram: we can possibly move log statement up here just before throwing an exception.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -230,6 +236,20 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
...
 
+    @staticmethod
+    def _invalid_apig_response(output):
+        allowable = {
+            "statusCode",
+            "body",
+            "headers",
+            "isBase64Encoded"
+        }
+        # In Python 2.7, need to explicitly make the Dictionary keys into a set
```
TheSriram: Good to know!

samcli/local/apigw/local_apigw_service.py
awood45: To do this I'd probably want to actually return the invalid keys and then perform the boolean if at this top level. Should we move to that approach?

TheSriram: Yeah that sounds good, provides more context as well.

awood45: Addressed in most recent change.

PR #398, opened 2018-05-10
Allow for case insensitve header names

koblas:
In testing with a local served files, I noted that Content-Type headers were being compared in a case-sensitive way.  Which means that for an express application which is serving headers up in lowercase, sam local was adding a second 'Content-Type: application/json header.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1166, opened 2019-05-08
Local support for response multiValueHeaders

dsanders11:
*Issue #, if available:*

#830 

*Description of changes:*

This is a rework of #842 by @martysweet, which seems to have stalled. This PR uses Flask's built-in method of providing multiple header values, which makes it act like API GW. In particular, multi-value headers are folded into a comma-separated list *unless* it's `Set-Cookie` which is a special-case.

There's two parts to this PR. First drops `CaseInsensitiveDict` as it doesn't seem to be necessary. The `Headers` class in Flask is already case-insensitive and swapping in `Headers` in the test case for `CaseInsensitiveDict` passed the tests. There are some minor differences:

  * `Headers` isn't JSON serializable by default
  * `Headers` won't pass equality comparison with a dict
  * `Headers` raises `werkzeug.exceptions.BadRequestKeyError` (a subclass of `KeyError`) on bad key

Those could be addressed with a custom subclass of `Headers` if necessary, but at the moment all tests pass just fine.

The second commit is mostly the same as #842 but uses the default behavior of `Headers` to handle multi-value headers instead of manually merging them as a comma-separated list. This lets Flask do the special-case for `Set-Cookie`.

@martysweet, @jfuss, @everyonce, you guys want to take a look since you were active on #842?

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -5,8 +5,9 @@ import logging
 import base64
 
 from flask import Flask, request
+from werkzeug.datastructures import Headers
 
-from samcli.local.services.base_local_service import BaseLocalService, LambdaOutputParser, CaseInsensitiveDict
+from samcli.local.services.base_local_service import BaseLocalService, LambdaOutputParser
 from samcli.lib.utils.stream_writer import StreamWriter
 from samcli.local.lambdafn.exceptions import FunctionNotFound
 from samcli.local.events.api_event import ContextIdentity, RequestContext, ApiGatewayLambdaEvent
@@ -207,7 +208,7 @@ class LocalApigwService(BaseLocalService):
             raise TypeError("Lambda returned %{s} instead of dict", type(json_output))
 
         status_code = json_output.get("statusCode") or 200
-        headers = CaseInsensitiveDict(json_output.get("headers") or {})
+        headers = Headers(json_output.get("headers") or {})
         body = json_output.get("body") or "no data"
         is_base_64_encoded = json_output.get("isBase64Encoded") or False
 
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -166,7 +166,7 @@ class LocalApigwService(BaseLocalService):
                                                                      route.binary_types,
                                                                      request)
         except (KeyError, TypeError, ValueError):
-            LOG.error("Function returned an invalid response (must include one of: body, headers or "
+            LOG.error("Function returned an invalid response (must include one of: body, headers, multiValueHeaders or "
                       "statusCode in the response object). Response received: %s", lambda_response)
             return ServiceErrorResponses.lambda_failure_response()
 
@@ -208,7 +208,8 @@ class LocalApigwService(BaseLocalService):
             raise TypeError("Lambda returned %{s} instead of dict", type(json_output))
 
         status_code = json_output.get("statusCode") or 200
-        headers = Headers(json_output.get("headers") or {})
+        headers = LocalApigwService._merge_response_headers(json_output.get("headers") or {},
+                                                            json_output.get("multiValueHeaders") or {})
         body = json_output.get("body") or "no data"
         is_base_64_encoded = json_output.get("isBase64Encoded") or False
 
@@ -245,6 +246,7 @@ class LocalApigwService(BaseLocalService):
             "statusCode",
             "body",
             "headers",
+            "multiValueHeaders",
             "isBase64Encoded"
         }
         # In Python 2.7, need to explicitly make the Dictionary keys into a set
@@ -262,7 +264,7 @@ class LocalApigwService(BaseLocalService):
             Corresponds to self.binary_types (aka. what is parsed from SAM Template
         flask_request flask.request
             Flask request
-        lamba_response_headers dict
+        lamba_response_headers werkzeug.datastructures.Headers
             Headers Lambda returns
         is_base_64_encoded bool
             True if the body is Base64 encoded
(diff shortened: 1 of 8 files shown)


dsanders11:
> Nothing should have changed to not make this true anymore, unless I am miss understanding the problem or solution.

@jfuss, I think you may have missed that `CaseSensitiveDict` was replaced with `Headers`, which also acts in a case-insensitive way. Prior to that PR the headers were being passed around as a simple `dict`. Basically `Headers` is a class from Flask which does the same thing as `CaseSensitiveDict` with a few extra bells and whistles (handles multi-value headers), and a few minor differences in functionality which I mentioned in the opening comment. When providing headers to a Flask `Response` it will turn the `dict` into a `Headers` object if it's not already.

jfuss:
>  I think you may have missed that CaseSensitiveDict was replaced with Headers, which also acts in a case-insensitive way.

@dsanders11 Yes, yes I did. I think that generally makes sense. Sorry for the oversight.

jfuss:
@dsanders11 I believe `CaseInsensitiveDict` is required. It was initially brought up here: #398. Nothing should have changed to not make this true anymore, unless I am miss understanding the problem or solution.

dsanders11:
I assume by the testdata directory you mean https://github.com/awslabs/aws-sam-cli/tree/develop/tests/integration/testdata/start_api? Taking a quick look, I see there's a `swagger-template.yaml` and a `template.yaml`. They look similar, but the `swagger-template.yaml` hasn't been touched in a year, and the test you linked refers to just `template.yaml`. Do you know if `swagger-template.yaml` is in use?

jfuss:
@dsanders11

> I assume by the testdata directory you mean https://github.com/awslabs/aws-sam-cli/tree/develop/tests/integration/testdata/start_api? 

Correct. That is the test data for that Test Class I linked.

The `swagger-template.yaml` is used in [these Test Cases](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/local/start_api/test_start_api.py#L180). The main difference between the `swagger-template.yaml` and `template.yaml` is how they define the API to test. The `swagger` uses [inline swagger](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/testdata/start_api/swagger-template.yaml#L16) while the `template` uses the [`Event` Api structure](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/testdata/start_api/template.yaml#L18). Since what you are adding is based on the output of the function and not how it is defined, which one you use doesn't really matter all that much. The Test Class I pointed to seemed like a good fit and is already bootstrapped with a template file. It is much easier to define the `template.yaml` than the `swagger-template.yaml`.

dsanders11:
@jfuss, thanks for the info. I'll try to get a test written there in the next couple days, and confirm the behavior of `Content-Type` on API Gateway.

Nevett:
This is an even bigger issue now, since v0.16.0 adds overly strict checks to the response, and kills any responses that contain `multiValueHeaders`. This includes any responses that come from [Amazon's own AspNetCoreServer](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.AspNetCoreServer).

See: https://github.com/awslabs/aws-sam-cli/blob/43a7c37/samcli/local/apigw/local_apigw_service.py#L242

johnjelinek:
@dsanders11: how's this looking? I'm blocked on testing golang lambda handlers until this is merged.

jfuss:
@dsanders11 I added integ tests and rebased on top of develop (required a force push sorry). Not fair to put you in the hot path due to something we just released.

@sanathkr, @awood45, or @TheSriram Can you take over reviewing as a different set of eyes.

dsanders11:
@jfuss, I was actually writing the integration test today (no big deal), but got sidetracked investigating the `Content-Type` behavior in API Gateway.

That led to me noticing that API Gateway doesn't seem to be behaving like the documentation says it should (and how this PR is written). Emphasis on the important part:

> [If you specify values for both headers and multiValueHeaders, API Gateway merges them into a single list. **If the same key-value pair is specified in both, only the values from multiValueHeaders will appear in the merged list.**](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format)

From my testing, here's what I'm seeing in the AWS Console when testing:

> Thu May 16 00:02:36 UTC 2019 : Endpoint response body before transformations: {"body":"aGVsbG8=","statusCode":200,"headers":{"X-Foo":"bar","Content-Type":"text/html"},"multiValueHeaders":{"Content-Type":["application/json"],"X-Foo":["foo"]},"isBase64Encoded":true}

> Thu May 16 00:02:36 UTC 2019 : Method response headers: {Content-Type=application/json,text/html, X-Foo=foo,bar, X-Amzn-Trace-Id=Root=1-5cdca899-c034cb57145c72f94b4d4592;Sampled=0}

Am I misreading the documentation, or is that not behaving like it says it should behave? It's merging the values from `headers` and `multiValueHeaders` into a combined list for the same header.

P.S. Regarding the breakage for some people, seems like you should be able to grab the very isolated change of allowing `multiValueHeaders` in the lambda response until this PR is done.

jfuss:
@dsanders11 My understanding from the docs is: the final header list is the set of headers and multiValueHeaders. I think what it’s trying to say is if multiValueHeaders also has text/html in the list (in your example), text/html would only appear once is the final headers that apigateway returns.

dsanders11:
@jfuss, that does seem to be the actual behavior on testing. Unfortunate that the documentation is ambiguous as both @martysweet and I interpreted it differently. I think it's the plural nature of 'only the values from multiValueHeaders will appear in the merged list.'

I've updated the code to act more like API Gateway does in actuality. This includes mimicking the order of header values when the same header appears in both `header` and `multiHeaderValues`, since this i important. From testing it appends the `header` value onto the end of the `multiHeaderValues` list.

The integration test will need to be updated to match this change.

Unfortunately I'm out of time to work on this anymore this week. I spent an hour trying to get base64 decoding working on API Gateway so I could confirm behavior, but was not able to do so despite seemingly following everything to a T. As such that part of this PR is still a question mark on if it exactly follows API Gateway.

jfuss:
@dsanders11 We really appreciate you (and @martysweet) diving into and ensuring this accurately reflects Api Gateway's behavior. I can take over the last bit here.

AyushG3112:
Is this the root cause of #1190? If yes, can this be merged and released soon as this is causing issues during our development.

PR #1186, opened 2019-05-17
Blackify some files to show style differences

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
This PR is meant to show what the differences would be if we started to use [Black](https://github.com/python/black). A good read about benefits and tradeoffs can be found under the [Django dep project (dep 0008)](https://github.com/django/deps/blob/master/accepted/0008-black.rst).

Happy to have a formal write up but wanted to show how this could look in our codebase through a small PR. If the team decides that this is worth while, I can work on formalize this with pre-commit hooks to ensure formatting is always done and therefore we should no longer need to care about formatting :). We could do something similar to how the [attrs](https://github.com/python-attrs/attrs) project is using Black.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


sanathkr:
Orange is the new black!

TheSriram:
Overall I like it. 👍

TheSriram:
Shall we adopt this? Given the line length is configurable, I'm okay with us taking this on.

jfuss:
I haven't heard a strict no yet. There is still some work to do here and we need to figure out a good time to do the flip. I am hoping to have this ready sometime early next week.

jfuss:
This should be ready to go.

PR includes:
* the current HEAD of development in formatted
* updated DEVELOPMENT_GUIDE with instructions on installing black and or pre-commit
* updated Makefile with a target to run the formatter
* travis running `black --check` to make sure source code is black compliant
* Integrated with pre-commit for those who want to run git hooks with black
* Integrated with `pyproject.toml` to hold our black config. More info in [PEP518](https://www.python.org/dev/peps/pep-0518/)

pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+requires = ["setuptools", "wheel"]  # PEP 508 specifications.
+
+
+[tool.black]
+line-length = 120
+target_version = ['py37', 'py27', 'py36']
+exclude = '''
+
+(
+  /(
```
sanathkr: Can you also exclude init template directories?

Makefile
```diff
@@ -26,5 +26,10 @@ lint:
 # Command to run everytime you make changes to verify everything works
 dev: flake lint test
 
+black-format:
+	black samcli/* tests/*
+
 # Verifications to run before sending a pull request
 pr: init dev
+
```
sanathkr: remove empty lines

Makefile
```diff
@@ -26,5 +26,10 @@ lint:
 # Command to run everytime you make changes to verify everything works
 dev: flake lint test
 
+black-format:
```
sanathkr: call it just `black`?

pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
+requires = ["setuptools", "wheel"]  # PEP 508 specifications.
+
+
+[tool.black]
+line-length = 120
+target_version = ['py37', 'py27', 'py36']
```
sanathkr: remove py27?

.travis.yml
```diff
@@ -39,12 +39,17 @@ before_install:
...
+  - chmod +x /tmp/black
+  - /tmp/black --version
 
 install:
   # Install the code requirements
   - make init
 
 script:
+  # Verify code is black compliant
+  - /tmp/black --check setup.py tests samcli
```
sanathkr: `make black` instead?

pyproject.toml
jfuss: good idea. sure can do.

Makefile
jfuss: done

Makefile
jfuss: sure

.travis.yml
jfuss: `make black` will format the files are write it back to the filesystem. Instead this commands `black --check` makes sure that the files are formatted and if not the build will fail. This ensures our codebase is always formatted. I can create a `make black-check`? command or something. I thought about putting it in `make pr` but didn't want to go through the hassle of pip installing on appveyor too (you can only [pip] install black in 3.6 or above).

pyproject.toml
jfuss: I kept it because we still have py27 support :(. So wasn't comfortable making formatting related this and risk breaking it. I just quickly ran it and the only changes were removing `u` from strings like `u"help"`. For now, I would prefer to just leave it.

pyproject.toml
jfuss: I have excluded the templates used for `sam init`. I will un-format them once, I do the update from develop and rerun black dance.

TheSriram:
I wonder if we should run `make black` on a github post hook or something and completely forget about formatting. it gets auto taken care of.

viksrivat:
If we run it as a github post hook, do we still need to run `make lint/flake` ?

jfuss:
| I wonder if we should run make black on a github post hook or something and completely forget about formatting. it gets auto taken care of.

@TheSriram That is what the `pre-commit` installation instructions are for. I didn't want to make it required. I don't like pip installing other tools, because it can mess with the global or user installation on a system. So I made this optional but travis does run a check to ensure the files are 'blackified', so we should never be checking in code that is not formatted. 

| If we run it as a github post hook, do we still need to run make lint/flake ?

@viksrivat I think we can remove flake8, but would like to keep pylint around until we know black handles everything.

DEVELOPMENT_GUIDE.md
```diff
@@ -36,7 +36,20 @@ easily setup multiple Python versions.
 5.  Make Python versions available in the project:
     `pyenv local 3.6.8 2.7.14 3.7.2`
 
-### 2. Activate Virtualenv
+### 2. Install Additional Tooling
+#### Black
+We format our code using [Black](https://github.com/python/black) and verify the source code is black compliant
+in Travis during PRs. You can find installation instructions on [Black's docs](https://black.readthedocs.io/en/stable/installation_and_usage.html).
```
TheSriram: Appveyor

samcli/commands/local/start_lambda/cli.py
```diff
@@ -51,33 +51,74 @@
...
-           debugger_path, docker_volume_basedir, docker_network, log_file, layer_cache_basedir, skip_pull_image,
-           force_image_build, parameter_overrides):
+    do_cli(
+        ctx,
+        host,
+        port,
+        template,
+        env_vars,
+        debug_port,
+        debug_args,
```
TheSriram: Not sure why these random empty lines show up here, github formatting?

Makefile
```diff
@@ -24,17 +24,15 @@ smoke-test:
...
 # Command to run everytime you make changes to verify everything works
-dev: flake lint test
+dev: lint test
+
+black:
+	black samcli/* tests/*
 
 # Verifications to run before sending a pull request
-pr: init dev
+pr: init dev
```
sanathkr: Why isn't `black` run as part of `make pr`?

jfuss: the `make black` command requires you to commit the changes. So I didn't want the `make pr` to run that and make it look like it works. I can add another `make` command that does the check instead, which is probably what we want anyways.

jfuss: If you install the git hooks, running the check on pr is just redundant. This is because you can't push unless black passes.

Issue #913, opened 2019-01-04
sam build --use-container fails hello_world when behind proxy.


DevOpsChris:
**Description:**

The hello_world example bombs everytime when following the written instructions when behind a proxy.  Need to create a proxy flag or introduce passing proxy envs with the build command. Better yet, honor the .docker/config.json and service settings in your container code.

Also, really need to make logging or stderr MUCH better in https://github.com/awslabs/aws-lambda-builders, because it just says 

```
{"jsonrpc": "2.0", "id": 1, "error": {"code": 400, "message": "PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: requests==2.20.0"}}
```

instead of 

```
Collecting foo
  Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x10c04e780>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known')': /simple/foo/
  Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x10c04e588>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known')': /simple/foo/
......
  Could not find a version that satisfies the requirement foo (from versions: )
No matching distribution found for foo

```

It really misleads people when debugging.

**Dirtiest Fix**, put in proxy envs in env dictionary https://github.com/DevOpsChris/aws-sam-cli/blob/develop/samcli/local/docker/lambda_build_container.py#L77-L81

**Best fix**: Honor the environment settings the docker service provides.

**Steps to reproduce the issue:**
1. Follow the 0.10.0 hello world instructions
2. `sam build --use-container`

**Observed result:**
```
aws_lambda_builders.exceptions.WorkflowFailedError: PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: requests==2.20.0
2019-01-04 13:37:35 Build inside container returned response {"jsonrpc": "2.0", "id": 1, "error": {"code": 400, "message": "PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: requests==2.20.0"}}
Build Failed
Error: PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: requests==2.20.0
```

**Expected result:**

```
[...]

PR #1196, opened 2019-05-24
Docker proxy config

cooper3330:
*Issue #, if available:*

#913 

*Description of changes:*

This updates the docker create command to configure the proxy settings inside the docker container using global docker configuration file.  This will allow users who sit behind a proxy and have configured their proxy settings in docker to pass through their proxy for things such as dependency installation (Pip, etc.).

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -99,7 +99,9 @@ class Container(object):
                 }
             },
             # We are not running an interactive shell here.
-            "tty": False
+            "tty": False,
+            # Set proxy configuration from global Docker config file
+            "use_config_proxy": True
         }
 
         if self._container_opts:
--- a/samcli/local/docker/utils.py
+++ b/samcli/local/docker/utils.py
@@ -3,8 +3,8 @@ Helper methods that aid with changing the mount path to unix style.
 """
 
 import os
-import posixpath
 import re
+import posixpath
 try:
     import pathlib
 except ImportError:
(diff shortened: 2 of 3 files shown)


TheSriram:
Thanks for the contribution! 💯 

On first pass this looks good. I have some follow up questions

* https://github.com/docker/docker-py/blob/e2c1c9caace22098f7030049f0ad2f6d3f4e0f0a/docker/api/container.py#L396 states that appropriate env vars will be injected if proxy information is found in config file, [Docker proxy](https://docs.docker.com/network/proxy/) states that this is supported in Docker 17.07 and higher, does that mean this is not supported in the world of docker toolbox users? (mac and windows)
  * EDIT: https://github.com/docker/toolbox/releases -> Looks like it might be.
* what is an effective way to test this?

cooper3330:
Testing this is an interesting case.  Since this is only modifying the Docker Container Create SDK call, I think we just need to test that the _container.create_ SDK call is kicked off with the expected parameter.  It is equivalent to including the _'tty=False'_ in the container create call.

To test these today, we are verifying the _Container Create_ call was kicked off with the correct parameters passed.  The proposed PR maintains that standard of testing with the underlying Docker SDKs by asserting _docker.create_ passed in the _use_config_proxy=True_ parameter.

Any issues/disagreement with that train of thought?

sanathkr:
@cooper3330 Your testing rationale makes sense.

I digged thru Docker-Py's recent commit of https://github.com/docker/docker-py/pull/2202 to understand the implementation. I am confident that this doesn't have any side effects until you set proxy config in global Docker configuration files.

But this feature is available in Docker SDK >= 3.7.0. We need to update the version in `requirements/base.txt`

sanathkr:
Bumping docker version separately - https://github.com/awslabs/aws-sam-cli/pull/1214 so we know this is safe before merging this.

Issue #925, opened 2019-01-10
Number not being honored when passed as parameter to sam local start-lambda


binarymist:
### Description

`--parameter-overrides` number not honored

### Steps to reproduce

Run:
`sam local start-lambda --parameter-overrides ParameterKey=TimeoutParameter,ParameterValue=3 --host 172.25.0.1 --docker-network compose_pt-net`

### Observed result
```
2019-01-11 00:28:51 local start_lambda command is called
2019-01-11 00:28:51 Collected default values for parameters: {'TimeoutParameter': 10}
2019-01-11 00:28:51 2 resources found in the template
2019-01-11 00:28:51 Found Serverless function with name='provisionAppSlaves' and CodeUri='./app-slave-provisioner'
2019-01-11 00:28:51 Found Serverless function with name='provisionSeleniumStandalones' and CodeUri='./selenium-standalone-provisioner'
2019-01-11 00:28:51 Trying paths: ['.docker/config.json', '.dockercfg']
2019-01-11 00:28:51 No config file found
2019-01-11 00:28:51 Trying paths: ['.docker/config.json', '.dockercfg']
2019-01-11 00:28:51 No config file found
2019-01-11 00:28:51 http://localhost:None "GET /v1.35/_ping HTTP/1.1" 200 2
2019-01-11 00:28:51 Trying paths: ['.docker/config.json', '.dockercfg']
2019-01-11 00:28:51 No config file found
2019-01-11 00:28:51 Trying paths: ['.docker/config.json', '.dockercfg']
2019-01-11 00:28:51 No config file found
2019-01-11 00:28:51 Starting the Local Lambda Service. You can now invoke your Lambda Functions defined in your template through the endpoint.
2019-01-11 00:28:51 Localhost server is starting up. Multi-threading = True
2019-01-11 00:28:51  * Running on http://172.25.0.1:3001/ (Press CTRL+C to quit)
```

When I invoke the lambda via the aws-sdk in code, sam local says:

```
2019-01-11 00:29:55 Found one Lambda function with name 'provisionAppSlaves'
2019-01-11 00:29:55 Found one Lambda function with name 'provisionSeleniumStandalones'
2019-01-11 00:29:55 Invoking index.provisionAppSlaves (nodejs8.10)
2019-01-11 00:29:55 Invoking index.provisionSeleniumStandalones (nodejs8.10)
2019-01-11 00:29:55 Environment variables overrides data is standard format
[...]

PR #1202, opened 2019-05-31
fix:  Lambda timeout from parameter (#925)

setrofim:
*Issue #, if available:*
https://github.com/awslabs/aws-sam-cli/issues/925

*Description of changes:*

This fixes the specific case used as the example in the issue of using  parameters to specify the Timeout for Lambdas. 

As per CloudFormation documentation (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html) the fact that the the Type of the parameter is ignored and the value is passed as a string appears to be expected behaviour.  Therefore, I'm doing the numeric conversion just before passing the value to the `Function`.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


TheSriram:
are you not able pass timeout as parameter override with the current model? and thats you are evaluating Timeout?

setrofim:
Please could you provide details of what broke? I tried testing `sam local invoke` on my setup, but it seems to be working fine for me.

setrofim:
Ah I see. Sorry, I misunderstood.

TheSriram:
Can we add an integ test for this as well, happy to merge it after its done.

setrofim:
I have added an integration test for this (and confirmed that it fails without the fix).

samcli/commands/local/lib/sam_function_provider.py
```diff
@@ -123,11 +124,15 @@ def _convert_sam_function_resource(name, resource_properties, layers):
 
         LOG.debug("Found Serverless function with name='%s' and CodeUri='%s'", name, codeuri)
 
+        timeout = resource_properties.get("Timeout")
+        if isinstance(timeout, str):
+            timeout = ast.literal_eval(timeout)
```
jfuss: Why ast here instead of doing an `int()` conversation?

In either case if the string can't be casted into a int, `ast.literal_eval` will throw `SyntaxError` and `int(timeout)` will throw `ValueError`. We should be catching and handling them gracefully here.

setrofim: I did an `ast.literal_eval` here  in case the value is a float (as the parameter type is `Number`). I shall catch and re-raise as `ValueError` -- thanks!

setrofim: I have made the change. It looks like now appveyor tests are failing, however, looking a the errors, I don't _believe_ it's related to my change? (Seems to be a permissions issue.)

setrofim:
@TheSriram  @jfuss is there anything else required from me for this?

TheSriram:
No, I dont think so, I re-kicked the build now.

samcli/commands/local/lib/sam_function_provider.py
```diff
@@ -123,11 +124,18 @@ def _convert_sam_function_resource(name, resource_properties, layers):
 
         LOG.debug("Found Serverless function with name='%s' and CodeUri='%s'", name, codeuri)
 
+        timeout = resource_properties.get("Timeout")
+        if isinstance(timeout, str):
+            try:
+                timeout = ast.literal_eval(timeout)
+            except SyntaxError:
+                raise ValueError('Invalid Number for Timeout: {}'.format(timeout))
```
jfuss: This will raise a ValueError up the chain and wont be caught. This will show up as a command failure with a stack track. We should raise a custom error/exception here to avoid that and provide a better customer experience.

setrofim: Is there an existing exception used for this? Is there an existing convention for how this is handled in the codeline?

jfuss: @setrofim You can follow what we are currently doing for Invalid Layer ARNs.

We throw [there](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/local/lib/sam_function_provider.py#L247) and the exception is defined [here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/local/cli_common/user_exceptions.py#L57)

There is currently an exception for `InvalidSamTemplateException`. This might be sufficient here with a good error message (what you currently have is fine). You can define a more specific to improve code readability but either works.

When defining exceptions, make sure the inherit from `UserException`. This will make sure the correct information is added to the exception so Click will work as expected (no stack trace and printing the error message to console). Only other this is to just update your current unit test to expect the new exception instead of ValueError.

setrofim: Sorry for the delayed response. I have now made the change -- I stuck with `InvalidSamTemplateException` as that see to cover this case.

setrofim:
Are there any outstanding issues with this?

PR #1215, opened 2019-06-07
Use function name instead of logical id local lambda

matrinox:
*Issue #, if available:*
https://github.com/awslabs/aws-sam-cli/issues/1025

*Description of changes:*
Uses `FunctionName` property (instead of logical ID) when:
1. invoking Lambda functions locally (with `sam local invoke`)
2. starting local Lambda server (with `sam local start-lambda`)

This mirrors the behaviour in AWS environments, where `FunctionName` takes precedent.

For example, given the following `template.yaml`:
```yaml
Resources:
  FooBar:
    Properties:
      FunctionName: foo-bar
      CodeUri: fooBar/
      Handler: index.handler
      Runtime: nodejs8.10

  BarFoo:
    Properties:
      CodeUri: barFoo/
      Handler: index.handler
      Runtime: nodejs8.10

  FooBarLambda:
    Type: AWS::Lambda::Function
    Properties:
      FunctionName: foo-bar-lambda
      Code: fooBar/
      Handler: index.handler
      Runtime: nodejs8.10
```

It can be invoked with the sam CLI with (in order of `template.yaml`):
```sh
sam local invoke foo-bar --event event.json
sam local invoke BarFoo --event event.json
sam local invoke foo-bar-lambda --event event.json
```

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/local/lib/local_lambda.py
```diff
@@ -75,10 +75,13 @@ def invoke(self, function_name, event, stdout=None, stderr=None):
         """
 
         # Generate the correct configuration based on given inputs
-        function = self.provider.get(function_name)
+        function = None
+        for f in self.provider.get_all():
+            if f.functionname == function_name:
+                function = f
```
jfuss: Not sure I understand the move from `.get(function_name)` to `.get_all()`. We should still be able to fetch the function based on the name here right? I think the decision of logical id or function name can be determined directly in the `.get(function_name)` call on the provider.

matrinox: I was worried about changing too much but I think you're approach is correct

matrinox:
@jfuss I've addressed your comments and added the integration tests. Let me know if it's up to par

floydding:
Thanks for the change, this is going to be extremely helpful! looking forward to the release of this change!

PR #1234, opened 2019-06-18
(feat) CloudFormation ApiGateway support design

viksrivat:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,377 @@
...
+        RequestTemplates:
+          applicationjson: "{\"statusCode\": $input.json('$.statusCode'), \"message\": $input.json('$.message')}"
+
+```
+
+The SAM template requires people to use only a single RestApi to define all of their resources.
+
+### **AWS::ApiGateway::Stages**
+
+The stage allows for defining environments for different parts of the pipeline in CloudFormation. Since a single stage is attached to a deployment, the behavior of multiple deployment/stages locally is currently undefined, which is similar to the current support or syntax in the SAM template.
```
jfuss: "the behavior of multiple deployment/stages locally is currently undefined, which is similar to the current support or syntax in the SAM template."

Kind of. Since AWS::Serverless::Api only allows one stage, we never have to pick if other stages are in the template.

designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,377 @@
...
+  Description: Prod Stage
+  Variables:
+    Stack: Prod
+  RestApiId: RestApi`
+```
+
+Since the stages can show up in any order as the code is iterating through the resources, There are two approaches to consider too quickly stitch the api and stages together. 
+
+*Approach #1:* 
+Arbitrarily pick one of the stages and use that on all the Api
```
jfuss: Picking arbitrarily is fine but we should make sure it is reproducible. Meaning if the customers gives the same template in two different command runs, the same stage should be selected, otherwise this could become frustrating to customers.

I think Arbitrarily here means we pick the first one in the template we find?

designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,377 @@
...
+
+I recommend approach #1 as it follows the current standard and there is some data that very little people use the <Stage Name> when defining it. 
+
+**AWS::ApiGateway::Deployment **
+The deployment type defines which Apis objects should be available to the public. Normally, in the SAM template everything defined will have a deployment component, but the CloudFormation templates may have more bloat and information in them defining certain apis and deployments for Multi-stages.
+
+One approach is to support multiple stages with multiple deployments and the code will filter out the stages that are not deployed or defined. One problem with this approach is that the RestApiId, a characteristic of the RestApi, for a deployment is only localhost and a single endpoint and the current schema doesn't allow deploying on multiple ports and multiple local domains. This will bring in unneeded complexity such as requiring the customer defining N valid ports and N valid domains.
+
+Another approach, which I recommend, is to ignore the flag altogether as it always associated with a stage and to randomly pick a stage.
```
jfuss: "ignore the flag" Deployment is not a flag, it is a resource

designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,377 @@
...
+
+One approach is to support multiple stages with multiple deployments and the code will filter out the stages that are not deployed or defined. One problem with this approach is that the RestApiId, a characteristic of the RestApi, for a deployment is only localhost and a single endpoint and the current schema doesn't allow deploying on multiple ports and multiple local domains. This will bring in unneeded complexity such as requiring the customer defining N valid ports and N valid domains.
+
+Another approach, which I recommend, is to ignore the flag altogether as it always associated with a stage and to randomly pick a stage. 
+
+## **Updating the code**
+
+First, the Apis in AWS::ApiGateway::RestApi  will be parsed in the SAM Template in a similar way to the current AWS::Serverless:Api code. Stages and other Meta information in the Api will be parsed to support the CloudFormation template. The code will then be refactored to first translate the SAM templates inputted into the CloudFormation template using the SamTranslator. Once the Api code has been refactored, the function code can be refactored so that Serverless code is refactored out.
```
jfuss: Not now but when you get to doing the Transformation to CFN, we can expand this doc or create a new one. I think there is enough interesting tradeoffs and decisions in that to warrant it.

designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,377 @@
...
+
+**Are you connecting to a remote API? If so explain how is this connection secured**
+No.
+
+**Are you reading/writing to a temporary folder? If so, what is this used for and when do you clean up?**
+The setup is not written for local development.
+
+# What is your Testing Plan (QA)?
+
+## Validation
```
jfuss: Validation isn't in scope right? Is this section needed then?

designs/cloud_formation_api_gateway_support.md
```diff
@@ -0,0 +1,371 @@
...
+        sam_template_data = self._get_template_data(self._template_file)
+        self._template_dict = self._get_clouformation_data(sam_template_data)
+        self._function_provider = CloudFormationFunctionProvider(self._template_dict,self.parameter_overrides
+    
+    # translate the template data to cloudformation template and return 
+    def _get_clouformation_data():
+        sam_translator = Translator(managed_policy_map=self.__managed_policy_map(),sam_parser=sam_parser, plugins=self.extra_plugins)
+        return sam_translator.translate(sam_template=template_copy, parameter_values=parameter_values)
+
+class CloudFormationFunctionProvider(SamBaseProvider):
```
jfuss: This doesn't parse the Method and resources right? That happens within the CfnProvider.

Can you update this section to match the implementation?

PR #1238, opened 2019-06-20
(feat) CloudFormation AWS::ApiGateway::RestApi support

viksrivat:
*Issue #, if available:*
Related pr: #1234 
*Description of changes:*
Add support for the CloudFormation AWS::ApiGateway::RestApi type for swagger.
*Checklist:*

- [x] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [x] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [x] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1239, opened 2019-06-24
(feat) CloudFormation AWS::ApiGateway::Stage Support

viksrivat:
*Issue #, if available:*

*Description of changes:*
Design Doc: #1234
AWS:ApiGateway::RestApi pr: #1238

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/route_collector.py
```diff
@@ -0,0 +1,59 @@
...
+from collections import defaultdict
+
+LOG = logging.getLogger(__name__)
+
+
+class RouteCollector(object):
+
+    def __init__(self):
+        # Route properties stored per resource.
+        self.by_resource = defaultdict(list)
```
jfuss: It is not clear what `by_resource` means here. I know this is what it was before, but this might be a good time to find a new name.

Also, since this is an iterator, I think this should be a private variable to the class. Unless I am missing a use case for directly accessing this var.

samcli/commands/local/lib/route_collector.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __iter__(self):
+        """
+        Iterator to iterate through all the routes stored in the collector. In each iteration, this yields the
+        LogicalId of the route resource and a list of routes available in this resource.
+        Yields
+        -------
+        str
+            LogicalID of the AWS::Serverless::Api resource
```
jfuss: `AWS::Serverless::Api` is no longer true :)

samcli/commands/local/lib/swagger/parser.py
```diff
@@ -62,15 +62,13 @@ def get_apis(self):
...
         -------
-        list of samcli.commands.local.lib.provider.Api
+        list of list of samcli.commands.local.apigw.local_apigw_service.Route
             List of APIs that are configured in the Swagger document
         """
 
         result = []
         paths_dict = self.swagger.get("paths", {})
 
-        binary_media_types = self.get_binary_media_types()
```
jfuss: Are we not reading binary media types out of the swagger?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -18,35 +18,85 @@
...
-        self.methods = methods
+        self.method = method.upper()
         self.function_name = function_name
         self.path = path
-        self.binary_types = binary_types or []
-        self.stage_name = stage_name
-        self.stage_variables = stage_variables
+
+    @staticmethod
+    def normalize_routes(routes):
```
jfuss: Not sure I understand the reason for moving the normalization methods into the Route. Previously the Route was a data holder but now a caller needs to understand which method to call? I think these belong up the call stack and outside of this class (similar to how it was before). But maybe I am missing the need for it here specifically.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -18,35 +18,85 @@
...
+
+    def __eq__(self, other):
+        return isinstance(other, Route) and \
+               self.method == other.method and self.function_name == other.function_name and self.path == other.path
+
+    def __hash__(self):
+        return hash(self.function_name) * hash(self.path) * hash(self.method)
+
+    @staticmethod
+    def normalize_http_methods(http_method):
```
jfuss: `ANY` is an API Gateway specific 'verb'. I don't think this belongs in the Route class. The Route is meant to be generic and follow valid HTTP routes/methods.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -18,35 +18,85 @@
...
+                        "POST",
+                        "HEAD",
+                        "OPTIONS",
+                        "PATCH"]
+
+    def __init__(self, function_name, path, method):
         """
         Creates an ApiGatewayRoute
 
         :param list(str) methods: List of HTTP Methods
```
jfuss: update? But I think this should still be a list.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -77,16 +127,15 @@ def create(self):
...
-                                                       path):
-                self._dict_of_routes[route_key] = api_gateway_route
+            route_key = self._route_key(api_gateway_route.method, path)
+            self._dict_of_routes[route_key] = api_gateway_route
 
             self._app.add_url_rule(path,
                                    endpoint=path,
                                    view_func=self._request_handler,
-                                   methods=api_gateway_route.methods,
+                                   methods=[api_gateway_route.method],
```
jfuss: Why change this from a list? I think the Routes should be modeled with a list of methods to a given function. This will help us avoid some of how flask resolves routes: https://github.com/awslabs/aws-sam-cli/issues/400#issuecomment-389397491

tests/integration/testdata/start_api/swagger-rest-api-template.yaml
```diff
@@ -61,6 +61,13 @@ Resources:
           - image/gif
       StageName: prod
     Type: AWS::ApiGateway::RestApi
+  Dev:
+    Type: AWS::ApiGateway::Stage
+    Properties:
+      StageName: Dev
+      RestApiId: !Ref MyApi
+      Variables:
+      Stack: Dev
```
jfuss: This should be indented under `Variables` right?

tests/integration/testdata/start_api/swagger-rest-api-template.yaml
```diff
@@ -61,6 +61,13 @@ Resources:
           - image/gif
       StageName: prod
     Type: AWS::ApiGateway::RestApi
+  Dev:
+    Type: AWS::ApiGateway::Stage
```
jfuss: Is this used in the integ tests? I don't see any tests that were updated or added that use this template.

tests/unit/commands/local/lib/test_cfn_api_provider.py
```diff
@@ -200,16 +199,217 @@ def test_with_binary_media_types_in_swagger_and_on_resource(self):
...
+        self.assertIn(route1, provider.routes)
+        self.assertEquals(provider.api.stage_name, "dev")
+        self.assertEquals(provider.api.stage_variables, {
+            "vis": "data",
+            "random": "test",
+            "foo": "bar"
+        })
+
+    def test_multi_stage_get_all(self):
+        template = OrderedDict({
```
jfuss: Can we just statically define this like the other tests? This is much harder to read and understand what the content of `template` is. Also remains consistent with all the other tests we have here.

tests/unit/commands/local/lib/test_local_api_service.py
```diff
@@ -80,50 +76,27 @@ def test_must_start_service(self,
...
         local_service = LocalApiService(self.lambda_invoke_context_mock, self.port, self.host, self.static_dir)
-
+        local_service.api_provider.api.routes = routing_list
         with self.assertRaises(NoApisDefined):
             local_service.start()
 
 
-class TestLocalApiService_make_routing_list(TestCase):
-
-    def test_must_return_routing_list_from_apis(self):
```
jfuss: Why was this tests removed? Do we have similar coverage if this was refactored out?

tests/unit/commands/local/lib/test_local_api_service.py
```diff
@@ -181,39 +154,3 @@ def test_must_return_none_if_path_not_exists(self, os_mock):
 
         result = LocalApiService._make_static_dir_path(cwd, static_dir)
         self.assertIsNone(result)
-
-
-class TestRoutingList(TestCase):
```
jfuss: Similar here. Why was this removed? Do we still have this covered in other areas.

tests/unit/commands/local/lib/test_sam_api_provider.py
```diff
@@ -855,27 +839,21 @@ def test_both_apis_must_get_binary_media_types(self):
 
         # Because of Globals, binary types will be concatenated on the explicit API
         expected_explicit_binary_types = ["explicit/type1", "explicit/type2", "image/gif", "image/png"]
-        expected_implicit_binary_types = ["image/gif", "image/png"]
+        # expected_implicit_binary_types = ["image/gif", "image/png"]
```
jfuss: This is commented out but is be asserted below?

tests/unit/commands/local/lib/test_sam_api_provider.py
```diff
@@ -1013,125 +985,119 @@ def test_provider_stage_variables(self):
...
-                                            "type": "aws_proxy",
-                                            "uri": {
-                                                "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31"
-                                                           "/functions/${NoApiEventFunction.Arn}/invocations",
-                                            },
-                                            "responses": {},
-                                        },
-                                    }
-                                }
+        template = OrderedDict({
```
jfuss: define statically

tests/unit/commands/local/lib/test_local_api_service.py
viksrivat: An Api is now defined as a list of routes and it's attributes. All the function was doing was converting a list of Apis -> list of Routes, which isn't needed in the current model as it is directly passed through.

samcli/commands/local/lib/swagger/parser.py
viksrivat: We are still reading the binary_media_types inside the swagger, but in a different way. Before, each route had the binary_media_types information instead of keeping a global copy. When we are extracting the swagger route in cfn_base_api_provider, we add it to the Api object's global copy.

tests/unit/commands/local/lib/test_local_api_service.py
viksrivat: Similar reasoning to the other comment

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -66,4 +74,23 @@ def _extract_cloud_formation_api(self, logical_id, api_resource, collector, cwd=
...
+       ----------
+        api_resource : dict
+            Resource definition, including its properties
+        api: samcli.commands.local.lib.provider.Api
+            Resource definition, including its properties
+        """
+        properties = api_resource.get("Properties", {})
+        stage_name = properties.get("StageName")
+        stage_variables = properties.get("Variables")
+        logical_id = properties.get("RestApiId")
```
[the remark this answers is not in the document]
viksrivat: The only other case to worry about I believe is dealing with intrinsic with Ref, which will be fixed in the other pr. Other than that, I'm not sure if I need to check if the logical_id is in the template and bubble up the error if so.

tests/unit/commands/local/lib/test_sam_api_provider.py
viksrivat: The problem with defining with statically is that when we parse the template, we use the order of the template to determine which stage to process. In python3 by default, dictionaries have order ,but in python2, there is no guarantee.

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -159,55 +170,68 @@ def _convert_event_api(lambda_logical_id, event_properties):
...
+        for route in result:
+            key = route.function_name + route.path
+            if key in merged_route_methods:
+                new_methods = list(set(merged_route_methods[key].methods + route.methods))
+                merged_route_methods[key].methods = new_methods
+            else:
+                merged_route_methods[key] = route
 
+        result = list(merged_route_methods.values())
         return list(result)
```
jfuss: You are doing a list of a list here. This can be combined with the line above.

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -159,55 +170,68 @@ def _convert_event_api(lambda_logical_id, event_properties):
...
+        merged_route_methods = {}
+        for route in result:
+            key = route.function_name + route.path
+            if key in merged_route_methods:
+                new_methods = list(set(merged_route_methods[key].methods + route.methods))
+                merged_route_methods[key].methods = new_methods
+            else:
+                merged_route_methods[key] = route
 
+        result = list(merged_route_methods.values())
```
jfuss: Not sure this belongs here. This method is to combine implicit and explicit apis. This changes the scope of the funciton. While this isn't wrong, it doesn't make the code as readable as it could. For example, we already do this function name + path when we print out routes in local_api_service.py. We should find a better how to do this deduping.

samcli/commands/local/lib/provider.py
```diff
@@ -265,35 +257,31 @@ def normalize_http_methods(http_method):
...
         """
-
-        result = list()
-        for api in apis:
-            for normalized_method in AbstractApiProvider.normalize_http_methods(api.method):
-                # _replace returns a copy of the namedtuple. This is the official way of creating copies of namedtuple
-                result.append(api._replace(method=normalized_method))
-
-        return result
+        return Route(function_name, path, methods=AbstractApiProvider.normalize_http_methods(method))
```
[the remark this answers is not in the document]
viksrivat: Before, each Api had a single method with the format Api(method="post"), Api(method="get"), which was then simplified later in the pipeline into Route(methods=["post","get"]). In the context of this function, all it's doing is converting ApiGateway specific types like "Any" into a list of methods within a single methods.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -18,35 +18,63 @@
...
-        self.stage_variables = stage_variables
+
+    def __eq__(self, other):
+        return isinstance(other, Route) and \
+               sorted(self.methods) == sorted(
+            other.methods) and self.function_name == other.function_name and self.path == other.path
+
+    def __hash__(self):
+        route_hash = hash(self.function_name) * hash(self.path)
+        for method in self.methods:
```
jfuss: order matters here. If we really need hash, we should sort methods.

samcli/commands/local/lib/cfn_base_api_provider.py
```diff
@@ -1,39 +1,44 @@
...
 from samcli.commands.local.lib.swagger.parser import SwaggerParser
-from samcli.commands.local.lib.swagger.reader import SamSwaggerReader
+from samcli.commands.local.lib.swagger.reader import SwaggerReader
 
 LOG = logging.getLogger(__name__)
 
 
 class CfnBaseApiProvider(object):
     RESOURCE_TYPE = "Type"
-
```
jfuss: This isn't needed anymore right?

tests/unit/commands/local/lib/test_api_provider.py
```diff
@@ -3,25 +3,26 @@
...
 
+from samcli.commands.local.lib.provider import Api
 from samcli.commands.local.lib.api_provider import ApiProvider
 from samcli.commands.local.lib.sam_api_provider import SamApiProvider
 from samcli.commands.local.lib.cfn_api_provider import CfnApiProvider
 
 
 class TestApiProvider_init(TestCase):
 
-    @patch.object(ApiProvider, "_extract_apis")
```
jfuss: remove print statements.

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -590,3 +586,23 @@ def test_should_base64_encode_returns_true(self, test_case_name, binary_types, m
...
+
+
+class TestRouteEqualsHash(TestCase):
+
+    def test_route_in_list(self):
+        route = Route(function_name="test", path="/test", methods=["POST"])
+        routes = [route]
+        self.assertIn(route, routes)
+
+    def test_route_method_order(self):
```
jfuss: You should have eq tests for same object different data, same object, two objects same data, two objects different data, two different objects/types.

PR #1242, opened 2019-06-26
(feat) Cors Support

viksrivat:
*Issue #, if available:*

*Description of changes:*
Adds Cors support to ApiGateway. This completes the #1009 pr

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -146,9 +146,43 @@ def _extract_from_serverless_api(self, logical_id, api_resource, collector):
...
+        ----------
+        properties : dict
+            Resource properties
+        """
+        cors_prop = properties.get("Cors")
+        cors = None
+        if cors_prop and isinstance(cors_prop, dict):
+            allow_methods = cors_prop.get("AllowMethods", ','.join(SamApiProvider._ANY_HTTP_METHODS))
+
+            if allow_methods and "OPTIONS" not in allow_methods:
```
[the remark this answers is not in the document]
viksrivat: The browser will do the validation and return the error if the incorrect values are set. We don't have to bother with checking. 

This is also true with post requests that require Access-Control-Allow-Methods to be *

jfuss:
@viksrivat Can you handle the merge conflicts here? I will do a quick pass over once you do that.

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -86,6 +89,41 @@ def _extract_from_serverless_api(self, logical_id, api_resource, collector, cwd=
...
+                allow_methods += ",OPTIONS"
+
+            cors = Cors(
+                allow_origin=cors_prop.get("AllowOrigin"),
+                allow_methods=allow_methods,
+                allow_headers=cors_prop.get("AllowHeaders"),
+                max_age=cors_prop.get("MaxAge")
+            )
+        elif cors_prop and isinstance(cors_prop, string_types):
```
jfuss: If cors in not a dict, then it must be a string right? we dont accept anything else, so do we need this if else?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -142,10 +144,15 @@ def _request_handler(self, **kwargs):
         Response object
         """
         route = self._get_current_route(request)
+        cors_headers = LocalApigwService.cors_to_headers(route.cors)
+
+        if 'OPTIONS' in route.methods:
```
jfuss: is this right? Shouldn't we only be doing this if the request was an Option request instead of if the route has a method of options?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -455,3 +468,26 @@ def _should_base64_encode(binary_types, request_mimetype):
...
+        ----------
+        cors list(samcli.commands.local.lib.provider.Cors)
+            CORS configuration objcet
+        Returns
+        -------
+            Dictionary with CORS headers
+        """
+        if not cors:
+            return {}
+        headers = {
```
jfuss: You could encapsulate this into the cors object.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -455,3 +468,26 @@ def _should_base64_encode(binary_types, request_mimetype):
...
+            'Access-Control-Allow-Origin': cors.allow_origin,
+            'Access-Control-Allow-Methods': cors.allow_methods,
+            'Access-Control-Allow-Headers': cors.allow_headers,
+            'Access-Control-Max-Age': cors.max_age
+        }
+
+        return {h_key: h_value for h_key, h_value in headers.items() if h_value is not None}
```
jfuss: What is this doing? This is hard for me to read.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -142,10 +144,15 @@ def _request_handler(self, **kwargs):
         Response object
         """
         route = self._get_current_route(request)
+        cors_headers = LocalApigwService.cors_to_headers(route.cors)
```
jfuss: If `cors_to_headers` returned an empty dict instead of None, then all the if checks on cors_headers go away and makes the code cleaner. Avoiding None unless a specific reason is not a bad practice to have. :)

samcli/local/apigw/local_apigw_service.py
jfuss: Can you answer the above question? This seems wrong when reading but could be just missing something obvious

samcli/local/apigw/local_apigw_service.py
```diff
@@ -404,7 +413,7 @@ def _query_string_params(flask_request):
         return query_string_dict, multi_value_query_string_dict
 
     @staticmethod
-    def _event_headers(flask_request, port):
+    def _event_headers(flask_request, port, cors_headers=None):
```
jfuss: Why not just make this required? the one place that calls it passes it

samcli/local/apigw/local_apigw_service.py
viksrivat: It's taking the cors_headers dict and filtering out properties from Cors object that don't exist.

samcli/commands/local/lib/sam_api_provider.py
viksrivat: The other case is the default case of no cors.

samcli/commands/local/lib/provider.py
```diff
@@ -242,3 +275,12 @@ def get_all(self):
         :yields Api: namedtuple containing the API information
         """
         raise NotImplementedError("not implemented")
+
+    @staticmethod
+    def normalize_http_methods(api):
```
jfuss: Was this a rebase mistake?

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -77,17 +80,51 @@ def _extract_from_serverless_api(self, logical_id, api_resource, collector, cwd=
...
+        Parameters
+        ----------
+        cors_prop : dict
+            Resource properties for Cors
+        """
+        cors = None
+        if cors_prop and isinstance(cors_prop, dict):
+            allow_methods = cors_prop.get("AllowMethods", ','.join(Route.ANY_HTTP_METHODS))
+
+            if allow_methods and "OPTIONS" not in allow_methods and "options" not in allow_methods:
```
jfuss: You need to normalize `allow_methods` here, just comparing OPTIONS and options is not enough, unless SAM or API gateway only allow these two ways for the strings.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -169,10 +170,16 @@ def _request_handler(self, **kwargs):
...
+
+        method, _ = self.get_request_methods_endpoints(request)
+        if method == 'OPTIONS':
+            headers = Headers(cors_headers)
+            return self.service_response('', headers, 200)
 
         try:
             event = self._construct_event(request, self.port, self.api.binary_media_types, self.api.stage_name,
-                                          self.api.stage_variables)
+                                          self.api.stage_variables, cors_headers)
```
jfuss: Is this right? Cors_headers always get added to the event?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -461,7 +481,9 @@ def _event_headers(flask_request, port):
 
         headers_dict["X-Forwarded-Port"] = str(port)
         multi_value_headers_dict["X-Forwarded-Port"] = [str(port)]
-
+        if cors_headers:
+            headers_dict.update(cors_headers)
+            multi_value_headers_dict.update(cors_headers)
```
jfuss: multi_value_headers value is a list, so this breaks that.

samcli/local/apigw/local_apigw_service.py
viksrivat: Yes. I believe CORS needs to be added to every request.

samcli/local/apigw/local_apigw_service.py
viksrivat: I don't think they are lists. They are dictionaries defined in _event_headers() as        
headers_dict = {}
multi_value_headers_dict = {}

TheSriram:
Is this branch ready to merge to candidate 20 branch? I see there are new commits added after the approval.

viksrivat:
The only changes that I added were updating some of the integration tests for Travis CI to pass.

jfuss:
Updated this to track and merge into develop.

jfuss:
@viksrivat Can you fix the flake8 errors:
`tests/unit/commands/local/lib/test_sam_api_provider.py:1065:9: E303 too many blank lines (2)`

Running `make pr` locally and making sure it passes should cover unit and linting.

PR #1247, opened 2019-06-29
(feat) CloudFormation AWS::ApiGateway::Methods and AWS::ApiGateway::Resources support

viksrivat:
*Issue #, if available:*

*Description of changes:*
Extends of #1239 to support AWS::ApiGateway::Resources
*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
 
         collector.stage_name = stage_name
         collector.stage_variables = stage_variables
+
+    def _extract_cloud_formation_method(self, resources, logical_id, api_resource, collector):
```
jfuss: This is a method_resource not an api_resource right?

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
...
+        """
+        Extract APIs from AWS::ApiGateway::Method and work backwards up the tree to resolve and find the true path.
+
+        Parameters
+        ----------
+        logical_id : str
```
jfuss: nit: keep the order they are defined in the function.

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
...
+        collector : ApiCollector
+            Instance of the API collector that where we will save the API information
+        """
+
+        properties = api_resource.get("Properties", {})
+        resource_id = properties.get("ResourceId")
+        rest_api_id = properties.get("RestApiId")
+        method = properties.get("HttpMethod")
+
+        resource_path = ""
```
jfuss: Would it make sense to default this to "/"?

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
...
+        if isinstance(resource_id, string_types):
+            resource = resources.get(resource_id)
```
jfuss: It would help in being more explicit about what resource this is. It took me awhile to understand what was going on here. Maybe a comment above the `if isinstance` would be helpful to.

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
...
+            if resource:
+                resource_path = self.resolve_resource_path(resources, resource, "")
+            else:
+                resource_path = resource_id  # In this case, the path was resolved to a string
```
jfuss: Is this comment right? This line gets executed if resource is falsey, not that the path was resolved to a string. This assigns the resource_path to be the resource_id, but should be the root path then right? Sorry having a hard time following this path right now.

samcli/commands/local/lib/cfn_api_provider.py
```diff
@@ -103,3 +120,102 @@ def _extract_cloud_formation_stage(resources, stage_resource, collector):
...
+        resource : dict
+            AWS::ApiGateway::Resource definition and its properties
+
+        current_path : str
+            Current path resolved so far
+        """
+
+        properties = resource.get("Properties", {})
+        parent_id = properties.get("ParentId")
+        path_part = properties.get("PathPart")
```
jfuss: At first I thought path_part was the separator (aka "/"), I know this matches what the API Gateway definition is but consider naming this to be `path` or `resource_path`. Maybe `path_part` does make sense and I am just the one jumping to conclusions.

samcli/commands/local/lib/route_collector.py
```diff
@@ -0,0 +1,59 @@
...
+routes in a standardized format
+"""
+
+import logging
+from collections import defaultdict
+
+LOG = logging.getLogger(__name__)
+
+
+class RouteCollector(object):
```
jfuss: Was this a merge/rebase mistake? It doesn't appear that RouteCollector is being used.

tests/integration/local/start_api/test_start_api.py
```diff
@@ -683,4 +682,110 @@ def test_global_stage_variables(self):
 
         response_data = response.json()
 
-        self.assertEquals(response_data.get("stageVariables"), {"Stack": "Dev"})
+        self.assertEquals(response_data.get("stageVariables"), {'Stack': 'Dev'})
+
+
+class TestStartApiWithMethodsAndResources(StartApiIntegBaseClass):
```
jfuss: Should we have any tests with `{+proxy}` or whatever the way you define a proxy route? What about a route that has a path variable in it?

tests/integration/testdata/start_api/methods-resources-api-template.yaml
```diff
@@ -0,0 +1,127 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
jfuss: I assume once the intrinsics PR gets merged and integrated, you will add some testdata and integ tests for resource and methods that are defined with intrinsics?

tests/integration/testdata/start_api/methods-resources-api-template.yaml
```diff
@@ -0,0 +1,127 @@
...
+      PathPart: "nofunctionfound"
+      ParentId: "RootApiResource"
+      ResourceId: "TestApi"
+  NoFunctionFoundMethod:
+    Type: "AWS::ApiGateway::Method"
+    Properties:
+      HttpMethod: "GET"
+      Integration:
+        Uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${WhatFunction.Arn}/invocations
```
jfuss: What if the Uri is defined just by a string instead of an Intrinsic Fn::Sub?

tests/integration/testdata/start_api/methods-resources-api-template.yaml
viksrivat: Ye. A lot of the code resource and methods is based on intrinsic references. When that is merged, all the tests will be updated to use intrinsics.

samcli/commands/local/lib/cfn_api_provider.py
viksrivat: I updated the comment to reflect it. This comment is referring to the case that is a ref of a raw resolved string such as  { "Fn::GetAtt": ["MyRestApi", "RootResourceId"] }

samcli/commands/local/lib/cfn_api_provider.py
viksrivat: The if instance can be removed once we merge in the intrinsic pr since the Refs will resolve to a string.

tests/integration/testdata/start_api/methods-resources-api-template.yaml
viksrivat: It doesn't really matter because all properties will be resolved using intrinsics

jfuss:
@viksrivat Looks like `tests.integration.local.start_api.test_start_api.TestStartApiWithMethodsAndResources testMethod=test_binary_request` is failing on all python versions. Can you please address the failing test?

PR #1260, opened 2019-07-09
Intriniscs Design Doc

viksrivat:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


designs/intrinsics_design.md
```diff
@@ -0,0 +1,391 @@
...
+
+    processed_resource = self.intrinsic_property_resolver(val)
+    processed_template[processed_key] = processed_resource
+```
+
+If there is an error with the resource, the item will be ignored and processed.
+
+## Integration into SAM-CLI
+
+This can be very easily plugged into the current SAM-CLI code. This needs to be run after the SamBaseTranslator runs in the ApiProvider. This will go through every property and check if it requires intrinsic resolution.
```
jfuss: `SamBaseTranslator` runs in any provider not just the ApiProvider.

designs/intrinsics_design.md
```diff
@@ -0,0 +1,391 @@
...
...ast-1d", "us-east-1e", "us-east-1f"],           "us-west-1": ["us-west-1b", "us-west-1c"],           "eu-north-1": ["eu-north-1a", "eu-north-1b", "eu-north-1c"],           "ap-northeast-3": ["ap-northeast-3a"],           "ap-northeast-2": ["ap-northeast-2a", "ap-northeast-2b", "ap-northeast-2c"],           "ap-northeast-1": ["ap-northeast-1a", "ap-northeast-1c", "ap-northeast-1d"],           "sa-east-1": ["sa-east-1a", "sa-east-1c"],           "ap-southeast-1": ["ap-southeast-1a", "ap-southeast-1b", "ap-southeast-1c"],           "ca-central-1": ["ca-central-1a", "ca-central-1b"],           "ap-southeast-2": ["ap-southeast-2a", "ap-southeast-2b", "ap-southeast-2c"],           "us-west-2": ["us-west-2a", "us-west-2b", "us-west-2c", "us-west-2d"],           "us-east-2": ["us-east-2a", "us-east-2b", "us-east-2c"],           "ap-south-1": ["ap-south-1a", "ap-south-1b", "ap-south-1c"],           "eu-central-1": ["eu-central-1a", "eu-central-1b", "eu-central-1c"],           "eu-west-1": ["eu-west-1a", "eu-west-1b", "eu-west-1c"],           "eu-west-2": ["eu-west-2a", "eu-west-2b", "eu-west-2c"],           "eu-west-3": ["eu-west-3a", "eu-west-3b", "eu-west-3c"],           "cn-north-1": []}
```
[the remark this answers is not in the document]
viksrivat: I'm not sure if there is a nicer or automatic way to deal with this other than updating it as new items are added. Is there a source of truth url I can look at?

designs/intrinsics_design.md
viksrivat: I was talking about how it can be integrated into the current codebase where this can be added into the ApiProvider

designs/intrinsics_design.md
```diff
@@ -0,0 +1,391 @@
...
+
+Different Types have different paramaters attached with them. This feels like it needs to be hardcoded for each one.
+
+### `Fn::Sub`
+
+```
+{ "Fn::Sub" : String } or { "Fn::Sub" : [string, "{test}", "{test2}"] }
+```
+
+This intrinsic does string substitution. Both types are supported. For the string, Regex will be used to replace the property of the string with the variables that are refs or pseudo refs such as ${AWS::RegionName}. This same process is done with the list, but every item in the list is recursively resolved and then approached the same way.
```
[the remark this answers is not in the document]
viksrivat: We are already using regex when dealing with Fn::Sub and when looking at getting the integration_uri. I don't think string formatting handles ${}

designs/intrinsics_design.md
```diff
@@ -0,0 +1,390 @@
...
+
+```
+{
+    "AWS::ApiGateway::RestApi": {
+        "RootResourceId": "/"
+    }
+}
+```
+
+* common_attribute_resolver: resolves common attributes that will be true across all
```
jfuss: Is this used for helping resolve `Fn::GetAtt` for all the 'common' attributes a resource could typically have?

designs/intrinsics_design.md
```diff
@@ -0,0 +1,390 @@
...
+
+```
+{    "Ref": lambda p,r: "",
+     "Arn:": arn_resolver}
+```
+
+
+First pseudo types are checked. If item is present in the logical_id_translator it is returned. Otherwise, it falls back to the default_pseudo_resolver. Then the default_type_resolver is checked, which has common attributes and functions for each types. Then the common_attribute_resolver is run, which has functions that are common for each attribute.
```
jfuss: All of this is for Pseudo Parameters? Pseudo parameters are all predefined. Seems like a lot of extra work that isn't needed.

designs/intrinsics_design.md
```diff
@@ -0,0 +1,390 @@
...
+```
+processed_template = {}
+for key, val in self.resources.items():
+    processed_key = self.symbol_resolver.get_translation(key, IntrinsicResolver.REF) or key
+
+    processed_resource = self.intrinsic_property_resolver(val)
+    processed_template[processed_key] = processed_resource
+```
+
+If there is an error with the resource, the item will be ignored and processed.
```
jfuss: What do you mean by error with the resource? Without understanding what you mean, it's hard to tell if this really should just ignore or bubble up the error to the customer.

designs/intrinsics_design.md
viksrivat: Yes

designs/intrinsics_design.md
viksrivat: The paragraph might need to be clarified. The paragraph describes the flow of logic for resolving symbol table attributes.

designs/intrinsics_design.md
viksrivat: The idea was to not break any workflows for unresolved symbol table or reference errors. Currently, in SAM these properties are ignored and wouldn't cause any errors, so by ignoring the errors we are using the same functionality. This is true for refs that exist to cloud instances. To ignore the error, we just copy the resource as is. 
I will update this in the description.

designs/intrinsics_design.md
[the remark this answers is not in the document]
TheSriram: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/ec2-example-regions-avail-zones.html

I think you can list regions without network access, but not availability zones.

```
from boto3.session import Session

Session().get_available_regions('cloudformation')
```

Issue #528, opened 2018-07-04
Missing !Sub and !If instrinsic functions


ericraymond:
**Description:**

Related to issue #490, please add support for !Sub !Ref, and a minimal !If implementation for environment vars in serverless function or in a globals section. 

Not having this functionality means copy-paste duplication within a template, maintaining a duplicate template for local execution,  copy-paste duplication to specify env vars on the command line, and/or the inability to override default values of environment variables via parameters at runtime.

Bonus points for supporting all variations expressing the fn in  Yaml and Json 

In 0.3.0 this was refactored and appears only to be supporter in a very limited context (e.g., swagger urls?).

**Output of `sam --version`**: 0.4.0

jfuss:
This is also related to #450 and #476.

Going to update labels.

charsleysa:
@jfuss is there any progress on this?

mnapoli:
@charsleysa exactly! I face the same problem with region and even layer versions (I want to use environment variables for those). Related: https://github.com/awslabs/serverless-application-model/issues/751

mnapoli:
Just a heads up for those interested, `!Ref` has been implemented in #657.

jfuss:
I have (slowly) been working on some Sub support. I have pushed the code to my [branch/repo](https://github.com/jfuss/aws-sam-cli/tree/sub-intrinsics). I have some functional tests that are working but missing:

* Unit tests
* A note/warning on not resolving. This is really important since we cannot resolve commuted properties of resources.
* Integ tests

If anyone is looking for a good way to contribute and need/want sub support, feel free to take what I started as the base.

henrikbjorn:
Running into this also for Layers.

joshpeng:
How do we get `!Ref` to work? I keep running into this:
```
Invalid value in Template for key: MY_ENV_VAR: expected string
```
Here is my `template.yml`:
```yml
Parameters:
  SomeTestParam:
    Type: String
    Default: foobar

...

SomeLambdaFunction:
    Type: AWS::Serverless::Function
    Properties:

      ...

      Environment:
        Variables:
          MY_ENV_VAR: !Ref SomeTestParam
```
I am on SAM CLI v0.19.0.

daveykane:
Also having trouble with !Sub not working with Layers

jfuss:
Release in [v0.21.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.21.0).

Closing

[...]

PR #1261, opened 2019-07-09
Intrinsics support

viksrivat:
*Issue #, if available:*
#528
*Description of changes:*
#1260 

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+        self.resources = resources or self.template.get("Resources", {})
+        self.mapping = mappings or self.template.get("Mappings", {})
+        self.parameters = parameters or self.template.get("Parameters", {})
+        self.conditions = conditions or self.template.get("Conditions", {})
+
+        self.symbol_resolver = symbol_resolver
+        self.symbol_resolver.resources = self.resources
+        self.symbol_resolver.parameters = self.parameters
+
+        self.intrinsic_key_function_map = {
```
jfuss: This seems like another class that could be created? Any reason you elected to do this mapping vs. making it into an explicit class?

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+            IntrinsicResolver.FN_BASE64: self.handle_fn_base64,
+            IntrinsicResolver.FN_FIND_IN_MAP: self.handle_find_in_map,
+            IntrinsicResolver.FN_TRANSFORM: self.handle_fn_transform,
+            IntrinsicResolver.FN_GET_AZS: self.handle_fn_get_azs,
+            IntrinsicResolver.REF: self.handle_fn_ref,
+            IntrinsicResolver.FN_GET_ATT: self.handle_fn_getatt,
+            IntrinsicResolver.FN_IMPORT_VALUE: self.handle_fn_import_value
+        }
+
+        self.conditional_key_function_map = {
```
jfuss: Same here as the `self.intrinsic_key_function_map`

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+            IntrinsicResolver.FN_AND: self.handle_fn_and,
+            IntrinsicResolver.FN_OR: self.handle_fn_or,
+            IntrinsicResolver.FN_IF: self.handle_fn_if,
+            IntrinsicResolver.FN_EQUALS: self.handle_fn_equals,
+            IntrinsicResolver.FN_NOT: self.handle_fn_not
+        }
+
+    def intrinsic_property_resolver(self, intrinsic, parent_function="template"):
```
jfuss: Having a hard time understanding what `parent_function="template"` is.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+            raise InvalidIntrinsicException("Missing Intrinsic property in {}".format(parent_function))
+
+        if any(isinstance(intrinsic, object_type) for object_type in [string_types, list, bool, int]):
+            return intrinsic
+
+        if intrinsic == {}:
+            return intrinsic
+
+        if not isinstance(intrinsic, dict):
+            raise InvalidIntrinsicException(
```
jfuss: This would never happen right? You are explicitly checking for other values above. Won't this be a dict?

The message also says "It is not a int, list, str, or dict" but you are checking for if it is not a dict.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+                "Invalid Intrinsic type. It is not a int, list, str, or dict {}".format(intrinsic))
+
+        keys = list(intrinsic.keys())
+        key = keys[0]
+
+        if key in self.intrinsic_key_function_map:
```
jfuss: Why are we not just doing 

`for key, value in intrinsics.items():`

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+            intrinsic_value = intrinsic.get(key)
+            return self.conditional_key_function_map.get(key)(intrinsic_value)
+
+        # In this case, it is a dictionary that doesn't directly contain an intrinsic resolver and must be
+        # re-parsed to resolve.
+        sanitized_dict = {}
+        for key, val in intrinsic.items():
+            sanitized_key = self.intrinsic_property_resolver(key, parent_function=parent_function)
+            sanitized_val = self.intrinsic_property_resolver(val, parent_function=parent_function)
+            verify_intrinsic_type_str(sanitized_key,
```
jfuss: Fail after each invalid intrinsic might not be a good user experience. Imagine you are a customers that has 5 different 'bugs' with intrinsics in your template. In order to make a single invoke work, you will be running this 5 times. Could be worth adding a collector that captures errors and prints them all at once.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+        This will parse through every entry in a CloudFormation template and resolve them based on the symbol_resolver.
+        Customers can optionally ignore resource errors and default to whatever the resource provides.
+
+        Return
+        -------
+        A resolved template with all references possible simplified
+        """
+        processed_template = {}
+        for key, val in self.resources.items():
+            processed_key = self.symbol_resolver.get_translation(key, IntrinsicResolver.REF) or key
```
jfuss: `IntrinsicResolver.REF` isn't need here, `get_translation` already defaults to this.

What does get_translation do here? It's not clear what the translation is referring too.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+            try:
+                processed_resource = self.intrinsic_property_resolver(val)
+                processed_template[processed_key] = processed_resource
```
jfuss: Not sure I follow what is going on here. Isn't `processed_key` always equal to `key` (aka. the logical id)?

viksrivat: There are cases when processed_key != key. The example in the SAM Ref code was RestApi.Deployment -> NewRestApi. This will be updated based on parameters to the logical_id_translator.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: get_translation simplifies something like to RestApi.Deployment -> NewRestApi.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: That is a good idea when dealing with errors. One issue with this is with objects that are looking for a certain type. For example, Fn::Select requires the first argument to be an integer. If there is an error during nested resolution, can we continue execution to resolve further errors? This could also possibly be issues with templates that point to deployed cloud resources.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: This simplifies figuring out where an error has occured since this function is called generically to provide a better error message. By default, when parsing the template the error will message will assume the location is the "template". However, inside nested Intrinsics having something like self.intrinsic_property_resolver(intrinsic_value, parent_function=IntrinsicResolver.FN_JOIN) simplifies the handling by telling users their errors are in Fn::Join. 

This happens in two cases. The object property resolved is missing or it is a nested dictionary within the object that needs to be resolved.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: The only case I was consider was if an object was there. I'll remove it since we are assuming that it's valid CloudFormation read from a file.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: It seemed nicer in terms of it's functional formatting and structure.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: It does the same thing as that. It's just going to enter the loop one time so it seemed unneeded or uglier to make it a loop.

samcli/commands/local/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,759 @@
...
+        """
+        self.template = template or {}
+        self.resources = resources or self.template.get("Resources", {})
+        self.mapping = mappings or self.template.get("Mappings", {})
+        self.parameters = parameters or self.template.get("Parameters", {})
+        self.conditions = conditions or self.template.get("Conditions", {})
+
+        self.symbol_resolver = symbol_resolver
+        self.symbol_resolver.resources = self.resources
+        self.symbol_resolver.parameters = self.parameters
```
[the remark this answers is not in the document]
viksrivat: It seemed nicer that we only had to parse the template once for both. The other case is that if someone specified the logical_id_translation in symbol_resolver before hand and at a different time calls IntrinsicsResolver with the run time parameters.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+
+        Return
+        -------
+        A string with the resolved attributes
+        """
+        data = self.intrinsic_property_resolver(intrinsic_value, parent_function=IntrinsicResolver.FN_BASE64)
+
+        verify_intrinsic_type_str(data, IntrinsicResolver.FN_BASE64)
+        # Encoding then decoding is required to return a string of the data
+        return base64.b64encode(data.encode()).decode()
```
jfuss: Are we sure this is the correct one? How did you validate this is correct

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+
+        Return
+        -------
+        A string with the resolved attributes
+        """
+        macro_name = intrinsic_value.get("Name")
+        name = self.intrinsic_property_resolver(macro_name, parent_function=IntrinsicResolver.FN_TRANSFORM)
+
+        if name not in IntrinsicResolver.SUPPORTED_MACRO_TRANSFORMATIONS:
+            raise InvalidIntrinsicException(
```
jfuss: Log a debug message here?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+        """
+        { "Fn::ImportValue" : sharedValueToImport }
+        This intrinsic function requires handling multiple stacks, which is not currently supported by SAM-CLI.
+        Thus, it will thrown an exception.
+
+        Return
+        -------
+        An InvalidIntrinsicException
+        """
+        raise InvalidIntrinsicException("Fn::ImportValue is currently not supported by IntrinsicResolver")
```
jfuss: One thing we need to make sure is that we log debug messages in these cases (maybe in info level) so customers understand what we support or don't support.

Sidenote: We will not want to fail resolving intrinsics initially, or at least until we know we have all things covered correctly. Otherwise, this will become a blocker for customers when they try invoking/building/etc.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
```diff
@@ -0,0 +1,424 @@
...
+        resource_attribute: str
+            The resource attribute of the resource in question or Ref for psuedo types.
+        ignore_errors: bool
+            An optional flags to not return errors. This used in sub
+
+        Return
+        -------
+        This resolves the attribute
+        """
+        # pylint: disable-msg=too-many-return-statements
```
jfuss: This is an indication that we should split this function out or find a better way to manage the control flows to reduce return statements (in some sane way).

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
```diff
@@ -0,0 +1,424 @@
...
+        The resolved Arn
+        """
+        aws_region = self.logical_id_translator.get(IntrinsicsSymbolTable.AWS_REGION) or self.handle_pseudo_region()
+        account_id = self.logical_id_translator.get(
+            IntrinsicsSymbolTable.AWS_ACCOUNT_ID) or self.handle_pseudo_account_id()
+        partition_name = self.handle_pseudo_partition()
+        function_name = logical_id
+        function_name = self.logical_id_translator.get(function_name) or function_name
+        str_format = "arn:{partition_name}:{service_name}:{aws_region}:{account_id}:{function_name}"
+        if service_name == "lambda":
```
jfuss: I wonder what other services have similar (custom) things in their arns.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -0,0 +1,1443 @@
...
+        [True, False, {}, 42, None]
+    ])
+    def test_fn_join_items_all_str(self, name, single_obj):
+        with self.assertRaises(InvalidIntrinsicException, msg=name):
+            self.resolver.intrinsic_property_resolver({
+                "Fn::Join": ["", ["test", single_obj, "abcd"]]
+            })
+
+
+class TestIntrinsicFnSplitResolver(TestCase):
```
jfuss: This goes for all of them, we should have some negative case tests too. Like what happens when you split but the delimiter is not in the string.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -0,0 +1,1443 @@
...
+        [True, False, "Test", {}, 42]
+    ])
+    def test_fn_split_arguments_invalid_formats(self, name, intrinsic):
+        with self.assertRaises(InvalidIntrinsicException, msg=name):
+            self.resolver.intrinsic_property_resolver({
+                "Fn::Split": intrinsic
+            })
+
+    @parameterized.expand([
+        ("Fn::Split should fail if the first argument does not resolve to a string: {}".format(item), item)
```
jfuss: I don't know the answer to this question but, could I have a ref to a parameter as the first argument? Does that work in CloudFormation?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+            not mutate the template.
+        """
+        self._template = template or {}
+        self._resources = self._template.get("Resources", {})
+        self._mapping = self._template.get("Mappings", {})
+        self._parameters = self._template.get("Parameters", {})
+        self._conditions = self._template.get("Conditions", {})
+
+        self._symbol_resolver = symbol_resolver
+        self._symbol_resolver.resources = self._resources
```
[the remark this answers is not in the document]
viksrivat: I decided to set it here because IntrinsicResolver is parsing the template and the resources here. The symbol resolver doesn't need worry about parsing.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+
+        if intrinsic == {}:
+            return intrinsic
+
+        keys = list(intrinsic.keys())
+        key = keys[0]
+
+        if key in self.intrinsic_key_function_map:
+            intrinsic_value = intrinsic.get(key)
+            return self.intrinsic_key_function_map.get(key)(intrinsic_value)
```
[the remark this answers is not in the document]
viksrivat: If we have intrinsic inside intrinsic, it is resolved within each specific intrinisic_function. This is because there are special cases to consider for each type when dealing with functions within functions. For example, Fn::Select needs to have the first element be an integer index.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+        elif key in self.conditional_key_function_map:
+            intrinsic_value = intrinsic.get(key)
+            return self.conditional_key_function_map.get(key)(intrinsic_value)
+
+        # In this case, it is a dictionary that doesn't directly contain an intrinsic resolver and must be
+        # re-parsed to resolve.
+        sanitized_dict = {}
+        for key, val in intrinsic.items():
```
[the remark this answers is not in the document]
viksrivat: This is dealing with the case where we have a dictionary of the format:
{"Property":
    "Item": {
        "IntrinsicPropertyKey": Intrinsic 
}
}
I updated the comment in the code to clarify it.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: Currently, I was handling for that in resolve_template() function. This has a property called ignore_errors, which we can keep on till we have covered everything. If ignore_errors is turned on, then it will run LOG.error("Unable to process properties of %s.%s", key, resource_type) and keep the current resource unmodified.
Ex Code around line 170: 
            except (InvalidIntrinsicException, InvalidSymbolException) as e:
                resource_type = val.get("Type", "")
                if ignore_errors:
                    LOG.error("Unable to process properties of %s.%s", key, resource_type)
                    processed_template[key] = val

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: I was following the spec from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-base64.html. I also tested some base64 encoded responses.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: Same as other comment about handling InvalidIntrinsicExceptions

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,743 @@
...
+        self._resources = self._template.get("Resources", {})
+        self._mapping = self._template.get("Mappings", {})
+        self._parameters = self._template.get("Parameters", {})
+        self._conditions = self._template.get("Conditions", {})
+
+        self._symbol_resolver = symbol_resolver
+        self._symbol_resolver.resources = self._resources
+        self._symbol_resolver.parameters = self._parameters
+
+        self.intrinsic_key_function_map = {
```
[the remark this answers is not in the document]
viksrivat: wouldn't a class like the following handles extending it:
```
class NewResolver(IntrinsicResolver):
    def __init__(self):
         super(IntrinsicResolver, self).__init__()
         self.intrinsic_key_function_map.extend({dic of resolvers})
```
 I'm on the fence about injecting it as a class. It feels unnecessary/unpythonic to have a class just for the dictionary rather than have it extended.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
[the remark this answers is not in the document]
viksrivat: This solves the problem in a top down fashion instead of bottom up in order to create a functional immutable style.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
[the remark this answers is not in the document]
viksrivat: I cleaned up the code to expose intrinsic_key_function_map to a user. I also exposed some of the properties such as adding the template when resolving it.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+
+        self.intrinsic_key_function_map = self.default_intrinsic_function_map()
+        self.conditional_key_function_map = self.default_conditional_key_map()
+
+    @property
+    def template(self):
+        return self._template
+
+    @template.setter
+    def template(self, template):
```
jfuss: Should this be mutable after `__init__`?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+        """
+        Set the attributes from the given template. The template is copied to avoid risking mutation of the original
+        template.
+
+        This is defined outside init for readability purposes.
+        :param template: the template dict to parse
+        """
+        # pylint: disable-msg=attribute-defined-outside-init
```
jfuss: Should these be defaulted in `__init__` and then overwritten?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+
+    def __init__(self, template=None, symbol_resolver=None):
+        """
+        Initializes the Intrinsic Property class with the default intrinsic_key_function_map and
+        conditional_key_function_map.
+
+        In the future, for items like Fn::ImportValue multiple templates can be provided
+        into the function.
+        """
+        self.template = template
```
jfuss: Are there any downsides to this pattern? This does allow `self.template` to be mutable, which we might not want. You could just move everything in the setter to the `__init__`, to avoid any confusion but maybe there is a case for when we would want to mutate this?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+        ignore_errors: bool
+            An option to ignore
+        symbol_resolver: IntrinsicSymbolTable
+            An Instance of the IntrinsicSymbolTable object contaning resolution for attributes like Ref, Fn::GetAtt
+        Return
+        -------
+        A resolved template with all references possible simplified
+        """
+        if template:
+            self.template = template
```
jfuss: O I see why you did the setter this way. You don't need to have the template pass in the `__init__` if you also pass it here. It is weird to me that you can default in the `__init__` but then overwritten here. Pick one (its ok to have a required parameter).

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+        if symbol_resolver:
+            self._symbol_resolver = symbol_resolver
+
+        processed_template = {}
+        for key, val in self._resources.items():
```
jfuss: This and any of the other _ properties (`self._mapping`, `self._parameters`, or `self._conditions`) can result in an AttributeError due to the early exit if the template is None or empty (line 95). This is also why it's important not to ignore the attribute-defined-outside-init that pylint throws.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+        """
+        This will parse through every entry in a CloudFormation template and resolve them based on the symbol_resolver.
+        Customers can optionally ignore resource errors and default to whatever the resource provides.
+
+        Parameters
+        -----------
+        template: dict
+            A dictionary containing all the resources and the attributes of the CloudFormation template. The class does
+            not mutate the template.
+        ignore_errors: bool
```
jfuss: Should `ignore_errors` ignore any and all errors/exceptions or only certain ones?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,811 @@
...
+        A resolved template with all references possible simplified
+        """
+        if template:
+            self.template = template
+        if symbol_resolver:
+            self._symbol_resolver = symbol_resolver
+
+        processed_template = {}
+        for key, val in self._resources.items():
+            processed_key = self._symbol_resolver.get_translation(key) or key
```
jfuss: Should `self._symbol_resolver.get(key)` just default to returning the key back if one can't be found?

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
jfuss: Can we look at a few other services? This function looks like it should be generic but has `function_name` in all arns, so it reads a little off. Also, I have the general concern that Lambda isn't the only special arn one here. Keeping this updated with every service, could be a big pain.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        Gets the region from the environment and defaults to a the default region from the global variables.
+
+        This is only run if it is not specified by the logical_id_translator as a default.
+
+        Return
+        -------
+        The region from the environment or a default one
+        """
+        return os.getenv("AWS_REGION") or SamBaseProvider.DEFAULT_PSEUDO_PARAM_VALUES.get(
```
jfuss: This only accounts for if they set `AWS_REGION`, what happens if a customer passes `--region` into the command. How do we resolve to that region instead of the default or one set in an env var?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: The reason I think we should continue to return None for self._symbol_resolver.get(key) is because we may want to differentiate between keys that don't exist and symbols that references to a logical_id in the future. This also makes it feel like dict.get(), which returns a default of None.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
viksrivat: The arn types are 
arn:partition:service:region:account-id:resource
arn:partition:service:region:account-id:resourcetype/resource
arn:partition:service:region:account-id:resourcetype/resource/qualifier
arn:partition:service:region:account-id:resourcetype/resource:qualifier
arn:partition:service:region:account-id:resourcetype:resource
arn:partition:service:region:account-id:resourcetype:resource:qualifier

The ones I looked at sns/sqs followed the default 
         str_format = "arn:{partition_name}:{service_name}:{aws_region}:{account_id}:{function_name}" 

Before, we also only supported dealing with Lambda Functions. This function will also only be called as a default case if they do not specify and override for resolving GetAtt for Arns.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
viksrivat: My plan for this during integration would be to add to add it to the logical_id_translator and override it.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,789 @@
...
+            return self.intrinsic_key_function_map.get(key)(intrinsic_value)
+        elif key in self.conditional_key_function_map:
+            intrinsic_value = intrinsic.get(key)
+            return self.conditional_key_function_map.get(key)(intrinsic_value)
+
+        # In this case, it is a dictionary that doesn't directly contain an intrinsic resolver, we must recursively
+        # resolve each of it's sub properties.
+        sanitized_dict = {}
+        for key, val in intrinsic.items():
+            sanitized_key = self.intrinsic_property_resolver(key, parent_function=parent_function)
```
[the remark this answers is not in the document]
TheSriram: does parent function just tell you where you are in the recursive stack?

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -0,0 +1,789 @@
...
+        arguments = self.intrinsic_property_resolver(intrinsic_value, parent_function=IntrinsicResolver.FN_AND)
+        verify_intrinsic_type_list(arguments, IntrinsicResolver.FN_AND)
+
+        for i, argument in enumerate(arguments):
+            if isinstance(argument, dict) and "Condition" in argument:
+                condition_name = argument.get("Condition")
+                verify_intrinsic_type_str(condition_name, IntrinsicResolver.FN_AND)
+
+                condition = self._conditions.get(condition_name)
+                verify_non_null(condition, IntrinsicResolver.FN_AND, position_in_list="{} th".format(str(i)))
```
TheSriram: "i th" might not always hold, since there are 1st, 2nd, 3rd.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -0,0 +1,1424 @@
...
+            "Mappings": self.mappings,
+            "Conditions": self.conditions,
+            "Resources": resources,
+        }
+        symbol_resolver = IntrinsicsSymbolTable(template=template,
+                                                logical_id_translator=self.logical_id_translator)
+        resolver = IntrinsicResolver(template=template,
+                                     symbol_resolver=symbol_resolver)
+        result = resolver.resolve_template(ignore_errors=True)
+        expected_template = {'HelloHandler2E4FBA4D': {'Properties': {'handler': 'main.handle'},
```
TheSriram: can the larger templates be added as separate files, that can be checked against. maybe under testdata/?

TheSriram:
One general feedback is to submit smaller PRs, for eg: instrintic property resolver with one just one instrinsic tested, and to build on them.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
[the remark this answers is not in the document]
viksrivat: The recursion limit in python is a 1000. I don't think we need to explicitly enforce this. SAM translator deals with it given the python limit.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
```diff
@@ -0,0 +1,403 @@
...
+               "us-east-2": ["us-east-2a", "us-east-2b", "us-east-2c"],
+               "ap-south-1": ["ap-south-1a", "ap-south-1b", "ap-south-1c"],
+               "eu-central-1": ["eu-central-1a", "eu-central-1b", "eu-central-1c"],
+               "eu-west-1": ["eu-west-1a", "eu-west-1b", "eu-west-1c"],
+               "eu-west-2": ["eu-west-2a", "eu-west-2b", "eu-west-2c"],
+               "eu-west-3": ["eu-west-3a", "eu-west-3b", "eu-west-3c"],
+               "cn-north-1": [],
+               "us-gov-west-1": []}
+
+    DEFAULT_PARTITION = "aws"
```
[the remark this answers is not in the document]
viksrivat: I wanted to avoid using boto3 for this, because it may bring in authentication constraints. There are ways to get around it, but it seems very ugly/hacky.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: All the template initialization has been moved up to the __init__

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: All the template initialization has been moved up to the __init__

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: Same as the other comment about moving things up to __init__

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: All of these properties are now initialized before calling this to prevent an AttributeError

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
viksrivat: I updated the docstring to reflect that ignore_errors only ignores InvalidIntrinsicExceptions

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
```diff
@@ -0,0 +1,419 @@
...
+            }
+        common_attribute_resolver: dict
+            This is a clean way of specifying common attributes across all types.
+            The value can either be a function of the form string or (logical_id) => string
+            {
+                "Ref": lambda p,r: "",
+                "Arn:": arn_resolver
+            }
+        """
+        if template is None:
```
jfuss: You don't do anything be set template to `self._template` so this check can be wrapped into that assignment.

samcli/lib/intrinsic_resolver/intrinsics_symbol_table.py
jfuss: That's fine but then we shouldn't have the `str_format` be tied to `function_name`. My callout is an arn for sqs doesn't have a function_name, it has a queue_name. Since this is broad we need to make sure these reads as such. Reading it as it is written, makes it seems like lambda is only supported.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -0,0 +1,1347 @@
+import json
+from copy import deepcopy
+from pathlib import Path
```
jfuss: pathlib is not in the Python2.7 standard library. You will need to do 

```
try:
    import pathlib
except ImportError:
    import pathlib2 as pathlib
```

to make the Py2.7 tests pass.

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
daveykane: It would be helpful if these logs were only printed if the --debug flag is set, or if the log only printed if the property was one that should be possible to process but for whatever reasons couldn't be. Currently there is a lot of noise in terminal for unprocessed properties that SAM CLI can not process i.e. ImportValue

Issue #1099, opened 2019-04-04
SAM Build support for Provided runtime


gkrizek:
### Describe your idea/feature/enhancement

I am trying to use [NodeSource's N|Solid Lambda Layer](https://nodesource.com/products/nsolid-aws-lambda) with SAM. This is a replacement for Node.js and therefore uses the `provided` runtime. Everything else with my function is identical to the Node.js runtime. I have a `package.json` that included dependencies for my application.

When running `sam build` I get the follow error:

```
$ sam build
2019-04-04 13:26:52 Found credentials in shared credentials file: ~/.aws/credentials
2019-04-04 13:26:52 Building resource 'Function'
Build Failed
Error: 'provided' runtime is not supported
```
I can't use SAM to build and deploy my Node.js functions when they use a provided runtime.

**I would like `sam build` to support the `provided` runtime.**

### Proposal

Obviously, the `provided` runtime is a little tricky for building because the runtime could be anything and there's no way to know from the `template.yaml` file.

My proposal would be to add an optional flag to the `sam build` command called `--builder`. With this flag you could tell `sam build` which builder to use regardless of runtime. This way we can still use the `provided` runtime, but still use `sam build` to package our Node.js applications.

gkrizek:
I think we should perhaps only allow this flag when the runtime is `provided`. If we let someone build a Python function with `--builder node` that's just asking for problems.

sanathkr:
@gkrizek This is a great usecase for `.samrc`(#975) - basically a config file that changes the default behavior for SAM CLI. I like the rc file over command line arguments because you can setup how SAM CLI behaves (in this case "sam build") one time, check into your repository, and be reused across your team without learning a new flag and typing it every time.

Thoughts?

gkrizek:
@sanathkr Yeah I think this sounds like a good idea. I like the reusability of it. Is there any ETA on supporting a `.samrc`?

sanathkr:
@gkrizek Not yet.

gkrizek:
@sanathkr Since there's no date on getting a `.samrc` file implemented, do you think a CLI flag would be a viable option in the meantime? Then users would have the ability to either use the `.samrc` or the CLI flag once both exist.

sanathkr:
@gkrizek You know what, there is a better solution to this: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/build/workflow_config.py#L107-L112

[...]

PR #1279, opened 2019-07-18
[WIP] build: added support for building using 'provided' runtime.

mster:
Issue #1099

*Description of changes:*

Allow for the building of a lambda function when specifying `Runtime: provided`, with a corresponding layer ARN. SAM will automagically determine the runtime by determining the workflow configuration based off of the project manifest.

For example, with the sample `sam-app`:
```yml
...
      Runtime: provided
      Layers:
        - <YOUR-PROVIDED-RUNTIME-ARN>
...
```

* Added a `ManifestWorkflowSelector` to add support for building using the `provided` runtime flag, as suggested [here](https://github.com/awslabs/aws-sam-cli/issues/1099#issuecomment-488877333). 
* Added logging to inform user that the `provided` runtime is being used. 
* Fails when attempting to build Lambda Function within a container. (can't determine an appropriate image to pull)

*Checklist:*

- [x] Write unit tests
- [-] Write/update functional tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/workflow_config.py
+++ b/samcli/lib/build/workflow_config.py
@@ -111,12 +111,22 @@ def get_workflow_config(runtime, code_dir, project_dir):
             JAVA_KOTLIN_GRADLE_CONFIG._replace(executable_search_paths=[code_dir, project_dir]),
             JAVA_MAVEN_CONFIG
         ]),
+
+         "provided": ManifestWorkflowSelector([
+            PYTHON_PIP_CONFIG,
+            RUBY_BUNDLER_CONFIG,
+            DOTNET_CLIPACKAGE_CONFIG,
+            NODEJS_NPM_CONFIG,
+        ]),
     }
 
     if runtime not in selectors_by_runtime:
         raise UnsupportedRuntimeException("'{}' runtime is not supported".format(runtime))
 
     selector = selectors_by_runtime[runtime]
+
+    if runtime == "provided":
+        LOG.info("Building resource using 'provided' runtime")
 
     try:
         config = selector.get_config(code_dir, project_dir)
--- a/samcli/lib/build/workflow_config.py
+++ b/samcli/lib/build/workflow_config.py
@@ -112,11 +112,14 @@ def get_workflow_config(runtime, code_dir, project_dir):
             JAVA_MAVEN_CONFIG
         ]),
 
-         "provided": ManifestWorkflowSelector([
+        "provided": ManifestWorkflowSelector([
             PYTHON_PIP_CONFIG,
             RUBY_BUNDLER_CONFIG,
             DOTNET_CLIPACKAGE_CONFIG,
             NODEJS_NPM_CONFIG,
+            JAVA_GRADLE_CONFIG._replace(executable_search_paths=[code_dir, project_dir]),
+            JAVA_KOTLIN_GRADLE_CONFIG._replace(executable_search_paths=[code_dir, project_dir]),
+            JAVA_MAVEN_CONFIG,
         ]),
     }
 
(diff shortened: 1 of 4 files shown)


samcli/lib/build/workflow_config.py
```diff
@@ -111,13 +111,23 @@ def get_workflow_config(runtime, code_dir, project_dir):
             JAVA_KOTLIN_GRADLE_CONFIG._replace(executable_search_paths=[code_dir, project_dir]),
             JAVA_MAVEN_CONFIG
         ]),
+
+         "provided": ManifestWorkflowSelector([
+            PYTHON_PIP_CONFIG,
+            RUBY_BUNDLER_CONFIG,
+            DOTNET_CLIPACKAGE_CONFIG,
+            NODEJS_NPM_CONFIG,
+        ]),
```
sanathkr: Is there any reason why you did not add other builders?

sanathkr:
Can you explain how you test this?

samcli/lib/build/workflow_config.py
mster: No particular reason, mainly just apprehension for testing.

tests/unit/lib/build_module/test_workflow_config.py
```diff
@@ -70,6 +70,30 @@ def test_must_work_for_java(self, runtime, build_file, dep_manager, os_mock):
...
+        self.assertEquals(result.language, language)
+        self.assertEquals(result.dependency_manager, dep_manager)
+        self.assertEquals(result.application_framework, None)
+        self.assertEquals(result.manifest_name, build_file)
+
+        if dep_manager == "gradle":
+            self.assertEquals(result.executable_search_paths, [self.code_dir, self.project_dir])
+        else:
+            self.assertIsNone(result.executable_search_paths)
+
```
mster: @sanathkr Would this be sufficient?

TheSriram: fair enough, the appropriate builder gets selected. A integration test for this, would be great as well.

mster:
> This is a great feature! thanks for adding this. The PR mentions the "appropriate layer ARN", is this left as an exercise to the user of the CLI to figure out? or should we mention that for example "Java runtime layer ARN needs to be provided since a java builder was picked on the basis of the manifest file?"

Excellent point. I do feel as though it should be up to the user to figure this out, however a log message during the build phase would be a useful addition. I will add that, let me know what you think.

mster:
@TheSriram for example:

```sh
2019-08-12 14:51:25 Building resource using 'provided' runtime.
2019-08-12 14:51:25 nodejs workflow detected. Please make sure you are using the appropriate layer ARN.
```

TheSriram:
@mster Awesome, thanks for updating the PR. I think the log message definitely makes sense. I still have some concerns on if we need to provide a mechanism for a `true` provided runtime that does not have a native builder yet? Do you have any thoughts on how we could integrate that? like a build plugin? (Thinking out loud here)

mster:
> have some concerns on if we need to provide a mechanism for a `true` provided runtime that does not have a native builder yet? Do you have any thoughts on how we could integrate that? like a build plugin? (Thinking out loud here)
 
What about requiring something in the `template.yaml` that specifies the builder to be used? 

```
...
      Runtime: provided
      Layers:
        - <provided-runtime-arn>
      Builder: <valid-native-builder>
...
```

Not sure on how to accomplish without the native builder.

jfuss:
Closing this as the issue linked has been closed and has been supported in SAM CLI. See https://github.com/aws/aws-sam-cli/issues/1099#issuecomment-661401086 for details.

PR #1289, opened 2019-07-25
fix: More robust connections to telemetry backend

sanathkr:
- Increase read timeout to 100ms. Backend needs a bit more time than 1ms to reliably receive & process data
- Don't crash if offline
- Integration tests to validate offline behavior & opt-in/opt-out behavior

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/telemetry/test_telemetry_contract.py
+++ b/tests/integration/telemetry/test_telemetry_contract.py
@@ -1,0 +1,73 @@
+
+from .integ_base import IntegBase, TelemetryServer
+
+
+class TestTelemetryContract(IntegBase):
+    """
+    Validates the basic tenets/contract Telemetry module needs to adhere to
+    """
+
+    def test_must_not_send_metrics_if_disabled_using_envvar(self):
+        """
+        No metrics should be sent if "Enabled via Config file but Disabled via Envvar"
+        """
+        # Enable it via configuration file
+        self.set_config(telemetry_enabled=True)
+
+        with TelemetryServer() as server:
+            # Start the CLI, but opt-out of Telemetry using env var
+            process = self.run_cmd(optout_envvar_value="0")
+            (_, stderrdata) = process.communicate()
+            retcode = process.poll()
+            self.assertEquals(retcode, 0, "Command should successfully complete")
+            all_requests = server.get_all_requests()
+            self.assertEquals(0, len(all_requests), "No metrics should be sent")
+
+            # Now run again without the Env Var Opt out
+            process = self.run_cmd()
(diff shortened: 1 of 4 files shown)


tests/unit/lib/telemetry/test_telemetry.py
```diff
@@ -121,15 +121,27 @@ def test_must_swallow_timeout_exception(self, requests_mock):
         #
 
         requests_mock.exceptions.Timeout = requests.exceptions.Timeout
+        requests_mock.exceptions.ConnectionError = requests.exceptions.ConnectionError
```
jfuss: Is this needed here if the this test is only for Timeout exception?

tests/integration/telemetry/test_telemetry_contract.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Validates the basic tenets/contract Telemetry module needs to adhere to
+    """
+
+    def test_must_not_send_metrics_if_disabled_using_envvar(self):
+        """
+        No metrics should be sent if "Enabled via Config file but Disabled via Envvar"
+        """
+        # Enable it via configuration file
+        self.set_config(telemetry_enabled=True)
```
jfuss: Personally, I am not crazy about setting this config through code, as these are integ tests and should be set how customers would enable/disable. I am ok with it, just would prefer always going through customer paths in our integ tests.

sanathkr: This is how customers would enable telemetry. Thru the config.

tests/unit/lib/telemetry/test_telemetry.py
sanathkr: Yeah lbecause of how Python exception mocking works

tests/integration/telemetry/test_telemetry_contract.py
jfuss: There is a sight nuance to me though. This function gets called through command executions for the customers. The customer would never explicitly call this method. I know it’s what happens for them but since a customer never directly interacts with it, calling the method in an integ test looks off to me. Not blocking on this, just something I noticed

sanathkr: Sure, but we need some way of mimicing customer's environment. Calling this method is a way to do that

sanathkr:
Couple of StartApi tests failed in Travis 2.7 only. This should be transient because I have never seen them fail before. Merging the PR anyway

PR #1303, opened 2019-07-30
(feat) sam destroy design

viksrivat:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


designs/sam_destroy_cmd.md
```diff
@@ -0,0 +1,180 @@
...
+This is the design for a command to **destroy** a CloudFormation stack.
+
+What is the problem?
+--------------------
+
+Customers create CloudFormation stacks using `sam deploy`. After deploying the stacks, customers may have to delete
+a stack to cleanup all of their resources and stacks because of stacks that landed in ROLLBACK_FAILED, 
+rapid testing of development tasks, and removing defective code that may have a large security risk.
+
+Currently, Customers would have to open up the CloudFormation console every time they have to delete, forcing them to leave the cli or use the `aws cloudformation` command.
```
TheSriram: Agreed! we need a more seamless CLI experience.

designs/sam_destroy_cmd.md
```diff
@@ -0,0 +1,180 @@
...
+            "cloudformation:delete" 
+            ], 
+            "Resource": "*" 
+        } 
+    ] 
+}
+```
+
+Customers that delete the stack may want to wait for the stack to be deleted before continuing. The following waiter
+can be used to wait for the stack to be deleted.
```
TheSriram: Is there an eventstream that we can print on the CLI? with what is the status of the destroy?

designs/sam_destroy_cmd.md
```diff
@@ -0,0 +1,180 @@
...
+with the following format to the user account.
+{ 
+    "Version": "2012-10-17", 
+    "Statement": [ 
+        { 
+            "Effect": "Allow", 
+            "Action": [ 
+            "cloudformation:delete" 
+            ], 
+            "Resource": "*"
```
TheSriram: Lets just add the corresponding resource instead of all the resources. also does the rolearn specified require permissions to delete other resources within the stack? say sqs, sns, apigateway? how do we handle that?

designs/sam_destroy_cmd.md
```diff
@@ -0,0 +1,180 @@
...
+---------------------
+
+In this proposal, we will be providing a new command, `sam destroy`, to
+destroy CloudFormation stacks in the cloud. 
+Customers will be able to view failures of the commands and other events directly inside the cli.
+
+
+Success criteria for the change
+-------------------------------
+- Customers will be able to delete a stack in the cloud
```
TheSriram: Do we need any constraints on the broadness of the destroy command? eg: --force to delete a stack even if it means there are resources that might be left behind?

viksrivat: Do you mean ignore resources that have receive a DELETE_FAILED state?

designs/sam_destroy_cmd.md
viksrivat: One option we could add is a -v or a verbose command that describes the stack after it has been deleted. Currently, It will print out the error message if it fails or provides a message to delete the stack.

TheSriram:
I think this design looks mostly good to me. 👍

jinixdme:
I really like the idea of automatically cleanup the resources created through '$ sam deploy'. But why do you want to implement a new sam cli function when you already can use '$ aws cloudformation delete-stack --stack-name sam-app'?

Why should you add the stack name to '$ sam destroy'? The stack names are already defined in the config files.

What I want is a cleanup function that does a smart deletion of all resources created by '$ sam deploy' with eg empty and delete s3 buckets first to avoid delete failures.

designs/sam_destroy_cmd.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+CLI Changes
+-----------
+```sh
+Usage: sam destroy [OPTIONS] [ARGS]...
+
+  The sam destroy command destroys a Cloudformation Stack.
+
+  e.g. sam destroy -stack-name sam-app
```
timoschilling, line 83:
```suggestion
  e.g. sam destroy --stack-name sam-app
```

morefaey:
Closing as we are tracking it in a related issue: https://github.com/aws/aws-sam-cli/issues/2191

PR #1197, opened 2019-05-25
[WIP]Destroy commands

zamirajaupaj:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #789, opened 2018-11-16
Feature Request: 'undeploy' CLI option


kjhosein:
**Description:**

(Am new to SAM, so this hopefully this makes sense!)

How about an `undeploy` or `delete` CLI option to delete an already deployed app/stack. Since the `deploy` is an alias to `aws cloudformation deploy`, couldn't there be a `delete` option that is an alias of `aws cloudformation delete-stack`?
Thanks!

**Potential Workflow:**

    cd <app-dir>
    sam undeploy OR sam delete

**Observed result:**
`Success. Your app (<name of your app>) has been deleted.`
OR
`Failed. <error message>`

**Expected result:**
Deletes your deployed app.

---
_Note to the maintainers: Although the **How to Contribute** doc says to add the `feature-request` label, I did not appear to have the permissions to do so._

StoneCypher:
i consider the inability to undeploy potentially defective code to be a fairly significant security risk

deploying over it may not be an option (cloudfront has a tendency to get wedged under sls)

is there an eta for this?

rl-michaelbrandeis:
you can just use the aws cli. not as convenient, but not a critical issue

`aws cloudformation delete-stack --stack-name xyz`

StoneCypher:
@rl-michaelbrandeis - that's actually pretty different than undeploy

We're looking for rollback, not destruction.  Deleting a stack can take an hour or more to recreate if there are complex interrelated resources, and occasionally much longer if you're renting capacity limited things like GPU machines

nickhalden:
This would be a good feature to have, I would be up for using the sam cli if cleaning up the resources is easier.

lestephane:
Will this include the deletion of the artifact S3 bucket if it was created by `sam deploy`?
That would be my expectation...

bitsofinfo:
agreed, this is definitely needed

jthomerson:
Adding my 👍 to this. Especially for wooing Serverless Framework users over, this will be important - we're used to `sls deploy` and `sls remove`

sboettger:
Thumbs up for this one!

DogFortune:
👍

jacurtis:
[...]

PR #1304, opened 2019-07-30
(feat) sam destroy

viksrivat:
*Issue #, if available:*
 #1197 and Issue #789

*Description of changes:*
Adds support for the `sam destroy` command to SAM CLI.

Design at #1303

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,197 @@
...
+                with the following format to the user account.
+                {
+                    "Version": "2012-10-17",
+                    "Statement": [
+                        {
+                            "Effect": "Allow",
+                            "Action": [
+                            "cloudformation:delete"
+                            ],
+                            "Resource": "*"
```
TheSriram: Should it be on the particular stack instead?

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,197 @@
...
+        ignore_cli_prompt=False):
+    """
+    Destroys the stack
+    """
+    # All logic must be implemented in the `do_cli` method. This helps ease unit tests
+    do_cli(ctx, stack_name, retain_resources, role_arn, client_request_token, wait, wait_time,
+           ignore_cli_prompt)  # pragma: no cover
+
+
+def verify_stack_exists(client, stack_name, required_status=None):
```
TheSriram: How about cross stack references? Especially if the stack we are about to delete has exports that are imported in other stacks? how can we be extra careful in those cases?

viksrivat: With nested stacks, I can log or output all the stacks that are references or being deleted. Cfn currently deletes the nested stacks and all its references if `aws-cli cloudformation delete-stack` is run. If termination protection is turned on, It seems to only be for the whole nested group. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-protect-stacks.html

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            click.secho("Failed to destroy Stack: {}".format(str(e.response["Error"]["Message"])), fg="red")
+
+        sys.exit(1)
+
+    if wait or wait_time:
+        waiter = cfn.get_waiter('stack_delete_complete')
+        try:
+            delay = 15
+            waiter.wait(StackName=stack_name,
```
TheSriram: is this a poller implementation that has been surfaced?

tests/unit/commands/destroy/test_cli.py
```diff
@@ -0,0 +1,249 @@
...
+        verify_stack_retain_resources(client, 'stack-name')
+        secho_client.assert_called_with(
+            'The logicalId test of the resource in the stack stack-name must be included in retain_resource'
+            ' since the deletion failed',
+            fg='red')
+        sys_exit.assert_called_with(1)
+
+    @patch('boto3.client')
+    @patch('click.confirm')
+    def test_check_nested_stack(self, confirm_client, boto_client):
```
TheSriram: we still need a integration test for this command too.

samcli/commands/destroy/__init__.py
viksrivat: Ye. It uses a poller behind the scenes

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+            if logical_id not in retain_resources:
+                click.secho(
+                    "The logicalId {} of the resource in the stack {} must be included in retain_resource since the "
+                    "deletion failed".format(logical_id, stack_name),
+                    fg="red",
+                )
+                sys.exit(1)
+
+
+def do_cli(  # pylint: disable=too-many-statements
```
TheSriram: did you run black on this? the formatting seems to be way different.

tests/integration/destroy/test_destroy_command_integ.py
```diff
@@ -0,0 +1,57 @@
...
+
+import boto3
+from botocore.exceptions import ClientError
+
+from samcli.commands.destroy import verify_stack_exists
+from tests.integration.destroy.test_destroy_integ_base import DestroyIntegBase
+
+LOG = logging.getLogger(__name__)
+
+# Publish tests require credentials and Travis will only add credentials to the env if the PR is from the same repo.
```
TheSriram: destroy tests, also our build system is all appveyor now, so this may need to change.

samcli/commands/destroy/__init__.py
viksrivat: Ye. I ran black on it prettify and cleanup the style

tests/integration/destroy/test_destroy_command_integ.py
viksrivat: I updated the tests to use the AppVeyor format

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+from samcli.lib.telemetry.metrics import track_command
+
+LOG = logging.getLogger(__name__)
+
+SHORT_HELP = "Destroys a deployed CloudFormation stack."
+
+HELP_TEXT = """The sam destroy command destroys a Cloudformation Stack.
+
+\b
+e.g. sam destroy -stack-name sam-app --retain-resources LOGICAL_ID_1 --retain-resources LOGICAL_ID_2
```
awood45: I think we should accept a `--profile` argument and use it for the profile input into the boto3 client. This gives users flexibility in choosing which credentials to use.

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+    """
+    Implementation of the ``cli`` method, just separated out for unit testing purposes
+    """
+    if not force:
+        click.confirm(
+            "Are you sure you want to delete the stack {}?".format(stack_name),
+            default=True,
+            abort=True,
+        )
+    cfn = boto3.client("cloudformation", region_name="us-west-1")
```
awood45: Why do we have a hard-coded region here? This is going to break for most customers. We should actually accept a region argument.

samcli/commands/destroy/__init__.py
awood45: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html - accepting a profile will change how we construct our boto3 client.

samcli/commands/destroy/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+        events = [
+            event
+            for event in iterator.get("StackEvents")
+            if event.get("ResourceStatus") == "DELETE_FAILED"
+        ]
+        for event in events:
+            logical_id = event.get("LogicalResourceId")
+            if logical_id not in retain_resources:
+                click.secho(
+                    "The logicalId {} of the resource in the stack {} must be included in retain_resource since the "
```
awood45: I think this error might be a bit incorrectly described. When a deletion fails to clean up a resource, sometimes you should mark it retained, but sometimes it's actually something you want to resolve first, then proceed. I'd recommend this wording:

```python
"The logicalId {} of the resource in the stack {} failed to delete. You'll need to resolve the root cause of the failure to delete this resource or add it to the set of retained resources in order to complete deletion of your stack.".format(logical_id, stack_name),
fg="red"
```

cabrinha:
It'd be great if we could get this issue looked at again...

bitsofinfo:
when might this get in?

aahung:
Close because this implementation is outdated

Issue #239, opened 2017-12-22
Feature request: make it possible to keep docker container warm


jandockx:
I understand from other issues that a new docker container is started for each request. This makes some experiments or automated tests undoable in practice. SAM Local is much too slow in the context where more then 1 request is to be handled.

I suspect that hot reloading depends on this feature.

I think it would be a good idea to make it possible to choose, while this project evolves further, to forego hot reloading, but to keep the docker container warm.

Something like 

    sam local start-api -p <PORT> --profile <AWS PROFILE> --keep-it-warm

This would broaden the applicability of sam local enormously.

Thank you for considering this suggestion. This looks like an awesome project.

aldegoeij:
+1 Python container takes too long to start for simple debugging...

zippadd:
+1. This currently makes local automated testing painful at best.

Thanks for the continued work on this project!

dannymcpherson:
Have there been any eyes on this?  The benefit would be so huge.

cagoi:
+1

daveykane:
+1

adrian1358:
+1

CRogers:
+1, even a simple hello world java8 lambda takes 3/4 seconds for each request!

monofonik:
+1

luisvsm:
+1 Very interested in this feature

luketn:
+1 Yes please!

nodeit:
+1, throwing my hat in the ring on this too

jfuss:
As a note: Please use the reaction feature on the top comment. We do look at issues sorted by thumbs up (as well as other reactions). Commenting +1 does not good for that and adds noise to the issue.

scoates:
@jfuss I agree (and had done this). Any feedback from your team would be helpful here, though. The closest thing we had to knowing if this is on your radar (before your comment) was duplicate issue consolidation and labeling.

ejoncas:
+1, this would be very beneficial for people using java + spring boot.

thoratou:
+1, around 1s for golang case

peachepe:
1 second is a world's difference when building an API and you expect to serve more than 1 request.

I think it's well worth the feature.

sanathkr:
@kevanpng Hey I was looking through your code to understand what exactly you did.. So basically, you create the container once with a fixed name, run the function, and on next invocation look for container with same name and simply `container.exec_run` instead of creating it from scratch again. Is my summary correct?

[...]

PR #1305, opened 2019-07-30
Warm invoke containers for start-api and start-lambda

jfuss:
*Issue #, if available:*
#239

*Description of changes:*

This is the start of providing the capability of a 'warm container' invoke. This PR changes some behavior, which I think has a net positive benefit.

New behavior:
When running `sam local start-api` or `sam local start-lambda` and calling a lambda (through the api or using the aws cli/sdks), we will now spin up a container but name it based on the logical Id. After the function executes, the container will be stopped (previously it was deleted). The next invoke of that function, will be directed to the stopped container, giving a 'warm' container invoke. This avoids the (expensive) cost of creating a container. In my rough testing (on my own machine), this reduced invoke time by almost 50%; 4.1s to 1.9s for the full command and 2.5s to 1.3s for the time around all the docker calls (create, invoke, delete containers).

Caveats:
* Since we name the container based on the logical id (function_name in the code), we can only have one container up at a time.
* We delete containers in a finally block (not trilled about this), so any failure will crash the cli.
* I imported some docker stuff into the services files, which leaks the docker into those areas (in my opinion, it doesn't belong there).

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


j:
Unless I did it wrong (probably), I didn't see a speed boost from a node.js app.  But it could be that I'm mounting my whole project directory?  Not sure.

jfuss:
@j The first invoke will still be 'slow'. I didn't add starting up all containers on command run. So we have the same behavior as before on the first invoke, but invokes after should be much faster. I did this to make the start up cost not huge and unless you are constantly changing your template, you can 'reload' code changes by invoking again. Let me know if you have further questions or are still not seeing any speed boost.

samcli/local/docker/manager.py
```diff
@@ -99,7 +99,15 @@ def run(self, container, input_data=None, warm=False):
...
+
+        :param samcli.local.docker.container.Container container: Container to stop
+        """
+        container.stop()
+
+    def delete(self, container):
+        """
+        Delete the container
 
         :param samcli.local.docker.container.Container container: Container to stop
```
dsanders11: Container to stop => Container to delete

lopun:
Is there any working progress on this PR? I'm really looking forward for this feature to be deployed :smile:

yang-wei:
Is there any working progress on this PR? I'm really looking forward for this feature to be deployed 😄

c2tarun:
Hi,

We are working on other items currently. We do plan to work on this feature (faster invokes) but do not have an eta now. This PR is left open as a base to when this worked gets picked again.

Thanks

jfuss:
Closing in favor of the newer pr: #2383

PR #1307, opened 2019-07-31
fix: fix unit tests so that they pass on windows

TheSriram:
first pass at getting unit tests to pass on windows (with Appveyor)

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/commands/local/lib/swagger/test_reader.py
+++ b/tests/unit/commands/local/lib/swagger/test_reader.py
@@ -184,7 +184,7 @@ class TestSamSwaggerReader_download_swagger(TestCase):
         expected = "parsed result"
         yaml_parse_mock.return_value = expected
 
-        with tempfile.NamedTemporaryFile(mode='w') as fp:
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as fp:
             filepath = fp.name
 
             json.dump(data, fp)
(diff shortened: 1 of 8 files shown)


TheSriram:
> 
> 
> Tests are failing on Linux. Py3 actually. Can you fix them?

Yeah, pushing new set up. Tried to use appveyor for checking the results, but decided to develop on a windows machine and re-submit the PR instead.

tests/unit/commands/local/lib/swagger/test_reader.py
```diff
@@ -184,7 +184,7 @@ def test_must_read_from_local_file(self, yaml_parse_mock):
         expected = "parsed result"
         yaml_parse_mock.return_value = expected
 
-        with tempfile.NamedTemporaryFile(mode='w') as fp:
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as fp:
```
TheSriram: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

```
This function operates exactly as TemporaryFile() does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object. Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed. The returned object is always a file-like object whose file attribute is the underlying true file object. This file-like object can be used in a with statement, just like a normal file.
```

PR #1315, opened 2019-08-05
Commits since 0.19.0 branch freeze

awood45:
We were keeping changes since we froze the develop branch for the 0.19.0 release in this branch. Now that 0.19.0 is out, returning these to the develop branch.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,16 +1,16 @@
 six~=1.11.0
 chevron~=0.12
-click~=6.7
+click~=7.0
 enum34~=1.1.6; python_version<"3.4"
 Flask~=1.0.2
 boto3~=1.9, >=1.9.56
-PyYAML~=3.12
+PyYAML~=5.1
 cookiecutter~=1.6.0
 aws-sam-translator==1.10.0
-docker~=3.7.0
+docker~=4.0
 dateparser~=0.7
 python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
 requests==2.22.0
-serverlessrepo==0.1.8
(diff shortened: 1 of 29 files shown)


requirements/base.txt
```diff
@@ -1,16 +1,16 @@
 six~=1.11.0
 chevron~=0.12
-click~=6.7
+click~=7.0
 enum34~=1.1.6; python_version<"3.4"
 Flask~=1.0.2
 boto3~=1.9, >=1.9.56
-PyYAML~=3.12
+PyYAML~=5.1
```
milind-shakya-sp: 👍

jfuss:
And Py2 bites again.... 

`os.symlink` is not available in python2.7's standard library: https://docs.python.org/2.7/library/os.html#os.symlink. This was added in #1140 to address the symlink issue with layers. To unblock this PR, which has a lot of other goodies, I vote to revert #1140 and then take another pass at it separately.

bubba-h57:
Aww man, people still use Windows? :-) Any appetite for a switch, if windows+2.7 then the old method, otherwise, *respect my authorit (symlinks)* :-)

jfuss:
@bubba-h57 Let me have a conversation with the team. I am totally for the switch (mainly because Py2.7 is being removed in Oct as a support Python version), but want to make sure the team is on the same page first. Will respond back here once I have that conversation.

jfuss:
@bubba-h57 The "keep old behavior with Windows+2.7 only" is good. Are you willing to update and push a new PR with this?

bubba-h57:
Certainly. I'll have something updated by the end of the week.

PR #1319, opened 2019-08-07
(feat) Add --persist-containers flag to start-api

frankh:
There's still some work to do for this, e.g. Design doc(?), more tests, and documentation

I want to see if this is something that would be considered to be merged before I spend too much time on it

There exists a PR already for a similar feature (https://github.com/awslabs/aws-sam-cli/pull/1305), however this has much better performance (roughly 0.2s per invocation, whereas using docker start instead was between 0.5 and 1.5s for me)

*Issue #, if available:*
#239 

*Description of changes:*
Adds `--persist-containers` as a flag to start api.

This causes lambda containers to stay running (with `sleep infinity`) the whole time, and lambdas are invoked by calling `exec` on the containers with the lambda handler. This means we only pay the container startup cost once (the first time the lambda is run), instead of at every request. Anecdotally this lowers runtime of a hello world lambda from ~3s to ~0.2s.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -188,6 +188,22 @@ class Container(object):
         # Start the container
         real_container.start()
 
+    def exec_run(self, cmd, stdout=None, stderr=None):
+        if not self.is_created():
+            raise RuntimeError("Container does not exist. Cannot exec in this container")
+
+        real_container = self.docker_client.containers.get(self.id)
+        if real_container.status != 'running':
+            raise RuntimeError("Container is not running, cannot exec in this container.")
+
+        # demux=True splits output into (stdout, stderr)
+        result = real_container.exec_run(cmd, demux=True)
+        container_stdout, container_stderr = result.output
+        if stdout:
+            stdout.write(container_stdout)
+        if stderr:
+            stderr.write(container_stderr)
+
     def wait_for_logs(self, stdout=None, stderr=None):
 
         # Return instantly if we don't have to fetch any logs
@@ -226,6 +242,10 @@ class Container(object):
 
             with tarfile.open(fileobj=fp, mode='r') as tar:
                 tar.extractall(path=to_host_path)
+
+    def get_image_entrypoint(self):
+        real_container = self.docker_client.containers.get(self.id)
+        return real_container.image.attrs['ContainerConfig']['Entrypoint']
 
     @staticmethod
     def _write_container_output(output_itr, stdout=None, stderr=None):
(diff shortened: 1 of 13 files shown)


samcli/local/docker/container.py
```diff
@@ -185,8 +190,27 @@ def start(self, input_data=None):
...
+        if not self.is_created():
+            raise RuntimeError("Container does not exist. Cannot exec in this container")
+
+        real_container = self.docker_client.containers.get(self.id)
+        if real_container.status != 'running':
+            raise RuntimeError("Container is not running, cannot exec in this container.")
+
+        entrypoint = self._entrypoint or []
+        # demux=True splits output into (stdout, stderr)
+        # entrypoint and cmd are both lists, concaternating them works as expected.
```
dsanders11: Looks like there's a minor typo: concaternating => concatenating.

emorneau:
Question, how can one set this up. I have sam baseline working. Any step by step instruction would be great so I can review/approve. Would love this sam warm lambda testing framework.

emorneau:
Hi @frankh, would it be possible to add this feature to "start-lambda"? I think you already most of the code right?

emorneau:
> Hi @frankh, would it be possible to add this feature to "start-lambda"? I think you already have most of the code to make it work. I decided to modify the aws-sam-cli/samcli/commands/local/start_lambda/cli.py to be similar to what start_api looks like and I was able to run the start lambda and run one lambda invocation but the second failed with _"TypeError: 'NoneType' does not have the buffer interface"_ which I do not know what it is . See in attachment:
> [execution.txt](https://github.com/awslabs/aws-sam-cli/files/3522268/execution.txt)
> 
> Any help would be greatly appreciated

@frankh you agree, any ETA?

frankh:
@emorneau it's done now, I moved the flag to the common options for all invoke commands

This has the possibly undesirable side-effect that the `sam invoke` command now accepts it but it doesn't behave as expected (the container will be cleaned up when sam exits, which happens immediately after the invocation)

jfuss:
@frankh and @emorneau I have had some discussions with the team on this. I think the best way forward is:

1. Finish #1305, which has some updates on how we can better manage the container lifecycle for the commands that spin up an endpoint (`start-api` and `start-lambda`). 
2. Update this PR with the container management.

Together (#1305 and this PR) will really help in performance. I have talked with @frankh on slack a little about this. I think my biggest concern with the sleep is how it plays with 'provided' runtimes. I need to do some deeper investigation here to really understand but want to focus on getting #1305 first.

wmonk:
Do you think this'll get any movement soon?

jfuss:
After further discussion, we do not think this PR is currently needed. We still want to support faster invokes but do not think a flag is needed to enable it. Between #1305 and recent changes to docker-lambda. If those do not end up being enough, we can revisit adding an option to the CLI.

Given we currently want to enable this without the option, I am going to close this PR.

PR #1325, opened 2019-08-08
Add boto3 client telemetry logging

viksrivat:
*Issue #, if available:*

*Description of changes:*
Add extra user_agent information to boto3 commands when telemetry is enabled.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/metrics.py
+++ b/samcli/lib/telemetry/metrics.py
@@ -2,23 +2,24 @@
 Provides methods to generate and send metrics
 """
 
-
 import platform
 import logging
 
 from timeit import default_timer
 
+import boto3
+from botocore.config import Config
+
 from samcli.cli.context import Context
 from samcli.commands.exceptions import UserException
 from samcli.cli.global_config import GlobalConfig
 from .telemetry import Telemetry
-
+from samcli import __version__
 
 LOG = logging.getLogger(__name__)
 
 
 def send_installed_metric():
-
     LOG.debug("Sending Installed Metric")
 
     telemetry = Telemetry()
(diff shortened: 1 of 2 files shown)


samcli/lib/telemetry/metrics.py
```diff
@@ -98,6 +99,17 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+def sam_boto3_client(*args, **kwargs):
+    user_agent_extra = "aws-sam-{}".format(__version__)
+    if _telemetry_enabled():
+        installation_id = GlobalConfig().installation_id
+        user_agent_extra += installation_id
```
TheSriram: should we use another '-' delimiter when you join them? also might be better to just use 
```
'-'.join() instead of +
```

samcli/lib/telemetry/metrics.py
```diff
@@ -98,6 +99,17 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+def sam_boto3_client(*args, **kwargs):
+    user_agent_extra = "aws-sam-{}".format(__version__)
+    if _telemetry_enabled():
+        installation_id = GlobalConfig().installation_id
+        user_agent_extra += '-'.join([user_agent_extra, installation_id])
+    return boto3.client(*args, config=Config(user_agent_extra=user_agent_extra), **kwargs)
```
TheSriram: Thanks for this change, can you add a unit test here? to see what are the actual arguments boto3.client gets called with.

sriram-mv:
#1757 addresses this.

Issue #1336, opened 2019-08-14
sam fails to build when build-dir or template-file exists on non-C drive


feldschlachtiv:
### Description
Sam fails when template or build-dir exists on non-C drive, and because of that pycharm cannot run lambda function locally.

Looking at SAM source code, it seems sam tries to use relative path on L161 in command.py. I guess this is to shorten the command to avoid hitting command length limit, however, it doesn't work when either ctx.build_dir or ctx.output_template_path is on non-C drive as there's no relative path exist, assuming sam is installed in C drive.

>            msg = gen_success_msg(os.path.relpath(ctx.build_dir),
>                                  os.path.relpath(ctx.output_template_path),
>                                  os.path.abspath(ctx.build_dir) == os.path.abspath(DEFAULT_BUILD_DIR))

I replaced the two os.path.relpath calls above with os.path.abspath, and sam succeeded to build my function without error (and hence pycharm was able to run the function locally)

### Steps to reproduce
Run the following command in windows 10 command prompt. (content of template.yaml is the "Hello World" one)
sam build --template N:\projects\git\kuraberu\template.yaml --build-dir N:\projects\git\kuraberu\.aws-sam\build HelloWorldFunction --debug

### Observed result

2019-08-14 00:56:25 Changing event name from creating-client-class.iot-data to creating-client-class.iot-data-plane
2019-08-14 00:56:25 Changing event name from before-call.apigateway to before-call.api-gateway
2019-08-14 00:56:25 Changing event name from request-created.machinelearning.Predict to request-created.machine-learning.Predict
2019-08-14 00:56:25 Changing event name from before-parameter-build.autoscaling.CreateLaunchConfiguration to before-parameter-build.auto-scaling.CreateLaunchConfiguration
2019-08-14 00:56:25 Changing event name from before-parameter-build.route53 to before-parameter-build.route-53
2019-08-14 00:56:25 Changing event name from request-created.cloudsearchdomain.Search to request-created.cloudsearch-domain.Search
2019-08-14 00:56:25 Changing event name from docs.*.autoscaling.CreateLaunchConfiguration.complete-section to docs.*.auto-scaling.CreateLaunchConfiguration.complete-section
2019-08-14 00:56:25 Changing event name from before-parameter-build.logs.CreateExportTask to before-parameter-build.cloudwatch-logs.CreateExportTask
[...]

PR #1337, opened 2019-08-14
Try relpath first, if failed use the user specified path

feldschlachtiv:
For compatibility with Windows. relpath() fails when the specified path is on a different drive.

*Issue #, if available:* #1336

*Description of changes:* use abspath() instead of relpath(), so cli command succeeds even when running on windows where sam and build_dir/template are located on different drive.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -158,8 +158,8 @@ def do_cli(function_identifier,  # pylint: disable=too-many-locals
 
             click.secho("\nBuild Succeeded", fg="green")
 
-            msg = gen_success_msg(os.path.relpath(ctx.build_dir),
-                                  os.path.relpath(ctx.output_template_path),
+            msg = gen_success_msg(os.path.abspath(ctx.build_dir),
+                                  os.path.abspath(ctx.output_template_path),
                                   os.path.abspath(ctx.build_dir) == os.path.abspath(DEFAULT_BUILD_DIR))
 
             click.secho(msg, fg="yellow")
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -158,8 +158,19 @@ def do_cli(function_identifier,  # pylint: disable=too-many-locals
 
             click.secho("\nBuild Succeeded", fg="green")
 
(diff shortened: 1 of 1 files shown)


samcli/commands/build/command.py
```diff
@@ -158,8 +158,19 @@ def do_cli(function_identifier,  # pylint: disable=too-many-locals
...
-                                  os.path.relpath(ctx.output_template_path),
+            # try to use relpath so the command is easier to understand, however,
+            # under Windows, when SAM and (build_dir or output_template_path) are
+            # on different drive, relpath() fails.
+            try:
+                build_dir = os.path.relpath(ctx.build_dir)
+                output_template_path = os.path.relpath(ctx.output_template_path)
+            except ValueError:
+                LOG.debug("Failed to retrieve relpath - using abspath instead")
+                build_dir = os.path.abspath(ctx.build_dir)
```
sanathkr: Since we got a ValueError computing relative path, I would just use the original path without computing absolute path.

samcli/commands/build/command.py
```diff
@@ -158,8 +158,19 @@ def do_cli(function_identifier,  # pylint: disable=too-many-locals
 
             click.secho("\nBuild Succeeded", fg="green")
 
-            msg = gen_success_msg(os.path.relpath(ctx.build_dir),
-                                  os.path.relpath(ctx.output_template_path),
+            # try to use relpath so the command is easier to understand, however,
+            # under Windows, when SAM and (build_dir or output_template_path) are
+            # on different drive, relpath() fails.
+            try:
+                build_dir = os.path.relpath(ctx.build_dir)
```
sanathkr: nit: can you rename `build_dir` and `output_template_path` to be something more descriptive? I just don't want someone to use these values without knowing what it means.

feldschlachtiv:
@sanathkr - Thank you for your comment. I think your suggestion makes sense, so I made a couple of changes. Let me know your thoughts about it - especially the renaming looks in-line with what you meant?

sanathkr:
Appreciate the changes. Approved.

PR #1343, opened 2019-08-15
bugfix/integrate_intrinsics

viksrivat:
*Issue #, if available:*
There was a bug when dealing with layers and intrinsics. This is an update to add tests and to address it. 
*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/cfn_api_provider.py
+++ b/samcli/commands/local/lib/cfn_api_provider.py
@@ -107,12 +107,9 @@ class CfnApiProvider(CfnBaseApiProvider):
         stage_name = properties.get("StageName")
         stage_variables = properties.get("Variables")
 
-        # Currently, we aren't resolving any Refs or other intrinsic properties that come with it
-        # A separate pr will need to fully resolve intrinsics
         logical_id = properties.get("RestApiId")
         if not logical_id:
             raise InvalidSamTemplateException("The AWS::ApiGateway::Stage must have a RestApiId property")
-
         rest_api_resource_type = resources.get(logical_id, {}).get("Type")
         if rest_api_resource_type != CfnApiProvider.APIGATEWAY_RESTAPI:
             raise InvalidSamTemplateException(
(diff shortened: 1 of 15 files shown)


viksrivat:
The integrate_intrinsics pr failed because LambdaLayers were handling Refs in a different way than the default resolver. I now handle LambdaLayers as a special case since unresolved Refs become {"Ref": "logical_id"}.

Before: SamTranslator was run and left attributes such as {"Ref": "value"} -> {"Ref": "value"}. The translator would leave the following resource as is. 
```yaml
StageTest:
     Properties:
          RestApiId: !Ref Value
``` 
However, we need to resolve these attributes in order to fully resolve intrinsics. After running a pass through the SamTranslator then IntrinsicResolver. Now, what we do is convert it to  
```yaml
StageTest:
     Properties:
          RestApiId: Value
``` 
This is because SamTranslator resolves a ref such as  {"Ref": "AWS::Region"} to {"Ref": "us-east-1"}. It also replaces Global Attributes directly into the relevant resources. However, we want {"Ref": "AWS::Region"} to become "us-east-1". Thus, by default we take {"Ref": "value"} -> "value". 

To provide support for the special LambdaLayer case, I extended the IntrinsicSymbolTable to use
```
        {
            "AWS::ApiGateway::RestApi": {
                "RootResourceId": "/"  # It usually used as a reference to the parent id of the RestApi,
            },
            "AWS::Lambda::LayerVersion": {
                IntrinsicResolver.REF: lambda logical_id: {IntrinsicResolver.REF: logical_id}
            },
            "AWS::Serverless::LayerVersion": {
                IntrinsicResolver.REF: lambda logical_id: {IntrinsicResolver.REF: logical_id}
            }
        }
```

In terms of dealing and managing intrinsics cases such as this, this implementation is not inclusive of all ways to process Refs and Fn::GetAtt. There doesn't seem to be a good standard that CloudFormation has set for all Refs and Fn::GetAtts. We need to either add special support for more types as time goes on or find another way to deal with maintaining intrinsics in the longer term.

PR #1363, opened 2019-08-22
Loosen requests version bounds

jonringer:
*Issue #, if available:*
Nope

*Description of changes:*
Use compatible versions of requests instead of pinning. Requests respects SemVer according to https://2.python-requests.org/en/master/dev/philosophy/#semantic-versioning , so there's no need to pin to a specific version. Anything in the >=2.22,<3 range should be compatible.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,6 +11,6 @@ docker~=4.0
 dateparser~=0.7
 python-dateutil~=2.6
 pathlib2~=2.3.2; python_version<"3.4"
-requests==2.22.0
+requests~=2.22
 serverlessrepo==0.1.9
 aws_lambda_builders==0.3.0


sanathkr:
We pin to specific versions for 1) determinism of builds and 2) test & validate SAM CLi works with this version. 

Can you explain why you want to expand this?

jonringer:
> 1) determinism of builds and 2) test & validate SAM CLi works with this version.

For tracing errors back to a given dependency, this makes sense. However, for many people, they most likely aren't just installing aws-sam-cli alone, but rather as a series of other packages. If someone has an environment where another package does similar pinning to requests==2.21, then they will get an error.

How about instead i talk about the benefits of using compatible versions with pip:
- Pip will choose the latest available version (in the version bound constraint) so you should be getting the most-up-to-date package
- Dont have to worry about CVE's which were patched in later versions and get PR's trying to fix them like https://github.com/awslabs/aws-sam-cli/pull/1200
- Version breathing room so that some flexibility allows you to play nicely with other packages. (Another package may have `requests~=2.23` when it gets released)

Not all packages follow SemVer, however, requests does, so we can leverage some of those guarantees of compatibility.

As an aside, requests is the only non-aws dependency that is pinned:
```
six~=1.11.0
chevron~=0.12
click~=7.0
enum34~=1.1.6; python_version<"3.4"
Flask~=1.0.2
boto3~=1.9, >=1.9.56
PyYAML~=5.1
cookiecutter~=1.6.0
aws-sam-translator==1.13.1
docker~=4.0
dateparser~=0.7
python-dateutil~=2.6
requests==2.22.0
serverlessrepo==0.1.9
aws_lambda_builders==0.3.0
```

My original use case was when packaging with nixpkgs, we supply the version that is available (usually the latest). And since the requests is pinned, we have to adjust for the bumps.

Examples:
- https://github.com/NixOS/nixpkgs/pull/62011/files#diff-ebb3457fdd733e81337d6b1c75aee925R60
- https://github.com/NixOS/nixpkgs/pull/67267/files#diff-ebb3457fdd733e81337d6b1c75aee925R59

jfuss:
@jonringer We did some deeper thinking and discussions about the pinning and how we want to handle this. What we ended up on is creating a new `requirements.txt` file that we could solely use for our 'binary' distributions. What this gives is people using pip (even though strongly advise not to), can still install through that for now and we still get a way to create a full reproducible build.

We are looking to invest more into our installations so hopefully Nix will be able to use some non-pip way in the future.

jonringer:
now have merge conflicts, and I've lost interest in up-streaming semver versioning

SnarfleBunny:
FWIW I am having many issues installing a version of the sam CLI tool that understands my authorizers work and at this point I am giving up on SAM and moving efforts to terraform where I don't need to worry that the command-line tool wont work (because it's a simple binary). I spent hours on this problem as a user of sam. I am not a package manager for this project and the notion that users don't have heterogeneous environments or multiple projects using different versions of things including python is not true for all of my coworkers at the last three employers.

mbbx6spp:
@jfuss I wasted hours from bad error messages from sam because I was on an old version, only because the newer version that support authorizers can't run for me. Can you switch to building fully self-contained executables to avoid this? It is a CLI tool after all and users shouldn't have to care exactly how the CLI tools they use are packaged.

jonringer:
Python doesn't have an officially supported way of creating an executable, and the "bundlers" that I'm aware of will create massive binaries.

Unfortunately python's module system is just "blindly import stuff at runtime", which means you can't really separate out what's exposed. The bandaid is to use "environments" (conda/venv) to at least mitigate the packages to a specific project, but if you want a global install, then you need to "play well" with the rest of the python exosystem

PR #1391, opened 2019-08-30
feat: Provide pinned requirements to allow for reproducible builds

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


TheSriram:
Closing https://github.com/awslabs/aws-sam-cli/pull/1369 ,as this PR supersedes it.

jfuss:
`boto3` is going to be a pain in this model since it is released so often. For now, I am going to add a lesser check on `boto3` in our checking script. I am basically re-implementing how a `.lock` file works and time is better spent moving over to a tool that does that for us but doing this in the short term seems ok.

scripts/check-isolated-needs-update.py
```diff
@@ -0,0 +1,55 @@
...
+        if line:
+            # remove markers from the line, which are seperated by ';'
+            pkgs_versions.append(line.split(";")[0])
+
+    return pkgs_versions
+
+# Don't try and compare the isolated list with the Python2 version. SAM CLI installers
+# all use Python3.6+ and Python2.7 is going EOL
+if sys.version_info[0] < 3:
+    sys.exit(0)
```
TheSriram: 🥇

scripts/check-isolated-needs-update.py
```diff
@@ -0,0 +1,58 @@
...
+
+for installed_pkg_version in all_installed_pkgs_list:
+    for base_req in base_req_list:
+        # a base requirement can be defined with different specifiers (>, <, ==, etc.). Instead of doing tons of string parsing,
+        # brute force the check by assuming the installed_pkgs will have == as a specifier. This is true due to how pip freeze
+        # works. So check to make sure the installed pakcage we are looking at is in the base.txt file, if so make sure the
+        # full requirement==version is within the isolated list.
+        installed_pkg = installed_pkg_version.split("==")[0]
+        # There is a py library we use but due to how we are comparing requirements, we need to handle this as a special case. :(
+        if installed_pkg not in ("py", "boto3") and base_req.startswith(installed_pkg):
```
TheSriram: Nit: can this be another file (list of ignored deps for fidelity checks)

jfuss: I am going to leave as a tuple instead of a file. This is a temporary solution to make sure builds going forward are deterministic. I want to move us into a tool that does the lockfile for us, which is way better than this script :).

Issue #1381, opened 2019-08-28
sam local does not accept additional quotes in CORS configuration 


stalar:
### Description

According to https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#cors-configuration CORS configuration strings should have an extra pair of quotes like this:

    Cors:
      AllowMethods: "'OPTIONS,GET,POST,DELETE'"
      AllowHeaders: "'X-Forwarded-For'"
      AllowOrigin: "'*'" 

Running `sam local start-api` with a template containing the above gives me: `Error: The method 'OPTIONS is not a valid CORS method`. (notice the dangling quote).

### Steps to reproduce

Add the Cors configuration above to an API definition. Run `sam local start-api`.

Sample template: 

    AWSTemplateFormatVersion: '2010-09-09'
    Transform: AWS::Serverless-2016-10-31
    Resources:
      HelloApi:
        Type: AWS::Serverless::Api
        Properties:
          StageName: prod
          Cors: 
            AllowOrigin: "'*'"
            AllowMethods: "'OPTIONS,HEAD,GET,PUT,POST'"
            AllowHeaders: "'Content-Type,X-Amz-Date,Authorization,X-Api-Key,X-Amz-Security-Token'"
      HelloWorldFunction:
        Type: AWS::Serverless::Function
        Properties:
          CodeUri: hello-world/
          Handler: app.lambdaHandler
          Runtime: nodejs10.x
          Events:
            HelloWorld:
              Type: Api
              Properties:
                Path: /hello
                Method: get
                RestApiId: 
                  Ref: HelloApi

### Observed result

    $ sam --debug local start-api
    Error: The method 'OPTIONS is not a valid CORS method

If I remove the extra singe quotes in `AllowMethods`, `sam local start-api` works as expected, but then I cannot deploy the template.

### Expected result

Expected a running API with CORS enabled.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS 10.14.6
2. `sam --version`: SAM CLI, version 0.21.0

TheSriram:
Ah, good catch! Looks like we actually test similar to way you reported it. https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/testdata/start_api/swagger-template.yaml

iceback:
sam --version
SAM CLI, version 0.19.0
on ubuntu18.04
[...]

PR #1395, opened 2019-09-03
Require Cors header values to be specified with double quotes.

stalar:
*Issue #, if available:* #1381

*Description of changes:*

Require Cors header values to have double quoting for start-api. If quotes are missing, fail with an exception.

Verified by running the template from #1381 both locally and deployed.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


TheSriram:
Have re-kicked the build, somehow it didnt start, will review this once the builds are okay.

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -103,23 +103,58 @@ def extract_cors(self, cors_prop):
...
             cors = Cors(
-                allow_origin=cors_prop.get("AllowOrigin"),
-                allow_methods=allow_methods,
-                allow_headers=cors_prop.get("AllowHeaders"),
-                max_age=cors_prop.get("MaxAge"),
+                allow_origin=allow_origin, allow_methods=allow_methods, allow_headers=allow_headers, max_age=max_age
             )
         elif cors_prop and isinstance(cors_prop, string_types):
+            allow_origin = cors_prop
+            if allow_origin[0] != "'" or allow_origin[-1] != "'":
```
TheSriram: if this is a string, can we just do startswith or endswith instead?

samcli/commands/local/lib/sam_api_provider.py
```diff
@@ -103,23 +103,58 @@ def extract_cors(self, cors_prop):
...
+        cors_dict : dict
+            Resource properties for Cors
+
+        Return
+        ------
+        A string with the extra quotes removed
+        """
+        prop = cors_dict.get(prop_name)
+        if prop:
+            if (not isinstance(prop, string_types)) or prop[0] != "'" or prop[-1] != "'":
```
TheSriram: Similarly here

samcli/commands/local/lib/sam_api_provider.py
stalar: Will fix!

jfuss:
@stalar Thanks for the PR. Things look good. I did add two additional unit tests to verify exceptions are thrown correctly. 

@TheSriram Can you take a peek at this? I think its good to go but wanted another approval since I added some tests.

Issue #1157, opened 2019-04-29
sam build --build-dir . will nuke current working directory


m1keil:
### Description
Executing `sam build --build-dir .` will fail and result in **removal of all files in current working directory**.

### Steps to reproduce
```bash
sam init -r python3.7
cd sam-app
sam build --build-dir .
```

### Observed result
```
$ sam build --debug --build-dir .
2019-04-30 02:18:11 Using SAM Template at /private/tmp/sam-app/template.yaml
2019-04-30 02:18:11 Changing event name from creating-client-class.iot-data to creating-client-class.iot-data-plane
2019-04-30 02:18:11 Changing event name from before-call.apigateway to before-call.api-gateway
2019-04-30 02:18:11 Changing event name from request-created.machinelearning.Predict to request-created.machine-learning.Predict
2019-04-30 02:18:11 Changing event name from before-parameter-build.autoscaling.CreateLaunchConfiguration to before-parameter-build.auto-scaling.CreateLaunchConfiguration
2019-04-30 02:18:11 Changing event name from before-parameter-build.route53 to before-parameter-build.route-53
2019-04-30 02:18:11 Changing event name from request-created.cloudsearchdomain.Search to request-created.cloudsearch-domain.Search
2019-04-30 02:18:11 Changing event name from docs.*.autoscaling.CreateLaunchConfiguration.complete-section to docs.*.auto-scaling.CreateLaunchConfiguration.complete-section
2019-04-30 02:18:11 Changing event name from before-parameter-build.logs.CreateExportTask to before-parameter-build.cloudwatch-logs.CreateExportTask
2019-04-30 02:18:11 Changing event name from docs.*.logs.CreateExportTask.complete-section to docs.*.cloudwatch-logs.CreateExportTask.complete-section
2019-04-30 02:18:11 Changing event name from before-parameter-build.cloudsearchdomain.Search to before-parameter-build.cloudsearch-domain.Search
2019-04-30 02:18:11 Changing event name from docs.*.cloudsearchdomain.Search.complete-section to docs.*.cloudsearch-domain.Search.complete-section
2019-04-30 02:18:11 Changing event name from creating-client-class.iot-data to creating-client-class.iot-data-plane
2019-04-30 02:18:11 Changing event name from before-call.apigateway to before-call.api-gateway
2019-04-30 02:18:11 Changing event name from request-created.machinelearning.Predict to request-created.machine-learning.Predict
[...]

PR #1403, opened 2019-09-12
fix(build): Disallow --build-dir to resolve to cwd

jfuss:
*Issue #, if available:*
#1157 

*Description of changes:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -85,10 +85,7 @@ class BuildContext(object):
         build_path = pathlib.Path(build_dir)
 
         if os.path.abspath(str(build_path)) == os.path.abspath(str(pathlib.Path.cwd())):
-            exception_message = "'build-dir' should not be the current workding directoy as the directory " \
-                "will be deleted. In the intention is to use the current working directory as the where the " \
-                "output should be place, either remove the 'build-dir' option and use the default or create a " \
-                "sub-directoy and pass that path to the option instead."
+            exception_message = "Failing build: Running a build with build-dir as current working directory is extremely dangerous since the build-dir contents is first removed. This is no longer supported, please remove the '--build-dir' option from the command to allow the build artifacts to be placed in the directory your template is in."
             raise InvalidBuildDirException(exception_message)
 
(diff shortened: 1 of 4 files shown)


samcli/commands/build/command.py
```diff
@@ -63,7 +63,7 @@
 @click.option('--build-dir', '-b',
               default=DEFAULT_BUILD_DIR,
               type=click.Path(file_okay=False, dir_okay=True, writable=True),  # Must be a directory
-              help="Path to a folder where the built artifacts will be stored")
+              help="Path to a folder where the built artifacts will be stored. If a clean build, this directory will be first removed.")
```
sanathkr: we always do a clean build. So the 'if' shouldn't be necessary

samcli/commands/build/build_context.py
```diff
@@ -83,6 +84,13 @@ def __exit__(self, *args):
     def _setup_build_dir(build_dir, clean):
         build_path = pathlib.Path(build_dir)
 
+        if os.path.abspath(str(build_path)) == os.path.abspath(str(pathlib.Path.cwd())):
+            exception_message = "'build-dir' should not be the current workding directoy as the directory " \
```
sanathkr: can we instead skip the "clean" if build-dir == cwd? It could give to a more intuitive experience

samcli/commands/build/command.py
jfuss: The wording was future looking. I can remove. Do we even mention clean build is all we support anywhere today?

samcli/commands/build/build_context.py
jfuss: Not sure I completely understand. My current understanding of the question is:

"Instead of always doing a "clean" build, we would not do a clean build if build-dir == cwd"?

If that is a fair rewording, then I don't think we should go down this route. Having a conditional do clean or not base upon a parameter that does not relate to a 'clean build' is not intuitive at all. This will make supporting clean builds more difficult in the future.

PR #45, opened 2017-08-14
Only use the last line of stdout for the JSON result

mhart:
The user's function may have emitted text directly to the Docker container's stdout using System.out.println or similar – this won't parse as JSON.

As the function result is always a JSON string, and it's the last thing written to stdout, then we just want to use the last non-empty line. This is the same functionality as that used by the docker-lambda test runner: https://github.com/lambci/docker-lambda/blob/92f59a60884a1809d94046e3555a39f6e0539eab/index.js#L59-L61

cc @sanathkr

PR #1428, opened 2019-09-24
fix: appveyor runs windows build tests per language

TheSriram:
*Description of changes:*

Enable separation of windows appveyor projects for running build integration tests per language. This allows to bring down build time per project, as the set of the dependencies installed is lower. The new gates will be non blocking in appveyor.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


appveyor-windows-build-dotnet.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - "echo %PATH%"
+    - "python --version"
+
+    # Upgrade setuptools, wheel and virtualenv
+    - "python -m pip install --upgrade setuptools wheel virtualenv"
+
+    # Create new virtual environment with chosen python version and activate it
+    - "rm -rf venv"
+    - "dir C:\\Python36-x64"
+    - "virtualenv -p %PYTHON_EXE% venv"
```
jfuss: Instead, you can add `PYTHON_HOME` to the PATH and then python will point to the correct version. Makes this a littler easier and then things like line #45 will work on the correct python version.

TheSriram: Adding to path was notoriously hard in https://github.com/awslabs/aws-sam-cli/pull/1269. There were a fair few commits in there to get binaries added to the path, and they didnt work as expected, so I skipped it this time.

TheSriram: Ported over those changes.

PR #1454, opened 2019-10-09
Interactive Init

awood45:
This is not quite ready to merge yet - I have tested code paths manually, but there is an expansion of unit and integration testing coverage required to provide a sufficient regression suite.

However, the code proper should be ready to review in tandem with these efforts.

**IMPORTANT NOTE**: As implemented, this depends on the Python 2.7 deprecation being complete.


samcli/commands/init/__init__.py
```diff
@@ -39,107 +55,57 @@
...
+    runtime,
+    dependency_manager,
+    output_dir,
+    name,
+    app_template,
+    no_input,
+    auto_clone=True,
+):
+    # check for mutually exclusive parameters
+    if location and app_template:
```
TheSriram: can potentially lift exception cases into decorators on `do_cli`

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,128 @@
...
+            runtime, dependency_manager
+        )
+        raise UserException(msg)
+
+    def _clone_repo(self):
+        shared_dir = global_cfg.config_dir
+        expected_path = os.path.normpath(os.path.join(shared_dir, self._repo_name))
+        if self._should_clone_repo(expected_path):
+            try:
+                subprocess.check_output(["git", "clone", self._repo_url], cwd=shared_dir, stderr=subprocess.STDOUT)
```
TheSriram: on windows you might need to look for git.exe or git.cmd, Not sure which.

samcli/commands/init/__init__.py
```diff
@@ -39,107 +55,57 @@
...
+            no_input = True
+            extra_context = {"project_name": name, "runtime": runtime}
+        if not output_dir:
+            output_dir = "."  # default - should I lift this to overall options and seed default?
+        do_generate(location, runtime, dependency_manager, output_dir, name, no_input, extra_context)
+    elif no_interactive:
+        error_msg = """
+ERROR: Missing required parameters, with --no-interactive set.
+
+Must provide one of the following required parameter combinations:
```
TheSriram: this is not clear on what constitues a combination, I would put valid combinations in a list, so that its easy to grok for the user.

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,128 @@
...
+        self.repo_path = None
+        self.clone_attempted = False
+        self._no_interactive = no_interactive
+        self._auto_clone = auto_clone
+
+    def prompt_for_location(self, runtime, dependency_manager):
+        options = self.init_options(runtime, dependency_manager)
+        choices = map(str, range(1, len(options) + 1))
+        choice_num = 1
+        for o in options:
```
TheSriram: for option in options, instead of `o`

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,128 @@
...
+                if overwrite:
+                    shutil.rmtree(expected_path)  # fail hard if there is an issue
+                    return True
+            self.repo_path = expected_path
+            return False
+        else:
+            if self._no_interactive:
+                return self._auto_clone
+            do_clone = click.confirm(
+                "This process will clone app templates from https://github.com/awslabs/aws-sam-cli-app-templates - is this ok?"
```
TheSriram: do we prompt even in a non interactive mode here?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+import click
+
+from samcli.local.common.runtime_template import RUNTIMES, RUNTIME_TO_DEPENDENCY_MANAGERS
+from samcli.commands.init.init_generator import do_generate
+from samcli.commands.init.init_templates import InitTemplates
+
+
+def do_interactive(location, runtime, dependency_manager, output_dir, name, app_template, no_input):
+    if location:
```
TheSriram: Can we convert if elses into: dictionary dispatch tables? https://medium.com/better-programming/dispatch-tables-in-python-d37bcc443b0b

samcli/commands/init/init_templates.py
awood45: We do not.

samcli/commands/init/init_templates.py
awood45: Good callout, adding an investigation item on this.

samcli/commands/init/__init__.py
awood45: That turned out to be a bit problematic when I tried it, as we don't always move to failure - rather, in some cases where we have missing required parameters we move to interactive, in other cases we fail. It's difficult to lift a partial of this out without making the code look more confusing to me.

samcli/commands/init/init_templates.py
awood45: Took a modified approach based on what we do for AWS CLI. On the one hand, we may want to abstract that to a utility, on the other hand, AWS CLI dependency is going away anyways.

samcli/commands/init/__init__.py
awood45: There are only two, what format did you suggest?

samcli/commands/init/__init__.py
```diff
@@ -28,118 +40,72 @@
...
+@click.option(
+    "--app-template",
+    help="Identifier of the managed application template you want to use. If not sure, call 'sam init' without options for an interactive workflow.",
+)
 @click.option(
     "--no-input",
     is_flag=True,
     default=False,
-    help="Disable prompting and accept default values defined template config",
+    help="Disable Cookiecutter prompting and accept default values defined template config",
```
TheSriram: can you be more verbose on what constitutes template config?

samcli/commands/init/__init__.py
TheSriram: [--name--runtime , --dependency-manager, --app-template] or [--location]. Just having all the `ands` together in `--name and --runtime and --dependency-manager and --app-template` made it difficult for me to spot the combination.

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+            templates = manifest_body.get(runtime)
+            if templates is None:
+                # Fallback to bundled templates
+                return self._init_options_from_bundle(runtime, dependency_manager)
+            if dependency_manager is not None:
+                templates_by_dep = filter(lambda x: x["dependencyManager"] == dependency_manager, templates)
+                return list(templates_by_dep)
+            return templates
+
+    def _init_options_from_bundle(self, runtime, dependency_manager):
```
TheSriram: bundle -> bundled

samcli/commands/init/init_templates.py
TheSriram: ah I misread the indentation.

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+        self.clone_attempted = False
+        self._no_interactive = no_interactive
+        self._auto_clone = auto_clone
+
+    def prompt_for_location(self, runtime, dependency_manager):
+        options = self.init_options(runtime, dependency_manager)
+        choices = map(str, range(1, len(options) + 1))
+        choice_num = 1
+        for o in options:
+            if o.get("displayName") is not None:
```
TheSriram: is displayname part of cookiecutter options? in the cookiecutter.json file. I see it used in the manifest in a test below. Never mind looks to be part of manifest.json

https://github.com/awslabs/aws-sam-cli-app-templates/blob/master/manifest.json

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def init_options(self, runtime, dependency_manager):
+        if self.clone_attempted is False:
+            self._clone_repo()
+        if self.repo_path is None:
+            return self._init_options_from_bundle(runtime, dependency_manager)
+        return self._init_options_from_manifest(runtime, dependency_manager)
+
+    def _init_options_from_manifest(self, runtime, dependency_manager):
+        manifest_path = os.path.join(self.repo_path, "manifest.json")
```
TheSriram: can we set manifest file as `manifest.json`, and reference that as self.manifest_file or self.manifest?

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+            if dependency_manager is not None:
+                templates_by_dep = filter(lambda x: x["dependencyManager"] == dependency_manager, templates)
+                return list(templates_by_dep)
+            return templates
+
+    def _init_options_from_bundle(self, runtime, dependency_manager):
+        for mapping in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values()))):
+            if runtime in mapping["runtimes"] or any([r.startswith(runtime) for r in mapping["runtimes"]]):
+                if not dependency_manager or dependency_manager == mapping["dependency_manager"]:
+                    mapping["appTemplate"] = "hello-world"  # when bundled, use this default template name
```
TheSriram: do we want the logic of falling back to the bundled be driven by manifest in an external repository? https://github.com/awslabs/aws-sam-cli-app-templates/blob/master/manifest.json#L102

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+    def _clone_repo(self):
+        shared_dir = global_cfg.config_dir
+        expected_path = os.path.normpath(os.path.join(shared_dir, self._repo_name))
+        if self._should_clone_repo(expected_path):
+            try:
+                subprocess.check_output(
+                    [self._git_executable(), "clone", self._repo_url], cwd=shared_dir, stderr=subprocess.STDOUT
+                )
+                self.repo_path = expected_path
+            except OSError:
```
TheSriram: do we catch permissions errors as well within OSError?

samcli/commands/init/init_templates.py
```diff
@@ -0,0 +1,150 @@
...
+                click.echo("WARN: Can't clone app repo, git executable not found.")
+            except subprocess.CalledProcessError as clone_error:
+                output = clone_error.output.decode("utf-8")
```
TheSriram: can we just look at exit codes instead?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -0,0 +1,55 @@
...
+        if valid_dep_managers is None:
+            dependency_manager = None
+        else:
+            dependency_manager = click.prompt(
+                "Dependency Manager", type=click.Choice(valid_dep_managers), default=valid_dep_managers[0]
+            )
+    templates = InitTemplates()
+    if app_template is not None:
+        location = templates.location_from_app_template(runtime, dependency_manager, app_template)
+        no_input = True
```
TheSriram: can no_input be moved out of the if else, we set it to true in both cases.

tests/unit/commands/init/test_templates.py
```diff
@@ -0,0 +1,59 @@
...
+            ]
+        }
+        manifest_json = json.dumps(manifest)
+
+        m = mock_open(read_data=manifest_json)
+        with patch("samcli.cli.global_config.GlobalConfig.config_dir", new_callable=PropertyMock) as mock_cfg:
+            mock_cfg.return_value = "/tmp/test-sam"
+            with patch("samcli.commands.init.init_templates.open", m):
+                location = it.location_from_app_template("ruby2.5", "bundler", "hello-world")
+                assert location == "/tmp/test-sam/aws-sam-cli-app-templates/mock-ruby-template"
```
TheSriram: self.assertEqual might be better. in all places where we assert.

awood45: Done

samcli/commands/init/interactive_init_flow.py
awood45: Refactored that out, good catch.

samcli/commands/init/init_templates.py
awood45: I am not sure that's possible. If we get to this piece of code, the manifest is not accessible. Or is this a separate scenario you're thinking about?

TheSriram: No, I was just thinking that falling back to bundled templates (within the same repo), required direction from a manifest file in another repo (app templates repo), introduces some coupling between the repos.

samcli/commands/init/init_templates.py
awood45: Per it's documentation, I believe that it should.

samcli/commands/init/init_templates.py
awood45: This follows the pattern we use for the AWS CLI, what are you proposing?

TheSriram: aah never mind, looks like CalledProcessError actually looks at exit codes natively :)

samcli/commands/init/__init__.py
```diff
@@ -2,24 +2,36 @@
...
 
 import click
 
-from samcli.cli.main import pass_context, common_options
+from samcli.cli.main import pass_context, common_options, global_cfg
+from samcli.commands.exceptions import UserException
+from samcli.local.common.runtime_template import RUNTIMES, SUPPORTED_DEP_MANAGERS
 from samcli.lib.telemetry.metrics import track_command
-from samcli.local.common.runtime_template import INIT_RUNTIMES, SUPPORTED_DEP_MANAGERS, DEFAULT_RUNTIME
+from samcli.commands.init.init_generator import do_generate
```
jfuss: Some of these imports should be pushed down into the `do_cli` function. This is to keep `--help` text load time low, otherwise we will be loading all of these modules.

awood45: Which imports?

awood45: This seems to create other problems, we can discuss in private.

Issue #1427, opened 2019-09-24
Function with 'Fn::Transform' as first events causes 'sam local start-api' to crash


koenaad:
**Description:**

If a template contains a Function, of which the top-most event is a `Fn::Transform` entry. Then `sam local start-api` fails to start with something like:

```
...
File "/usr/local/Cellar/aws-sam-cli/0.21.0/libexec/lib/python3.7/site-packages/samcli/commands/local/lib/sam_api_provider.py", line 191, in extract_routes_from_events
    for _, event in serverless_function_events.items():
AttributeError: 'str' object has no attribute 'items'
```

This template does not cause any issues with other commands, including `sam local invoke`, `sam package` and `sam deploy`.

**Steps to reproduce the issue:**
1. Given the template:
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Resources:
  Func1:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri:                      "./"
      Handler:                      "main"
      Runtime:                      "go1.x"
      Events:
        Api1:
          Type: Api
          Properties:
            Path:                   "/func1"
            Method:                 "GET"

  Func2:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri:                      "./"
      Handler:                      "main"
      Runtime:                      "go1.x"
      Events:
        Fn::Transform:
          Name: AWS::Include
          Parameters:
            Location:               "events.yaml"
```
2. Run `sam local start-api`

**Observed result:**
`sam local start-api` fails to start:

```
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 11, in <module>
    load_entry_point('aws-sam-cli==0.21.0', 'console_scripts', 'sam')()
  File "/usr/local/Cellar/aws-sam-cli/0.21.0/libexec/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/0.21.0/libexec/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/aws-sam-cli/0.21.0/libexec/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

Issue #1418, opened 2019-09-18
Command 'sam local start-api' fails when using AWS::Include transform to add a Swagger file for 'DefinitionBody'


sivarosh:
### Description

Command 'sam local start-api' fails when using AWS::Include transform to add a local swagger file for an AWS::Serverless::Api resource under the 'DefinitionBody' property.

This issue occurs with AWS SAM CLI v0.21, and does not occur with v0.19. I believe that this is different from the issue resolved by PR #1410 as the resolution of the 'Fn::Transform' intrinsic appears to cause the issue. Further details below. 

### Steps to reproduce

Template, Lambda function and Swagger file included at https://gist.github.com/sivarosh/bb7290c16fe15628c8105b842916f454

Install aws-sam-cli v0.21
Run command `sam local start-api`

Install aws-sam-cli v0.19
Run command `sam local start-api`

### Observed result
v0.21
Error Log : https://gist.github.com/sivarosh/905ce2b49a85b6584cd53580e62888c1#file-v0-21_error-log

v0.19
Successful Log : https://gist.github.com/sivarosh/905ce2b49a85b6584cd53580e62888c1#file-v0-21_error-log

### Investigation

In v0.19, when printing the `self.resources` object under `__init__` function of class `SamApiProvider` in `sam_api_provider.py`, we see the value:
`{'StageName': 'test', 'DefinitionBody': {'Fn::Transform': {'Name': 'AWS::Include', 'Parameters': {'Location': './swagger.yml'}}}}`

In v0.21, when printing the `self.resources` object under `__init__` function of class `ApiProvider` in `api_provider.py`, we see the value: 
`{'StageName': 'test', 'DefinitionBody': './swagger.yml'}`

In v0.21, it appears that the template dict which is fetched from `get_template `function of  `SamBaseProvider` class attempts to resolve all intrinsic functions by calling the `resolve_template` function of class `IntrinsicResolver` in `intrinsic_property_resolver.py`.

Under the `IntrinsicResolver` class, function `handle_fn_transform` transforms `'DefinitionBody': {'Name': 'AWS::Include', 'Parameters': {'Location': './swagger.yml'}}` to `'DefinitionBody': './swagger.yml'`

Due to this, when attempting to the `_read_from_definition_body` function in `reader.py` invokes `parse_aws_include_transform` which does not detect an AWS::Include transform as the template snippet has been replaced and the contents of the `DefinitionBody` are left as `./swagger.yml`

### Expected result

[...]

PR #1472, opened 2019-10-17
fix: Inline data that was defined with AWS::Include

jfuss:
*Issue #, if available:*
#1418 and possibly all of #1427 (haven't tested with that yet)

*Description of changes:*

`Fn::Transform` with `AWS::Include` allows for inlining data that is found in a separate file. The current implementation only took the value from the `Parameter.Location`, which is breaks `start-api`. This PR will take the value of `Parameter.Location` and try to load the local file and return that as the value of the `Fn::Transform`. If the file is not a valid local file, the `Fn::Transform` is left alone (say if the value is an s3 Uri). Our other code paths already handle these cases, so no special logic is needed, but we can expand this in the future to do all the right things.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -17,4 +17,4 @@ backports.tempfile==1.0
 pytest-xdist==1.20.0
 pytest-forked==1.0.2
 pytest-timeout==1.3.3
-pytest-rerunfailures==7.0
+pytest-rerunfailures==5.0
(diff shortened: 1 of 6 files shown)


dmitry-weirdo:
Will this work with setting a s3 url in `DefinitionBody` (containing the OAS)? 

Or should I use the local file for local `sam build` + `sam local start-api`?

PR #1477, opened 2019-10-17
Make --no-event the default for local invoke

awood45:
If no explicit `--event` is passed in, the default assumption is that no
event is desired. STDIN events must be explicit.

This is a breaking change.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/local/invoke/cli.py
```diff
@@ -14,14 +14,17 @@
...
+\b
+Invoking a Lambda function without an input event
+$ sam local invoke "HelloWorldFunction"\n
 \b
 Invoking a Lambda function using an event file
 $ sam local invoke "HelloWorldFunction" -e event.json\n
 \b
 Invoking a Lambda function using input from stdin
-$ echo '{"message": "Hey, are you there?" }' | sam local invoke "HelloWorldFunction" \n
+$ echo '{"message": "Hey, are you there?" }' | sam local invoke "HelloWorldFunction" --event - \n
```
TheSriram: does `--event -` automatically trigger reading from stdin?

jfuss: Yes. Pretty sure it is standard in unix.

http://click.palletsprojects.com/en/5.x/arguments/#file-arguments

samcli/commands/local/invoke/cli.py
```diff
@@ -31,11 +34,10 @@
     "--event",
     "-e",
     type=click.Path(),
-    default=STDIN_FILE_NAME,  # Defaults to stdin
     help="JSON file containing event data passed to the Lambda function during invoke. If this option "
-    "is not specified, we will default to reading JSON from stdin",
+    "is not specified, no event is assumed. Pass in the value '-' to input JSON via stdin",
```
jfuss: The default value is '-', so there is no need for them to pass '-' to the option `--event` is sufficient.

samcli/commands/local/invoke/cli.py
```diff
@@ -31,11 +34,10 @@
...
 )
-@click.option("--no-event", is_flag=True, default=False, help="Invoke Function with an empty event")
+@click.option("--no-event", is_flag=True, default=True, help="DEPRECATED: By default no event is assumed.", hidden=True)
```
jfuss: Do we have a plan to remove this option all together?

samcli/commands/local/invoke/cli.py
awood45: It does, I've tested this.

samcli/commands/local/invoke/cli.py
awood45: That's not the default behavior anymore. If you do not pass in `--event`, we will not trigger STDIN. That's the point of this change. Unless I'm missing something or made a mistake.

samcli/commands/local/invoke/cli.py
awood45: I'm open to that, question is if it's worth breaking people who are already using it. But by making it hidden, we're more or less de facto deleting it. The next step would be to drop values on the floor if someone goes to the trouble of doing `--no-no-event` or similar.

TheSriram: Can we actually add a line somewhere "DEPRECATED" onto stdout, before we start looking for values on stdin. Similair to how python2.7 was deprecated.

samcli/commands/local/invoke/cli.py
jfuss: My mistake. I thought the default value was still there (which makes no sense and it was removed in the diff). Please ignore my comment.

samcli/commands/local/invoke/cli.py
jfuss: You don't have `--no-no-event` defined. So customers shouldn't be doing that at least. (https://click.palletsprojects.com/en/7.x/options/#boolean-flags). 

I am ok not deleting it fully. You marked it as `DEPRECATED` in the help text, which would indicate we would remove it at some point. Maybe the right thing to do is not to that at all, and allow people to explicitly define `--no-event`. Once this is released, I am not sure why people would type more but it really doesn't hurt anything. If we are planning on removing completely, I would like us to have a way to do that was all. This is a non-blocking comment but worth thinking about.

awood45: I think if someone explicitly defines `--no-event` there is no harm done - we're basically making it a no-op regardless, and I believe the deprecation flag removes it from help text anyways.

So perhaps the best way forward is to just leave it in place, but have it rendered inert.

sparrowt:
The [documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html) still says:
> -e, --event PATH  ... If you don't specify this option, the default is reading the JSON from stdin.

which is no longer true, and it took me a while to find that this breaking change was the reason I could no longer invoke via stdin without explicitly specifying `--event -`.

Where is the best place to raise a ticket for updating the docs?

PR #1478, opened 2019-10-18
fix: add timeouts on docker local-invoke, start-api and start-lambda integration tests

TheSriram:
- cause the builds to fail on appveyor rather than halting the test
  forever, resulting in a lengthy build timeout.
- had to play around with requirements to be able to adopt `pytest-timeout`, also do we need `tox`?

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,13 +1,13 @@
 coverage==4.3.4
-tox==2.2.1
 pytest-cov==2.4.0
 # astroid > 2.0.4 is not compatible with pylint1.7
 astroid>=1.5.8,<2.1.0
 pylint==1.7.2
 
 # Test requirements
-pytest==3.1.0
-py==1.4.33
+pytest==3.6.0
+py==1.5.1
+pluggy==0.6.0
 mock==2.0.0
 parameterized==0.6.1
 pathlib2==2.3.2; python_version<"3.4"
@@ -16,3 +16,4 @@ futures==3.2.0; python_version<"3.2.3"
 backports.tempfile==1.0
 pytest-xdist==1.20.0
 pytest-forked==1.0.2
+pytest-timeout==1.3.3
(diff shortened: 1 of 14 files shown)


TheSriram:
Might need to adopt https://github.com/pytest-dev/pytest-rerunfailures as well.

requirements/dev.txt
```diff
@@ -1,13 +1,13 @@
 coverage==4.3.4
-tox==2.2.1
```
sanathkr: No we don't need tox. +1

requirements/dev.txt
```diff
@@ -1,13 +1,13 @@
...
 pytest-cov==2.4.0
 # astroid > 2.0.4 is not compatible with pylint1.7
 astroid>=1.5.8,<2.1.0
 pylint==1.7.2
 
 # Test requirements
-pytest==3.1.0
-py==1.4.33
+pytest==3.6.0
```
sanathkr: This might cause some weird test failures/warnings I think.

tests/integration/local/invoke/runtimes/test_with_runtime_zips.py
```diff
@@ -30,6 +31,7 @@ def setUp(self):
     def tearDown(self):
         os.remove(self.events_file_path)
 
+    @pytest.mark.timeout(timeout=300, method='thread')
```
sanathkr: is this seconds?

requirements/dev.txt
jfuss: Are you talking about removing py or the pytest upgrade?

tests/integration/local/invoke/runtimes/test_with_runtime_zips.py
TheSriram: yeah its seconds.

requirements/dev.txt
TheSriram: I had to upgrade pytest to even be able to use pytest-timeouts.

TheSriram:
Gonna increase the lambda timeouts on the integration test templates. Looks like 3 seconds is too low (sporadically) when running docker linux containers on appveyor windows.

PR #1479, opened 2019-10-22
Fix #1463. Allow to expose multiple ports when running in debug mode

### LINKED ISSUE #1463: Expose multiple ports through --debug-port when start lambda in debug mode
I would like to have multiple debug ports to be exposed on start lambda in debug mode.
JetBrains Rider IDE debugger requires two ports to be exposed to work properly (the process set up two sockets). We would like to have an option to start dotnet core docker container with multiple ports available.

### Proposal
It would be great to be able to specify a multiple coma-separated ports in `--debug-port` option, e.g. `--debug-port 5600, 5601`.

Add details of how to add this to the product.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)
No.

### Additional Details
None.

sdubov:
*Issue #, if available:*
#1463

*Description of changes:*
* Add ability to specify multiple ports to be exposed from docker instance when running lambda in debug mode
* Update and add new tests to verify multiple debug ports

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


designs/debug_mode_multiple_exposed_ports.md
```diff
@@ -0,0 +1,122 @@
...
+
+What will be changed?
+---------------------
+SAM CLI has a ``--debug-port`` parameter that provide a port. This parameter is stored in DebugContext object.
+``DebugContext`` should store an array of ports instead of one single port. This array should be transformed
+into a map containing each stored port when passing to docker container arguments.
+
+Success criteria for the change
+-------------------------------
+All ports (comma separated) specified in ``--debug-port`` SAM CLI parameter should be exposed by docker container.
```
jfuss: Why comma separated? This seems like a good use-case for multiple options: `--debug-port 5600 --debug-port 5601`.

sdubov: Good point. Updated with `multiple=true` option parameter.

samcli/local/docker/lambda_container.py
```diff
@@ -96,16 +96,21 @@ def _get_exposed_ports(debug_options):
         bind to same port both inside and outside the container ie. Runtime process is started in debug mode with
         at given port inside the container and exposed to the host machine at the same port
 
-        :param int debug_port: Optional, integer value of debug port
+        :param list(int) debug_port: Optional, List of integer values of debug ports
```
jfuss: This doc is wrong (was wrong previously). The parameter passed in is `debug_options` not the `debug_port`. Could you update this to be correct?

samcli/local/docker/lambda_container.py
```diff
@@ -96,16 +96,21 @@ def _get_exposed_ports(debug_options):
...
-        return {
-            # container port : host port
-            debug_options.debug_port: debug_options.debug_port
-        }
+        if not debug_options.debug_port:
+            return None
+
+        # container port : host port
+        ports_map = {}
+        for port in debug_options.debug_port:
```
jfuss: Can you update the docs for `debug_options.debug_port` to reflect changes to the type?

samcli/commands/local/cli_common/options.py
```diff
@@ -95,6 +95,8 @@ def invoke_common_options(f):
                 help="When specified, Lambda function container will start in debug mode and will expose this "
                 "port on localhost.",
                 envvar="SAM_DEBUG_PORT",
+                type=click.INT,
```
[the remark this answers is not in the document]
sdubov: No issues with passing integer as a string here. Looked to the `click` parameters type processing logic - it tries to case a value to integer. Verified it works well with strings unless you specify an invalid value, e.g. "abc" that should be an expected behavior.

Looked through `click` implementation for handling `multiple` flag - it wraps values into `tuple` if flag is set and return them or return an empty tuple if no values. So, I would expect we always have `tuple` after processing `--debug-port` option. Verified with single and multiple values.

samcli/local/docker/lambda_container.py
sdubov: Sure. Fixed doc here and in `DebugContext` to specify that debug port is a collection of ports to be exposed.

samcli/commands/local/cli_common/options.py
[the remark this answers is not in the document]
jfuss: Perfect. Thanks for digging into that.

samcli/commands/local/lib/debug_context.py
```diff
@@ -5,6 +5,13 @@
 
 class DebugContext(object):
     def __init__(self, debug_port=None, debugger_path=None, debug_args=None):
+        """
+        Initialize the Debug Context with Lambda debugger options
+
+        :param tuple(int) debug_port: Collection of debugger ports to be exposed from a docker container
```
jfuss: Nit: Should this be `debug_ports` instead of `debug_port`

jfuss:
@sdubov I had one last comment. It's purely for readability. Everything else looks good to me.

samcli/commands/local/lib/debug_context.py
sdubov: Agreed. I thought about the same naming change. My original thought here was to persist the original CLI option `--debug-port` down to usages. Updated parameter names from `InvokeContext` down to `DebugContext`.

jfuss: Ya we can keep the option the same (`--debug-port`). This is just an internal object, so no concerns on the updates. 

Will do a re-review here shortly.

sdubov:
@jfuss, there is a merge conflict in debug_context.py file. Would you like me to squash my commits into one and rebase? or merge `develop` into my branch is ok?

sdubov:
Not sure sure `make-pr` configuration is failed. Checked agent details and did not find any issues from logs.

jfuss:
@sdubov its failing due to a format check. We run a tool called black to make sure everything is formatted the same way throughout the codebase and avoids us having to do things manually. I can pull a update that part tomorrow if you don’t want to set that up.

sdubov:
@jfuss, Thank you for clarifying. No worries. Done with code formatting.

samcli/local/docker/lambda_container.py
```diff
@@ -169,17 +174,20 @@ def _get_entry_point(runtime, debug_options=None):  # pylint: disable=too-many-b
...
         :return list: List containing the new entry points. Each element in the list is one portion of the command.
             ie. if command is ``node index.js arg1 arg2``, then this list will be ["node", "index.js", "arg1", "arg2"]
         """
 
         if not debug_options:
             return None
 
-        debug_port = debug_options.debug_port
+        debug_ports = debug_options.debug_ports
+        if not debug_ports:
```
jfuss: This check is fine but I think it is not needed. `__bool__` on DebugContext is `bool(self.debug_ports)`. This should never execute given the first if statement (`if not debug_options`).

jfuss:
@sdubov Thanks for contributing the feature and going through and addressing all the feedback!

sdubov:
@jfuss, no problem :) thank you for your feedback and a quick review 👍

PR #1416, opened 2019-09-17
Deprecate: Remove Python2.7 installation support

jfuss:
*Issue #, if available:*
#1273

*Description of changes:*
Removes the ability to install SAM CLI into a Py2 environment.

I want to take this time to also upgrade some of our dev dependencies that were stuck in the stone age and get rid of six. We will need to do similar things in `aws_lambda_builders` but that will be completely separate than this. If some of the dev dependency upgrades become to much, I will move that to a different PR but for now keeping this as a Draft PR.

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1437, opened 2019-09-30
feat: sam package without awscli pre-installed

TheSriram:
Will be making additional changes and adding tests.

- rely on boto3 instead
- port over code from aws cli

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1482, opened 2019-10-25
Revert "Revert "fix: Respect zipped symlinks (#1140)""

jfuss:
This reverts commit 360d90aa560e0b2bb926ad037729ae160ebf17f6.

This commit originally reverted #1140 due to os.symlink not being
available on windows in Python2.7 stdlib, details here: https://github.com/awslabs/aws-sam-cli/pull/1315#issuecomment-518939441. We recently removed Python2.7 support in #1416, so this commit
is a revert of the revert, with some additional black formating to make `make pr`
pass.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/lambdafn/zip.py
+++ b/samcli/local/lambdafn/zip.py
@@ -101,11 +101,15 @@ def unzip(zip_file_path, output_dir, permission=None):
         # For each item in the zip file, extract the file and set permissions if available
         for file_info in zip_ref.infolist():
             extracted_path = _extract(file_info, output_dir, zip_ref)
-            _set_permissions(file_info, extracted_path)
 
-            _override_permissions(extracted_path, permission)
+            # If the extracted_path is a symlink, do not set the permissions. If the target of the symlink does not
+            # exist, then os.chmod will fail with FileNotFoundError
+            if not os.path.islink(extracted_path):
+                _set_permissions(file_info, extracted_path)
+                _override_permissions(extracted_path, permission)
 
-    _override_permissions(output_dir, permission)
+    if not os.path.islink(extracted_path):
+        _override_permissions(output_dir, permission)
 
 
 def _override_permissions(path, permission):
@@ -138,7 +142,7 @@ def _set_permissions(zip_file_info, extracted_path):
     """
 
     # Permission information is stored in first two bytes.
-    permission = (zip_file_info.external_attr >> 16) & 511
+    permission = zip_file_info.external_attr >> 16
     if not permission:
         # Zips created on certain Windows machines, however, might not have any permission information on them.
         # Skip setting a permission on these files.
--- a/samcli/local/lambdafn/zip.py
+++ b/samcli/local/lambdafn/zip.py
@@ -15,6 +15,70 @@ from samcli.lib.utils.progressbar import progressbar
 
 LOG = logging.getLogger(__name__)
 
+S_IFLNK = 0xA
+
+
+def _is_symlink(file_info):
+    """
+    Check the upper 4 bits of the external attribute for a symlink.
+    See: https://unix.stackexchange.com/questions/14705/the-zip-formats-external-file-attribute
+
+    Parameters
+    ----------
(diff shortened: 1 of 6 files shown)


mhart:
Will this also fix https://github.com/awslabs/aws-sam-cli/issues/477 ?

jfuss:
@mhart 
> Will this also fix #477 ?

No. This is only for unzipping and we are still relying on `aws cloudformation package` until #1437 is done. Then we will have full control over fixing issues like #477

samcli/local/lambdafn/zip.py
```diff
@@ -77,7 +138,7 @@ def _set_permissions(zip_file_info, extracted_path):
     """
 
     # Permission information is stored in first two bytes.
-    permission = zip_file_info.external_attr >> 16
+    permission = (zip_file_info.external_attr >> 16) & 511
```
awood45: Can we consider adding a comment explaining the `& 511` part especially? May be useful to note for posterity/later development.

awood45:
Current plan is to add integ tests before merging.

samcli/local/lambdafn/zip.py
mnapoli: Since @bubba-h57 wrote the initial pull request he might be able to give some background.

bubba-h57:
`zip_file_info.external_attr` contains the files "attributes" (see [stat.h](https://minnie.tuhs.org/cgi-bin/utree.pl?file=4.4BSD/usr/include/sys/stat.h))

```
#define S_IFIFO  0010000  /* named pipe (fifo) */
#define S_IFCHR  0020000  /* character special */
#define S_IFDIR  0040000  /* directory */
#define S_IFBLK  0060000  /* block special */
#define S_IFREG  0100000  /* regular */
#define S_IFLNK  0120000  /* symbolic link */
#define S_IFSOCK 0140000  /* socket */
#define S_ISUID 0004000 /* set user id on execution */
#define S_ISGID 0002000 /* set group id on execution */
#define S_ISTXT 0001000 /* sticky bit */
#define S_IRWXU 0000700 /* RWX mask for owner */
#define S_IRUSR 0000400 /* R for owner */
#define S_IWUSR 0000200 /* W for owner */
#define S_IXUSR 0000100 /* X for owner */
#define S_IRWXG 0000070 /* RWX mask for group */
#define S_IRGRP 0000040 /* R for group */
#define S_IWGRP 0000020 /* W for group */
#define S_IXGRP 0000010 /* X for group */
#define S_IRWXO 0000007 /* RWX mask for other */
#define S_IROTH 0000004 /* R for other */
#define S_IWOTH 0000002 /* W for other */
#define S_IXOTH 0000001 /* X for other */
#define S_ISVTX 0001000 /* save swapped text even after use */
```

We shift it right to drop off the unnecessary bits _(pun included for no charge)_ leaving naught but the permissions we are interested in setting and then we then do a bitwise AND operation with `511`.

Note that `chmod 511 *` would set the file permissions to *-r-x--x--x*? The bitwise AND guarantees that the resulting file permissions will always, at a minimum, be *-r-x--x--x*. 

This is necessary because we really have no idea what sort of `umask` the developer is using, nor even what operating system _(Windows? OSX? *NIX)_ so any file permission schema may have been used.

However, when run on AWS Lambda, the `owner` *must have* read access to all files.

We also don't know which files *need* to be executable, so we simply assume all files need to be executable.

As for testing, the various test environments can have some strange accounts/permissions setup. So we go ahead and ensure anyon can execute any file. Why? Because in the testing environments, it isn't always clear who the owner/executor is. Group/Other will only have read/write access if that is what the development version of the files has.

So, that's what is going on. Take the permission on the development file, overlay that with `511` permissions, and we guarantee that the file will be properly accessed and work in either testing or production environments.

mhart:
Any news here? It seems that layers with symlinks still don't work with sam cli?

PR #1486, opened 2019-10-28
feat: add additional volumes when in debug mode

### LINKED ISSUE #1485: Allow to mount additional volumes when invoking lambda in debugger mode
We use a custom debugger when running .net lambda under debug mode. This debugger process produce some logs. I would like to have an option to get those logs for investigation purposes, e.g. investigate debugger process failures during lambda run. It would be great to have an option to pass an additional option through SAM CLI to mount additional volumes in Docker.

### Proposal
Extend SAM CLI to have an additional option `--additional-volume` with `multiple=True` flag that mount directories in a Docker container that run lambda.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)
No changes needed.

### Additional Details
The mapping should use the following rules:
- All specified volumes are mounted inside `/tmp/lambci_volumes` folder
- Each volume is mounted to a remote volume directory + host directory base name (e.g. `/Users/<user>/mount_this` should be mounted to `/tmp/lambci_volumes/mount_this` directory)
- Remote directory should have write access to be able to write logs into a folder.

sdubov:
*Issue #, if available:*
- #1485

*Description of changes:*
- Add --additional-volume option to SAM CLI invoke call to mount a directory inside a container
- Update existing tests
- Add new tests to verify mapping between remote and host directories logic

*Checklist:*
- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


tests/unit/commands/local/invoke/test_cli.py
```diff
@@ -28,6 +28,7 @@ def setUp(self):
         self.debug_ports = [123]
         self.debug_args = "args"
         self.debugger_path = "/test/path"
+        self.additional_volumes = [Path.home()]
```
awood45: If this feature supports multiple additional volumes, can we explicitly test that?

samcli/local/docker/lambda_container.py
```diff
@@ -201,3 +217,29 @@ def _get_entry_point(runtime, debug_options=None):  # pylint: disable=too-many-b
...
+            return None
+
+        volume_map = {}
+        for host_volume in host_volumes:
+            host_base_names = os.path.basename(host_volume)
+
+            if not host_base_names:
+                continue
+
+            remote_volume_path = os.path.join(LambdaContainer._VOLUME_MOUNT_PATH, host_base_names)
```
sanathkr: Would this work on Windows?

samcli/local/docker/lambda_container.py
```diff
@@ -201,3 +217,29 @@ def _get_entry_point(runtime, debug_options=None):  # pylint: disable=too-many-b
...
+
+            volume_map[host_volume] = {"bind": remote_volume_path, "mode": ""}
```
sanathkr: This means, users can't mount two different host folders that have the name name. Example: if you mount host paths `/foo/bar` and `/tmp/bar`, only one will get mounted. It's also not clear which one. I think this will be a security risk.

Can you instead hard fail in this situation?

sanathkr: We have historically taken the stance that we won't do "write" mounts. Primarily for security, but it also helps performance.

What is your use case for writable mounts?

samcli/local/docker/lambda_container.py
```diff
@@ -31,7 +32,10 @@ class LambdaContainer(Container):
     # This is the dictionary that represents where the debugger_path arg is mounted in docker to as readonly.
     _DEBUGGER_VOLUME_MOUNT = {"bind": _DEBUGGER_VOLUME_MOUNT_PATH, "mode": "ro"}
 
-    def __init__(
+    # Additional volumes Mount path for files in docker
+    _VOLUME_MOUNT_PATH = "/tmp/lambci_volumes"
```
sanathkr: Why do you need the additional directory? What do you think about mounting directly under /tmp instead?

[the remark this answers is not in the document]
sdubov: Sure. Fixed folder name.

tests/unit/commands/local/invoke/test_cli.py
sdubov: @awood45, updated `test_cli.py` with multiple volumes by default. I've added some tests to verify cases when there are multiple volumes in `tests/unit/local/docker/test_lambda_container.py`.

samcli/local/docker/lambda_container.py
sdubov: Good catch. Thank you. Updated to use PosixPath here.

samcli/local/docker/lambda_container.py
sdubov: Fixed. Fail in case a basename already added.

sdubov: We use a custom debugger application in Rider to debug Lambda in container. This debugger app file some log files that I would like to collect to investigate in cases of any issues. We configure our debugger with a custom log folder that I would like to map inside a container. We need write access to be able to write logs into a folder.

samcli/local/docker/lambda_container.py
sdubov: Additional directory is just for better folders structure. `/tmp/` I found might be pretty common place for this purpose. Similar to `/tmp/lambci_debug_files` that is mounted for all debugger files. Please let me know if you would prefer to have these files inside `/tmp/`.

samcli/local/docker/lambda_container.py
sanathkr: Understood. Writable folders can potentially cause security attacks. We have to do a threat model and security posture assessment before we can enable this. If this is exclusively for debugging during development, I can help you find a workaround to `docker cp` logs.

samcli/local/docker/lambda_container.py
sanathkr: Makes sense. To Alex's comments, can you make this folder more descriptive like `/tmp/additional_volumes`?

samcli/local/docker/lambda_container.py
sanathkr: Would this work for youi?

sdubov:
Windows instances are failed with:
```
File "c:\projects\aws-sam-cli\venv\lib\site-packages\win32\lib\pywin32_bootstrap.py", line 14, in <module>
265        for maybe in site.getsitepackages():
266    AttributeError: module 'site' has no attribute 'getsitepackages'
```
that looks like an internal environment error.
@awood45 @sanathkr, do you know if there is anything I can do from my side about this error?

jfuss:
@sdubov We have fixed that issue with windows and pywin32. If you pull from develop, that should solve the issue.

JoeDupuis:
I am trying to run the dev database for our lambdas through sockets only to prevent port clashing. I thought of mounting the socket through docker volumes and found this PR. This is pretty old so I guess support for it is dead, but anyway I could do something similar or do I really need to use ports?

It's for our development environments only.

sdubov:
Sorry for such a long response in this PR. We have found a workaround as per @sanathkr suggested with `docker cp` before we kill a debugger session and stop a Docker container. I understand the reasons from the security standpoint to prevent additional volumes mount. From my side, please feel free to close the PR if it is still the issue with this change or please let me know if you would like to proceed.

samcli/local/docker/lambda_container.py
sdubov: @sanathkr, sorry for such a long response. The main issue with `docker cp` approach is that we do not have a handler that allows us to copy logs before container is deleted. SAM CLI runs with `--rm` flag and we have no guarantee that container is alive at the moment we do `docker cp`. As a workaround, we can make this API on Rider side to make a sync call before we kill the process we want our logs from. I just would like to understand if this is the only option here. I totally understand the security issue with suggested approach. Please let me know if you think it is the stopper for this PR. If yes, please feel free to close and we will do a workaround on our side.

jfuss:
This PR looks like it has stalled and has many conflicts. The referenced issue has been closed and linked to #2589. If this is still something we should address, please 👍 #2589.

Closing

ffMathy:
Wow, that's very sad 😢

Issue #1375, opened 2019-08-27
SAM 0.17+ prompts for MFA credentials on every request when running API locally


johnc44:
### Description

We use Multi-Factor Authentication to access our AWS accounts.

In previous versions (0.19.0), SAM CLI would prompt for my MFA token when I first invoked the lambda, and would then retain those credentials for future requests.  Eventually they would expire and I'd have to quit/restart to get it to work again, but it was quite workable.

In 0.20.1, however, I am asked for my MFA credentials on literally every request.  This makes it completely unusable for me, so I've had to downgrade again.

### Steps to reproduce

Using Ubuntu WSL:

export AWS_PROFILE=<my-mfa-profile>
aws local start-api -n environment.json -p 3100

Then hit the API.  It sits there twiddling its thumbs.  Switch back to Ubuntu and it's asking for the MFA code.  Enter it in, everything works.

Hit the API a second time.  Again, it sits there thinking about it.  Switch back to Ubuntu and it's asking for the MFA again.  This is not the case in 0.19.0 which would remember my token.

### Observed result

Asked for MFA code on every single request.

### Expected result

MFA token should be cached between requests.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10, patched up to date.  Using Ubuntu WSL.
2. `sam --version`: 0.20.1

johnc44:
Hmm.. OK on closer inspection, dropping down to 0.19.0 did not fix the issue.  Neither has going down to 0.18.0.

It used to work as expected.  Last week I upgraded to sam cli 0.20.1 and presumed it was the culprit.  Maybe it's something else?  (I can't guarantee what version I was on before, I think it was 0.18.0)

Is there anything I can do to work around this?  Having to type the MFA code in every time is driving me crazy.

johnc44:
OK, I went back to sam 0.16 and the problem has now cleared up.  So either in 0.17 or 0.18 it seems to have changed behaviour.

As a side note - I've also noticed that if I Ctrl+C to quit the process whilst it's asking me for an MFA then it renders the whole terminal unusable.  (Well, by that I mean I can't see what I'm typing - but I can actually type blindly in and it works).

But, anyway... any idea what's changed in SAM that would affect this behaviour?  Any workaround or do I just need to stick to version 0.16?

TheSriram:
Interesting is this just a windows 10 issue, do you see it on any other platform?

johnc44:
[...]

PR #1494, opened 2019-10-31
fix: Cache AWS Creds instead of getting a Session on every invoke

jfuss:
Addresses Issue #1375.

Due to Boto3.session.Session objects not being thread safe, we
create a Session object on every invoke to get credentials to
mount into the container. This works well except for customers
using MFA. When using MFA, this dehavior requires customers to
provide MFA code/pin on every invoke. This creates much longer
debug sessions for customers and pauses invokes until the MFA
code/pin is provided. This PR changes that behavior to one we
had between version v0.8.0 and v0.16.0, which is caching the
creds/session to be reused on other invokes until the credentials
expire. Once the credentials expire, customers will need to restart
the command (which was the same behavior on version 0.8.0 to 0.16.0.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/local_lambda.py
```diff
@@ -205,6 +207,22 @@ def _make_env_vars(self, function):
             aws_creds=aws_creds,
         )
 
+    def _get_session_creds(self):
```
TheSriram: we could make this a lazy property in the future.

jfuss: Good idea. Going to leave it as is for now but something we should consider for the future when dealing with things like this.

PR #1495, opened 2019-10-31
Match local APIG API slashes behavior to AWS APIG

deltalus-aws:
*Issue #, if available:*

*Description of changes:*
When defining an API, if the APIG template defines a path at `/foo`, local APIG will return an error that the URL is not found when sending a request formatted as `/foo/`. This does not match the behavior of AWS APIG. The CloudWatch logs receive the `/foo/` request but is able to answer suing the method defined for `/foo`.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -110,6 +110,8 @@ def create(self):
             static_folder=self.static_dir,  # Serve static files from this directory
         )
 
+        self._app.url_map.strict_slashes = False
```
sriram-mv: Can we  add a comment on what this means.

jfuss: Sure can do.

PR #1500, opened 2019-11-05
exposed extra_context CLI option to pass cookiecutter parameter

gupta-n:
*Issue #, if available:*
NA 
*Description of changes:*
Exposing extra_context as SAM CLI argument to pass cookiecutter parameter
*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/init/__init__.py
```diff
@@ -126,12 +147,20 @@ def do_cli(
...
             templates = InitTemplates(no_interactive, auto_clone)
             location = templates.location_from_app_template(runtime, dependency_manager, app_template)
             no_input = True
-            extra_context = {"project_name": name, "runtime": runtime}
+            default_context = {"project_name": name, "runtime": runtime}
+            if extra_context is None:
+                extra_context = default_context
+            else:
+                merged_context = default_context.copy()
+                extra_context_dict = json.loads(extra_context)
```
awood45: How are we handling it when users provide invalid JSON?

tests/unit/commands/init/test_cli.py
```diff
@@ -244,9 +249,91 @@ def test_init_cli_generate_project_fails(self, generate_project_patch, sd_mock):
...
+        # THEN we should receive no errors
+        generate_project_patch.assert_called_once_with(
+            ANY,
+            self.runtime,
+            self.dependency_manager,
+            ".",
+            self.name,
+            True,
+            {"project_name": "testing project", "runtime": "python3.6", "schema_name": "events", "schema_type": "aws"},
+        )
```
awood45: Can we add some invalid input tests to confirm proper behavior?

samcli/commands/init/__init__.py
gupta-n: added test.

tests/unit/commands/init/test_cli.py
gupta-n: added test

samcli/commands/init/__init__.py
```diff
@@ -3,6 +3,8 @@
 Init command to scaffold a project app from a template
 """
 import logging
+import json
+from json import JSONDecodeError
```
awood45: I'd recommend not having both a full and a partial import of `json` - you can use `json.JSONDecodeError` below, or import any json methods you use as a part of the `from` statement.

awood45: I'd probably go with the former approach since `JSONDecodeError` is only used once.

samcli/commands/init/__init__.py
```diff
@@ -126,12 +149,25 @@ def do_cli(
...
-            extra_context = {"project_name": name, "runtime": runtime}
+            default_context = {"project_name": name, "runtime": runtime}
+            if extra_context is None:
+                extra_context = default_context
+            else:
+                merged_context = default_context.copy()
+                try:
+                    extra_context_dict = json.loads(extra_context)
+                except JSONDecodeError:
+                    raise UserException(
```
jfuss: Can we make an explicit Exception here? Something like:

```
class InvalidCliParameterValue(UserException):
    pass
```

samcli/commands/init/__init__.py
```diff
@@ -126,12 +149,25 @@ def do_cli(
     # check for required parameters
     if location or (name and runtime and dependency_manager and app_template):
         # need to turn app_template into a location before we generate
-        extra_context = None
         if app_template:
             templates = InitTemplates(no_interactive, auto_clone)
             location = templates.location_from_app_template(runtime, dependency_manager, app_template)
             no_input = True
-            extra_context = {"project_name": name, "runtime": runtime}
+            default_context = {"project_name": name, "runtime": runtime}
```
awood45: Can we actually pull this into another method? The method would handle all of the default context and merging of extra context, and return the new context value to pass to later calls.

samcli/commands/init/__init__.py
```diff
@@ -126,12 +149,25 @@ def do_cli(
...
+            if extra_context is None:
```
jfuss: This is all pretty dense. Why can't we do something like:

```
default_context = {"project_name": name, "runtime": runtime}
extra_context_dict = {}
try:
    extra_context_dict = json.loads(extra_context)
except JSONDecodeError:
   raise

default_context = {**default_context , **extra_context_dict}
```

gupta-n: I was trying to avoid override for default parameter by switching order this does the work.

samcli/commands/init/__init__.py
```diff
@@ -86,16 +89,36 @@
...
+        output_dir,
+        name,
+        app_template,
+        no_input,
+        extra_context,
     )  # pragma: no cover
 
 
-# pylint: disable=too-many-locals
+# pylint: disable=too-many-locals, too-many-statements
```
awood45: Can we remove the too-many-statements disable flag if no longer needed?

PR #1502, opened 2019-11-05
feat: configuration file for sam cli commands

TheSriram:
- add passthroughs to the CLI command line interface via a configuration
  file
- local defaults are set at: `samconfig.toml` at project root.
- This commit contains code copied and modified from https://github.com/phha/click_config_file/blob/master/click_config_file.py under MIT license
- Design doc PR is separate: https://github.com/awslabs/aws-sam-cli/pull/1503

- [ ] Design doc
- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -12,5 +12,5 @@ python-dateutil~=2.6, <2.8.1
 requests==2.22.0
 serverlessrepo==0.1.9
 aws_lambda_builders==0.5.0
-# https://github.com/mhammond/pywin32/issues/1439
 pywin32 < 226; sys_platform == 'win32'
+toml==0.10.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -12,5 +12,6 @@ python-dateutil~=2.6, <2.8.1
 requests==2.22.0
 serverlessrepo==0.1.9
 aws_lambda_builders==0.5.0
+# https://github.com/mhammond/pywin32/issues/1439
 pywin32 < 226; sys_platform == 'win32'
 toml==0.10.0
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -1,0 +1,158 @@
+"""
+CLI configuration decorator to use TOML configuration files for click commands.
+"""
+
+## This section contains code copied and modified from [click_config_file][https://github.com/phha/click_config_file/blob/master/click_config_file.py]
+## SPDX-License-Identifier: MIT
+
+import functools
+import os
+import logging
+
+import click
+import toml
+
+__all__ = ("TomlProvider", "configuration_option")
+
+LOG = logging.getLogger(__name__)
+DEFAULT_CONFIG_FILE_NAME = "sam-app-config"
+
+
+class TomlProvider:
+    """
+    A parser for toml configuration files
+
+    Parameters
+    ----------
+    section : str
+        If this is set to something other than the default of `None`, the
+        provider will look for a corresponding section inside the
+        configuration file and return only the values from that section.
+    """
+
+    def __init__(self, section=None):
+        self.section = section
+
+    def __call__(self, file_path, cmd_name):
+        """
+        Parse and return the configuration parameters.
+
+        Parameters
+        ----------
+        file_path : str
+            The path to the configuration file
+        cmd_name : str
+            The name of the click command
+
+        Returns
+        -------
+        dict
+            A dictionary containing the configuration parameters.
+        """
+        config = toml.load(file_path)
+        if self.section:
+            try:
+                config = config[self.section]
+            except KeyError:
+                LOG.warning("Specified section: %s does not exist in the config file", self.section)
+        return config
+
+
+def configuration_callback(cmd_name, option_name, config_file_name, saved_callback, provider, ctx, param, value):
+    """
+    Callback for reading the config file.
+
+    Also takes care of calling user specified custom callback afterwards.
+
+    cmd_name : str
+        The command name. This is used to determine the configuration directory.
+    option_name : str
+        The name of the option. This is used for error messages.
+    config_file_name : str
+        The name of the configuration file.
+    saved_callback: callable
+        User-specified callback to be called later.
+    provider : callable
+        A callable that parses the configuration file and returns a dictionary
+        of the configuration parameters. Will be called as
+        `provider(file_path, cmd_name)`. Default: `toml_provider()`
+    ctx : object
+        Click context.
+    """
+    ctx.default_map = ctx.default_map or {}
+    cmd_name = cmd_name or ctx.info_name
+
+    default_value = os.path.join(".aws-sam", config_file_name)
+
+    param.default = default_value
+    value = value or default_value
+
+    if os.path.isfile(value):
+        try:
+            LOG.info("Config file location: %s", os.path.abspath(value))
+            config = provider(value, cmd_name)
+
+        except Exception as e:
+
+            raise click.BadOptionUsage(option_name, "Error reading configuration file: {}".format(e), ctx)
+        ctx.default_map.update(config)
+
+    return saved_callback(ctx, param, value) if saved_callback else value
+
+
+def configuration_option(*param_decls, **attrs):
+    """
+    Adds configuration file support to a click application.
+
+    This will create an option of type `click.File` expecting the path to a
+    configuration file. When specified, it overwrites the default values for
+    all other click arguments or options with the corresponding value from the
+    configuration file.
+
+    The default name of the option is `--config`.
+
+    This decorator accepts the same arguments as `click.option` and
+    `click.Path`. In addition, the following keyword arguments are available:
+
+    cmd_name : str
+        The command name. This is used to determine the configuration
+        directory. Default: `ctx.info_name`
+    config_file_name : str
+        The name of the configuration file. Default: `'sam-app-config'``
+    provider : callable
+        A callable that parses the configuration file and returns a dictionary
+        of the configuration parameters. Will be called as
+        `provider(file_path, cmd_name)`. Default: `toml_provider()`
+        """
+    param_decls = param_decls or ("--config",)
+    option_name = param_decls[0]
+
+    def decorator(f):
+
+        attrs.setdefault("is_eager", True)
+        attrs.setdefault("help", "Read configuration from FILE.")
+        attrs.setdefault("expose_value", False)
+        cmd_name = attrs.pop("cmd_name", None)
+        config_file_name = attrs.pop("config_file_name", DEFAULT_CONFIG_FILE_NAME)
+        provider = attrs.pop("provider", TomlProvider())
+        path_default_params = {
+            "exists": False,
+            "file_okay": True,
+            "dir_okay": False,
+            "writable": False,
+            "readable": True,
+            "resolve_path": False,
+        }
+        path_params = {k: attrs.pop(k, v) for k, v in path_default_params.items()}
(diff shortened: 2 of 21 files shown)


TheSriram:
Integration tests are working locally, need to investigate appveyor gates. Unrelated PR: https://github.com/awslabs/aws-sam-cli/pull/1501 is experiencing similar issues.

TheSriram:
open to modification of `--identifier`. it could be environment, mode, config.

TheSriram:
@awood45 thanks for the review! yes those tests will be added in next series of commits.

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+import os
+import logging
+
+import click
+import toml
+
+__all__ = ("TomlProvider", "configuration_option")
+
+LOG = logging.getLogger(__name__)
+DEFAULT_CONFIG_FILE_NAME = "sam-config"
```
sanathkr: i guess this will be changed to `samconfig.toml`

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+        self.section = section
+
+    def __call__(self, file_path, identifier, cmd_name):
+        """
+        :param file_path: The path to the configuration file
+        :param identifier: The name of the sectional identifier within configuration file.
+        :param cmd_name: sam command name as defined by click
+        :returns dictionary containing the configuration parameters under specified identifier
+        """
+        config = toml.load(file_path)
```
sanathkr: - What happens if the file does not exist?
- What happens if the TOML file cannot be parsed?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+        resolved_config = {}
+        if self.section:
+            try:
+                resolved_config = self._get_identifier(config, identifier)[self.cmd][self.section]
+            except KeyError:
+                LOG.debug(
```
sanathkr: Should we print a error message to customers?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    :param cmd: sam command name as defined by click
+    :param section: section defined in the configuration file nested within `cmd`
+    """
+
+    def __init__(self, cmd=None, section=None):
+        self.cmd = cmd
+        self.section = section
+
+    def __call__(self, file_path, identifier, cmd_name):
+        """
```
sanathkr: Quick line explaining what the method does?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+        :param file_path: The path to the configuration file
```
sanathkr: Is this param a string or pathlib?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    :param saved_callback: User-specified callback to be called later.
+    :param provider:  A callable that parses the configuration file and returns a dictionary
+        of the configuration parameters. Will be called as
+        `provider(file_path, identifier, cmd_name)`.
+    :param ctx: Click context
+    :param param: Click parameter
+    :param value: Specified value for identifier
+    :returns specified callback or the specified value for identifier.
+    """
+    ctx.default_map = ctx.default_map or {}
```
sanathkr: In which scenario, can `ctx` be None? How do handle that?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    cmd_name = cmd_name or ctx.info_name
+
+    config_file = os.getenv("SAM_CONFIG", os.path.abspath(os.path.join(".", DEFAULT_CONFIG_FILE_NAME)))
```
sanathkr: change envvar to `SAM_CLI_CONFIG`. 

- Can you write make this a class/file level constant?
- Also, why the `.` in os.path.join?
- Is `os.cwd()` the correct working directory to resolve the file relative to? According to the design doc, this was relative to the `template.yaml` file. Am I missing something?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+
+    param.default = DEFAULT_IDENTIFER
```
sanathkr: How is `default` added to param?

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    :param cmd_name: `sam` command name derived from click.
+    :param option_name: The name of the option. This is used for error messages.
+    :param identifier_name: `top` level section within configuration file
+    :param saved_callback: User-specified callback to be called later.
+    :param provider:  A callable that parses the configuration file and returns a dictionary
+        of the configuration parameters. Will be called as
+        `provider(file_path, identifier, cmd_name)`.
+    :param ctx: Click context
+    :param param: Click parameter
+    :param value: Specified value for identifier
```
sanathkr: can you provide a better variable name than `value`? It is too vague.

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    cmd_name = cmd_name or ctx.info_name
+
+    config_file = os.getenv("SAM_CONFIG", os.path.abspath(os.path.join(".", DEFAULT_CONFIG_FILE_NAME)))
+
+    param.default = DEFAULT_IDENTIFER
+    identifier_name = value or identifier_name
+
+    if os.path.isfile(config_file):
+        try:
+            LOG.debug("Config file location: %s", os.path.abspath(config_file))
```
sanathkr: so this `os.path.abspath` again here is dangerous. It is a good practice to do relative->absolute path conversion once because it's tricky getting the right cwd

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,152 @@
...
+    :param identifier_name: The identifier name. This is used to determine which part of the configuration
+        needs to be read.
+    :param provider:         A callable that parses the configuration file and returns a dictionary
+        of the configuration parameters. Will be called as
+        `provider(file_path, identifier, cmd_name)
+    """
+    param_decls = param_decls or ("--identifier",)
+    option_name = param_decls[0]
+
+    def decorator(f):
```
sanathkr: I don't understand the Click API a whole lot. So if you want me to dig into those and review this method in detail, let me know.

samcli/commands/local/start_lambda/cli.py
```diff
@@ -49,6 +50,7 @@
     help=HELP_TEXT,
     short_help="Starts a local endpoint you can use to invoke your local Lambda functions.",
 )
+@configuration_option(provider=TomlProvider(cmd="local_start_lambda", section="parameters"))
```
sanathkr: Document the "algorithm" to create the command name here.

samcli/cli/cli_config_file.py
TheSriram: changed.

samcli/cli/cli_config_file.py
TheSriram: there are checks for that.

samcli/cli/cli_config_file.py
TheSriram: changed to error, although might be polluting if they choose not to use a config file.

samcli/cli/cli_config_file.py
TheSriram: its a string. understandable by os.path, can change to pathlib

samcli/cli/cli_config_file.py
TheSriram: ctx should not be none, the decorators need to be explicitly added to a click command.

samcli/cli/cli_config_file.py
TheSriram: addressed.

samcli/cli/cli_config_file.py
TheSriram: ctx, param, value are used in click callbacks, so re-using them.

samcli/commands/local/start_lambda/cli.py
TheSriram: done.

samcli/cli/cli_config_file.py
TheSriram: We are essentially creating a hidden option, whose callback can help populate defaults for that command.

requirements/base.txt
```diff
@@ -12,5 +12,5 @@ python-dateutil~=2.6, <2.8.1
 requests==2.22.0
 serverlessrepo==0.1.9
 aws_lambda_builders==0.5.0
-# https://github.com/mhammond/pywin32/issues/1439
-pywin32 < 226; sys_platform == 'win32'
\ No newline at end of file
+pywin32 < 226; sys_platform == 'win32'
```
sanathkr: Can you add the GH Issue link back? It will useful to track when the issue is fixed

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    param_decls = param_decls or ("--config-env",)
+    option_name = param_decls[0]
+
+    def decorator(f):
+
+        attrs.setdefault("is_eager", True)
+        attrs.setdefault("help", "Read config-env from Configuration File.")
+        attrs.setdefault("expose_value", False)
+        # --config-env is hidden and can potentially be opened up in the future.
```
sanathkr: can we make sure we don't respect user specififc value?

requirements/base.txt
TheSriram: yes, lost it in a rebase conflict.

samcli/cli/cli_config_file.py
TheSriram: Done.

samcli/cli/cli_config_file.py
```diff
@@ -0,0 +1,185 @@
...
+    If a given cmd_name is start-api, the parsed name is "local_start_api".
+    provider is called with `config_file`, `config_env_name` and `parsed_cmd_name`.
+
+    :param cmd_name: `sam` command name
+    :param provider: provider to be called for reading configuration file
+    :param ctx: Click context
+    :param config_env_name: config-env within configuration file
+    :return: dictionary of defaults for parameters
+    """
+    config_file = os.path.join(os.getcwd(), DEFAULT_CONFIG_FILE_NAME)
```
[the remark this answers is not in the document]
TheSriram: This makes sense for commands that do take in a template file. how does this work on commands where there are no template files? logs and init dont have such a concept.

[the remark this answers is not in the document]
sanathkr: os.getcwd()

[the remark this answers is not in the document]
TheSriram: addressed.

samcli/commands/_utils/options.py
```diff
@@ -43,8 +44,10 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
             if os.path.exists(option):
                 provided_value = option
                 break
-
     result = os.path.abspath(provided_value)
+
+    if ctx:
+        setattr(ctx, "config_path", os.path.dirname(provided_value))
     LOG.debug("Using SAM Template at %s", result)
```
sanathkr: Clever! But you need to use 'result' and not provided_value because there could be relative paths that's provided

TheSriram: Good catch! done.

PR #1503, opened 2019-11-05
design: samconfig

TheSriram:
- a configuration that provides parameter pass throughs for
  sam cli commands

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/designs/sam-app-config.md
+++ b/designs/sam-app-config.md
@@ -77,6 +77,15 @@ region="us-east-1"
 profile="srirammv"
 ```
 
+The configuration file can then be potentially intialized
+
+* all sam init projects come with a sample .sam-app-config file
+
+Open Questions
+-------------------------------
+
+* Potentially every sam command could have functionality to have a series of command line parameters exported into a configuraion file.
+
 Tenets
 -------------------------------
 
--- a/designs/sam-app-config.md
+++ b/designs/sam-app-config.md
@@ -20,7 +20,7 @@ Have a look at the following series of workflows
 
 * `sam package --s3-bucket ... --template-file ... --output-template-file ... --s3-prefix ... --kms-key-id ...`
 
-* `sam deploy deploy --template-file ... --stack-name ... --capabilities ... --tags ... --parameter-overrides ... --kms-key-id ...`
+* `sam deploy --template-file ... --stack-name ... --capabilities ... --tags ... --parameter-overrides ... --kms-key-id ...`
 
 If this could be condensed into a series of workflows that look like
 
@@ -33,15 +33,26 @@ That would be a huge user experience win.
 What will be changed?
 ---------------------
 
-The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located to the template.yaml by default. 
+The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located at the project root where template.yaml is located by default. 
 
-`.aws-sam/sam-app-config`
+`.sam-app-config`
 
 This configuration would solely be used for specifiying the parameters that each of SAM CLI commands use and would be in TOML format.
 
+Running a SAM CLI command now automatically looks for `.sam-app-config` file and if its finds it goes ahead with parameter passthroughs to the CLI.
+
+```
+sam build
+Default Config file location: .sam-app-config
+..
+..
+..
+```
+
+
 A configuration file can also optionally be specified by each of the commands themselves as well.
 
-`sam build --config .aws-sam/sam-app-dev-config`
+`sam build --config .sam-app-dev-config`
 
 Sample configuration file
 
@@ -66,7 +77,7 @@ region="us-east-1"
 profile="srirammv"
 ```
 
-Success criteria for the change
+Tenets
 -------------------------------
 
 * Resolution of command line parameters should always favor explicit versus implicit. A native command line parameter specified directly on the command line should override a parameter specified in the configuration file.
@@ -92,11 +103,8 @@ Implementation
 CLI Changes
 -----------
 
-New command line argument is added per command called `--config` to be able to specify non default locations of config files
+New command line argument is added per command called `--config` to be able to specify non default locations of config files.
 
-For example:
-
-`sam build --config /tmp/sam-api-app-config`
 
 ### Breaking Change
 
--- a/designs/sam-app-config.md
+++ b/designs/sam-app-config.md
@@ -1,0 +1,201 @@
+SAM CLI App Level Config
+====================================
+
+
+What is the problem?
+--------------------
+
+Today users of SAM CLI need to invoke the CLI directly with all parameters supplied to its commands.
+
+for e.g: `sam build --use-container --debug`
+
+But often, during the lifecycle of building and deploying a serverless application. the same commands get run repeatedly to build, package and deploy, before solidifying into the final application. 
+
+These CLI commands are often long and have many changing parts.
+
+Have a look at the following series of workflows
+
+
+* `sam build --use-container --template ... --parameter-overrides=... --skip-pull-image --manifest ...`
+
+* `sam package --s3-bucket ... --template-file ... --output-template-file ... --s3-prefix ... --kms-key-id ...`
+
+* `sam deploy deploy --template-file ... --stack-name ... --capabilities ... --tags ... --parameter-overrides ... --kms-key-id ...`
+
+If this could be condensed into a series of workflows that look like
+
+* `sam build`
+* `sam package`
+* `sam deploy`
+
+That would be a huge user experience win.
+
+What will be changed?
+---------------------
+
+The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located to the template.yaml by default. 
+
+`.aws-sam/sam-app-config`
+
+This configuration would solely be used for specifiying the parameters that each of SAM CLI commands use and would be in TOML format.
+
+A configuration file can also optionally be specified by each of the commands themselves as well.
+
+`sam build --config .aws-sam/sam-app-dev-config`
+
+Sample configuration file
+
+```
+[build]
+profile="srirammv"
+debug=true
+skip_pull_image=true
+use_container=true
+
+[package]
+profile="srirammv"
+region="us-east-1"
+s3_bucket="sam-bucket"
+output_template_file="packaged.yaml"
+
+[deploy]
+template_file="packaged.yaml"
+stack_name="using_config_file"
+capabilities="CAPABILITY_IAM"
+region="us-east-1"
+profile="srirammv"
+```
+
+Success criteria for the change
+-------------------------------
+
+* Resolution of command line parameters should always favor explicit versus implicit. A native command line parameter specified directly on the command line should override a parameter specified in the configuration file.
+
+
+Out-of-Scope
+------------
+
+* Not focusing on a global configuration. SAM CLI already has a notion of a global config at `~/.aws-sam/metadata.json`
+
+User Experience Walkthrough
+---------------------------
+
+Once a configuration file is appropriately populated, day to day workflows per application developed with SAM CLI become much simpler.
+
+* `sam build` -> `sam package` -> `sam deploy`
+* `sam build` -> `sam local invoke`
+* `sam build` -> `sam package` -> `sam publish`
+
+Implementation
+==============
+
+CLI Changes
+-----------
+
+New command line argument is added per command called `--config` to be able to specify non default locations of config files
+
+For example:
+
+`sam build --config /tmp/sam-api-app-config`
+
+### Breaking Change
+
+* No breaking changes to the CLI, in absence of the configuration file, CLI continues to work as normal.
+
+Design
+------
+
+*Explain how this feature will be implemented. Highlight the components
+of your implementation, relationships* *between components, constraints,
+etc.*
+
+A custom decorator to `click.option` is developed which reads from a configuration file the sections that are pertinent to that particular command and populates the click's `map` context.
+
+The configuration file parser is a custom provider that can be made to understand any configuration file format in a pluggable manner.
+
+This decorator benefits from the same type checking that some SAM CLI parameters already use.
+
+`.samrc` Changes
+----------------
+
+This design emphasizes parameter pass throughs with a configuration file and does not change the core working of the SAM CLI itself. The SAM CLI continues to be working just as it was with efficiency gains in usability.
+
+Security
+--------
+
+*Tip: How does this change impact security? Answer the following
+questions to help answer this question better:*
+
+**What new dependencies (libraries/cli) does this change require?**
+
+toml
+
+**What other Docker container images are you using?**
+
+N/A
+
+**Are you creating a new HTTP endpoint? If so explain how it will be
+created & used**
+
+N/A
+
+**Are you connecting to a remote API? If so explain how is this
+connection secured**
+
+N/A
+
+**Are you reading/writing to a temporary folder? If so, what is this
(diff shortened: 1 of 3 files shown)


designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+These CLI commands are often long and have many changing parts.
+
+Have a look at the following series of workflows
+
+
+* `sam build --use-container --template ... --parameter-overrides=... --skip-pull-image --manifest ...`
+
+* `sam package --s3-bucket ... --template-file ... --output-template-file ... --s3-prefix ... --kms-key-id ...`
+
+* `sam deploy deploy --template-file ... --stack-name ... --capabilities ... --tags ... --parameter-overrides ... --kms-key-id ...`
```
awood45: `sam deploy deploy`?

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+
+If this could be condensed into a series of workflows that look like
+
+* `sam build`
+* `sam package`
+* `sam deploy`
+
+That would be a huge user experience win.
```
awood45: I'd add to the problem some more considerations:

* Dev/Beta/PreProd/Prod workflows may vary and need to be reflected somehow.
* There is a tie-in with `~/.aws/credentials` and `~/.aws/config` to respect, or at least to keep consistent in some way.
* A solution should allow for overriding of any defaults - explicit parameter choices should always be respected.

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+
+What will be changed?
+---------------------
+
+The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located to the template.yaml by default. 
+
+`.aws-sam/sam-app-config`
```
awood45: Is this an `.aws-sam` directory relative to the project? Would we want to conflate this with build directories like that?

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+
+This configuration would solely be used for specifiying the parameters that each of SAM CLI commands use and would be in TOML format.
+
+A configuration file can also optionally be specified by each of the commands themselves as well.
+
+`sam build --config .aws-sam/sam-app-dev-config`
+
+Sample configuration file
+
+```
```
awood45: INI format?

awood45: Also, do we want to consider user profiles in a way that aligns with AWS CLI?

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+Success criteria for the change
+-------------------------------
+
+* Resolution of command line parameters should always favor explicit versus implicit. A native command line parameter specified directly on the command line should override a parameter specified in the configuration file.
+
+
+Out-of-Scope
+------------
+
+* Not focusing on a global configuration. SAM CLI already has a notion of a global config at `~/.aws-sam/metadata.json`
```
awood45: Different in nature. In the future, could always support a global configuration file in `~/.aws-sam/` - it doesn't conflict with `metadata.json` existing. It's an intentional choice because a global config may not make sense for this problem.

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+output_template_file="packaged.yaml"
+
+[deploy]
+template_file="packaged.yaml"
+stack_name="using_config_file"
+capabilities="CAPABILITY_IAM"
+region="us-east-1"
+profile="srirammv"
+```
+
```
awood45: It may be worth mentioning how this config file will be bootstrapped. By hand? By a helper command? Via init? Some combination?

designs/sam-app-config.md
TheSriram: Good catch.

designs/sam-app-config.md
TheSriram: Not necessarily. ~/.aws/credentials and ~/.aws/config are global files. Yes, explicit parameter will always be respected. I can add that under a section called tenets.

designs/sam-app-config.md
TheSriram: yes, its where build artifacts go today.

designs/sam-app-config.md
awood45: Those are global config files, but do we want the same profile identifier to work for both? Why or why not? Let's make a conscious decision.

designs/sam-app-config.md
awood45: I'd want to separate these concepts. I don't think there's any harm to having this expected to exist at project root, or to be otherwise specified. (On that note, we should support an ENV variable for this possibly).

designs/sam-app-config.md
TheSriram: this is toml. looks similair to ini but understands types better. pip today respects `pyproject.toml`. I'd rather have multiple configuration files, similar to how parameter overrides usually denote different environments.

designs/sam-app-config.md
TheSriram: This design does not cover for that scenario, but can account for potential directions forward.

designs/sam-app-config.md
TheSriram: We can always leave the door open for identifiers, but I'd prefer not to start with one. It may conflate expectations with aws profiles.

designs/sam-app-config.md
jfuss: Why `sam-app-config`? Should we simplify this further to samconfig? 

I am confused on the placement within `.aws-sam`. To this point, this directory is generated by the cli directly (to ensure correct permissions, mainly). With this, we are suggestion to customers that they need to now create this directory. This might be ok if the entry point was through a command (sam config generate) but we aren't that far yet. Why not place this within the root of the project?

designs/sam-app-config.md
sanathkr: +1 to start with simple, as long as we keep door open to add profiles in future if needed

designs/sam-app-config.md
jfuss: We should make sure toml supports what we want. I am ok with it but I prefer us to not support multiple configuration files. The upkeep on that might be more than we originally think.

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+* `sam build` -> `sam local invoke`
+* `sam build` -> `sam package` -> `sam publish`
+
+Implementation
+==============
+
+CLI Changes
+-----------
+
+New command line argument is added per command called `--config` to be able to specify non default locations of config files
```
jfuss: Is there a reason we wouldn't want a command to not have the `--config` option?

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+* `sam build`
+* `sam package`
+* `sam deploy`
+
+That would be a huge user experience win.
+
+What will be changed?
+---------------------
+
+The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located to the template.yaml by default.
```
jfuss: How does this work with build then? We have (at least try to) trained customers to operate on their source. In cases of build, we generate a new template for them (that they shouldn't care about). This means in some cases, we would pick the config in the source but others (that have run build before) we will be looking at the built template. Lets be strong on where this should live, otherwise this can be a point of confusion for customers.

designs/sam-app-config.md
```diff
@@ -0,0 +1,201 @@
...
+Expected Results
+----------------
+* Works on all platforms
+* Resolution of parameters follows.
+ * CLI parameters -> Config file parameters
+
+Documentation Changes
+=====================
+
+* Addition of a new `--config` file parameter per command
```
jfuss: What about docs page about the config, what it does, how/why you should use it?

designs/sam-app-config.md
TheSriram: We can move it up to the project root, to me this configuration was very specific to the app the user was working on at that point. Also might prevent accidental checking it in to source control, since we ignore the .aws-sam directory.

designs/sam-app-config.md
TheSriram: generate-event consists of tons of subcommands, that may be a place where config might not be used as much. But we can add it across all commands.

designs/sam-app-config.md
TheSriram: `sam-app-config` because it doesn't change the inherent behavior of SAM CLI, its a pass through of parameters.

designs/sam-app-config.md
TheSriram: Moved to project level with an option to set the location of the configuration file with a environment variable.

TheSriram:
open to modification of `--identifier`. it could be environment, mode, config.

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+```
+sam build
+Default Config file location: .sam-config
+..
+..
+..
+```
+
+
+The default location of a .sam-config can be replaced by overriding an environment variable called `SAM_CONFIG`
```
sanathkr: Is this consistent with other env vars names? Should we make this SAM_CLI_CONFIG?

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+* `sam deploy`
+
+That would be a huge user experience win.
+
+What will be changed?
+---------------------
+
+The suite of commands supported by SAM CLI would be aided by looking for a configuration file thats locally located at the project root where template.yaml is located by default. 
+
+`.sam-app-config`
```
sanathkr: change?

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
+SAM CLI App Level Config
+====================================
+
+
+What is the problem?
```
sanathkr: Provide a bigger scope

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
+SAM CLI App Level Config
```
sanathkr: Can we call this "SAM Config" in general then?

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+`
+export SAM_CONFIG=~/Users/username/mysamconfig
+`
+
+Users can pass an identifier for the section that will be scanned within the configuration file to pass parameters through.
+
+Be default the `default` section of the configuration is chosen.
+
+```
+[default]
```
sanathkr: awesome!

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+output_template_file="packaged.yaml"
+
+[default.deploy]
+[default.deploy.parameters]
+stack_name="using_config_file"
+capabilities="CAPABILITY_IAM"
+region="us-east-1"
+profile="srirammv"
+
+[dev]
```
sanathkr: This makes much more sense to me. I like this

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+
+The default location of a .sam-config can be replaced by overriding an environment variable called `SAM_CONFIG`
+
+`
+export SAM_CONFIG=~/Users/username/mysamconfig
+`
+
+Users can pass an identifier for the section that will be scanned within the configuration file to pass parameters through.
+
+Be default the `default` section of the configuration is chosen.
```
jfuss: nit: By default

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+[default]
+
+[default.build]
+[default.build.paramaters]
+profile="srirammv"
+debug=true
+skip_pull_image=true
+use_container=true
+
+[default.package]
```
jfuss: Not overly familiar with TOML but is this required? What would we place in this section?

designs/sam-config.md
```diff
@@ -0,0 +1,282 @@
...
+The configuration file can then be potentially intialized
+
+* all sam init projects could come with a sample .sam-config file
+
+Open Questions
+-------------------------------
+
+* Potentially every sam command could have functionality to have a series of command line parameters exported into a configuraion file.
+
+Tenets
```
jfuss: Nit: consider moving this up before "What will be changed?". Might help set some context and be easier to reference and update when we make changes in the future.

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+
+
+Config file in Git Repos
+------------------------
+
+`samconfig.toml` file can be checked into a git repo, so that its ready to use on cloning the repo. if the configuration file does not present all the necesssary parameters, the command fails just as if one had specified the same arguments on the command line directly.
+
+Optionally, if multiple configuration files are checked in. One can change the `SAM_CLI_CONFIG` environment variable to point a different configuration file.
+
+`--env` can also be passed in to deal with custom environments defined in the configuration file.
```
awood45: Could we use a different parameter name than this, something more descriptive/connected to the config file?

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+------
+
+The design can be built in phases.
+
+* No option to specify configuration file or env name
+* Specify configuration file with an environment variable
+* Read `--env` to make sure we can select an appropriate portion in configuration file.
+
+
+`.samrc` Changes
```
sanathkr: This PR is `samrc`. Can you actually rename this part of the design doc template to call it `samconfig` instead?

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+`--env` can also be passed in to deal with custom environments defined in the configuration file.
+
+Error Messages
+---------------
+
+When a custom env is passed in, and such an environment is not found. The error message can highlight all the environments that were found in the given configuration file.
+
+`
+sam build --env devo
+Error: Environment 'devo' was not found in .aws-sam/samconfig.toml , Possible environments are : ['dev', 'prod']
```
sanathkr: Good job on the thinking ahead about error messages.

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+
+Pre-requesites
+--------------
+
+N/A
+
+Test Scenarios/Cases
+--------------------
+
+* Integration tests for every command with identifer based overrides, and command line overrides on existing sam configuration file and custom configuration file through environment variables.
```
sanathkr: s/identifier/env

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+----
+
+Configuration files are tested alongside SAM CLI and are expected to work seamlessly with meaningful error messages to steer users towards using configuration file to manage their app workflows.
+
+Pre-requesites
+--------------
+
+N/A
+
+Test Scenarios/Cases
```
sanathkr: How do you plan to verify that the config file with every single CLI command we have today and in future?

designs/sam-config.md
```diff
@@ -0,0 +1,369 @@
...
+Expected Results
+----------------
+* Works on all platforms
+* Resolution of parameters follows.
+ * CLI parameters -> Config file parameters
+
+Documentation Changes
+=====================
+
+* Addition of a new `--env` parameter per command
```
sanathkr: Also need documentation on how to write the config file. How to specify flags like `--debug` that don't have explicit values. How to verify whether a config file is parsable or not.

designs/sam-config.md
TheSriram: I'm open to suggestions.

designs/sam-config.md
TheSriram: sure.

designs/sam-config.md
TheSriram: Lets add a configuration file version number

PR #1504, opened 2019-11-05
feat: Type hinting for for nodejs hello-world template

iRoachie:
*Description of changes:*
Adds type hints and autocomplete for context, and event parameters on the lambdaHandler.

Knowing what values exist in these parameters can be daunting, and usually require going to the links provided in the jsdoc. Having types here reduces that friction and keeps the developer in the editor.

### DIFF
--- a/samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/tsconfig.json
+++ b/samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es2017",
+    "checkJs": true,
+    "allowJs": true,
+    "noEmit": true
+  },
+  "exclude": ["**/node_modules", "tests/**"]
+}
--- a/samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/app.js
+++ b/samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/app.js
@@ -5,13 +5,13 @@ let response;
 /**
  *
  * Event doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
- * @param {Object} event - API Gateway Lambda Proxy Input Format
+ * @param {AWSLambda.APIGatewayEvent} event - API Gateway Lambda Proxy Input Format
  *
  * Context doc: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-context.html 
- * @param {Object} context
+ * @param {AWSLambda.Context} context
  *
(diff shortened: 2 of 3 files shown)


samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
```
jfuss: Forgive me, as I am not super familiar with Nodejs. It seems you are converting this example from Nodejs to Typescript? I can see the benefit but we want to keep the examples usable with Nodejs without the need to convert Typescript to Nodejs.

iRoachie: Nope, the project stays as NodeJs. What's happening here is that we're using typescript power the types. So you write normal javascript, but get type-hinting and completion from typescript

aahung: I think if we want to do this, it is better to do it for all templates in https://github.com/aws/aws-sam-cli-app-templates

iRoachie: I'm willing to add the other node templates to this PR if you all are interested still

aahung:
Close in favor of PRs in https://github.com/aws/aws-sam-cli-app-templates

We also have a PR adding typescript template https://github.com/aws/aws-sam-cli-app-templates/pull/77

If you are still interested, can you create PRs there?

samcli/local/init/templates/cookiecutter-aws-sam-hello-nodejs/{{cookiecutter.project_name}}/hello-world/tsconfig.json
iRoachie: Hey @aahung when you say all templates, you're referring to all the templates in `nodejs10.x`, `nodejs12.x`, `nodejs14.x`, including docker ones?

aahung: Yeah, I think it is good for consistency across templates.

iRoachie: One more thing @aahung. Some templates have more than one lambda in them so I personally don't think it makes sense to add the dependency to each function's package.json. At work, we do a top-level package.json. However it means we'd need to update the documentation as well.

Here's an example https://github.com/iRoachie/aws-sam-cli-app-templates/commit/9c303959532bce8c325b02794b489e789545efb6.

Notice the directory tree:

```
{{cookiecutter.project_name}}
 ┣ events
 ┃ ┗ event.json
 ┣ hello-world
 ┃ ┣ tests
 ┃ ┃ ┗ unit
 ┃ ┃ ┃ ┗ test-handler.js
 ┃ ┣ .npmignore
 ┃ ┣ app.js
 ┃ ┗ package.json
 ┣ .gitignore
 ┣ README.md
 ┣ package.json // this is where @types/aws-lambda will be installed
 ┗ template.yaml
```


Let me know what you think about this approach.

PR #1505, opened 2019-11-05
feat: Remove nodejs6.10 init template

iRoachie:
*Description of changes:*
Since April 30, 2019, it's no longer possible to create new functions using nodejs6.10 https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html. We should remove this option from sam init

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/init/init_generator.py
```diff
@@ -37,9 +37,6 @@ def do_generate(location, runtime, dependency_manager, output_dir, name, no_inpu
         "python3.7",
         "python3.6",
         "python2.7",
-        "nodejs",
```
sanathkr: Don't remove `nodejs`. It is just a short cut for the latest version.

iRoachie: That may've been a mistake 🤦‍♂  Good catch

iRoachie:
@sanathkr can you update your review? I made the change a while ago

PR #1521, opened 2019-11-11
design: `sam deploy` also packages built artifacts

TheSriram:
- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


designs/package_during_deploy.md
```diff
@@ -0,0 +1,175 @@
...
+Parameters that dont need to be added.
+
+* `--output-template-file`
+ * 	An explicit `output-template-file` is created on the fly during packaging in the deploy phase.
+
+If the expectation is to package and deploy in one command, One can now do.
+
+`sam deploy --stack-name sam-package-on-deploy --capabilities CAPABILITY_IAM --s3-bucket sam-package-bucket`
+
+There is no explicit need to pass in a `--template-file` parameter, if one is not passed in it to defaults to trying to find the template.yaml that was generated by `sam build` which is located at `.aws-sam/build/template.yaml`
```
jfuss: If `sam build` was never run what happens? Do we pick up the default template from the cwd or fail? All of our other commands pick up the default but could see a case to fail and leave the door open for a deploy that builds, packages, and deploys. Would be worth describing what will happen here.

designs/package_during_deploy.md
```diff
@@ -0,0 +1,175 @@
...
+
+The old method of deploying pre-packaged artifacts will continue to work as before.
+
+* `sam deploy --template-file packaged.yaml --stack-name sam-package-on-deploy --capabilities  CAPABILITY_IAM`
+
+Future of `sam package`?
+---------------------
+
+* `sam package` will continue to exist in the state it is today and will continue to be improved upon separately.
```
jfuss: Just for clarification, if we update package this new deploy flow will also get those updates?

designs/package_during_deploy.md
```diff
@@ -0,0 +1,175 @@
...
+Provide identical experiences in terms of a deploying the same stack, with exactly same artifacts.
+
+
+Implementation
+==============
+
+CLI Changes
+-----------
+
+* Add new arguments `--metadata`, `--use-json` and modify existing `--template-file` to look for a default `template.yaml`  that exists under `.aws-sam/build/`
```
jfuss: What value does --use-json provide in the deploy experience?

designs/package_during_deploy.md
```diff
@@ -0,0 +1,175 @@
...
+
+Design
+------
+
+* Changes to Deploy command's click options
+* Attempt to package on every deploy if an appropriate s3 bucket is specified and deploy using the output template file during package. 
+* If a pre-packaged template is specified, an attempt to package does not change the template and the same template is used for deploy.
+* The parameters that share the same name across package and deploy are collapsed together. eg: `--kms-key-id` , if a kms-key-id is specified that same key is used across both packaging and deploy purposes.
+
+`.samrc` Changes
```
jfuss: We should update the template in the sam config so this gets reflected in new designs going forward. Just noticing this now but not really apart of this pr directly.

designs/package_during_deploy.md
```diff
@@ -0,0 +1,175 @@
...
+
+Expected Results
+----------------
+
+* Regresssion and Integration tests pass.
+
+Documentation Changes
+=====================
+* Required nature of `--template-file` parameter has a series of defaults that are looked at during `sam deploy` similair to `sam package`. 
+* If `--template-file` points to a non-packaged template-file, `--s3-bucket` becomes required to be able to effectively package and deploy in one command using `sam deploy`.
```
jfuss: Also --template right?

designs/package_during_deploy.md
TheSriram: yeah, we pick up the default template. Can describe that case.

designs/package_during_deploy.md
TheSriram: correct. all changes will go into `PackageContext` instead, which deploy will reap the benefits of.

designs/package_during_deploy.md
TheSriram: If a given template file is above 51,200 bytes. that corresponding template also gets uploaded to an s3 bucket, if `--use-json` is provided, that uploaded template_file will be in the json format.

PR #1526, opened 2019-11-13
Managed S3 Bucket via optional bootstrap command

awood45:
Provides logic for creating or reading from a managed CloudFormation stack which contains an S3 bucket.

This is a WIP in that it should be integrated with ongoing config work, it is not ready to merge until it is tied in to seeding such a config file with the retrieved bucket name. Please do feel free to use a different branch as the base.


samcli/cli/command.py
```diff
@@ -21,6 +21,7 @@
     "samcli.commands.deploy",
     "samcli.commands.logs",
     "samcli.commands.publish",
+    "samcli.commands.setup",
```
sanathkr: can we call this `configure` to be consistent with awscli? Or `bootstrap` is also popular

awood45: We cannot call this configure as that would conflict with a future command with a different purpose. bootstrap is an option

designs/sam_setup_cmd.md
```diff
@@ -0,0 +1,25 @@
...
+* `--profile` This is associated with a user’s AWS profile, and defaults to `"default"` if not provided. It will be used for sourcing credentials for CloudFormation commands used when setting up the bucket, and for doing S3 ListBucket calls to see if a suitable bucket already exists.
+
+## Challenges
+
+Both S3 buckets and CloudFormation stacks do not have sufficiently efficient ways to search by tags. Simply put, there’s likely to be some computational inefficiency as up to hundreds of API calls might be required to identify an existing bucket that was created to be a source bucket. This means that to avoid severe performance issues, we need to make compromises. Proposed:
+
+* The default managed bucket uses a fixed stack name per region, such as “aws-sam-cli-managed-source-bucket”. If the user for some reason has a stack with that name, then we cannot support a managed bucket for them.
```
sanathkr: I'd propose we use a more generic stack name so we can support other functionality on the bucket.

samcli/commands/setup/command.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        raise UserException(msg)
+    # This bucket name is what we would write to a config file
+    msg = "Source Bucket: " + bucket_name
+    click.echo(msg)
+
+
+def _create_stack(cloudformation_client):
+    change_set_name = "InitialCreation"
+    change_set_resp = cloudformation_client.create_change_set(
```
sanathkr: Shouldn't we use `sam deploy` instead?

samcli/commands/setup/command.py
```diff
@@ -0,0 +1,125 @@
...
+    create_or_get_stack(cloudformation_client)
+
+
+def create_or_get_stack(cloudformation_client):
+    stack = None
+    try:
+        ds_resp = cloudformation_client.describe_stacks(StackName=SAM_CLI_STACK_NAME)
+        stacks = ds_resp["Stacks"]
+        stack = stacks[0]
+        click.echo("Found managed SAM CLI stack.")
```
sanathkr: Should you want to update if the stack already exists?

designs/sam_setup_cmd.md
TheSriram: I suggest we add a suffix in the end when we create it. Paves the way forward for managing multiple sam cli stacks in the future.

`aws-sam-cli-managed-source-bucket-d8f22d1e9`

samcli/lib/bootstrap/bootstrap.py
```diff
@@ -0,0 +1,108 @@
...
+from botocore.config import Config
+from botocore.exceptions import ClientError
+
+from samcli.commands.exceptions import UserException
+
+
+LOG = logging.getLogger(__name__)
+SAM_CLI_STACK_NAME = "aws-sam-cli-managed-stack"
+
+MANAGED_STACK_DEFINITION = """
```
TheSriram: * this lives in a separate file?
* can we add the version number of sam cli with which this stack was created?
Any other cheap metadata we can add today?

sanathkr: * separate file: Yeah, the template is small. it's cheap to have it here. we will move out later
* metadata: Good idea!

PR #1532, opened 2019-11-15
Package and deploy in one command

TheSriram:
- if the template is left unspecified, built artifacts are chosen and s3 bucket is optionally needed to deploy in one command.
- existing workflows to package and deploy do not break.

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -12,6 +12,7 @@ from samcli.commands._utils.options import (
     notification_arns_override_option,
     template_click_option,
 )
+from samcli.cli.cli_config_file import configuration_option, TomlProvider
 from samcli.cli.main import pass_context, common_options, aws_creds_options
 from samcli.lib.telemetry.metrics import track_command
 
@@ -100,6 +101,7 @@ e.g. sam deploy --template-file packaged.yaml --stack-name sam-app --capabilitie
 @tags_override_option
 @parameter_override_option
 @capabilities_override_option
+@configuration_option(provider=TomlProvider(section="deploy"))
 @aws_creds_options
 @common_options
 @pass_context
@@ -158,7 +160,21 @@ def do_cli(
     region,
     profile,
 ):
+    from samcli.commands.package.package_context import PackageContext
     from samcli.commands.deploy.deploy_context import DeployContext
+
+    import ipdb
+    ipdb.set_trace()
+
+    with PackageContext(
+        template_file=template_file,
+        s3_bucket=s3_bucket,
+        s3_prefix=s3_prefix,
+        output_template_file='sam-cli-packaged-'+template_file,
+        region=region,
+        profile=profile
+    ) as package_context:
+        package_context.run()
 
     with DeployContext(
         template_file=template_file,
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -2,9 +2,10 @@
 CLI command for "deploy" command
 """
 
+import tempfile
+
 import click
 
-from samcli.cli.cli_config_file import configuration_option, TomlProvider
 from samcli.commands._utils.options import (
     parameter_override_option,
     capabilities_override_option,
@@ -12,6 +13,7 @@ from samcli.commands._utils.options import (
     notification_arns_override_option,
     template_click_option,
 )
+
 from samcli.cli.cli_config_file import configuration_option, TomlProvider
 from samcli.cli.main import pass_context, common_options, aws_creds_options
 from samcli.lib.telemetry.metrics import track_command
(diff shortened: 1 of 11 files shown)


samcli/commands/deploy/command.py
```diff
@@ -28,14 +31,14 @@
...
 
-@configuration_option(provider=TomlProvider(section="parameters"))
 @click.command(
     "deploy",
     short_help=SHORT_HELP,
     context_settings={"ignore_unknown_options": False, "allow_interspersed_args": True, "allow_extra_args": True},
     help=HELP_TEXT,
 )
-@template_click_option(include_build=False)
+@configuration_option(provider=TomlProvider(section="parameters"))
```
sanathkr: is the order of decorators important?

samcli/commands/deploy/command.py
```diff
@@ -28,14 +31,14 @@
...
+@template_click_option(include_build=True)
```
sanathkr: Is this a breaking change or a new feature?

`sam build` -> `sam deploy` will now start to work, right?
`sam build` -> `sam package` -> `sam deploy without --template` will use the built template instead of the packaged template, right?

samcli/commands/deploy/command.py
```diff
@@ -96,6 +99,14 @@
     "changes to be made to the stack. The default behavior is  to  return  a"
     "non-zero exit code.",
 )
+@click.option(
```
sanathkr: I know we talked about this before, why is this explicitly required? can't we default to using JSON for the intermediate template?

samcli/commands/deploy/command.py
```diff
@@ -154,27 +169,47 @@ def do_cli(
...
+            output_template_file=output_template_file.name,
+            kms_key_id=kms_key_id,
+            use_json=use_json,
+            force_upload=force_upload,
+            metadata=metadata,
+            on_deploy=True,
+            region=region,
+            profile=profile,
+        ) as package_context:
+            package_context.run()
```
sanathkr: why are you packaging *always*? This would be a breaking change

samcli/commands/package/exceptions.py
```diff
@@ -72,3 +72,12 @@ def __init__(self, template_file, ex):
         super(PackageFailedError, self).__init__(
             message=message_fmt.format(template_file=self.template_file, ex=self.ex)
         )
+
+
+class NoSuchBucketError(UserException):
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
+
+        message_fmt = "\n S3 Bucket does not exist."
```
sanathkr: Space at start of line - right after `\n`.

Why did you remove the use `aws s3 mb` text from the message?

samcli/commands/package/package_context.py
```diff
@@ -67,6 +68,7 @@ def __init__(
         self.metadata = metadata
         self.region = region
         self.profile = profile
+        self.on_deploy = on_deploy
```
sanathkr: so this is the special case for package happening as part of deploy?

samcli/commands/deploy/command.py
TheSriram: not after, we made template `is_eager=True`

samcli/commands/deploy/command.py
TheSriram: Yes. but today sam deploy requires --template-file to be specified.

being able to run sam deploy without --template-file is a new feature.

samcli/commands/deploy/command.py
TheSriram: Sure, we can. user had the option with package, do we gain anything by removing it from deploy?

samcli/commands/deploy/command.py
TheSriram: No. its an attempt to package, if the codeuri's already contain packaged contents, we do not repackage.

samcli/commands/package/exceptions.py
TheSriram: `sam bootstrap` should come here logically.

samcli/commands/package/package_context.py
TheSriram: I'm just muting some output that comes on the terminal, afaik.

samcli/commands/deploy/command.py
sanathkr: okay

samcli/commands/deploy/command.py
sanathkr: Ah okay. We have to consolidate the myriad of options sometime in the future.

samcli/commands/deploy/command.py
sanathkr: This is skirting the lines of a breaking change. Is there any other mechanism to create two clear code paths, one for previous experience and other for new one?

And, this relies on the fact that `sam package` is idempotent. That's an undocumented assumption of `sam package`.

samcli/commands/package/exceptions.py
sanathkr: right. can you add that?

samcli/commands/package/package_context.py
sanathkr: sure, but this feels hacky. May be in the future we will find more general use of this variable.

samcli/commands/deploy/command.py
sanathkr: There is no obvious user benefit of this flag in `sam deploy`. Why expose then?

samcli/commands/deploy/command.py
TheSriram: https://github.com/awslabs/aws-sam-cli/issues/1533. Added this question here.

PR #1534, opened 2019-11-15
fix: deploy error on not supplying s3 bucket

TheSriram:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


sanathkr:
`deploy` smoke tests failed because with this change you have allowed package to run before deploy. Now package is trying to access creds and fails.

We need to think more deeply about the back-compat story of always running package before deploy. I think it is complicated.

PR #1536, opened 2019-11-15
feat: Guided interactive deployment

sanathkr:
*Description of changes:*
`sam deploy -i` now provides an interactive guided deployment experience. 

Remaining:
- [x] Store parameters in samconfig.toml
- [ ] Unit tests are still pending. 

**Screenshot**
Screen Shot 2019-11-15 at 1 40 31 AM

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -6,17 +6,15 @@ CLI configuration decorator to use TOML configuration files for click commands.
 ## SPDX-License-Identifier: MIT
 
 import functools
-import os
 import logging
 
 import click
-import toml
+from samcli.cli.context import get_cmd_names
+from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV
 
 __all__ = ("TomlProvider", "configuration_option", "get_ctx_defaults")
 
-LOG = logging.getLogger("samcli")
-DEFAULT_CONFIG_FILE_NAME = "samconfig.toml"
-DEFAULT_IDENTIFER = "default"
+LOG = logging.getLogger(__name__)
 
 
 class TomlProvider:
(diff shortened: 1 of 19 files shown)


samcli/commands/deploy/command.py
```diff
@@ -172,12 +185,23 @@ def do_cli(
...
     tags,
     metadata,
+    interactive,
     region,
     profile,
 ):
     from samcli.commands.package.package_context import PackageContext
     from samcli.commands.deploy.deploy_context import DeployContext
 
+    confirm_changeset = False
```
TheSriram: confirm changeset also needs to be its own flag in the CLI

sanathkr: Good point. Will add

samcli/commands/deploy/command.py
```diff
@@ -30,18 +34,21 @@
 \b
 """
 
+CONFIG_SECTION = "parameters"
```
TheSriram: At a later stage we can move such section defintions to a utils file.

samcli/commands/deploy/command.py
```diff
@@ -172,12 +189,33 @@ def do_cli(
...
+                template_file,
+                stack_name=stack_name,
+                s3_bucket=s3_bucket,
+                region=region,
+                profile=profile,
+                confirm_changeset=confirm_changeset,
+            )
+
+        # We print deploy args only on interactive.
+        # Should we print this always?
```
TheSriram: +1 to printing always

samcli/cli/cli_config_file.py
```diff
@@ -29,44 +28,41 @@ class TomlProvider:
...
         :returns dictionary containing the configuration parameters under specified config_env
         """
+
         resolved_config = {}
-        try:
-            config = toml.load(file_path)
-        except Exception as ex:
-            LOG.error("Error reading configuration file :%s %s", file_path, str(ex))
+
+        samconfig = SamConfig(config_dir)
```
awood45: Is there any ability for users to pass a custom file name? As far as I can tell, we don't have a code path for that here.

samcli/commands/_utils/options.py
```diff
@@ -149,6 +154,8 @@ def capabilities_click_option():
     return click.option(
         "--capabilities",
         cls=OptionNargs,
+        default=("CAPABILITY_IAM",),
```
awood45: Remove this.

samcli/commands/deploy/command.py
```diff
@@ -30,18 +36,21 @@
...
+
 
 @click.command(
     "deploy",
     short_help=SHORT_HELP,
     context_settings={"ignore_unknown_options": False, "allow_interspersed_args": True, "allow_extra_args": True},
     help=HELP_TEXT,
 )
-@configuration_option(provider=TomlProvider(section="parameters"))
+@configuration_option(provider=TomlProvider(section=CONFIG_SECTION))
```
awood45: I get what you're doing here, but I wonder if this should be a shared constant, and you're saying `CONFIG_PARAM_SECTION`, because if there are ever other sections, this won't be easy to understand.

Your call on this, it can always be refactored later.

samcli/commands/deploy/command.py
```diff
@@ -172,12 +199,45 @@ def do_cli(
...
+    interactive,
+    confirm_changeset,
     region,
     profile,
 ):
     from samcli.commands.package.package_context import PackageContext
     from samcli.commands.deploy.deploy_context import DeployContext
 
+    changeset_decision = False
+    _capabilities = ("CAPABILITY_IAM",)
```
awood45: Why is this set here?

samcli/cli/cli_config_file.py
TheSriram: In the future, not now.

samcli/commands/deploy/command.py
TheSriram: I called out on this, as well. later refactor

samcli/commands/_utils/options.py
TheSriram: Done.

samcli/commands/deploy/command.py
TheSriram: set at one place.

PR #1538, opened 2019-11-15
fix: override botocore logging to a NullHandler

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/main.py
+++ b/samcli/cli/main.py
@@ -97,6 +97,8 @@ def cli(ctx):
     sam_cli_logger = logging.getLogger("samcli")
     sam_cli_formatter = logging.Formatter("%(message)s")
     lambda_builders_logger = logging.getLogger("aws_lambda_builders")
+    botocore_logger = logging.getLogger("botocore")
 
     SamCliLogger.configure_logger(sam_cli_logger, sam_cli_formatter, logging.INFO)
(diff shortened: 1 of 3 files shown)


samcli/cli/main.py
```diff
@@ -97,6 +97,8 @@ def cli(ctx):
     sam_cli_logger = logging.getLogger("samcli")
     sam_cli_formatter = logging.Formatter("%(message)s")
     lambda_builders_logger = logging.getLogger("aws_lambda_builders")
+    botocore_logger = logging.getLogger("botocore")
 
     SamCliLogger.configure_logger(sam_cli_logger, sam_cli_formatter, logging.INFO)
     SamCliLogger.configure_logger(lambda_builders_logger, sam_cli_formatter, logging.INFO)
+    SamCliLogger.configure_null_logger(botocore_logger)
```
sanathkr: Does this silence all logs, including debug logs?

jfuss: https://docs.python.org/3.6/howto/logging.html#configuring-logging-for-a-library explains it. Basically it will prevent the library’s logged event being output to sys.stderr and is documented as "A do-nothing handler". This is also what boto3 is doing but botocore is not (for whatever reason).

From reading the docs, my understanding is yes.

PR #1546, opened 2019-11-18
feedback on sam config design

diablodale:
feedback as requested by @sanathkr in issue https://github.com/awslabs/aws-sam-cli/issues/1490


designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+Default Config file location: .aws-sam/samconfig.toml
+..
+..
+..
+```
+
+Why samconfig under `.aws-sam`
+---------------------------------
+
+> The below two don't answer the above question. Instead, they give technical implementation answers to questions not asked like "How is the .aws-sam directory made?" and "Are any files in .aws-sam git ignored?". The ideas surfaced in the previous PR's discussion were lost. For example, in the PR threads it was for a short time in the root of the project but then put back into .aws-sam. I recommend clearly stating here an answer to the above question in terms of goals, not technical implementations.
```
TheSriram: This is a good point and callout on the back and forth during the PR discussion. Will update the design doc with the decision and why it was taken.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+
+The `.gitignore` specified in the init apps also only have `.aws-sam/build` ignored and not anything else.
+
+
+Config file versioning
+-----------------------
+
+The configuration file: `samconfig.toml` will come with a top level version key that specifies the grammar version of the configuration file. This version can then be used to determine if a given configuration file works with a version of SAM CLI. A given version of SAM CLI may support one or more configuration file grammar versions.
+
+> The above clarifies that the version is not the version of the file itself (like an ever increasing commit version), instead it is the version of the grammar/DTD/spec of the file.
```
TheSriram: 👍 can make this more explicit.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+
+The default location of a .aws-sam/samconfig can be replaced by overriding an environment variable called `SAM_CLI_CONFIG`
+
+`
+export SAM_CLI_CONFIG=~/Users/username/mysamconfig.toml
+`
+
+Users can pass an environment `--env` for the section that will be scanned within the configuration file to pass parameters through.
+
+> I hesitate in using `--env` as the param that controls what prefix of TOML tables/sections is parsed/loaded. Using `--env` strongly associates a specific usage scenario (environment choice) to a feature that is much broader in potential usage. Also, `--env` doesn't to me clarify what it controls/changes. At first read, I think it controls environment variables. Perhaps an alternative like `--config-prefix` which is no longer limited to a specific usage scenario while more clearly describing what it controls.
```
TheSriram: Yes, this is the reason why the `--env` is hidden and ignored right now. To be able to support a better naming nomenclature.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+
+> The above clarifies that the version is not the version of the file itself (like an ever increasing commit version), instead it is the version of the grammar/DTD/spec of the file.
+
+It also paves the forward when major changes need to be made to the configuration file and add a version bump to the config file version
+
+```
+version = 0.1
+```
+
+> What is behavior of SAM CLI for different scenarios, e.g.:
```
TheSriram: 1 & 2 are covered in the implementation, whereas 3 is not. Can document what the behaviors will be.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+[default.deploy]
+[default.deploy.parameters]
+stack_name="using_config_file"
+capabilities="CAPABILITY_IAM"
+region="us-east-1"
+profile="srirammv"
+
+```
+
+> I am not in favor of having empty TOML tables/sections like `[default.package]`.
```
TheSriram: This might probably be due to implementation details with tomlkit, Will create an explicit issue for this.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+```
+.aws-sam/samconfig-dev.toml
+.aws-sam/samconfig-beta.toml
+.aws-sam/samconfig-prod.toml
+```
+
+`--env` dev will refer to `.aws-sam/samconfig-dev.toml` and so on.
+
+> I'm not in favor of parsing file names in lieu of TOML table/section prefixes. What are the TOML tables in the first file? Does it contain `[dev.build.parameters]` or just `[build.parameters]` or `[default.build.parameters]`. First is redundant, second is out-of-spec, third is potentially confusing since there could be defaults elsewhere.
+> Instead, I support loading of multiple config files in order they are specified. Similar to how `docker-compose` can have multiple `-f filename` parameters. However, then comes the discussion of merge rules: overwrite/append keys, extend arrays, and application of these merge rules across files across defaults, merge rules for config files of different spec versions, etc.?
```
TheSriram: This was infact part of the first implementation and then removed for design on this particular hierachy first. But definitely on the radar and will create another issue for this.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+
+```
+.aws-sam/samconfig-dev.toml
+.aws-sam/samconfig-beta.toml
+.aws-sam/samconfig-prod.toml
+```
+
+`--env` dev will refer to `.aws-sam/samconfig-dev.toml` and so on.
+
+> I'm not in favor of parsing file names in lieu of TOML table/section prefixes. What are the TOML tables in the first file? Does it contain `[dev.build.parameters]` or just `[build.parameters]` or `[default.build.parameters]`. First is redundant, second is out-of-spec, third is potentially confusing since there could be defaults elsewhere.
```
TheSriram: default is the name of the profile (config-env) here  and showcases granularity and depth of where the configuration values affect the behavior of samcli.

designs/sam-config.md
```diff
@@ -0,0 +1,413 @@
...
+
+* Potentially every sam command could have functionality to have a series of command line parameters exported into a configuraion file.
+
+
+Out-of-Scope
+------------
+
+* Not focusing on a global configuration. SAM CLI already has a notion of a global config at `~/.aws-sam/metadata.json`
+
+> This is confusing/conflicting with a few paragraphs above is written, "This way, configuration that might be global can be placed...". While technically such could be possible, I recommend not writing of a technically possible scenario that conflicts with what is stated as out-of-scope.
```
TheSriram: I think this was the effect of the back and forth on the design, will take this into consideration with the refactor on the design doc to accurately reflect present day and paths forward.

sriram-mv:
Closing this PR, Thanks for the feedback on the design, really useful.

PR #1547, opened 2019-11-18
enabled extra_context to work with custom cookiecutter template

gupta-n:
updated extra_context cli parameter to extra-context

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/init/__init__.py
```diff
@@ -90,7 +90,7 @@
     help="Disable Cookiecutter prompting and accept default values defined template config",
 )
 @click.option(
-    "--extra_context",
+    "--extra-context",
```
nikp: while this would be more consistent with the other SAM CLI parameters, it may be too late to change this since this was already released as v 0.31.0 :(

samcli/commands/init/__init__.py
```diff
@@ -149,15 +149,13 @@ def do_cli(
     # check for required parameters
     if location or (name and runtime and dependency_manager and app_template):
         # need to turn app_template into a location before we generate
+        default_context = None
         if app_template:
             templates = InitTemplates(no_interactive, auto_clone)
             location = templates.location_from_app_template(runtime, dependency_manager, app_template)
             no_input = True
             default_context = {"project_name": name, "runtime": runtime}
```
nikp: why is default_context only initialized with the name/runtime parameters in the app_template flow? is there no expetations to be provided those parameters in location flow? I suppose I can understand, but is that a documented assumption? 

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-init.html only says that "This parameter is required if --no-interactive is specified and --runtime, --name, and --app-template are not provided.", but that doesn't necessarily imply that the reverse is true.

Should we just use this as the default in all cases? (assuming there is a name && runtime)

gupta-n: As per my understanding currently other parameters like runtime, name are use to determine template and then passed in cookie cutter template. Adding them in location flow means I am assuming customer is going to use these parameters in cookie cutter template.

nikp: for sure - "location" can be any github repo or directory, and they could have parameterized runtime and project_name or others.

might be a safe call

jfuss: I agree with @nikp here. I don't think `{"project_name": name, "runtime": runtime}` is specific to app_template and should be moved up to be included with `--location`. This gives the ability to other authors to have access to these values.

samcli/commands/init/__init__.py
jfuss: Talked with @awood45 on this. Consistency with all our other options is important. This is a new field so we should adjust it now.

samcli/commands/init/__init__.py
gupta-n: User will have access to {"project_name": name, "runtime": runtime} by setting this as part of extra_context. I am assuming SAM init with location will only use extra_context no other parameter. If it's not true I am happy to make it as default parameter.

tests/unit/commands/init/test_cli.py
```diff
@@ -139,6 +139,7 @@ def test_init_cli_interactive_multiple_dep_mgrs(self, generate_project_patch, sd
 2
 test-project
 N
+1
```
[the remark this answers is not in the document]
gupta-n: Sure

samcli/commands/init/__init__.py
jfuss: But they can also pass `--name` and `--runtime` into the command. Why would we want customers to be duplicating these values? How will customers know they need to do that vs just pass the option to the command?

gupta-n: In case of location if customer passes name and runtime we assume its cookie cutter template variable is it right understanding?

gupta-n: I am happy to update if this is the expectation.

awood45: I think it's reasonable to map the `--name` param to the context `project_name` and the `--runtime` param to the `runtime` context.

jfuss: @gupta-n Yes. We want to start passing this information to the templates that are given through `--location`. This should simplify some of this code as well.

samcli/commands/init/__init__.py
```diff
@@ -177,11 +173,24 @@ def do_cli(
...
+    if runtime is not None:
+        default_context = {"runtime": runtime}
+    if name is not None:
+        if default_context is not None:
+            default_context = {**default_context, **{"project_name": name}}
+        else:
+            default_context = {"project_name": name}
+    if extra_context is not None:
+        try:
+            extra_context = extra_context.encode().decode("unicode_escape")
```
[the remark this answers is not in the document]
gupta-n: this is to enable usecase where user is passing escaped quotes. I am not able to test as black formatting removes that. I have tested that manually. Let me know if you know how I can test that?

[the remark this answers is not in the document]
gupta-n: e.g: --extra-context='{\"AWS_Schema_registry\":\"a\",\"AWS_Schema_name\":\"b\",\"AWS_Schema_source\":\"c\",\"AWS_Schema_detail_type\":\"d\",\"user_agent\":\"AWSToolkit\"}'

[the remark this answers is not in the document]
gupta-n: funny :) it got removed here as well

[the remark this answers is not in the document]
jfuss: You should turn black off with `fmt:off` and then back on after the line `fmt:on`.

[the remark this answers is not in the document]
gupta-n: updated.

PR #1549, opened 2019-11-19
chore: Support Java11, Python3.8, and Nodejs12.x runtimes

jfuss:
*Issue #, if available:*

*Description of changes:*
What is left to do:
- [x] Update appveyor to run java11 for in-process builds for windows
- [x] Update appveyor to run java11 for in-process builds on linux
- [x] Update appveyor to run python3.8 for in-process builds
- [x] Update aws-lambda-builders to 0.6.0 after it is released

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mhart:
`aws-lambda-builders` has been updated to `0.6.0` on all `build-*` images 👍

appveyor.yml
```diff
@@ -113,7 +132,10 @@ for:
       - sh: "pytest -vv tests/integration"
       - sh: "pytest -vv -n 4 tests/regression"
       - sh: "/tmp/black --check setup.py tests samcli scripts"
-      - sh: "python scripts/check-isolated-needs-update.py"
+
+      # Set JAVEHOME to java11
```
c2tarun: typo.

appveyor.yml
```diff
@@ -113,7 +132,10 @@ for:
...
+      - sh: "JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64"
```
c2tarun: JAVA_HOME for java8 is not set. I am not sure how appveyor works, but can this be an issue on next run, when JAVA_HOME will be for java11 and test would run expecting them to be 8?

c2tarun: Nevermind, now that I think about it, it shouldn't be an issue since it'll be a new session.

appveyor.yml
jfuss: Done

appveyor.yml
jfuss: We set `JAVA_HOME` to java8 on line 82 (first step in the install step). The first pass of build tests will run with java8. I needed a way to also run the Java11 build tests. Instead of creating another appveyor project, I edit the `JAVA_HOME` env var to be java11 within the test_script phase of appveyor.

appveyor.yml
```diff
@@ -98,6 +100,25 @@ for:
...
+      - sh: "sudo apt-get -y install python3.7"
+      - sh: "sudo apt-get update"
+      - sh: "sudo apt-get -y install python3.8"
+
+      - sh: "which python3.8"
+      - sh: "which python3.6"
+      - sh: "which python3.7"
+      - sh: "which python2.7"
+
+      - sh: "PATH=$PATH:/usr/bin/python3.8:/usr/bin/python3.7"
```
TheSriram: do 3.6 and 2.7 also need to be added to path?

samcli/local/common/runtime_template.py
```diff
@@ -99,17 +102,27 @@
     itertools.chain(*[c["runtimes"] for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))])
 )
 
+# Order here should be a the group of the latest versions of runtimes followed by runtime groups
 INIT_RUNTIMES = [
-    "nodejs10.x",
-    "python3.7",
+    # latest of each runtime version
```
TheSriram: I think this list is what is used to showcases the runtimes during `init`, we should make this ordered, with a set maybe.

jfuss: This is ordered just in a list. I can update it to a set but there isn't a whole difference here (other than duplicate entry guard).

This is used within interactive init not in the help text.

appveyor.yml
jfuss: 2.7 and 3.6 are already there: https://www.appveyor.com/docs/linux-images-software/#python

I am considering moving off of the pre-installed everything in favor how installing python versions through apt-get. This will make things a little more consistent in appveyor, but didn't want to go that deep here.

jfuss:
@TheSriram I added debug support for Python3.8 and Node12.x, which are based on the previous runtimes. Java11 needs more investment before we can confidently support debugging. The reason for this is because Java11 containers are based on provided and the overriding mechanism is slightly different. I don't want to hold up releasing everything else so my recommendation is to go forward without debug support. In the meantime, I will be spending time on ensuring things work before adding that related debug config.

Issue #1537, opened 2019-11-15
ScriptLocation is not updated for Glue::Job when building


wnkz:
### Description

When running `sam build` ; `AWS::Glue::Job` `Command.ScriptLocation`'s path is not updated in `.aws-sam/build/template.yaml`.

My real template includes Serverless::Functions so I need to run build and cannot just use package.
This works just fine for other resources supported by `sam package` (e.g. `AWS::CloudFormation::Stack`) but not for Glue Jobs.

### Steps to reproduce

template.yml
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Resources:
  GlueJob:
    Type: AWS::Glue::Job
    Properties:
      GlueVersion: '1.0'
      MaxCapacity: 5
      Command:
        Name: glueetl
        PythonVersion: '3'
        ScriptLocation: glue.py
      Role: !GetAtt GlueServiceRole.Arn

  GlueServiceRole:
    Type: AWS::IAM::Role
    Properties:
      Path: /service-role/
      AssumeRolePolicyDocument:
        Version: 2012-10-17
        Statement:
          - Effect: Allow
            Principal:
              Service:
                - glue.amazonaws.com
            Action:
              - sts:AssumeRole
      ManagedPolicyArns:
        - arn:aws:iam::aws:policy/service-role/AWSGlueServiceRole
```

```
sam build
```

### Observed result

.aws-sam/build/template.yaml
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Resources:
  GlueJob:
    Type: AWS::Glue::Job
    Properties:
      GlueVersion: '1.0'
      MaxCapacity: 5
      Command:
        Name: glueetl
        PythonVersion: '3'
        ScriptLocation: glue.py
      Role:
        Fn::GetAtt:
        - GlueServiceRole
        - Arn
  GlueServiceRole:
    Type: AWS::IAM::Role
    Properties:
      Path: /service-role/
      AssumeRolePolicyDocument:
        Version: 2012-10-17
        Statement:
        - Effect: Allow
          Principal:
            Service:
            - glue.amazonaws.com
          Action:
          - sts:AssumeRole
      ManagedPolicyArns:
      - arn:aws:iam::aws:policy/service-role/AWSGlueServiceRole
```

Command.ScriptLocation is left untouched.

### Expected result

.aws-sam/build/template.yaml
[...]

PR #1550, opened 2019-11-19
fix: Handle dot notations in the path_prop_name variables (#1537)

wingkwong:
*Issue #, if available:*
#1537

*Description of changes:*
Support dot notations in the path_prop_name variables under ``samcli/commands/_utils/template.py``

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


sanathkr:
There are some changes in this codepath that could change your code. Can you rebase on latest `develop`?

wingkwong:
@sanathkr Merged with the latest ``develop``.

jfuss:
@wingkwong I am not sure this PR actually solves the problem of the issue. If all of our code was went through this Template class, this would work but build does not (Template is used only by the package command from what I could trace and understand). The package command in the `artifact_exporter.py` file uses a library called (jmespath)[https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/package/artifact_exporter.py#L133] to handle the querying for nested attributes in json. Package handles this properly but build updates the template differently (due to when things where built) [here](https://github.com/awslabs/aws-sam-cli/blob/4f2c345c5e415ec443cb55b82dd7cbb7b0ba2ad6/samcli/lib/build/app_builder.py#L108). This method is not aware of the '.' notation and therefore won't update the template correctly.

Maybe I am missing something, where this PR is actually solving the issue. I think what we will want to do here is create another dummy glue resource [here](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/unit/lib/build_module/test_app_builder.py#L49) and update this [test](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/unit/lib/build_module/test_app_builder.py#L57) to assert the dot notation update is actually happening. Would you be able to add this tests? This will help us with ensure this doesn't break in the future and help us understand this is doing what we are expecting.

wingkwong:
@jfuss Thanks for the clarification. I'll take a look later.

wingkwong:
@jfuss I think my fix updates the path inside [move_template](https://github.com/wingkwong/aws-sam-cli/blob/fix-dot-notation/samcli/commands/build/command.py#L195) -> [_update_relative_paths](https://github.com/wingkwong/aws-sam-cli/blob/fix-dot-notation/samcli/commands/_utils/template.py#L76) instead of [update_template](https://github.com/wingkwong/aws-sam-cli/blob/fix-dot-notation/samcli/commands/build/command.py#L193) . If you run ``sam build``, ScriptLocation is able to be updated to ``../../glue.py`` because the template is updated before moving it to the destination path. Or you prefer to move this logic back to build app_builder -> update_template and add the tests as you stated?

wingkwong:
@jfuss on second thought, there is no functions_to_build in this case which means ``artifacts`` should be {} . ``update_template`` returns the original dict. Besides, based on the description of the method ``update_template`` in ``app_builder.py``, I think we should update the path in ``_update_relative_paths`` instead.

jfuss:
@wingkwong Good catch. I think `_update_relative_paths` is the right place. I thought the `update_template` did both for some reason.

wnkz:
@jfuss @wingkwong How close are we to getting this merged?  
I really need this to work and it's been almost two months now. I can help with reviewing, testing or even code if needed. Please let me know.

wingkwong:
@wnkz I don't have further changes to make. I've been waiting for someone to review it.

jfuss:
@wingkwong Apologize on the delay here. The missing piece is another test. I think this got lost in one of my comments above.

I think what we will want to do here is create another dummy glue resource [here](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/unit/lib/build_module/test_app_builder.py#L49) and update this [test](https://github.com/awslabs/aws-sam-cli/blob/develop/tests/unit/lib/build_module/test_app_builder.py#L57) to assert the dot notation update is actually happening. Would you be able to add this tests? This will help us with ensure this doesn't break in the future and help us understand this is doing what we are expecting.

I know you have added a test within the `test_template.yaml` file but I think we need one (or an integ tests) that covers this case for build specifically.

wingkwong:
@jfuss Added. Please take a look to see if it is what you want. Thanks.

jfuss:
@wingkwong Thanks for adding the extra tests. While reviewing, I noticed that you were not using jmespath (allows searching json based on dot notation). In the package command, we already use jmespath to do something very similar and why we have the dot notation defined in the [resources.py](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/_utils/resources.py#L35). I updated your implementation to mimic what we do in package already.

I also added an integ tests to validate that we update the nested property correctly. This is what I thought the `test_app_builder` unit tests was more or less doing but on deeper inspection, I noticed I was wrong. I went ahead and updated it to reduce back and forth. Really, appreciate the effort you put into this and keeping it updated over time!

PR #1552, opened 2019-11-19
fix: Remove attach_api and replace with docker library calls

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/docker/container.py
```diff
@@ -200,7 +199,7 @@ def wait_for_logs(self, stdout=None, stderr=None):
         real_container = self.docker_client.containers.get(self.id)
 
         # Fetch both stdout and stderr streams from Docker as a single iterator.
-        logs_itr = attach(self.docker_client, container=real_container, stdout=True, stderr=True, logs=True)
+        logs_itr = real_container.attach(stream=True, logs=False, demux=True)
```
TheSriram: Is this with the latest version of docker? do we need any version bumps?

jfuss: This was released in 3.7 (looking at https://github.com/docker/docker-py/issues/1952). We are on 4.0 or greater, so we should be covered and not need any bump.

thomas-fossati: Setting `logs=False` changes the previous behaviour.  Is this intentional?

jfuss: @thomas-fossati What behavior. I have asked a similar question on #1566. It is not clear to me what you are @setrofim are referring to. Please be more specific.

thomas-fossati: Apologies for the extreme terseness of my previous comment.

If the lambda business logics completes before SAM attaches to the container, the lambda's output will only be available through the logs.  So if you don't ask the logs to be forwarded on attach, you might end up seeing no output, including the bits that SAM cares about (ie., `{ "status": ... }`) and get stuck.

jfuss: @setrofim had a small back and forth here: https://github.com/awslabs/aws-sam-cli/pull/1566#issuecomment-557115763 I think I understand the reasoning more.

It was intentionally `False` because I took this from the 'warm container' PR in which we do not want all the logs from the container. Since we are not in that mode yet, making this `True` makes sense to me. We will need to filter logs by invocation for the warm container case but we can tackle all that at once.

thomas-fossati: Sounds like a good plan to me, thanks!

Issue #1551, opened 2019-11-19
The attach API implementation does not work as expected


thomas-fossati:
### Description

The [Attach API](https://docs.docker.com/engine/api/v1.30/#operation/ContainerAttach) implementation does not work as expected.

In particular, the status report sent by the Lambda
```
{"statusCode":...,"headers":{...},"multiValueHeaders":...,"body":...}
```
might end up being ignored, resulting in a 502 status response with the following
```
Function returned an invalid response (must include one of: body, headers, multiValueHeaders or statusCode in the response object)
```

### Steps to reproduce

This might be tricky to repro.  In particular, for us it never showed up on bare metal.  We have started seeing it consistently when running SAM inside a VM.

What we can do at this stage is to give you the BoM of our failing VM.  See below in the "Additional environment details" section.

### Observed result

The Attach API implementation is split into two logically separate bits: first the client does a `POST /containers/.../attach?...` using the `requests` library machinery, asking to be attached to the running container.  Then, after receiving the 101 "switching protocols" response from the Docker engine, the client pilfers the underlying socket descriptor from the `requests.Response` instance and consumes the binary protocol which muxes stdout and stderr coming from the container.

The problem is that at the time the socket descriptor is stolen, the `requests` library might have already read, either partially or entirely, the binary stream following the 101 from Docker.  Therefore, when the actual raw read happens, there might be no data left to consume from the socket -- the bytes are securely stashed inside the `requests.Response` buffer (and can be looked at using `peek()`) but are ignored by the binary protocol processor.

(Note that this behaviour has never been observed on bare metal.  It started showing up in a VM where the socket read buffer is apparently coalesced more aggressively.)

The following extract is from a `strace(1)` of a successful invocation. The `sendto` is the client doing the POST to the attach API endpoint. The first of the subsequent `recvfrom` is done by the `requests` library.  Note that ~580 bytes of "good" output get swallowed and won't be available to the attached client.  However, it doesn't really matter because the next `recvfrom` (after `sd=13` has been pilfered from the `requests.Response`) has data to consume at a sane boundary and the only thing that SAM cares about for declaring victory is the `{"statusCode":...,"headers":{...},"multiValueHeaders":...,"body":...}` which is due to arrive later.
So this, at least partially, succeeds.
```
[pid 15304] sendto(13, "POST /containers/027cc9a5f6d03c6"..., 302, 0, NULL, 0) = 302
[...]

PR #1566, opened 2019-11-21
fix: collect prior output from attached containers

setrofim:
When attaching to a docker container, specify logs=True in order to
receive the output the container has generated prior to it being
attached.

*Issue #, if available:*

Related to #1551

PR #1552 switched to using Docker API for attaching to containers. However, it is invoked with `logs=False` so that only the output generated by the container after attaching is collected. This means that some initial output may get lost.

*Description of changes:*

Specify `logs=True` when attaching to a container to ensure the complete output is collected.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -199,7 +199,7 @@ class Container:
         real_container = self.docker_client.containers.get(self.id)
 
         # Fetch both stdout and stderr streams from Docker as a single iterator.
-        logs_itr = real_container.attach(stream=True, logs=False, demux=True)
+        logs_itr = real_container.attach(stream=True, logs=True, demux=True)
 
         self._write_container_output(logs_itr, stdout=stdout, stderr=stderr)
 
--- a/tests/unit/local/docker/test_container.py
+++ b/tests/unit/local/docker/test_container.py
@@ -451,7 +451,7 @@ class TestContainer_wait_for_logs(TestCase):
 
         self.container.wait_for_logs(stdout=stdout_mock, stderr=stderr_mock)
 
-        real_container_mock.attach.assert_called_with(stream=True, logs=False, demux=True)
+        real_container_mock.attach.assert_called_with(stream=True, logs=True, demux=True)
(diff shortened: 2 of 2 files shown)


jfuss:
@setrofim Can you be more specific on the output that is lost? I am not sure what you are referring to.

setrofim:
@jfuss The output between when the container is stared by the `container_manager`  inside `invoke()` and is attached to inside `wait_for_logs()`. Not much happens between these two points, but it  seems that occasionally, the attach happens too late, and we miss output from some of our lambdas, resulting in intermittent failures (basically, the same symptoms as in  #1551, but due to a different timing issue). Setting `logs=True` ensures that we collect output from the start of the container, regardless of the timing of the `attach()`.

jfuss:
@setrofim Makes sense. I don't know if we really want to do `logs=True`. I took this from another PR I had open for `warm containers` #1305. In that case, we actually do not want all the logs from the container. 

I initially made the change in #1552 thinking the bug was within our attach implementation and could instead just use dockers. As you and @thomas-fossati have pointed out, this is actually not the case. Instead of just introducing another failure/case into the mix, I am going to revert the #1552 PR so it does not get released. I want to do a little deeper understanding of these things to make a better call on next steps.

Really appreciate the dive you guys have done here! It is very helpful.

setrofim:
FWIW, I have submitted a PR to docker with a proposed fix for this on their end: https://github.com/docker/docker-py/pull/2467

jfuss:
@setrofim That is awesome! I think the part we are missing is timestamp information, so we can either tell docker-py to filter logs or do it ourselves. It looks like this should be possible, given the `attach()` method is a wrapper around [`logs()`](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.logs). So maybe the right thing here is to actually update to `logs=True` now and just fix the timestamp related concern when we finish up the 'warm container' pr.

Sorry thinking out loud/as I type.

PR #1571, opened 2019-11-21
Move to using communicate with a timeout for integ tests

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-windows.yml
+++ b/appveyor-windows.yml
@@ -75,7 +75,7 @@ test_script:
     - "venv\\Scripts\\activate"
     - "docker system prune -a -f"
     - "pytest -vv tests/integration"
-    - "pytest -n 4 -vv tests/regression"
+    - "pytest -vv tests/regression"
 
 # Uncomment for RDP
 # on_finish:
--- a/appveyor-windows.yml
+++ b/appveyor-windows.yml
@@ -74,8 +74,8 @@ test_script:
     - "git --version"
     - "venv\\Scripts\\activate"
     - "docker system prune -a -f"
-    - "pytest -vv tests/integration"
-    - "pytest -vv tests/regression"
+    - "pytest -n 4 -vv tests/integration"
+    - "pytest -n 4 -vv tests/regression"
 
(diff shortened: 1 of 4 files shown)


appveyor-windows.yml
```diff
@@ -75,7 +75,7 @@ test_script:
     - "venv\\Scripts\\activate"
     - "docker system prune -a -f"
     - "pytest -vv tests/integration"
-    - "pytest -vv tests/regression"
+    - "pytest -n 4 -vv tests/regression"
```
sanathkr: YES!!!!1

appveyor.yml
```diff
@@ -128,11 +118,11 @@ for:
       - "pylint --rcfile .pylintrc samcli"
       # There are some functional tests that are currently broken due to not being updated with changed code or still running with node4.3 runtimes
       # We need to update those but this allows us to at least runs the ones we currently have working
-      - "pytest tests/functional/commands/validate tests/functional/commands/cli/test_global_config.py"
+      - "pytest -n 4 tests/functional/commands/validate tests/functional/commands/cli/test_global_config.py"
 
       # Runs only in Linux
       - sh: "pytest -vv tests/integration"
```
sanathkr: in the future, we should parallelize this too

jfuss: Agreed. I tried to do that in this PR but required some additional changes in the build tests. We need to fix this part of the tests: https://github.com/awslabs/aws-sam-cli/blob/develop/tests/integration/buildcmd/test_build_cmd.py#L72 because we validate the image was deleted. I think we did this to ensure new tests would all start with downloading images and from the same place. While I agree, this actually added mins to the test runs as well. I was going to try and run all other integ tests but build but wanted to try and keep things scoped and not solve everything at once here.

jfuss: Correction: That method verifies we are deleting the container not the image.

PR #1581, opened 2019-11-22
fix: add version to `samconfig.toml` file

TheSriram:
- support version key, any float is okay.
- if a config file is present and the version key is missing, we do not
  process it.
- if a config file is missing, thats fine. this check does not get in
  the way.
- validation logic to determine if a SAM CLI version is compatible can
  be written later.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -10,6 +10,9 @@ import functools
 import logging
 
 import click
+
+from samcli.commands.exceptions import ConfigException
+from samcli.lib.config.exceptions import SamConfigVersionException
 from samcli.cli.context import get_cmd_names
 from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV
 
@@ -53,6 +56,7 @@ class TomlProvider:
 
             # NOTE(TheSriram): change from tomlkit table type to normal dictionary,
             # so that click defaults work out of the box.
+            samconfig.sanity_check()
             resolved_config = {k: v for k, v in samconfig.get_all(cmd_names, self.section, env=config_env).items()}
             LOG.debug("Configuration values read from the file: %s", resolved_config)
 
@@ -65,6 +69,11 @@ class TomlProvider:
                 self.section,
                 str(ex),
             )
+
+        except SamConfigVersionException as ex:
+            LOG.debug("%s %s", samconfig.path(), str(ex))
+            raise ConfigException(f"Syntax invalid in samconfig.toml: {str(ex)}")
+
         except Exception as ex:
             LOG.debug("Error reading configuration file: %s %s", samconfig.path(), str(ex))
 
--- a/samcli/lib/config/samconfig.py
+++ b/samcli/lib/config/samconfig.py
@@ -9,6 +9,9 @@ from pathlib import Path
 
 import tomlkit
 
+from samcli.lib.config.version import SAM_CONFIG_VERSION, VERSION_KEY
+from samcli.lib.config.exceptions import SamConfigVersionException
+
 LOG = logging.getLogger(__name__)
 
 DEFAULT_CONFIG_FILE_NAME = "samconfig.toml"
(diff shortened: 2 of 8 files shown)


samcli/lib/config/samconfig.py
```diff
@@ -148,24 +151,38 @@ def config_dir(template_file_path=None):
...
+            # NOTE(TheSriram): Do not attempt to re-write an existing version
+            pass
         self.filepath.write_text(tomlkit.dumps(self.document))
 
+    def _version(self):
+        return self.document.get(VERSION_KEY, None)
+
+    def _version_sanity_check(self, version):
+        if not isinstance(version, float):
+            raise SamConfigVersionException(f"'{VERSION_KEY}' key is not present " f"or is in unrecognized format. ")
```
awood45: This is well and good, but what is the purpose of a version number? The AWS CLI config has never needed this, what does it mean when we increment it?

TheSriram: This was discussed as part of design, but at the very least, it provides us a mechanism depending on the growth of scope of the configuration file in the future.

samcli/lib/config/samconfig.py
```diff
@@ -148,24 +151,38 @@ def config_dir(template_file_path=None):
...
+            self._version_sanity_check(self._version())
         return self.document
 
     def _write(self):
         if not self.document:
             return
         if not self.exists():
             open(self.filepath, "a+").close()
+        current_version = self._version() if self._version() else SAM_CONFIG_VERSION
+        try:
```
c2tarun: Just curious: any reason why we are not checking for key to be present and add only if it isn't?

samcli/lib/config/samconfig.py
```diff
@@ -148,24 +151,38 @@ def config_dir(template_file_path=None):
...
-        except OSError:
-            self.document = tomlkit.document()
-
+        if not self.document:
+            try:
+                txt = self.filepath.read_text()
+                self.document = tomlkit.loads(txt)
+                self._version_sanity_check(self._version())
+            except OSError:
+                self.document = tomlkit.document()
```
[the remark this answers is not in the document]
TheSriram: Sure.

samcli/cli/cli_config_file.py
```diff
@@ -63,6 +67,11 @@ def __call__(self, config_dir, config_env, cmd_names):
                 cmd_names,
                 self.section,
             )
+
+        except SamConfigVersionException as ex:
+            LOG.debug("%s %s", samconfig.path(), str(ex))
+            raise ConfigException(f"Syntax invalid in samconfig.toml: {str(ex)}")
```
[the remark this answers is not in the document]
TheSriram: One is a click exception, and the other isnt. str(ex) gives us version based information.

samcli/lib/config/samconfig.py
```diff
@@ -148,24 +151,38 @@ def config_dir(template_file_path=None):
...
 
     def _write(self):
         if not self.document:
             return
         if not self.exists():
             open(self.filepath, "a+").close()
+        current_version = self._version() if self._version() else SAM_CONFIG_VERSION
+        try:
+            self.document.add(VERSION_KEY, current_version)
+        except tomlkit.exceptions.KeyAlreadyPresent:
```
TheSriram: We check it here.

samcli/lib/config/samconfig.py
TheSriram: https://github.com/awslabs/aws-sam-cli/pull/1581/files#r349833716

TheSriram: we do.

samcli/lib/config/samconfig.py
```diff
@@ -101,8 +103,8 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
             If the data is invalid
         """
 
-        self._read()
         if not self.document:
+            self._read()
```
TheSriram: bugfix: do not attempt to read everytime a put is made, read once and load.

PR #1589, opened 2019-11-23
fix: move tests to a `static` bucket solution

TheSriram:
- s3 bucket is created is appropriate permissions and exposed via
  environment variable.
- serverlessrepo.amazonaws.com has access to the static s3 bucket.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/package/package_integ_base.py
```diff
@@ -12,18 +10,15 @@ class PackageIntegBase(TestCase):
...
         cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "package")
 
-        # Create S3 bucket
+        # Intialize S3 client
         s3 = boto3.resource("s3")
         # Use a pre-created KMS Key
         cls.kms_key = os.environ.get("AWS_KMS_KEY")
+        # Use a pre-created S3 Bucket
         cls.s3_bucket = s3.Bucket(cls.bucket_name)
-        cls.s3_bucket.create()
```
jfuss: Do we want to still do this as a fallback if `AWS_S3` is not set? Thinking it is good for local where devs do not need to have additional things preconfigured

TheSriram: OK, valid concern.

PR #1595, opened 2019-11-25
feat: Initialize a project from non-cookiecutter github repo or zip file

sanathkr:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/osutils.py
+++ b/samcli/lib/utils/osutils.py
@@ -6,8 +6,13 @@ import sys
 import os
 import shutil
 import tempfile
+import logging
+import contextlib
 
 from contextlib import contextmanager
+
+
+LOG = logging.getLogger(__name__)
 
 
 @contextmanager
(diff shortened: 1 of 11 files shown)


samcli/lib/utils/temp_file_utils.py
```diff
@@ -2,25 +2,4 @@
 Helper functions for temporary files
 """
 import os
```
sanathkr: I moved these to `osutils.py` to consolidate all OS related helpers. This is not required for this PR, but I thought it might be a good cleanup to do while I am here

tests/integration/init/test_init_command.py
```diff
@@ -33,13 +37,13 @@ def test_init_command_passes_and_dir_created(self):
                 raise
 
             self.assertEqual(process.returncode, 0)
-            self.assertTrue(os.path.isdir(temp + "/sam-app"))
+            self.assertTrue(Path(temp, "sam-app").is_dir())
```
sanathkr: Refactored to remove hardcoded `/` while I am here

samcli/lib/utils/osutils.py
```diff
@@ -62,3 +67,71 @@ def stderr():
...
+        ``ignore`` property of ``shutils.copytree`` method
+    """
+
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+
+        try:
+            # Let's try to copy the directory metadata from source to destination
+            shutil.copystat(source, destination)
+        except OSError as ex:
```
TheSriram: this seems broad to be just windows errors, can we get more granular?

samcli/lib/utils/osutils.py
```diff
@@ -62,3 +67,71 @@ def stderr():
...
+        ignored_names = ignore(source, names)
+    else:
+        ignored_names = set()
+
+    for name in names:
+        # Skip ignored names
+        if name in ignored_names:
+            continue
+
+        new_source = os.path.join(source, name)
```
TheSriram: Nit: I see some parts of the code changing to use pathlib, do we need to do the same here?

tests/integration/init/test_init_command.py
```diff
@@ -147,12 +151,48 @@ def test_init_command_with_extra_context_parameter(self):
...
+        zip_path_no_extension = str(Path(self.tempdir, "myfile"))
+
+        self.zip_path = shutil.make_archive(zip_path_no_extension, "zip", root_dir=self.tempdir, base_dir="zipdata")
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+
+    @parameterized.expand([(None,), ("project_name",)])
+    def test_arbitrary_project(self, project_name):
+        with tempfile.TemporaryDirectory() as temp:
```
TheSriram: are there any gotchas with temporary directories on windows?

samcli/local/init/arbitrary_project.py
```diff
@@ -0,0 +1,94 @@
...
+    if repository.is_zip_file(location):
+        LOG.debug("%s location is a zip file", location)
+        download_fn = functools.partial(
+            repository.unzip, zip_uri=location, is_url=repository.is_repo_url(location), no_input=no_input
+        )
+
+    # Else, treat it as a git/hg/ssh URL and try to clone
+    elif repository.is_repo_url(location):
+        LOG.debug("%s location is a source control repository", location)
+        download_fn = functools.partial(repository.clone, repo_url=location, no_input=no_input)
```
c2tarun: Just curious, if cookiecutter supports `zip` and `git` both, then why do we need two if conditions?

samcli/lib/utils/osutils.py
sanathkr: eventually. I don't want to change much now

tests/integration/init/test_init_command.py
sanathkr: it didn't break before ;) so no

samcli/lib/utils/osutils.py
sanathkr: this is not even my code. it's copy pasted ;)

samcli/lib/utils/osutils.py
```diff
@@ -62,3 +67,71 @@ def stderr():
...
     """
     return sys.stderr.buffer
+
+
+def remove(path):
+    if path:
+        try:
+            os.remove(path)
+        except OSError:
+            pass
```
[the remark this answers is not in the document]
sanathkr: it's only used in a test. I think we should clean this up later.

samcli/local/init/arbitrary_project.py
sanathkr: because cookiecutter's internal code verifies the presence of cookiecutter.json. we don't want that

samcli/lib/utils/osutils.py
```diff
@@ -62,3 +67,71 @@ def stderr():
...
+    try:
+        yield _tempfile
+    finally:
+        _tempfile.close()
+        remove(_tempfile.name)
+
+
+# NOTE: Py3.8 or higher has a ``dir_exist_ok=True`` parameter to provide this functionality.
+#       This method can be removed if we stop supporting Py37
+def copytree(source, destination, ignore=None):
```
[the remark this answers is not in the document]
sanathkr: because this was copied from lambda-builders :)

PR #1599, opened 2019-11-26
Ignore .git when init from arbitrary github repo

sanathkr:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/init/arbitrary_project.py
+++ b/samcli/local/init/arbitrary_project.py
@@ -3,8 +3,10 @@ Initialize an arbitrary project
 """
 
 import functools
+import shutil
 import logging
 
+from pathlib import Path
 from cookiecutter import repository
 from cookiecutter import config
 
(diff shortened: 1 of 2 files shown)


sanathkr:
Yeah, we can maybe add integ tests. But I don't think it's necessary because we are relying on stdlib functions

samcli/local/init/arbitrary_project.py
```diff
@@ -89,6 +91,6 @@ def _download_and_copy(download_fn, output_dir):
 
     with osutils.mkdir_temp() as tempdir:
         downloaded_dir = download_fn(clone_to_dir=tempdir)
-        osutils.copytree(downloaded_dir, output_dir)
+        osutils.copytree(downloaded_dir, output_dir, ignore=shutil.ignore_patterns("*.git"))
```
jfuss: Are we sure this works as we expect? This looks like the issue the Ruby builder had: https://github.com/awslabs/aws-lambda-builders/pull/130

Reading docs, this should work but maybe I just misunderstand the ruby builder fix.

sanathkr: yeah we use this the same way lambda builders does. It supports glob patterns so we don't copy .git from anywhere, including submodules.

PR #1620, opened 2019-11-29
Fix for stackTrace not returned in new Lambda containers.

c2tarun:
*Issue 1458:*

*Description of changes:* During `sam local start-api` we expect container to return response. We then parse this response with some caveat explained [here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/services/base_local_service.py#L151-L155) and expect a particular set of keys. Due to recent changes in lambda one of the key was removed from the response. The key removed was `stackTrace`, we rely on this key to add `FunctionError` in our response to client. Since the key was removed we need to handle the scenario so that we don't loose the information of function error.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -231,7 +231,17 @@ def do_cli(
 
         read_config_showcase(template_file=template_file)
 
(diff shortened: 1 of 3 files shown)


sanathkr:
Can you give a bit more context on why this change is right? I can't understand

c2tarun:
> Can you give a bit more context on why this change is right? I can't understand

Hi @sanathkr, I updated PR description as well as issue number.

PR #1639, opened 2019-12-04
Support/ruby inprocess builds

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
@TheSriram 

> Curious that doing away with a virtualenv makes this work.

Yes. Installing SAM CLI into a virtualenv (or venv) causes Ruby builds to fail. From output, it seemed Ruby (Gem or Bundler) thought that a man-in-the-middle attack was happening. Looking at what we do in `aws-lambda-builders` on appveyor, I noticed that we do not use venv and instead install it into the system's site-packages. I still don't fully understand what the deeper root cause is here but removing venv works. For now, this is ok since it is isolated by virtue of it being a build system but I don't think it is ideal. This at least unblocks us and allows us to run all tests every time. :)

PR #1634, opened 2019-12-04
feat(debug): Java11 debug support

jfuss:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1246, opened 2019-06-28
NODE_PATH is not set in the Container when in debug mode


guijob:
### Description

I was using sam 0.13 and I could debug my function using VS Code with no problems, after upgrading my lambdas do nodejs10.x (coming from 8.10) I had to update my sam version due compatbility, and now I no longer can debug my code, either in Ubuntu or macOS environment. I'm getting following error:
```
{
  "errorType": "Runtime.ImportModuleError",
  "errorMessage": "Error: Cannot find module 'aws-sdk'"
}
```

Does this package shouldn't be inside container? 

Additional point:
- It runs normally without debbuging mode (without -d 5858).
- Steps provided used to work using SAM 0.13 with node8.10 Runtime.

EDIT: I've just tried using SAM 0.17.0 with node8.10 Runtime and it worked. But ,using node10.x it doesn't.

### Steps to reproduce

project structure
```
node_modules/
template.yml
webpack.config.js
src/
-- lambdas/
---- records/
------ index.js
```
template.yml
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: ["AWS::Serverless-2016-10-31", "CorsFixer"]
Description: >
  auditing-back-app
    
Globals:
  Function:
    Runtime: nodejs10.x
    Timeout: 60
    MemorySize: 128
  Api:
    Cors:
      AllowOrigin: "'*'" 
      AllowHeaders: "'Content-Type,Authorization'" 

Parameters:
  Environment:
    Type: String

  RecordsEndpointsFunction:
    Type: AWS::Serverless::Function
    Properties:
      Description: !Sub ${Environment}
      CodeUri: dist/records/records.zip
      Handler: index.handler
      MemorySize: 1536
      Timeout: 500
      Policies:
        - AWSLambdaBasicExecutionRole
        - AmazonDynamoDBFullAccess
        - AmazonS3FullAccess
        - AmazonSESFullAccess
        - AmazonSQSFullAccess
      Events:
        UserEndpoints:
          Type: Api
          Properties:
            Path: /users/{userId}/parkinglots/{parkingLotId}/records
            Method: any
            RestApiId: !Ref HttpApiGateway
        NewFileUploaded:
          Type: SQS
          Properties:
            Queue: !Sub arn:aws:sqs:${AWS::Region}:{hidden}:auditing-new-file-to-process-${Environment}
            BatchSize: 1
        RealTimeUpdater:
          Type: Schedule
          Properties:
[...]

PR #1646, opened 2019-12-06
fix(debug): Add NODE_PATH env var when debugging Node10/12

jfuss:
To support debugging for Nodejs10.x and Nodejs12.x, SAM CLI overrides
the entrypoint to the container. In newer runtimes (that are based
on the provided runtime), we overrided the bootstrap file which can
have side-effects. In the case of Nodejs, the bootstrap file adds
the NODE_PATH env var. Since we override this, NODE_PATH does not
get setup on the system and causes modules to not load (in some
cases, #1246). With #1634 (Java11 Debug support), we added a
capability to attach Env Vars into the container for debug mode
invokes. We will use this to apply the NODE_PATH onto the container
for debug.

I attempted to move all the args we pass into NODE_OPTIONS, which would
make Nodejs work like Java11 for debug. That is, allow us not to override
the entrypoint and still be able to apply the needed flags to the interpreted.
Unfortunately, NODE_OPTIONS restricts values that can be passed in,
which includes `--nolazy`. This flag is used to tell Nodejs to parse
all JavaScript files first instead of on first access. This ensures that
breakpoints to not "jump" to a different location in the file. It could
be worth exploring whether `--nolazy` is strictly required but for now
leaving as is.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_debug_settings.py
+++ b/samcli/local/docker/lambda_debug_settings.py
@@ -157,7 +157,9 @@ class LambdaDebugSettings:
                         ]
                     ),
                 ],
-                debug_env_vars={},
+                debug_env_vars={
+                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node10/node_modules:/var/runtime/node_modules"
+                },
             ),
             Runtime.nodejs12x.value: DebugSettings(
                 [
@@ -177,7 +179,9 @@ class LambdaDebugSettings:
                         ]
                     ),
                 ],
-                debug_env_vars={},
+                debug_env_vars={
+                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node12/node_modules:/var/runtime/node_modules"
+                },
             ),
             Runtime.python27.value: DebugSettings(
                 ["/usr/bin/python2.7"] + debug_args_list + ["/var/runtime/awslambda/bootstrap.py"], debug_env_vars={}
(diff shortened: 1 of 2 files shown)


tests/unit/lib/bootstrap/test_bootstrap.py
```diff
@@ -13,7 +13,7 @@
 
 class TestBootstrapManagedStack(TestCase):
     def _stubbed_cf_client(self):
-        cf = botocore.session.get_session().create_client("cloudformation")
+        cf = botocore.session.get_session().create_client("cloudformation", region_name="us-west-2")
```
sanathkr: why hardcode?

samcli/local/docker/lambda_debug_settings.py
```diff
@@ -157,7 +157,9 @@ def get_debug_settings(debug_port, debug_args_list, runtime, options):
                         ]
                     ),
                 ],
-                debug_env_vars={},
+                debug_env_vars={
+                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node10/node_modules:/var/runtime/node_modules"
```
sanathkr: Can you add a comment explaining what this is and how it works? I can't understand

tests/unit/lib/bootstrap/test_bootstrap.py
jfuss: This is a unit tests and the tests will fail unless you have defined a region or have it on the system as an env var.

PR #1647, opened 2019-12-06
fix(ci): force chocolatey-core.extension 1.3.3

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
Ruby and Python installed through choco needs 1.3.3 or choco installs fail.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-windows-build-python.yml
+++ b/appveyor-windows-build-python.yml
@@ -38,6 +38,7 @@ install:
 
     # Make sure the temp directory exists for Python to use.
     # Install python3.8
+    - "choco install chocolatey-core.extension --version 1.3.3 --force -y"
     - "choco install python3 --version 3.8.0"
     - "C:\\Python38\\python.exe -m pip freeze"
     - "refreshenv"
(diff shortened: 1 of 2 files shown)


appveyor-windows-build-python.yml
```diff
@@ -38,6 +38,7 @@ install:
 
     # Make sure the temp directory exists for Python to use.
     # Install python3.8
+    - "choco install chocolatey-core.extension --version 1.3.3 --force -y"
```
c2tarun: Why do we have to pass a version at all? Can't we always install the latest version? Or it don't work with later versions?

jfuss: https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-q551u/builds/29367692

From what we are installing we need 1.3.3, otherwise they fail. This has to do with what the package (or whatever choco calls them). Something got updated underneath us (not entirely sure what), so this is to get us back to a good state

Issue #1644, opened 2019-12-06
SAM Deploy fails when using InlineCode


keetonian:
### Description

`sam deploy` doesn't work when using InlineCode. I use `sam deploy` with InlineCode to quickly test items and avoid any extra time spent on package/deploy steps.

### Steps to reproduce

Template:

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Resources:
  MyFunction2:
    Type: AWS::Serverless::Function
    Properties:
      InlineCode: |
        def handler(event, context):
            print(event)
            return {'body': 'Hello World 2!', 'statusCode': 200}
      Handler: index.handler
      Runtime: python3.7
      Events:
        Api:
          Type: Api
          Properties:
            Path: /
            Method: get
```

Command: 

```
sam deploy --template-file template.yaml --stack-name testing-inline-api --capabilities CAPABILITY_IAM
```
### Observed result

```
Unable to export
Traceback (most recent call last):
  File "/usr/local/Cellar/aws-sam-cli/0.35.0/libexec/lib/python3.7/site-packages/samcli/lib/package/artifact_exporter.py", line 255, in export
    self.do_export(resource_id, resource_dict, parent_dir)
  File "/usr/local/Cellar/aws-sam-cli/0.35.0/libexec/lib/python3.7/site-packages/samcli/lib/package/artifact_exporter.py", line 271, in do_export
    uploaded_url = upload_local_artifacts(resource_id, resource_dict, self.PROPERTY_NAME, parent_dir, self.uploader)
  File "/usr/local/Cellar/aws-sam-cli/0.35.0/libexec/lib/python3.7/site-packages/samcli/lib/package/artifact_exporter.py", line 151, in upload_local_artifacts
    return zip_and_upload(local_path, uploader)
  File "/usr/local/Cellar/aws-sam-cli/0.35.0/libexec/lib/python3.7/site-packages/samcli/lib/package/artifact_exporter.py", line 162, in zip_and_upload
    return uploader.upload_with_dedup(zip_file)
  File "/usr/local/Cellar/aws-sam-cli/0.35.0/libexec/lib/python3.7/site-packages/samcli/lib/package/s3_uploader.py", line 127, in upload_with_dedup
    return self.upload(file_name, remote_path)
[...]

PR #1661, opened 2019-12-11
Fix InlineCode Packaging Bug

awood45:
This also improves stack cleanup in the deploy integ tests. As of this change, all unit tests and deploy integ tests pass. Still want to add another integ test to validate that the short-circuit check for inline code isn't over-broad.

*Issue #, if available:* #1644


samcli/lib/package/artifact_exporter.py
```diff
@@ -156,6 +159,11 @@ def upload_local_artifacts(resource_id, resource_dict, property_name, parent_dir
 
     raise exceptions.InvalidLocalPathError(resource_id=resource_id, property_name=property_name, local_path=local_path)
 
+def resource_not_packageable(resource_dict):
+    inline_code = jmespath.search("InlineCode", resource_dict)
+    if inline_code is not None:
```
c2tarun: can we simply do `return inline_code is not None`?

awood45: We could at this point, wrote it this way with the idea that more such checks would follow. Also added a `make black` refactor just now.

tests/integration/deploy/test_deploy_command.py
```diff
@@ -283,6 +283,7 @@ def test_deploy_with_s3_bucket_switch_region(self, template_file):
         template_path = self.test_data_path.joinpath(template_file)
 
         stack_name = "a" + str(uuid.uuid4()).replace("-", "")[:10]
+        self.stack_names.append(stack_name)
```
jfuss: Nice catch. Not needed now, but we should adjust this in the future to where we only delete the stack if the tests pass. Would help us have something inspect-able after the fact, instead of just errors in the output.

tests/integration/deploy/test_deploy_command.py
```diff
@@ -436,6 +439,23 @@ def test_deploy_twice_with_fail_on_empty_changeset(self, template_file):
...
+
+        deploy_command_list = self.get_deploy_command_list(template_file=template_path, stack_name=stack_name, capabilities="CAPABILITY_IAM")
+        deploy_process_execute = Popen(deploy_command_list, stdout=PIPE, stderr=PIPE, stdin=PIPE)
+        try:
+            _, stderr = deploy_process_execute.communicate(timeout=TIMEOUT)
+        except TimeoutExpired:
+            deploy_process_execute.kill()
+            raise
+        stderr = stderr.strip()
+        print(stderr)
```
jfuss: remove print

PR #1664, opened 2019-12-13
Revert "Set the CORS "Access-Control-Allow-Credentials" for local running"

sanathkr:
Reverts awslabs/aws-sam-cli#1648


CGreenburg:
@sanathkr @jfuss I just noticed the merge request I had made is being reverted. Is there an issue either of you can link for why?

jfuss:
@CGreenburg I don't specifically recall but probably something to do with not enough testing around it (guessing by looking at the pr).

PR #1671, opened 2019-12-16
Removes API Event from Inline Test

awood45:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/testdata/package/aws-serverless-inline.yaml
+++ b/tests/integration/testdata/package/aws-serverless-inline.yaml
@@ -12,9 +12,3 @@ Resources:
           return {'body': 'Hello World 2!', 'statusCode': 200}
       Handler: index.handler
       Runtime: python3.7
-      Events:
-        Api:
-          Type: Api
-          Properties:
-            Path: /
-            Method: get


tests/integration/testdata/package/aws-serverless-inline.yaml
```diff
@@ -12,9 +12,3 @@ Resources:
           return {'body': 'Hello World 2!', 'statusCode': 200}
       Handler: index.handler
       Runtime: python3.7
-      Events:
```
c2tarun: Without API gateway how are we validating if Lambda is working? Are we invoking it directly?

tests/integration/testdata/package/aws-serverless-inline.yaml
```diff
@@ -12,9 +12,3 @@ Resources:
...
-        Api:
```
c2tarun: Also why are we removing it?

tests/integration/testdata/package/aws-serverless-inline.yaml
TheSriram: That brings up a good point, what we check for here is that we can deploy a particular template and check for return codes, but we dont attempt to check if the components that are deployed actually work. This may be a new set of end to end tests we may need to adopt.

jfuss: while it could be valuable to test the function, it’s not the scope of these tests. We want to make sure the deploy can happen successfully. I can see value in verifying configuration of what was deployed, but that kind of a grey area to me and not really testing lambda either.

tests/integration/testdata/package/aws-serverless-inline.yaml
jfuss: Not need, and would require more permissions for our integration tests to work.

PR #1674, opened 2019-12-17
refactor: Move providers to lib folder

sanathkr:
* Move `samcli/commands/local/lib/*providers.py` to `samcli/lib/providers/`
* Update imports
* Update patch annotations
* Move `InvalidLayerReference` to an exception.py file within the destination

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -143,8 +146,9 @@ def functions_to_build(self):
 
             if not function:
                 all_functions = [f.name for f in self._function_provider.get_all()]
-                available_function_message = "{} not found. Possible options in your template: {}" \
-                    .format(self._function_identifier, all_functions)
+                available_function_message = "{} not found. Possible options in your template: {}".format(
```
jfuss: Super nit (just noticing as I look through): this could be a f-string to improve readability and speed.

sanathkr: Agreed. I am doing several large scale code moves. I don't want to change these now.

jfuss: We should track these changes some how. We will encounter them as we write new code, but probably won't get to update the code base completely (which is fine). These kind of improvements will go a long way if we can standardize on Py3 things.

PR #1678, opened 2019-12-17
Change How App Templates Are Updated for `sam init`

awood45:
This was initially triggered by a need to support silent updates for the IDEs using `--no-interactive` as an option for init. To properly handle that, while still handling cases where a user might have issues copying the new init templates down for some reason (internet connections, possibly file permissions), this PR now changes our general approach for copying app templates.

Now, we use a sort of state machine logic - we use a temp directory to stage the `git clone` of the app templates repo, deleting and copying over the new repo as appropriate to the AWS SAM shared directory. If the clone fails, no change is made and the existing app templates copy is preserved.


awood45:
Hold on this PR, going for a deeper approach.

samcli/commands/init/init_templates.py
```diff
@@ -118,23 +120,72 @@ def _shared_dir_check(self, shared_dir):
...
+                expected_temp_path = os.path.normpath(os.path.join(tempdir, self._repo_name))
+                LOG.info("\nCloning app templates from %s", self._repo_url)
                 subprocess.check_output(
-                    [self._git_executable(), "clone", self._repo_url], cwd=shared_dir, stderr=subprocess.STDOUT
+                    [self._git_executable(), "clone", self._repo_url, self._repo_name],
+                    cwd=tempdir,
+                    stderr=subprocess.STDOUT,
                 )
+                # Now we need to delete the old repo and move this one.
+                try:
```
TheSriram: Suggestion: can we move this try except block which is inside another try except block to another method, helps with readability.

samcli/commands/init/init_templates.py
```diff
@@ -118,23 +120,72 @@ def _shared_dir_check(self, shared_dir):
...
+                self.repo_path = expected_path
+            except OSError as ex:
+                LOG.warning("WARN: Could not clone app template repo.", exc_info=ex)
+            except subprocess.CalledProcessError as clone_error:
+                output = clone_error.output.decode("utf-8")
+                if "not found" in output.lower():
+                    click.echo("WARN: Could not clone app template repo.")
+
+    def _clone_new_app_templates(self, shared_dir, expected_path):
+        with osutils.mkdir_temp(ignore_errors=True) as tempdir:
```
TheSriram: if we are using ignore_errors, i also suggest we add a callback for `onerror` to log a debug message.

samcli/commands/init/init_templates.py
```diff
@@ -118,23 +120,72 @@ def _shared_dir_check(self, shared_dir):
...
             except OSError as ex:
                 LOG.warning("WARN: Can't clone app repo, git executable not found", exc_info=ex)
             except subprocess.CalledProcessError as clone_error:
                 output = clone_error.output.decode("utf-8")
                 if "not found" in output.lower():
                     click.echo("WARN: Could not clone app template repo.")
-        self.clone_attempted = True
+
+    def _template_directory_exists(self, expected_path):
+        path = Path(expected_path)
```
TheSriram: nit/non-blocking: I notice we are using both pathlib and os.path, is it possible to standardize on one or other.

samcli/lib/utils/osutils.py
```diff
@@ -16,7 +16,7 @@
 
 
 @contextmanager
-def mkdir_temp(mode=0o755):
+def mkdir_temp(mode=0o755, ignore_errors=False):
```
TheSriram: 👍

samcli/commands/init/init_templates.py
awood45: Done in new version.

samcli/commands/init/init_templates.py
awood45: Done in latest change

PR #1681, opened 2019-12-18
refactor: `sam deploy`

sriram-mv:
### refactor: `sam deploy`

Why is this change necessary?

* This change is necessary to address overall tech debt introduced by lumping all functions into one `command.py` file which handled both interactive and non-interactive modes when  `sam deploy` was used.
* Fix for https://github.com/awslabs/aws-sam-cli/issues/1652
* Previously user agent used to be passed via `AWS_EXECUTION_ENV`, Moving away from AWS CLI regressed that functionality.

* Interactive prompts and their lifecycles during the guided deploy phase are moved to its own class, and the results of the prompts are then used to start the package-deploy process with all the parameters that were gathered. This removes tight coupling and allows for guided prompts to be changed independently.
* Dealing with configuration files (also its own class) are also now masked within the interactive prompts because the normal deploy process does not need to write to the configuration file. This configuration file manager can also deal with other sections other than just parameters in the future.
* Printing of deploy arguments have been moved its own utils file to showcase separation of concerns as this happens irrespective of the deploy mode.
* User agent is passed to `boto session` based on the SAM CLI version and installation id, depending on telemetry being enabled.

* The refactors do not introduce any side effects.
* All parameter values will be quoted when saving to configuration file.
* Behavior change on submitting user agent information, which differs from the original method of just sending the installation id and not the SAM CLI version by default.

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/utils.py
+++ b/samcli/commands/deploy/utils.py
@@ -1,0 +1,59 @@
+"""
+Utilities for sam deploy command
+"""
+
+import json
+import click
+
+
+def print_deploy_args(stack_name, s3_bucket, region, capabilities, parameter_overrides, confirm_changeset):
+    """
+    Print a table of the values that are used during a sam deploy
+
+    Example below:
+
+        Deploying with following values
+        ===============================
+        Stack name                 : sam-app
+        Region                     : us-east-1
+        Confirm changeset          : False
+        Deployment s3 bucket       : aws-sam-cli-managed-default-samclisourcebucket-abcdef
+        Capabilities               : ["CAPABILITY_IAM"]
+        Parameter overrides        : {'MyParamater': '***', 'Parameter2': 'dd'}
+
+    :param stack_name: Name of the stack used during sam deploy
+    :param s3_bucket: Name of s3 bucket used for packaging code artifacts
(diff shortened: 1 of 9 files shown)


samcli/commands/deploy/utils.py
```diff
@@ -0,0 +1,59 @@
+"""
+Utilities for sam deploy command
```
sanathkr: ah another utilities!

samcli/commands/package/package_context.py
```diff
@@ -79,9 +82,17 @@ def __exit__(self, *args):
...
 
+        gc = GlobalConfig()
         session = boto3.Session(profile_name=self.profile if self.profile else None)
         s3_client = session.client(
-            "s3", config=Config(signature_version="s3v4", region_name=self.region if self.region else None)
+            "s3",
+            config=Config(
+                signature_version="s3v4",
+                region_name=self.region if self.region else None,
+                user_agent_extra=f"aws-sam-cli/{__version__}/{gc.installation_id}"
```
sanathkr: can we move this to utils? don't want to duplicate

TheSriram: Good call out.

PR #1686, opened 2019-12-19
feat: Update handling of UserExceptions that are raised

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
Updates how we handle UserExceptions so that when telemetry is on, we can see the more specific errors than just "UserException".

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/exceptions.py
+++ b/samcli/commands/exceptions.py
@@ -19,6 +19,11 @@ class UserException(click.ClickException):
 
     exit_code = 1
 
+    def __init__(self, message, wrapped_from=None):
+        self.wrapped_from = wrapped_from
+
+        click.ClickException.__init__(self, message)
+
 
 class CredentialsError(UserException):
     """
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -147,7 +147,7 @@ You can run 'sam init' without any options for an interactive initialization flo
     --name and --runtime and --app-template and --dependency-manager
     --location
         """
-        raise UserException(msg)
+        raise click.UsageError(msg)
     # check for required parameters
     if location or (name and runtime and dependency_manager and app_template):
         # need to turn app_template into a location before we generate
@@ -170,7 +170,7 @@ Must provide one of the following required parameter combinations:
 
 You can also re-run without the --no-interactive flag to be prompted for required values.
         """
-        raise UserException(error_msg)
+        raise click.UsageError(error_msg)
     else:
         # proceed to interactive state machine, which will call do_generate
         do_interactive(location, runtime, dependency_manager, output_dir, name, app_template, no_input)
@@ -190,7 +190,7 @@ def _get_cookiecutter_template_context(name, runtime, extra_context):
         try:
             extra_context_dict = json.loads(extra_context)
         except JSONDecodeError:
-            raise UserException(
+            raise click.UsageError(
                 "Parse error reading the --extra-context parameter. The value of this parameter must be valid JSON."
             )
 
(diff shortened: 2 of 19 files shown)


samcli/commands/init/__init__.py
```diff
@@ -190,7 +190,7 @@ def _get_cookiecutter_template_context(name, runtime, extra_context):
         try:
             extra_context_dict = json.loads(extra_context)
         except JSONDecodeError:
-            raise UserException(
+            raise click.UsageError(
```
c2tarun: How is raising click UsageError different from UserException? I mean what'll be difference on user side due to this change?

samcli/commands/init/init_templates.py
```diff
@@ -59,7 +63,7 @@ def prompt_for_location(self, runtime, dependency_manager):
             return (template_md["init_location"], "hello-world")
         if template_md.get("directory") is not None:
             return (os.path.join(self.repo_path, template_md["directory"]), template_md["appTemplate"])
-        raise UserException("Invalid template. This should not be possible, please raise an issue.")
+        raise InvalidInitTemplateError("Invalid template. This should not be possible, please raise an issue.")
```
c2tarun: For template issue we used name `TemplateNotFoundException`, but here we are calling it an `Error`. How are we differentiating between `Error` and `Exception` terms?

samcli/commands/exceptions.py
```diff
@@ -19,6 +19,11 @@ class UserException(click.ClickException):
 
     exit_code = 1
 
+    def __init__(self, message, wrapped_from=None):
+        self.wrapped_from = wrapped_from
```
TheSriram: I like the idea. Does this information need to be surfaced somewhere though?

samcli/lib/telemetry/metrics.py
```diff
@@ -65,7 +65,10 @@ def wrapped(*args, **kwargs):
             # Capture exception information and re-raise it later so we can first send metrics.
             exception = ex
             exit_code = ex.exit_code
-            exit_reason = type(ex).__name__
+            if ex.wrapped_from is None:
+                exit_reason = type(ex).__name__
+            else:
+                exit_reason = ex.wrapped_from
```
TheSriram: ah Nice!

tests/unit/commands/init/test_cli.py
```diff
@@ -230,7 +231,7 @@ def test_init_cli_int_from_location(self, generate_project_patch, sd_mock):
     def test_init_cli_missing_params_fails(self):
         # WHEN we call init without necessary parameters
         # THEN we should receive a UserException
-        with self.assertRaises(UserException):
+        with self.assertRaises(click.UsageError):
```
TheSriram: Something to note, usage errors are exit code 2. Would we be breaking anyone?

jfuss: This is only for init and the all the errors are based on either passing in the wrong data (not json) or invalid collection of parameters. Given it is only an exit code and for init, I am not concerned with this breaking anyone.

samcli/commands/init/__init__.py
jfuss: see https://github.com/awslabs/aws-sam-cli/pull/1686#discussion_r360044936

samcli/commands/init/init_templates.py
jfuss: An Error typically indicates serious problem that an application should not try to catch, while an Exception indicates conditions that an application might want to catch.

We can't really do anything about an Invalid template here so decided to go with Error instead of Exception.

samcli/commands/exceptions.py
jfuss: In telemetry (looks like you found it, base on your next comment :))

PR #1693, opened 2019-12-25
fix: botocore sessions across samcli

sriram-mv:
### fix: botocore sessions across samcli

Why is this change necessary?

* This change is necessary to address MFA usecases when caching of credentials is essential. There have been multiple reports in github issues: [1682](https://github.com/awslabs/aws-sam-cli/issues/1682), [1623](https://github.com/awslabs/aws-sam-cli/issues/1623)

* Use botocore's native json file cache manager.

* Sessions are managed at the cli context level and not anywhere else.

TODO?

* Need to check entire codebase for session management.

Depends on:

* https://github.com/awslabs/aws-sam-cli/pull/1681

### DIFF
--- a/designs/botocore_sessions.md
+++ b/designs/botocore_sessions.md
@@ -1,0 +1,117 @@
+CLI: Botocore Sessions
+====================================
+
+What is the problem?
+--------------------
+
+SAM CLI does not have first class support for MFA based authentication, which leads to multiple requests for such tokens degrading the overall CLI experience.
+
+What will be changed?
+---------------------
+* SAM CLI will support default sessions with caching of temporary credentials to a json cache.
+
+Success criteria for the change
+-------------------------------
+* SAM CLI only ever asks for a MFA token once and caches it under the default boto location. This allows for other application that uses default location for caching credentials.
+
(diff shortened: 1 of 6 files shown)


sriram-mv:
will rebase once https://github.com/awslabs/aws-sam-cli/pull/1681 is merged.

designs/botocore_sessions.md
```diff
@@ -0,0 +1,117 @@
+CLI: Botocore Sessions
+====================================
+
+What is the problem?
+--------------------
+
+SAM CLI does not have first class support for MFA based authentication, which leads to multiple requests for such tokens degrading the overall CLI experience.
```
sanathkr: Thank you for adding a design doc!

samcli/cli/context.py
```diff
@@ -144,6 +147,11 @@ def _refresh_session(self):
         """
         try:
             boto3.setup_default_session(region_name=self._aws_region, profile_name=self._aws_profile)
+            # get botocore session and setup caching for MFA based
```
sriram-mv: Code changes

PR #1695, opened 2019-12-26
Build and Package once per CodeUri instead of once per resource

beck3905:
*Description of changes:*

- Update `sam build` to cache the built directory for each unique CodeUri.  This will prevent unnecessary duplication of CodeUri build directories when multiple functions share the same CodeUri.
- Update `sam package` artifact exporter to use the folder checksum as the suffix of the package name rather than a UUID.  This will prevent changing the output template file when the application hasn't changed.  If the application or template has not changed then the output template file will contain the same updated paths as previous `sam package` outputs.
- Update `sam package` artifact exporter to cache the package generated for each property name/value combination.  This will prevent generating and uploading multiple copies of the same artifact to S3 when multiple serverless resources use the same path.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

This is a change to the internal process that happens in `sam build` and `sam package` in order to improve efficiency and to avoid redeploying applications that haven't changed.  I'm not sure what if any new documentation is required.  Please advise.

I have not been able to get integration tests running locally on my machine.  I'd be happy to work on those if I can get them running.  Can anyone help me with that?

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -83,6 +83,7 @@ class ApplicationBuilder:
         self._container_manager = container_manager
         self._parallel = parallel
         self._mode = mode
+        self._package_cache = {}
 
     def build(self):
         """
@@ -169,15 +170,19 @@ class ApplicationBuilder:
         str
             Path to the location where built artifacts are available
         """
+        # TODO: implement if code_dir already exists then reuse rather than rebuilding
 
         # Create the arguments to pass to the builder
         # Code is always relative to the given base directory.
         code_dir = str(pathlib.Path(self._base_dir, codeuri).resolve())
 
+        if code_dir in self._package_cache:
+            return self._package_cache[code_dir]
+
         config = get_workflow_config(runtime, code_dir, self._base_dir)
 
         # artifacts directory will be created by the builder
-        artifacts_dir = str(pathlib.Path(self._build_dir, function_name))
+        artifacts_dir = str(pathlib.Path(self._build_dir, pathlib.Path(code_dir).name))
 
         with osutils.mkdir_temp() as scratch_dir:
             manifest_path = self._manifest_path_override or os.path.join(code_dir, config.manifest_name)
@@ -187,12 +192,16 @@ class ApplicationBuilder:
             if self._container_manager:
                 build_method = self._build_function_on_container
 
-            return build_method(config,
+            artifacts = build_method(config,
                                 code_dir,
                                 artifacts_dir,
                                 scratch_dir,
                                 manifest_path,
                                 runtime)
+
+            self._package_cache[code_dir] = artifacts
+
+        return artifacts
 
     def _build_function_in_process(self,
                                    config,
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -170,7 +170,6 @@ class ApplicationBuilder:
         str
             Path to the location where built artifacts are available
         """
-        # TODO: implement if code_dir already exists then reuse rather than rebuilding
 
         # Create the arguments to pass to the builder
         # Code is always relative to the given base directory.
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -252,6 +252,7 @@ class Resource:
     # Set this property to True in base class if you want the exporter to zip
     # up the file before uploading This is useful for Lambda functions.
     FORCE_ZIP = False
+    _package_cache = {}
 
     def __init__(self, uploader):
         self.uploader = uploader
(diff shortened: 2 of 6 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -174,10 +175,13 @@ def _build_function(self, function_name, codeuri, runtime):
         # Code is always relative to the given base directory.
         code_dir = str(pathlib.Path(self._base_dir, codeuri).resolve())
 
+        if code_dir in self._package_cache:
+            return self._package_cache[code_dir]
```
sriram-mv: Ideally there shouldn't be a build neccessary if the contents have not changed. We would still need a way to conduct a force build

sriram-mv:
Thanks for the PR! Which integration tests are you having trouble with?

samcli/lib/build/app_builder.py
beck3905: Not necessarily true in all cases.  For example, I am building and deploying SAM apps in a CI/CD pipeline.  My repository contains the SAM app as well as unit tests and integration tests.  If I update my tests, it will force a whole build and deploy of the SAM app.  By using this hash, my CI/CD system will still generate the correct artifacts (packaged.yaml referencing the same package in S3) without actually updating the application.

These changes also save a lot of time when my application does change.  Currently, my application has several Lambda functions that all share the same CodeUri.  `sam build` ends up making a copy of my CodeUri folder for each function and `sam package` zips and uploads a copy of my CodeUri folder for each function, which takes more time and generates many duplicate artifacts in S3.

beck3905:
@sriram-mv I was having trouble with all of them.  Last night I finally had a breakthrough and was able to get some of them working.  It was not clear in the DEVELOPMENT_GUIDE.md that there are several prerequisities required to run integration tests.  For example, I am running on a Mac and I had to install JDK, Gradle, Maven, and dotnet in order to make some progress on the integration tests.  I'm still having trouble with some of the dotnet tests where it is telling me that it cannot find the `dotnet` CLI executable, but in my terminal I can run `dotnet` command and get usage instructions.

beck3905:
@sriram-mv I just ran the integration tests locally on the latest `develop` branch and had 7 failures.

- test_with_default_requirements_0_python2_7
- test_with_default_requirements_1_python3_6

both failed with:

```
  File "~/.pyenv/versions/3.7.5/lib/python3.7/logging/__init__.py", line 1025, in emit
    msg = self.format(record)
  File "~/.pyenv/versions/3.7.5/lib/python3.7/logging/__init__.py", line 869, in format
    return fmt.format(record)
  File "~/.pyenv/versions/3.7.5/lib/python3.7/logging/__init__.py", line 608, in format
    record.message = record.getMessage()
  File "~/.pyenv/versions/3.7.5/lib/python3.7/logging/__init__.py", line 369, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
```

on `LOG.info("Running Command: {}", cmdlist)`

- test_with_dotnetcore_0_dotnetcore2_0
- test_with_dotnetcore_1_dotnetcore2_1
- test_with_dotnetcore_2_dotnetcore2_0
- test_with_dotnetcore_3_dotnetcore2_1

all failed with:

```
Error: DotnetCliPackageBuilder:RunPackageAction - Dotnet CLI Failed: Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET Core program, but dotnet-lambda does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```

I have the dotnet-sdk installed and added `~/.dotnet/tools` to my PATH.  `amazon.lambda.tools` shows when I run `dotnet tools list --global`

- test_create_application_infer_region_from_env

failed with:

```
# Get console link application id from stdout
        pattern = r"arn:[\w\-]+:serverlessrepo:[\w\-]+:[0-9]+:applications\~[\S]+"
        match = re.search(pattern, process_stdout.decode("utf-8"))
        self.application_id = match.group().replace("~", "/")
>       self.assertIn(self.region_name, self.application_id)
E       TypeError: 'in <string>' requires string as left operand, not NoneType
publish/test_command_integ.py:170: TypeError
```

Now that most of the tests are running and passing locally on the `develop` branch.  I will try to run them on my branch with changes.  Hopefully I can address any issues there related to my changes.  The failures above don't seem to be related to anything my changes impact.

beck3905:
@sriram-mv Even though my develop branch ran the integration tests, my feature branch did not.  I ran tests in develop and rebased my feature branch onto develop, but still no luck.  All of the tests seem to be getting the same logging error that just those 2 python tests got in develop.

jfuss:
The build updates in the PR as similar to #1180. Please see my comment about building to a single artifact: https://github.com/awslabs/aws-sam-cli/pull/1180#issuecomment-614078419, on how we need to go forward to accept a PR on this feature. 

In addition, we have already released an update to `sam package` that hashes the directory contents instead of the zip so that we do not always upload code that has not changed.

Closing as we cannot accept the PR in the current state.

PR #1711, opened 2020-01-04
Reading Globals properties for Function.

c2tarun:
*Issue #1706*

This change ensures that we respect values declared in Globals for SAM resources.

This change reads properties from Globals and merges them with local properties of resources. Local property value will override global property value.

None

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/artifact_exporter.py
```diff
@@ -616,11 +614,17 @@ def export(self):
...
         self.template_dict = self.export_global_artifacts(self.template_dict)
+        global_function_properties = self.template_dict.get("Globals", {}).get("Function", {})
 
         for resource_id, resource in self.template_dict["Resources"].items():
 
             resource_type = resource.get("Type", None)
-            resource_dict = resource.get("Properties", None)
+            if global_function_properties:
+                # Merging global and local Function properties. Local will override global.
+                resource_dict = {**global_function_properties, **resource.get("Properties", {})}
```
sanathkr: Hang on. You don't need to do this here. Just run SAM Translator to merge Globals for you before package command starts. This is a more scalable approach because there are other parts of SAM template that you might need to "merge" before packaging.

This functionality should be accessible through SamFunctionProvider. Look at the InvokeContext code paths to get an idea.

jfuss: We might need to think deeper here. While I agree what @sanathkr is suggesting Is the right thing to do, SAM Translator requires additional permissions, which some depend on the resources in the template (SAR). Ideally, we should be able to do a dry run of the transform without needing permissions but that is way more work. So we should make sure adding additional permissions on package is something we want to support long term.

sanathkr: But how do we run local invoke without these permissions?

jfuss: Fair. Validate has those requirements. Ignore me and my memory :)

sriram-mv: The question to ask is, did package have those permissions before? If we are adding it now, does it mean the permissions required to package change, and hence breaking tighlty scoped roles used in CI/CD for sam package, sam deploy?

c2tarun: > Hang on. You don't need to do this here. Just run SAM Translator to merge Globals for you before package command starts. 

Updated code to use SamFunctionProvider before exporting Resources.

samcli/lib/package/artifact_exporter.py
```diff
@@ -532,6 +533,15 @@ def include_transform_export_handler(template_dict, uploader, parent_dir):
 GLOBAL_EXPORT_DICT = {"Fn::Transform": include_transform_export_handler}
 
 
+def normalized_sam_template(template_yaml):
+    """Takes sam yaml template as input and returns normalized template.
+
+    :param template_yaml: template need to be normalized.
+    :return:
+    """
+    return SamFunctionProvider(template_yaml).template_dict
```
sanathkr: Looking at SamFunctionProvider code, looks all you need is the BaseProvider so you don't rely on any unintended side-effects introduced on the template_dict variable by SamFunctionProvider.

```
        self.template_dict = SamBaseProvider.get_template(template_dict, parameter_overrides)
```

PR #1726, opened 2020-01-14
fix: Nested properties in RESOURCES_WITH_LOCAL_PATHS

wnkz:
*#1537*

Nested properties in `RESOURCES_WITH_LOCAL_PATHS` are not handled and tested correctly. Also, using dot notation is not really future-proof in case a real property contains a dot in its name.

This PR is an alternative to #1550 

Uses lists instead of dot notation.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/resources.py
+++ b/samcli/commands/_utils/resources.py
@@ -32,7 +32,7 @@ RESOURCES_WITH_LOCAL_PATHS = {
     AWS_SERVERLESS_APPLICATION: ["Location"],
     AWS_LAMBDA_LAYERVERSION: ["Content"],
     AWS_SERVERLESS_LAYERVERSION: ["ContentUri"],
-    AWS_GLUE_JOB: ["Command.ScriptLocation"],
+    AWS_GLUE_JOB: [["Command", "ScriptLocation"]],
 }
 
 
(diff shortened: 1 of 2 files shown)


wnkz:
I'm aware of failing tests but they are already ineffective for nested properties.

https://github.com/awslabs/aws-sam-cli/blob/a0680096b22ed307e06e0b13559b1cb8be58e9e5/tests/unit/commands/_utils/test_template.py#L122-L154

In this example we end up testing a template looking like 

```yaml
Resources:
  GlueJob:
    Type: AWS::Glue::Job
    Properties:
      Command.ScriptLocation: glue.py
```

instead of

```yaml
Resources:
  GlueJob:
    Type: AWS::Glue::Job
    Properties:
      Command:
        ScriptLocation: glue.py
```

I might need some help because the root cause of this issue is much deeper than I anticipated.

jfuss:
@wnkz Thanks for the PR. I made some additions to #1550 and that is now in an approved state (waiting on builds to finish). Going to close this one since this was attempt to solve the same problem. Let me know if there was something different about this one that we should consider adding as well.

PR #1727, opened 2020-01-15
Allow the characters _:./+- to be used in tag keys during deploy

RobReus:
Using the `aws cloudformation deploy --tags` command, you are allowed to use these characters in the tag keys. Documentation also lists these characters as supported and allowed. With `sam deploy --tags`, it did not work properly due to a regex limiting supported characters to A-Za-z0-9.

By adding a regex specifically for tag keys. By using this new regex, `sam deploy --tags 'owner:name=son of anton'` no longer converts the tag key to `name` instead of `owner:name`.
Before:
```
{"name": "son of anton"}
```
After:
```
{"owner:name": "son of anton"}
```

None. More characters allowed in keys and fully backwards compatible.

*Checklist:*

- [ ] ~~Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))~~
- [ ] ~~Write unit tests~~
- [ ] ~~Write/update functional tests~~
- [X] `make pr` passes
- [ ] ~~Write documentation~~

### DIFF
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -14,6 +14,7 @@ def _value_regex(delim):
 
 
 KEY_REGEX = '([A-Za-z0-9\\"]+)'
+TAG_KEY_REGEX = '([A-Za-z0-9\\"_:\\./\\+-]+)'
 # Use this regex when you have space as delimiter Ex: "KeyName1=string KeyName2=string"
 VALUE_REGEX_SPACE_DELIM = _value_regex(" ")
 # Use this regex when you have comma as delimiter Ex: "KeyName1=string,KeyName2=string"
@@ -168,7 +169,7 @@ class CfnTags(click.ParamType):
 
     _EXAMPLE = "KeyName1=string KeyName2=string"
 
-    _pattern = r"{key}={value}".format(key=KEY_REGEX, value=VALUE_REGEX_SPACE_DELIM)
+    _pattern = r"{key}={value}".format(key=TAG_KEY_REGEX, value=VALUE_REGEX_SPACE_DELIM)
 
     # NOTE(TheSriram): name needs to be added to click.ParamType requires it.
     name = ""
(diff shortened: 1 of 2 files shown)


vladb-metapack:
Hi, 

I was trying to tag my entire stack with some tags which include colons in them 
and using sam deploy to deploy them (`sam version 0.40 and awscli version 1.16.230`) 
and had the same issue as this. Everything before the `:` including the `:` itself was removed in the tags. 

I have upgraded the aws versions and downgraded sam to 0.15 which enabled me to use this, but alas for safety reasons we went ahead with just using `aws cloudformation deploy`.

RobReus:
> Hi,
> 
> I was trying to tag my entire stack with some tags which include colons in them
> and using sam deploy to deploy them (`sam version 0.40 and awscli version 1.16.230`)
> and had the same issue as this. Everything before the `:` including the `:` itself was removed in the tags.
> 
> I have upgraded the aws versions and downgraded sam to 0.15 which enabled me to use this, but alas for safety reasons we went ahead with just using `aws cloudformation deploy`.

This was the exact reason for me to create this PR. As per AWS [best practices](https://d1.awsstatic.com/whitepapers/aws-tagging-best-practices.pdf), we too use colons to separate between parts of the tag keys. I got tired of having to use `sam package` and then `aws cloudformation deploy`.

@sriram-mv I just realized I did not make the regex a bit nicer. I left some other allowed characters out (like &) and did not add the constraints of the key value (must start and end with A-Za-z0-9). Would you want me to fix this before merging this PR?

sriram-mv:
@RobReus Yes please, I think its a good idea to add some more constraints on this.

sriram-mv:
Closing because I was not able to push to the fork.

https://github.com/awslabs/aws-sam-cli/pull/1798 is the new PR.

PR #1731, opened 2020-01-16
Support SAM Build with Go Modules

awood45:
*Issue #, if available:*

Numerous

This change adds support for building Go Lambda applications to aws-sam-cli, integrating the already merged addition to aws-lambda-builders.

*Checklist:*
- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation (verifying now if any documentation needs to be updated explicitly)


tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -528,6 +528,97 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
+        cmdlist = self.get_command_list(use_container=False, parameter_overrides=overrides)
+
+        # Need to pass GOPATH ENV variable to match the test directory when running build
+
+        LOG.info("Running Command: {}".format(cmdlist))
+        LOG.info("Running with SAM_BUILD_MODE={}".format(mode))
+
+        newenv = os.environ.copy()
+        if mode:
+            newenv["SAM_BUILD_MODE"] = mode
```
jfuss: This will produce a binary but the binary is not debugable. Not sure if we want to add this test now or when we enable debug builds for go. Just wanted to call it out.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -528,6 +528,97 @@ def _verify_built_artifact(self, build_dir, function_logical_id, expected_files)
...
+
+        newenv["GOPROXY"] = "direct"
+        newenv["GOPATH"] = str(self.working_dir)
+
+        process = Popen(cmdlist, cwd=self.working_dir, env=newenv, stderr=PIPE)
+        try:
+            stdout, stderr = process.communicate(timeout=TIMEOUT)
```
jfuss: You do not use stdout or stderr here. You can remove them and remove the `stderr=PIPE` from the Popen call. 

Or you can leave it and log the output on failures, would lead to a better debug experience in ci maybe? If you go this route, you will need to add `stdout=PIPE` to the Popen call.

tests/integration/testdata/buildcmd/Go/.gitignore
```diff
@@ -0,0 +1 @@
+go.sum
```
jfuss: Should this be in the base .gitignore? Will git even pick this up for the repo?

tests/integration/buildcmd/test_build_cmd.py
awood45: This is leftover from some debugging I was doing, will clean it up.

tests/integration/testdata/buildcmd/Go/.gitignore
awood45: Git does pick it up, but I can move it to the base of the repo.

jfuss: O cool. I didn't actually know that. I don't have a preference where it lives, just didn't realize git was smart like that :)

appveyor-windows-build-go.yml
```diff
@@ -0,0 +1,79 @@
...
+    - "go env"
+
+    # Echo final Path
+    - "echo %PATH%"
+
+test_script:
+    # Reactivate virtualenv before running tests
+    - "venv\\Scripts\\activate"
+    - "docker system prune -a -f"
+    - "pytest -vv tests/integration/buildcmd/test_build_cmd.py -k test_with_go"
```
jfuss: Can't we just add this to the 'in-process' java and ruby appveyor file?

appveyor-windows-build-go.yml
```diff
@@ -0,0 +1,79 @@
...
+    - "pytest -vv tests/integration/buildcmd/test_build_cmd.py -k test_go_must_fail_with_container"
```
jfuss: Does this tests need docker installed? I don't remember the code paths but if we don't need docker installed to test this, then we can run it in the 'java-ruby-inprocess-build' appveyor project.

appveyor-windows-build-go.yml
```diff
@@ -0,0 +1,79 @@
+version: 1.0.{build}
```
jfuss: Did you create a appveyor project for this? Can you add it to the repo so we can see the build results from the github UI?

jfuss: O it is named: `continuous-integration/appveyor/pr`. Can you update the name in appveyor so the name we see here explains which tests that build belongs too?

jfuss:
This PR looks good to me. Some general questions:

* Are we planning to follow up with updating init?
* Assuming we are still planning to update aws-lambda-builders to support the debug build use-case? This will be needed if we want IDE support.
* We should probably update https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/build/command.py#L32. I am ok with a follow up, since I think we are missing runtimes we launched just before re:invent (python3.8, node12.x, java11).

awood45:
- Are we planning to follow up with updating init?
    - Done in the last change, will add an app-templates update when the release approaches.
- Assuming we are still planning to update aws-lambda-builders to support the debug build use-case? This will be needed if we want IDE support.
    - This will be a later PR.
- We should probably update https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/build/command.py#L32. I am ok with a follow up, since I think we are missing runtimes we launched just before re:invent (python3.8, node12.x, java11).
    - Updated the documentation, we had several out of date documentation lines around build.

samcli/commands/build/command.py
```diff
@@ -31,11 +31,12 @@
 \b
 Supported Runtimes
 ------------------
-1. Python 2.7, 3.6, 3.7 using PIP\n
-2. Nodejs 10.x, 8.10, 6.10 using NPM\n
+1. Python 2.7, 3.6, 3.7, 3.8 using PIP\n
```
sriram-mv: Awesome! Thanks for updating this!

PR #1743, opened 2020-01-23
Revert "Revert "fix: botocore session cache across samcli (#1693)""

sriram-mv:
Reverts awslabs/aws-sam-cli#1717

### DIFF
--- a/samcli/cli/context.py
+++ b/samcli/cli/context.py
@@ -2,10 +2,13 @@
 Context information passed to each CLI command
 """
 
-import uuid
 import logging
+import uuid
+
 import boto3
 import botocore
+import botocore.session
+from botocore import credentials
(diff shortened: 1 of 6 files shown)


samcli/cli/context.py
```diff
@@ -144,6 +147,11 @@ def _refresh_session(self):
         """
         try:
             boto3.setup_default_session(region_name=self._aws_region, profile_name=self._aws_profile)
+            # get botocore session and setup caching for MFA based
+            boto3.DEFAULT_SESSION._session.get_component("credential_provider").get_provider(  # pylint: disable=W0212
```
jfuss: This is using a private member. We should find a way not to rely on this. It boto changes this at all, we break.

PR #1744, opened 2020-01-23
fix: cache MFA based credentials

sriram-mv:
*Issue #, if available:*

https://github.com/awslabs/aws-sam-cli/issues/1682

SAM CLI does not cache MFA based credentials and asks for them each time as the credentials are not cached.

MFA based credentials are cached at ~/.aws/boto/cache

* Only one prompt for MFA based authentication, and if the MFA code is successful, only one prompt till the credentials that were retrieved expire.

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/designs/botocore_sessions.md
+++ b/designs/botocore_sessions.md
@@ -1,0 +1,117 @@
+CLI: Botocore Sessions
+====================================
+
+What is the problem?
+--------------------
+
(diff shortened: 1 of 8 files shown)


jfuss:
@sriram-mv Looks like there are some unit tests that need to be adjusted for the new implementation, but otherwise things look good.

PR #1766, opened 2020-01-31
fix: Handle global value for Function CodeUri in Package

awood45:
*Issue #, if available:* #1706

The `sam package` command does not read any global values over the course of packaging - traditionally, only build and deploy functionality would use any values. Issue #1706 points to an uncommon use case that exposes that some global values do need to be considered, at minimum so that package won't break deploy.

Creates an `apply_global_values` method that fetches global values we use (currently just the one) and applies them if/when they are applicable.

None

*Checklist:*

- [N/A] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [X] Write unit tests
- [X] `make pr` passes
- [X] Write documentation

### DIFF
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -616,7 +616,7 @@ class Template:
             resource_type = resource.get("Type", None)
             resource_dict = resource.get("Properties", None)
 
-            if "CodeUri" not in resource and resource_type == AWS_SERVERLESS_FUNCTION:
+            if "CodeUri" not in resource_dict and resource_type == AWS_SERVERLESS_FUNCTION:
                 code_uri_global = self.template_dict.get("Globals", {}).get("Function", {}).get("CodeUri", None)
                 if code_uri_global is not None and resource_dict is not None:
                     resource_dict["CodeUri"] = code_uri_global
@@ -636,8 +636,8 @@ class Template:
         if "Resources" not in self.template_dict:
             return self.template_dict
 
-        self.template_dict = self.export_global_artifacts(self.template_dict)
         self.template_dict = self.apply_global_values(self.template_dict)
+        self.template_dict = self.export_global_artifacts(self.template_dict)
 
         for resource_id, resource in self.template_dict["Resources"].items():
 
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -602,6 +602,22 @@ class Template:
 
         return template_dict
 
+    def apply_global_values(self, template_dict):
+        """
+        Takes values from the "Global" parameters and applies them to resources where needed for packaging.
+        """
+        for resource_id, resource in self.template_dict["Resources"].items():
+
+            resource_type = resource.get("Type", None)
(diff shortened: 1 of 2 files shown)


samcli/lib/package/artifact_exporter.py
```diff
@@ -616,6 +637,7 @@ def export(self):
             return self.template_dict
 
         self.template_dict = self.export_global_artifacts(self.template_dict)
+        self.template_dict = self.apply_global_values(self.template_dict)
```
sanathkr: To be future compat, you might wanna move this before `export_global_values`

samcli/lib/package/artifact_exporter.py
```diff
@@ -602,6 +602,27 @@ def export_metadata(self, template_dict):
...
+        packaging, even when there may not be a build step (such as the source being an S3 file). This is the only
+        known use case for using any global values in the package step, so any other such global value applications
+        should be scoped to this method if possible.
+        """
+        for _, resource in self.template_dict["Resources"].items():
+
+            resource_type = resource.get("Type", None)
+            resource_dict = resource.get("Properties", None)
+
+            if "CodeUri" not in resource and resource_type == AWS_SERVERLESS_FUNCTION:
```
sanathkr: this should be `CoreUri not in resource_dict` -> you wanna look into the properties.

samcli/lib/package/artifact_exporter.py
```diff
@@ -602,6 +602,27 @@ def export_metadata(self, template_dict):
...
+                code_uri_global = self.template_dict.get("Globals", {}).get("Function", {}).get("CodeUri", None)
+                if code_uri_global is not None and resource_dict is not None:
```
sanathkr: nit: `if not core_uri_global and not resource_dict` woudl be better

samcli/lib/package/artifact_exporter.py
sanathkr: The bug doesn't state, but `DefinitionUri` of `Globals->Api` also needs same treatment

samcli/lib/package/artifact_exporter.py
awood45: Done

jfuss:
Does Globals also support Layers? If so, that also will need to be added.

samcli/lib/package/artifact_exporter.py
awood45: My forehead moves quickly to meet my waiting palm.

awood45:
I believe that Globals would indeed support layers. I'll look in to that.

samcli/lib/package/artifact_exporter.py
awood45: Changed

samcli/lib/package/artifact_exporter.py
awood45: Will check this with the Layers issue as well.

awood45:
Sounds like Globals do not yet support Layers, so no-op there.

PR #1768, opened 2020-02-01
Efficient deploy log fetching

smatsumt:

This PR improves `sam deploy` inefficient log fetching. Current implementation fetches events repeatidly until deploy finished. And it fetches "all" log events for every fetches. It includes already shown events and old events before the current deployment starts.

So, when we update a stack in many times, sam-cli event showing becomes quite slow. And it takes a little bit longer time after actual stack deployment finished.

The PR implementation skip older events and prevent to fetch all, old events.

I guess almost nothing. Maybe the number of API calls increases because it does not fetch all old events.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -340,10 +340,9 @@ class Deployer:
                 is_reached_old_entry = False
                 for event_items in response_iterator:
                     for event in event_items["StackEvents"]:
-                        current_time_stamp_marker = utc_to_timestamp(event["Timestamp"])
-                        if event["EventId"] not in events and current_time_stamp_marker > time_stamp_marker:
+                        if event["EventId"] not in events and utc_to_timestamp(event["Timestamp"]) > time_stamp_marker:
                             events.add(event["EventId"])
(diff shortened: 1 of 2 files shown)


samcli/lib/deploy/deployer.py
```diff
@@ -336,10 +336,14 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
                 paginator = self._client.get_paginator("describe_stack_events")
                 response_iterator = paginator.paginate(StackName=stack_name)
                 stack_status = describe_stacks_resp["Stacks"][0]["StackStatus"]
+                latest_time_stamp_marker = time_stamp_marker
                 for event_items in response_iterator:
                     for event in event_items["StackEvents"]:
                         if event["EventId"] not in events and utc_to_timestamp(event["Timestamp"]) > time_stamp_marker:
                             events.add(event["EventId"])
+                            latest_time_stamp_marker = max(
```
sriram-mv: I understand the logic here, could you add a unit test to showcase this?

smatsumt: Thank you for the review!
I've added the unit test. Could you review it?

PR #1771, opened 2020-02-04
chore: version bump on aws-lambda-builders to 0.7.0

awood45:

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,5 +11,5 @@ dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.22.0
 serverlessrepo==0.1.9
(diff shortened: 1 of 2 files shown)


jfuss:
You need to update `base.txt` not isolated.

c2tarun:
> Ned to update the base.txt.

If this is the case, can we remove it from isolate.txt?

jfuss:
| Do we have any reason of not letting sam-cli to pick latest available minor version of lambda builders?

To provide a better install experience. If we do not pin, we run the risk of a deployment of aws-lambda-builders impacting already released version of SAM CLI. Ideally, we should be pinning all of our dependencies, but we allow customers to use pip so we only do some of the dependencies instead (mainly ones we control).

PR #1776, opened 2020-02-06
fix: appveyor windows job - python version

sriram-mv:
Why is this change necessary?

* With release v0.41.0 - MSIs have version 3.7.6 as their base version.
Appveyor testing needs to reflect that.

* Appveyor test frameworks are now using python 3.7.6. The environment
variables are set according to: https://www.appveyor.com/docs/windows-images-software/

* None

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-windows.yml
+++ b/appveyor-windows.yml
@@ -20,7 +20,6 @@ environment:
   PYTHON_HOME: "C:\\Python37-x64"
   PYTHON_SCRIPTS: "C:\\Python37-x64\\Scripts"
   PYTHON_EXE: "C:\\Python37-x64\\python.exe"
-  PYTHON_VERSION: '3.7.5'
   PYTHON_ARCH: '64'
   HOME: 'C:\Users\appveyor'
   HOMEDRIVE: 'C:'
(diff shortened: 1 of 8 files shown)


appveyor-windows.yml
```diff
@@ -16,11 +16,11 @@ environment:
...
-  # MSI Installers only use Py3.6.6. It is sufficient to test with this version here.
-  PYTHON_HOME: "C:\\Python36-x64"
-  PYTHON_SCRIPTS: "C:\\Python36-x64\\Scripts"
-  PYTHON_EXE: "C:\\Python36-x64\\python.exe"
-  PYTHON_VERSION: '3.6.8'
+  # MSI Installers only use Py3.7.6. It is sufficient to test with this version here.
+  PYTHON_HOME: "C:\\Python37-x64"
+  PYTHON_SCRIPTS: "C:\\Python37-x64\\Scripts"
+  PYTHON_EXE: "C:\\Python37-x64\\python.exe"
+  PYTHON_VERSION: '3.7.5'
```
jfuss: I think this is only for linux? Do we need this? Looks unused in the file.

jfuss:
Don't we need to update all the windows build projects as well?

appveyor-windows.yml
sriram-mv: It still adds visibility when looking at a appveyor test run.

PR #1789, opened 2020-02-11
fix: packaging of built artifacts

sriram-mv:
Why is this change necessary?

* sam deploy of the same built artifacts can result in new uploads to s3
during the package step.

* Hash of the directory is constructed rather than on the packaged zip
file.

* Potentially None

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/hash.py
+++ b/samcli/lib/utils/hash.py
@@ -1,0 +1,60 @@
+"""
+Hash calculation utilities for files and directories.
+"""
+import os
+import hashlib
+
+BLOCK_SIZE = 4096
+
+
+def file_checksum(file_name):
+    """
+
+    Parameters
+    ----------
+    file_name: file name of the file for which md5 checksum is required.
+
+    Returns
+    -------
+    md5 checksum of the given file.
+
+    """
+    with open(file_name, "rb") as file_handle:
+        md5 = hashlib.md5()
+
+        # Save current cursor position and reset cursor to start of file
+        curpos = file_handle.tell()
+        file_handle.seek(0)
+
+        buf = file_handle.read(BLOCK_SIZE)
+        while buf:
+            md5.update(buf)
+            buf = file_handle.read(BLOCK_SIZE)
+
+        # Restore file cursor's position
+        file_handle.seek(curpos)
+
+        return md5.hexdigest()
+
+
+def dir_checksum(directory, followlinks=True):
+    """
+
+    Parameters
+    ----------
+    directory : A directory with an absolute path
+    followlinks: Follow symbolic links through the given directory
+
+    Returns
+    -------
+    md5 checksum of the directory.
+
+    """
+    md5_dir = hashlib.md5()
+    # Walk through given directory and find all directories and files.
+    for dirpath, _, filenames in os.walk(directory, followlinks=followlinks):
+        # Go through every file in the directory and sub-directory.
+        for filepath in [os.path.join(dirpath, filename) for filename in filenames]:
+            # Encode file's checksum to be utf-8 and bytes.
+            md5_dir.update(file_checksum(filepath).encode("utf-8"))
+    return md5_dir.hexdigest()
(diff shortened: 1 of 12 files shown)


jfuss:
Is there a way we can integ test this? It will help guard against regressions in the future (especially as this has been a pain point for customers for some time now).

samcli/lib/utils/hash.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+import os
+import hashlib
+
+BLOCK_SIZE = 4096
+
+
+def file_checksum(file_name):
+    """
+
```
sanathkr: can you explain here it is specifically used with `sam deploy` and might potentially be back-compat issues if we break this?

samcli/lib/utils/hash.py
```diff
@@ -0,0 +1,60 @@
...
+    ----------
+    directory : A directory with an absolute path
+    followlinks: Follow symbolic links through the given directory
+
+    Returns
+    -------
+    md5 checksum of the directory.
+
+    """
+    md5_dir = hashlib.md5()
```
sanathkr: is md5 the right one? I am not sure if md5 is stable enough or has issues.

samcli/lib/utils/hash.py
```diff
@@ -0,0 +1,60 @@
...
+    # Walk through given directory and find all directories and files.
+    for dirpath, _, filenames in os.walk(directory, followlinks=followlinks):
```
sanathkr: Do you have to catch circular links?

samcli/lib/utils/hash.py
```diff
@@ -0,0 +1,60 @@
...
+        # Go through every file in the directory and sub-directory.
+        for filepath in [os.path.join(dirpath, filename) for filename in filenames]:
+            # Encode file's checksum to be utf-8 and bytes.
+            md5_dir.update(file_checksum(filepath).encode("utf-8"))
```
sanathkr: Hang on, you are just checksumming the file path and not the content of the files. Is this intentional? If so, Why is this sufficient?

sriram-mv: No, I'm passing it through to the file_checksum method.

samcli/lib/utils/hash.py
sriram-mv: Not sure how os.walk handles that, can look into it.

samcli/lib/utils/hash.py
sriram-mv: for checksum purposes it should be fine (I dont think its cryptographically secure though), we can also look at sha256 as well.

samcli/lib/utils/hash.py
sriram-mv: Makes sense, I can add a series of comments there.

samcli/lib/utils/hash.py
sriram-mv: Tried to create a series of files that had a circular symlink reference.

```
 ~/w/ap/py37/s/.aws-sam/build/HelloWorldFunction ▓▒░ ls -lrth a                                                                                 ░▒▓ ✔ | py3.7 Py | 12:46:18
lrwxr-xr-x  1 srirammv  ANT\Domain Users     1B Feb 12 12:46 a -> c
 ~/w/ap/py37/s/.aws-sam/build/HelloWorldFunction ▓▒░ ls -lrth b                                                                                 ░▒▓ ✔ | py3.7 Py | 12:46:23
lrwxr-xr-x  1 srirammv  ANT\Domain Users     1B Feb 12 12:46 b -> a
 ~/w/ap/py37/s/.aws-sam/build/HelloWorldFunction ▓▒░ ls -lrth c                                                                                 ░▒▓ ✔ | py3.7 Py | 12:46:25
lrwxr-xr-x  1 srirammv  ANT\Domain Users     1B Feb 12 12:46 c -> b
```

Get this error:

```
Error: Unable to upload artifact HelloWorldFunction referenced by CodeUri parameter of HelloWorldFunction resource.
[Errno 62] Too many levels of symbolic links: '/Users/srirammv/workspace/apps/py37_app/sam-app/.aws-sam/build/HelloWorldFunction/a'
```

So Looks like circular references are taken care of.

PR #1790, opened 2020-02-12
fix: display no outputs if outputs are not defined.

sriram-mv:
Why is this change necessary?

* bug from previous PR which always looked for outputs in deployed
templates.

* Explicitly check for outputs, before attempting to display them.
* Added extra tests to verify so that it does not cause regressions in
the future.

* None.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -401,7 +401,8 @@ class Deployer:
             raise deploy_exceptions.DeployFailedError(stack_name=stack_name, msg=str(ex))
 
         outputs = self.get_stack_outputs(stack_name=stack_name, echo=False)
-        self._stack_outputs(outputs)
+        if outputs:
+            self._display_stack_outputs(outputs)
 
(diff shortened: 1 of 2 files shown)


samcli/lib/deploy/deployer.py
```diff
@@ -401,7 +401,8 @@ def wait_for_execute(self, stack_name, changeset_type):
             raise deploy_exceptions.DeployFailedError(stack_name=stack_name, msg=str(ex))
 
         outputs = self.get_stack_outputs(stack_name=stack_name, echo=False)
-        self._stack_outputs(outputs)
+        if outputs:
+            self._stack_outputs(outputs)
```
sanathkr: should this be given a more descriptive name? When I saw this code in your old PR, I thought it was functional - turns out this is just cosmetic.

sriram-mv: yeah true, I'll change it to display stack outputs.

PR #1793, opened 2020-02-12
test: run by canary

c2tarun:
*Issue #, if available:*

Working on creating Canary to run all integration tests in secure environment. This code change disable skipping test if tests are triggered by Canary.

Canary Trigger will setup an environment variable which will be monitored by tests. If detected then this will disable skipping tests.

None

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


tests/testing_utils.py
```diff
@@ -8,6 +8,7 @@
 RUNNING_ON_CI = os.environ.get("APPVEYOR", False)
 RUNNING_TEST_FOR_MASTER_ON_CI = os.environ.get("APPVEYOR_REPO_BRANCH", "master") != "master"
 CI_OVERRIDE = os.environ.get("APPVEYOR_CI_OVERRIDE", False)
+RUN_BY_CANARY = os.environ.get("BY_CANARY", None) is not None
```
sriram-mv: Can we follow similar convention here? We say False on the rest.

c2tarun: Updated.

PR #1798, opened 2020-02-17
Allow the characters _:./+-@= to be used in tag keys during deploy

sriram-mv:
Copied from https://github.com/awslabs/aws-sam-cli/pull/1727

Using the `aws cloudformation deploy --tags` command, you are allowed to use these characters in the tag keys. Documentation also lists these characters as supported and allowed. With `sam deploy --tags`, it did not work properly due to a regex limiting supported characters to A-Za-z0-9.

By adding a regex specifically for tag keys. By using this new regex, `sam deploy --tags 'owner:name=son of anton'` no longer converts the tag key to `name` instead of `owner:name`.
Before:
```
{"name": "son of anton"}
```
After:
```
{"owner:name": "son of anton"}
```

None. More characters allowed in keys and fully backwards compatible.

*Checklist:*

- [ ] ~~Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))~~
- [ ] ~~Write unit tests~~
- [ ] ~~Write/update functional tests~~
- [X] `make pr` passes
- [ ] ~~Write documentation~~


msambol:
Thanks! Just discovered this when deploying.

samcli/cli/types.py
```diff
@@ -9,15 +9,47 @@
 import click
 
 
-def _value_regex(delim):
-    return f'(\\"(?:\\\\.|[^\\"\\\\]+)*\\"|(?:\\\\.|[^{delim}\\"\\\\]+)+)'
+def _match_regex(match, delim):
+    return f'(\\"(?:\\\\{match}|[^\\"\\\\]+)*\\"|(?:\\\\{match}|[^{delim}\\"\\\\]+)+)'
```
jfuss: This is one crazy to read regex... :(

samcli/cli/types.py
```diff
@@ -9,15 +9,47 @@
 import click
 
 
-def _value_regex(delim):
-    return f'(\\"(?:\\\\.|[^\\"\\\\]+)*\\"|(?:\\\\.|[^{delim}\\"\\\\]+)+)'
+def _match_regex(match, delim):
```
jfuss: Can you throw a quick doc string on this? Finding it slightly hard to understand what/why you would need the match here.

samcli/cli/types.py
```diff
@@ -9,15 +9,47 @@
...
-VALUE_REGEX_SPACE_DELIM = _value_regex(" ")
+VALUE_REGEX_SPACE_DELIM = _match_regex(match=".", delim=" ")
 # Use this regex when you have comma as delimiter Ex: "KeyName1=string,KeyName2=string"
-VALUE_REGEX_COMMA_DELIM = _value_regex(",")
+VALUE_REGEX_COMMA_DELIM = _match_regex(match=".", delim=",")
+
+
+def _unquote(value):
+    r"""
+    Removes wrapping double quotes and any '\ ' characters. They are usually added to preserve spaces when passing
```
jfuss: You say "They are usually added". So does that mean there is a case that we wouldn't want to replace the double or '\ ' characters from the input?

samcli/cli/types.py
```diff
@@ -9,15 +9,47 @@
...
+    value : str
+        Input to unquote
+
+    Returns
+    -------
+    Unquoted string
+    """
+    if value and (value[0] == value[-1] == '"'):
+        # Remove quotes only if the string is wrapped in quotes
+        value = value.strip('"')
```
jfuss: The intention here is to only remove the first and last quotes right? Do we need to worry about the case where there are two `"` in the beginning or end of the string input?

Below was run in Python3.8
```
>>> value = '"a\""'
>>> value.strip('"')
'a'
>>> value = '""a""'
>>> value.strip('"')
'a'
```

Doing a slice might be better here:
```
>>> value = '"a\""'
>>> value[1:-1]
'a"'
>>> value = '""a""'
>>> value[1:-1]
'"a"'
```

sriram-mv: This only cares about wrapping quotes and not the entire nesting, can change the function name to reflect that.

samcli/cli/types.py
sriram-mv: I dont think so, we will want to replace it. "Usually added" maybe a misnomer, preserving spaces inside quotes or even escaping quotes within quotes requires a `\` (escape) character.

samcli/cli/types.py
sriram-mv: Will do

samcli/cli/types.py
sriram-mv: Will add a docstring.

Issue #1161, opened 2019-05-03
Sam local converts an empty response to the string "no data"


johnc44:
### Description

We have an API that, for PUTS, POSTS and DELETES returns an HTTP status 200 with no content.

When running the lambda through sam local start-api, SAM seems to change the content to be the string "no data".  I had a quick look in the source (local_apigw_service) and can see this:

body = json_output.get("body") or "no data"

The problem manifested to us because out front-end is Angular and were were getting a JSON deserialization error when running locally that we weren't seeing when running through AWS.

So it looks intentional.  But this behaviour is not the same as when I run through the lambda and API gateway.

### Steps to reproduce

Lambda returns an HTTP status 200 with no body.  In our example, it looks like this:

{'isBase64Encoded': False, 'statusCode': 200, 'headers': {'Access-Control-Allow-Origin': 'localhost'}, 'body': None}

### Observed result

When running through AWS, I get Content-Length: 0, empty body, http 200.

When running through SAM local, I get Content-Length: 7, with a body "no data"

### Expected result

sam local start-api should work the same way as running through AWS, and should never tamper with my data.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10
2. `sam --version`:  0.15.0

johnc44:
I should say that we have a workaround in our angular app, which is to add , responseType: 'text' to the calls.

But it's just another frustrating difference with start-api and the real implementation.

nickbreen:
I've been chasing this in some automated tests of my own!

Further:
- when using `sam local invoke` I get the expected empty body and `Content-Length: 0`
- when using `sam local start-api` and using `curl` I get the unexpected `no data` body and `Content-Length: 7

It'd be appropriate for `sam local` to _log_ that there is no data, rather than corrupting the response.

awood45:
I can definitely recreate this and this is indeed odd behavior.

As a workaround, this doesn't happen to HTTP 204 responses (which is technically the semantically correct way to have a no-body response), but this should be fixed for other response codes as well.

johnc44:
Setting http 204 response is a great solution, thanks @awood45 awood45.  (Ashamed to say I didn't know that this even existed).  Should still be fixed in sam as it shouldn't be modifying responses, but happy to change my coe to use 204 for no responses.

jfuss:
Released in [v0.44.0](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.44.0).

Closing

PR #1813, opened 2020-02-21
Feature/lambda empty return

francilioaraujo:
*Issue #, if available:* #1161

*Why is this change necessary?* The API response when lambda doesn't provide a response was unintuitive.

*How does it address the issue?* The API return value was changed when lambda returns nothing.

*What side effects does this change have?* The API return value was changed to some cases and a log message was added

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -266,7 +266,9 @@ class LocalApigwService(BaseLocalService):
         headers = LocalApigwService._merge_response_headers(
             json_output.get("headers") or {}, json_output.get("multiValueHeaders") or {}
         )
-        body = json_output.get("body") or "no data"
+        body = json_output.get("body")
+        if body is None:
+            LOG.warning("Lambda returned empty body!")
         is_base_64_encoded = json_output.get("isBase64Encoded") or False
 
         try:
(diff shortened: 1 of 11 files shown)


francilioaraujo:
@jfuss, I can't reproduce this error locally. I set up an ubuntu VM using vagrant and ran only the TestBuildCommand_Go_Modules inside test_build_cmd.py. Installed Go version 1.14 and python 3.8.0 inside the VM, But all 5 tests passed.
The lambda go function isn't returning the "{'message': 'Hello World'}" string on AppVeyor. On my VM the test passes. Could you give some clarification?

jfuss:
@francilioaraujo I re-kicked off the failed tests. We have been seeing this randomly fail in appveyor but have not been able to root cause the issue.

francilioaraujo:
@jfuss, thank you for re-executing the test. Now AppVeyor status is 'Completed' but the AppVeyor app on Github still shows as pending. Do we wait for it to update?

jfuss:
@francilioaraujo Ya I just noticed that as well. There looks to be some kind of Github outage: https://www.githubstatus.com/incidents/q07bfjh7jf1t. This could be related to that. I wouldn't worry about it. I am going do another review pass right now though.

jfuss:
@francilioaraujo I am going to check back here in a couple hours to see if Github catches up on the status so we don't have to do any manually overriding (I really don't want to kick off the whole build again, as it takes some time to actually finish). 

There shouldn't be any need for further updates from you at this point. Thanks for the contribution!

francilioaraujo:
Thanks for all the support @jfuss. I'll be watching for updates.

Issue #1404, opened 2019-09-12
`start-lambda` does not catch runtime errors, so calls to it never fail


mereghost:
### Description

With `sam local start-lambda` the failure is only taken into account if the actual function returns the specified json with all 3 keys (even if just 1 is mandatory) (`errorType`, `errorMessage` and `stackTrace`).

When a function fails due to other issues, a wrong method call or some kind of runtime error for example, `start-lambda` will not catch it, therefore marking the task as succeeded.

This hampers any effort of testing step functions with retry mechanisms or branching based on failure. 

### Steps to reproduce

A simple app that should trigger failure (ruby in this case, but language is irrelevant):

```ruby
def handle(context:, event:)
  not_defined_value.someMethod
end
```

### Observed result

The response from the service is `{ "statusCode" : 200 }`

### Expected result

A response that indicates that an error occured: `{ "statusCode" : 200, "FunctionError" : "Unhandled" }`

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Linux
2. `sam --version`: 0.21.0

anthonymq:
This behaviour implies that we can't develop step functions locally as every failing step is considered a Success.

oben59:
Same issue in Linux

anthonymq:
> Same issue in Linux with 0.21.0

Same in 0.22.0

anton-pfc:
Same problem here on MacOS. 
Please improve SAM so we can use it for local development. 

Cmon Amazon you can do it better!

zhulik:
Same here, this issue makes SAM absolutely useless for integration testing.

adamrecsko:
+1
 same issue here

ashleyfrieze:
I had the exact same problem as the above in Java, even though Java was correctly throwing an Exception and the Lambda was finishing with the right JSON exception output.

Looking at the code, it seems that the valid output for an error is expected to have 2 or 3 properties. In Java, the `cause` field provides a fourth. I worked around this in my code so that the exception thrown by the handler did not provide a `cause`. This resolved the issue for me.

However, it seems to me that the correct logic should be `2, 3, 4` and when I get a moment, I'll create a PR for that.

ashleyfrieze:
I have a fix for this issue in #1820 - any chance of someone reviewing it?

jfuss:
[...]

PR #1820, opened 2020-02-25
Support 4 property errors for Java lambdas

ashleyfrieze:
*Issue #1404 *

When a Java thrown exception has a cause, the `start-lambda` runtime does not mark it as a failure. When it does not have a cause, it does.

Supports 4 property dicts as sent by Java.

Unknown?

*Checklist:*

- [X] Write unit tests
- [X] `make pr` passes
- [X] Write documentation

### DIFF
--- a/samcli/local/services/base_local_service.py
+++ b/samcli/local/services/base_local_service.py
@@ -152,11 +152,14 @@ class LambdaOutputParser:
             # Error/Exception that was raised/returned/thrown from the container. To ensure minimal false positives in
             # this checking, we check for all the keys that can occur in Lambda raised/thrown/returned an
             # Error/Exception. This still risks false positives when the data returned matches exactly a dictionary with
-            # the keys 'errorMessage', 'errorType' and 'cause'
+            # the keys 'errorMessage', 'errorType', 'stackTrace' and 'cause'.
+            # This also accounts for a situation where there are three keys returned, two of which are
+            # 'errorMessage' and 'errorType', for languages with different error signatures
             if (
                 isinstance(lambda_response_dict, dict)
-                and len(lambda_response_dict.keys() & {"errorMessage", "errorType", "stackTrace", "cause"})
-                  == len(lambda_response_dict)
+                and ((len(lambda_response_dict.keys() & {"errorMessage", "errorType", "stackTrace", "cause"})
+                  == len(lambda_response_dict))
+                  or (len(lambda_response_dict) - len(lambda_response_dict.keys() & {"errorMessage", "errorType"}) <= 1))
             ):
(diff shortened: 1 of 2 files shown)


ashleyfrieze:
This needs a fix to the failing unit test. It may also need the python to specifically check for allowing `cause` as an extra field. If someone can pitch in and help, then please do. Otherwise, I'll look into this one soon.

ashleyfrieze:
I've submitted some rather clunky Python to attempt to fix the failing test and I've added a test for the use case I'm trying to cover.

ashleyfrieze:
@jfuss - please can I have some review/help on this one. It looks like the CI build is working ok, but then encountering some errors also visible in other PRs.

This should be a minor change. Please advise what I need to do to bring it up to a point where you can merge it. We'd rather like to have our `cause` passed through to the logs in our Java lambda, which we test using `sam local start-lambda` with a StepFunction runtime driving the process. At the moment, we're using a workaround where we swallow the `cause`. This will probably fix that.

Thanks!

ashleyfrieze:
Hey @jfuss @sriram-mv @sanathkr @awood45  - I've just squashed and rebased this branch against the latest `develop` in order to see if it will build any better. As far as I'm aware, this is a minor change that the slightly extended unit tests are happy with.

Without this modification, any Java exception, thrown in a lambda with a `cause` field will make the sam lambda runtime fail to recognise that an error has occurred.

I'm sure you guys are busy, but if I could have a review on this one, that would be great.

ashleyfrieze:
Hmmm... integration test failure. I'm guessing that the int test relies on an error being caught or not caught in a way which disagrees with the spec suggested by the unit tests and code. I'll submit changes if someone can give me a clue what's expected here.

ashleyfrieze:
Hey @jfuss - thanks for the approval... I'm still not sure whether I should be worried about the test failures. I usually worry about test failures, but I'm not sure in this case whether they're caused by this change. Any ideas?

jfuss:
@ashleyfrieze It looks like the files are not formatted (which is causing the failure). Have you run `make black`? (https://github.com/psf/black needs to be installed).

ashleyfrieze:
@jfuss - thanks for the tip. I must confess, as a non Python developer (I've written some in the past, but never for more than a few hours at a time), I just edited the files in the GitHub WebIDE!

However, I've run them through `black` on my computer and updated the PR. Here's hoping it will pass everything.

ashleyfrieze:
@jfuss - phew, it looks like this one's now building to green. Thanks for you help!

jfuss:
@ashleyfrieze Thank you for contributing the fix!

PR #1828, opened 2020-02-27
feat: adding separate bucket based on python version for integration tests.

c2tarun:
*Issue #, if available:*

        Our integration tests use S3 bucket to run several tests. Given that S3 objects are eventually consistent and we are using same bucket for lot of integration tests, we want to have multiple buckets to reduce transient failures. This change will allow referring correct bucket based on the python version for which test is running.

This spreads total number of objects across several buckets and improving sync time to reduce time it takes by S3 to be consistent.

NA

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/package/package_integ_base.py
+++ b/tests/integration/package/package_integ_base.py
@@ -29,7 +29,10 @@ class PackageIntegBase(TestCase):
 
         For backwards compatibility we are falling back to reading AWS_S3 so that current tests keep working.
         """
-        cls.pre_created_bucket = os.environ.get(os.environ.get("AWS_S3", False), os.environ.get("AWS_S3"))
+        cls.pre_created_bucket = os.environ.get(
+            os.environ.get("AWS_S3", False),  # Reading env var key for specific bucket.
+            os.environ.get("AWS_S3"),  # Fallback to old logic if separate env key not present.
+        )
         cls.bucket_name = cls.pre_created_bucket if cls.pre_created_bucket else str(uuid.uuid4())
         cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "package")
 
--- a/tests/integration/package/package_integ_base.py
+++ b/tests/integration/package/package_integ_base.py
@@ -29,10 +29,7 @@ class PackageIntegBase(TestCase):
 
         For backwards compatibility we are falling back to reading AWS_S3 so that current tests keep working.
         """
-        cls.pre_created_bucket = os.environ.get(
-            os.environ.get("AWS_S3", False),  # Reading env var key for specific bucket.
-            os.environ.get("AWS_S3"),  # Fallback to old logic if separate env key not present.
-        )
+        cls.pre_created_bucket = os.environ.get(os.environ.get("AWS_S3"), False)
         cls.bucket_name = cls.pre_created_bucket if cls.pre_created_bucket else str(uuid.uuid4())
         cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "package")
 
--- a/tests/integration/package/package_integ_base.py
+++ b/tests/integration/package/package_integ_base.py
@@ -14,7 +14,22 @@ class PackageIntegBase(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.region_name = os.environ.get("AWS_DEFAULT_REGION")
-        cls.pre_created_bucket = os.environ.get("AWS_S3", False)
+        """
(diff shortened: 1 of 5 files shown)


tests/integration/package/package_integ_base.py
```diff
@@ -14,7 +14,22 @@ class PackageIntegBase(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.region_name = os.environ.get("AWS_DEFAULT_REGION")
```
jfuss: This should be below doc strings.

tests/integration/package/package_integ_base.py
```diff
@@ -14,7 +14,22 @@ class PackageIntegBase(TestCase):
...
+        and we are using same bucket for lot of integration tests, we want to have multiple buckets to reduce
+        transient failures. In order to achieve this we created 3 buckets one for each python version we support (3.6,
+        3.7 and 3.8). Tests running for respective python version will use respective bucket.
+
+        AWS_S3 will point to a new environment variable AWS_S3_36 or AWS_S3_37 or AWS_S3_38. This is controlled by
+        Appveyor. These environment variables will hold bucket name to run integration tests. Eg:
+
+        For Python36:
+        AWS_S3=AWS_S3_36
+        AWS_S3_36=aws-sam-cli-canary-region-awssamclitestbucket-forpython36
```
jfuss: Do no document bucket names.

tests/integration/package/package_integ_base.py
```diff
@@ -14,7 +14,22 @@ class PackageIntegBase(TestCase):
...
+
+        For backwards compatibility we are falling back to reading AWS_S3 so that current tests keep working.
+        """
+        cls.pre_created_bucket = os.environ.get(os.environ.get("AWS_S3", False), os.environ.get("AWS_S3"))
```
jfuss: Break this line up, I am having a hard time following the nested calls here.

tests/integration/publish/publish_app_integ_base.py
```diff
@@ -16,7 +16,8 @@ class PublishAppIntegBase(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.region_name = os.environ.get("AWS_DEFAULT_REGION")
-        cls.pre_created_bucket = os.environ.get("AWS_S3", False)
+        """Please read comments in package_integ_base.py for more details around this."""
+        cls.pre_created_bucket = os.environ.get(os.environ.get("AWS_S3", False), os.environ.get("AWS_S3"))
```
jfuss: same

tests/regression/package/regression_package_base.py
```diff
@@ -17,7 +17,8 @@ class PackageRegressionBase(TestCase):
     @classmethod
     def setUpClass(cls):
         cls.region_name = os.environ.get("AWS_DEFAULT_REGION")
-        cls.pre_created_bucket = os.environ.get("AWS_S3", False)
+        """Please read comments in package_integ_base.py for more details around this."""
+        cls.pre_created_bucket = os.environ.get(os.environ.get("AWS_S3", False), os.environ.get("AWS_S3"))
```
jfuss: same

tests/integration/package/package_integ_base.py
c2tarun: This is not a real bucket name. Something I made up based on how real name looks like.

tests/integration/package/package_integ_base.py
c2tarun: Actually doc string is only for the line where we are reading `AWS_S3` and not for the whole method. That is why I didn't move this below docstring. Thoughts?

tests/integration/package/package_integ_base.py
c2tarun: Done.

tests/integration/package/package_integ_base.py
```diff
@@ -14,7 +14,25 @@ class PackageIntegBase(TestCase):
...
+
+        For Python36:
+        AWS_S3=AWS_S3_36
+        AWS_S3_36=aws-sam-cli-canary-region-awssamclitestbucket-forpython36
+
+        For backwards compatibility we are falling back to reading AWS_S3 so that current tests keep working.
+        """
+        cls.pre_created_bucket = os.environ.get(
+            os.environ.get("AWS_S3", False),  # Reading env var key for specific bucket.
+            os.environ.get("AWS_S3"),  # Fallback to old logic if separate env key not present.
```
jfuss: This is not the same fallback. The previous fallback had a value of False.

PR #1829, opened 2020-02-27
fix: bug around using temporary credentials in event bridge integration test.

c2tarun:
*Issue #, if available:*

Event bridge integration tests do no respect session token to run secure tests. This code change fixes that.

Using session token if present.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/init/schemas/schemas_test_data_setup.py
+++ b/tests/integration/init/schemas/schemas_test_data_setup.py
@@ -134,7 +134,8 @@ Y
 aws_access_key_id = {access_key}
 aws_secret_access_key = {secret_key}
 """
-        cred_profile_content += f"aws_session_token={session_token}\n"
+        if session_token:
+            cred_profile_content += f"aws_session_token={session_token}\n"
         return cred_profile_content
 
 
--- a/tests/integration/init/schemas/schemas_test_data_setup.py
+++ b/tests/integration/init/schemas/schemas_test_data_setup.py
@@ -17,14 +17,6 @@ AWS_PROFILE = "AWS_PROFILE"
 SLEEP_TIME = 1
 
 
-def _print_relevant_environment_vars(environ):
-    session = Session()
-    print("Session current region={}".format(session.region_name))
-    print("Session available_profiles={}".format(session.available_profiles))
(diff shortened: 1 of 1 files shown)


tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -128,17 +115,28 @@ def _create_config_file(self, profile, region):
             file.write(config_file_content)
         return custom_config
 
-    def _create_cred_file(self, profile, access_key, secret_key):
-        cred_file_content = "[default]\naws_access_key_id = {1}\naws_secret_access_key = {2}\n"
+    def _create_cred_file(self, profile, access_key, secret_key, session_token=None):
+        cred_file_content = self._create_cred_profile("default", access_key, secret_key, session_token)
```
jfuss: Doesn't this need to be `[default]`?

jfuss: O... this is moved down to the helper method below.

c2tarun: That is taken care of in the method line 133. It is `[{profile_name}]`

tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -128,17 +115,28 @@ def _create_config_file(self, profile, region):
...
+    def _create_cred_profile(self, profile_name, access_key, secret_key, session_token=None):
+        """
+        Method to create aws credentials entry similar to ~/.aws/credentials file format.
+        """
+        cred_profile_content = f"""
+[{profile_name}]
+aws_access_key_id = {access_key}
+aws_secret_access_key = {secret_key}
+"""
+        cred_profile_content += f"aws_session_token={session_token}\n"
```
jfuss: Why not have this in the `cred_profile_content` above?

c2tarun: Ohh that is a miss, there is supposed to be an if in there to add this only if session_token is passed. Lost somewhere in merging. I'll fix it.

PR #1832, opened 2020-02-28
refactor: deploy integration tests to use same level of logging.

c2tarun:
*Issue #, if available:*

Our Deploy integration tests sometimes fail without any log message or CFN stack id. This change is extracting some code out around command execution and adding some logging which will be displayed when tests fail.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/deploy/test_deploy_command.py
+++ b/tests/integration/deploy/test_deploy_command.py
@@ -3,6 +3,7 @@ import shutil
 import tempfile
 import uuid
 import time
+from collections import namedtuple
 from subprocess import Popen, PIPE, TimeoutExpired
 from unittest import skipIf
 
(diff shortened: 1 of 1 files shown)


tests/integration/deploy/test_deploy_command.py
```diff
@@ -21,6 +22,8 @@
 CFN_SLEEP = 3
 TIMEOUT = 300
 
+CommandResult = namedtuple("CommandResult", "process stdout stderr")
```
sriram-mv: No commas? are they not required?

c2tarun: Named tuple have two formats, one is array of var names, second is space separated var names. Both work.

PR #1850, opened 2020-03-05
fix: removing object deletion and using same stack name for integration tests to avoid NoSuchKey errors.

c2tarun:
*Issue #, if available:*

We often see NoSuchKey error when running integration tests when trying to use object uploaded to S3. For our integration tests we rarely change our objects.

 This change will stop object deletion at the end of integration test, so that we have higher chance of successful integration test.

None

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/deploy/test_deploy_command.py
+++ b/tests/integration/deploy/test_deploy_command.py
@@ -53,7 +53,7 @@ class TestDeploy(PackageIntegBase, DeployIntegBase):
 
             self.assertEqual(package_process.process.returncode, 0)
 
-            stack_name = "test-package-and-deploy-no-s3-bucket-all-args" + CFN_PYTHON_VERSION_SUFFIX
+            stack_name = f"test-package-and-deploy-no-s3-bucket-all-args-{CFN_PYTHON_VERSION_SUFFIX}"
             self.stack_names.append(stack_name)
 
             # Deploy and only show changeset.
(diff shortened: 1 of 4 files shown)


tests/integration/deploy/test_deploy_command.py
```diff
@@ -52,7 +53,7 @@ def test_package_and_deploy_no_s3_bucket_all_args(self, template_file):
 
             self.assertEqual(package_process.process.returncode, 0)
 
-            stack_name = "a" + str(uuid.uuid4()).replace("-", "")[:10]
+            stack_name = f"test-package-and-deploy-no-s3-bucket-all-args-{CFN_PYTHON_VERSION_SUFFIX}"
```
sriram-mv: can we do something like

https://docs.python.org/3/library/unittest.html#unittest.TestCase.id

instead?

c2tarun: Man, I was looking for this but was searching for `how to get current method name` :) I'll make the change.

tests/integration/deploy/test_deploy_command.py
```diff
@@ -21,6 +21,7 @@
 SKIP_DEPLOY_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
 CFN_SLEEP = 3
 TIMEOUT = 300
+CFN_PYTHON_VERSION_SUFFIX = os.environ.get("PYTHON_VERSION", "3.6.9").replace(".", "-")
```
jfuss: Should we default to something that is not a valid Python version? This could be confusing to tell if there was no Python Version env var or the value of the env var was 3.6.9.

c2tarun: Yeah, thats fair, this would go out of sync if python minor version is updated. I'll make that change.
Thanks

tests/integration/deploy/test_deploy_command.py
```diff
@@ -184,7 +187,8 @@ def test_no_package_and_deploy_with_s3_bucket_all_args_confirm_changeset(self, t
     def test_deploy_without_s3_bucket(self, template_file):
         template_path = self.test_data_path.joinpath(template_file)
 
-        stack_name = "a" + str(uuid.uuid4()).replace("-", "")[:10]
+        method_name = self.id().split(".")[-1]
```
sriram-mv: Nit: can probably move this to a method.

PR #1853, opened 2020-03-06
add missing --no-confirm-changeset option

fbeister:

It is impossible to switch changeset confirmation off

### DIFF
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -115,6 +115,7 @@ LOG = logging.getLogger(__name__)
 )
 @click.option(
     "--confirm-changeset/--no-confirm-changeset",
+    default=False,
     required=False,
     is_flag=True,
     help="Prompt to confirm if the computed changeset is to be deployed by SAM CLI.",
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -114,7 +114,7 @@ LOG = logging.getLogger(__name__)
     "non-zero exit code.",
 )
 @click.option(
-    "--confirm-changeset",
+    "--confirm-changeset/--no-confirm-changeset",
     required=False,
     is_flag=True,
     help="Prompt to confirm if the computed changeset is to be deployed by SAM CLI.",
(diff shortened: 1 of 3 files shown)


jfuss:
@fbeister You can turn it off by not supplying the option though. What's the use-case for strictly needing a flag?

fbeister:
If I don't supply the option, "confirm changeset" defaults to True (at least for the version I have) and there is no way to get a non-interactive mode.

fbeister:
Cross reference: https://github.com/awslabs/aws-sam-cli/issues/1600

jfuss:
@fbeister I don't think that is the intended behavior. What version are you running? Do you have a `samconfig.toml` file where `confirm_changeset` is set to `True`?

I do think have complementary flags is a good thing. I want to take a little time and think this completely through and run some cases through to make sure we don't regress or change behavior (especially with guided).

fbeister:
> @fbeister I don't think that is the intended behavior. What version are you running? Do you have a `samconfig.toml` file where `confirm_changeset` is set to `True`?

Well, how do I put this, ...
`confirm_changeset = true` --- my bad :-D
I created this samconfig when I was starting to experiment with SAM and never came back to really look at it when it became my copy-and-paste blueprint for samconfigs to come. Let me try to remove this statement, then :roll_eyes:

jfuss:
@fbeister No worries. So I think there are a couple things here then.

1. We need to output when we are reading from the config. While this doesn't solve the issue, I think part of this is no communication was given that we were picking up the config. Giving the clue of, hey we are reading values from here, would have helped you realize something was happening 'magically'.
2. This pr.

I should have time today to do look at this one deeper.

kmcarter:
Can this change be documented [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-deploy.html)? I just went on a goosechase looking for this option and only found this PR after implementing a hack and then deciding to create an issue (which now I don't have to do). TIL that the --help documentation is more up to date than the online documentation.

For anyone curious, here is my use case: I have a samconfig.toml file in my project that is intended for local deploy testing only (confirm changeset is as expected, etc), so `confirm_changeset = true` in that file. `sam build` [cannot exclude files](https://github.com/awslabs/serverless-application-model/issues/308), so this config file goes into the build and when the build enters our CI/CD pipeline, `sam deploy` cannot override `confirm_changeset = true` without something like `--no-confirm-changeset`.

PR #1867, opened 2020-03-16
fix: default capabilities on guided mode to be a list

sriram-mv:
Why is this change necessary?

* Previously, the string was passed along which leads to a validation
failure.

* The default is now a list.

* None.

*Issue #, if available:*

https://github.com/awslabs/aws-sam-cli/issues/1848

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
@sriram-mv 

No unit tests needed to be updated?

Should we add an integration tests for this?

PR #1869, opened 2020-03-17
fix: adding better logging to build_integ_test base's run of Popen command

c2tarun:
*Issue #, if available:*
This change will print better logs, stderr and stdout around the command that fails when running `test_with_go` integration tests.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/buildcmd/build_integ_base.py
+++ b/tests/integration/buildcmd/build_integ_base.py
@@ -13,7 +13,7 @@ import jmespath
 from pathlib import Path
 
 from samcli.yamlhelper import yaml_parse
-from tests.testing_utils import IS_WINDOWS
+from tests.testing_utils import IS_WINDOWS, _run_command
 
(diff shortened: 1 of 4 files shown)


tests/integration/buildcmd/build_integ_base.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
 
 from samcli.yamlhelper import yaml_parse
-from tests.testing_utils import IS_WINDOWS
+from tests.testing_utils import IS_WINDOWS, _run_command
```
jfuss: Can we make `_run_command` not private since it is being imported into another file?

c2tarun: Updated, this made PR little long, but it is mostly rename by by IDE.

Issue #1432, opened 2019-09-26
Error: AllowOrigin must be a quoted string when using !Sub/!ImportValue


johnc44:
### Description

Template uses !Sub to determine the CORS string to use.  When using sam local start-pi the following error is thrown:
b'Error: AllowOrigin must be a quoted string (i.e. "\'value\'" is correct, but "value" is not).\n'

### Steps to reproduce

In template.yml specify AllowOrigin field to use !Sub, for example:

      AllowOrigin: !Sub
        - "'https://${FrontEndDomainName}.${dnszone}'"
        - {dnszone: !ImportValue mydomain-com-ZoneName}

Then sam build --use-container && sam local start-api

### Observed result

Found '15' API Events in Serverless function with name 'myprojectApiLambda'
Sending Telemetry: {'metrics': [{'commandRun': {'awsProfileProvided': False, 'debugFlagProvided': True, 'region': '', 'commandName': 'sam local start-api', 'duration': 135, 'exitReason': 'UserException', 'exitCode': 1, 'requestId': '226f246e-d2cb-4cd9-a496-a706b32e3284', 'installationId': '450286d8-42a1-4f22-9c69-81080b051b56', 'sessionId': '0d0e0c51-2097-4d57-9fb5-8569204975c9', 'executionEnvironment': 'CLI', 'pyversion': '3.7.2', 'samcliVersion': '0.22.0'}}]}
HTTPSConnectionPool(host='aws-serverless-tools-telemetry.us-west-2.amazonaws.com', port=443): Read timed out. (read timeout=0.1)
Error: AllowOrigin must be a quoted string (i.e. "'value'" is correct, but "value" is

### Expected result

Should work...!

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: OSX Mojave
2. `sam --version`: 0.22

awood45:
Something I'm noticing here:

```
"'https://${FrontEndDomainName}.${dnszone}'"
```

The inside of the string would thus be:

```
'https://${FrontEndDomainName}.${dnszone}'
```

Is it intended to have two wrapping sets of quotes inside the YAML file?

awood45:
Past this, it looks like we don't yet have built in support for Import Values for sam local - noting the feature request.

johnc44:
> Something I'm noticing here:
> 
> ```
> "'https://${FrontEndDomainName}.${dnszone}'"
> ```
> 
> The inside of the string would thus be:
> 
[...]

PR #1874, opened 2020-03-18
fix: Log warning if there is unresolved intrinsics in cors

jfuss:
*Issue #, if available:*
#1432

This change allows for the Lambda behind an API to be reachable when SAM CLI fails to resolve an intrinsic function in the API Cors definition.

When reading the CORS configuration, we check to see if the value is a string that starts with `!` (indicating it is an intrinsics, !Ref for example) or anything but a string. This will then log a message to tell customers we are ignoring this property due to the intrinsic being unresolved.

No

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/sam_api_provider.py
+++ b/samcli/lib/providers/sam_api_provider.py
@@ -148,7 +148,14 @@ class SamApiProvider(CfnBaseApiProvider):
         """
         prop = cors_dict.get(prop_name)
         if prop:
-            if (not isinstance(prop, string_types)) or (not (prop.startswith("'") and prop.endswith("'"))):
+            if not isinstance(prop, string_types) or prop.startswith("!"):
+                LOG.warning(
+                    "CORS Property %s was not fully resolved. Will proceed as if the Property was not defined.",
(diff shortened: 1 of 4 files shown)


samcli/lib/providers/sam_api_provider.py
```diff
@@ -148,7 +148,14 @@ def _get_cors_prop(cors_dict, prop_name):
         """
         prop = cors_dict.get(prop_name)
         if prop:
-            if (not isinstance(prop, string_types)) or (not (prop.startswith("'") and prop.endswith("'"))):
+            if not isinstance(prop, string_types) or prop.startswith("!"):
```
c2tarun: Some intrinsic functions starts with `Fn::`, should we check for those as well?

jfuss: That are covered under the first check. Basically, anything that isn't a string gets caught. For `FN::` intrinsics, they are all dictionaries (key is `FN::`, value is a string, list, or dictionary).

PR #1889, opened 2020-03-24
fix: logging for samconfig

sriram-mv:
Why is this change necessary?

* Make it abundantly clear that a config file is being looked for during
cli invocation.

* Removes obfuscation of the config file and the options with which SAM
CLI is being invoked with.

* There will always be a blurb printed to the console on "Samconfig",
irrespective of the command being used.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -15,10 +15,15 @@ from samcli.commands.exceptions import ConfigException
 from samcli.lib.config.exceptions import SamConfigVersionException
 from samcli.cli.context import get_cmd_names
 from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV
+from samcli.lib.utils.sam_logging import SamCliLogger
 
 __all__ = ("TomlProvider", "configuration_option", "get_ctx_defaults")
 
+# Adding Samconfig prefix to the formatter here to show explicitly that the log messages generated
+# are exclusively from the configuration file context.
+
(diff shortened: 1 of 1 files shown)


samcli/cli/cli_config_file.py
```diff
@@ -15,10 +15,15 @@
...
 from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV
+from samcli.lib.utils.sam_logging import SamCliLogger
 
 __all__ = ("TomlProvider", "configuration_option", "get_ctx_defaults")
 
+# Adding Samconfig prefix to the formatter here to show explicitly that the log messages generated
+# are exclusively from the configuration file context.
+
 LOG = logging.getLogger(__name__)
+SamCliLogger.configure_logger(LOG, logging.Formatter("Samconfig: %(message)s"), logging.INFO)
```
c2tarun, line 26: What does this do? Modify log format for every log or just of this file?

c2tarun: Ohhh... never mind it is accept LOG object, so probably just for this file.

sriram-mv: There is no logger configured at this point, because the configuration file decorator is called in before https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/cli/main.py#L97

PR #1897, opened 2020-03-26
fix: split BuildError - showcase deeper exception

sriram-mv:
Why is this change necessary?

* Telemetry does not capture correct level of detail today.

* Show the deeper cause of Exception and emit to Telemetry.

* None

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -12,6 +12,7 @@ from samcli.commands._utils.options import (
     parameter_override_option,
 )
 from samcli.cli.main import pass_context, common_options as cli_framework_options, aws_creds_options
+from samcli.lib.build.app_builder import BuildInsideContanerError
 from samcli.lib.telemetry.metrics import track_command
 from samcli.cli.cli_config_file import configuration_option, TomlProvider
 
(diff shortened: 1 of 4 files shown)


samcli/commands/build/command.py
```diff
@@ -217,13 +218,19 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
...
+                UnsupportedBuilderLibraryVersionError,
+                ContainerBuildNotSupported,
         ) as ex:
             click.secho("\nBuild Failed", fg="red")
-            raise UserException(str(ex), wrapped_from=ex.__class__.__name__)
+
+            # Some Exceptions have a deeper wrapped exception that needs to be surfaced
+            # from deeper than just one level down.
+            deep_wrap = getattr(ex, "wrapped_from", None)
+            wrapped_from = deep_wrap if deep_wrap else ex.__class__.__name__
```
c2tarun: What are other types of value (except class names) which can be used for wrapped_from?

sriram-mv: technically anything, we can formalize on a specific class maybe if it helps in the future.

Issue #1825, opened 2020-02-27
sam build fails with go+python template due to --use-container requirement for python


tebruno99:
I have a multi-developer project that has several lambda functions. Some are written in Go and some are in Python. The python functions require a few native compiled libraries.

When building natively in codepipeline/travis/and on linux desktop using sam build works great since no -u (--use-container) is required to do the build. However, the Python developers all use Windows or OSX.  They would like to use sam build -u to test their code before pushing up and triggering the C.I. pipelines.

```
sam build -u
Starting Build inside a container
Building resource 'CrmImportHandlerFunction'

Build Failed
Error: We do not support building Go Lambda functions within a container. Try building without the container. Most Go functions will build successfully.
```
The Go functions don't really need to be run by these particular developers but they do need to use `sam build -u` to build their own projects. It would be good if we could specify exactly which function we want to build so these users can build and test their functions.

Splitting the project is also a possibility, but it was super great to have these functions that share a lot of resources in the same template.

Possible suggestions:
.) Make it so you can specify to each run-time whether to use a container or not.
.) Have `sam build -u` build with containers when supported, but NOT use a container for unsupported function. This is best imo because everything would be allowed to succeed.

sriram-mv:
``
It would be good if we could specify exactly which function we want to build so these users can build and test their functions.
``

Usage: sam build [OPTIONS] [FUNCTION_IDENTIFIER]

You can specify just a function name.

tebruno99:
Wow, sorry I must have had an older version on that machine or something. Thanks for pointing that out.

tebruno99:
@sriram-mv That was a great tip, we have since found though that deploying with sam cli is still not possible.

Since sam build always clears the .aws-sam/build folder there isn't a way for a windows or osx user to end up with all the functions built for a deploy.
Your suggestion works great for single function invokes but it would be nice for these developers to be able to deploy the whole build.

stevenj:
I have the same problem, but its because I am using Python 3.8 and my linux box is on 3.7.
I was using the -u because that handled the mismatch in python version.  But when i just added my go function, i ran into the same issue.
My solution is to create a python virtual environment with python 3.8.  I can then run sam build and it will build both the go code and the python code for deployment.

I know this isn't a perfect solution because if you use python libraries that have binaries they may not run properly if built outside the container.

jfuss:
[...]

PR #1898, opened 2020-03-27
Allow building lambda functions which do not support containers, even if -u is specified.

stevenj:
*Issue #, if available:*
#1825

If you have a project that mixes lambda functions with the need to build inside a container (say a python project with a native library) and say a Go function which is not supported by container build.  Then the build fails, and there is no effective workaround.

I simply check if the runtime attempting to be built supports building inside a container, BEFORE i change the build to be a containered build. Currently the code checks after and raises an exception which terminates the build.

Builds which are currently broken will build properly.  No other side effects.

*Checklist:*

- [X ] Write unit tests
- [? ] Write/update functional tests
- [X ] Write/update integration tests
- [X ] `make pr` passes
- [X ] Write documentation

I didn't write any functional tests, because I didn't see any functional tests which exercise container builds.

### DIFF
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -81,7 +81,9 @@ $ sam build && sam package --s3-bucket <bucketname>
     "-u",
     is_flag=True,
     help="If your functions depend on packages that have natively compiled dependencies, use this flag "
-    "to build your function inside an AWS Lambda-like Docker container",
+    "to build your function inside an AWS Lambda-like Docker container.  IFF the runtime does not support "
+    "building inside a container, a warning will be logged, and the function will attempt to be built "
+    "outside the container.",
 )
 @click.option(
     "--manifest",
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -186,7 +186,11 @@ class ApplicationBuilder:
             # By default prefer to build in-process for speed
             build_method = self._build_function_in_process
             if self._container_manager:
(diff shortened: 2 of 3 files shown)


stevenj:
Both integration tests windows-build-dotnet and windows-build-go are now failing because they now no longer fail to build code if building in a container is specified.  I can't find those integration tests in the repo, but i am happy to fix them.

sriram-mv:
@stevenj You can directly specify that function you want to build in sam build.

```
Usage: samdev build [OPTIONS] [FUNCTION_IDENTIFIER]

  Use this command to build your AWS Lambda Functions source code to
  generate artifacts that target AWS Lambda's execution environment.
```

samdev build abc should work.

stevenj:
That does not provide a work around to the issue.  As @tebruno99 said in the linked issue:

> Since sam build always clears the .aws-sam/build folder there isn't a way for a windows or osx user to end up with all the functions built for a deploy.
Your suggestion works great for single local function invokes but it would be nice for these developers to be able to deploy more than 1 python lambda at a time (the others are missing from the build package since sam build -u deletes the previous builds and sam build -u fails on the Go builds.)

> (Also the resulting template only has the correct codeuri for the function build that just ran)

jfuss:
@stevenj Could you please add unit tests to cover the cases you are adding and handle the merge conflict with HEAD of develop?

stevenj:
I am not using aws-sam anymore,  i will try and add the test cases when time permits.

Issue #954, opened 2019-01-23
After I deploy Lambda Function Resource of Go1.x with SAM and Windows 10 onto AWS, it occurs an error.


MenEnger:
### Description

I cannot run a lambda function of Go1.x on AWS console.

### Steps to reproduce

Prepare Windows10 Pro.
Setup  Python3.7, AWS-CLI, AWS-SAM-CLI, Go1.11.4.
Configure AWS API key etc....
Do "sam init -r go"
Create S3 bucket to store package on AWS.
Build an executable file with "GOOS=linux" .
Do "sam package" and "sam deploy" commands.
Test the function on AWS Lambda console with "hello-world"-based event.

### Observed result
```
{
  "errorMessage": "fork/exec /var/task/hello-world: permission denied",
  "errorType": "PathError"
}
```

### Expected result

Show IP adress of "https://checkip.amazonaws.com"

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS:Windows 10 Pro
2. `sam --version`:0.9.0

`Add --debug flag to command you are running`

### Cause to guess
I suspect that SAM-CLI does not use "aws-lambda-go/cmd/build-lambda-zip" when zipping executables of Golang.
Zip in Python module doesn't change the attribute like `chmod 777`, does it?

jfuss:
SAM CLI does not use that go package and shouldn't for zipping. Instead we should be updating the external_attr attribute on the zipfile.ZipInfo before zipping.

spaceraccoon:
@jfuss Are there plans to merge the fix for `external_attr`? `sam build` on the default `hello-world` for `go` still creates a lambda function with the error:

```
{
  "errorMessage": "fork/exec /var/task/hello-world: permission denied",
  "errorType": "PathError"
}
```

Thanks!

jfuss:
@spaceraccoon There currently isn't a PR for this and not aware of anyone directly working on it at the moment. It's something we should defiantly fix. Are you willing to try and implement it?

spaceraccoon:
Started a PR at #1986.

PR #1912, opened 2020-04-05
Add go_zip_and_upload function to us aws-lambda-go/cmd/build-lambda-zip

hottestchilipepper:
*Issue #, if available:*

#954 

Because on Windows occurs 'permission denied', 'PathError' in invoking lambda functions, if we us aws-sam-cli( sam deploy ).
You can see the problem in #954 .

According to README.md of [aws-lambda-go](https://github.com/aws/aws-lambda-go), on Windows we have to do some special things. We have to use 'build-lambda-zip'.
So i added code which does it.

Yet, if we build .go file with 'sam build', there will be only one output file. 'build-lambda-zip' of aws-lambda-go can make a zip file for only one file. So, if the number of output files by 'sam build' is more than one, 'sam deploy' will fail.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -25,11 +25,11 @@ import pathlib
 import contextlib
 from contextlib import contextmanager
 import uuid
-from samcli.lib.utils.osutils import is_windows, popen
 from urllib.parse import urlparse, parse_qs
 import shutil
 from botocore.utils import set_value_from_jmespath
 import jmespath
+from samcli.lib.utils.osutils import popen, is_windows
 
 from samcli.commands._utils.resources import (
     AWS_SERVERLESSREPO_APPLICATION,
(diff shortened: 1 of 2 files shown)


hottestchilipepper:
Users should install Go and execute 'go.exe get -u github.com/aws/aws-lambda-go/cmd/build-lambda-zip' before 'sam deploy'

jfuss:
@hottestchilipepper Thanks for the PR. While this solves the initial issue, we cannot force someone to install another tool like this and shell out to it. We do this in `sam build` so we can support building with language native tools but we should not be doing this in `package`. Instead we should be updating the `external_attr` attribute on the `zipfile.ZipInfo` before zipping.

I am going to close this but feel free to submit a new PR that does this through setting the `external_attr` on `zipfile.ZipInfo`.

hottestchilipepper:
@jfuss Thank you for your guide. I will do that

PR #1917, opened 2020-04-10
feat: Allow pinned requirements for linux installs from source

jfuss:
*Issue #, if available:*
N/A

This will enable installing pinned versions of all dependencies of SAM CLI when installing from source.

Added a make target to create a new requirements file called `reproducible-linux.txt`. This will allow someone to download the source, pip install this requirements file and then the CLI. 

None that I know of.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
hmm.. the `requirements/reproducible-linux.txt` is empty. Something is messed up with my make target. Investigating what is going on.

jfuss:
Redirecting in the makefile wasn't working like I expected it too. So moved to use the `-o` option on `pip-compile` command.

Issue #1920, opened 2020-04-13
Bootstrap doesn't support --role_arn


jplock:
### Description

Running `sam deploy --guided` the first time to create the S3 bucket via a CloudFormation template, the `--role_arn` parameter isn't supported in https://github.com/awslabs/aws-sam-cli/blob/master/samcli/lib/bootstrap/bootstrap.py

### Steps to reproduce

Run `sam deploy --guided` and it fails because we have policies that require all CloudFormation stacks to use a specific role.

### Observed result

Please provide command output with `--debug` flag set.

### Expected result

I'm able to run `sam deploy --guided --role_arn <my_role_arn>` and SAM will provision it's CloudFormation stack with the role I specified.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac
2. `sam --version`:  SAM CLI, version 0.47.0

Would you accept a pull request adding this?

sriram-mv:
sam deploy does support `--role-arn`

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-deploy.html

jplock:
@sriram-mv correct, but it needs to bootstrap the S3 bucket, the role_arn does not get used.

sriram-mv:
Makes sense, in the meantime a workaround is setup the role_arn in a aws profile and pass in that profile during guided.

sam deploy -g --profile role_profile

jplock:
@sriram-mv the CloudFormation role we are using isn't assume-able by users, it's a service role, so I don't think what you proposed would work

sriram-mv:
@jplock Makes sense. Thanks for clarifying.

jplock:
@sriram-mv thoughts on including https://github.com/awslabs/aws-sam-cli/pull/1933 in an upcoming release?  It would simplify our getting started activities

PR #1933, opened 2020-04-22
bug: Support --role_arn during guided bootstrap

jplock:
*Issue #1920 *

To allow the guided bootstrap S3 bucket creation to use a pre-defined CloudFormation IAM role.

`sam deploy --guided --role_arn=<role>` now works as expected

None

No

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/bootstrap/command.py
```diff
@@ -22,9 +22,9 @@
 @pass_context
 @track_command
 def cli(ctx):
-    do_cli(ctx.region, ctx.profile)  # pragma: no cover
+    do_cli(ctx.region, ctx.profile, ctx.role_arn)  # pragma: no cover
```
jfuss, line 25: `role_arn` isn't apart of the context that I know of. You will need to add an `@option` annoation, you look at other commands to see how this is done.

samcli/commands/bootstrap/command.py
```diff
@@ -22,9 +22,9 @@
...
 
 
-def do_cli(region, profile):
-    bucket_name = bootstrap.manage_stack(profile=profile, region=region)
+def do_cli(region, profile, role_arn=None):
```
jfuss, line 28: Remove `=None`

samcli/lib/bootstrap/bootstrap.py
```diff
@@ -83,16 +83,21 @@ def _create_or_get_stack(cloudformation_client):
...
+
+    params = {
+        "StackName": SAM_CLI_STACK_NAME,
+        "TemplateBody": _get_stack_template(),
+        "Tags": [{"Key": "ManagedStackSource", "Value": "AwsSamCli"}],
+        "ChangeSetType": "CREATE",
+        "ChangeSetName": change_set_name,  # this must be unique for the stack, but we only create so that's fine
+    }
+    if role_arn:
+        params["RoleARN"] = role_arn
```
jfuss, line 98: Curious, is setting `RoleARN` to None not work? I would hope that Boto3 or the CloudFormation APIs handle this already, which would allow us to just set it no matter what and remove the need for a if check.

jplock: Yea it didn’t work with a value of None

Issue #1934, opened 2020-04-23
Runtime deprecation on May 1st 2020


jfuss:
**SAM CLI will stop supporting dotnetcore2.0, nodejs4.3, nodejs6.10, and nodejs8.10 runtimes as of May 1st, 2020.**

Runtimes Nodejs4.3, Nodejs6.10, Nodejs8.10, and .NetCore 2.0 are no longer supported by AWS Lambda. SAM CLI still currently supports these runtimes in `sam init`, `sam local invoke|start-api|start-lambda` and `sam build`. 

**What is the deprecation schedule?**
In an upcoming release after May 1st, SAM CLI will no longer allow customers use these depreciated runtimes within the CLI. 

**Am I impacted by this deprecation?**
You are impacted by this if you are still using one of the following runtimes: nodejs4.3, nodejs6.10, nodejs8.10, or dotnetcore2.0. The recommended path forward is to upgrade your functions to a new AWS Lambda supported runtime. 

For Nodejs runtimes, you will need to upgrade to nodejs10.x or nodejs12.x.
For .NetCore, you will need to upgrade to dotnetcore2.1 or dotnetcore3.1.

jfuss:
Was released in https://github.com/awslabs/aws-sam-cli/releases/tag/v0.49.0

Closing

PR #1935, opened 2020-04-23
feat: Print warning about using depreciated runtimes

jfuss:
*Issue #, if available:*
#1934

This is to add warnings for customers using build, invoke|start-api|start-lambda, and init commands with runtimes that are no longer supported by AWS Lambda.

This is the first phase, so we can safely remove these depreciated runtimes in a future release.

None, just prints warnings when a customer uses a depreciated runtime.

N/A

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -9,8 +9,7 @@ from json import JSONDecodeError
 import click
 
 from samcli.cli.cli_config_file import configuration_option, TomlProvider
-from samcli.commands.exceptions import UserException
-from samcli.cli.main import pass_context, common_options, global_cfg
+from samcli.cli.main import pass_context, common_options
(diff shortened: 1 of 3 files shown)


samcli/commands/init/__init__.py
```diff
@@ -176,6 +177,19 @@ def do_cli(
         do_interactive(location, runtime, dependency_manager, output_dir, name, app_template, no_input)
 
 
+def _deprecate_notification(runtime):
+    from samcli.lib.utils.colors import Colored
+
+    depreciated_runtimes = {"dotnetcore1.0", "dotnetcore2.0"}
```
awood45, line 183: Nitpick: Spelling s/depreciated/deprecated

PR #1936, opened 2020-04-23
Build for Layers

c2tarun:
*Issue #, if available:*
Adding ability to build layers. If a user want to build a layer then simply add following:
```
Metadata:
  BuildMethod: <runtime>
```
to the layer in the template. This will pick the build method for that runtime (as it does for Lambda Function). After that Layer can be build by one of the following two commands:
`sam build`
`sam build LayerLogicalId`

*Checklist:*

- [x] Write unit tests
- [ ] Write/update integration tests *(WIP...)*
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -9,23 +9,22 @@ import pathlib
 
 from samcli.local.docker.manager import ContainerManager
 from samcli.lib.providers.sam_function_provider import SamFunctionProvider
+from samcli.lib.providers.sam_layer_provider import SamLayerProvider
 from samcli.commands._utils.template import get_template_data
-from samcli.commands.exceptions import UserException
-from samcli.local.lambdafn.exceptions import FunctionNotFound
-from samcli.commands.build.exceptions import InvalidBuildDirException
+from samcli.local.lambdafn.exceptions import ResourceNotFound
+from samcli.commands.build.exceptions import InvalidBuildDirException, MissingBuildMethodException
 
 LOG = logging.getLogger(__name__)
 
 
 class BuildContext:
-
     # Build directories need not be world writable.
     # This is usually a optimal permission for directories
     _BUILD_DIR_PERMISSIONS = 0o755
 
     def __init__(
         self,
-        function_identifier,
+        resource_identifier,
         template_file,
         base_dir,
         build_dir,
@@ -38,7 +37,7 @@ class BuildContext:
         skip_pull_image=False,
     ):
 
-        self._function_identifier = function_identifier
+        self._resource_identifier = resource_identifier
         self._template_file = template_file
         self._base_dir = base_dir
         self._build_dir = build_dir
@@ -51,6 +50,7 @@ class BuildContext:
         self._mode = mode
 
         self._function_provider = None
+        self._layer_provider = None
         self._template_dict = None
         self._app_builder = None
         self._container_manager = None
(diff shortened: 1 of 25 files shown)


samcli/lib/build/workflow_config.py
```diff
@@ -74,6 +72,29 @@ class UnsupportedRuntimeException(Exception):
     pass
 
 
+def get_layer_subfolder(runtime):
+    subfolders_by_runtime = {
```
sriram-mv, line 76: how did we arrive at this?

c2tarun: you mean subfolders? I simply copied from here: https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path

samcli/lib/build/app_builder.py
```diff
@@ -151,14 +160,46 @@ def update_template(self, template_dict, original_template_path, built_artifacts
...
+        # Code is always relative to the given base directory.
+        code_dir = str(pathlib.Path(self._base_dir, codeuri).resolve())
+
+        config = get_workflow_config(runtime, code_dir, self._base_dir)
+        subfolder = get_layer_subfolder(runtime)
+
+        # artifacts directory will be created by the builder
+        artifacts_dir = str(pathlib.Path(self._build_dir, layer_name, subfolder))
+
+        with osutils.mkdir_temp() as scratch_dir:
```
sriram-mv, line 185: something to check that this works as well on windows as well. I know we have a utility for platform independent temp file, but thats for temp files.

samcli/lib/providers/sam_base_provider.py
```diff
@@ -17,6 +17,90 @@ class SamBaseProvider:
     Base class for SAM Template providers
     """
 
+    SERVERLESS_FUNCTION = "AWS::Serverless::Function"
```
sriram-mv, line 20: This is just refactoring and moving stuff around, am I right?

samcli/commands/build/build_context.py
```diff
@@ -137,18 +143,71 @@ def mode(self):
...
+        Returns
+        -------
+
+        """
+        layer = self._layer_provider.get(resource_identifier)
+        if not layer:
+            # No layer found
+            return
+        if layer and layer.build_method is None:
+            LOG.error("Layer %s is missing BuildMethod Metadata.", self._function_provider)
```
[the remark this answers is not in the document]
c2tarun: Layers can choose not to have build method, which is correct. However, if there is no build method, we do want to fail on `sam build MyLayer` command. Current there is no build for Layer, so this is a completely new feature, without any need to maintain backwards compatibility.

samcli/lib/build/app_builder.py
```diff
@@ -106,13 +108,20 @@ def build(self):
...
-                                                                lambda_function.handler)
+        for function in self._resources_to_build.functions:
+            LOG.info("Building function '%s'", function.name)
+            result[function.name] = self._build_function(function.name,
+                                                         function.codeuri,
+                                                         function.runtime,
+                                                         function.handler)
+        for layer in self._resources_to_build.layers:
+            LOG.info("Building layer '%s'", layer.name)
+            if layer.build_method is None:
```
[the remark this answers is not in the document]
c2tarun: Explained above, today there is no build for layer. So failing on `sam build MyLayer` which fails today, won't break any backwards compatibility.

samcli/lib/providers/sam_base_provider.py
c2tarun: Yup.

samcli/lib/build/app_builder.py
c2tarun: I'll validate on windows, didn't do that.

c2tarun: Validated on windows: https://pastebin.com/Sj6ZwbZr

PR #1943, opened 2020-04-26
fix: showcase prompts and messages on deploying - auth

sriram-mv:
Why is this change necessary?

* If Auth is determined to be not present with a degree of possibility,
a prompt is shown to ask the user if they are explicitly okay with such
a configuration.

* Showcase to the user that there may be a potential issue with how the
template resources have been configured.

* Extra prompt on guided mode.

*Issue #, if available:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/auth_utils.py
+++ b/samcli/commands/deploy/auth_utils.py
@@ -1,0 +1,140 @@
+"""
+Utilities for checking authorization of certain resource types
+"""
+import logging
+
+from samcli.commands.local.lib.swagger.reader import SwaggerReader
+from samcli.lib.providers.sam_function_provider import SamFunctionProvider
+
+LOG = logging.getLogger(__name__)
+
+
+def auth_per_resource(parameter_overrides, template_dict):
+    """
+    Check if authentication has been set for the function resources defined in the template that have `Api` Event type.
+
+    Parameters
+    ----------
+    parameter_overrides: dict
+        list of parameter overrides for the parameters defined in the template
+    template_dict: dict
+        Raw dictionary of the defined SAM template
+
+    Returns
+    -------
+
+    List of tuples per function resource that have the `Api` or `HttpApi` event types, that describes the resource name
+    and if authorization is required per resource.
+
+    """
+
+    _auth_per_resource = []
+
+    sam_functions = SamFunctionProvider(
+        template_dict=template_dict, parameter_overrides=parameter_overrides, ignore_code_extraction_warnings=True
+    )
+    for sam_function in sam_functions.get_all():
+        # Only check for auth if there are function events defined.
+        if sam_function.events:
+            _auth_resource_event(sam_functions, sam_function, _auth_per_resource)
+
+    return _auth_per_resource
+
+
+def _auth_resource_event(sam_functions, sam_function, auth_resource_list):
+    """
+
+    Parameters
+    ----------
+    sam_functions: List of all functions with intrinscis resolved.
+    sam_function: Current function which has all intrinsics resolved.
+    auth_resource_list: List of tuples with function name and auth. eg: [("Name", True)]
+
+    Returns
+    -------
+
+    """
+    for event in sam_function.events.values():
+        for event_type, identifier in [("Api", "RestApiId"), ("HttpApi", "ApiId")]:
+            if event.get("Type") == event_type:
+                # Is there any auth defined directly on the function resource?
+                # NOTE(sriram-mv): How do we check this in more detail?
+                # Currently just checking for presence of Auth, not the details.
+                # https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-apifunctionauth.html
+                # https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-httpapifunctionauth.html
+                if event.get("Properties", {}).get("Auth", False):
+                    auth_resource_list.append((sam_function.name, True))
+                # Is there any auth defined on the referred http api or serverless api through the `id` construct?
+                elif _auth_id(sam_functions.resources, event.get("Properties", {}), identifier):
+                    auth_resource_list.append((sam_function.name, True))
+                else:
+                    auth_resource_list.append((sam_function.name, False))
+
+
+def _auth_id(resources_dict, event_properties, identifier):
+    """
+
+    Parameters
+    ----------
+    resources_dict: dict
+        Resolved resources defined in the SAM Template
+    event_properties: dict
+        Properties of given event supplied to a function resource
+    identifier: str
+        Id: `ApiId` or `RestApiId`
+
+    Returns
+    -------
+    bool
+        Returns if the given identifier under the event properties maps to a resource and has authorization enabled.
+
+    """
+    resource_name = event_properties.get(identifier, "")
+    api_resource = resources_dict.get(resource_name, {})
+    return any(
+        [
+            api_resource.get("Properties", {}).get("Auth", False),
+            _auth_definition_body_and_uri(
+                definition_body=api_resource.get("Properties", {}).get("DefinitionBody", {}),
+                definition_uri=api_resource.get("Properties", {}).get("DefinitionUri", None),
+            ),
+        ]
+    )
+
+
+def _auth_definition_body_and_uri(definition_body, definition_uri):
+    """
+
+    Parameters
+    ----------
+    definition_body: dict
+        inline definition body defined in the template
+    definition_uri: string
+        Either an s3 url or a local path to a definition uri
+
+    Returns
+    -------
+    bool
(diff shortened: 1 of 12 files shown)


samcli/commands/deploy/auth_utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns
+    -------
+    bool
+        Is security defined on the swagger or not?
+
+
+    """
+
+    reader = SwaggerReader(definition_body=definition_body, definition_uri=definition_uri)
```
c2tarun, line 120: What happens if Swagger definition have an error? Is it possible? If yes will it throw an error which might sound like it is due to missing auth or something?

samcli/commands/deploy/auth_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    # NOTE(sriram-mv): Authorization and Authentication is indicated by the `security` scheme.
+    # https://swagger.io/docs/specification/authentication/
+    for _, verb in swagger.get("paths", {}).items():
+        for _property in verb.values():
+            _auths.append(bool(_property.get("security", False)))
+
+    _auths.append(bool(swagger.get("security", False)))
+
+    # This is not an exhaustive check, but to check if there is some form of security setup.
+    return any(_auths)
```
c2tarun, line 134: just a suggestion, can we have list of check methods which all take same parameter and return true of false after validating a type of auth. It might be easy to extend by simply adding more auth methods?

samcli/commands/deploy/guided_context.py
```diff
@@ -114,6 +117,18 @@ def guided_prompts(self, parameter_override_keys):
         self.save_to_config = save_to_config
         self.confirm_changeset = confirm_changeset
 
+    def prompt_authorization(self, parameter_overrides):
+        auth_required_per_resource = auth_per_resource(parameter_overrides, get_template_data(self.template_file))
+
+        for resource, authorization_required in auth_required_per_resource:
+            if not authorization_required:
+                auth_confirm = confirm(
+                    f"\t{self.start_bold}{resource} may not have authorization defined, Is this okay?{self.end_bold}",
```
c2tarun, line 126: if a user have samconfig.toml file with guided deploy configured and is being used in CI/CD, will this prompt break them?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -26,7 +26,7 @@ class SamFunctionProvider(FunctionProvider):
     _LAMBDA_LAYER = "AWS::Lambda::LayerVersion"
     _DEFAULT_CODEURI = "."
 
-    def __init__(self, template_dict, parameter_overrides=None):
+    def __init__(self, template_dict, parameter_overrides=None, ignore_code_extraction_warnings=False):
```
c2tarun, line 29: what is `code extraction warning`?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -40,15 +40,17 @@ def __init__(self, template_dict, parameter_overrides=None):
         :param dict template_dict: SAM Template as a dictionary
         :param dict parameter_overrides: Optional dictionary of values for SAM template parameters that might want
             to get substituted within the template
+        :param bool ignore_code_extraction_warnings: Ignores Log warnings
         """
 
         self.template_dict = SamBaseProvider.get_template(template_dict, parameter_overrides)
+        self.ignore_code_extraction_warnings = ignore_code_extraction_warnings
```
c2tarun, line 47: why save something in class variable when all we have is static methods and we keep passing them to every method? *Hint: Lets change static methods to class methods and use class variable* :)

tests/unit/commands/deploy/test_command.py
```diff
@@ -99,15 +100,83 @@ def test_all_args(self, mock_deploy_context, mock_deploy_click, mock_package_con
...
+    @patch("samcli.commands.deploy.command.click")
+    @patch("samcli.commands.deploy.deploy_context.DeployContext")
+    @patch("samcli.commands.deploy.command.print_deploy_args")
+    @patch("samcli.commands.deploy.guided_context.manage_stack")
+    @patch("samcli.commands.deploy.guided_context.auth_per_resource")
+    @patch("samcli.commands.deploy.guided_context.get_template_parameters")
+    @patch("samcli.commands.deploy.guided_context.get_template_data")
     @patch.object(GuidedConfig, "get_config_ctx", MagicMock(return_value=(None, get_mock_sam_config())))
     @patch("samcli.commands.deploy.guided_context.prompt")
     @patch("samcli.commands.deploy.guided_context.confirm")
```
c2tarun, line 172: That is lot of patching.

samcli/lib/providers/sam_function_provider.py
sriram-mv: Yeah, i thought about this, this a dirty way to do this without changing a lot of code.

samcli/lib/providers/sam_function_provider.py
sriram-mv: on a processed template. the code uri has changed to become s3 url, on deploy we actually pass in the packaged template. and this prompt kept appearing otherwise.

samcli/commands/deploy/guided_context.py
sriram-mv: we dont save this value into samconfig

samcli/commands/deploy/auth_utils.py
sriram-mv: agreed with a series of handler methods. or just a series of filters.

samcli/commands/deploy/auth_utils.py
sriram-mv: just returns None, which we handle.

samcli/commands/deploy/auth_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    -------
+
+    List of tuples per function resource that have the `Api` or `HttpApi` event types, that describes the resource name
+    and if authorization is required per resource.
+
+    """
+
+    _auth_per_resource = []
+
+    sam_functions = SamFunctionProvider(
```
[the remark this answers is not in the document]
sriram-mv: yeah template_dict is passed around during build as well. this is post transfomation phase with the template though

samcli/lib/providers/sam_function_provider.py
jfuss: @c2tarun This is historically how we have done things and the reason we did it this was was for testing. We test a lot of the private methods through out the codebase, and to make some of that easier we have `@staticmethod` on methods so we can pass all the values (expected and not expected) into the method. This allows us deeper control and testing over different parts of the codebase. We can change it but not sure the value out weighs the effort in doing this.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -177,12 +196,13 @@ def _extract_sam_function_codeuri(name, resource_properties, code_property_key):
         # CodeUri can be a dictionary of S3 Bucket/Key or a S3 URI, neither of which are supported
         if isinstance(codeuri, dict) or (isinstance(codeuri, str) and codeuri.startswith("s3://")):
             codeuri = SamFunctionProvider._DEFAULT_CODEURI
-            LOG.warning(
-                "Lambda function '%s' has specified S3 location for CodeUri which is unsupported. "
-                "Using default value of '%s' instead",
-                name,
-                codeuri,
-            )
+            if not ignore_code_extraction_warnings:
```
jfuss, line 199: Why are we changing this warning?

samcli/commands/deploy/guided_context.py
sriram-mv: no, its only on guided.

samcli/lib/providers/sam_function_provider.py
sriram-mv: since we pass in a packaged template from the deploy context where the code uri has been changed to an s3 uri. Without that ignore code extractions, it will throw a message saying code uri with s3 is not supported.

samcli/commands/deploy/auth_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        swagger = {}
+    # NOTE(sriram-mv): Authorization and Authentication is indicated by the `security` scheme.
+    # https://swagger.io/docs/specification/authentication/
+    for _, verb in swagger.get("paths", {}).items():
+        for _property in verb.values():
+            _auths.append(bool(_property.get("security", False)))
+
+    _auths.append(bool(swagger.get("security", False)))
+
+    # This is not an exhaustive check, but to check if there is some form of security setup.
```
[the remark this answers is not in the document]
sriram-mv: being exhaustive is a whole different ball game, I agree that we can log that the check is not exhaustive.

PR #1957, opened 2020-04-30
feat: allow a custom builder workflow selection

sriram-mv:
Why is this change necessary?

* Build for provided runtimes and for existing runtimes to have their
own build process.

* Alternative build mechanisms for those use-cases where existing build
mechanisms don't solve the use-case.

* Function Metadata is updated as part of function properties, for easy
wiring. Have not seen side-effects in current testing yet.

*Issue #, if available:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/build/workflow_config.py
```diff
@@ -97,6 +106,11 @@ def get_workflow_config(runtime, code_dir, project_dir):
     namedtuple(Capability)
         namedtuple that represents the Builder Workflow Config
     """
+    specified_build_workflow = metadata.get("BuildMethod", None) if metadata else None
```
c2tarun, line 109: I understand why you are using `BuildMethod` to avoid conflict with existing runtime. However, I though the initial idea in [this](https://github.com/awslabs/aws-lambda-builders/pull/166/files#diff-cf7e02516f5f279dc897b4a2243e2d3e) design doc was to simply check if `makefile` exist with `build-{logical-id}` then use it or fallback to current runtime. This might confuse users.

samcli/lib/build/workflow_config.py
```diff
@@ -69,12 +69,21 @@
...
+
 
 class UnsupportedRuntimeException(Exception):
     pass
 
+class UnsupportedBuilderException(Exception):
+    pass
 
-def get_workflow_config(runtime, code_dir, project_dir):
+def get_workflow_config(runtime, code_dir, project_dir, metadata=None):
```
c2tarun, line 86: Also if you decide to use `BuildMethod` anyway, I don't think it is good idea to pass metadata all the way to this level. We have `build_function` method that does translate `codeuri` -> `code_dir` which `get_workflow_config` method uses. Now if you pass raw metadata, then get_workflow_config now have the logic to do that transformation to runtime. Now this cleaner function has convoluted logic of choosing between two runtimes. I would suggest two approaches:
1. Do this conversion in build_function method and pass the correct runtime in this method based on your logic.
2. Abstract out these conversions to Function class so that it is responsible to provide a local build_runtime based on all parameters.

I like approach one better because it is easier to do, and keeps all build related logic at one place. Probably abstract out whole thing in a _runtime_resolver method.

sriram-mv: yeah moving selector process to its own method was what I had in mind.

samcli/lib/build/workflow_config.py
sriram-mv: fair call out, the doc needs to be updated. But the change is not backwards incompatible and can be covered in docs to showcase its usage.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -118,6 +118,10 @@ def _extract_functions(resources, ignore_code_extraction_warnings=False):
 
             resource_type = resource.get("Type")
             resource_properties = resource.get("Properties", {})
+            resource_metadata = resource.get("Metadata", None)
+            # Add extra metadata information to properties under a separate field.
+            if resource_metadata:
+                resource_properties["Function_Metadata"] = resource_metadata
```
sriram-mv, line 124: Need to change this to just metadata

Issue #867, opened 2018-12-18
Sam Package strips quotes from account Ids


rhboyd:
### Description

When I use the following mapping to provide configurations based on the AWS::AccountId
```
Mappings:
  "FOO":
    "0123456789":
      KEY: VALUE
```

after running `sam package --template-file /PATH/TO/template.yaml --s3-bucket`, the quotes are stripped from the account ID in the cfn template that is generated. This only happens when the accountID starts with a 0 and does not contain any "8"s or "9"s

### Steps to reproduce

-create a mapping that has keys which start with a 0 and also contain an 8 or 9 (i.e. not valid Octal)
-execute `sam package`
-inspect output template

### Observed result
```
2018-12-18 09:14:09 package command is called
... cfn template ...
2018-12-18 09:14:11 package command successful
```

### Expected result

I expected the quotes around string literals to remain in place

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac Mojave
2. `sam --version`: SAM CLI, version 0.8.1

rhboyd:
Sorry for the spam. I misread the results when I tried it in 0.9.0, the problem exists in the new release as well.

jfuss:
This is an issue with PyYaml: https://github.com/yaml/pyyaml/issues/98 only supporting 1.1. 

Note: AWS CloudFormation only supports [YAML 1.1](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-formats.html). We need to remain compatible with what AWS CloudFormation supports and therefore their isn't much we can do here.

Closing

ecmonsen:
So is this saying that AWS SAM is producing templates that are incompatible with CloudFormation, and this won't be fixed? 

My issue is that `aws cloudformation create-stack` succeeds with the following template, but `sam deploy` fails because it strips the quotes from "011111111118" (but not the other keys in the mapping) before feeding it to CloudFormation.

```
AWSTemplateFormatVersion: "2010-09-09"

Mappings:
  AdminRoleArns:
    "111111111118":
      Arn: arn:aws:iam::111111111111:role/accounts-AWSIAMAdminRole-1111111111118
    "011111111111":
      Arn: arn:aws:iam::111111111111:role/accounts-AWSIAMAdminRole-1111111111111
    "011111111118":
      Arn: arn:aws:iam::011111111118:role/accounts-AWSIAMAdminRole-0111111111118

Resources:
  KmsKey:
[...]

PR #1965, opened 2020-05-06
fix: Force strings to be outputted when string starts with 0

jfuss:
*Issue #, if available:*
#867

When build outputs a template, the yaml dumper will omit the quotes when the number could not parse as a valid octal number. This will then lead to CFN stripping the 0 off the string (mostly accountids) and cause CFN stacks to fail.

We will force the yaml dumper to output a string if there is a scalar with that starts with 0. 

None that I know of

N/A

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/yamlhelper.py
```diff
@@ -23,6 +23,15 @@
 import yaml
 from yaml.resolver import ScalarNode, SequenceNode
 
+TAG_STR = "tag:yaml.org,2002:str"
+
+
+def string_representer(dumper, value):
```
sriram-mv, line 29: could you add a docstring here, just to explain what happens here.

jfuss: Good Callout, I meant to go back and completely forgot. Will get a new revision with that here soon.

Issue #1960, opened 2020-05-01
sam deploy hangs with SAM CLI version higher than 0.41.0


piersf:
### Description

We have noticed a strange behavior when upgrading SAM CLI to any version higher than `0.41.0`.
The `sam deploy` command hangs for a long time without doing/outputting anything.

Even at this point, we have a `sam deploy` process on a VM still running after 2 days. We haven't killed the process in order to be able to troubleshoot.

This happens(for us) with every version higher than `0.41.0`. With versions prior to `0.41.0`, we don't see this behavior.

We are running SAM CLI inside a Python virtual environment. At first, we thought this could be the version of Python inside the virtual environment(`3.7.4`). So we created a virtual environment with Python `3.7.7`. But the problem still remained and this did not make a difference.

The command we are running is:

```
/path/to/virtualenv/bin/python3.7 /path/to/virtualenv/bin/sam deploy --region REGION --profile PROFILE --template-file /path/to/packaged/template.yaml --stack-name STACK_NAME --s3-bucket S3_BUCKET_NAME --s3-prefix S3_BUCKET_PREFIX --capabilities CAPABILITY_NAMED_IAM --tags Name=value --parameter-overrides Parameter1=value1 Parameter2=value2 Parameter3=value3 ParameterN=valueN --no-fail-on-empty-changeset
```

Looking at the CloudFormation console while the `sam deploy` command is hanging, there's no stack update happening whatsoever.

We tried to add the `--debug` flag, but no additional output is being shown.

I want to point out that with versions prior to `0.41.0` the deployment works just fine. We don't experience the same behavior.

Is there something wrong with our command above? Are we missing some kind of flag that is required with versions higher than `0.41.0`?

Has anyone else experienced something similar? As a result of this, we're unable to go any higher than version `0.41.0`.

### Steps to reproduce

Upgrade to SAM CLI version > `0.41.0` and run `sam deploy`

### Observed result

`sam deploy` with SAM CLI version higher than `0.41.0` hangs

### Expected result

`sam deploy` with SAM CLI version higher than `0.41.0` should not hang

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: `Redhat 7, Debian`
2. `sam --version`: `0.42.0, 0.43.0, 0.44.0, 0.45.0, 0.46.0, 0.47.0, 0.48.0`
3. Python version: `3.7.4, 3.7.7`

Thank you in advance!

c2tarun:
[...]

Issue #1845, opened 2020-03-01
"sam deploy" does not execute if a tag has a long value


AaronChelvan:
**Description:**
The `sam deploy` command hangs if a tag with a long value (approx 30+ characters long) is provided. Tested on SAM CLI v0.43.0.

**Steps to reproduce the issue:**
1. Create an empty directory and navigate to it.
2. Execute `sam deploy --stack-name test --tags key=012345678901234567890123456789`

**Observed result:**
The command hangs and does not produce any output.

**Expected result:**
```
	Deploying with following values
	===============================
	Stack name                 : test
	Region                     : None
	Confirm changeset          : False
	Deployment s3 bucket       : None
	Capabilities               : null
	Parameter overrides        : {}

Initiating deployment
=====================
Error: Failed to package template: /tmpDir/template.yml. 
 [Errno 2] No such file or directory: '/tmpDir/template.yml'

```

However, if a shorter tag value is provided (e.g. `sam deploy --stack-name test --tags key=01234567890123456789`) then the correct output gets printed.

beck3905:
This appears to be caused by this regex that was added in version 0.42.0: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/cli/types.py#L207.

I isolated this code and tried it with a short tag and a long tag and it is the `re.findall` call that is hanging.  Not exactly sure what the regex pattern is meant to do as it is a rather long and complex pattern.  A little more commenting for that would be helpful.

dslove:
In my case there is no any error msg showing up. It took me a day to figure out what the problem is and it was because I was lucky to google and find this page... I am using 0.44.

iDVB:
I too am hitting this issue as well. 
I can provide a bit more testing cases.

Using the AWS CLI directly work 
```bash
aws cloudformation deploy \
  --template .aws-sam/build/packaged.yaml \
  --stack-name 'my-stack-name' \
  --capabilities CAPABILITY_NAMED_IAM \
  --tags \
      SERVICE=ggg-ggggggggg-ggggggg \
      GIT_BRANCH=master \
      GIT_REPOSITORY=gggggg-ggggggggg-ggg-gggggggggggggg
```

Using the same call with AWS SAM either does not work... or takes way too long
```bash
sam deploy \
 --template-file .aws-sam/build/packaged.yaml \
  --stack-name 'my-stack-name' \
  --capabilities CAPABILITY_NAMED_IAM \
--no-fail-on-empty-changeset \
[...]

PR #1968, opened 2020-05-06
Using split instead of regex for simple key=value type tags.

c2tarun:
*Issue #, if available:*
Issue #1960 and #1845 

This change will unblock lot of users who use long tag value to upgrade from 0.41.0.

For simple `key=value` type values we are not splitting on `=` instead of using regex.

None

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/cli/types.py
```diff
@@ -221,3 +226,23 @@ def convert(self, value, param, ctx):
                 )
 
         return result
+
+    @staticmethod
+    def _standard_key_value_parser(tag_value):
+        """
+        Method to parse simple `Key=Value` type tags without using regex. This is similar to how CFN does this.
```
sriram-mv, line 233: aws cli, not cfn.

samcli/cli/types.py
```diff
@@ -204,14 +204,19 @@ def convert(self, value, param, ctx):
...
-            groups = re.findall(self._pattern, val)
-
-            if not groups:
-                fail = True
-            for group in groups:
-                key, v = group
-                # assign to result['KeyName1'] = string and so on.
-                result[_unquote_wrapped_quotes(key)] = _unquote_wrapped_quotes(v)
+            # Using standard parser first. We should implement other type parser like JSON and Key=key,Value=val type format.
+            parsed, k, v = self._standard_key_value_parser(val)
```
sriram-mv, line 208: should we check for more than 1 count, given that aws cli already does it.

samcli/cli/types.py
c2tarun: ahhhh.. missed it. Will fix now.

samcli/cli/types.py
c2tarun: aws-cli simply pick first `=`. I think our existing solution already handles more edge cases than aws-cli, we should keep using it until we implement separate parses.

sriram-mv: right, but this solution only looks for one key value pair right?

c2tarun: Nope, we get list of `key=value` tags. We parse each one. They are already separated by command and handled by click.

piersf:
In which release will be this fix available?

c2tarun:
@piersf Code is merged, we haven't finalized exact date yet, but there are few more minor fixes we want to include as well. You can expect this release by early next week.

piersf:
- @c2tarun Thank you! I look forward to the next release.

PR #1986, opened 2020-05-14
fix: Update external_attr attribute on the zipfile.ZipInfo before zipping

spaceraccoon:
*Issue #, if available:* #954

When exporting artifacts on Windows, file permissions are not preserved. This causes binaries to lose their executable permission, causing errors when pushing the zips to Lambda.

For example, the `hello-world` Golang sample fails on Lambda with `fork/exec /var/task/hello-world: permission denied: PathError`.

As suggested by @jfuss at #1912, 

> we should be updating the external_attr attribute on the zipfile.ZipInfo before zipping.

Due to the complexities of determining the exact file permissions before zipping, it simply adds a `777` permission bit to all files in the upload.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/artifact_exporter.py
```diff
@@ -200,7 +200,11 @@ def make_zip(file_name, source_root):
                 for filename in files:
                     full_path = os.path.join(root, filename)
                     relative_path = os.path.relpath(full_path, source_root)
-                    zf.write(full_path, relative_path)
+                    with open(filename, 'rb') as data:
+                        file_bytes = data.read()
+                        info = zipfile.ZipInfo(relative_path)
+                        info.external_attr = 0o777 << 16L
```
jfuss, line 206: Can we scope this to only files that actually need it? We should not be mutating all file permissions in the zip, seems like an escalation of privileged.

spaceraccoon: I'll dive a bit deeper and check if we can extract the permission bits using python standard library and match it exactly. Failing that, I'll only try to match this for executable files.

jfuss: We should also scope this to only windows. From my understanding, this is really only a problem on windows because you can't set linux like permissions.

jiminygreen:
@spaceraccoon I'm so pleased to see that you have been working on this. I'm a windows developer and I've been running in circles for the last few days.

It looks like it is really close to being completed with the only failure on the 'make-pr'.

Is there anything else to do once that is addressed?

MaddoxDevelopment:
Any updates on this? Running into this issue as well.

sriram-mv:
#2193 is the latest PR on this. Progress can be tracked there.

Closing this PR.

PR #1991, opened 2020-05-17
Skip non-verb items when parsing swagger security

leobasilio:
*Issue #, if available:*

No issue found.

The OpenAPI Specification states that a [Path Item Object](https://swagger.io/specification/#path-item-object) may have entries that are not verbs.

When parsing an API description with such features, SAM CLI fails with the following error:

>   File "/home/linuxbrew/.linuxbrew/Cellar/aws-sam-cli/0.49.0/libexec/lib/python3.7/site-packages/samcli/commands/deploy/auth_utils.py", line 132, in _auth_definition_body_and_uri
>    _auths.append(bool(_property.get("security", False)))
>AttributeError: 'list' object has no attribute 'get'

I was using a "parameters" item, which is a list, and [the code](https://github.com/awslabs/aws-sam-cli/blob/23212ef827a105d23339712adc9422c6cc55e12c/samcli/commands/deploy/auth_utils.py#L132) expects a dict.

Given the current specification, only verbs are objects, so only verbs are parsed as a dict. It should be enough to check the variable type before trying to get the desired property.

None.

No changes.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/auth_utils.py
+++ b/samcli/commands/deploy/auth_utils.py
@@ -129,7 +129,11 @@ def _auth_definition_body_and_uri(definition_body, definition_uri):
     # https://swagger.io/docs/specification/authentication/
     for _, verb in swagger.get("paths", {}).items():
         for _property in verb.values():
-            _auths.append(bool(_property.get("security", False)))
+            # A path object may have entries that are not verbs.
+            # We should first check whether _property is a dict
+            # https://swagger.io/specification/#path-item-object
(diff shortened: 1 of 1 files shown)


jfuss:
@leobasilio Thanks for the patch. Would you be able to add unit tests for this as well?

sriram-mv:
This is similar to #1988 fix. we can take in, one of the two.

PR #1996, opened 2020-05-18
Adding HttpApi to artifact exporter

leobasilio:
*Issue #, if available:*

awslabs/serverless-application-model#305, kind of.

The documentation on [AWS Docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-httpapi.html#sam-httpapi-definitionuri) states that a local file path is a valid input for the `DefinitionUri` property of a `AWS::Serverless::HttpApi` resource. Trying that leads to this error:

> Error: Failed to create changeset for the stack: my-stack, ex: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [GatewayHttpApi] is invalid. 'DefinitionUri' is not a valid S3 Uri of the form 's3://bucket/key' with optional versionId query parameter.

Although this scenario is supported for `AWS::Serverless::Api` for a while now, it is still missing for `HttpApi`. The definition file is not uploaded to S3.

Support for `DefinitionUri` as a local file path when using `HttpApi` was added following the already existing code for `Api`.

None.

No.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss:
@leobasilio Thanks for adding this. One thing we will want to do as well is add HttpApi here: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/validate/lib/sam_template_validator.py#L96 so that way when you reference a file, we can also run validate correctly.

tests/regression/package/test_package_regression.py
```diff
@@ -23,6 +23,7 @@ def tearDown(self):
     @parameterized.expand(
         [
             "aws-serverless-api.yaml",
+            "aws-serverless-httpapi.yaml",
```
jfuss, line 26: Not sure we want to add it here. These tests are to test regressions against AWS CLI. I don't think we need this covered for this addition.

leobasilio:
It's done. :+1:

PR #2000, opened 2020-05-20
feat: adding makefile workflow for layers build.

c2tarun:
*Issue #, if available:*

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -122,7 +122,8 @@ class ApplicationBuilder:
                     f"Layer {layer.name} cannot be build without BuildMethod. Please provide BuildMethod in Metadata.")
             result[layer.name] = self._build_layer(layer.name,
                                                    layer.codeuri,
-                                                   layer.build_method)
+                                                   layer.build_method,
+                                                   layer.compatible_runtimes)
 
         return result
 
@@ -172,7 +173,7 @@ class ApplicationBuilder:
 
         return template_dict
 
-    def _build_layer(self, layer_name, codeuri, specified_workflow):
+    def _build_layer(self, layer_name, codeuri, specified_workflow, compatible_runtimes):
         # Create the arguments to pass to the builder
         # Code is always relative to the given base directory.
         code_dir = str(pathlib.Path(self._base_dir, codeuri).resolve())
@@ -187,9 +188,15 @@ class ApplicationBuilder:
             manifest_path = self._manifest_path_override or os.path.join(code_dir, config.manifest_name)
 
             # By default prefer to build in-process for speed
+            build_runtime = specified_workflow
             build_method = self._build_function_in_process
             if self._container_manager:
-                build_method = self._build_function_on_container
+                build_method = self._build_function_in_process
+                if config.language == "provided":
+                    LOG.warning(
+                        "For container layer build, first compatible runtime is chosen as build target for container.")
+                    # Only set to this value if specified workflow is makefile which will result in config language as provided
+                    build_runtime = compatible_runtimes[0]
             options = ApplicationBuilder._get_build_options(layer_name, config.language, None)
 
             build_method(config,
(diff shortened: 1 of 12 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -187,17 +188,24 @@ def _build_layer(self, layer_name, codeuri, runtime):
             manifest_path = self._manifest_path_override or os.path.join(code_dir, config.manifest_name)
 
             # By default prefer to build in-process for speed
+            build_runtime = specified_workflow
             build_method = self._build_function_in_process
             if self._container_manager:
-                build_method = self._build_function_on_container
+                build_method = self._build_function_in_process
+                if config.language == "provided":
+                    LOG.warning(
```
sriram-mv, line 196: Nit: should we say which layer we are trying to build in here? or is it already covered elsewhere?

samcli/lib/build/workflow_config.py
```diff
@@ -140,6 +140,8 @@ def get_layer_subfolder(runtime):
         "ruby2.7": "ruby/lib",
         "java8": "java",
         "java11": "java",
+        # User is responsible for creating subfolder in these workflows
```
sriram-mv, line 143: Nit: we may need to change the name of this method later, since makefile is not a runtime.

it could be `get_layer_subfolder(build_workflow)` maybe?

samcli/lib/build/workflow_config.py
```diff
@@ -211,7 +213,9 @@ def get_workflow_config(runtime, code_dir, project_dir, specified_workflow=None)
         "provided": BasicWorkflowSelector(PROVIDED_MAKE_CONFIG)
     }
     # First check if the runtime is present and is buildable, if not raise an UnsupportedRuntimeException Error.
-    if runtime not in selectors_by_runtime:
+    # If runtime is present it should be in selectors_by_runtime, however for layers there will be no runtime so in that case
+    # we move ahead and resolve to any matching workflow from both types.
+    if runtime and runtime not in selectors_by_runtime:
```
sriram-mv, line 218: 👍

samcli/lib/build/app_builder.py
```diff
@@ -187,17 +188,24 @@ def _build_layer(self, layer_name, codeuri, runtime):
...
+                        "For container layer build, first compatible runtime is chosen as build target for container.")
+                    # Only set to this value if specified workflow is makefile which will result in config language as provided
+                    build_runtime = compatible_runtimes[0]
```
sriram-mv, line 199: If no compatible runtimes are specified, Im hoping the transform itself would fail before we get to this code path, so we should be good?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -736,6 +736,29 @@ def test_build_single_layer(self, runtime, use_container, layer_identifier):
             "python",
         )
 
+    @parameterized.expand(
+        [("makefile", False, "LayerWithMakefile"), ("makefile", "use_container", "LayerWithMakefile")]
+    )
+    def test_build_layer_with_makefile(self, runtime, use_container, layer_identifier):
```
sriram-mv, line 742: nit: instead of `runtime`, it can be `build_method`

samcli/lib/build/app_builder.py
c2tarun: Everytime a resource is build we see a log `Building layer 'LayerOne'` or `Building function 'FunctionOne'`.

samcli/lib/build/workflow_config.py
c2tarun: Sure.

samcli/lib/build/app_builder.py
c2tarun: Yeah, with no compatible runtime we don't reach here.

tests/integration/buildcmd/test_build_cmd.py
c2tarun: Sure.

Issue #2005, opened 2020-05-22
Error: PythonPipBuilder:ResolveDependencies - {pyrsistent==0.16.0(sdist)}


brysontyrrell:
# Description

Possibly related to #1840 but discovered while using new features in 0.51.0.

I implemented `Metadata` feature for Layers to allow `sam build` to install requirements.

```
    Metadata:
      BuildMethod: python3.8
```

When running `sam build --use-container` on the template I am getting this error:

```
Error: PythonPipBuilder:ResolveDependencies - {pyrsistent==0.16.0(sdist)}
```

The Layer triggering this error is installing the following packages:

```
jsonschema==3.2.0
PyJWT==1.7.1
```

If I downgrade from 0.51.0 to 0.50.0 which does not include the updates for supporting `Metadata` on Lambda Functions the build command completes without error.

### Steps to reproduce

Install SAM CLI 0.51.0.

Add `Metadata` build option as shown above to a Lambda Layer in a template.

Build the template.

### Observed result

This test template contains multiple Layers - only one is triggering the error.

```
 sam build --use-container --debug
Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
'build' command is called
Starting Build inside a container
No Parameters detected in the template
4 resources found in the template
No Parameters detected in the template
Building layer 'AwsXraySdk'
Loading workflow module 'aws_lambda_builders.workflows'
Registering workflow 'PythonPipBuilder' with capability 'Capability(language='python', dependency_manager='pip', application_framework=None)'
Registering workflow 'NodejsNpmBuilder' with capability 'Capability(language='nodejs', dependency_manager='npm', application_framework=None)'
Registering workflow 'RubyBundlerBuilder' with capability 'Capability(language='ruby', dependency_manager='bundler', application_framework=None)'
Registering workflow 'GoDepBuilder' with capability 'Capability(language='go', dependency_manager='dep', application_framework=None)'
Registering workflow 'GoModulesBuilder' with capability 'Capability(language='go', dependency_manager='modules', application_framework=None)'
Registering workflow 'JavaGradleWorkflow' with capability 'Capability(language='java', dependency_manager='gradle', application_framework=None)'
Registering workflow 'JavaMavenWorkflow' with capability 'Capability(language='java', dependency_manager='maven', application_framework=None)'
[...]

PR #2008, opened 2020-05-26
Fixing bug for build for layers in container.

c2tarun:
*Issue #2005 , if available:*
Ran integration test on local and it works.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -191,7 +191,7 @@ class ApplicationBuilder:
             build_runtime = specified_workflow
             build_method = self._build_function_in_process
             if self._container_manager:
-                build_method = self._build_function_in_process
+                build_method = self._build_function_on_container
                 if config.language == "provided":
                     LOG.warning(
                         "For container layer build, first compatible runtime is chosen as build target for container.")
--- a/tests/unit/lib/build_module/test_app_builder.py
+++ b/tests/unit/lib/build_module/test_app_builder.py
@@ -57,6 +57,69 @@ class TestApplicationBuilder_build(TestCase):
         )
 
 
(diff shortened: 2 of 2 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -191,7 +191,7 @@ def _build_layer(self, layer_name, codeuri, specified_workflow, compatible_runti
             build_runtime = specified_workflow
             build_method = self._build_function_in_process
             if self._container_manager:
-                build_method = self._build_function_in_process
+                build_method = self._build_function_on_container
```
sriram-mv, line 194: Is there a unit test that needs to be updated?

tests/unit/lib/build_module/test_app_builder.py
```diff
@@ -57,6 +57,69 @@ def test_must_iterate_on_functions_and_layers(self):
...
+        osutils_mock.mkdir_temp.return_value.__exit__ = Mock()
+
+        get_workflow_config_mock.return_value = config_mock
+        build_function_on_container_mock = Mock()
+
+        self.builder._build_function_on_container = build_function_on_container_mock
+        self.builder._build_layer("layer_name", "code_uri", "python3.8", ["python3.8"])
+
+        build_function_on_container_mock.assert_called_once()
+        args, _ = build_function_on_container_mock.call_args_list[0]
```
sriram-mv, line 111: 👍

samcli/lib/build/app_builder.py
sriram-mv: thanks for the unit test, this looks good to me.

PR #2023, opened 2020-06-05
Remove --location /path/to/template/folder docstr

gabehollombe-aws:
the `--location /path/to/template/folder` option listed in the doc string for the `sam init` command isn't supported by SAM, and https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/init/arbitrary_project.py#L41 indicates that it's explicitly not supported. 

This fix just removes this option from the doc string.

The doc string is misleading. You cannot pass a local path to the `--location` flag.

This PR removes this option from `init`'s doc string.

None.

No

### DIFF
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -48,10 +48,6 @@ Common usage:
     $ sam init --location /path/to/template.zip
     \b
     $ sam init --location https://example.com/path/to/template.zip
-    \b
-    Initializes a new SAM project using custom template in a local path
-    \b
-    $ sam init --location /path/to/template/folder
 """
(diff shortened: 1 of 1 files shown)


jfuss:
We still support `--location` for local filesystems, though it needs to be a cookiecutter example. What you are linking to is explicitly for arbitrary examples that are not cookiecutter not that we do not support location.

jfuss:
Closing because the comment linked is explicitly for non cookiecutter examples not that we do not support `--location' at all: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/init/arbitrary_project.py#L30

Issue #1955, opened 2020-04-30
`sam deploy --guided` doesn't seem to respect additional CLI args


afoeder:
### Steps to reproduce

Both invoke
a) `sam deploy --guided --capabilities CAPABILITY_AUTO_EXPAND`
b) `sam deploy                 --capabilities CAPABILITY_AUTO_EXPAND`

### Observed result

a) Deployment process is invoked with `capabilities = "CAPABILITY_IAM"`
b) Deployment process is invoked with `capabilities = "CAPABILITY_AUTO_EXPAND"`

### Expected result

Should be `CAPABILITY_AUTO_EXPAND` also in case a). 

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac
2. `sam --version`: `SAM CLI, version 0.48.0`

`Add --debug flag to command you are running`

sriram-mv:
Guided is currently a walkthrough process and you cant explicitly set those arguments that will be asked as questions during the walkthough process. However, this is a good request and I definitely see the value add. Marking it as a request.

wayne-folkes:
It would be helpful if the guided mode gave you an option to pick the type of capability you needed. In that wizard because it will just use `CAPABILITY_IAM` by default.

jfuss:
Reopening since the PR is merged but not released.

mndeveci:
Patched in [0.53.0 release](https://github.com/awslabs/aws-sam-cli/releases/tag/v0.53.0), closing

PR #2041, opened 2020-06-11
fix: guided deploy doesn't update default capabilities that is provid…

mndeveci:
…ed via --capabilites option as CLI argument

*Issue #1955

As explained in the issue above, running guided deploy with --capabilities parameter doesn't respect the given parameters. Where it should update the default capabilities with the ones that is given by CLI parameter.

In the guided context, default capabilities were assigned statically to `CAPABILITY_IAM`. With this commit it is changed to assign the default capabilities that is read from the CLI or from samconfig.toml file

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes

### DIFF
--- a/samcli/commands/deploy/guided_context.py
+++ b/samcli/commands/deploy/guided_context.py
@@ -69,7 +69,7 @@ class GuidedContext:
     def guided_prompts(self, parameter_override_keys):
         default_stack_name = self.stack_name or "sam-app"
         default_region = self.region or "us-east-1"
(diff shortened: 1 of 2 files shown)


sriram-mv:
This looks good in general @mndeveci were you able to deploy and check these changes?

mndeveci:
@sriram-mv as we synced offline, I tested with couple of scenarios and application has been deployed successfully.

Issue #1947, opened 2020-04-28
Improve exception logging for API Gateway Local Service


lfpetrini:
This is a very simple request, I didn't know if it could be considered a bug since it's just related to logging, but I have spent hours trying to figure out what is wrong with the response I'm sending back from my lambda function and the only error message I get is this one:

> [32mREPORT RequestId: ac7ae76f-bdb4-1dd8-b184-6a7d2ad94f20     Init Duration: 544.64 ms        Duration: 23.88 ms      Billed Duration: 100 ms Memory Size: 128 MB     Max Memory Used: 56 MB      [0m
> **<class 'samcli.local.apigw.local_apigw_service.LambdaResponseParseException'>**

I'm new to SAM so I decided to take a look at the [source code](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/apigw/local_apigw_service.py).
Even though the exceptions are raised with very informative messages, at the end it will just log its type instead of anything useful (line 214):
```
except LambdaResponseParseException:
            LOG.error(str(LambdaResponseParseException))
            return ServiceErrorResponses.lambda_failure_response()
```

I suppose an \__str__ function should be added to the exception, and it should log its instance instead of its type.

awood45:
That makes sense, we'd be happy to take a PR for this as well.

mndeveci:
Hi @lfpetrini ,

Thanks for raising this issue up. I took a look at the concern here, and try to re-produce it by throwing an exception inside the lambda function. But I can see the exception has been logged into console.

Example code piece that I added inside handler function to throw the exception;
```
if true {
	panic(errors.New("Error message here"))
}
```
And the logs here;

```
Fetching lambci/lambda:go1.x Docker container image......
Mounting /Volumes/unix/workplace/other/sam-internal-go-app/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
START RequestId: 34908c02-e2e5-13be-8dfc-fe1422c06da1 Version: $LATEST
END RequestId: 34908c02-e2e5-13be-8dfc-fe1422c06da1
REPORT RequestId: 34908c02-e2e5-13be-8dfc-fe1422c06da1  Init Duration: 266.24 ms        Duration: 471.41 ms     Billed Duration: 500 ms    Memory Size: 128 MB     Max Memory Used: 26 MB
Lambda returned empty body!
[...]

PR #2053, opened 2020-06-17
fix: print exception details for LambdaResponseParseException and rem…

mndeveci:
…ove redundant logging in upstreams methods

*Issue #1947

When debugging lambda functions locally, some of the exceptions are not printed in the console.

Instead of printing the exception type, it is been changed to print exception details. Any upstream logging is removed for not having duplicate logs in the console.

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -214,8 +214,8 @@ class LocalApigwService(BaseLocalService):
             (status_code, headers, body) = self._parse_lambda_output(
                 lambda_response, self.api.binary_media_types, request
             )
-        except LambdaResponseParseException:
-            LOG.error(str(LambdaResponseParseException))
+        except LambdaResponseParseException as ex:
+            LOG.error("Invalid lambda response received: %s", ex)
             return ServiceErrorResponses.lambda_failure_response()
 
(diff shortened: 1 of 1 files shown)


samcli/local/apigw/local_apigw_service.py
```diff
@@ -287,25 +287,21 @@ def _parse_lambda_output(lambda_output, binary_types, flask_request):
...
 
         # API Gateway only accepts statusCode, body, headers, and isBase64Encoded in
         # a response shape.
         invalid_keys = LocalApigwService._invalid_apig_response_keys(json_output)
         if bool(invalid_keys):
-            msg = "Invalid API Gateway Response Keys: " + str(invalid_keys) + " in " + str(json_output)
-            LOG.error(msg)
-            raise LambdaResponseParseException(msg)
+            raise LambdaResponseParseException(
+                f"Invalid API Gateway Response Keys: {str(invalid_keys)} in {str(json_output)}"
```
jfuss, line 303: No need call `str()`. F-strings will do this for you.

mndeveci: Oh didn't know that, thanks! :) Will update in next revision

PR #2066, opened 2020-06-24
feat: Use aws-sam-cli docker images

jfuss:
*Issue #, if available:*
N/A

This changes updates SAM CLI to use new `amazon/aws-sam-cli-build-image-<runtime>` and `amazon/aws-sam-cli-emulation-image-<runtime>` images.

Updates relevant code around image usage.

I have verified that invoke and build images works for runtimes but .Net and provided. These run integ tests in Appveyor so will verify them that way. I have not tested the debug experience. The entrypoints are updated but they could be incorrect in the current state. This is going into a non-develop branch so this will not impact any upcoming releases or customers unless the cli is explicitly installed from the `release-v1.0.0` branch.

N/A

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mhart:
Not sure if y'all are tracking changes to docker-lambda – but just landed CORS support too (which Lambda prod also has, and a couple of users asked for) – pretty minor change: https://github.com/lambci/docker-lambda/commit/20fca7e5121e4b8e40fe68d6a826220d0389d012

Issue #2028, opened 2020-06-09
Region shows as "None" after stack deploys using manual parameters


jbesw:
### Description

Briefly describe the bug you are facing.

In the Cloud9 IDE, if you deploy using "sam deploy" with manual parameters, and do not supply a region in the command line, it uses the correct region but the output message shows "None". 

Partially described in issue https://github.com/awslabs/aws-sam-cli/issues/1704.

### Steps to reproduce

1. Open Cloud9.
2. Deploy a stack with manual parameters, e.g.:
sam deploy --template-file packaged.yaml --stack-name ride-times --capabilities CAPABILITY_IAM
3. See output message ("Successfully created/updated stack - ride-times in None").

The stack deploys implicitly to the correct region (where the Cloud9 instance is located) but the output message is misleading/wrong.

### Observed result

Please provide command output with `--debug` flag set.

jbeswick@amazon.com:~/environment/theme-park-backend/1-app-deploy/ride-controller (master) $ sam deploy --template-file packaged.yaml --stack-name ride-times --capabilities CAPABILITY_IAM

        Deploying with following values
        ===============================
        Stack name                 : ride-times
        Region                     : None
        Confirm changeset          : False
        Deployment s3 bucket       : None
        Capabilities               : ["CAPABILITY_IAM"]
        Parameter overrides        : {}

Initiating deployment

### Expected result

It should display the region code where the stack was deployed.

Describe what you expected.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Cloud9
2. `sam --version`: SAM CLI, version 0.38.0

`Add --debug flag to command you are running`

timoschilling:
maybe related https://github.com/awslabs/aws-sam-cli/issues/1704

prenx4x:
This issue is fixed and merged.

PR #2068, opened 2020-06-24
fix: Region shows as "None" after stack deploys using manual parameters

prenx4x:
*Issue #, if available:*
#2028 

- When you manually deploy without providing region, the console prints show region as None.

- Used the default region information from default profile in boto session object.

- None

- No

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/context.py
+++ b/samcli/cli/context.py
@@ -32,7 +32,7 @@ class Context:
         Initialize the context with default values
         """
         self._debug = False
-        self._aws_region = boto3.session.Session().region_name
+        self._aws_region = None
         self._aws_profile = None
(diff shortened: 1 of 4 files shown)


samcli/cli/context.py
```diff
@@ -32,7 +32,7 @@ def __init__(self):
         Initialize the context with default values
         """
         self._debug = False
-        self._aws_region = None
+        self._aws_region = boto3.session.Session().region_name
```
sriram-mv, line 35: botocore.session.get_session() might be better, this might create a brand new session.

Issue #2095, opened 2020-07-06
RC1 failing local invoke for Node12.


c2tarun:
### Description

`sam local invoke` fails for nodejs12 project with following error:

```
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http+docker://localhost/v1.35/containers/b750d7791ae2d2b1ab5b57a27a0b50f8379e170e06e9205c9ab150827f69fcea/start
```

### Steps to reproduce

1. `sam init` a node12 project.
2. `sam local invoke` in that project.

### Observed result

```
❯ sam local invoke --debug                                                                                                                                                               ─╯
Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
local invoke command is called
No Parameters detected in the template
2 resources found in the template
Found Serverless function with name='HelloWorldFunction' and CodeUri='HelloWorldFunction'
Found one Lambda function with name 'HelloWorldFunction'
Invoking app.lambdaHandler (nodejs12.x)
No environment variables found for function 'HelloWorldFunction'
Environment variables overrides data is standard format
Loading AWS credentials from session with profile 'None'
Resolving code path. Cwd=/Users/tarumall/workplace/byol-testbed/sam-app-n12/.aws-sam/build, CodeUri=HelloWorldFunction
Resolved absolute path to code is /Users/tarumall/workplace/byol-testbed/sam-app-n12/.aws-sam/build/HelloWorldFunction
Code /Users/tarumall/workplace/byol-testbed/sam-app-n12/.aws-sam/build/HelloWorldFunction is not a zip/jar file
Skipping building an image since no layers were defined

Fetching amazon/aws-sam-cli-emulation-image-nodejs12.x:latest Docker container image......
Mounting /Users/tarumall/workplace/byol-testbed/sam-app-n12/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
[...]

PR #2098, opened 2020-07-07
Move Mounted Files to .aws-sam Folder

awood45:
*Issue #, if available:* Resolves issue #2095 


samcli/commands/local/cli_common/options.py
```diff
@@ -117,6 +145,20 @@ def invoke_common_options(f):
                 help="Specifies the location basedir where the Layers your template uses will be downloaded to.",
                 default=get_default_layer_cache_dir(),
             ),
+            click.option(
+                "--rapid-basedir",
```
c2tarun, line 149: Do we want to expose this functionality?

samcli/local/docker/lambda_container.py
```diff
@@ -143,20 +144,20 @@ def _get_additional_options(runtime, debug_options):
         return opts
 
     @staticmethod
-    def _get_additional_volumes(runtime, debug_options):
+    def _get_additional_volumes(runtime, debug_options, mounted_file_provider):
```
c2tarun, line 147: missing in doc

samcli/local/mount/mounted_file_provider.py
```diff
@@ -0,0 +1,48 @@
...
+        return self._rapid_basedir
+
+    @property
+    def go_bootstrap_basedir(self):
+        return self._go_bootstrap_basedir
+
+    @staticmethod
+    def _copy_source_files(rapid_basedir, go_bootstrap_basedir):
+        LOG.debug("Creating basedirs if they do not yet exist.")
+        Path(rapid_basedir).mkdir(mode=0o700, parents=True, exist_ok=True)
```
c2tarun, line 36: I am assuming that python somehow handles this for Windows.

samcli/local/mount/mounted_file_provider.py
```diff
@@ -0,0 +1,48 @@
...
+        rapid_dest = "{}/init".format(rapid_basedir)
+        LOG.debug("Copying RAPID stub server from %s to %s.", str(MountedFileProvider._RAPID_SOURCE), rapid_dest)
+        copyfile(MountedFileProvider._RAPID_SOURCE, rapid_dest)
+        rapid_st = os.stat(MountedFileProvider._RAPID_SOURCE)
+        os.chmod(rapid_dest, rapid_st.st_mode)
+
+        go_bootstrap_dest = "{}/aws-lambda-go".format(go_bootstrap_basedir)
+        copyfile(MountedFileProvider._GO_BOOTSTRAP_SOURCE, go_bootstrap_dest)
+        go_bootstrap_st = os.stat(MountedFileProvider._GO_BOOTSTRAP_SOURCE)
+        os.chmod(go_bootstrap_dest, go_bootstrap_st.st_mode)
```
c2tarun, line 48: nitpick: probably can be extracted in a method and called for both files.

samcli/local/mount/mounted_file_provider.py
```diff
@@ -0,0 +1,48 @@
...
+import os
+import stat
+
+from pathlib import Path
+from shutil import copyfile
+
+LOG = logging.getLogger(__name__)
+
+
+class MountedFileProvider:
```
c2tarun, line 14: Some doc would be helpful here when we'll see this in future 😄

samcli/commands/local/cli_common/options.py
awood45: Will look at an alternative approach tomorrow, but short answer is "yes" - if the user has folder permission issues and we don't provide this option, they're basically stuck with no ability to invoke. However, we should carefully document that most users will never want to touch this.

samcli/local/mount/mounted_file_provider.py
awood45: It does, we use this for global config and layer caching.

awood45:
We may pick this back up later if needed, but closing in favor of #2100

PR #2100, opened 2020-07-08
refactor: Bake Rapid into image on the fly

jfuss:
*Issue #, if available:*
This is an alternative solution to #2098 

We need RAPID inside the container at runtime. The options are mounting or created an image on they fly. Mounting requires the path to be sharable to docker and if it not, the invoke will fail which is a regressions in rc1.

Instead of mounting or allowing customers to give the path to an alternative rapid, we will just build the image locally that has rapid baked into it. This will ensure we can set permissions on this binary (if needed). This should get around the initially issue found in #2095 where we could not mount the rapid binary due to share settings of docker. The implementation creates a tar and sends that to the docker build (through the sdk). If we have access to the file (which we should) then we can create the tar and therefore not have to deal with docker share settings.

None that I know of

N/A

Extra details:
If we decide to go this route, we should consider doing the same for the custom go boostrap that enabled debugging for go.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_container.py
+++ b/samcli/local/docker/lambda_container.py
@@ -24,8 +24,6 @@ class LambdaContainer(Container):
     # The Volume Mount path for debug files in docker
     _DEBUGGER_VOLUME_MOUNT_PATH = "/tmp/lambci_debug_files"
     _DEFAULT_CONTAINER_DBG_GO_PATH = _DEBUGGER_VOLUME_MOUNT_PATH + "/dlv"
-    _RAPID_SOURCE_PATH = Path(__file__).parent.joinpath("..", "rapid").resolve()
-    _RAPID_DESTINATION_MOUNT = {"bind": "/var/rapid", "mode": "ro"}
     _GO_BOOTSTRAP_SOURCE_PATH = Path(__file__).parent.joinpath("..", "go-bootstrap").resolve()
     _GO_BOOTSTRAP_DESTINATION_MOUNT = {"bind": "/var/runtime", "mode": "ro"}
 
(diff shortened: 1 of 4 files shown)


jfuss:
To get an idea of performance impact:
If the image is already cached:
With this PR: `samdev local invoke  0.63s user 0.33s system 21% cpu 4.500 total`
With v0.47.0: `sam local invoke  0.57s user 0.21s system 22% cpu 3.416 total`

If the baked image is already cached:
with this PR: `samdev local invoke  0.77s user 0.33s system 30% cpu 3.591 total`

So the first invoke will be able 1 second slower with this PR. I can profile the actually time it takes to build this image if need be to better understand the difference (without having network being a factor).

Note: This is building off of what we already do for support Layers.

samcli/local/docker/lambda_container.py
```diff
@@ -24,8 +24,6 @@ class LambdaContainer(Container):
     # The Volume Mount path for debug files in docker
     _DEBUGGER_VOLUME_MOUNT_PATH = "/tmp/lambci_debug_files"
     _DEFAULT_CONTAINER_DBG_GO_PATH = _DEBUGGER_VOLUME_MOUNT_PATH + "/dlv"
-    _RAPID_SOURCE_PATH = Path(__file__).parent.joinpath("..", "rapid").resolve()
-    _RAPID_DESTINATION_MOUNT = {"bind": "/var/rapid", "mode": "ro"}
     _GO_BOOTSTRAP_SOURCE_PATH = Path(__file__).parent.joinpath("..", "go-bootstrap").resolve()
```
awood45, line 27: What are we going to do with these? Go debugging is going to have the same issue.

samcli/local/docker/lambda_image.py
```diff
@@ -85,15 +86,15 @@ def build(self, runtime, layers):
         """
         base_image = "{}-{}:latest".format(self._INVOKE_REPO_PREFIX, runtime)
 
-        # Don't build the image if there are no layers.
-        if not layers:
-            LOG.debug("Skipping building an image since no layers were defined")
-            return base_image
+        # Default image tag to be the base image with a tag of 'rapid' instead of latest
+        image_tag = "{}-{}:rapid".format(self._INVOKE_REPO_PREFIX, runtime)
```
awood45, line 90: How will this pick up any updates made to the `init` binary?

PR #2102, opened 2020-07-08
feature: Added support for parameter --resolve-s3 for non-guided deployments

prenx4x:
*Issue #, if available:*

Adding support for non-guided deployments which resolves s3-buckets similar to guided deployments.

By adding new parameter --deploy-s3

None

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/deploy/command.py
```diff
@@ -222,6 +234,14 @@ def do_cli(
         )
         guided_context.run()
 
+    if resolve_s3:
+        if s3_bucket or guided:
```
awood45, line 238: I think if `resolve_s3` is true, and `--guided` is supplied, we don't need to error out and we can just continue. `--guided` implies this to be true.

samcli/commands/deploy/command.py
```diff
@@ -127,6 +127,13 @@
     help="Indicates whether to use JSON as the format for "
     "the output AWS CloudFormation template. YAML is used by default.",
 )
+@click.option(
```
awood45, line 130: This is a good start, but we should ALSO plumb this into the package command. It's possible for the package command to be run on its own and that's where S3 bucket resolution will happen.

samcli/commands/deploy/exceptions.py
```diff
@@ -67,3 +67,12 @@ def __init__(self):
         )
 
         super(DeployBucketRequiredError, self).__init__(message=message_fmt)
+
+
+class DeployResolveS3AndS3Error(UserException):
+    def __init__(self):
+        message_fmt = (
+            "Cannot use both --deploy-s3 and guided or --s3-bucket parameter simultaneously." "Please use either one."
```
awood45, line 75: Will want to update this error message to reflect that guided is not part of the conflict.

samcli/commands/deploy/command.py
```diff
@@ -222,6 +234,14 @@ def do_cli(
         )
         guided_context.run()
 
+    if resolve_s3:
+        if s3_bucket or guided:
+            raise deploy_exceptions.DeployResolveS3AndS3Error()
+
+        s3_bucket = manage_stack(profile=profile, region=region)
+        click.echo(f"\n\t\tManaged S3 bucket: {s3_bucket}")
+        click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
```
awood45, line 243: Or by specifying `--s3-bucket` explicitly.

samcli/commands/deploy/exceptions.py
```diff
@@ -67,3 +67,10 @@ def __init__(self):
         )
 
         super(DeployBucketRequiredError, self).__init__(message=message_fmt)
+
+
+class DeployResolveS3AndS3Error(UserException):
+    def __init__(self):
+        message_fmt = "Cannot use both --deploy-s3 and --s3-bucket parameter simultaneously." "Please use either one."
```
awood45, line 74: s/deploy/resolve in the parameter name here.

samcli/commands/package/command.py
```diff
@@ -112,8 +131,20 @@ def do_cli(
     metadata,
     region,
     profile,
+    resolve_s3,
 ):
     from samcli.commands.package.package_context import PackageContext
+    from samcli.lib.bootstrap.bootstrap import manage_stack
+    from samcli.commands.deploy import exceptions as deploy_exceptions
```
awood45, line 138: Let's not pull in all deploy exceptions, can we just pull down the single exception? Or move it into the shared exceptions file at `samcli.command.exceptions` since it's shared across methods.

samcli/commands/package/command.py
```diff
@@ -112,8 +131,20 @@ def do_cli(
...
+
+    if resolve_s3:
+        if s3_bucket:
```
awood45, line 141: `s3_bucket` is currently a required parameter. You'll need to make it no longer required, and add a check that either `s3_bucket` or `resolve_s3` is specified. That might also be the right place to check that you raise an error if both are specified.

tests/unit/commands/deploy/test_command.py
```diff
@@ -39,6 +39,7 @@ def setUp(self):
         self.metadata = {}
         self.guided = False
         self.confirm_changeset = False
+        self.resolve_s3 = False
```
awood45, line 42: Fine as a default, but can we add unit tests where this is set to true, and where we stub out the act of actually setting up the bucket?

tests/unit/commands/package/test_command.py
```diff
@@ -17,6 +17,7 @@ def setUp(self):
         self.metadata = {"abc": "def"}
         self.region = None
         self.profile = None
+        self.resolve_s3 = False
```
awood45, line 20: Fine as a default, but can we add unit tests where this is set to true, and where we stub out the act of actually setting up the bucket?

Especially important here as our unit tests didn't notice that the basic case would never work (you can't set this param because s3_bucket is required).

tests/unit/commands/deploy/test_command.py
```diff
@@ -563,3 +571,53 @@ def test_all_args_guided_no_params_no_save_config(
...
+                profile=self.profile,
+                use_json=self.use_json,
+                metadata=self.metadata,
+                guided=False,
+                confirm_changeset=True,
+                resolve_s3=True,
+            )
+
+    def test_resolve_s3_and_s3_bucket_both_not_set(self):
+        with self.assertRaises(DeployResolveS3AndS3Error):
```
awood45, line 601: This is probably the wrong error to raise if neither is specified.

samcli/commands/deploy/exceptions.py
```diff
@@ -67,3 +67,23 @@ def __init__(self):
         )
 
         super(DeployBucketRequiredError, self).__init__(message=message_fmt)
+
+
+class DeployResolveS3AndS3SetError(UserException):
+    def __init__(self):
+        message_fmt = (
+            "Cannot use both --resolve-s3 and --s3-bucket parameters in non-guided deployments."
```
awood45, line 75: Please add a space here, the two sentences are concatenated with no space.

samcli/commands/deploy/exceptions.py
```diff
@@ -67,3 +67,23 @@ def __init__(self):
...
+        )
+
+        super(DeployResolveS3AndS3SetError, self).__init__(message=message_fmt)
+
+
+class DeployResolveS3AndS3NotSetError(UserException):
+    def __init__(self):
+        message_fmt = (
+            "Cannot skip both --resolve-s3 and --s3-bucket parameters in non-guided deployments."
+            "Please use only one or use the --guided option for a guided deployment."
```
awood45, line 86: Same space comment.

samcli/commands/package/exceptions.py
```diff
@@ -90,3 +90,17 @@ def __init__(self, **kwargs):
...
+class PackageResolveS3AndS3SetError(UserException):
+    def __init__(self):
+        message_fmt = "Cannot use both --resolve-s3 and --s3-bucket parameters." "Please use only one."
+
+        super(PackageResolveS3AndS3SetError, self).__init__(message=message_fmt)
+
+
+class PackageResolveS3AndS3NotSetError(UserException):
+    def __init__(self):
+        message_fmt = "Cannot skip both --resolve-s3 and --s3-bucket parameters." "Please use only one."
```
awood45, line 104:
```
message_fmt = "Cannot skip both --resolve-s3 and --s3-bucket parameters. Please provide one of these arguments."
```

samcli/commands/package/exceptions.py
```diff
@@ -90,3 +90,17 @@ def __init__(self, **kwargs):
         message_fmt = "\nS3 Bucket not specified, use --s3-bucket to specify a bucket name or run sam deploy --guided"
 
         super(BucketNotSpecifiedError, self).__init__(message=message_fmt.format(**self.kwargs))
+
+
+class PackageResolveS3AndS3SetError(UserException):
+    def __init__(self):
+        message_fmt = "Cannot use both --resolve-s3 and --s3-bucket parameters." "Please use only one."
```
awood45, line 97:
```
message_fmt = "Cannot use both --resolve-s3 and --s3-bucket parameters. Please use only one."
```

PR #2105, opened 2020-07-12
add new flag that allow to send event into lambda docker function thr…

### LINKED ISSUE #188: Feature Request: Support large event payload input to Lambda
I’m trying to build an API endpoint lambda function that accepts an uploaded binary file.

Try this:

```bash
git clone git@github.com:awslabs/aws-sam-local.git
cd aws-sam-local/samples/api-event-source
sam local start-api
```
and then in a new shell, from the `aws-sam-local` folder:

```bash
curl --request POST --url 'http://localhost:3000/' --data-binary '@./sam-banner.png'
```

The response I get to `curl` is `{ "message": "Internal server error" }`.

Back in the first shell, I see this output:

```
2017/11/01 23:05:55 Connected to Docker 1.32
2017/11/01 23:05:55 Fetching lambci/lambda:nodejs6.10 image for nodejs6.10 runtime...
nodejs6.10: Pulling from lambci/lambda
Digest: sha256:7eb4ced6a15ae3c30effc4ec0cd3aabb2bd57c9a8330b37920c3d5d722d81083
Status: Image is up to date for lambci/lambda:nodejs6.10

Mounting index.handler (nodejs6.10) at http://127.0.0.1:3000/ [OPTIONS GET HEAD POST PUT DELETE TRACE CONNECT]
Mounting static files from public at /

You can now browse to the above endpoints to invoke your functions.
You do not need to restart/reload SAM CLI while working on your functions,
changes will be reflected instantly/automatically. You only need to restart
SAM CLI if you update your AWS SAM template.

2017/11/01 23:06:19 Invoking index.handler (nodejs6.10)
standard_init_linux.go:185: exec user process caused "argument list too long"
2017/11/01 23:06:25 Function index.handler timed out after 3 seconds
2017/11/01 23:06:25 Error invoking nodejs6.10 runtime: io: read/write on closed pipe
```

If I remove the `--data-binary`, to send a POST request without a body, I get the expected `OK` response.

If I change `--data-binary '@./sam-banner.png'` to `--data-binary '@./README.md'`, replacing the 129 KB file with a 21 KB one, I also get the `OK` response. So I gather that there must be some (fairly low) limit on request size? Is there some way I can increase it?

Also, I feel like `aws-sam-local` should print an informative error message, rather than (I presume) passing invalid parameters onward to Docker for it to throw an exception on.

DKurilo:
add new flag that allow to send event into lambda docker function through stdin

Fixes #188

Current version can send event only as environmental variable, that is sent through command line argument. This way has limitations that impossible to change in OS (at least in MacOS). So now it's impossible to debug locally lambda that should work with images, for example.

This PR adds flag `--use_stdin` and with this flag API Gateway sends data into lambda docker container using stdin.

I hope it don't have any side effects. It's still possible to use system and its components the same way as now. Only one place that contains breaking changes is here:
https://github.com/DKurilo/aws-sam-cli/blob/docker-invoke-stdin/samcli/commands/local/start_api/cli.py#L95
But as I understand, it's not supposed to use `do_cli` from any other modules.

I didn't change a dependency in `requirements/base.txt`

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


CoshUS (approved):

moelasmar:
This PR is not needed any more, as after we moved the new rapid-rie server, and no need to send the event input in the container environment variables.

This PR https://github.com/aws/aws-sam-cli/pull/2506 is resolving this issue.

PR #2112, opened 2020-07-17
Build output dots

awood45:
Helps anything that might depend on STDOUT output...

### DIFF
--- a/tests/unit/local/docker/test_lambda_image.py
+++ b/tests/unit/local/docker/test_lambda_image.py
@@ -1,3 +1,5 @@
+import io
+
 from unittest import TestCase
 from unittest.mock import patch, Mock, mock_open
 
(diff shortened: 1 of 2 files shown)


tests/unit/local/docker/test_lambda_image.py
```diff
@@ -84,16 +86,22 @@ def test_force_building_image_that_doesnt_already_exists(
         docker_client_mock = Mock()
         docker_client_mock.images.get.side_effect = ImageNotFound("image not found")
 
+        stream = io.StringIO()
```
sriram-mv, line 89: Nit: BytesIO instead?

awood45: Matches our `manager.py` unit tests.

PR #2113, opened 2020-07-19
Bumping Boto3

chadbrewbaker:
*Issue #, if available:*
It generated an error when building the lambci python3.8 Docker image.

Bumps to parity with aws-cli and gets rid of build errors on lambci python3.8 Docker image

Eliminates the build error.

It bumps boto3 from 1.13.x to 1.14.x. 

Yes.
No. For security reasons I would feel better if a core maintainer updated the hashcodes.
*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

Yes.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.0
 Flask~=1.0.2
 #Need to add Schemas latest SDK.
-boto3~=1.13.0, >=1.13.0
+boto3~=1.14.23
 jmespath~=0.9.5
 PyYAML~=5.1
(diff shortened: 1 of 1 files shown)


jfuss:
Please see my comment here: https://github.com/lambci/docker-lambda/issues/296#issuecomment-661104076.

This is only a short term fix. Docker-lambda (and any other tool) should be installing python tools into isolated venv, if they use pip. Otherwise, this will happen again in the future with your function code or AWS CLI.

awood45:
#2130 takes Jacob's feedback and replaces this PR - thank you for the contribution!

PR #2124, opened 2020-07-23
Fix: template path (#1685)

stackchain:
*Issue #, if available:*

I don't know how it was working so far after the mega refactor made by @sanathkr the template folder was missing, it was moved from local -> lib.

Fixing the path to the right folder where the templates live

Cache?

No
*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
@stackchain We only used the ones packaged in the CLI as a default. We have moved everything to a separate repo (https://github.com/awslabs/aws-sam-cli-app-templates/) and download that with the examples instead.

aahung:
This has been fixed in #2425

PR #2132, opened 2020-07-24
Fix: Encoding for Reading Template File

CoshUS:
*Issues
https://github.com/awslabs/aws-sam-cli/issues/1248
https://github.com/awslabs/aws-sam-cli/issues/1628
https://github.com/awslabs/aws-sam-cli/issues/1786

Template file is being read using system default encoding instead of UTF-8. In certain situations like Windows with Japanese cp 932, UTF-8 characters will cause a decoding error.
In other cases, the template file is read using UTF-8 without BOM which appends BOM to the start of the actual content.

Template file is forced to be read with UTF-8.

Non UTF-8 compatible encoded template files will not be supported.

No

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/_utils/template.py
```diff
@@ -39,7 +39,7 @@ def get_template_data(template_file):
     if not pathlib.Path(template_file).exists():
         raise TemplateNotFoundException("Template file not found at {}".format(template_file))
 
-    with open(template_file, "r") as fp:
+    with open(template_file, "r", encoding="UTF-8") as fp:
```
jfuss, line 42: nit: Let's lower case `uft-8` to match what we do in other places.

tests/integration/testdata/invoke/sam-template.yaml
```diff
@@ -10,3 +10,4 @@ Resources:
       Runtime: python3.6
       CodeUri: .
       Timeout: 600
+# 😃 UTF-8 Decoding Test
```
jfuss, line 13: Question: Without this change, does this actually fail?

CoshUS: It fails if the environment is Windows and uses cp932 or some other encodings. For now, I don't think any of the test environments will fail this, which means we might need to add another one.

CoshUS: Added system encoding change script for Windows in AppVeyor.

appveyor-windows.yml
```diff
@@ -80,6 +80,17 @@ test_script:
     - "docker system prune -a -f"
     - "pytest -vv tests/integration"
     - "pytest -vv tests/regression"
+    
+    # Change encoding to cp932
+    - ps: "Set-WinSystemLocale -SystemLocale ja-JP"
```
jfuss, line 85: This will add a lot of time into our tests. Can we not accomplish the same thing by create a specific test that will change the Locale for the duration of that test? 

What this gives us is a specific test that can be run anywhere (not just windows) that ensures if the system locale is different than utf8, things still work as expected. This will allow any developer to be able to verify locally, and not have to rely on specific setup/tests run on the ci/cd system. When this test will fail, we will know exactly what is broken vs thinking things are working, seeing ci/cd fail, and have to dig through failures which will look like all invokes are broken.

CoshUS: Removed locale system change in favor of using unittest tests to mock open with different locale.

tests/integration/local/invoke/test_integrations_do_cli.py
```diff
@@ -0,0 +1,54 @@
...
+
+        cls._original_getpreferredencoding = locale.getpreferredencoding
+        locale.getpreferredencoding = mock_getpreferredencoding
+
+    @classmethod
+    def tearDownClass(cls):
+        locale.getpreferredencoding = cls._original_getpreferredencoding
+
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
```
jfuss, line 39: I am not super familiar with pytest fixtures. What is this doing and why should we use this over the unittest module in the standard lib? Historically, we have tried to keep all our test using soely unittest, so everything is consistent. I am ok with using fixtures for this, but in the future I would prefer to match our existing setup and not rely on pytest other than a runner of tests.

CoshUS: caplog fixture is needed here for a issue with Click and Logging. 
https://github.com/pallets/click/issues/824
Not sure if there is another way around this problem.

jfuss: Works for me. Thanks for explaining.

PR #1952, opened 2020-04-30
Updated sam cli to read schemas region from boto3

gupta-n:
*Issue #, if available:*
EventBridge Schemas service regions are hardcoded in SAM CLI.

*Why is this change necessary?*
This change is necessary to avoid SAM CLI changes when schema service is available in new region.

*How does it address the issue?*
By reading schema regions from boto3.

*What side effects does this change have?*
Customer needs to use latest AWS SDK.

*Did you change a dependency in `requirements/base.txt`?*
    *If so, did you run `make update-reproducible-reqs`*
     Yes.

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [Done] Write unit tests
- [ Done] Write/update functional tests
- [ Done] Write/update integration tests
- [ Done] `make pr` passes
- [ NA] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2158, opened 2020-08-07
Loosen the pin on boto3 to allow >=1.14.23,<2

michael-k:
Originally the pin was `boto3~=1.5` ([meaning](https://www.python.org/dev/peps/pep-0440/#compatible-release) >=1.5,<2).  Followed by `~=1.9, >=1.9.56` and `~=1.10, >=1.10.29`, both allowing all later releases up to `<2`.

Then it changed ~ 3 months ago to `boto3~=1.13.0, >=1.13.0` in commit 2e73a0b6fb5073972d40bdecc8ab10fcf1f6f07e (PR awslabs/aws-sam-cli#1952).  Neither the commit message nor the pull request (there's no linked issue) gives any reason why suddenly the upper limit should be `<1.14`.  And `boto3~=1.13.0, >=1.13.0` is the same as `boto3~=1.13.0`.

Therefore I think adding `.0` to `boto3~=1.13` happened by accident.  Following boto3's release history, there's also no reason to believe that boto3 1.15 or later up to `<2` might include a breaking change.

----

**Issue #, if available:** n/a

**Why is this change necessary?**

The current release of aws-sam-cli cannot be installed alongside boto3 1.14.*.  This has been fixed in the `develop` branch (PR awslabs/aws-sam-cli#2130).  But the pin is too strict, leading to the same problem with later releases.

**How does it address the issue?**

Allowing all releases of boto3 between 1.14.23 (inclusive) and 2 (exclusive).

**What side effects does this change have?**

Later releases (<2) of boto3 might not be compatible with the upcoming releases of aws-sam-cli.  But following boto3's history, there's no reason to believe that this might happen.

**Did you change a dependency in `requirements/base.txt`?**

Yes

- **If so, did you run `make update-reproducible-reqs`?**

    No, I only made the range less restrictive.  The updated range is still compatible with the full pins generated by pip-compile. 

**Checklist:**

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

----

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.0
 Flask~=1.0.2
 #Need to add Schemas latest SDK.
-boto3~=1.14.0, >=1.14.23
+boto3~=1.14, >=1.14.23
 jmespath~=0.9.5
 PyYAML~=5.1
 cookiecutter~=1.6.0


michael-k:
Btw. @jfuss said

> Please see my comment here: lambci/docker-lambda#296 (comment).
> 
> This is only a short term fix. Docker-lambda (and any other tool) should be installing python tools into isolated venv, if they use pip. Otherwise, this will happen again in the future with your function code or AWS CLI.

(source: https://github.com/awslabs/aws-sam-cli/pull/2113#issuecomment-661114611)

But your own AWS CodeBuild is affected by that.

Here's a minimal `buildspec.yml` which fails without installing anything running on `aws/codebuild/standard:4.0` (CodeBuild Image):

```
---
version: 0.2

phases:
  install:
    runtime-versions:
      python: 3.8
    commands:
      - python -m pip freeze
      - python -m pip check
```

<details><summary>Partial build log</summary>

```
(…)
[Container] 2020/08/07 00:00:00 Running command python -m pip freeze
(…)
aws-sam-cli==0.52.0
(…)
boto3==1.14.1
(…)

[Container] 2020/08/07 00:00:00 Running command python -m pip check
aws-sam-cli 0.52.0 has requirement boto3>=1.13.0,~=1.13.0, but you have boto3 1.14.1.

[Container] 2020/08/07 00:00:00 Command did not exit successfully python -m pip check exit status 1
[Container] 2020/08/07 00:00:00 Phase complete: INSTALL State: FAILED
[Container] 2020/08/07 00:00:00 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: python -m pip check. Reason: exit status 1
```

</details>

michael-k:
boto3 1.15 is out and, as expected, the bump wasn't due to a breaking change.  The reason for the bump was that botocore no longer depends on [`docutils`](https://pypi.org/project/docutils/).

https://github.com/boto/boto3/blob/1.15.1/CHANGELOG.rst#1150

michael-k:
@c2tarun you've reviewed and merged PR #1952.  Could I ask you to review this PR too?

CoshUS (changes requested):
Can you run `make update-reproducible-reqs` and update the pinned requirements?
CI will run with those requirements and can check whether the newest version of boto3 has a breaking change or not.

michael-k:
I didn't notice the review due to a relocation and holidays.  This has been fixed in aws/aws-sam-cli#2493.

Issue #2154, opened 2020-08-05
SAM deploy idempotency expectation?


heffergm:
### Description

Currently, working from the same codebase, running `sam deploy` on one machine and then again on another will trigger a full deployment workflow (function code updates, codedeploy deployment, etc) even though no code has changed.

### Steps to reproduce

On system X, run `sam build && sam deploy`; if there are changes pending, apply them. Re-run so that you confirm an empty changeset.

On system Y, using the same codebase, run `sam build && sam deploy` and a full stack update/changeset is created and deployed, despite no code changes. If applied and re-run, the changeset will again be empty.

Now go back to system X and repeat the process, and again there is a full stack changeset pending.

### Expected result

When there are no code/template (indeed, not any) changes, I would not expect that the origin point of the deploy process would trigger a complete stack update.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

`SAM CLI, version 1.0.0`

In my example above, system X is a local OSX machine, system Y is a Linux GitLab CI runner.

My only inclination is that this has something to do with the local .aws-sam cache.

timoschilling:
Which runtime do you use. I can't confirm you case for ruby for example.

heffergm:
Python3.7

heffergm:
These are moments apart. It's probably worth noting our setup is on the complex end of the spectrum (pre/post deploy hooks, multiple functions, etc). If I went ahead and applied the change locally, the next run on gitlab would result in a full changeset. However if I re-run from gitlab in this case (clean changeset first run) subsequent runs will remain clean. The same holds true for the reverse.

## my local setup
```
Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Invoke Function: sam local invoke
[*] Deploy: sam deploy --guided

Uploading to ourapp-prod-app/2212e79ccbcb3fbb6ee85a6f767b69e2  7362084 / 7362084.0  (100.00%)
Uploading to ourapp-prod-app/d7440d6438d501a69d79aeee146bec91  536556 / 536556.0  (100.00%)

	Deploying with following values
	===============================
	Stack name                 : ourapp-prod-app
	Region                     : us-east-2
	Confirm changeset          : True
	Deployment s3 bucket       : a-obfuscated-bucket
	Capabilities               : ["CAPABILITY_IAM"]
	Parameter overrides        : {'EnvType': 'prod'}

[...]

PR #2126, opened 2020-07-23
Update hash function to be location independent

anvar:
*Issue #, if available:*

https://github.com/awslabs/aws-sam-cli/issues/2110

*Why is this change necessary?*

AWS CodeBuild uses transient locations as root folders for a build. As the absolute path was used to calculated the directory hash it always resulted in a new hash thus prompting a redeploy even when no code change had been made.

*How does it address the issue?*

It uses the relative path of the directory instead of the absolute.

*What side effects does this change have?*

None

*Did you change a dependency in `requirements/base.txt`?*
    *If so, did you run `make update-reproducible-reqs`*

No

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [x] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2163, opened 2020-08-11
fix: md5 of directories are inconsistent without sorting the file paths

mndeveci:
*Issue #2154

Creating md5 of directories without sorting the file paths, do not generate the same result across different operating systems.

This change also contains the relative path change from this PR; https://github.com/awslabs/aws-sam-cli/pull/2126

Sorts the file paths before generating md5 of the directory

With this change running `build` and `package` will generate different md5 even though the project didn't change. But after this change it will be consistent

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/hash.py
```diff
@@ -51,13 +51,19 @@ def dir_checksum(directory, followlinks=True):
...
         # Go through every file in the directory and sub-directory.
         for filepath in [os.path.join(dirpath, filename) for filename in filenames]:
             # Look at filename and contents.
             # Encode file's checksum to be utf-8 and bytes.
-            md5_dir.update(filepath.encode("utf-8"))
-            filepath_checksum = file_checksum(filepath)
-            md5_dir.update(filepath_checksum.encode("utf-8"))
+            files.append(filepath)
+
+    files.sort()
```
sriram-mv, line 63: Great catch! I think the native behavior is different for `ls` on mac and linux, can we add a unit test for this? otherwise this looks good.

mndeveci: Thanks, added unit tests for sorting changes and from this PR: #2126

tests/unit/lib/utils/test_hash.py
```diff
@@ -13,6 +14,42 @@ def setUp(self):
...
+            f.write("Testfile")
+        checksum1 = dir_checksum(temp_dir1)
+
+        temp_dir2 = shutil.move(temp_dir1, os.path.join(self.temp_dir, "temp-dir-2"))
+        checksum2 = dir_checksum(temp_dir2)
+
+        self.assertEqual(checksum1, checksum2)
+
+    def test_dir_hash_independent_of_file_order(self):
+        file1 = tempfile.NamedTemporaryFile(delete=False, dir=self.temp_dir)
```
sriram-mv, line 30: small nit: can we name the files ourselves? this way we can showcase the sorting to work.

Because otherwise its non-determinstic, the names of the files will be different every run.

sriram-mv: we pass it both ways, and the name of the files can never be the same, so I rescind my nit. Approving.

PR #2166, opened 2020-08-11
Fix: Use Docker Python SDK Version 4.2.x Instead of 4.x

CoshUS:
*Issue #, if available:*
Docker Python SDK updated to version 4.3.0. Breaks support for docker API version 1.37.

~=4.0 installs latest docker sdk version of 4.x

Changed ~=4.0 to ~=4.2.0 which installs latest version within 4.2

Docker will not update to latest breaking changes. Manual update to requirements.txt is required.

Yes
Yes

*Checklist:*

- [x] Write unit tests
- [ ] `make pr` passes
- [x] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -7,6 +7,7 @@ jmespath~=0.9.5
 PyYAML~=5.1
 cookiecutter~=1.6.0
 aws-sam-translator==1.25.0
+#docker minor version updates can include breaking changes. Auto update micro version only.
 docker~=4.2.0
 dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -7,7 +7,7 @@ jmespath~=0.9.5
 PyYAML~=5.1
 cookiecutter~=1.6.0
 aws-sam-translator==1.25.0
-docker~=4.0
+docker~=4.2.0
 dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.23.0
--- a/requirements/reproducible-linux.txt
+++ b/requirements/reproducible-linux.txt
@@ -4,9 +4,9 @@
 #
 #    pip-compile --allow-unsafe --generate-hashes --output-file=requirements/reproducible-linux.txt
 #
-arrow==0.15.5 \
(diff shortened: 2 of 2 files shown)


requirements/base.txt
```diff
@@ -7,7 +7,7 @@ jmespath~=0.9.5
 PyYAML~=5.1
 cookiecutter~=1.6.0
 aws-sam-translator==1.25.0
-docker~=4.0
+docker~=4.2.0
```
jfuss, line 10: Can we comment on why we need to specify it this way? This will help us remember and communicate to any customers who might wonder why this is like this. We will need to figure out a path out of this but for now this is fine.

CoshUS: Added notice for docker minor version having breaking changes.

requirements/reproducible-linux.txt
```diff
@@ -200,9 +200,9 @@ serverlessrepo==0.1.9 \
     --hash=sha256:0c340d0e4437b5043eed2f2aafcb8fd6b16ab3d62ace19e70186542f4f7ac0f5 \
     --hash=sha256:7b58bd86f4ef1d0189fdaee17b7a322c59ef5bbf5373a3d2ceaf440886e35236 \
     # via aws-sam-cli (setup.py)
-six==1.14.0 \
-    --hash=sha256:236bdbdce46e6e6a3d61a337c0f8b763ca1e8717c03b369e87a7ec7ce1319c0a \
-    --hash=sha256:8f3cd2e254d8f793e7f3d6d9df77b92252b52637291d0f0da013c76ea2724b6c \
+six==1.15.0 \
+    --hash=sha256:30639c035cdb23534cd4aa2dd52c3bf48f06e5f4a941509c8bafd8ce11080259 \
+    --hash=sha256:8b74bedcbbbaca38ff6d7491d76f2b06b3592611af620f8426e82dddb04a5ced \
```
mndeveci, line 205: Just wondering why version of this library has changed, since this is not our direct dependency? Is it dependency of docker sdk?

sriram-mv: Yep: https://github.com/docker/docker-py/blob/master/requirements.txt

CoshUS: Yea, six is a dependency of docker sdk.

mndeveci: 👍

PR #2168, opened 2020-08-12
Add Support for Java8 and Provided AL2 Runtimes

awood45:


awood45:
We may see some integ test failures on container build tests, as the containers do not yet exist. They will before we release, so I'll have this open to re-run those failing tests as needed.

PR #2170, opened 2020-08-12
Fix: Permission Set to Default for Files Added to Docker Image When Building

CoshUS:
*Issue #, if available:*
https://github.com/awslabs/aws-sam-cli/issues/1014

Layer file permission is set to default(0666) if the docker image tar file is built with a system that's not unix(Windows).

Set all file permission to 0777 for files added to the docker image.

All files added to the docker image will have permission of 0777.

No

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_image.py
+++ b/samcli/local/docker/lambda_image.py
@@ -195,7 +195,13 @@ class LambdaImage:
             for layer in layers:
                 tar_paths[layer.codeuri] = "/" + layer.name
 
-            with create_tarball(tar_paths) as tarballfile:
+            # Set tar file permission to 777
+            # This is need for systems without unix like permission bits(Windows) while creating a unix image
+            def set_item_permission(tar_info):
+                tar_info.mode = 777
+                return tar_info
+
+            with create_tarball(tar_paths, filter=set_item_permission) as tarballfile:
                 try:
                     resp_stream = self.docker_client.api.build(
                         fileobj=tarballfile, custom_context=True, rm=True, tag=docker_tag, pull=not self.skip_pull_image
--- a/samcli/lib/utils/tar.py
+++ b/samcli/lib/utils/tar.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 
 
 @contextmanager
-def create_tarball(tar_paths):
+def create_tarball(tar_paths, filter=None):
     """
     Context Manger that creates the tarball of the Docker Context to use for building the image
 
@@ -25,7 +25,7 @@ def create_tarball(tar_paths):
 
     with tarfile.open(fileobj=tarballfile, mode="w") as archive:
         for path_on_system, path_in_tarball in tar_paths.items():
-            archive.add(path_on_system, arcname=path_in_tarball)
+            archive.add(path_on_system, arcname=path_in_tarball, filter=filter)
 
     # Flush are seek to the beginning of the file
     tarballfile.flush()


samcli/local/docker/lambda_image.py
```diff
@@ -197,7 +197,13 @@ def _build_image(self, base_image, docker_tag, layers, is_debug_go, stream=None)
...
                 tar_paths[layer.codeuri] = "/" + layer.name
 
-            with create_tarball(tar_paths) as tarballfile:
+            # Set tar file permission to 777
+            # This is need for systems without unix like permission bits(Windows) while creating a unix image
+            def set_item_permission(tar_info):
+                tar_info.mode = 777
+                return tar_info
+
+            with create_tarball(tar_paths, tar_filter=set_item_permission) as tarballfile:
```
mndeveci, line 206: Should we do this only for Windows? I was just wondering if this introduces any issues with other operating systems. Can we verify with other OS?

tests/unit/lib/utils/test_tar.py
```diff
@@ -19,7 +19,10 @@ def test_generating_tarball(self, temporary_file_patch, tarfile_open_patch):
 
         tarfile_file_mock.add.assert_called()
         tarfile_file_mock.add.assert_has_calls(
-            [call("/some/path", arcname="/layer1"), call("/some/dockerfile/path", arcname="/Dockerfile")],
+            [
+                call("/some/path", arcname="/layer1", filter=None),
+                call("/some/dockerfile/path", arcname="/Dockerfile", filter=None),
+            ],
```
mndeveci, line 25: It looks like these tests only cover the case with `filter=None`, can we add test case for the new use case?

samcli/local/docker/lambda_image.py
CoshUS: Updated to only set in Windows.

tests/unit/lib/utils/test_tar.py
CoshUS: Added test case with filter.

samcli/local/docker/lambda_image.py
```diff
@@ -197,7 +198,15 @@ def _build_image(self, base_image, docker_tag, layers, is_debug_go, stream=None)
             for layer in layers:
                 tar_paths[layer.codeuri] = "/" + layer.name
 
-            with create_tarball(tar_paths) as tarballfile:
+            # Set tar file permission to 777
+            # This is need for systems without unix like permission bits(Windows) while creating a unix image
+            def set_item_permission(tar_info):
+                tar_info.mode = 777
```
jfuss, line 204: This sets the permission on the host system and for every single file? If so, we should find another way to deal with this. 

Do we really need 777 on all files? Can we not scope this at all?

CoshUS: This sets permission for all the files in the tarball. File permission on the host is not changed.

tests/unit/lib/utils/test_tar.py
```diff
@@ -19,7 +19,42 @@ def test_generating_tarball(self, temporary_file_patch, tarfile_open_patch):
...
+    @patch("samcli.lib.utils.tar.TemporaryFile")
+    def test_generating_tarball_with_filter(self, temporary_file_patch, tarfile_open_patch):
+        temp_file_mock = Mock()
+        temporary_file_patch.return_value = temp_file_mock
+
+        tarfile_file_mock = Mock()
+        tarfile_open_patch.return_value.__enter__.return_value = tarfile_file_mock
+
+        def tar_filter(tar_info):
+            tar_info.mode = 0o500
```
jfuss, line 44: Can we add an assert that verifies the `tar_info.mode` for each file is set correctly?

CoshUS: tarfile_file_mock.add.assert_has_calls(..., filter=tar_filter) asserts that the filter must be used as parameter for tarfile library.
It will not be actually called in the unittest.

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -416,8 +416,8 @@ def test_invoke_returns_expected_results_from_git_function(self):
         process_stdout = stdout.strip()
         self.assertEqual(process_stdout.decode("utf-8"), '"git init passed"')
 
+    # For Windows, this test must run with administrator privilege
```
jfuss, line 419: Can you explain why this is?

CoshUS: Git layer zip file contains symbolic link. Python uses os.symlink which on newer versions of Windows 10 requires Developer Mode enabled, SeCreateSymbolicLinkPrivilege privilege, or running as Administrator.

CoshUS: Added Click based exception.

Issue #1940, opened 2020-04-24
Show error message when aws-sam-cli-managed-default CloudFormation stack not in CREATE_COMPLETE status


pyk:
### Describe your idea/feature/enhancement

I wish `sam-cli` show error message when there is something wrong in the default cloudformation stack.

Currently if the state of cloudformation stack is not complete it raise an key error, due to the `Outputs` object is not available (#1939).

If `aws-sam-cli-managed-default` CloudFormation stack is in bad state, the subsequent deploy command will always failed.

### Proposal

Show the cloudformation stack error message

awood45:
Guidance to implementers: We should also accept states like `UPDATE_COMPLETE`, and behavior when the stack is in a bad state may have to be set to a helpful error message (if we auto-deleted there is a risk of deleting uploaded code).

mndeveci:
In addition to Alex's comments, the state might look like it is in invalid state (ex: `DELETE_FAILED`) but it is actually healthy.

I accidentally tried to delete my `aws-sam-cli-managed-default` stack and it is now in `DELETE_FAILED` state. Which looks like an invalid one, but it is actually healthy, resources exists, and I can do my deployments. Looking at the [stack status codes](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-describing-stacks.html) it is hard to distinguish healthy stack from unhealthy one.

For that reason, instead of checking state of the stack, I [added validation of some information](https://github.com/awslabs/aws-sam-cli/pull/2171) that we expect from the stack (like, `Tags` and `Outputs` information).

mndeveci:
Released with [SAM CLI v1.2.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.2.0)

PR #2171, opened 2020-08-13
fix: add checks whether aws-sam-cli-managed-default contains valid resources for deployments

mndeveci:
*Issue #1940, if available:*

Deployments are failing if `aws-sam-cli-managed-default` is in invalid state

Check if stack contains correct information and resources

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/bootstrap/bootstrap.py
```diff
@@ -83,6 +66,42 @@ def _create_or_get_stack(cloudformation_client):
...
+        msg = (
+            f"Stack {SAM_CLI_STACK_NAME} is missing Tags and/or Outputs information and therefore not in a "
+            f"healthy state (Current state:{stack_state}). Failing as the stack was likely not created "
+            f"by the AWS SAM CLI"
+        )
+        raise UserException(msg)
+
+    # Sanity check for non-none stack? Sanity check for tag?
+    try:
+        sam_cli_tag = next(t for t in tags if t["Key"] == "ManagedStackSource")
```
c2tarun, line 86: I am guessing you are getting first value here. Probably pass a default parameter to `next` if `ManagedStackSource` is missing.

c2tarun: Nevermind I noticed late that you are catching StopIteration.

mndeveci: This was the previous code actually, I just moved it another method :)

PR #2174, opened 2020-08-14
Fix: string params starting with '0'

prenx4x:
*Issue #, if available:*
[200](https://github.com/awslabs/serverless-application-model/issues/200)

Templates with string params leading with '0' are converted to integers. This patch fixes this issues.

Check output template and put quotes around such params

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/regression/deploy/test_deploy_regression.py
```diff
@@ -67,6 +67,7 @@ def test_deploy_with_all_args(self, template_file):
             "aws_stack_name": aws_stack_name,
             "sam_stack_name": sam_stack_name,
             "capabilities": "CAPABILITY_IAM",
+            "s3_bucket": self.s3_bucket.name,
```
jfuss, line 70: Curious, why did you add these? Trying to understand why this change required these to be added to the tests.

tests/regression/package/test_package_regression.py
```diff
@@ -26,9 +26,7 @@ def tearDown(self):
             "aws-appsync-graphqlschema.yaml",
             "aws-appsync-resolver.yaml",
             "aws-appsync-functionconfiguration.yaml",
-            "aws-apigateway-restapi.yaml",
```
jfuss, line 29: Can you explain why these were removed?

It's ok to remove them but I want to ensure removing them does not loose coverage in testing unexpectedly.

tests/integration/testdata/package/aws-serverlessrepo-application.yaml
```diff
@@ -12,5 +12,5 @@ Metadata:
     ReadmeUrl: ./README.md
     Labels: ['tests']
     HomePageUrl: https://github.com/user1/my-app-project
-    SemanticVersion: 0.0.1
+    SemanticVersion: 1.0.1
```
jfuss, line 15: Why shouldn't we leave this as `0.0.1`?

tests/regression/deploy/test_deploy_regression.py
prenx4x: So when I was working on this, we had the resolve-s3 feature merged, which needed either s3_bucket or -resolve-s3 parameter for deploy. But since last week, I pushed another PR which removed this check in deploy as it was not needed in deploy (only needed in package), which got merged yesterday. Thus I can now remove this as we don't need it.

tests/regression/package/test_package_regression.py
prenx4x: For this test, the hash generated inside SAM has a leading zero which is causing it to fail. I could maybe modify the value of the property a bit so we have a different hash (hopefully without leading zeros) and see if that works.

tests/integration/testdata/package/aws-serverlessrepo-application.yaml
prenx4x: this is failing because our output is '0.0.1' and awscli's output is 0.0.1 which is failing. Thus I changed the leading 0 to 1 so we don't wrap it in quotes.

tests/regression/package/test_package_regression.py
prenx4x: Tried this and modified the SemanticVersion property. Now there is no value generated with leading zero. As far as I can tell, this changes does not affect the crux of the tests.

tests/integration/testdata/package/samconfig.toml
```diff
@@ -0,0 +1,10 @@
+version = 0.1
+[default]
+[default.deploy]
+[default.deploy.parameters]
+stack_name = "test-deploy-guided-set-parameter-0-aws-serverless-function-yaml-0-0-0"
```
jfuss, line 5: I don't think we should be committing this. The current tests should not require this. Committing this, runs the risk of other tests using these defaults and failing. 

For example, there can only be one S3 Bucket with a given name (names are global for S3). Hardcoding a bucket (event in a config file like this), will cause seemingly random failures for our package tests if someone other than you tries to run the tests.

prenx4x: thats right. this file shouldn't be here.

jfuss (approved):

Issue #1615, opened 2019-11-28
sam deploy command option for different config file


ericallam:
### Describe your idea/feature/enhancement

I'd like to be able to have multiple different samconfig.toml files for deploying to different environments in my CI/CD process.

### Proposal

Add an option flag to `sam deploy` that allows a user to specify a path to a `samconfig.toml` path, like so:

```bash
$ sam deploy --config-file ./samconfig-dev.toml
```

The guided mode should also be updated to ask for the path name, possible right after the `Save arguments to samconfig.toml:` input step.

sanathkr:
This is a good feature request. We had made the explicit decision to skip this to support more automatic resolution mechanisms of the config file.

What do you think about an automatic resolution mechanism where you specify the env `sam deploy --env dev` and SAM CLI automatically chooses `samconfig-<env>.toml`?

ericallam:
@sanathkr That would work for my use case 👍

rowanu:
I like to have multiple stacks in my application repo (e.g. splitting front-end and back-end resources), so I don't think the `--env` approach works.

Really I just want to be able to support a custom configuration file name, so that I can specify where to load settings from, just like I would when specifying different template files - it feels like there is always going to be a 1:1 mapping between template file and config file, so why not support specifying them the same way?

benkehoe:
I agree with @rowanu. I'd rather start with a simple generic mechanism, `--config-file`, that can enable all sorts of use cases. Then we can see what people use it for, and build an opinionated version (like `--env`) based on that.

benkehoe:
Additionally, with the ability to use a separate configuration file should come the ability for "guided" mode to not actually deploy but write the values to a new config file location.

benkehoe:
I need multiple configs for all of: 
1. the same stack copied to multiple accounts
2. a slightly different stack in dev vs test vs prod
3. a set of different stacks together in the same place comprising an application

I can accomplish all of them with `--config-file`.

jdonboch:
I would also be in favor of being able to define multiple profiles inside of a single `samconfig.toml` similar to AWS credentials file.   There are already namespaes in there for `[default]`, why not also support other profile names that you could choose to override the default.

Does this belong in a separate issue?

darylounet:
It seems there is a hidden feature (or at least not yet documented) [here](https://github.com/awslabs/aws-sam-cli/blob/master/samcli/cli/cli_config_file.py#L163) with "--config-env" :
```toml
[...]

PR #2176, opened 2020-08-19
feature: Enable --config-env and --config-file to support multiple configuration environments and files

wchengru:
*Issue #, if available:*
#1615 

sam deploy option "--config-env" was not enabled. I enabled this option and allow samconfig.toml to save/load multiple configuration profiles.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/cli/cli_config_file.py
```diff
@@ -31,42 +31,60 @@ class TomlProvider:
...
         :param cmd_names list(str): sam command name as defined by click
         :returns dictionary containing the configuration parameters under specified config_env
         """
 
         resolved_config = {}
 
-        samconfig = SamConfig(config_dir)
+        # Use default sam config file name if config_path only contain the directory
+        samconfig = (
+            SamConfig(os.path.split(config_path)[0], os.path.split(config_path)[1])
```
sriram-mv, line 50: you can probably use pathlib for this.
pathlib.parent should give you the directory without needing to do splits.

samcli/cli/cli_config_file.py
```diff
@@ -31,42 +31,60 @@ class TomlProvider:
...
+            if config_path
+            else SamConfig(os.getcwd())
+        )
+        # Enable debug level logging by environment variable "SAM_DEBUG"
+        if os.environ.get("SAM_DEBUG", "").lower() == "true":
+            LOG.setLevel(10)
```
sriram-mv, line 56: What is 10? can we use ENUMs instead?

samcli/cli/cli_config_file.py
```diff
@@ -145,44 +165,88 @@ def configuration_option(*param_decls, **attrs):
...
-        attrs.setdefault("is_eager", True)
-        attrs.setdefault("help", "Read config-env from Configuration File.")
-        attrs.setdefault("expose_value", False)
-        # --config-env is hidden and can potentially be opened up in the future.
-        attrs.setdefault("hidden", True)
-        # explicitly ignore values passed to --config-env, can be opened up in the future.
-        config_env_name = DEFAULT_ENV
+
+    def decorator_configuration_setup(f):
+        configuration_setup_param_devls = ()
```
sriram-mv, line 185: what does `configuration_setup_param_devls` mean, the meaning is not apparent from the name.

samcli/commands/deploy/command.py
```diff
@@ -42,7 +42,7 @@
     context_settings={"ignore_unknown_options": False, "allow_interspersed_args": True, "allow_extra_args": True},
     help=HELP_TEXT,
 )
-@configuration_option(provider=TomlProvider(section=CONFIG_SECTION))
+@configuration_option(provider=TomlProvider(section=CONFIG_SECTION),)
```
sriram-mv, line 45: is the extra comma required?

samcli/lib/config/samconfig.py
```diff
@@ -106,7 +106,16 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
         if not self.document:
             self._read()
             # Empty document prepare the initial structure.
-            self.document.update({env: {self._to_key(cmd_names): {section: {key: value}}}})
+            # self.document is a nested dict, we need to check each layer and add new tables, otherwise duplicated key
+            # in parent layer will override the whole child layer
+            if self.document.get(env, None):
```
sriram-mv, line 111: is there a testcase to showcase all the cases shown here?

tests/unit/commands/deploy/test_guided_context.py
```diff
@@ -121,3 +123,45 @@ def test_guided_prompts_with_given_capabilities(
             call(f"\t{self.gc.start_bold}Capabilities{self.gc.end_bold}", default=expected_capabilities, type=ANY),
         ]
         self.assertEqual(expected_prompt_calls, patched_prompt.call_args_list)
+
+    @patch("samcli.commands.deploy.guided_context.prompt")
+    @patch("samcli.commands.deploy.guided_context.confirm")
+    @patch("samcli.commands.deploy.guided_context.manage_stack")
+    @patch("samcli.commands.deploy.guided_context.auth_per_resource")
+    @patch("samcli.commands.deploy.guided_context.get_template_data")
+    def test_guided_prompts_check_defaults_public_resources(
```
sriram-mv, line 132: should the test name be any different?

sriram-mv (changes requested):

samcli/commands/deploy/command.py
wchengru: I is a typo and have fixed. Thanks for pointing it out.

samcli/cli/cli_config_file.py
wchengru: It is the parameters of this decorator, as it does not accept any parameters from command line. I put it empty. The name has been changed to "configuration_setup_params".

samcli/cli/cli_config_file.py
wchengru: Changed it to loggings.DEBUG.

samcli/lib/config/samconfig.py
wchengru: Yes, they are tested in test_samconfig.py.

tests/unit/commands/deploy/test_guided_context.py
wchengru: This has been fixed.

samcli/cli/cli_config_file.py
wchengru: Contributer of SamConfig() is using strings of directory and filename. Here I think to pass in strings would be consistent.

wchengru: I reconsidered this and change to use pathlib to avoid some edge cases.

tests/unit/commands/_utils/test_options.py
```diff
@@ -83,6 +83,15 @@ def test_verify_ctx(self, get_template_data_mock, os_mock):
         self.assertEqual(ctx.template_dict, "dummy_template_dict")
         os_mock.path.abspath.assert_called_with(expected)
 
+    def test_verify_ctx_template_file_param(self):
```
sriram-mv, line 87: can we add a specific test for this in test_samconfig.py file

sriram-mv (changes requested):

tests/unit/commands/_utils/test_options.py
wchengru: sure, new test case is added in that file

faboulaye:
When this issue will be available ?

tests/unit/cli/test_cli_config_file.py
```diff
@@ -26,33 +27,41 @@ def setUp(self):
...
         self.assertEqual(self.toml_provider(config_dir, self.config_env, [self.cmd_name]), {})
 
+    def test_toml_invalid_file_name(self):
+        config_dir = tempfile.gettempdir()
+        config_path = Path(config_dir, "mysamconfig.toml")
+        config_path.write_text("version=0.1\n[config_env.topic.parameters]\nword='clarity'\n")
+        config_path_invalid = Path(config_dir, "samconfig.toml")
+
+        self.assertEqual(self.toml_provider(config_path_invalid, self.config_env, [self.cmd_name]), {})
+
```
awood45, line 65: Can we add a test with an absolute path? May need to have another function call generate it so it works on whatever machine runs this test, but I didn't notice that case being covered.

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -714,6 +718,27 @@ def test_override_with_cli_params_and_envvars(self, do_cli_mock):
                 {"A": "123", "C": "D", "E": "F12!", "G": "H"},
             )
 
+    @patch("samcli.commands.validate.validate.do_cli")
+    def test_secondary_option_name_template_validate(self, do_cli_mock):
+        # "--tempalte" is an alias of "--template-file"
```
awood45, line 723: s/tempalte/template

awood45 (approved):
Looks good, ideally we can add one more test and there's a comment that has a typo.

tests/unit/commands/samconfig/test_samconfig.py
wchengru: Will fix this in later PRs.

tests/unit/cli/test_cli_config_file.py
wchengru: I think we have a temp directory generated by `config_dir = tempfile.gettempdir()`, and that will be the absolute path.

sriram-mv (approved):

PR #2178, opened 2020-08-20
Stop showing 'Failed to download ...' when using local rapid docker images

wchengru:
*Issue #, if available:*
#2165 

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/local/docker/manager.py
+++ b/samcli/local/docker/manager.py
@@ -8,6 +8,7 @@ import sys
 import docker
 import requests
 
+from samcli import __version__ as SAM_CLI_VERSION
 from samcli.lib.utils.stream_writer import StreamWriter
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


samcli/local/docker/manager.py
```diff
@@ -152,6 +157,21 @@ def has_image(self, image_name):
...
 
+    def _is_rapid_image(self, image_name):
+        """
+        Is the image tagged as a RAPID clone?
+
+        : param string image_name: Name of the image
+        : return bool: True, if the image name ends with rapid-$SAM_CLI_VERSION. False, otherwise
+        """
+
+        rapid_suffix = f"rapid-{SAM_CLI_VERSION}"
```
awood45, line 168: Do we possibly just want to match on a regular expression for `rapid-\d+\.\d+.\d+$` to match any version? In practice it should usually be the SAM CLI version, but if we ever allow some sort of selection scheme this could fail.

wchengru: Updated this part to use regular expression.

samcli/local/docker/manager.py
```diff
@@ -152,6 +157,18 @@ def has_image(self, image_name):
...
 
+    def _is_rapid_image(self, image_name):
+        """
+        Is the image tagged as a RAPID clone?
+
+        : param string image_name: Name of the image
+        : return bool: True, if the image name ends with rapid-$SAM_CLI_VERSION. False, otherwise
+        """
+
+        if not re.search(r"rapid-\d+\.\d+.\d+$", image_name):
```
sriram-mv, line 168: it is tagged as :rapid-SAMCLIVersion right? should we include the ":"

sriram-mv: because we are checking the image tag, not the image itself.

wchengru: Changed this to match tag name only.

Issue #1291, opened 2019-07-25
Less verbose logging


zafarella:
### Less verbose logging of the progress - reducing noise
the sam cli should have an option to be less verbose or just be less verbose but provide a flag to be more verbose.

### Proposal

For instance, `sam package` command print progress like:
```
Uploading to 6c10f7319c36f4474bddc098cc2ea299  262144 / 30227673.0  (0.87%)
Uploading to 6c10f7319c36f4474bddc098cc2ea299  524288 / 30227673.0  (1.73%)
Uploading to 6c10f7319c36f4474bddc098cc2ea299  786432 / 30227673.0  (2.60%)
...
...
```
which is not really necessary in the CI environment.

awood45:
Noted as a feature request - also happy to take a PR for this if anyone wants to take this on as a first issue.

qingchm:
Implemented and released

PR #2187, opened 2020-08-27
feat: Added the option to hide progress bars during s3 upload for sim…

qingchm:
…plicity through --no-progressbar

*Issue #1291*

*Why is this change necessary? *
Customers using CI/CD environment doesn't want to see the progress bars when uploading to s3

Added an option "--no-progressbar" to have the choice of not printing the progress bars

Not known so far

No

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/_utils/options.py
```diff
@@ -162,6 +162,22 @@ def parameter_override_option(f):
     return parameter_override_click_option()(f)
 
 
+def no_progressbar_click_option():
+    return click.option(
+        "--no-progressbar",
+        default=False,
+        required=False,
+        is_flag=True,
+        help="Indicates whether to simplify the logging "
```
sriram-mv, line 171: help section does need to go into implementation details. it can simply say

```
Does not showcase a progress bar when uploading artifacts to s3
```

samcli/lib/package/s3_uploader.py
```diff
@@ -91,12 +92,18 @@ def upload(self, file_name, remote_path):
             if self.artifact_metadata:
                 additional_args["Metadata"] = self.artifact_metadata
 
-            print_progress_callback = ProgressPercentage(file_name, remote_path)
+            if not self.no_progressbar:
+                print_progress_callback = ProgressPercentage(file_name, remote_path)
```
sriram-mv, line 96: both this line and line 102 can go under the same if condition from line 95.

jfuss:
Have we considered adding an Env Var instead of a CLI option to control this behavior? 

To me, a 'no progress bar' flag requires us to change this in every command. If we instead use an Env Var, customers will have the ability to control this across the whole CLI. I think the intention from customers is much broader than a single command, especially as the CLI grows.

qingchm:
> Have we considered adding an Env Var instead of a CLI option to control this behavior?
> 
> To me, a 'no progress bar' flag requires us to change this in every command. If we instead use an Env Var, customers will have the ability to control this across the whole CLI. I think the intention from customers is much broader than a single command, especially as the CLI grows.

That's a good suggestion, maybe let's have a discussion on how to implement that if possible?

qingchm:
Discussed with Jacob and we decided to move ahead with adding command line option for this issue, since this no progress bar option is only for deploy and package commands, setting an env var sounds like a more general idea. Might create a env var for hiding all progress bars in the future if customers like this idea. For now it'll be a command line option since it's merely for deploy and package commands.

sriram-mv (changes requested):
This looks good, can we add unit test to actually see that there is no progress bar in stderr stream?

qingchm:
> This looks good, can we add unit test to actually see that there is no progress bar in stderr stream?

Added a set of integration tests into package integration test

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -85,17 +85,17 @@ def _extract_cloud_formation_route(self, logical_id, api_resource, collector, cw
     @staticmethod
     def _extract_cloud_formation_stage(resources, stage_resource, collector):
         """
-        Extract the stage from AWS::ApiGateway::Stage resource by reading and adds it to the collector.
-        Parameters
-       ----------
-        resources: dict
-            All Resource definition, including its properties
+         Extract the stage from AWS::ApiGateway::Stage resource by reading and adds it to the collector.
```
sriram-mv, line 88: is there some formatting issues here?

tests/integration/package/test_package_command.py
```diff
@@ -440,6 +440,39 @@ def test_package_with_resolve_s3(self, template_file):
...
+        template_path = self.test_data_path.joinpath("aws-serverless-function.yaml")
+        s3_prefix = "integ_test_prefix"
+
+        with tempfile.NamedTemporaryFile(delete=False) as output_template:
+            command_list = self.get_command_list(
+                template_file=template_path,
+                s3_prefix=s3_prefix,
+                output_template_file=output_template.name,
+                force_upload=True,
+                no_progressbar=bool(int(no_progressbar)),
```
sriram-mv, line 454: you can just pass in True, False instead of "1", "0"

sriram-mv (changes requested):

samcli/lib/providers/cfn_api_provider.py
qingchm: It was a forced reformatting by black 20, but since I switched back to black 19 I don't think either one will cause pr problems. but i guess i'll just leave it this way

tests/integration/package/test_package_command.py
qingchm: @sriram-mv Changed this in the latest commit

sriram-mv (approved):

PR #2190, opened 2020-08-28
fix: template warning

c2tarun:
*Issue #, if available:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -54,6 +54,7 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
         # to be .aws-sam/build/
         setattr(ctx, "samconfig_dir", os.path.dirname(original_template_path))
         try:
+            # FIX-ME: figure out a way to insert this directly to sam-cli context and not use click context.
             template_data = get_template_data(result)
             setattr(ctx, "template_dict", template_data)
         except TemplateNotFoundException:
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -8,13 +8,13 @@ from functools import partial
 
 import click
 from click.types import FuncParamType
+
+from samcli.commands._utils.template import get_template_data, TemplateNotFoundException
 from samcli.cli.types import CfnParameterOverridesType, CfnMetadataType, CfnTags
 from samcli.commands._utils.custom_options.option_nargs import OptionNargs
 
-
 _TEMPLATE_OPTION_DEFAULT_VALUE = "template.[yaml|yml]"
 DEFAULT_STACK_NAME = "sam-app"
-
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


samcli/commands/_utils/options.py
```diff
@@ -53,6 +53,13 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
         # sam configuration file should always be relative to the supplied original template and should not to be set
         # to be .aws-sam/build/
         setattr(ctx, "samconfig_dir", os.path.dirname(original_template_path))
+        try:
+            template_data = get_template_data(result)
+            setattr(ctx, "template_dict", template_data)
```
jfuss, line 58: We should not be doing setattr directly. Can we not follow the same pattern as profile or region which have explicit setter methods?

sriram-mv: https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/options.py#L33, something like that makes sense.

samcli/lib/warnings/sam_cli_warning.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Top level class which all warnings should extend from.
+    """
+
+    def check(self, template_dict):
+        raise Exception("NotImplementedException")
+
+
+def _get_deployment_preferences_status(function):
+    deployment_preference = function["Properties"].get("DeploymentPreference", None)
```
sriram-mv, line 19: can we add a docstring for this?

samcli/lib/warnings/sam_cli_warning.py
```diff
@@ -0,0 +1,85 @@
...
+    def check(self, template_dict):
+        raise Exception("NotImplementedException")
+
+
+def _get_deployment_preferences_status(function):
+    """
+    Takes a AWS::Serverless::Function resource and checks if resource have a deployment preferences applied
+    to it. If DeploymentPreference found then it returns its status if it is enabled or not.
+    """
+    deployment_preference = function["Properties"].get("DeploymentPreference", None)
```
sriram-mv, line 23: should we do function.get("Properties", {}).get("DeploymentPreference", None) just to avoid failures.

sriram-mv (commented):
This looks okay to me, we should come back and aim to fix the line you mentioned. we are now making two calls for telemetry instead of one, I dont think it will cause issues becuase timeouts are built-in. Can we check that there is no performance impact?

samcli/commands/_utils/options.py
```diff
@@ -53,6 +53,14 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
         # sam configuration file should always be relative to the supplied original template and should not to be set
         # to be .aws-sam/build/
         setattr(ctx, "samconfig_dir", os.path.dirname(original_template_path))
+        try:
+            # FIX-ME: figure out a way to insert this directly to sam-cli context and not use click context.
```
awood45, line 57: Can we close out the FIXME here?

awood45 (changes requested):
This generally looks good - can we either resolve or remove the `FIX-ME` item?

awood45 (approved):

PR #2193, opened 2020-08-31
Fix: Missing Unix Permission Bits for Artifact ZIP Created in Windows

CoshUS:
Original PR: 
https://github.com/aws/aws-sam-cli/pull/1986

*Issue:*
https://github.com/aws/aws-sam-cli/issues/954

ZIP created in Windows will have no Unix permission bits and host OS is set to Windows.
When unzipping in Unix system, like lambda function, permission is set to default 0666.

 Lambda function needs execution permission bit set for "global/other".
Permission bits will be set to 0755 in Windows.
Host OS will also be changed to Unix so the permission will be preserved while unzipping.

All artifact ZIP files created in Windows will be have 0755 permission.

No

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/package/artifact_exporter.py
```diff
@@ -203,7 +204,19 @@ def make_zip(file_name, source_root):
...
                     full_path = os.path.join(root, filename)
                     relative_path = os.path.relpath(full_path, source_root)
-                    zf.write(full_path, relative_path)
+                    if platform.system().lower() == "windows":
+                        with open(full_path, "rb") as data:
+                            file_bytes = data.read()
+                            info = zipfile.ZipInfo(relative_path)
+                            # Clear external attr set for Windows
+                            info.external_attr = 0
+                            # Set external attr with Unix 0755 permission
```
jfuss, line 213: How did we select to use 755 permissions for this?

CoshUS: This is set to match the behavior on Ubuntu which sets the permission to 755.

jfuss: Why are we matching Ubuntu here? We should be setting these permissions to the min required for Lambda. We should validate but I think it only requires 005.

CoshUS: Changed to 005 and added unit tests.

CoshUS: After doing more investigation on this matter, I think we should move forward with setting all file permissions in the artifact to 005 in Windows.

In Lambda, file owner and group is set to "slicer" and "933", and execution user of the function is not "slicer" or in that group. Setting the first two(user and group) permissions does not matter.

Changing from the default 644(equivalent to 004) to 005 when packaging in Windows will not be a breaking change as it only adds more permission. Go is the only language gets affected by this issue since it build a binary executable instead of requiring only read permission like others (java .class, python .py, C# .json and .dll, etc).

In terms of whether we should allow write permission on the artifacts, I could not find any documentation on modifying artifact files at runtime. Most of the aws documentation states 644 or 755 which means this use case is either not supported or very rarely encountered.

jfuss (approved):
🎉

aahung (approved):

PR #2201, opened 2020-09-03
chore: bump lambda-builder version v1.1.0 and sam-cli version v1.2.0.

mingkun2020:

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,5 +13,5 @@ dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.23.0
 serverlessrepo==0.1.9
-aws_lambda_builders==1.0.0
+aws_lambda_builders==1.1.0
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -13,5 +13,5 @@ dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.23.0
 serverlessrepo==0.1.9
-aws_lambda_builders==1.0.0
+aws_lambda_builders==1.1.0
```
c2tarun, line 16: You need to update requirements file.

c2tarun (changes requested):

requirements/base.txt
c2tarun: Check https://github.com/aws/aws-sam-cli/blob/develop/Makefile#L51

mingkun2020: updated

Issue #998, opened 2019-02-11
Locally debugging API Gateway (start-api  -d) can't be terminated from Eclipse


phijacobs:
### Description

When locally debugging with the API Gateway (sam local start-api -d), the API Gateway process can't be stopped from the IDE. SIGTERM is caught to properly to clean up the running container, but it isn't propagated to terminate the Python process running sam local.

Once the API Gateway request completes, the SIGTERM handler is still in effect. The SIGTERM handler needs to be set back to the default handler after the API Gateway request completes. The IDE can't stop the API Gateway process after a request is completed and the debugger detaches.

### Steps to reproduce

1. Test API Gateway locally as documented at https://docs.aws.amazon.com/toolkit-for-eclipse/v1/user-guide/tke-sam-local.html
2. After debugging the request, stop the running process using the stop button, and you'll get a Terminate failed.

### Observed result

Please provide command output with `--debug` flag set.

### Expected result

Describe what you expected.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac OS X 10.11.6 El Capitan
2. `sam --version`: SAM CLI, version 0.11.0

`Add --debug flag to command you are running`

morefaey:
@phijacobs can you please submit this as a PR for review?

phijacobs:
@morefaey I rebased and will have the PR shortly. Thank you!

phijacobs:
Hi @morefaey, I created PR #2205

CoshUS:
Closing as stated in https://github.com/aws/aws-sam-cli/pull/2205#issuecomment-687976761
Debugger should be attached with `sam local invoke` instead of `sam local start-api`.

PR #2205, opened 2020-09-05
PR: Locally debugging API Gateway (start-api -d) can't be terminated from Eclipse

phijacobs:
*Issue #, if available:*
#998 
The SIGTERM signal handler wasn't being propagated to stop all threads in sam. Once the signal handler stops the docker container, the signal handler is set to default and the signal is fired again.

Upon terminating sam using an IDE debugger, it no longer leaves the python program running as a child process of the IDE.

None that I know of.

No

*Checklist:*

- [ X] Write unit tests
- [ X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,5 +13,5 @@ dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.23.0
 serverlessrepo==0.1.9
-aws_lambda_builders==1.0.0
+aws_lambda_builders==1.1.0
 tomlkit==0.5.8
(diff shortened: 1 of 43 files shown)


jfuss:
@phijacobs The current experience is actually by design. The first SIGTERM handles killing the running function, the second (or the first if a function is not running) will kill the whole process. `sam local start-api` is meant to be run once and then used to trigger different functions, update your code, and then re-trigger the the function again without needed to restart the process. This breaks that behavior.

If you want to debug the function directly, the way to do this would be through `sam local invoke` and pass it an event that you want to test.

CoshUS:
Closing due to the reason above.

Issue #758, opened 2018-11-12
Support for custom message via --message for sqs receive message generation


terlar:
Feature request:
Just like you can specify a custom message for `sam local generate-event sns notification`. It would be nice to be able to do the same for `sam local generate-event sqs receive-message`.

E.g.:
```
sam local generate-event sqs receive-message --message '{"custom": "message"}'
```

kvernon:
or body

PR #2209, opened 2020-09-09
feat: Add support for customized message body in generated SQS message

hawflau:
*Issue #, if available:*
#758 

Customer wants to be able to customize message body when generating an SQS event

To support customized `body` in the SQS was straightforward, but customizing `body` would also trigger an update in `md5OfBody`. (see [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html))
I introduced an additional `children` attribute in `event-mapping.json` to describe such relation, and made corresponding update in `events.py` to consume the `children` attribute. I also added supported for calculating the md5 digest of `body`.

No, the change is mostly contained with the `Events` class in `samcli/lib/generated_sample_events/events.py` and the class is only used in `samcli/commands/local/generate_event/event_generation.py`

No 
*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/generated_sample_events/events.py
+++ b/samcli/lib/generated_sample_events/events.py
@@ -5,9 +5,9 @@ Methods to expose the event types and generate the event jsons for use in SAM CL
 import os
 import json
 import base64
-from requests.utils import quote
-
+from requests.utils import quote as url_quote
 from chevron import renderer
+from samcli.lib.utils.hash import str_checksum
 
 
 class Events:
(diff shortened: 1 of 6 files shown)


samcli/lib/generated_sample_events/events.py
```diff
@@ -33,63 +33,106 @@ def __init__(self):
...
         -------
-        string: the base64_utf encoded value
+        hashed: string
+            the hashed value
         """
+        if hashing_scheme == "md5":
+            return str_checksum(val)
 
-        return base64.b64encode(value.encode("utf8")).decode("utf-8")
+        # returns original val if hashing_scheme not recognized
```
awood45, line 134: It's consistent with previous behavior, but is this what we want to do? It seems to me that at minimum we should log a warning, and perhaps we should raise an exception in this case. For our pre-baked events, this exception should never occur and we would want an explicit failure and for someone to raise an issue if it does.

awood45 (changes requested):
I really like the refactor and testing scheme here, this is a well done change. I do have a suggested change to make, but this looks like a quality change.

samcli/lib/generated_sample_events/events.py
hawflau: yep it makes sense to raise an exception. Will update this part.

awood45 (approved):

PR #2211, opened 2020-09-10
Add "Build All Images" Script

awood45:
Could be useful for local development, for example. Should be kept up to date as new runtimes are launched.


build-image-src/build_all_images.sh
```diff
@@ -0,0 +1,17 @@
+# You can use this to build any changes you make to Docker build images to your local machine.
+# Of course, you can also run a single one of these commands manually.
+# If you use this script, ensure that you run with --skip-pull-image, else the remote image may be used.
+
+docker build -f Dockerfile-nodejs10x -t amazon/aws-sam-cli-build-image-nodejs10.x --build-arg SAM_CLI_VERSION=$SAM_CLI_VERSION .
```
sriram-mv, line 5: can we add a comment to set the SAM_CLI_VERSION env variable?

sriram-mv (changes requested):
one nit, otherwise looks okay

sriram-mv (approved):

build-image-src/build_all_images.sh
awood45: Added a check to the script instead.

Issue #2210, opened 2020-09-10
sam deploy is showing parameters with NoEcho: true


kstro21:
### Description

If I add a parameter to my template and set the `NoEcho` option to `true`, then, when I run `sam deploy` with `--parameter-overrides` I can see the value of the parameter in the command output.

### Steps to reproduce

1. Add a parameter with `NoEcho: true`
   ```yaml
   MyParam:
      Type: String
      Description: My description
      NoEcho: true
   ```
2. Invoke `sam deploy` like this:
   ```bash
   sam deploy --stack-name my-stack --parameter-overrides MyParam=$SOME_ENV_VAR \
    --region $AWS_REGION --s3-bucket $FZ_BUCKET_NAME \
    --s3-prefix my-app-prefix --capabilities CAPABILITY_IAM --force-upload
   ```

### Observed result

```bash

Deploying with following values
--
253 | ===============================
254 | Stack name                 : my-stack
255 | Region                     : us-east-1
256 | Confirm changeset          : False
257 | Deployment s3 bucket       : my-bucket
258 | Capabilities               : ["CAPABILITY_IAM"]
259 | Parameter overrides        : {'MyParam': 'ShouldntSeeThis'}
```

### Expected result

I shouldn't see parameters with `NoEcho: true` in the cli output. It is ok if I'm local, but I don't want those values in my CI/CD pipeline's log

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. `sam --version`: 1.2.0

vaunus:
+1 am also seeing the same issue. Would be good to get this merged in.

aahung:
it is released in v1.16.0, please reopen if the issue still persists

snoblenet:
I am experiencing this issue with SAM CLI v1.20.0

PR #2212, opened 2020-09-11
fix: hide NoEcho parameters in sam deploy console output

wchengru:
*Issue #, if available:*
#2210 

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


kstro21:
Is this going to be merged any time in the near future?

samcli/commands/deploy/utils.py
```diff
@@ -25,15 +27,17 @@ def print_deploy_args(stack_name, s3_bucket, region, capabilities, parameter_ove
...
+    :param template_params: Parameters from template, need to hide "NoEcho" paramters
     :param parameter_overrides: Cloudformation parameter overrides to be supplied based on the stack's template
     :param confirm_changeset: Prompt for changeset to be confirmed before going ahead with the deploy.
     :return:
     """
 
     _parameters = parameter_overrides.copy()
     for key, value in _parameters.items():
-        if isinstance(value, dict):
-            _parameters[key] = value.get("Value", value) if not value.get("Hidden") else "*" * len(value.get("Value"))
```
moelasmar, line 36: I think you maintain the same logic.
In your change, you neglect if the Parameter value is a dict not just a simple string

moelasmar (changes requested):
Can you add unit/integration test cases that cover this change

samcli/commands/deploy/utils.py
wchengru: Thanks. This has been updated to use only 5 stars in case value is a dict.

sriram-mv (approved):

aahung (approved):

moelasmar (approved):

samcli/commands/deploy/utils.py
```diff
@@ -89,3 +87,13 @@ def sanitize_parameter_overrides(parameter_overrides):
     :return:
     """
     return {key: value.get("Value") if isinstance(value, dict) else value for key, value in parameter_overrides.items()}
+
+
+def hide_noecho_parameter_overrides(template_parameters, parameter_overrides):
+    hidden_params = copy.deepcopy(parameter_overrides)
+    params = template_parameters.get("Parameters", None)
+    for key, value in hidden_params.items():
+        if isinstance(params, dict) and key in params and isinstance(params[key], dict):
```
elbayaaa, line 96: `key in params` will raise a "TypeError" if params is None?

elbayaaa, line 96: Is it better to:
1. Move `isinstance(params, dict)` out of the loop, something like 
```
if not isinstance(params, dict):
  return hidden_params
for ...
```
2. better to remove the check `key in params` and use `params.get(key)` instead of `params[key]`?

PR #2218, opened 2020-09-14
fix: displaying warning if Condition and DeploymentPreferences are us…

c2tarun:
…ing on SAM Function resource.

*Issue #, if available:*

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/warnings/sam_cli_warning.py
+++ b/samcli/lib/warnings/sam_cli_warning.py
@@ -29,7 +29,10 @@ def _get_deployment_preferences_status(function):
 
 class TemplateWarningsChecker:
     def __init__(self):
-        self.all_warnings = {CodeDeployWarning.__name__: CodeDeployWarning()}
+        self.all_warnings = {
+            CodeDeployWarning.__name__: CodeDeployWarning(),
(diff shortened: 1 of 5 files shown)


sriram-mv (approved):

samcli/lib/warnings/sam_cli_warning.py
```diff
@@ -83,3 +86,36 @@ def check(self, template_dict):
...
+    def check(self, template_dict):
+        """
+        Checking if template dictionary have Function with Condition and DeploymentPreferences which
+        will trigger this warning.
+        """
+        functions = [
+            resource
+            for (_, resource) in template_dict.get("Resources", {}).items()
+            if resource.get("Type", "") == "AWS::Serverless::Function"
+        ]
```
mndeveci, line 109: I think this should work, but we already have function provider which lists all the Functions from the template, should we use that? https://github.com/aws/aws-sam-cli/blob/39add9acd0c7532129a6543134919eb6ae9df4f3/samcli/lib/providers/sam_function_provider.py#L15

Issue #1424, opened 2019-09-22
Provide linux installation without homebrew


SkeLLLa:
Homebrew is not common tool for linux systems. And it (homebrew) not only consume space and interfere with other tools, but it also works rather bad. Also SAM cli is only one tool that requires homebrew to be installed. 

There are a lot of better tools that provide linux system installation. Consider using them instead of uncommon homebrew and will make this tool linux-friendly.

TiS:
For starters, I suggest snapcraft (highly), or flatpak (somehow reluctantly), if you want to have one install for multiple environments.

SkeLLLa:
@TiS I don't think that this actually need functionality that snapcraft provides. I think it would just some installation script like nvm or even homebrew does `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash`.

rapofran:
please this

rapofran:
@SkeLLLa did you find some workaround?

TiS:
I suggested these tools, because they bundle dependencies. This is good for tools like that, if we want them to be independent from OS details and packages.

Coming from Ubuntu, Snapcraft is the most natural choice for me, but I heard many people chooses flatpak as better alternative (I don't know why, though, as my personal flatpak experience is quite bad). 

Using just installation script can be tricky, as one has to somehow take into account all of package version differences between systems. Also, installation script makes things like uninstall / upgrade harder to do, while snapcraft / flatpak, being package managers, suport them out of the box.

rapofran:
I think the docs are bad written

running `pip install aws-sam-cli` installs everything OK

siwyd:
It is a bit peculiar to suggest Linux users to install Homebrew when a simple `pip3 install --user aws-sam-cli` works.

dwhitla:
Plenty of Mac users refuse to install Homebrew also because it interferes with their alternative distribution preference. Please get rid of Homebrew completely as a dependency for aws-sam-cli. It is completely unnecessary.

dwhitla:
apt get install

darrenweiner:
Agreed, brew is not the recommended approach for linux.  Pip makes a lot more sense, and will perform much better in a ci/cd pipeline.  Brew is incredibly slow.

SmoothGlide:
[...]

PR #2224, opened 2020-09-16
add spec file for pyinstaller and circleci config to build package

dsayling:
*Issue #, if available:* #1424 

Linuxbrew is not ideal

This uses pyinstaller to provide a reproducible package. 
With the way it's build in the current circleci config, systems with Debian GLIBC 2.19 would be able to use this pacakge sam application. (This is most modern debian distros)

There's still plenty of work to be done.

- [ ] a location to upload the resulting artifact
- [ ] building for other systems/distros

None of the aws-sam-cli itself

n/a

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,39 @@
+version: 2.1
+
+executors:
+  docker-cimg:
+    docker:
+      - image: circleci/python:3.6-jessie
+  # macos-11-4:
+  #   macos:
+  #     xcode: 11.4.0
+  # machine-16:
+  #   machine:
+  #     image: ubuntu-1604:201903-01
+
+
+jobs:
+  build-sam-package:
+    # executor: <<parameters.executor>>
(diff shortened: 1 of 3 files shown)


awsjeffg:
@dsayling thanks for taking the time to create this PR, and I apologize if issue #1424 hasn't been updated to reflect the SAM team's current plans. We are also moving forward with changes to how we distribute SAM cli in a way that is similar to your approach here, and we expect to move towards binary-style installers similar to [how AWS cli currently works](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html) in the near future. 

If the changes you have included here unblock you in the near term, feel free to create a pyinstaller-based installer. We will not be accepting this PR at this time however as we expect to move forward with our own installer improvements shortly.

Thanks again!

PR #2226, opened 2020-09-16
Fix: fixed an integration test error with pull #2187

qingchm:
*Issue #, if available:*
Issue1291

The previous integration test caused a failure in canary test

Changed the integration test to capture the stderr

Not known

No

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/s3_uploader.py
```diff
@@ -101,6 +101,7 @@ def upload(self, file_name, remote_path):
                     file_name, self.bucket_name, remote_path, additional_args, [print_progress_callback]
                 )
             else:
+                sys.stderr.flush()
```
CoshUS, line 104: Is flush() required here? Can you try running the failing test without flush()?

CoshUS (changes requested):

samcli/lib/package/s3_uploader.py
qingchm: It is not required, will remove this line

CoshUS (approved):

Issue #1236, opened 2019-06-20
`sam build` builds the same CodeUri multiple times


benkehoe:
### Describe your idea/feature/enhancement

`sam build` treats each function separately, even if they share the same CodeUri. This leads to unnecessarily longer build times.

### Proposal

People may rely on the structure of `.aws-sam`, so maybe stick with a directory per function, but if a function uses the same CodeUri as a function that we've already built in this build, just copy the build directory rather than run a new build.

TheSriram:
https://github.com/awslabs/aws-sam-cli/pull/1180 is a step in that direction.

Arekusei:
any news?

sriram-mv:
https://github.com/aws/aws-sam-cli/issues/1236 makes sure that if functions reference the same exact codeuri, they are built only once.

benkehoe:
@sriram-mv were you intending to reference a PR or comment on a different issue? This issue is #1236

sriram-mv:
Oops you are right. https://github.com/aws/aws-sam-cli/pull/2232 this is the PR I wanted to reference.

benkehoe:
Awesome! Loving that the companion to "Hello World" in the tests is "Hello Mars"

forgivegod:
looks like 1.6.0 addresses this:
https://github.com/aws/aws-sam-cli/releases/tag/v1.6.0

awood45:
Resolved in `1.6.2`

Issue #1072, opened 2019-03-21
sam build - reuse lambda deployment packages, ideally one-per-manifest


adilosa:
### Describe your idea/feature/enhancement

`sam build` appears to re-create identical Lambda deployment packages for each of my Python functions. That is, all dependencies and code files are included – even handler files for Lambdas other than the Lambda being built. 

This adds increased build time as the same deployment package is built and uploaded over and over. 

### Proposal

It would be great if `sam build` could reuse Lambda deployment packages when the environment between Lambda functions is the same. Since only one manifest is supported right now, this should result in one deployment package.

Ideally, multiple manifests could be provided so Lambdas can have different dependencies. Then `sam build` would produce one deployment package per manifest, rather than re-building an uber deployment package for every lambda. 

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)
Perhaps, to add a way to specify which manifest a Lambda function should use

adilosa:
Looks like this can be somewhat mitigated by: 
* Creating multiple source folders with different `requirements.txt` files
* Storing common dependencies in a Lambda Layer so they don't need to be built into every handler (pending #840)

That said, it still seems like one deployment package per handler function could be unnecessary – even though their size can be mitigated with Layers.

TheSriram:
Looks like you have captured mitigation stratergies which make sense.

``
it still seems like one deployment package per handler function could be unnecessary – even though their size can be mitigated with Layers.
``

Could you explain a bit more on this, this is in essence the deployment artifacts that need to be uploaded for lambda to work.

adilosa:
Sure, I'm structuring my repo like this:

```
.
+-- some_lambdas
|   +-- __init__.py   # contains 3 Lambda handler functions
|   +-- requirements.txt
+-- some_lambdas_2
|   +-- __init__.py
|   +-- first_handler.py
|   +-- second_handler.py
|   +-- requirements.txt
+-- template.yml
```

Here I have 5 Lambda handler functions ij 2 directories. In theory, this requires 2 deployment packages – one for `some_lambdas` and one for `some_lambdas_2` – with the Lambda functions having different entry points.

However, since there are 5 `AWS::Serverless::Function` resources in my SAM template, `sam build` will accordingly build 5 deployment packages. Resulting in 3 copies of the `some_lambdas` environment and 2 copies of the `some_lambdas_2` environment. 

[...]

PR #2232, opened 2020-09-19
Dedup builds

mndeveci:
*Issue #, if available:*
- https://github.com/aws/aws-sam-cli/issues/1236
- https://github.com/aws/aws-sam-cli/issues/1072

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/lib/build_module/test_app_builder.py
```diff
@@ -41,21 +48,97 @@ def test_must_iterate_on_functions_and_layers(self):
...
+        build_function_mock = Mock()
+
+        # create 3 function resources where 2 of them would have same codeuri, runtime and metadata
+        function1_1 = generate_function("function1_1")
+        function1_2 = generate_function("function1_2")
+        function2 = generate_function("function2", runtime="different_runtime")
+        resources_to_build_collector = ResourcesToBuildCollector()
+        resources_to_build_collector.add_functions([function1_1, function1_2, function2])
+
+        with osutils.mkdir_temp() as temp_base_dir:
```
CoshUS, line 104: Shouldn't osutils.mkdir_temp be patched and mocked for unit tests? This also goes for other IO calls.

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -134,3 +134,20 @@ def _verify_invoke_built_function(self, template_path, function_logical_id, over
 
         process_stdout = process_execute.stdout.decode("utf-8")
         self.assertEqual(json.loads(process_stdout), expected_result)
+
+
+class DedupBuildIntegBase(BuildIntegBase):
```
CoshUS, line 139: Correct me if I'm wrong, but this does not verify there is no duplicate builds right?

CoshUS (changes requested):

tests/integration/buildcmd/build_integ_base.py
mndeveci: This is just BaseClass which contains some shared function definitions there. Actual tests for de-dup builds are here; https://github.com/mndeveci/aws-sam-cli/blob/dedup_builds/tests/integration/buildcmd/test_build_cmd.py#L1037-L1080

CoshUS: In those two tests, they just call base _verify_build_and_invoke_functions after running build. How are we checking the intended de-dup feature in which a function should be built only once instead of multiple times?

mndeveci: You are right, this is not specifically check for building once, this verifies that building once won't mess-up with different functions. Let me grab command line output and verify there.

samcli/lib/build/build_graph.py
```diff
@@ -0,0 +1,188 @@
...
+        If graph already contains a build definition which is same as the newly passed one, then it will add
+        the function to the existing one, discarding the new one
+
+        If graph doesn't contain such unique build definition, it will be added to the current build graph
+
+        :param build_definition: build definition which is newly read from template.yaml file
+        :param function: function details for this build definition
+        """
+        if build_definition in self._build_definitions:
+            previous_build_definition = self._build_definitions[self._build_definitions.index(build_definition)]
```
wchengru, line 85: What is the benefit of using a list instead of set here? I saw equality of build_definition is defined.

samcli/lib/build/build_graph.py
```diff
@@ -0,0 +1,188 @@
...
+            LOG.debug("Same build definition found, adding function (Previous: %s, Current: %s, Function: %s)",
+                      previous_build_definition, build_definition, function)
+            previous_build_definition.functions.append(function)
+        else:
+            LOG.debug("Unique build definition found, adding as new (Build Definition: %s, Function: %s)",
+                      build_definition, function)
+            build_definition.functions.append(function)
```
wchengru, line 92: We have add_function() already I think. Here and L124.

samcli/lib/build/build_graph.py
```diff
@@ -0,0 +1,188 @@
...
+        self.codeuri = codeuri
+        self.metadata = metadata if metadata else {}
+        self.uuid = str(uuid4())
+        self.functions = []
+
+    def add_function(self, function):
+        self.functions.append(function)
+
+    def get_function_name(self):
+        return self.functions[0].name
```
wchengru, line 167: Check in case functions is empty.

tests/unit/lib/build_module/test_app_builder.py
mndeveci: Sure, I verified that reading and writing to `build.toml` file has already been tested in another file, I will mock here.

samcli/lib/build/build_graph.py
mndeveci: Good catch will update it

samcli/lib/build/build_graph.py
mndeveci: Build process is calling `clean_redundant_functions_and_update` method to clean any build definitions without functions in it, but I will add secondary check here in case something goes bad in the future.

samcli/lib/build/app_builder.py
```diff
@@ -127,6 +123,49 @@ def build(self):
 
         return result
 
+    def _get_build_graph(self):
+        """
+        Converts list of functions into a build graph, where we can iterate on each unique build and trigger build
+        :return: BuildGraph, which represents list of unique build definitions
+        """
+        build_graph = BuildGraph(self._build_dir)
+        functions = self._resources_to_build.functions
```
sriram-mv, line 132: do we account for building layers too? Should we?

samcli/commands/build/build_context.py
```diff
@@ -169,6 +169,10 @@ def resources_to_build(self):
         result.add_layers([l for l in self._layer_provider.get_all() if l.build_method is not None])
         return result
 
+    @property
+    def is_specific_resource(self):
```
sriram-mv, line 173: This should get a better name, its hard to get what it means by just taking a glance. Also is this worthy of a docstring?

sriram-mv (changes requested):
I need to do another look at this, will do in the morning.

samcli/commands/build/build_context.py
mndeveci: Sure I will add some documentation, and for naming I could change it to `is_building_specific_resource`, what do you think?

samcli/lib/build/app_builder.py
mndeveci: No we don't, and we shouldn't. Because functions may have multiple function definition in the same codeuri and runtime, because of the entry-point (you can write multiple functions and keep them in the same package, but your entry-point should be different). But for layer's, I couldn't think of any reason, why you will crate 2 different layers out of same codeuri and runtime, where you could just create one layer and use it.

samcli/commands/build/build_context.py
```diff
@@ -170,7 +170,11 @@ def resources_to_build(self):
         return result
 
     @property
-    def is_specific_resource(self):
+    def is_building_specific_resource(self):
+        """
+        Checks whether customer requested to build specific resource
```
sriram-mv, line 175: customer requested to build a specific resource alone in isolation, by specifying function_identifier to the build command. 

Something like that makes it more clear.

sriram-mv (approved):
I'm approving, 

I have a comment on naming and docstring, which should be relatively small to address.

wchengru (approved):

CoshUS (approved):

Issue #1866, opened 2020-03-13
sam deploy --guided should pick up params set in samconfig.toml


insane-dreamer:
When a samconfig.toml file exists there may be additional parameters to be passed to sam deploy, notably passwords, that are not stored by default when doing --guided and which could be added manually to samconfig.toml file but that's not ideal from a security point of view. 

Granted, those pwd params could be passed with `--parameter-overrides` but that requires 1) a long string (easy to have a typo), 2) you have to lookup all the param names that have pwds 3) all your passwords are in your cli history. 

It would be helpful if running sam deploy --guided it picked up the settings from samconfig.toml and showed them, overriding the template defaults, when going through the params. This would not only be useful in the case of pwds but also other params which may have been forgotten or which one wants to do a bit differently than in samconfig.toml--i.e., they're mostly the same but just one is different. 

Perhaps this could be an option, such as `sam deploy --guided --defaults samconfig.toml` which would take the defauts from samconfig.toml. 

(--guided does presently give a message that it loads samconfig.toml, but when guiding through the params it shows the original template defaults, not those in samconfig.toml).

PR #2234, opened 2020-09-21
fix: use parameter overrides from configuration file or command line

wchengru:
*Issue #, if available:*
#1866 

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/deploy/guided_context.py
```diff
@@ -203,3 +212,17 @@ def run(self):
                 confirm_changeset=self.confirm_changeset,
                 capabilities=self._capabilities,
             )
+
+    def _get_parameter_default_value(self, parameter_key, parameter_properties, parameter_override_from_cmdline):
```
sriram-mv, line 216: we should just call this value, since we are fetching from config file if present, not just default values.

sriram-mv (changes requested):

samcli/commands/deploy/guided_context.py
wchengru: Changed the function name here.

sriram-mv (approved):

Issue #2230, opened 2020-09-18
SAM guided deploy does not respect AWS Region set in config file


rts-rob:
### Description

Briefly describe the bug you are facing.

When using guided deploy with `sam deploy --guided`, AWS SAM CLI does not respect the AWS Region set in the config file. This requires the user to manually enter the Region on every first deploy, placing an unnecessary mental burden on the developer. This can result in the unintended deployment of applications to _us-east-1_. This creates confusion for developers and a non-intuitive developer experience.

### Steps to reproduce

1. Create and activate an AWS profile with a default Region **other than** _us-east-1_ by running `aws configure` or manually editing _~/.aws/config_
1. Create an AWS SAM app with `sam init`
1. Change into the new app directory
1. [Optional depending on runtime and template] Run `sam build`
1. Run `sam deploy --guided`

### Observed result

SAM CLI presents: `AWS Region [us-east-1]: `

### Expected result

SAM CLI presents: `AWS Region [<profile Region>]: `

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: all
2. `sam --version`: SAM CLI, version 1.2.0

singledigit:
+1

awsjeffg:
The [docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-config.html) don't explicitly cover what should happen here, what they do say instead is:

> region
>
 >   Set the region you want to deploy to. If you omit region, then the AWS SAM CLI will try to resolve the region through AWS profiles and then environment variables.

I think I agree, it seems reasonable that if a customer has set a default region via aws cli it would follow that SAM would try to discover and honor that intent. I'll dig a bit further and see what the original design intent was.

mgrandis:
Fixed in 1.16

DanCookeWillow:
What is the workaround for this? 

I cannot imagine a CF template that doesn't use at least AWS::Region, are we just supposed to hardcode these values if we want to use `sam local` ?

Would be amazing to get some kind of support

Surely its as simple as just reading AWS_REGION or AWS_DEFAULT_REGION when running locally?

PR #2240, opened 2020-09-22
Set default region in sam deploy --guided

simoneBrancati:
*Issue #, if available:*
#2230 

This change sets the default region in `sam deploy --guided` as the region in the aws config file (found in `~/.aws/config`) or in the environment variables (as `AWS_DEFAULT_REGION`).

Currently `sam deploy --guided` tries to find the region in the `samconfig.toml`, otherwise falls back to an hardcoded value of `us-east-1`.
This change gets the current session, using botocore, and gets the 'region' variable for that session.
The session region is set by botocore from the `~/aws/config` configuration file or can be overridden by the `AWS_DEFAULT_REGION` environment variable.

None found.

*Checklist:*

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


sriram-mv:
@simoneBrancati : this looks great! could you add a unit test too for this, and we can take this in :)

CoshUS (approved):

simoneBrancati:
@sriram-mv Hi, i think this is ready for review

samcli/commands/deploy/guided_context.py
```diff
@@ -88,7 +89,7 @@ def guided_parameter_overrides(self):
     # pylint: disable=too-many-statements
     def guided_prompts(self, parameter_override_keys):
         default_stack_name = self.stack_name or "sam-app"
-        default_region = self.region or "us-east-1"
+        default_region = self.region or get_session().get_config_variable("region") or "us-east-1"
```
prenx4x, line 92: this seems redundant. we are already defaulting region to us-east-1. what is the benefit of getting region from boto session?

simoneBrancati: That's literally the only change this pull request makes. It looks for the user's default region in the aws config file (found in ~/.aws/config) or in the environment variables (as AWS_DEFAULT_REGION). If neither is found It defaults to us-east-1.

prenx4x: I understand what it does. Just want to understand why we need this? Are you trying to prevent some corner case failure here? or is it to just improve general UX?

simoneBrancati: Improve user experience

prenx4x (approved):

Issue #1773, opened 2020-02-05
No timestamp in --debug


literakl:
### Describe your idea/feature/enhancement

I need to find out why it takes 4-6 seconds to process my lambda, which takes 400 ms to initialize and 200 ms to run. I found --debug option. It really prints more information but without a time indication.  So I cannot compare where `sam `spent the time.

2020-02-05 18:36:22 127.0.0.1 - - [05/Feb/2020 18:36:22] "GET /v1/verify/Yk6QwCF1 HTTP/1.1" 403 -
Constructed String representation of Event to invoke Lambda. Event: {"httpMethod": "GET", "body": null, "resource": "/v1/verify/{token}", "requestContext": {"resourceId": "123456", "apiId": "1234567890", "resourcePath": "/v1/verify/{token}", "httpMethod": "GET", "requ
estId": "c6af9ac6-7b61-11e6-9a41-93e8deadbeef", "accountId": "123456789012", "stage": "production", "identity": {"apiKey": null, "userArn": null, "cognitoAuthenticationType": null, "caller": null, "userAgent": "Custom User Agent String", "user": null, "cognitoIdentity
PoolId": null, "cognitoAuthenticationProvider": null, "sourceIp": "127.0.0.1", "accountId": null}, "extendedRequestId": null, "path": "/v1/verify/{token}"}, "queryStringParameters": null, "multiValueQueryStringParameters": null, "headers": {"Host": "127.0.0.1:3000", "
Connection": "keep-alive", "Accept": "application/json, text/plain, */*", "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.130 Safari/537.36", "Origin": "http://localhost:8080", "Sec-Fetch-Site": "cross-s
ite", "Sec-Fetch-Mode": "cors", "Referer": "http://localhost:8080/verify/Yk6QwCF1", "Accept-Encoding": "gzip, deflate, br", "Accept-Language": "cs", "X-Forwarded-Proto": "http", "X-Forwarded-Port": "3000"}, "multiValueHeaders": {"Host": ["127.0.0.1:3000"], "Connection
[...]

PR #2266, opened 2020-10-01
log timestamp when --debug is presented (#1773)

aahung:
log entry will look like '2020-09-30 20:51:12,353 | ...'

*Issue #1773, if available:*

- Allow users to have better visibility on how long each process takes

- Timestamp is printed as well

- No side effect I am aware of

*Checklist:*

- [x] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


aahung:
A sample log:

```
(samcli37) ~/P/hello-world-nodejs ❯❯❯ samdev local invoke "HelloWorldFunction" -e events/event.json --debug --timestamp-output
2020-09-30 20:51:12,353 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2020-09-30 20:51:12,509 | local invoke command is called
2020-09-30 20:51:12,515 | No Parameters detected in the template
2020-09-30 20:51:12,539 | 2 resources found in the template
2020-09-30 20:51:12,539 | Found Serverless function with name='HelloWorldFunction' and CodeUri='HelloWorldFunction'
2020-09-30 20:51:12,548 | Found one Lambda function with name 'HelloWorldFunction'
2020-09-30 20:51:12,549 | Invoking app.lambdaHandler (nodejs12.x)
2020-09-30 20:51:12,549 | No environment variables found for function 'HelloWorldFunction'
2020-09-30 20:51:12,549 | Environment variables overrides data is standard format
2020-09-30 20:51:12,549 | Loading AWS credentials from session with profile 'None'
2020-09-30 20:51:12,574 | Resolving code path. Cwd=/Users/XXXXX/Projects/hello-world-nodejs/.aws-sam/build, CodeUri=HelloWorldFunction
2020-09-30 20:51:12,574 | Resolved absolute path to code is /Users/XXXXX/Projects/hello-world-nodejs/.aws-sam/build/HelloWorldFunction
2020-09-30 20:51:12,574 | Code /Users/XXXXX/Projects/hello-world-nodejs/.aws-sam/build/HelloWorldFunction is not a zip/jar file
2020-09-30 20:51:12,586 | Skip pulling image and use local one: amazon/aws-sam-cli-emulation-image-nodejs12.x:rapid-1.3.2.

2020-09-30 20:51:12,586 | Mounting /Users/XXXXX/Projects/hello-world-nodejs/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
2020-09-30 20:51:12,884 | Starting a timer for 3 seconds for function 'HelloWorldFunction'
START RequestId: XXXXX-XXXXXXX Version: $LATEST
END RequestId: XXXXX-XXXXXXXREPORT RequestId: XXXXX-XXXXXXX	Init Duration: 101.10 ms	Duration: 2.50 ms	Billed Duration: 100 ms	Memory Size: 128 MB	Max Memory Used: 39 MB
2020-09-30 20:51:13,132 | Sending Telemetry: {XXXXXXXXXXXXXXXXXXXXXXX}
^[[A2020-09-30 20:51:13,764 | HTTPSConnectionPool(host='aws-serverless-tools-telemetry.us-west-2.amazonaws.com', port=443): Read timed out. (read timeout=0.1)

{"statusCode":200,"body":"{\"message\":\"hello world\"}"}
```

tests/unit/cli/test_context.py
```diff
@@ -31,6 +34,39 @@ def test_must_unset_get_debug_flag(self):
...
+        for handler in handlers:
+            self.assertNotRegex(handler.formatter.format(message_record), timestamp_log_regex)
+
+        ctx.output_timestamp = True
+        self.assertEqual(ctx.output_timestamp, True, "output_timestamp must be set to True")
+        # timestamp should be output
+        for handler in handlers:
+            self.assertRegexpMatches(handler.formatter.format(message_record), timestamp_log_regex)
+
+    def test_must_unset_get_output_timestamp_flag(self):
```
CoshUS, line 60: Why do we need this test case?

samcli/cli/context.py
```diff
@@ -54,6 +60,22 @@ def debug(self, value):
...
+    @output_timestamp.setter
+    def output_timestamp(self, value):
+        """
+        Turn on timestamp logging to print a timestamp at the beginning of each logg entry.
+
+        :param value: Value of output-timestamp flag
+        """
+        self._output_timestamp = value
+
+        for handler in logging.getLogger("samcli").handlers + logging.getLogger("aws_lambda_builders").handlers:
```
CoshUS, line 72: Since we are using logging.getLogger("samcli") and logging.getLogger("aws_lambda_builders") in quite a few places, taking "samcli" and "aws_lambda_builders" out as constants would be more maintainable.

CoshUS (changes requested):

samcli/cli/context.py
```diff
@@ -40,6 +42,10 @@ def __init__(self):
     def debug(self):
         return self._debug
 
+    @property
+    def output_timestamp(self):
```
jfuss, line 46: Can't we just configure the logger to have the timestamp when debugging is enabled? Do we really need a new options for this?

jfuss (changes requested):

tests/unit/cli/test_context.py
aahung: We do have one for the `--debug` flag, so this one is to keep it consistent with it
https://github.com/aws/aws-sam-cli/blob/f23baa9dc99ecb76ce12f88c14ffa48df39c983a/tests/unit/cli/test_context.py#L24-L32

samcli/cli/context.py
aahung: done! good suggestion!

samcli/cli/context.py
aahung: Related comment from @awsjeffg 

https://github.com/aws/aws-sam-cli/issues/1773#issuecomment-695076282

I think the idea is, maybe, for non-debug mode, users will want still to see the timestamp. So having a dedicate flag gives such flexibility.

CoshUS (approved):

samcli/cli/context.py
awsjeffg: @jfuss You're not wrong, I am just being careful as we've run into other problems where changing stdout behaviour causes customers problems.

jfuss: @awsjeffg Yes, we have run into problems with other tools expecting certain output but that has been against our 'info' logs not 'debug'. Additionally, we should be able to freely add data without worrying. I see a stronger case for removing  output. I am not saying we shouldn't be cautious around changing output but as a product we need to have the flexibility to update and change output, otherwise we are always stuck with whatever we ship with. I am a strong believer in relying on stdout or stderr output is an anti-pattern and should be treated as behavior not an interface. I am only aware of the AWS Toolkit depending on our output, can't we just ask them if this would break them? My understanding is it shouldn't.

Long ago, we had the default logger for all of SAM CLI's logs. This was overly noisy and we ended up removing it. I personally do not see a difference between adding `--time-stamp-output` and using `--debug`. They are ways to configure the logging and output. We should be aware of the additional overhead on our customer when adding new options for things. Why should we have an option versus an environment variable? If you want this still to be intentionally, wouldn't something like allowing `--debug --debug` to allow customers to control/add additional output (similar to `-vv` which is common for command line tools)? 

I can understand wanting to introduce this in a new way somehow (though I don't fully agree), but the main thing I want to poke at is: Why a new command line option?

aahung: Both are valid in my opinion. Another thought against **ONLY** including timestamp in DEBUG mode: including timestamp into `--debug` which is used for configuring logging level is implicit in my opinion. 

Ideally I would prefer to always include timestamps regardless what logging level users chooses to configure. But as @awsjeffg pointed out, it will change stdout behaviour. I understand changing stdout is not something we like to do very often, but `--output-timetamp` does feel like a temporary solution. 

Considering users only use `--debug` mode for investigating (are they?), I think it is fairly safe to change stdout behavior there assuming debug stdout won't be used for postprcessing. What do you think?

aahung: As discussed offline with @awsjeffg, we are going with including timestamp when `--debug` is presented

samcli/cli/context.py
```diff
@@ -50,9 +55,12 @@ def debug(self, value):
...
 
         if self._debug:
-            # Turn on debug logging
-            logging.getLogger("samcli").setLevel(logging.DEBUG)
-            logging.getLogger("aws_lambda_builders").setLevel(logging.DEBUG)
+            # Turn on debug logging and display timestamps
+            for logger_name in [SAM_CLI_LOGGER_NAME, LAMBDA_BULDERS_LOGGER_NAME]:
+                logger = logging.getLogger(logger_name)
+                logger.setLevel(logging.DEBUG)
+                for handler in logger.handlers:
```
jfuss, line 62: Why are we setting on all handlers?

samcli/cli/context.py
```diff
@@ -50,9 +55,12 @@ def debug(self, value):
         self._debug = value
 
         if self._debug:
-            # Turn on debug logging
-            logging.getLogger("samcli").setLevel(logging.DEBUG)
-            logging.getLogger("aws_lambda_builders").setLevel(logging.DEBUG)
+            # Turn on debug logging and display timestamps
+            for logger_name in [SAM_CLI_LOGGER_NAME, LAMBDA_BULDERS_LOGGER_NAME]:
```
jfuss, line 59: Just noticing this. Why are we setting the logger directly instead of through `SamCliLogger.configure_logger` like we do in other places?

jfuss (changes requested):
Just a couple questions but overall this looks good.

samcli/cli/context.py
aahung: There is only one handlers I think. In order to get handler object from `logger`, only `logger.handlers` is available.

samcli/cli/context.py
aahung: `SamCliLogger.configure_logger` will add a new handler. We should change the existing handlers here.

aahung: But I do think we should pack this into `SamCliLogger` though!

aahung: (done)

samcli/cli/context.py
aahung: (no longer applicable in the latest iteration)

samcli/lib/utils/sam_logging.py
```diff
@@ -21,13 +28,18 @@ def configure_logger(logger, formatter, level):
         -------
         None
         """
-        log_stream_handler = logging.StreamHandler()
+        handlers = logger.handlers
+        if handlers:
+            log_stream_handler = handlers[0]
+        else:
+            log_stream_handler = logging.StreamHandler()
+            logger.addHandler(log_stream_handler)
```
jfuss, line 36: Do we not need to `addHandler()` in the if part?

aahung: If there is already a handler, it indicates this method is called before. We only need to change the existing handler instead of creating a new one.

jfuss: So then we don't need line 32 and 33 above? Can we update this to only do the StreamHanlder if handler is None?

jfuss (approved):

Issue #2252, opened 2020-09-24
SAM build should validate that layers exist


awsjeffg:
### Describe your idea/feature/enhancement

If I specify a layer in my template pointing to a local zip file and that file does not exist, I should get the warning from sam build instead of running into it later in sam local invoke or sam deploy.

aahung:
For invalid ARN, an `InvalidLayerVersionArn` is thrown. To make it consistent with that, I think it should be an exception instead of a warning.

PR #2275, opened 2020-10-06
[WIP] Verify local layer file exists whenever LayerVersion inits

aahung:
*Issue #, if available:*

#2252 

Quote original issue:

> If I specify a layer in my template pointing to a local zip file and that file does not exist, I should get the warning from sam build instead of running into it later in sam local invoke or sam deploy.

For invalid ARN, an `InvalidLayerVersionArn` is thrown when creating a `LayerVersion` with an invalid ARN. To make it consistent with that, I think it should also raise an exception when the `ContentUri` does not exist.

By checking the content uri at the LayerVersion creation time, we can ensure the verification is at at its earliest time.

I am not sure about that. Can someone evaluate it?

*Checklist:*

- [x] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -4,9 +4,14 @@ source
 """
 import hashlib
 import logging
+import os
 from collections import namedtuple
 
-from samcli.commands.local.cli_common.user_exceptions import InvalidLayerVersionArn, UnsupportedIntrinsic
+from samcli.commands.local.cli_common.user_exceptions import (
+    InvalidLayerVersionArn,
+    UnsupportedIntrinsic,
+    InvalidLayerVersionContentUri,
+)
 
 LOG = logging.getLogger(__name__)
 
@@ -105,6 +110,8 @@ class LayerVersion:
         self._version = LayerVersion._compute_layer_version(self.is_defined_within_template, arn)
         self._build_method = metadata.get("BuildMethod", None)
         self._compatible_runtimes = compatible_runtimes
+        if self.is_defined_within_template:
+            LayerVersion._verify_code_uri_existence(codeuri)
 
     @staticmethod
     def _compute_layer_version(is_defined_within_template, arn):
@@ -170,6 +177,11 @@ class LayerVersion:
         return LayerVersion.LAYER_NAME_DELIMETER.join(
             [layer_name, layer_version, hashlib.sha256(arn.encode("utf-8")).hexdigest()[0:10]]
         )
+
+    @staticmethod
+    def _verify_code_uri_existence(code_uri):
+        if not os.path.exists(code_uri):
+            raise InvalidLayerVersionContentUri(code_uri + " is an invalid ContentUri, file/directory does not exist.")
 
     @property
     def arn(self):
--- a/tests/unit/commands/local/lib/test_sam_function_provider.py
+++ b/tests/unit/commands/local/lib/test_sam_function_provider.py
@@ -1,8 +1,8 @@
 from unittest import TestCase
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 from parameterized import parameterized
 
-from samcli.commands.local.cli_common.user_exceptions import InvalidLayerVersionArn
+from samcli.commands.local.cli_common.user_exceptions import InvalidLayerVersionArn, InvalidLayerVersionContentUri
 from samcli.lib.providers.provider import Function, LayerVersion
 from samcli.lib.providers.sam_function_provider import SamFunctionProvider
 from samcli.lib.providers.exceptions import InvalidLayerReference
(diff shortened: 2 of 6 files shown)


tests/testing_utils.py
```diff
@@ -95,3 +96,18 @@ def full_path(self, filename):
         f.full_path('foo/bar.txt') -> /tmp/asdfasd/foo/bar.txt
         """
         return os.path.join(self.rootdir, filename)
+
+
+class TempFile(object):
```
jfuss, line 101: You can use `with tempfile.NamedTemporaryFile(delete=False)` instead. This is how all of our other tests are setup, so let's keep everything consistent.

samcli/lib/providers/provider.py
```diff
@@ -171,6 +178,11 @@ def _compute_layer_name(is_defined_within_template, arn):
             [layer_name, layer_version, hashlib.sha256(arn.encode("utf-8")).hexdigest()[0:10]]
         )
 
+    @staticmethod
+    def _verify_code_uri_existence(code_uri):
+        if not os.path.exists(code_uri):
+            raise InvalidLayerVersionContentUri(code_uri + " is an invalid ContentUri, file/directory does not exist.")
```
jfuss, line 184: What if the Layer is defined to be an S3 URI already?

jfuss (changes requested):
This generally looks good. I have a couple comments.

We should also be adding integration tests for these case, to ensure we have proper coverage.

tests/testing_utils.py
aahung: thanks for the hint. Didn't know there is this method. 

Why we are going with `delete=False`?

samcli/lib/providers/provider.py
aahung: good catch!

aahung: It does not seem to affect it. I tested on this template snippet:

```yarml
Resources:
  TestLambdaLayer:
    Properties:
      LayerName: test-layer
      Description: Lambda layer with all node module dependency and shared code between
        lambda
      ContentUri: s3://dummy-bucket/my-layer.zip
      CompatibleRuntimes:
      - nodejs12.x
      RetentionPolicy: Retain
```

The `code_url`'s value is `.`, so it will pass the `_verify_code_uri_existence`

tests/unit/commands/local/lib/test_sam_layer_provider.py
```diff
@@ -72,34 +74,29 @@ class TestSamLayerProvider(TestCase):
     }
 
     def setUp(self):
+        Path("PyLayer").mkdir(exist_ok=True)
```
jfuss, line 77: Unit tests should not be creating resources on the system. We should mock out the call instead of this.

As an additional note: When we do create things on the system, they should always be done in in a tmp directory (through tempfile)

tests/testing_utils.py
```diff
@@ -4,6 +4,7 @@
 import tempfile
 import shutil
 from collections import namedtuple
+from pathlib import Path
```
jfuss, line 7: remove unused imports.

tests/unit/commands/local/lib/test_sam_function_provider.py
```diff
@@ -516,28 +517,30 @@ def test_must_ignore_opt_in_AmazonLinux1803_layer(self):
             self.assertEqual(actual_layer, expected_layer)
 
     def test_layers_created_from_template_resources(self):
-        resources = {
-            "Layer": {"Type": "AWS::Lambda::LayerVersion", "Properties": {"Content": {"Bucket": "bucket"}}},
-            "ServerlessLayer": {"Type": "AWS::Serverless::LayerVersion", "Properties": {"ContentUri": "/somepath"}},
-        }
+        with tempfile.NamedTemporaryFile() as handle:
```
jfuss, line 520: Any reason we are not mocking out the call to check if the file exists instead of creating one? Typically unit tests do not create resources on the system.

tests/unit/commands/local/lib/test_sam_layer_provider.py
```diff
@@ -1,3 +1,5 @@
+import os
```
jfuss, line 1: Remove unused imports.

Not sure if black does this for us (pretty sure it does). If not, there should be a configuration in your IDE for this.

jfuss (changes requested):

tests/testing_utils.py
aahung: removed

tests/unit/commands/local/lib/test_sam_function_provider.py
aahung: thanks, changed to use `@patch`

aahung:
Checking in, this is still something I want to finish. To be rebased.

PR #2276, opened 2020-10-06
Update bug_report.md for Slack unfurl

bmicklea:
Trying to tweak template so unfurl in Slack is more readable.

*Issue #, if available:*

Should make reading the unfurled snippet in Slack easier which, in turn, makes it easier to reference the correct issue in conversation.

- moves help text to comments
- adds colons after headings so they're more readable without carriage returns in unfurl

Makes the issue template longer and maybe harder to use if people ignore comments (that's a bad idea though).

No

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss (approved):
Thanks for the update to our Issues templates.

sriram-mv (approved):

PR #2304, opened 2020-10-15
Upgrade dependency (pylint)

aahung:
*Issue #, if available:*

*Why is this change necessary?*

Maintain our dependencies up-to-date.

*How does it address the issue?*

Upgrade the dependencies and adjust our code accordingly.

For pylint, there are new rules such as:
```
samcli/local/lambda_service/local_lambda_invoke_service.py:33:8: R1725: Consider using Python 3 style super() without arguments (super-with-arguments)
samcli/commands/local/start_api/cli.py:117:4: C0415: Import outside toplevel (samcli.commands.exceptions.UserException) (import-outside-toplevel)
samcli/commands/publish/command.py:84:8: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
```

The rule "import-outside-toplevel" has been disabled since our "violation" looks reasonable to me.

*What side effects does this change have?*

*Did you change a dependency in `requirements/base.txt`?*
    *If so, did you run `make update-reproducible-reqs`* yes

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2305, opened 2020-10-15
Upgrade base & dev Dependences except pylint

aahung:
*Issue #, if available:*

Maintain our dependencies up-to-date.

Update all dependencies versions in requirement txts to match their latest available versions.
Except for pylint, which has its own PR #2304

    *If so, did you run `make update-reproducible-reqs`* yes

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


pytest.ini
```diff
@@ -1,7 +1,5 @@
 [pytest]
 log_cli = 1
 log_cli_level = INFO
-env =
```
jfuss, line 4: Why was this removed?

requirements/base.txt
```diff
@@ -1,17 +1,17 @@
 chevron~=0.12
-click~=7.0
-Flask~=1.0.2
+click~=7.1
```
jfuss, line 2: Python will install 7.1 automatically. Do we need a bump here?

requirements/base.txt
```diff
@@ -1,17 +1,17 @@
...
+PyYAML~=5.3
+cookiecutter~=1.7.2
 aws-sam-translator==1.27.0
 #docker minor version updates can include breaking changes. Auto update micro version only.
 docker~=4.2.0
 dateparser~=0.7
 python-dateutil~=2.6, <2.8.1
 requests==2.23.0
-serverlessrepo==0.1.9
+serverlessrepo==0.1.10
```
jfuss, line 15: What are the changes here? Do we need updated integ tests?

jfuss (changes requested):

requirements/base.txt
aahung: We don't need to, but keeping our requirement up to date looks better.
I put all these no-effect upgrade under this commit:
https://github.com/aws/aws-sam-cli/pull/2305/commits/db9f0221e3ff7db0fbe7e9de9d6504299dfe4212
chore: Update versions in in base.txt (no effective difference)

pytest.ini
aahung: pytest 6.1.1 does not support `env` (and report error), and I don't find the usage of this `env` either.

requirements/base.txt
aahung: See the commit message for detail: https://github.com/aws/aws-sam-cli/pull/2305/commits/f250d39e4bc3e678260b2397e0e69edb205a13c2

>  chore: Bump serverlessrepo to 0.1.10
only one change since 0.1.9:
Support LicenseBody and ReadmeBody application metadata properties

requirements/base.txt
jfuss: The point in defining it this way to so that when pip installs happen, patch versions get installed.

pytest.ini
jfuss: This is to set the env var on the system to make boto3 happy and not fail. What is the new way to do this then?

requirements/base.txt
aahung: `~=7.0` means the `7.x.y` can be installed, what the difference when I changed it to `~=7.1`?

requirements/base.txt
aahung: Oh I see what you mean. Do you know where an existing test example located? thanks!

pytest.ini
aahung: I will double check the new way. Do you know how to produce boto3 failure?

aahung: I don't think the existing env variable `AWS_DEFAULT_REGION` is being set.
I am at develop branch cd4c061de406626480625b9a8da97ee6194914e1 and try to run a test after `make init`:

image

jfuss: Interesting! Thanks for verifying this. Seems reasonable to remove then.

aahung: https://github.com/MobileDynasty/pytest-env

I searched online and it seems never an option in pytest.ini. It requires a pytest plugin `pytest-env` which we don't have 😅

requirements/base.txt
c2tarun: At some point I added integ tests here for this task :( https://github.com/c2tarun/aws-sam-cli/commit/792695851cbf988fede9dd399747f0d00757fcff

It fell off my radar due to some other issues. @aahung please feel free to grab tests from there. But since it was long time ago, please ensure that tests are still valid and correct.

aahung: Thanks @c2tarun!
Now integration tests are here now.

Note: you might find licensebody is only tested for application creation and readmebody is only tested for application update. It took me a while to find that out. I think this is intended behavior and respect the tests in serverlessrepo:

https://github.com/awslabs/aws-serverlessrepo-python/blob/b08cb944d93519e05087fa51993a778623b69ec0/tests/unit/test_publish.py#L314-L355

tests/integration/publish/test_command_integ.py
```diff
@@ -34,65 +37,39 @@ def tearDown(self):
...
+        self.assertIn(expected_msg, result_msg)
 
-        app_metadata_text = self.temp_dir.joinpath("metadata_create_app_version.json").read_text()
+        app_metadata_text = self.temp_dir.joinpath(expected_template_filename).read_text()
         app_metadata = json.loads(app_metadata_text)
-        self.assert_metadata_details(app_metadata, process_stdout.decode("utf-8"))
+        self.assert_metadata_details(app_metadata, result_msg)
 
-    def test_create_application_version_with_semantic_version_option(self):
+    def test_update_application_version_with_semantic_version_option(self):
```
jfuss, line 60: This tests creates not updates an application right?

jfuss (changes requested):

tests/integration/publish/test_command_integ.py
aahung: I don't think so. It is under the class `TestPublishExistingApp` and the expected message is "The following metadata of application "{}" has been updated." So I corrected the test name to "update" from "create"

requirements/base.txt
```diff
@@ -1,19 +1,19 @@
...
 aws_lambda_builders==1.1.0
-tomlkit==0.5.8
+tomlkit==0.7.0
```
sriram-mv, line 17: can we verify that that this does not break integ tests, we have to be careful with any library we upgrade, esp one which deals with configuration files.

sriram-mv (changes requested):

requirements/base.txt
aahung: I keep getting this timeout every time I ran integration test locally:

```
tests/integration/local/invoke/test_integrations_cli.py::TestSamPython36HelloWorldIntegration::test_skip_pull_image_in_env_var
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~~~~ Stack of MainThread (4356918720) ~~~~~~~~~~~~~~~~~~~~~~~
  File "/Users/xinhol/.pyenv/versions/samcli377/bin/pytest", line 10, in <module>
    sys.exit(console_main())
  File "/Users/xinhol/.pyenv/versions/3.7.7/envs/samcli377/lib/python3.7/site-packages/_pytest/config/__init__.py", line 187, in console_main
    code = main()
  File "/Users/xinhol/.pyenv/versions/3.7.7/envs/samcli377/lib/python3.7/site-packages/_pytest/config/__init__.py", line 165, in main
    config=config
  File "/Users/xinhol/.pyenv/versions/3.7.7/envs/samcli377/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/Users/xinhol/.pyenv/versions/3.7.7/envs/samcli377/lib/python3.7/site-packages/pluggy/manager.py", line 93, in _hookexec
```

So I decided to explore ways to run integration test on AppVeyor, and I notice `` is already checked in "continuous-integration/appveyor/pr/make-pr" why we still need to verify integration test again when CI already run the integration tests?

https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/35969063/job/9apg0g6e2t0oacbc?fullLog=true#L4975

jfuss (approved):

sriram-mv (approved):

requirements/base.txt
aahung: Tested against the production sam-cli with:
- production generate a config that matches this PR's generated one
- this PR can load production generated config without correct stdout

PR #2306, opened 2020-10-15
pyInstall configuration, and install script

wchengru:
*Issue #, if available:*

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


requirements/base.txt
```diff
@@ -17,3 +17,5 @@ aws_lambda_builders==1.1.0
 tomlkit==0.5.8
 # TODO: remove six and its usage since we don't support Python 2
 six==1.15.0
+# pyinstaller should not update to 4.0 until https://github.com/pyinstaller/pyinstaller/issues/4674 is resolved
+pyinstaller==3.5
```
jfuss, line 21: `pyinstaller` is not a runtime dependency of SAM CLI. We should put this into dev.txt or a new file specifically for building the binary. If we were to leave it here, pyinstaller (and its depedencies) would be installed if someone still `pip install`'s, which is not what we want.

installer/assets/install
```diff
@@ -0,0 +1,144 @@
+#!/bin/sh
+# Copyright 2012-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jfuss, line 2: This should be `Copyright Amazon.com Inc. or its affiliates.`

jfuss (changes requested):
Did we test that this generates the correct binary? Can we verify the 'installer' directory is not being packaged in our .whl?

installer/assets/install
```diff
@@ -0,0 +1,144 @@
+#!/bin/sh
```
hoffa, line 1: Can we use `set -eu` so that errors and unset variables abort the script? Should simplify error handling too.

wchengru: That's a good point. Thanks, Chris!

installer/assets/install
```diff
@@ -0,0 +1,145 @@
...
+  INSTALL_SAM_EXE="$INSTALL_BIN_DIR/$EXE_NAME"
+
+  CURRENT_INSTALL_DIR="$ROOT_INSTALL_DIR/current"
+  CURRENT_SAM_EXE="$CURRENT_INSTALL_DIR/bin/$EXE_NAME"
+
+  BIN_SAM_EXE="$BIN_DIR/$EXE_NAME"
+}
+
+create_install_dir() {
+  mkdir -p "$INSTALL_DIR" || exit 1
```
hoffa, line 88: When is `mkdir -p` expected to fail? Even if it does `set -e` will automatically abort the script.

installer/assets/install
```diff
@@ -0,0 +1,145 @@
...
+  parse_commandline "$@"
+  set_global_vars
+  check_preexisting_install
+  create_install_dir
+  create_bin_symlinks
+  echo "You can now run: $BIN_SAM_EXE --version"
+  exit 0
+}
+
+main "$@" || exit 1
```
hoffa, line 145: As above, `exit 1` is unnecessary as `set -e` will already abort if an error occurs.

Also unrelated nit: as warned by GitHub, POSIX lines must [end with a newline](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) character. This can cause issues in some scenarios with the last line being ignored (can't remember what from the top of my head, but it has bit me before). I suggest [`shfmt`](https://github.com/mvdan/sh) to never have to worry about formatting: `brew install shfmt && shfmt -w install`.

Issue #2292, opened 2020-10-13
sam build stop working on sam-cli 1.6.2 building ruby 2.5 project on CircleCI


wildodecisiv:
### Description

Sam build started throwing an exception when trying to `bundle install called by sam build` the gems of my ruby lambda project.
I have checked my historical builds on CircleCI and I found that `sam build` worked perfectly with `sam-cli 1.4.0` and it started failing with the new `sam-client 1.6.2.`

### Steps to reproduce
1. On CircleCI set docker image to: lambci/lambda:build-ruby2.5
**Note:** if you set to `lambci/lambda:20201004-build-ruby2.5` it works as expected because it uses `sam-cli 1.4.0`, but the last docker image is using  `sam 1.6.2`
2. Run "sam build -t template.yaml --debug" of a ruby with lambda and layers project

### Observed result

```
.....
.....
Found workflow 'RubyBundlerBuilder' to support capabilities 'Capability(language='ruby', dependency_manager='bundler', application_framework=None)'
Running workflow 'RubyBundlerBuilder'
Running RubyBundlerBuilder:CopySource
RubyBundlerBuilder:CopySource succeeded
Running RubyBundlerBuilder:RubyBundle
Running bundle install in /tmp/tmp6kibmw5f
executing Bundler: ['bundle', 'install', '--without', 'development', 'test']
RubyBundlerBuilder:RubyBundle failed
Traceback (most recent call last):
  File "/usr/local/pipx/venvs/aws-sam-cli/lib64/python3.6/site-packages/aws_lambda_builders/workflows/ruby_bundler/actions.py", line 31, in execute
    self.subprocess_bundler.run(["install", "--without", "development", "test"], cwd=self.source_dir)
  File "/usr/local/pipx/venvs/aws-sam-cli/lib64/python3.6/site-packages/aws_lambda_builders/workflows/ruby_bundler/bundler.py", line 55, in run
    raise BundlerExecutionError(message=out.decode("utf8").strip())
aws_lambda_builders.workflows.ruby_bundler.bundler.BundlerExecutionError: Bundler Failed: 

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/pipx/venvs/aws-sam-cli/lib64/python3.6/site-packages/aws_lambda_builders/workflow.py", line 269, in run
    action.execute()
[...]

PR #2309, opened 2020-10-17
fix: for de-dup builds, build into one of the function folder instead of temporary folder

mndeveci:
*Issue #, if available:*
#2292

Building some of the ruby projects into temp folder is not working because of shared Gemfile in the root folder

Instead of building into temp folder, it is building into one of the functions for same build definition

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


awood45 (changes requested):
For languages like "Go", I'm not certain that this fully fixes the issue - can we add regression tests around this?

It's possible we may want to have a way to disable this feature.

awood45 (approved):

PR #2314, opened 2020-10-20
chore: Fix inconsistent black version in precommit

aahung:
*Issue #, if available:*

to make black version consistent with appveyor and dev guide. Even better, by only specifying black version in one place , we can reduce our maintenance cost. 

- make black installation managed
- only specify black version in one place
- avoid installing black in pre-commit (reuse existing black installed by `make init`)
- simplify dev guide

*Checklist:*

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/python/black
-    rev: 19.3b0
+    rev: 19.10b0
     hooks:
       - id: black
         language_version: python3.7
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/python/black
-    rev: 19.10b0
+    rev: 20.8b1
     hooks:
       - id: black
         language_version: python3.7
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,9 @@
 repos:
-  - repo: https://github.com/python/black
-    rev: 20.8b1
-    hooks:
-      - id: black
-        language_version: python3.7
-        exclude_types: ['markdown', 'ini', 'toml']
+- repo: local
+  hooks:
+    - id: black
+      name: black
+      entry: black
+      language: system
+      require_serial: true
+      types: [python]
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -52,6 +52,28 @@ in Appveyor during PRs. Black will be installed automatically with `make init`.
 
 After installing, you can run our formatting through our Makefile by `make black` or integrating Black directly in your favorite IDE (instructions
 can be found [here](https://black.readthedocs.io/en/stable/editor_integration.html))
+ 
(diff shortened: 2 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/python/black
-    rev: 19.3b0
+    rev: 19.10b0
```
jfuss, line 3: Latest is 20.8b1 of black. Should we upgrade to that?

aahung: Good point. In addition, I made a few changes to how we install black and sync appveyor as well. Waiting for AppVeyor's check.

aahung:
The black's pre-commit-config.yaml is taken from https://github.com/psf/black/blob/master/.pre-commit-config.yaml

aahung:
This PR proposes a change on how we install black, I would like to get more feedbacks, thanks!

DEVELOPMENT_GUIDE.md
```diff
@@ -48,10 +48,9 @@ easily setup multiple Python versions.
 ### 2. Install Additional Tooling
 #### Black
 We format our code using [Black](https://github.com/python/black) and verify the source code is black compliant
-in Appveyor during PRs. You can find installation instructions on [Black's docs](https://black.readthedocs.io/en/stable/installation_and_usage.html).
-Install version 19.10b0 as this is what is currently used in the CI/CD pipeline.
+in Appveyor during PRs. Black will be installed automatically with `make init`.
```
hoffa, line 51: It doesn't seem it will; no reference to `black` in any of the requirements files. 🤔

requirements/dev.txt
```diff
@@ -8,4 +8,7 @@ parameterized==0.7.0
 pytest-xdist==1.30.0
 pytest-forked==1.1.3
 pytest-timeout==1.3.3
-pytest-rerunfailures==7.0
\ No newline at end of file
+pytest-rerunfailures==7.0
+
+# formatter
+black==20.8b1
```
sriram-mv, line 14: I like that we have this effectively defined, instead of relying on user to have a version of black installed. This has caused conflicts in the past, with files being formatted with different versions of black.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,9 @@
 repos:
-  - repo: https://github.com/python/black
-    rev: 19.3b0
-    hooks:
-      - id: black
-        language_version: python3.7
-        exclude_types: ['markdown', 'ini', 'toml']
+- repo: local
```
sriram-mv, line 2: could you explain a bit what the new changes are?

aahung: The black's pre-commit-config.yaml is taken from https://github.com/psf/black/blob/master/.pre-commit-config.yaml

aahung:
(just did a rebase on latest upstream/develop)

hoffa (approved):
LGTM 👍 Big improvement versus having versions scattered around!

aahung:
Just did a rebase and update the dev guide with a workaround how to configure black in IDE.

CoshUS (approved):

Issue #2311, opened 2020-10-18
`sam deploy` does not warn the user of an invalid samconfig.toml


benkehoe:
### Description

With an invalid `samconfig.toml`, `sam deploy` does not warn the user of this and proceeds, including failing because necessary arguments provided in the config file are now missing. With `--guided`, it does warn the user of this.

### Steps to reproduce

Create a syntax error in your samconfig.toml, e.g. a missing start quote
```toml
[default.deploy.parameters]
stack_name=my_stack"
```

1. Run `sam deploy`
2. Run `sam deploy --guided`

### Observed result

1.
```
$ sam deploy --debug
Usage: sam deploy [OPTIONS]
Try 'sam deploy --help' for help.

Error: Missing option '--stack-name', 'sam deploy --guided' can be used to provide and save needed parameters for future deploys.
```

2.
```
$ sam deploy --guided --debug
Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics

Configuring SAM deploy
======================

        Looking for config file [samconfig.toml] :  Found
        Reading default arguments  :  Failure
Sending Telemetry: <snip>
Error: Syntax invalid in samconfig.toml; save values through sam deploy --guided to overwrite file with a valid set of values.
```

### Expected result

The error provided with `--guided` is helpful, and and a similar error should be provided whenever `samconfig.toml` has invalid syntax.

### Additional environment details

1. OS: WSL (Ubuntu 18.04) on Windows 10
2. `sam --version`: SAM CLI, version 1.6.2

hoffa:
Thanks for reporting the issue! Fixed in #2316.

mndeveci:
The fix has been released with [SAM CLI v1.7.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.7.0).

PR #2316, opened 2020-10-20
fix: fail on invalid config

hoffa:
*Issue #, if available:* #2311 

- Prevent issues in the config from being ignored.
- Improve UX by stating the root cause as early as possible.

### How does it address the issue?

Fail immediately if a non-`KeyError` exception is thrown when processing the config.

For example, with the following `samconfig.toml`:

```toml
version = 0.1
[default.deploy.parameters]
stack_name = _sam-app"
```

Instead of cryptic error messages as in #2311, `sam deploy` and `sam deploy --guided` will now show the following output:

```bash
Error: Error reading configuration: Unexpected character: '_' at line 3 col 13
```

A few notes regarding the PR:
- Keeping the `KeyError` handling as-is for now as it's not directly relevant to the issue and there might be marginal benefit to it (e.g. allow specifying a custom `config_env` without the corresponding TOML table).
- Removed the [`sanity_check()`](https://github.com/aws/aws-sam-cli/blob/3edb55cd8dba9898ddf2e1c8bd9eeba0434c08de/samcli/lib/config/samconfig.py#L134)  as the `get_all()` below it already validates the config by calling `_read()` (which is also what `sanity_check()` does, except by swallowing exceptions).

### What side effects does this change have?

`sam deploy` fails early if there's a problem reading the config.

### Did you change a dependency in `requirements/base.txt`?

No.

### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -13,7 +13,6 @@ from pathlib import Path
 import click
 
 from samcli.commands.exceptions import ConfigException
-from samcli.lib.config.exceptions import SamConfigVersionException
 from samcli.cli.context import get_cmd_names
 from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV, DEFAULT_CONFIG_FILE_NAME
 
@@ -76,7 +75,6 @@ class TomlProvider:
 
             # NOTE(TheSriram): change from tomlkit table type to normal dictionary,
             # so that click defaults work out of the box.
-            samconfig.sanity_check()
             resolved_config = dict(samconfig.get_all(cmd_names, self.section, env=config_env).items())
             LOG.debug("Configuration values successfully loaded.")
             LOG.debug("Configuration values are: %s", resolved_config)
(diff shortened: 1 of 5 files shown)


samcli/cli/cli_config_file.py
```diff
@@ -76,7 +75,6 @@ def __call__(self, config_path, config_env, cmd_names):
 
             # NOTE(TheSriram): change from tomlkit table type to normal dictionary,
             # so that click defaults work out of the box.
-            samconfig.sanity_check()
```
hoffa, line 79: Note: unnecessary as the `get_all()` below already validates the config by calling `_read()` (which is also what [`sanity_check()`](https://github.com/aws/aws-sam-cli/blob/3edb55cd8dba9898ddf2e1c8bd9eeba0434c08de/samcli/lib/config/samconfig.py#L134) does, except by swallowing the useful `tomlkit.exceptions.TOMLKitError`).

c2tarun, line 79: What was this doing here? We don't need it anymore?

hoffa: Right, I explain the why in the summary. Basically `get_all()` already does what `sanity_check()` does, except that `sanity_check()` swallows the important TOML exception with the useful error message.

c2tarun (approved):

jfuss (changes requested):
Can we add an integ tests that covers the invalid samconfig case?

hoffa:
Added unit/integ tests!

jfuss (approved):

sriram-mv (approved):
Nice fix!

PR #2318, opened 2020-10-21
Export the ApplicationBuilder parallel option to the cli

ItayGoren:
*Issue #, if available:* no issue

The application builder supports building the different function parallelly.
Yet, this option was always set to false. This commit adds the option to toggle
it to true from the command line as a flag.
To keep backward compatibility, the default value is False - single thread.

*What side effects does this change have?* none

*Did you change a dependency in `requirements/base.txt`?* no - no reason

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -33,6 +33,7 @@ class BuildContext:
             manifest_path=None,
             clean=False,
             use_container=False,
+            parallel=False,
             parameter_overrides=None,
             docker_network=None,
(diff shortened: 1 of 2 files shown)


jfuss:
The library was originally built with parallel in mind but was never implemented. This is why it was never exposed. 

However, we do have a WIP PR for parallel builds: #2274 

Closing as parallel builds is not implemented and the reason it is not exposed in the command line options.

Issue #2324, opened 2020-10-23
--debugger-path is not validated without --debug-port


justinmk3:
### Description

1. `--debugger-path` is not (fully) validated unless `--debug-port` is given. https://github.com/aws/aws-sam-cli/blob/2201b17bff0a438b934abbb53f6c76eff9ccfa6d/samcli/commands/local/cli_common/invoke_context.py#L340
2. Help description of `--debugger-path` does not mention anything about a directory.

### Steps to reproduce

If `--debug-port` is not given, setting `--debugger-path` to a non-directory is silently ignored (lambda is invoked without debugging):

    sam local invoke Foo --template py38/template.yaml --debugger-path /venv/bin/python
    Invoking app.lambda_handlerPy38 (python3.8)
    ...

Setting `--debugger-path` to a binary shows an error (correctly) if `--debug-port` is given:

    sam local invoke Foo --template ./template.yaml --debugger-path /venv/bin/python --debug-port 42
    Error: '/Volumes/workplace/aws-toolkit-vscode/venv/bin/python' should be a directory with the debugger in it.

### Expected result

1. Mention "directory" in `sam local invoke --help` description of `--debugger-path`: 
    - > "Host path to a debugger **directory** that will be mounted into the Lambda container."
2. Always validate `--debugger-path`.
    - Or always require `--debug-port` if `--debugger-path` is given. Possibly a breaking change :(
### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS
2. `sam --version`: `SAM CLI, version 1.2.0`

PR #2330, opened 2020-10-27
validate debugger-path parameter if it is provided

moelasmar:
*Issue #, if available:* https://github.com/aws/aws-sam-cli/issues/2324

to always validate the debugger-path parameter even if the debug-port is not provided

The change is to validate the debugger-path if it is provided regardless the value of the debug-port.

no side effects.

no

*Checklist:*

- [x] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/cfn_api_provider.py
+++ b/samcli/lib/providers/cfn_api_provider.py
@@ -401,7 +401,9 @@ class CfnApiProvider(CfnBaseApiProvider):
     def _parse_route_key(route_key):
         """
         parse the route key, and return the methods && path.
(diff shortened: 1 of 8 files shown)


jfuss (commented):
Do we want to validate the debugger_path if we are not in debug mode? Currently, `debugger_path` only matters if you pass a `debug_port`. If you do not pass a `debug_port`, SAM CLI is not in debug mode and is run like a regular invoke.

moelasmar:
So we should not accept debugger_path unless the debug_port is provided .. right ?

PR #2338, opened 2020-10-29
refactor: get rid of hardcoded test template list

hoffa:
*Issue #, if available:*

Whenever we release a new SAM Translator, we:

1. Copy test templates from
[`serverless-application-model/tests/translator/input`](https://github.com/aws/serverless-application-model/tree/develop/tests/translator/input) to
[`aws-sam-cli/tests/functional/commands/validate/lib/models`](https://github.com/aws/aws-sam-cli/tree/develop/tests/functional/commands/validate/lib/models)
2. Update `VALID_TEST_TEMPLATES` to include every file copied to `models`.

### How does it address the issue?

This PR gets rid of step (2), instead getting the list of test templates at runtime.

It will simplify the deployment process as we don't need to maintain error-prone scripts that patch `VALID_TEST_TEMPLATES`.

There are no differences between the list of hardcoded files and files in `models`:

```bash
DIR=tests/functional/commands/validate/lib
diff \           
    <(ls $DIR/models/* | sort) \
    <(grep $DIR/models $DIR/test_sam_template_validator.py | cut -d\" -f2 | sort)
```

### What side effects does this change have?

Scripts that patch `VALID_TEST_TEMPLATES` won't work.

### Did you change a dependency in `requirements/base.txt`?

No.

### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


aahung (commented):
This is much cleaner. Do the numbers match?

tests/functional/commands/validate/lib/test_sam_template_validator.py
```diff
@@ -7,216 +8,13 @@
 from samcli.commands.validate.lib.sam_template_validator import SamTemplateValidator
 from samcli.commands.validate.lib.exceptions import InvalidSamDocumentException
 
+# Out of scope because https://stackoverflow.com/a/47224266
```
aahung, line 11:
```suggestion
# Out of TestValidate's scope because https://stackoverflow.com/a/47224266
```

aahung: If I did not have the context, I would not get what "out of scope" means here.. out of this PR's scope or out of this file's scope 😂

aahung (approved):

sriram-mv (approved):

c2tarun (approved):

Issue #2344, opened 2020-10-31
Layer won't deploy: "Layer conversion failed" error


dan-mba:
### Description

Deploying my application fails on layer creation with a file permission error.
I tried to `chmod -R 755` the layer build directory in git bash, but it had no effect.

### Steps to reproduce

1. clone my [repo](https://github.com/dan-mba/aws-sam-rekognition)
1. sam build --use-container
1. sam deploy --guided

### Observed result

CREATE_FAILED
AWS::Lambda::LayerVersion
OpencvLayer641c5be385
Layer conversion failed: Some files do not have read permissions;
(Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException;
Request ID: b212fad5-ef52-4386-b1de-d391533f9c68)

### Expected result

Deployment to complete

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 8.1
2. `sam --version`: `SAM CLI, version 1.7.0`

dan-mba:
I was able to get my code working by moving all of my dependencies into my Lambda.
See the dev branch of my repo.

But with my bundle being almost 190MB with 99% being dependencies, 
this is not a good way to develop quickly having to constantly re-uploading my dependencies.

codestart:
I am getting this exact error with my Node.js app. Thankfully my bundle is only 13.8MB but still, not ideal. This is what I am doing as a workaround too.
I have the same code as I had a few weeks back (that worked) and my code runs fine when run locally. The only change (I haven't rolled-back) is the change to SAM CLI 1.7.0... unless I'm missing something :(

codestart:
@dan-mba I've uninstalled v1.7.0 and installed v1.6.2 and it runs fine 👍

singledigit:
Hey @dan-mba, is this layer being managed in the same template? Or is it an external layer? Can you post a copy of the SAM template?

dan-mba:
@singledigit The template is in my [repo here](https://github.com/dan-mba/aws-sam-rekognition).

The source for the layer is only the requirements.txt for the dependencies. They contain platform specific modules, which is why I have to build with container.

jholladay10:
[...]

PR #2356, opened 2020-11-06
Layer Deployment Permission Issue

CoshUS:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli/issues/2344
Lambda needs read permission as owner on artifacts.

Changed Windows mocked permission to 755, mimicking Unix system.
Previously, it was changed from default 644 to 005 to give execute permission for go artifacts. Owner and group owner permission bits were removed assuming Lambda itself (not the user executing the code) do not need permission on the artifacts. However, as regression issue https://github.com/aws/aws-sam-cli/issues/2344 stated, it does require read permission on the Layer artifacts.

There should be no security concern as owner and group owner are set and used by Lambda.

Owner and group owner permission bits will be set to 7 and 5 respectively.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mndeveci (approved):
LGTM, and I tested it on Windows and confirmed it is working.

qingchm (approved):

awsjeffg:
@CoshUS THANK YOU!

OK-Apps:
still getting same problem even with version 1.9.0 :

$ sam --version
SAM CLI, version 1.9.0

-----

PDATE_FAILED            AWS::Lambda::Function    StagingKinesisSenderEx   Layer conversion
                                                  tended                   failed: Some files do
                                                                           not have read
                                                                           permissions; (Service:
                                                                           AWSLambdaInternal;
                                                                           Status Code: 400;
                                                                           Error Code: InvalidPar
                                                                           ameterValueException;
                                                                           Request ID: 4db54727-3
                                                                           8f7-41c9-a648-655361a2
                                                                           2882; Proxy: null)
UPDATE_FAILED            AWS::Lambda::Function    StagingKinesisPutRecor   Layer conversion
                                                  d                        failed: Some files do
                                                                           not have read
                                                                           permissions; (Service:
                                                                           AWSLambdaInternal;
                                                                           Status Code: 400;
                                                                           Error Code: InvalidPar
                                                                           ameterValueException;
                                                                           Request ID: a05a655b-0
                                                                           cfb-4f57-9025-2ab308b2
                                                                           068d; Proxy: null)
UPDATE_ROLLBACK_IN_PRO   AWS::CloudFormation::S   Staging-StagingKinesis   The following
GRESS                    tack                                              resource(s) failed to
                                                                           update: [StagingKinesi
                                                                           sSenderExtended, Stagi
                                                                           ngKinesisPutRecor


---

CoshUS:
@OK-Apps It seems this could be caused by wrong permission set in the existing resource with previous version of SAM CLI. Can you try manually deleting those two resources and re-deploy again?

OK-Apps:
> @OK-Apps It seems this could be caused by the wrong permission set in the existing resources using the previous version of SAM CLI. Can you try manually deleting those two resources and re-deploy again?

@CoshUS by "resource" do you mean the lambda functions or the lambda layers ? (or both?)
and, 
in case if it is only about lambda functions (and not layers) - 
if I was able to successfully do a 'sam deploy; from a non-Windows machine (mac) - so does it count as same effect 
as "manually delete"? - so would we expect if we now try again from a windows machine with 1.9.0 

thank you!

CoshUS:
@OK-Apps 

> do you mean the lambda functions or the lambda layers ? (or both?)

My hunch would be only layers, but try delete or overriding both.

> if I was able to successfully do a 'sam deploy; from a non-Windows machine (mac) - so does it count as same effect
as "manually delete"? - so would we expect if we now try again from a windows machine with 1.9.0

If your deployment on Mac overrides the existing layer artifact with correct permission bits, then it should be the same as deleting and creating a new one.

Thanks!

OK-Apps:
@CoshUS 

an update of current state of affairs - everyone in a team has SAM CLI 1.9.0
when doing 'sam build' & 'sam deploy' from Mac - all works ok
when doing same from various Windows 10 machines - getting the error:

Layer conversion failed: Some files do not have read permissions; 
(Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException;
Request ID: 4df2bf50-f584-4e85-9538-5173260b072e; Proxy: null)

OK-Apps:
@CoshUS so puled how is it even possible - if things take place at server-side- at AWS, 
why it behaves differently when triggering it from Mac VS Windows ?

OK-Apps:
@CoshUS 
regarding layers themselves - some layers we did not even use sam to create them - they been created/uploaded "manually" into AWS console

so what we need to do with those layers? (if we need to do anything with layers to fix this)

CoshUS:
@OK-Apps 
A few clarifying questions:

CoshUS:
A few clarifying questions:
> regarding layers themselves - some layers we did not even use sam to create them - they been created/uploaded "manually" into AWS console

Were the layers created with sam deployed using SAM CLI 1.7.0 or 1.8.0 on Windows?

> an update of current state of affairs - everyone in a team has SAM CLI 1.9.0
when doing 'sam build' & 'sam deploy' from Mac - all works ok
when doing same from various Windows 10 machines - getting the error:

Are these deployments using the same layers and the same Lambdas? 

------

> how is it even possible - if things take place at server-side- at AWS

Lambda is based on Linux which uses unix permission bits. When packaging a layer artifact, Windows does not set permission bits in the zip file which can cause missing permission errors when uploaded and unpacked into Lambda. We changed this behavior in 1.7.0 and forced all file permissions to be 005 when packaged in Windows. This had unintended consequences in deploying layers that can result in `Layer conversion failed: Some files do not have read permissions;` We then changed the permission to 755 to be more permissible in SAM CLI 1.9.0. 

If your layers are created manually in Windows, it should have the default 644 permission and not cause read permission issue.

kurax:
I'm using the lastest release of SAM CLI as 1.10.0 and this issue still remains.
I've deleted all my deployed functions by comment them out in template.yml, then uncomment try to recreate and got this:
> Layer conversion failed: Some files do not have read permissions; (Service: AWSLambdaInternal; Status Code: 400; Error Code: InvalidParameterValueException; Request ID: 194fed35-a3b8-4ed5-8b17-d93823171bbc; Proxy: null)

Any ideas?

OK-Apps:
HI @CoshUS 
thank you for update - and yes it is our understadning s well ... we have observed the issue with Windows permisions in GoLang  lambda
and seen them fixed in 1.7 ... but then started seeing this "layers issue" with Python lambdas
and now even with 1.9 - still experiencing it (on Windows machines !!)
despite the fact that 
"layers are created manually in Windows, it should have the default 644 permission and not cause read permission issue."

yes all our Python layers been manually created on Windows and yet we still get this error with SAM CLI 1.9 when running 
on Windows machines

CoshUS:
@OK-Apps  @kurax 
Thanks for the information. Created issue #2392 to track this problem.

`Is any version of SAM CLI expected to fix/solve this?`
SAM CLI 1.6.2 does not have the Go permission fix. Can you give that a try?

CoshUS:
@OK-Apps 
Are you able to provide a sample template that reproduces the issue?

OK-Apps:
@CoshUS 

> @OK-Apps
> Are you able to provide a sample template that reproduces the issue?

pls see attached template.zip


also, 
yes it worked ok under SAM CLI 1.5/1.6 

thank you!
[template.yml.zip](https://github.com/aws/aws-sam-cli/files/5610079/template.yml.zip)

esscott2112:
I am having the exact same issue with Window machine.  SAM deployment works with Linux but not Windows for all the reasons listed.  When will this be fixed in the SAM so my development team can SAM deploy from windows machine?

PR #2367, opened 2020-11-10
ci: check PR title

hoffa:
*Issue #, if available:*

We'll be auto-generating release notes; ensures PR titles are somewhat consistent.

Run regex check on every PR title.

*Checklist:*

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -6,7 +6,6 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
       - name: Check PR title
         run: |
           title=$(jq -r .pull_request.title "$GITHUB_EVENT_PATH")
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -10,4 +10,7 @@ jobs:
       - name: Check PR title
         run: |
           title=$(jq -r .pull_request.title "$GITHUB_EVENT_PATH")
-          [[ $title =~ ^.*:\ .*$ ]] || exit 1
+          if [[ ! $title =~ ^.*:\ .*$ ]]; then
+            echo "Pull request titles must adhere to Conventional Commits: https://www.conventionalcommits.org"
+            exit 1
+          fi
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -1,0 +1,13 @@
+name: Check PR title
+on:
+  pull_request:
+    types: [opened, edited]
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_title.yml
```diff
@@ -0,0 +1,13 @@
...
+    types: [opened, edited]
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check PR title
+        run: |
+          title=$(jq -r .pull_request.title "$GITHUB_EVENT_PATH")
+          [[ $title =~ ^.*:\ .*$ ]] || exit 1
```
aahung, line 13: Can we output an error message when the title is not qualified? Like "PR title should follow this format: XYZ"

hoffa: I'm a bit ambivalent about this.

Adding that would essentially double the amount of code (need print message then exit, so need a separate `if` or subprocess), and require maintaining the error message. Also I trust any hacker will immediately understand the issue. I'm happy to be convinced otherwise if you think it'll be a valuable UX improvement.

hoffa: Eh maybe you're right, it might be obvious what's the issue, but not necessarily obvious how to fix it. Let me update it.

aahung: I am not sure whether everyone familiar with the regex. But an easy fix would be to update the github action stage name to "Check PR title to match format "*: *"

aahung:
My another question is, why we need to enforce PR title? or why we enfore PR title instead of commit messages?

aahung:
> We'll be auto-generating release notes; ensures PR titles are somewhat consistent.

Are we using PR titles or commit messages for release note generation?

hoffa:
> > We'll be auto-generating release notes; ensures PR titles are somewhat consistent.
> 
> Are we using PR titles or commit messages for release note generation?

Certainly looks like it: https://github.com/aws/aws-sam-cli/releases

aahung:
> > > We'll be auto-generating release notes; ensures PR titles are somewhat consistent.
> > 
> > 
> > Are we using PR titles or commit messages for release note generation?
> 
> PR titles: https://github.com/aws/aws-sam-cli/releases

It looks like the commit messages' first lines to me 🤔 when PR owners merge the PR, they can still change the title when squashing right? When the commit title differs from PR title, which will get picked up?

hoffa:
They are commit messages from `master`, which in turn are from the PR title. GitHub automatically appends the `(#xxx)` to the PR title too.

aahung (approved):

qingchm (approved):

jfuss (approved):

PR #2375, opened 2020-11-13
chore: Enable some pylint rules and add clarity to pylintrc file

aahung:

Currently there are multiple pylint rules we disable without clarity. It is hard to find which rules are necessarily to disable, and which are not.

This PR exposes the rule name in pylintrc file, and (re)enabled some rules

absolutely no side effect

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -2,8 +2,9 @@
 CLI configuration decorator to use TOML configuration files for click commands.
 """
 
-## This section contains code copied and modified from [click_config_file][https://github.com/phha/click_config_file/blob/master/click_config_file.py]
-## SPDX-License-Identifier: MIT
+# This section contains code copied and modified from
+# [click_config_file][https://github.com/phha/click_config_file/blob/master/click_config_file.py]
+# SPDX-License-Identifier: MIT
 
 import os
 import functools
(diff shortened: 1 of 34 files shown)


samcli/commands/build/build_context.py
```diff
@@ -24,20 +24,20 @@ class BuildContext:
     _BUILD_DIR_PERMISSIONS = 0o755
 
     def __init__(
-            self,
```
hoffa, line 27: Why is the formatting different?

mgrandis (approved):

samcli/commands/build/build_context.py
hoffa: Well as per @aahung 

>I just ran black on that file, it does not do anything, then I made a small change, it reformatted it

It seems `black` isn't yet stable. 😩😩😩

Great PR otherwise, approved.

hoffa (approved):

samcli/cli/cli_config_file.py
```diff
@@ -263,4 +264,4 @@ def decorator_customize_config_env(f):
     return click.option(*config_env_param_decls, **config_env_attrs)(f)
 
 
-# End section copied from [[click_config_file][https://github.com/phha/click_config_file/blob/master/click_config_file.py]
+# End section copied from [click_config_file][https://github.com/phha/click_config_file/blob/master/click_config_file.py
```
elbayaaa, line 267: [minor] will the hyperlink work with the last close bracket missed?

aahung: it is a python file so hyperlink won't work anyway

aahung: I don't feel the original comment was trying to follow format like markdown 🤔

PR #2376, opened 2020-11-14
chore: Fix black misses some files during formatting

aahung:

Some files that was left behind by black, caused by a bug in black: https://github.com/psf/black/issues/1820

Use `find` to list all python files instead of relying on black for directory traversal. 

No files will be missed:

```
$ make black
find setup.py samcli tests -type f -name "*.py" | xargs black
reformatted /Users/xinhol/Projects/aws-sam-cli/tests/integration/testdata/buildcmd/Provided/main.py
reformatted /Users/xinhol/Projects/aws-sam-cli/tests/integration/testdata/package/main.py
reformatted /Users/xinhol/Projects/aws-sam-cli/tests/integration/testdata/invoke/main.py
reformatted /Users/xinhol/Projects/aws-sam-cli/tests/integration/testdata/start_api/main.py
reformatted /Users/xinhol/Projects/aws-sam-cli/samcli/commands/build/build_context.py
reformatted /Users/xinhol/Projects/aws-sam-cli/samcli/commands/build/command.py
reformatted /Users/xinhol/Projects/aws-sam-cli/samcli/lib/build/build_strategy.py
reformatted /Users/xinhol/Projects/aws-sam-cli/samcli/lib/build/app_builder.py
All done! ✨ 🍰 ✨
8 files reformatted, 418 files left unchanged.

```

no side effect

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,10 @@ lint:
 dev: lint test
 
 black:
-    # command "black <directory>" will skip some files, it is a bug, see https://github.com/psf/black/issues/1820
-	find setup.py samcli tests -type f -name "*.py" | xargs black
+	black setup.py samcli tests
 
 black-check:
-	find setup.py samcli tests -type f -name "*.py" | xargs black --check
+	black --check setup.py samcli tests
 
 # Verifications to run before sending a pull request
 pr: init dev black-check
--- a/Makefile
+++ b/Makefile
@@ -40,10 +40,11 @@ lint:
 dev: lint test
 
 black:
-	black setup.py samcli tests
+    # command "black <directory>" will skip some files, it is a bug, see https://github.com/psf/black/issues/1820
+	find setup.py samcli tests -type f -name "*.py" | xargs black
 
 black-check:
-	black --check setup.py samcli tests
+	find setup.py samcli tests -type f -name "*.py" | xargs black --check
 
 # Verifications to run before sending a pull request
 pr: init dev black-check
--- a/tests/integration/local/start_lambda/test_start_lambda.py
+++ b/tests/integration/local/start_lambda/test_start_lambda.py
@@ -188,11 +188,11 @@ class TestLambdaService(StartLambdaIntegBaseClass):
         statement_raising_exception = 'raise Exception("Lambda is raising an exception")'
         statement_raising_exception_line_no = 0
         with open("tests/integration/testdata/invoke/main.py", "r") as f:
-            for i, line in enumerate(f):
+            for line_no, line in enumerate(f, 1):
                 if line.strip() == statement_raising_exception:
-                    statement_raising_exception_line_no = i + 1
+                    statement_raising_exception_line_no = line_no
         if not statement_raising_exception:
-            raise Exception(f'Cannot find statement "{statement_raising_exception}" in invoke/main.py')
+            self.fail(f'Cannot find statement "{statement_raising_exception}" in invoke/main.py')
 
         self.assertEqual(
             response_data,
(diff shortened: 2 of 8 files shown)


hoffa (approved):
Yay, great find and temporary workaround 👌

jfuss:
@aahung This looks good to me but there is an integ test that is failing. The `<tests.integration.local.start_lambda.test_start_lambda.TestLambdaService testMethod=test_lambda_function_raised_error>` validates that the exception response is correct. This has a line number from the source and we validate it is correct. Can you update that test to be the new line number?

Makefile
```diff
@@ -40,10 +40,11 @@ lint:
 dev: lint test
 
 black:
-	black setup.py samcli tests
+    # command "black <directory>" will skip some files, it is a bug, see https://github.com/psf/black/issues/1820
+	find setup.py samcli tests -type f -name "*.py" | xargs black
```
hoffa, line 44: Should we use `black --include build setup.py samcli tests` so that we don't deviate too much in the meantime?

aahung: good point! I did a big more digging and found we have a configuration file 🙂

Makefile
```diff
@@ -43,7 +43,7 @@ black:
 	black setup.py samcli tests
 
 black-check:
-	black --check setup.py samcli tests
+	black setup.py samcli tests --check
```
aahung, line 46: this change was not intentional. I don't want to change it back because pr checking is forever 😢

hoffa: If we're being pedantic the order _is_ `black [OPTIONS] [SRC]...`... I don't think `black` cares but other tools do.

aahung: please don't ask me to change this tiny thing back. It will be another hour running the PR check 🥺

hoffa: Haha I didn't revoke the approval!

aahung: since jacob is not likely to approve it today, let me just fix it

tests/integration/local/start_lambda/test_start_lambda.py
```diff
@@ -184,13 +184,28 @@ def test_lambda_function_raised_error(self):
 
         print(response_data)
 
+        # determine the line number of the statement that will raise the error
+        statement_raising_exception = 'raise Exception("Lambda is raising an exception")'
+        statement_raising_exception_line_no = 0
+        with open("tests/integration/testdata/invoke/main.py", "r") as f:
+            for i, line in enumerate(f):
```
jfuss, line 191: Instead of adding 1 to `i` on line 193, you can do `for i, line in enumerate(f, 1)`. Which will start `i` as 1 instead of 0. 

Not material here but just wanted to bring it up so you know for next time.

tests/integration/local/start_lambda/test_start_lambda.py
```diff
@@ -184,13 +184,28 @@ def test_lambda_function_raised_error(self):
...
+                if line.strip() == statement_raising_exception:
+                    statement_raising_exception_line_no = i + 1
+        if not statement_raising_exception:
+            raise Exception(f'Cannot find statement "{statement_raising_exception}" in invoke/main.py')
```
jfuss, line 195: We should use `self.fail()` here instead of an exception.

jfuss (changes requested):
Sorry I wasn't able to get to this yesterday afternoon. Consider the update to enumerate but only ask to to address the `self.fail` over raising exceptions in tests.

tests/integration/local/start_lambda/test_start_lambda.py
aahung: addressed, good one

tests/integration/local/start_lambda/test_start_lambda.py
aahung: addressed

aahung:
close in favor of #2517

PR #2385, opened 2020-11-18
ci: Enable basic mypy typechecking

aahung:

We don't have typecheck, typecheck can make our code safer.

[Warnings and errors output by mypy without changing any code](https://paste.amazon.com/show/xinhol/1605722224)

#####  ~I include a demo change how the code is safer without using Dict directly:~
~With the intro of the following dataclass, mypy can typecheck them so we won't mis-spell dict keys.~
- ~`TemplateOption`~
- ~`RuntimeDepInfo`~
this change has been separated from this PR and moved to https://github.com/aahung/aws-sam-cli/commit/b5e2b1bd34b7011cdd93192e1279e3675d77231f

##### Some benefits we see immediately:
- `test_configure_samcli_logger` is defined twice, and one of which cannot be triggered
- `test_download_source_code_binding` is defined twice, and they are identical
- mypy also caught a bunch of incorrect use of `.format()` in "tests/unit/lib/init/test_init.py"
- it caught a Python 2 import
- it enforces "getter" and "setter" should be next to each other, increasing readability instantly
- it caught some imports that are discouraged by package developers

N/A

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/provider.py
```diff
@@ -251,7 +251,7 @@ def binary_media_types(self):
 
 _CorsTuple = namedtuple("Cors", ["allow_origin", "allow_methods", "allow_headers", "max_age"])
 
-_CorsTuple.__new__.__defaults__ = (
+_CorsTuple.__new__.__defaults__ = (  # type: ignore
```
aahung, line 254: mypy won't acknowledge private attribute. So we need to disable it before we find an alternative

samcli/lib/schemas/schemas_aws_config.py
```diff
@@ -2,7 +2,7 @@
 
 import click
 
-from boto3 import Session
+from boto3.session import Session
```
aahung, line 5: mypy will be happier if we can import from the original class declaration file

samcli/local/common/runtime_template.py
```diff
@@ -85,14 +95,17 @@
     "java8.al2": ["maven", "gradle"],
 }
 
+
 SUPPORTED_DEP_MANAGERS = {
-    c["dependency_manager"]
-    for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))
-    if c["dependency_manager"]
+    c.dependency_manager
+    for c in list(itertools.chain.from_iterable(RUNTIME_DEP_TEMPLATE_MAPPING.values()))
```
aahung, line 101: `chain.from_iterable` is better at perserving types than `chain(*`

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -439,7 +439,7 @@ def test_invoke_returns_expected_results_from_git_function_with_parameters(self)
 
 
 class TestSamInstrinsicsAndPlugins(InvokeIntegBase):
-    template = "template-pseudo-params.yaml"
+    template = Path("template-pseudo-params.yaml")
```
aahung, line 442: All other subclasses of InvokeIntegBase assign `template` a `Path` value, except for this. Good catch by mypy

aahung (commented):
Here are some notes about why I made these changes

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -527,7 +530,7 @@ def test_fn_sub_second_argument_invalid_formats(self, name, intrinsic):
 
     @parameterized.expand(
         [
-            ("If Fn::Sub is a list, it should only have 2 arguments".format(item), item)
+            (f"If Fn::Sub is a list, it should only have 2 arguments {', '.join(item)}", item)
```
hoffa, line 533: Why the join here but `{item}` a few lines up? If it doesn't matter I'd go with less code.

tests/unit/commands/deploy/test_guided_config.py
```diff
@@ -3,7 +3,9 @@
 from unittest.mock import patch
 
 import click
-from click.globals import _local
+
+# TODO: importing private module property, to investigate alternative for stability
+from click.globals import _local  # type: ignore
```
hoffa, line 8: Why does mypy care about this? 🤔 Wouldn't setting types on the "private" method work?

samcli/settings/__init__.py
```diff
@@ -17,7 +17,7 @@
...
 import os
 
-if "__SAM_CLI_TELEMETRY_ENDPOINT_URL" not in os.environ:
-    telemetry_endpoint_url = "https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics"
-else:
-    telemetry_endpoint_url = os.getenv("__SAM_CLI_TELEMETRY_ENDPOINT_URL")
+
+telemetry_endpoint_url = os.getenv(
+    "__SAM_CLI_TELEMETRY_ENDPOINT_URL", "https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics"
+)
```
aahung, line 23: this change was addressing inconsistent types:

- above: `str`
- below: `Optional[str]`

requirements/base.txt
```diff
@@ -15,3 +15,4 @@ requests==2.23.0
 serverlessrepo==0.1.10
 aws_lambda_builders==1.1.0
 tomlkit==0.7.0
+dataclasses~=0.8; python_version<"3.7"
```
hoffa, line 18: Is there no other reasonable way than introducing a new dependency and using dataclasses? It seems like a reasonable change, but would be nice for the first PR to only change what's absolutely necessary.

samcli/local/lambdafn/env_vars.py
```diff
@@ -131,6 +131,10 @@ def memory(self, value):
     def handler(self):
         return self._function["handler"]
 
+    @handler.setter
+    def handler(self, value):
+        self._function["handler"] = value
+
```
aahung, line 137: setter / getter needs proximity

samcli/lib/providers/provider.py
```diff
@@ -194,6 +194,10 @@ def name(self):
     def codeuri(self):
         return self._codeuri
 
+    @codeuri.setter
+    def codeuri(self, codeuri):
+        self._codeuri = codeuri
+
```
aahung, line 200: setter / getter needs proximity

samcli/lib/logs/formatter.py
```diff
@@ -5,14 +5,8 @@
...
-try:
-    # Python2
-    from itertools import imap
-except ImportError:
-    # Python3 already has `map` defined, alias to imap
-    # We do this to prevent accidentally using the built-in ``map`` in Python2. In Python2, ``map`` does a full
-    # evaluation of the iterator whereas imap does a lazy evaluation. For performance reasons, we need to use ``imap``.
-    from builtins import map as imap
+
+from builtins import map
```
aahung, line 9: Python 2 is no longer used, mypy is not happy to see this failed import

requirements/base.txt
aahung: I agree, but without getting ride of that Dict, mypy cannot infer types in quite some places..

aahung: I was aiming for NamedTuple at first, and then found they need to be mutable

tests/unit/commands/deploy/test_guided_config.py
aahung: mypy seems to be unhappy about any private imports.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
aahung: yeah it does not matter, I can remove the join, sounds better

requirements/base.txt
aahung: `dataclasses` is a pretty reputable and simple portback, I think it is pretty safe to use https://github.com/ericvsmith/dataclasses

hoffa: Yeah fair enough if not using it would be a significant detriment. I'm assuming without this mypy would complain a lot?

(Also if we merge this we should let the team know they're free to use dataclasses!)

aahung: You know what, I changed my mind! I think it is a great suggestion to make this PR easier to go through! I made another branch with these changes and will file a PR immediately after this one is merged 😋

aahung: They are moved here: https://github.com/aahung/aws-sam-cli/commit/b5e2b1bd34b7011cdd93192e1279e3675d77231f

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -149,7 +149,6 @@ def _verify_invoke_built_function(self, template_path, function_logical_id, over
 
 
 class BuildIntegRubyBase(BuildIntegBase):
-    EXPECTED_FILES_GLOBAL_MANIFEST = set()
```
aahung, line 152: this is not being used

samcli/local/common/runtime_template.py
```diff
@@ -85,14 +86,16 @@
     "java8.al2": ["maven", "gradle"],
 }
 
-SUPPORTED_DEP_MANAGERS = {
-    c["dependency_manager"]
+SUPPORTED_DEP_MANAGERS: Set[str] = {
+    c["dependency_manager"]  # type: ignore
```
aahung, line 90: mypy infers it as `Any`

samcli/local/common/runtime_template.py
```diff
@@ -85,14 +86,16 @@
...
     for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))
     if c["dependency_manager"]
 }
 
-RUNTIMES = set(
-    itertools.chain(*[c["runtimes"] for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))])
+RUNTIMES: Set[str] = set(
+    itertools.chain(
+        *[c["runtimes"] for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))]  # type: ignore
```
aahung, line 97: mypy infers it as `Any`

samcli/lib/generated_sample_events/events.py
```diff
@@ -5,7 +5,7 @@
 import os
 import json
 import base64
-from requests.utils import quote as url_quote
+from urllib.parse import quote as url_quote
```
aahung, line 8: following the import in `requests.utils`, the `quote` is actually imported from `urllib.parse` (for Python 3)

wchengru (approved):

hoffa (approved):
Great PR, thanks! 🙌 Good for ensuring long-term health.

We should also make it a soft policy to add missing types to any code touched by new PRs, so we can cover the important stuff fairly quick.

aahung:
just manually merged from develop

PR #2389, opened 2020-11-18
chore: Bump sam translator version, boto3 version bump to 1.15

qingchm:
Required in SAM release 1.30.0

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


aahung (approved):

requirements/base.txt
```diff
@@ -2,11 +2,11 @@ chevron~=0.12
 click~=7.1
 Flask~=1.1.2
 #Need to add Schemas latest SDK.
-boto3~=1.14.0, >=1.14.23
+boto3~=1.15.0
```
aahung, line 5: Since we found why there was conflict before, we can revert this one

aahung: or bump it to 1.16

c2tarun (approved):

requirements/base.txt
IrinaAlexandra: What happened here? Can it be bumped to 1.16 now?

aahung: it had conflict with samtranslator's boto3 requirements. It was fixed.


> Can it be bumped to 1.16 now?

yes we will. Is there any reason you want us to bump this version?

IrinaAlexandra: I understand. The reason I am asking is that by the time this change was made, boto3 was at 1.16.x. This in turn puts us in a position where we need to either use an older version of `aws-sam-cli` (which is not ideal) or an older version of `boto3` (also not ideal).

aahung: Oh I see. Are you using pip to install sam-cli? We recommend to use homebrew (macOS, linux) or installer (windows) to install it. In that case, you would not have this dependencies issue.

aahung: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html

aahung: Besides, I created a PR to loose the boto3's version https://github.com/aws/aws-sam-cli/pull/2493

Issue #2332, opened 2020-10-28
Unable to call sam deploy without setting --capabilities parameter. No way to set it to no capabilities


hauboldj:
### Description
There is no way to use `sam deploy` with out specifying the `--capabilities` parameter. I'm deploying a template without IAM roles or transformations. I also want to prevent future developers from adding IAM roles (this template will be deployed via a CI process).

Similar issue previously raise here: https://github.com/aws/aws-sam-cli/issues/1675

I'm using `sam deploy` (instead of aws cloudformation deploy) in this case to maintain consistency in the CI process between different repos in this project.

### Steps to reproduce

```
sam deploy \
		--template-file artifacts/template.packaged.yaml \
		--stack-name temp-test-stack- \
		--no-fail-on-empty-changeset

	Deploying with following values
	===============================
	Stack name                 : temp-test-stack-
	Region                     : None
	Confirm changeset          : False
	Deployment s3 bucket       : None
	Capabilities               : null
	Parameter overrides        : {}

Initiating deployment
=====================
Error: Failed to create changeset for the stack: temp-test-stack-, Parameter validation failed:
Invalid type for parameter Capabilities, value: None, type: <class 'NoneType'>, valid types: <class 'list'>, <class 'tuple'>
```
### Observed result

```
sam deploy \
		--template-file artifacts/template.packaged.yaml \
		--stack-name temp-test-stack- \
		--no-fail-on-empty-changeset \
		--debug
Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
Property Command.ScriptLocation of GlueJobOne resource is not a URL

	Deploying with following values
	===============================
	Stack name                 : temp-test-stack-
	Region                     : None
	Confirm changeset          : False
	Deployment s3 bucket       : None
	Capabilities               : null
	Parameter overrides        : {}

Initiating deployment
=====================
Collected default values for parameters: {'Connection': 'Redshift_dev'}
2 resources found in the template
Unable to create changeset
Traceback (most recent call last):
  File "/Users/jhaubold/.local/pipx/venvs/aws-sam-cli/lib/python3.8/site-packages/samcli/lib/deploy/deployer.py", line 191, in _create_change_set
    resp = self._client.create_change_set(**kwargs)
  File "/Users/jhaubold/.local/pipx/venvs/aws-sam-cli/lib/python3.8/site-packages/botocore/client.py", line 337, in _api_call
[...]

PR #2415, opened 2020-11-25
don't pass the --capabilities argument to cloudformation create_chang…

elbayaaa:
…eset if it is none.

#2332
Currently there is no way to not pass the **optional** argument "capabilities" to cloudformation `create_change_set`. This change is to not pass the argument if the user didn't set it during `sam deploy`
By not letting `sam deploy` to pass the capabilities argument to cloudformation `create_change_set` if it is none.
None
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/lib/deploy/test_deployer.py
```diff
@@ -213,6 +213,58 @@ def test_create_changeset_ClientErrorException_generic(self):
                 tags={"unit": "true"},
             )
 
+    def test_create_changeset_pass_through_optional_arguments_only_if_having_values(self):
```
moelasmar, line 216: can you split this test case ... it validates 2 different things

samcli/lib/deploy/deployer.py
```diff
@@ -180,6 +179,8 @@ def create_changeset(
                 kwargs["TemplateURL"] = s3_uploader.to_path_style_s3_url(parts["Key"], parts.get("Version", None))
 
         # don't set these arguments if not specified to use existing values
+        if capabilities is not None:
```
moelasmar, line 182: can you make it `if capabilities:`
I know this ways is more readable, but I am thinking if some how the passed capabilities is not None, but an empty list.

moelasmar (changes requested):

samcli/lib/deploy/deployer.py
elbayaaa: This is to be consistent with the reset of arguments below, for example we consider an empty `notification_arns` array to be a valid value to pass  to cloudformation. I agree though that an empty array should be treated as none but I don't want to change the current behavior as it is a customer-facing change,

tests/unit/lib/deploy/test_deployer.py
elbayaaa: which 2 different things? it validates that `capabilities`, `role_arn` and `notification_arns` will be passed to cloudformation only if they have values, otherwise, they won't

moelasmar: I mean that you can have 2 test cases, one if these properties have values, and the second one for the None values.
any way, it is a very minor comment

samcli/lib/deploy/deployer.py
moelasmar: ok, thanks for the clarification.

moelasmar (approved):

CoshUS (approved):

PR #2409, opened 2020-11-23
adding cloudformation package support for ServiceCatalog

rafaelliu:
#### Which issue(s) does this change fix?
This relates to [AWS CLI issue 3388](https://github.com/aws/aws-cli/issues/3388), asking for Service Catalog to be added to the list of resources supporting CLoudFormation's `package` command

#### Why is this change necessary?
Current implementation of `package` does not exporting attributes that have arrays in its path (e.g. a.b[].c) due to a limitation with JMESPath library and boto's utility function `set_value_from_jmespath`, currently used by `package`.

This is an issue for resources like Service Catalog's products, which needs a [AWS::ServiceCatalog::CloudFormationProduct.ProvisioningArtifactParameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicecatalog-cloudformationproduct.html#cfn-servicecatalog-cloudformationproduct-provisioningartifactparameters).

#### How does it address the issue?
This change allows for setting up attributes that have arrays in it's path, so "a.b.c" would act as JMESPath "a.b[].c".

#### What side effects does this change have?
There should be no side effect

#### Checklist

- [NA] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write unit tests
- [NA] Write/update functional tests
- [NA] Write/update integration tests
- [X] `make pr` passes
- [NA] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #2422, opened 2020-11-27
Adds cloudformation package support for ServiceCatalog #2409

rafaelliu:
This relates to [AWS CLI issue 3388](https://github.com/aws/aws-cli/issues/3388), asking for Service Catalog to be added to the list of resources supporting CLoudFormation's `package` command

Current implementation of `package` does not exporting attributes that have arrays in its path (e.g. a.b[].c) due to a limitation with JMESPath library and boto's utility function `set_value_from_jmespath`, currently used by `package`.

This is an issue for resources like Service Catalog's products, which needs a [AWS::ServiceCatalog::CloudFormationProduct.ProvisioningArtifactParameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicecatalog-cloudformationproduct.html#cfn-servicecatalog-cloudformationproduct-provisioningartifactparameters).

This change allows for setting up attributes that have arrays in it's path, so "a.b.c" would act as JMESPath "a.b[].c".

There should be no side effect

#### Checklist

- [NA] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write unit tests
- [NA] Write/update functional tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/lib/package/artifact_exporter.py
```diff
@@ -257,65 +257,71 @@ class Resource:
...
-            return
-
-        if isinstance(property_value, dict):
-            LOG.debug("Property %s of %s resource is not a URL", self.PROPERTY_NAME, resource_id)
-            return
-
+        context = {"resource_id": resource_id, "parent_dir": parent_dir}
+        _transform_dict(resource_dict, self.PROPERTY_NAME, self.parse_property, context)
+
+    #    def parse_property(self, property_name, property_value, resource_id, parent_dir):
```
aahung, line 270: Looks like this one was left behind.

samcli/commands/_utils/template.py
```diff
@@ -161,6 +156,31 @@ def _update_relative_paths(template_dict, original_root, new_root):
...
+        value = sub_dict[next_prop]
+        if len(sub_path) == 1:
+            sub_dict[next_prop] = transformation(template_dict, path, next_prop, value, context)
+        else:
+            if isinstance(value, list):
+                sub_dict[next_prop] = [visit(v, sub_path[1:]) for v in value]
+            elif isinstance(value, dict):
+                sub_dict[next_prop] = visit(value, sub_path[1:])
+            else:
+                raise Exception("Unsupported path '%s' or dict structure '%s'" % path, template_dict)
```
mgrandis, line 177: `Exception` is too generic, you can maybe create a `DictTransformException` with `UsedException` as base class

samcli/commands/_utils/template.py
```diff
@@ -161,6 +156,31 @@ def _update_relative_paths(template_dict, original_root, new_root):
     return template_dict
 
 
+def _transform_dict(template_dict, path, transformation, context=None):
```
mgrandis, line 159: As you import this method in another file, it shouldn't be private.
Also you should add docstrings.

samcli/commands/_utils/template.py
```diff
@@ -119,6 +117,9 @@ def _update_relative_paths(template_dict, original_root, new_root):
 
     """
 
+    def transform(original_dict, original_path, property_name, value, context):
```
mgrandis, line 120: This could be renamed to `resolve_relative_transform`

mgrandis (changes requested):
First pass.
Please resolve the conflicts.
Can you also add some integration tests please?

samcli/commands/_utils/resources.py
```diff
@@ -39,6 +40,7 @@
     AWS_SERVERLESS_LAYERVERSION: ["ContentUri"],
     AWS_GLUE_JOB: ["Command.ScriptLocation"],
     AWS_STEPFUNCTIONS_STATEMACHINE: ["DefinitionS3Location"],
+    AWS_SERVICECATALOG_PRODUCT: ["ProvisioningArtifactParameters.Info.LoadTemplateFromURL"],
```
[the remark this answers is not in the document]
mgrandis: Scratch that.

samcli/commands/_utils/template.py
rafaelliu: You mean it's imported on the unit test? I besides file it's declared on, it should only be used in `test_template.py`. Can you link where else it's being imported?

mgrandis: In `artifact_explorer.py`: https://github.com/aws/aws-sam-cli/pull/2422/files#diff-9d41d813918d5d2434b3b5397b8017579cfbe1c1f53b0f242959d6eee874346aR268

rafaelliu: nvm, lost it on the merging. found it

samcli/lib/package/artifact_exporter.py
```diff
@@ -257,65 +257,70 @@ class Resource:
...
         # folder and send the temp folder to be zipped
-        temp_dir = None
-        if is_local_file(property_value) and not is_zip_file(property_value) and self.FORCE_ZIP:
-            temp_dir = copy_to_temp_dir(property_value)
-            set_value_from_jmespath(resource_dict, self.PROPERTY_NAME, temp_dir)
-
+        temp_dir = False
+        if is_local_file(value) and not is_zip_file(value) and self.FORCE_ZIP:
+            temp_dir = True
+            value = copy_to_temp_dir(value)
```
mndeveci, line 281: Variable name `value` is too generic here. Please update regarding to its function

samcli/lib/package/artifact_exporter.py
```diff
@@ -257,65 +257,70 @@ class Resource:
...
-            resource_id, resource_dict, self.PROPERTY_NAME, parent_dir, self.uploader, artifact_extension
+            resource_id, property_value, self.PROPERTY_NAME, parent_dir, self.uploader, artifact_extension
         )
         if should_sign_package:
             uploaded_url = self.code_signer.sign_package(
                 resource_id, uploaded_url, self.uploader.get_version_of_artifact(uploaded_url)
             )
-        set_value_from_jmespath(resource_dict, self.PROPERTY_NAME, uploaded_url)
+
+        if self.URL_PROTOCOL == "http":
```
mndeveci, line 313: How about `https`?

mndeveci (changes requested):

tests/unit/lib/package/test_artifact_exporter.py
```diff
@@ -507,6 +507,39 @@ class MockResourceNoForceZip(Resource):
...
+        self, is_local_file_mock, zip_and_upload_mock, copy_to_temp_dir_mock, is_zipfile_mock, rmtree_mock
+    ):
+        class MockResourceNoForceZip(Resource):
+            PROPERTY_NAME = "foo.bar"
+
+        resource = MockResourceNoForceZip(self.s3_uploader_mock, self.code_signer_mock)
+
+        resource_id = "id"
+        original_path = "/path/to/file"
+        resource_dict = {"foo": [{"bar": original_path}]}
```
mgrandis, line 525: You could maybe add more elements in this array:
- One with only a `baz` attribute
- One with `bar` and `baz`

This would help testing that we are going through all elements and not updating the baz attribute.

samcli/commands/_utils/template.py
```diff
@@ -161,6 +156,31 @@ def _update_relative_paths(template_dict, original_root, new_root):
     return template_dict
 
 
+def _transform_dict(template_dict, path, transformation, context=None):
+    def visit(sub_dict, sub_path):
```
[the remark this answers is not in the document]
mgrandis: Not sure about this one @mndeveci as it is only doing the heavy lifting for the parent method 
Adding a comment for the parent one should be enough

samcli/commands/_utils/template.py
```diff
@@ -161,6 +156,31 @@ def _update_relative_paths(template_dict, original_root, new_root):
...
+        if not sub_path:
+            return sub_dict
+
+        next_prop = sub_path[0]
```
mgrandis, line 164: Isn't this the current property rather than the next? `property` or `prop_name` might be better names

samcli/lib/package/artifact_exporter.py
```diff
@@ -489,23 +494,23 @@ class CloudFormationStackResource(Resource):
...
-        template_path = resource_dict.get(self.PROPERTY_NAME, None)
+        template_path = property_value
 
         if (
             template_path is None
             or is_s3_url(template_path)
-            or template_path.startswith(self.uploader.s3.meta.endpoint_url)
-            or template_path.startswith("https://s3.amazonaws.com/")
+            or template_path.startswith("http://")
+            or template_path.startswith("https://")
```
mgrandis, line 510: This doesn't check anymore that we are on AWS, we may want to change this.

samcli/lib/package/artifact_exporter.py
mgrandis: This is only a way to discriminate between http and s3 (set by each resource that extends this class).

tests/unit/lib/package/test_artifact_exporter.py
mgrandis: Ok this is done in unit tests, resolving this

tests/unit/commands/_utils/test_template.py
```diff
@@ -258,3 +259,39 @@ def test_must_update_and_write_template(self, yaml_dump_mock, update_relative_pa
         yaml_dump_mock.assert_called_with(modified_template)
         m.assert_called_with(dest, "w")
         m.return_value.write.assert_called_with(dumped_yaml)
+
+
+class Test_dict_transformation(TestCase):
```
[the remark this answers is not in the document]
mgrandis: We nay be missing some tests:
- A non list/dict value (raising the exception)
- Path is not in the dict (no change)
- Transform fails (attempting to do a sum on a dict for example, we may want to update the code to handle this case)
- Dict is {}
- Dict is None

jfuss:
Cleaning up old PRs. If this is still something you want to continue @rafaelliu, feel free to re-open and handle feedback. 

To other community members, if this is something you still need we are happy to accept support for ServiceCatalog in `sam package`.

Issue #2437, opened 2020-12-02
sam deploy --guided without running build first crashes for Image based apps


sriram-mv:
### Description:

running `sam deploy --guided` without `sam build` causes a crash, as the deploy guided is looking for an non existent resource to get a valid image repository for. 

### Steps to reproduce:
sam init
sam deploy --guided

### Observed result:

```

Configuring SAM deploy
======================

	Looking for config file [samconfig.toml] :  Not found

	Setting default arguments for 'sam deploy'
	=========================================
	Stack Name [sam-app]:
	AWS Region [us-east-1]:
	Image Repository []: asdfr
Traceback (most recent call last):
  File "/Users/srirammv/workspace/1.13.1_env/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/samcli/lib/telemetry/metrics.py", line 100, in wrapped
    return func(*args, **kwargs)
  File "/Users/srirammv/workspace/1.13.1_env/lib/python3.7/site-packages/samcli/commands/deploy/command.py", line 211, in cli
    config_env,
[...]

PR #2439, opened 2020-12-03
fix(deploy): guided deploy

sriram-mv:
- if attempting to deploy without build, sam cli does not crash
- stricter validation on `--image-repository` to always be an ECR URI
- if an ECR image repository is not provided not provided on guided, SAM
CLI does not crash.

* https://github.com/aws/aws-sam-cli/issues/2437

* Prevent SAM CLI crashes.

* Add type validation to `--image-repository` option type.
* image repository can only be valid ECR URI
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/regexpr.py
```diff
@@ -0,0 +1,4 @@
+"""
+Regular Expressions for Resources.
+"""
+ECR_URL = "(^[a-zA-Z0-9][a-zA-Z0-9-_]*).dkr.ecr.([a-zA-Z0-9][a-zA-Z0-9-_]*).amazonaws.com(.cn)?/.*"
```
mgrandis, line 4: Should the dots be escaped like this "\."? Otherwise you are matching any char.

sriram-mv: That makes sense!
```
"(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?\/.*"
```

is what i tested now and that seems to be more exact.

samcli/commands/deploy/guided_context.py
```diff
@@ -270,13 +271,18 @@ def prompt_image_repository(self, parameter_overrides):
                 type=click.STRING,
                 default=self.image_repository if self.image_repository else "",
             )
+            if not image_repository or not is_ecr_url(image_repository):
```
mgrandis, line 274: I believe the `not image_repository` check is already done inside `is_ecr_url`.

tests/unit/cli/test_types.py
```diff
@@ -285,3 +287,32 @@ def test_must_fail_on_invalid_format(self, input):
...
+
+class TestImageRepositoryType(TestCase):
+    def setUp(self):
+        self.param_type = ImageRepositoryType()
+        self.mock_param = Mock(opts=["--image-repository"])
+
+    @parameterized.expand(
+        [
+            # Just a string
+            ("some string"),
```
mgrandis, line 300: You could maybe add more failing tests, with strings missing key parts of the pattern, like `123456789012.us-east-1.amazonaws.com/test1`

tests/unit/cli/test_types.py
```diff
@@ -285,3 +287,32 @@ def test_must_fail_on_invalid_format(self, input):
...
+    def test_must_fail_on_invalid_format(self, input):
+        self.param_type.fail = Mock()
+        with self.assertRaises(BadParameter):
+            self.param_type.convert(input, self.mock_param, Mock())
+
+    @parameterized.expand(
+        [
+            (
+                "123456789012.dkr.ecr.us-east-1.amazonaws.com/test1",
+                "123456789012.dkr.ecr.us-east-1.amazonaws.com/test1",
```
mgrandis, line 312: you can add a test for a `.cn` url: `123456789012.dkr.ecr.us-east-1.amazonaws.com.cn/test1`

samcli/commands/deploy/guided_context.py
```diff
@@ -270,13 +271,18 @@ def prompt_image_repository(self, parameter_overrides):
...
+
             for _, function_prop in self.transformed_resources.functions.items():
                 if function_prop.packagetype == IMAGE:
                     image = function_prop.imageuri
                     try:
                         tag = tag_translation(image)
                     except NonLocalImageException:
                         pass
+                    except NoImageFoundException as ex:
+                        raise GuidedDeployFailedError("No images found to deploy, try sam build") from ex
```
mgrandis, line 285: I would add `running`: `try running sam build`

mgrandis (approved):

samcli/lib/package/regexpr.py
```diff
@@ -0,0 +1,4 @@
+"""
+Regular Expressions for Resources.
+"""
+ECR_URL = r"(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?\/.*"
```
hoffa, line 4: Just curious, is this documented anywhere? Or is it possible this will break some day?

hoffa (approved):

samcli/lib/package/regexpr.py
sriram-mv: what we could document here is the docs section based on which the regex is derived.

https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_Repository.html does showcase the pattern for the repository name but not the entire regex.

What we can do is add a comment to link this above and also change the regex to allow for checking on the repository name as well.

Issue #2432, opened 2020-12-02
error : 'dict' object has no attribute 'upload_with_dedup


danielnassab:
### Description:
Not able to deploy using sam deploy with paramaters

### Steps to reproduce:
          python -m pip install --upgrade pip
          pip install aws-sam-cli==1.13.1
          sam deploy \
        --parameter-overrides $parameters \
        --capabilities CAPABILITY_IAM \
        --stack-name ${SERVICE} \
        --s3-bucket $DEPLOYMENT_BUCKET \
        --s3-prefix  ${SERVICE} \
        --no-fail-on-empty-changeset

### Observed result:
error stacktrace

include_transform_export_handler
    template_dict["Parameters"]["Location"] = uploader.upload_with_dedup(abs_include_location)
AttributeError: 'dict' object has no attribute 'upload_with_dedup'

### Expected result:
downgrading to 1.12.0 works, expecting it to run as i dont see any changes to config or deploy command. 

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: ubuntu 
2. `sam --version`: 1.13.1

`Add --debug flag to command you are running`

afittz:
I'm also seeing this when run sam package. 

`
sam package --output-template-file .\packaged.yaml --s3-bucket <MY_BUCKET> --profile default --debug
2020-12-02 11:31:15,179 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2020-12-02 11:31:15,179 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2020-12-02 11:31:15,180 | Sending Telemetry: {'metrics': [{'templateWarning': {'awsProfileProvided': True, 'debugFlagProvided': True, 'region': '', 'warningName': 'CodeDeployWarning', 'warningCount': 0, 'requestId': 'a411e19e-1e67-4f82-a720-1b166a0946c5', 'installationId': 'd9024855-b3ba-41ab-9623-e6758d79650f', 'sessionId': '3f26609f-c9f4-428f-80ee-5604f3e5150e', 'executionEnvironment': 'CLI', 'pyversion': '3.9.0', 'samcliVersion': '1.13.1'}}]}
2020-12-02 11:31:15,603 | Telemetry response: 200
[...]

PR #2440, opened 2020-12-03
fix(package): regression with "Fn::Transform"

sriram-mv:
- default to s3 based uploader
- added integration test for Fn::Transform which was missing before.

https://github.com/aws/aws-sam-cli/issues/2432

* AWS Include defaults to s3 as the export mechanism. there was a bug in not supplying the correct uploader.

* Specified AWS Include macro now packages to s3
* Its a regression fix
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -151,7 +151,9 @@ class Template:
         """
         for key, val in template_dict.items():
             if key in GLOBAL_EXPORT_DICT:
-                template_dict[key] = GLOBAL_EXPORT_DICT[key](val, self.uploader, self.template_dir)
+                template_dict[key] = GLOBAL_EXPORT_DICT[key](
+                    val, self.uploader.get(ResourceZip.EXPORT_DESTINATION), self.template_dir
+                )
             elif isinstance(val, dict):
                 self.export_global_artifacts(val)
             elif isinstance(val, list):
--- a/tests/integration/package/test_package_command_zip.py
+++ b/tests/integration/package/test_package_command_zip.py
@@ -56,6 +56,7 @@ class TestPackageZip(PackageIntegBase):
             "aws-serverlessrepo-application.yaml",
             "aws-serverless-statemachine.yaml",
             "aws-stepfunctions-statemachine.yaml",
+            "aws-include-transform.yaml",
         ]
     )
     def test_package_barebones(self, template_file):
--- a/tests/integration/testdata/package/aws-include-transform.yaml
+++ b/tests/integration/testdata/package/aws-include-transform.yaml
@@ -1,0 +1,9 @@
+AWSTemplateFormatVersion : '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: Transform only app
+
+Resources:
+  'Fn::Transform':
+    Name: 'AWS::Include'
+    Parameters:
+      Location : ./main.py


BR0kEN-:
@sriram-mv here is another scenario where `self.uploader` wrongly used as a `dict` - https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/package/artifact_exporter.py#L72

BR0kEN-:
I cannot share the template but can confirm the proposed patch works. Checked on CodeBuild by patching the file:

```
cp deploy/artifact_exporter.py /root/.pyenv/versions/3.8.3/lib/python3.8/site-packages/samcli/lib/package/artifact_exporter.py
```

sriram-mv:
@BR0kEN- thanks for the catch, looking into it.

samcli/lib/package/artifact_exporter.py
```diff
@@ -89,11 +89,13 @@ def do_export(self, resource_id, resource_dict, parent_dir):
             temporary_file.write(exported_template_str)
             temporary_file.flush()
 
-            url = self.uploader.upload_with_dedup(temporary_file.name, "template")
+            url = self.uploader.get(self.EXPORT_DESTINATION).upload_with_dedup(temporary_file.name, "template")
```
BR0kEN-, line 92: I might be wrong but `self.uploader` in the `Resource` is not a dict but a specific uploader.

CoshUS (approved):

samcli/lib/package/artifact_exporter.py
```diff
@@ -69,7 +69,7 @@ def do_export(self, resource_id, resource_dict, parent_dir):
         if (
             template_path is None
             or is_s3_url(template_path)
-            or template_path.startswith(self.uploader.s3.meta.endpoint_url)
+            or template_path.startswith(self.uploader.get(self.EXPORT_DESTINATION).s3.meta.endpoint_url)
```
BR0kEN-, line 72: I think this must be exactly `self.uploader.s3` because `self.uploader` here is an instance of `S3Uploader` and `s3` is its property.

samcli/lib/package/artifact_exporter.py
sriram-mv: Thanks for the catch, updated

samcli/lib/package/artifact_exporter.py
```diff
@@ -89,11 +89,13 @@ def do_export(self, resource_id, resource_dict, parent_dir):
             temporary_file.write(exported_template_str)
             temporary_file.flush()
 
-            url = self.uploader.upload_with_dedup(temporary_file.name, "template")
+            url = self.uploader.get(self.EXPORT_DESTINATION).upload_with_dedup(temporary_file.name, "template")
 
             # TemplateUrl property requires S3 URL to be in path-style format
             parts = parse_s3_url(url, version_property="Version")
-            s3_path_url = self.uploader.to_path_style_s3_url(parts["Key"], parts.get("Version", None))
+            s3_path_url = self.uploader.get(self.EXPORT_DESTINATION).to_path_style_s3_url(
```
BR0kEN-, line 96: This needs to be reverted too.

BR0kEN-: Sorry, I brought the confusion by https://github.com/aws/aws-sam-cli/pull/2440#issuecomment-737715257. The https://github.com/aws/aws-sam-cli/pull/2440/commits/0214f4b440de77785fe5723e0ec866525f6653dc seems all we have to change.

mndeveci (approved):

samcli/lib/package/artifact_exporter.py
sriram-mv: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/package/s3_uploader.py#L57

So the uploader within the dict which is assigned to s3 is S3Uploader which has the s3_client stored as a `s3` member.

BR0kEN-: `self.uploader` is not a `dict` but `S3Uploader`, so I think this change needs to be reverted. The tests change are also wrong.

sriram-mv: yeah on further testing and running integration tests, seems like https://github.com/aws/aws-sam-cli/commit/0214f4b440de77785fe5723e0ec866525f6653dc is all we need.

samcli/lib/package/artifact_exporter.py
sriram-mv: Yep, on further testing, only have https://github.com/aws/aws-sam-cli/commit/0214f4b440de77785fe5723e0ec866525f6653dc in place now.

PR #2441, opened 2020-12-03
Fix schema loading when no default profile

kylelaker:
The region_name value when a profile does not specify a region or when
there is not a 'default' profile resulted in a crash since
_get_aws_region_choice attempted to perform concatenation using the `+`
operator (which is undefined for str + NoneType). Using an actual string
formatting method fixes that issue.

Additionally, it is useless to prompt the user whether not to use the
'default' profile since doing so is impossible. They need to be
prompted. Additionally, it is slightly more 'intelligent' to create a
new session after the user has chosen a new profile so that a better
default region can be chosen.

No issue was opened.

The `sam init` crashes if a user needs to pick a schema and they do not have a `default` profile or their `default` profile does not specify a `region`. (`TypeError` in `_get_aws_region_choice`)

Avoids performing a `+` operation to perform string concatenation.

Does not prompt the user to accept the defaults if doing so results in `region` being None. Just moves on to asking the user. 

#### Checklist

- [x] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


moelasmar (changes requested):
Can you add unit/integration test cases that cover this change

kylelaker:
@moelasmar: Unit test coverage for the new code path(s) has been added

CoshUS (approved):

moelasmar (approved):

Issue #2433, opened 2020-12-02
sam deploy --guided disregards region input


smirnoal:
### Description:

after execution of `sam deploy --guided` it asked me about the region to deploy lambda function to. I entered "us-east-2". A few steps later it printed out a huge python stack trace with message "You must specify region".

It affects customer experience as the tool disregards information entered by a user - which does not look holistic.

### Steps to reproduce:

- Run `aws configure`. Enter credentials. Do not put region information.
- Create a sample python lambda function from "Hello word" template using `sam init`
- Try to deploy it with `sam deploy --guided`. It will ask for a region. Enter "us-east-2"
- Follow further on-screen instructions

### Observed result:
```
% sam deploy --guided

Configuring SAM deploy
======================

	Looking for config file [samconfig.toml] :  Not found

	Setting default arguments for 'sam deploy'
	=========================================
	Stack Name [sam-app]: 
	AWS Region [us-east-1]: us-east-2
	#Shows you resources changes to be deployed and require a 'Y' to initiate deploy
	Confirm changes before deploy [y/N]: y
	#SAM needs permission to be able to create roles to connect to the resources in your template
	Allow SAM CLI IAM role creation [Y/n]: y
	HelloWorldFunction may not have authorization defined, Is this okay? [y/N]: y
	Save arguments to configuration file [Y/n]: y
	SAM configuration file [samconfig.toml]: 
	SAM configuration environment [default]: 

	Looking for resources needed for deployment: Not found.
	Creating the required resources...
	Successfully created!

		Managed S3 bucket: aws-sam-cli-managed-default-samclisourcebucket-1lljvk7s3ie0m
		A different default S3 bucket can be set in samconfig.toml

	Saved arguments to config file
	Running 'sam deploy' for future deployments will use the parameters saved above.
	The above parameters can be changed by modifying samconfig.toml
	Learn more about samconfig.toml syntax at 
	https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-config.html
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 33, in <module>
    sys.exit(load_entry_point('aws-sam-cli==1.12.0', 'console_scripts', 'sam')())
  File "/usr/local/Cellar/aws-sam-cli/1.12.0/libexec/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
[...]

PR #2455, opened 2020-12-04
fix: create signer client with region definition

mndeveci:
https://github.com/aws/aws-sam-cli/issues/2433

If user doesn't have any default region setting (env variable or in ~/.aws folder) guided deploy command will fail.

This change sets region from guided deploy parameters

None

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


prenx4x (approved):

jfuss (approved):
LGTM. Are there any tests we should be added for this?

mndeveci:
> LGTM. Are there any tests we should be added for this?

Good point. Added some unit tests for client creation of `s3`, `ecr` and `signer`.

PR #2458, opened 2020-12-05
Set operationName in local API request context

### LINKED ISSUE #2457: `operationName` missing from `sam local start-api` API Gateway proxy request context
### Description:

When running a Lambda behind an API Gateway REST API, each request context includes a field called `operationName`. This enables a single Lambda to handle multiple API routes:
https://github.com/aws/aws-lambda-go/blob/5bc50ab1c1dc8f32fce7f7745efcf601f8054fbe/events/apigw.go#L35

Unfortunately, this field isn't being set by `sam local start-api`, which prevents testing Lambdas that depend on `operationName`. The same Lambdas work fine on production AWS.

For clarity, both this issue and #2456 are about missing fields that AWS provides but that `sam local start-api` does not. However, I wasn't able to find any version of the SAM CLI that sets `operationName`, so this one doesn't look like a regression.

### Steps to reproduce:

Check the API request context for the `operationName` field.

### Observed result:

The `operationName` is omitted from the request context JSON blob.

### Expected result:

The `operationName` should contain the API route's operation name (e.g. `ConnectRoute` in the case of https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/tests/smoke/templates/sar/simple-websockets-chat-app-template.yaml#L29)

1. OS: Mac OS 10.15.7
2. `sam --version`: v1.13.2

ramosbugs:

Fixes #2457.

The `operationName` field is currently missing from the API Gateway proxy request context, which prevents certain Lambda functions from working correctly in the local environment.

This PR adds the missing `operationName` based on the Swagger `operationId`, as the production AWS Gateway does.

Hopefully none :-)

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


ramosbugs:
Oh shoot, this doesn't quite fix the issue because of the route deduping done here: https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/lib/providers/api_collector.py#L104

The deduping combines all of the routes that share the same path (but different methods) into a single `Route` object:
https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/lib/providers/api_collector.py#L152-L166

However, they should be treated as separate routes if they have different operation names. Given this, should we still be deduping the routes at all? I can't quite tell from the comments what the original rationale was. It looks like this was introduced in #1239. @viksrivat: do you happen to remember?

jfuss:
@ramosbugs SAM CLI uses Flask under the hood to create an API Gateway like endpoint. We need to dedup the urls and methods to [wire it up with Flask](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/apigw/local_apigw_service.py#L157).

ramosbugs:
> @ramosbugs SAM CLI uses Flask under the hood to create an API Gateway like endpoint. We need to dedup the urls and methods to [wire it up with Flask](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/apigw/local_apigw_service.py#L157).

@jfuss: thanks for the quick reply! the call site you linked to is passing the applicable HTTP method(s) to Flask as an argument: https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/local/apigw/local_apigw_service.py#L161

The [`add_url_rule` docs](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.add_url_rule) mention that the kwargs are passed to [Werkzeug `Rule`](https://werkzeug.palletsprojects.com/en/1.0.x/routing/#werkzeug.routing.Rule), which documents a `methods` argument for specifying which HTTP methods should be handled by the rule. My reading of this is that Flask supports different route handlers for different HTTP methods, which is true of every HTTP server framework I've used. Independent of SAM, it would be _very_ limiting if every REST API implemented in Flask had to use the same function to handle all HTTP methods for a given path.

Locally, I replaced this line with `routes = self.routes`, and `sam local start-api` properly routes to the correct Lambda depending on the HTTP method (with the same URL path for each): https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/lib/providers/api_collector.py#L104

Consequently, it seems safe to remove the dedupe functionality, which may have been based on a misunderstanding of Flask's behavior, or some limitation that used to exist but no longer does.

ramosbugs:
or maybe the issue is this: https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/lib/providers/api_collector.py#L141

the line above says that it should dedupe based on `the same function_name and method`, but what it actually does is dedupe based on `function_name` and `path`: https://github.com/aws/aws-sam-cli/blob/3ee79db8a83b7941482fe03315d6162483094e9d/samcli/lib/providers/api_collector.py#L153

ramosbugs:
@jfuss: any thoughts on my comments above?

iiuc, the current SAM CLI isn't capable of testing most REST APIs, which frequently use multiple HTTP methods (GET/POST/DELETE, etc.) for a single route to represent different operations (i.e., create the resource, fetch the resource, delete the resource). Flask definitely supports this, but not the way that the SAM CLI is currently leveraging Flask...

I'm happy to help contribute a fix if we can align on the approach :-)

ramosbugs:
See https://github.com/ramosbugs/aws-sam-rest-api-repro for an example repro of the route dedupe bug described above (on the latest version of this PR)

jfuss:
@ramosbugs Thanks for providing the repro app. It is super helpful. I need to do some deeper testing on what is going on here, as API Gateway doesn't document this (from what I can tell).

What I am going to try and answer:

1. Is operationid only supported in the payload version 1? Looking at https://github.com/aws/aws-lambda-go/blob/5bc50ab1c1dc8f32fce7f7745efcf601f8054fbe/events/apigw.go#L35 (which you linked in the issue) is only for v1. The v2 format doesn't seem to have operationid, at least in the aws-lambda-go library. 
2. operationid is part of the OpenApi Path object: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#operationObject. I understand how this is supported in the "get", "post", etc verbs but not sure on the x- extension. This is important for how this works with `x-amazon-apigateway-any-method`. 

From your example, it is clear how this will work in the non swagger model but we need to make sure we support both. I would like to get integrations tests added for various combinations, without that we will risk missing or breaking this. I am happy to add them myself but need to dig deeper and understand this all, especially because it is not documented behavior of API Gateway.

ramosbugs:
thank you @jfuss! great points.

once implemented, I think this should simplify the internal routing logic that Lambda functions need to implement: based on unique operation ID rather than (HTTP method, route) pair, which partially duplicates the routing logic API Gateway does for us.

jfuss:
@ramosbugs I am not sure I get your point on: "I think this should simplify the internal routing logic that Lambda functions need to implement: based on unique operation ID rather than (HTTP method, route) pair, which partially duplicates the routing logic API Gateway does for us."

My understanding of API Gateway is that the routing done is by HTTP method and route. Operation ID is optional, so I don't quite understand how this improves any routing logic in SAM CLI.

jfuss:
The OperationName is only sent to the Lambda Function from API Gateway V1 (Rest API). Rest API's only support payload version 1.0. However for Http Apis (v2), API Gateway never sends the OperationName (payload version 1 or 2).

With this change, we need to reflect the same behavior as API Gateway. @ramosbugs Are you willing to make changes to this PR or should I take it over completely?

ramosbugs:
> @ramosbugs I am not sure I get your point on: "I think this should simplify the internal routing logic that Lambda functions need to implement: based on unique operation ID rather than (HTTP method, route) pair, which partially duplicates the routing logic API Gateway does for us."
> 
> My understanding of API Gateway is that the routing done is by HTTP method and route. Operation ID is optional, so I don't quite understand how this improves any routing logic in SAM CLI.

it's a somewhat minor point, but if you have a single Lambda function that's serving multiple API endpoints (particularly useful for compiled languages to avoid emitting dozens of binaries), that Lambda function has to know which endpoint handler to invoke internally. if the operation name/ID from the OpenAPI spec is present in the request, that provides a convenient input for making that routing decision. it also means that only the API Gateway even needs to be aware of the URL paths (e.g., `/user/{user_id}`) for each endpoint (to translate it to an operation ID), and then the Lambda only needs to reason in terms of operation IDs (e.g., `getUser`).

> The OperationName is only sent to the Lambda Function from API Gateway V1 (Rest API). Rest API's only support payload version 1.0. However for Http Apis (v2), API Gateway never sends the OperationName (payload version 1 or 2).
> 
> With this change, we need to reflect the same behavior as API Gateway. @ramosbugs Are you willing to make changes to this PR or should I take it over completely?

it's been awhile since I've touched this code and you're much more familiar with this codebase than I am, so please feel free to take it over :-)

ramosbugs:
> @ramosbugs I have raised a PR against your PR branch to add the integration tests, and only let operationname send to APIGW v1. Can you please merge it and get it into this PR? [ramosbugs#1](https://github.com/ramosbugs/aws-sam-cli/pull/1)

merged, thank you!

wchengru (approved):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -319,8 +324,20 @@ def _request_handler(self, **kwargs):
...
+                else:
+                    operation_name = None
                 event = self._construct_v_1_0_event(
-                    request, self.port, self.api.binary_media_types, self.api.stage_name, self.api.stage_variables
+                    request,
+                    self.port,
+                    self.api.binary_media_types,
+                    self.api.stage_name,
+                    self.api.stage_variables,
+                    operation_name,
```
jfuss, line 340: We can change this later since I don't think we have push access to remote. Keep nesting like this are very challenging to maintain. Instead we should be doing "if some condition: do a thing", "if other thing: do this thing". 

So a much more maintainable way is:

```
# in all other cases but the if's below
create event

if event_type is `Route.API`: 
  create event

if event_type is Route.HTTP:
   create event

if event_type is Route.Http and some condition:
  create event
```

Its much easier to read, as you expand you don't have a magical path that defaults to something (sometimes that is better but can lead to unexpected bugs).

samcli/local/events/api_event.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         domain_name=None,
         request_time_epoch=int(time()),
         request_time=datetime.utcnow().strftime("%d/%b/%Y:%H:%M:%S +0000"),
+        operation_name=None,
```
jfuss, line 85: Is this used for all API Events? If so, shouldn't we model different RequestsContext's for different events?

tests/integration/local/start_api/test_start_api.py
```diff
@@ -664,6 +664,30 @@ def test_patch_call_with_path_setup_with_any_swagger(self):
...
+    @pytest.mark.flaky(reruns=3)
+    @pytest.mark.timeout(timeout=600, method="thread")
+    def test_http_api_payload_v2_should_not_have_operation_id(self):
+        response = requests.get(self.url + "/httpapi-operation-id-v2", timeout=300)
+        self.assertEqual(response.status_code, 200)
+
+        response_data = response.json()
+        self.assertEqual(response_data.get("version", {}), "2.0")
+        # operationName or operationId shouldn't be processed by Httpapi swaggers
+        self.assertIsNone(response_data.get("requestContext", {}).get("operationName"))
```
jfuss, line 688: How do you know `operationName` does not exist vs `operationName` is in the dict but with no value (`None`)?

jfuss (changes requested):

tests/integration/local/start_api/test_start_api.py
wchengru: Good catch! I updated the tests here and other places.

samcli/local/apigw/local_apigw_service.py
wchengru: Thanks for the suggestion. Updated.

jfuss (approved):
LGTM

ramosbugs:
thank you all!

PR #2463, opened 2020-12-08
Add .NET 5 image template configuration

ganeshnj:
- Adds ability to create .NET 5 image projects.

Successfully deployed the new project using below commands
```
python -m samcli init
python -m samcli build
python -m samcli local invoke
python -m samcli deploy --guided
```

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/common/runtime_template.py
```diff
@@ -38,7 +38,7 @@
     ],
     "dotnet": [
         {
-            "runtimes": ["dotnetcore3.1", "dotnetcore2.1"],
+            "runtimes": ["dotnet5.0", "dotnetcore3.1", "dotnetcore2.1"],
```
sriram-mv, line 41: you should not need this as these for zip based runtimes, can you confirm this is needed?

sriram-mv (changes requested):

samcli/local/common/runtime_template.py
ganeshnj: Yes this is required else template clone step fails with
```
Error: Lambda Runtime dotnet5.0 and dependency manager cli-package does not have an available initialization template.
```

samcli/local/docker/lambda_debug_settings.py
```diff
@@ -81,6 +81,10 @@ def get_debug_settings(debug_port, debug_args_list, _container_env_vars, runtime
                 entry + ["/var/runtime/bootstrap"] + debug_args_list,
                 container_env_vars={"_AWS_LAMBDA_DOTNET_DEBUGGING": "1", **_container_env_vars},
             ),
+            Runtime.dotnet5.value: DebugSettings(
+                entry + ["/var/runtime/bootstrap"] + debug_args_list,
+                container_env_vars={"_AWS_LAMBDA_DOTNET_DEBUGGING": "1", **_container_env_vars},
+            ),
```
mndeveci, line 87: I am not sure but can you confirm that we support debugging for image option as well (not just for zip)?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -539,7 +541,13 @@ def test_with_dotnetcore(self, runtime, code_uri, mode):
 
         self.verify_docker_container_cleanedup(runtime)
 
-    @parameterized.expand([("dotnetcore2.1", "Dotnetcore2.1"), ("dotnetcore3.1", "Dotnetcore3.1")])
+    @parameterized.expand(
+        [
+            ("dotnetcore2.1", "Dotnetcore2.1"),
+            ("dotnetcore3.1", "Dotnetcore3.1"),
+            ("dotnet5.0", "Dotnet5.0"),
```
mndeveci, line 548: I believe these integration tests are used for ZIP type building and they are failing in PR checks. I will check if there are image build tests as well.

mndeveci (changes requested):

samcli/local/common/runtime_template.py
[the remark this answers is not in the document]
ganeshnj: There is branding change and from .NET 5,[ Microsoft has dropped](https://docs.microsoft.com/en-us/dotnet/standard/frameworks) core from the target framework name.

samcli/local/docker/lambda_debug_settings.py
ganeshnj: I think yes, Dockerfile has [SAM_BUILD_MODE](https://github.com/aws/aws-sam-cli-app-templates/blob/master/dotnetcore3.1-image/cookiecutter-aws-sam-hello-dotnet-lambda-image/%7B%7Bcookiecutter.project_name%7D%7D/src/HelloWorld/Dockerfile#L16) which in my opinion used for debugging.

tests/integration/buildcmd/test_build_cmd.py
ganeshnj: Sure, let me know if any additional change required for tests.

samcli/local/docker/lambda_debug_settings.py
mndeveci: I see but were you able to start a debugging session by running `sam local invoke -d {debugport}` and see the execution have been halted?

samcli/local/common/runtime_template.py
jfuss: We should figure this out deeper. As Sriram said, this is for zips only. There could be something tied together that shouldn't be since this is the first language that we do not have both a zip and image package type for

samcli/local/common/runtime_template.py
```diff
@@ -86,6 +86,7 @@ def get_local_lambda_images_location(mapping, runtime):
     "ruby2.7": ["bundler"],
     "nodejs12.x": ["npm"],
     "nodejs10.x": ["npm"],
+    "dotnet5.0": ["cli-package"],
```
jfuss, line 89: Not needed.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -485,8 +485,10 @@ class TestBuildCommand_Dotnet_cli_package(BuildIntegBase):
         [
             ("dotnetcore2.1", "Dotnetcore2.1", None),
             ("dotnetcore3.1", "Dotnetcore3.1", None),
+            ("dotnet5.0", "Dotnet5.0", None),
```
jfuss, line 488: Not needed. These test are for zip package types.

samcli/local/docker/lambda_debug_settings.py
jfuss: This is only used for zips if I recall correctly. This shouldn't be needed at all.

jfuss (changes requested):

samcli/local/common/runtime_template.py
ganeshnj: I think, this is tied to L41 above. Both changes are required to `sam init` the template.

samcli/local/docker/lambda_debug_settings.py
ganeshnj: Without it, `make pr` fails 
```
E           samcli.local.docker.lambda_debug_settings.DebuggingNotSupported: Debugging is not currently supported for dotnet5.0

samcli/local/docker/lambda_debug_settings.py:137: DebuggingNotSupported
```

tests/integration/buildcmd/test_build_cmd.py
ganeshnj: Fixed.

samcli/local/common/runtime_template.py
ganeshnj: What's the next step here for this PR?

samcli/local/docker/lambda_image.py
```diff
@@ -37,6 +37,7 @@ class Runtime(Enum):
     go1x = "go1.x"
     dotnetcore21 = "dotnetcore2.1"
     dotnetcore31 = "dotnetcore3.1"
+    dotnet5 = "dotnet5.0"
```
jfuss, line 40: Is this needed for something? 

Off the top of my head, this enum is heavily used for Zips but not for images (or wouldn't expect it to). This is probably fine but not sure what this may affect.

jfuss (approved):

samcli/lib/init/__init__.py
```diff
@@ -12,12 +12,20 @@
 from samcli.local.common.runtime_template import RUNTIME_DEP_TEMPLATE_MAPPING
 from .exceptions import GenerateProjectFailedError
 from .arbitrary_project import generate_non_cookiecutter_project
+from ..utils.packagetype import IMAGE
```
sriram-mv, line 15: I dont see this convention elsewhere, can we use absolute import?

samcli/lib/init/__init__.py
```diff
@@ -51,7 +59,7 @@ def generate_project(
 
     template = None
 
-    if runtime:
+    if runtime and package_type != IMAGE:
```
sriram-mv, line 62: == ZIP may be more explicit.

samcli/local/docker/lambda_image.py
sriram-mv: I think this maybe because and these are used keys into the manifest.json from app templates. we need to decouple this.

sriram-mv (approved):
approving, but have some comments. could you please look into them?

samcli/local/docker/lambda_image.py
mndeveci: You are right, this is not needed anymore, tested with guided `init` and non-interactive one(`samdev init --no-interactive --name dotnet5-18-1738 --package-type Image --base-image amazon/dotnet5.0-base`) I was able to initiate the project.

mndeveci (approved):

Issue #2447, opened 2020-12-03
Sam Package/Deploy --image-repository Behavior


metaskills:
Many of the process I put in place for both open source and in company deploy pipelines take advantage of SAM CLI and the AWS CLI using conventions like AWS_PROFILE. I've been very happy that SAM CLI has followed these patterns. Today when working with the new container features I was surprised by this odd behavior of [sam package](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-package.html) when using the `--image-repository` option. Here is an example of my usage where the new image repo was added to my process.

```shell
sam package \
  --region ${AWS_DEFAULT_REGION} \
  --template-file ./.aws-sam/build/template.yaml \
  --output-template-file ./.aws-sam/build/packaged.yaml \
  --image-repository  "lambyc-starter" \
  --s3-bucket "${CLOUDFORMATION_BUCKET}" \
  --s3-prefix "lambyc-starter-${RAILS_ENV}"
```

These commands are run as either the default AWS_PROFILE or with specific ENV overrides. Given this was set and that the `--region` was set here, my expectation was this command was going to find and publish to the ECR repo within my AWS account. Instead, it tried to push to docker.io and failed with a user password. Digging into some guides and published SAM examples I can see what you expect folks to do is:

```shell
sam package \
  --region ${AWS_DEFAULT_REGION} \
  --template-file ./.aws-sam/build/template.yaml \
  --output-template-file ./.aws-sam/build/packaged.yaml \
  --image-repository  "123456789.dkr.ecr.us-east-1.amazonaws.com/lambyc-starter" \
  --s3-bucket "${CLOUDFORMATION_BUCKET}" \
  --s3-prefix "lambyc-starter-${RAILS_ENV}"
```

This feels like the wrong interface to me and against the grain of how the CLI operates given all my previous experiences. I can work around this if y'all disagree by adding more `aws` CLI commands to find the account ID and use the `AWS_DEFAULT_REGION` env and/or look that up as well. But it would cool if SAM did this. Thoughts?

sriram-mv:
Thanks for reporting this!

For the issue it trying to upload to dockerhub: there is a fix out for that: https://github.com/aws/aws-sam-cli/pull/2439

One of the things that makes it harder to figure out the full URI with just a name is to construct the neccessary accountid, the way to do that would sts get-caller-identity (or what if you wanted to do cross account), which is additional permissions for users.

[...]

PR #2465, opened 2020-12-08
feat(package/deploy): new option `--image-repositories`

sriram-mv:

https://github.com/aws/aws-sam-cli/issues/2447 
It does not fix this issue, but allows for mapping of function to different ECR if required.

Allow for function to ecr mapping. each function can now be uploaded to a different ECR.

Guided deploy prompts for an ECR per function and saves it.

if both `--image-repository` and `--image-repositories` are specifed. One has to be removed. Guided flow looks to fill `--image-repositories` but is not destructive to the configuration file and does not remove `--image-repository` if its already set.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

NOTE: WIP and needs more tests.


samcli/commands/_utils/options.py
```diff
@@ -9,10 +9,16 @@
 import click
 from click.types import FuncParamType
 
-from samcli.commands._utils.template import get_template_data, TemplateNotFoundException
+from samcli.commands._utils.option_validator import Validator
```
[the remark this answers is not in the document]
sriram-mv: ah, I was hoping that `make black` takes care of this, apparently it does not.

samcli/cli/types.py
```diff
@@ -388,5 +388,26 @@ def convert(self, value, param, ctx):
...
+        if len(key_value_pair) != 2:
+            raise click.BadParameter(
+                f"{param.opts[0]} is not a valid format, it needs to be of the form function_logical_id=ECR_URI"
+            )
+        else:
+            key = key_value_pair[0]
+            _value = key_value_pair[1]
+            if not is_ecr_url(_value):
+                raise click.BadParameter(f"{param.opts[0]} needs to have valid ECR URI as value")
+            return {key: _value}
```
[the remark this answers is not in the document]
sriram-mv: this is a specific way of click operates with multi value options, we can do the other way but its not native click. ( this is how --parameter-overrides work).

I still prefer one, because its native impl and better representation in configuration file.

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -0,0 +1,60 @@
...
+                exception=click.BadOptionUsage(
+                    option_name="--image-repositories",
+                    ctx=ctx,
+                    message=f"Incomplete list of function logical ids specified for '--image-repositories'",
+                ),
+            ),
+        ]
+        for validator in validators:
+            validator.validate()
+        func(*args, **kwargs)
```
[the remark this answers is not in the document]
sriram-mv: +1

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -0,0 +1,60 @@
+import click
+
+from samcli.commands._utils.option_validator import Validator
+from samcli.commands._utils.template import get_template_function_resource_ids, get_template_artifacts_format
+from samcli.lib.utils.packagetype import IMAGE
+
+
+def image_repository_validation(func):
```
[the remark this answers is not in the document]
sriram-mv: Definitely. will add. Good callout.

samcli/commands/deploy/guided_config.py
```diff
@@ -93,5 +100,10 @@ def _save_parameter_overrides(self, cmd_names, config_env, parameter_overrides,
             if _params:
                 samconfig.put(cmd_names, self.section, "parameter_overrides", " ".join(_params), env=config_env)
 
+    def _save_image_repositories(self, cmd_names, config_env, samconfig, image_repositories):
+        if image_repositories:
+            _image_repositories = [f"{key}={value}" for key, value in image_repositories.items()]
+            samconfig.put(cmd_names, self.section, "image_repositories", _image_repositories, env=config_env)
```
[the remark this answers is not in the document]
sriram-mv: But I need to save it in a certain format, thats why its this separate.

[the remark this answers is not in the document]
mndeveci: AFAIR, if you just pass the arguments in `**kwargs` this will automatically add subsection and populate items under there. So we don't need special handling here.

The reason why we have this with the others, is because of the escaping of these arguments, which we won't be doing for this.

samcli/commands/_utils/options.py
```diff
@@ -94,6 +94,22 @@ def guided_deploy_stack_name(ctx, param, provided_value):
...
+    :param param: Param name
+    :param provided_value: Value provided by Click, after being processed by ImageRepositoriesType.
+    :return: dictionary of function logic ids to ECR URIs.
+    """
+
+    image_repositories = {}
+    for value in provided_value:
+        image_repositories.update(value)
+
+    return image_repositories if image_repositories else None
```
mndeveci, line 110: Why do we returning `None` instead of empty dict? Asking because this might cause errors down the execution where we call operations on the dict.

samcli/commands/_utils/template.py
```diff
@@ -261,3 +263,25 @@ def get_template_artifacts_format(template_file):
...
+        AWS::Serverless::Function
+    :param template_file: template file location.
+    :param artifact: artifact of type IMAGE or ZIP
+    :return: list of artifact formats
+    """
+
+    template_dict = get_template_data(template_file=template_file)
+    _function_resource_ids = []
+    for resource_id, resource in template_dict.get("Resources", {}).items():
+        if resource.get("Properties", {}).get("PackageType", ZIP) == artifact and resource.get("Type") in [
```
mndeveci, line 282: There was an existing function collector, which is been used in build command, can we use that one instead?

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -0,0 +1,78 @@
...
+                    option_name="--image-repositories",
+                    ctx=ctx,
+                    message="Missing option '--image-repository' or '--image-repositories'",
+                ),
+            ),
+            Validator(
+                validation_function=lambda: not guided
+                and (
+                    set(image_repositories.keys()) != set(get_template_function_resource_ids(template_file, IMAGE))
+                    and image_repositories
```
mndeveci, line 64: Can customers define `key=value` pairs along with general image repository? Which will mean use specific image repositories for the ones which is in the dict, but use the other one which is not in the dict.

sriram-mv:
@mndeveci : if we dont special case how we save to the config file. It saves as below, thats why we need it.

```
[default.deploy.parameters.image_repositories]
HelloWorldFunction = "847756679349.dkr.ecr.us-east-1.amazonaws.com/sam-cli"
```

samcli/lib/cli_validation/image_repository_validation.py
sriram-mv: nope, its one or other.

samcli/commands/_utils/template.py
sriram-mv: does that process the template? I will take a look at it.

samcli/commands/_utils/options.py
sriram-mv: I had this because of previous validation issues when trying to do validation in the respective callbacks, we should just be able to return dict, now.

samcli/commands/_utils/template.py
sriram-mv: its very specific to build, we would need some refactoring to make it more generic so that we can share that.

mndeveci: I would still use `SamFunctionProvider` https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/providers/sam_function_provider.py#L16 which will handle parameter overrides etc.

sriram-mv: There are some circular dependencies that we need to take care of here, I vote we do this in a separate PR.

moelasmar (approved):

metaskills:
There is still no documentation on sam package (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-package.html) guides page. Trying to find the format to use here because I think this arg will allow me to deploy two functions with the same image.

JamesDougherty:
Hopefully this will help someone and save them time. There is no documentation on this as @metaskills mentioned and no examples on how to use it. So after spending more time than needed on it and looking through the source code on this, I was able to figure out what they intended.

Let's say you have two functions in your template file that use the Image package type, like the following:

```
FooFunction:   # <------------------This is a functions logical ID, take note of this.
  Properties:
    PackageType: Image
    ...
  Metadata:
    DockerTag: latest
    DockerContext: ./opt/dist/functions/foo-function/        # Wherever your Dockerfile is
    Dockerfile: Dockerfile

BarFunction:   # <------------------This is a functions logical ID, take note of this.
  Properties:
    PackageType: Image
    ...
  Metadata:
    DockerTag: latest
    DockerContext: ./opt/dist/functions/bar-function/        # Wherever your Dockerfile is
    Dockerfile: Dockerfile
```

After you run the `sam build` command it will add the following `ImageUri` to the template it generates in the .aws-sam/build directory. The ImageUri consists of the image repository name (functions logical ID in all lower case) and DockerTag concatenated. Here is an example of what it will add for the functions defined above:

```
FooFunction:
  Properties:
    PackageType: Image
    ImageUri: foofunction:latest  # <------------- Added by SAM Build
    ...
  Metadata:
    DockerTag: latest
    DockerContext: ./opt/dist/functions/foo-function/
    Dockerfile: Dockerfile

BarFunction:
  Properties:
    PackageType: Image
    ImageUri: barfunction:latest  # <------------- Added by SAM Build
    ...
  Metadata:
    DockerTag: latest
    DockerContext: ./opt/dist/functions/bar-function/
    Dockerfile: Dockerfile
```

Now when you run the `sam deploy` command you need to add the `--image-repositories` in the following format:

```
--image-repositories <Function Logical ID>=<Function Image Repository>, <Function Logical ID>=<Function Image Repository>, ...
```

Where:
**Function Logical ID** = The functions logical ID as defined in the SAM template
**Function Image Repository** = The functions ECR image repository that SAM created in ECR URL format

So using the above example we would call `sam deploy` using the following:

```
sam deploy ... --image-repositories FooFunction=<account id>.dkr.ecr.<region>.amazonaws.com/foofunction, BarFunction=<account id>.dkr.ecr.<region>.amazonaws.com/barfunction
```

Where:
**Account ID** = Your AWS account ID
**Region** = Your AWS region (like us-east-1)

I have this automated using a PowerShell script so you will just need to figure out how to handle it using your tooling / use case.

Enjoy.

ygorth:
Hi @JamesDougherty I know this is merged and closed, but before I create an issue, do you know why I am getting the error below for sam package:
```
sam package --image-repositories Function1=?.dkr.ecr.us-east-1.amazonaws.com/spn-int-ecr-lam-snstats-fun-ftppublisher, Function2=?.dkr.ecr.us-east-1.amazonaws.com/spn-int-ecr-lam-snstats-fun-scapfriendly --output-template-file functions/spn/sn-stats/packaged-template.yaml --region us-east-1
..
Error: Got unexpected extra argument (Function2=946043939236.dkr.ecr.us-east-1.amazonaws.com/spn-int-ecr-lam-snstats-fun-scapfriendly)
```

JamesDougherty:
Hello @ygorth - Thank you for the correction. I was only testing with one repository at the moment and made the assumption that they designed it how the parameter overrides work "sam deploy ... --parameter-overrides foo=bar, hello=world". Of course it would be different. This is why it would be nice to have some official documentation so we don't have to assume anything. Anyway, thanks again.

jmamin:
Hello,

Any advice how to fix this issue? image name is compatible with format and regex but still errors out!:

MacBook-Pro:lambda-local jamin$ sam deploy --guided
Usage: sam deploy [OPTIONS]
Try 'sam deploy -h' for help.

Error: Invalid value for '--image-repositories': --image-repositories is not a valid format, it needs to be of the form function_logical_id=ECR_URI
MacBook-Pro:lambda-local jamin$ 

samconfig.toml
version = 0.1
[default]
[default.deploy]
[default.deploy.parameters]
stack_name = "everflowclicks"
s3_bucket = "aws-sam-cli-managed-default-samclisourcebucket-1785ltpwmjlwc"
s3_prefix = "everflowclicks"
region = "us-west-2"
confirm_changeset = true
capabilities = "CAPABILITY_IAM"
image_repositories = ["880954911558.dkr.ecr.us-west-2.amazonaws.com/everflow-app"]
disable_rollback = true

PR #2472, opened 2020-12-11
feat(runtime telemetry): Added Runtime Telemetry for Local Commands 

CoshUS:
Data on runtime information is not available.

Restructured Metric and Telemetry.
Added Metric class and decorators to capture function parameters and return values into a Metric object which then can be emitted from Telemetry.
The decorators are accessible globally and provides a versatile way, similar to Logger, to store metrics.
Ex:
```
@capture_parameter("runtimeMetric", "runtimes", 1, parameter_nested_identifier="runtime", as_list=True)
def invoke(self, function_config, event, debug_context=None, stdout=None, stderr=None):
  ...
```

`pytest .\tests\integration\telemetry` Passed

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/telemetry/metric.py
```diff
@@ -0,0 +1,355 @@
...
+            return True, 'Warning applicable'
+    """
+
+    def decorator(func):
+        """
+        Actual decorator method with warning names
+        """
+
+        def wrapped(*args, **kwargs):
+            telemetry = Telemetry()
```
mgrandis, line 60: Shouldn't we check here that Telemetry is enabled so that we don't execute all this code if it isn't?

Maybe add a `is_enabled` function to Telemetry and use it

CoshUS: The check for telemetry enabled or not is done in Telemetry.emit(...). Building a Metric object is trivial amount of work and having a single point of checking reduces possible error.

hawflau (approved):

mgrandis (approved):

PR #2481, opened 2020-12-15
chore: update the warm-containers & debug-function options help wording

moelasmar:
Update the new options (warm-containers && debug-function) help description 

#### Checklist

- [ ] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -153,13 +153,25 @@ def warm_containers_common_options(f):
     warm_containers_options = [
         click.option(
             "--warm-containers",
-            help="Optional. Specifies how AWS SAM CLI manages containers for each function.",
+            help="""
+            \b
+            Optional. Specifies how AWS SAM CLI manages 
+            containers for each function.
+            Two modes are available:
+            EAGER: Containers for all functions are 
+            loaded at startup and persist between 
+            invocations.
+            LAZY:  Containers are only loaded when each 
+            function is first invoked. Those containers 
+            persist for additional invocations.
+            """,
             type=click.Choice(ContainersInitializationMode.__members__, case_sensitive=False),
         ),
         click.option(
             "--debug-function",
             help="Optional. Specifies the Lambda Function logicalId to apply debug options to when"
-            " --warm-containers is specified ",
+            " --warm-containers is specified.  This parameter applies to --debug-port, --debugger-path,"
+            " and --debug-args.",
             type=click.STRING,
             multiple=False,
         ),


wchengru (approved):

samcli/commands/local/cli_common/options.py
```diff
@@ -153,13 +153,25 @@ def warm_containers_common_options(f):
     warm_containers_options = [
         click.option(
             "--warm-containers",
-            help="Optional. Specifies how AWS SAM CLI manages containers for each function.",
+            help="""
+            \b
```
mndeveci, line 157: A dummy question, what is the `\b` here? I didn't see it in the other help texts.

samcli/commands/local/cli_common/options.py
```diff
@@ -153,13 +153,25 @@ def warm_containers_common_options(f):
...
+            Optional. Specifies how AWS SAM CLI manages 
+            containers for each function.
+            Two modes are available:
+            EAGER: Containers for all functions are
```
aahung, line 161: I did not find our existing multi-line `help=`. What does the stdout look like here?

samcli/commands/local/cli_common/options.py
moelasmar: \b is for click formatting to prevent rewrapping, as I need to have new lines in the help text.
without \b, the help text will look like this:
```
--warm-containers [EAGER|LAZY]    Optional. Specifies how AWS SAM CLI manages
                                  containers for each function. Two modes are
                                  available: EAGER: Containers for all
                                  functions are  loaded at startup and persist
                                  between  invocations. LAZY:  Containers are
                                  only loaded when each  function is first
                                  invoked. Those containers  persist for
                                  additional invocations.
```

but with \b, it will look like this
```
--warm-containers [EAGER|LAZY]    Optional. Specifies how AWS SAM CLI manages 
                                  containers for each function.
                                  Two modes are available:
                                  EAGER: Containers for all functions are 
                                  loaded at startup and persist between 
                                  invocations.
                                  LAZY:  Containers are only loaded when each 
                                  function is first invoked. Those containers 
                                  persist for additional invocations.
```

samcli/commands/local/cli_common/options.py
moelasmar: it looks like this:
```
--warm-containers [EAGER|LAZY]    Optional. Specifies how AWS SAM CLI manages 
                                  containers for each function.
                                  Two modes are available:
                                  EAGER: Containers for all functions are 
                                  loaded at startup and persist between 
                                  invocations.
                                  LAZY:  Containers are only loaded when each 
                                  function is first invoked. Those containers 
                                  persist for additional invocations.
```

aahung (approved):

samcli/commands/local/cli_common/options.py
aahung: this is pretty cool. Good to know, thx

samcli/commands/local/cli_common/options.py
mndeveci: Good to know, this is really helpful! Thanks!

hoffa: Good to see the new names! Much better! 👍

Issue #2417, opened 2020-11-25
capabilities in samconfig.toml does not support list


benkehoe:
In `samconfig.toml`, `parameter_overrides` is able to be a list of strings, which helps format the file better (though see #2253 for a request that this be improved).

I expected that I could do the same with `capabilities` when I have multiple capabilities. For example:

```toml
version = 0.1
[default]
[default.deploy]
[default.deploy.parameters]
capabilities = [
  "CAPABILITY_IAM",
  "CAPABILITY_NAMED_IAM"
]
parameter_overrides = [
  "Foo=Bar"
]
```

However, this gives me the error `AttributeError: 'Array' object has no attribute 'split'`.

It appears that this function should handle conversion to tuple for the type `tomlkit.items.Array` https://github.com/aws/aws-sam-cli/blob/876047ac35c4b2d5a62749a01c5fd3f1ef637953/samcli/commands/_utils/options.py#L276

aahung:
v1.16.0 is released, please reopen if the issue still persists

PR #2490, opened 2020-12-19
feat: Add support for list of capabilities in samconfig.toml

G4brym:
#2417 

To allow for capabilities to be configured in a list orientation, while maintaining support for the previous configurations

It adds support for the `tomlkit.items.Array` class
I decided to import the `tomlkit.items.Array` as `tomlkit_Array` to not cause confusion with python's array

None, that i found

#### Checklist
- [ ] Write documentation

This is my first pr, and i think the change is so small that the only item in the checklist that applies here is the Documentation, but i didn't find the write place to do it.


moelasmar (changes requested):
Could you add some uint/integration test cases that cover this change.

G4brym:
Just added some unit tests for my changes.
After digging more into the tomlkit, I found that the `tomlkit.items.Array` behaves like a normal list and the result of the `isinstance(value, list)` is also true when using the `tomlkit.items.Array` class.
So, I modified the code to use a list instead.

samcli/commands/_utils/options.py
```diff
@@ -362,7 +362,7 @@ def notification_arns_override_option(f):
 
 
 def _space_separated_list_func_type(value):
-    return value.split(" ") if not isinstance(value, tuple) else value
+    return value.split(" ") if not isinstance(value, tuple) and not isinstance(value, list) else value
```
mgrandis, line 365: Hi @G4brym, thank you for your contribution!

I think this code could be clearer:
```python
    if isinstance(value, str):
        return value.split(" ")
    if isinstance(value, (list, tuple)):
        return value
    raise ValueError()
```

What do you think?

G4brym:
Thanks for pointing it out @mgrandis
I'm used to make the code more compact than it should, but doing it your way does make the code much more readable.
I'm looking forward to help you guys with more issues, I'm just picking some smaller first to know the project better :smile:

mgrandis:
Thanks for the update @G4brym!
I see you added a test with an integer, could you also add ones for the other types that [TOML](https://toml.io) supports?
I think float, boolean, datetime, dict/object should be enough.

G4brym:
Hey, I just added those data types you mentioned

mgrandis (approved):

elbayaaa:
Do we need to make changes [here](https://github.com/aws/aws-sam-cli/blob/07fcd611d61eb2f213d03702391929ba22550adf/samcli/commands/deploy/guided_config.py#L59-L60) as well? I'm afraid that this change will successfully read the capabilities array from the toml file while executing `sam deploy` then override it to a space-separated string while writing to the file again.

benkehoe:
@elbayaaa I think it'd be preferable to have that use the list format as well.

elbayaaa:
@benkehoe yes I agree on that. I'm just worried about the inconsistency with the other options [here](https://github.com/aws/aws-sam-cli/blob/07fcd611d61eb2f213d03702391929ba22550adf/samcli/commands/deploy/guided_config.py#L59-L60). I think the scope should be expanded to include all possible options not just the capabilities.

mgrandis:
This could indeed be a good opportunity to use the same method (list) everywhere, we need to be cautious with regressions however.

elbayaaa (approved):
Approving with the understanding that this is going to be an ongoing improvement. The next step is to save the value as an array instead of space-separated string.

moelasmar (approved):

mgrandis:
Thank you for your contribution @G4brym!

PR #2503, opened 2020-12-23
ci: Bring mypy typechecking back

aahung:

This brings back basic mypy check which was originally implemented in #2385 and reverted in #2396 to make room for a massive merge.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


appveyor.yml
```diff
@@ -79,6 +79,7 @@ for:
       - "venv\\Scripts\\activate"
       - "pytest --cov samcli --cov-report term-missing --cov-fail-under 94 tests/unit"
       - "pylint --rcfile .pylintrc samcli"
+      - "mypy setup.py samcli tests"
```
hoffa, line 82: I assume it already has it installed?

tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -4,7 +4,7 @@
 import tempfile
 from unittest import TestCase
 
-from boto3 import Session
+from boto3.session import Session
```
hoffa, line 7: What was the issue? Are the `Session`s identical?

appveyor.yml
hoffa, line 82: I assume it already has it installed?

aahung: it is installed in `dev.txt`

tests/integration/init/schemas/schemas_test_data_setup.py
aahung: it is, mypy is not happy if we import `Session` from `boto3/__init__.py` which imports from `boto3.session`. I don't know how it works, but it seems when file A import X from B, mypy will prevent us from import X from A.

samcli/settings/__init__.py
```diff
@@ -17,7 +17,7 @@
 
 import os
 
-if "__SAM_CLI_TELEMETRY_ENDPOINT_URL" not in os.environ:
-    telemetry_endpoint_url = "https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics"
-else:
-    telemetry_endpoint_url = os.getenv("__SAM_CLI_TELEMETRY_ENDPOINT_URL")
+
+telemetry_endpoint_url = os.getenv(
```
sriram-mv, line 21: There is a reason why this was coded this way, just racking my brain to remember it.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
```diff
@@ -731,10 +734,7 @@ def test_fn_and_arguments_invalid_formats(self, name, intrinsic):
             self.resolver.intrinsic_property_resolver({"Fn::And": intrinsic}, True)
 
     @parameterized.expand(
-        [
-            ("Fn:And must have all arguments that resolves to booleans".format(item), item)
-            for item in [{}, 42, None, "test"]
-        ]
+        [(f"Fn:And must have all arguments that resolves to booleans {item}", item) for item in [{}, 42, None, "test"]]
```
sriram-mv, line 737: is there a reason, we are doing this change just here? do we need a larger sweep?

samcli/settings/__init__.py
aahung: But according to the [python doc](https://docs.python.org/3/library/os.html#os.getenv), this should be same effect.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
aahung: mypy complaints about this because we were calling `.format` on a string that does not have any placeholder. So I fixed them all. Because f-string are less prune to such error, I used it.

samcli/settings/__init__.py
sriram-mv: seems alright.

tests/unit/lib/intrinsic_resolver/test_intrinsic_resolver.py
sriram-mv: yes, I meant is it just these files, but as you said, looks like mypy only complained about these files.

sriram-mv (approved):
LGTM

hoffa (approved):
LGTM 👍

PR #2522, opened 2021-01-06
chore: Make pylintrc ignore fewer rules and pylintrc more readable

aahung:

This brings back some pylint improvements which were originally implemented in #2375 and reverted in #2400 to make room for a massive merge.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


CoshUS (approved):

samcli/commands/local/cli_common/click_mutex.py
```diff
@@ -41,7 +41,7 @@ def handle_parse_result(self, ctx, opts, args):
                     if mutex_opt in opts:
                         req_cnt -= 1
 
-                if req_cnt == 0:
+                if not req_cnt:
```
hoffa, line 44: Does it complain about this? It's weird using boolean logic on integers.

samcli/commands/publish/command.py
```diff
@@ -15,8 +15,9 @@
 
 LOG = logging.getLogger(__name__)
 
-SAM_PUBLISH_DOC = "https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-template-publishing-applications.html"  # noqa
-SAM_PACKAGE_DOC = "https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-package.html"  # noqa
+SAM_DOC_PREFIX = "https://docs.aws.amazon.com/serverless-application-model/latest/developerguide"
+SAM_PUBLISH_DOC = f"{SAM_DOC_PREFIX}/serverless-sam-template-publishing-applications.html"  # noqa
```
hoffa, line 19: What was the error here? Doesn't `noqa` prevent the complaints? Feels more of a fix to deal with `pylint`s whims.

samcli/lib/deploy/deployer.py
```diff
@@ -345,9 +345,8 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
 
         while stack_change_in_progress and retry_attempts <= self.max_attempts:
             try:
-
                 # Only sleep if there have been no retry_attempts
-                time.sleep(self.client_sleep if retry_attempts == 0 else 0)
+                time.sleep(0 if retry_attempts else self.client_sleep)
```
hoffa, line 349: What was the complaint here? As above, not sure why it prefers boolean logic on integers.

hoffa (commented):
A lot of these changes seem to be about `pylint` enforcing a draconian line length limit. `black` on the other hand already partially enforces this, but in a more pragmatic manner ("an inherently long URL is long, so it's okay"). I wonder if it might be better to turn of the length enforcement and delegate that to `black`?

aahung:
> black on the other hand already partially enforces this, but in a more pragmatic manner ("an inherently long URL is long, so it's okay"). 
black does nothing about strings, it won't split long string like this, either make a super long comment line shorter.
image

But I do agree we should make URLs exceptions

samcli/commands/local/cli_common/click_mutex.py
aahung: In pylint's source

> Most of the times you should use the fact that integers with a value of 0 are false.
    An exception to this rule is when 0 is allowed in the program and has a
    different meaning than None!

https://github.com/PyCQA/pylint/blob/master/pylint/extensions/comparetozero.py#L23

samcli/commands/publish/command.py
aahung: It was a fix. I will remove `# noqa` here instead of reverting in this case because making the common prefix into a variable does make it easier to maintain

aahung:
I reverted the changes that split URLs into multiple lines. It made copy/paste difficult. For all other splitting-the-line scenarios, I think they are pretty legit and I don't have a good reason not to do them.

hoffa (approved):
Looking great! Thank you! 🙌

PR #2524, opened 2021-01-08
feat: Support SHUTDOWN Events for Extensions

awood45:
This adds the `--shutdown` flag to `local invoke`, `local start-api`, and the `local start-lambda` commands. When set, local execution environments will send a SHUTDOWN event to any registered extensions when a container is stopped.

Add support for SHUTDOWN events when testing AWS Lambda Extensions locally.

Adds a `--shutdown` flag to `local` commands that will trigger a SHUTDOWN event when the execution environment containers are stopped for any reason.

#### Checklist

- [x] Write unit tests
- [X] `make pr` passes
- [X] Write documentation


awood45:
TODO: Still working on some new tests for the new code paths, existing test changes focus on getting existing tests to pass.

samcli/local/docker/container.py
```diff
@@ -180,6 +180,34 @@ def create(self):
...
+            LOG.debug("Container was not created, cannot run stop.")
+            return
+
+        try:
+            self.docker_client.containers.get(self.id).stop(timeout=time)
+        except docker.errors.NotFound:
+            # Container is already removed
+            LOG.debug("Container with ID %s does not exist. Cannot stop!", self.id)
+        except docker.errors.APIError as ex:
+            msg = str(ex)
```
mgrandis, line 203: Could you use `ex.message` instead of this? (does it contain both text parts you are searching?)

samcli/local/docker/container.py
```diff
@@ -180,6 +180,34 @@ def create(self):
...
+            removal_in_progress = ("removal of container" in msg) and ("is already in progress" in msg)
+
+            # When removal is already started, Docker API will throw an exception
+            # Skip such exceptions.
+            if not removal_in_progress:
+                raise ex
+
```
mgrandis, line 210: We should maybe `LOG.debug` something here if removal is in progress

tests/unit/local/docker/test_container.py
```diff
@@ -311,6 +311,72 @@ def test_must_fail_if_already_created(self):
...
+
+        self.container.is_created = Mock()
+
+    def test_stop_with_timeout(self):
+        self.container.is_created.return_value = True
+        real_container_mock = Mock()
+        self.mock_docker_client.containers.get.return_value = real_container_mock
+        real_container_mock.remove = Mock()
+
+        self.container.stop(time=3)
```
mgrandis, line 338: `3` is the default value, you should maybe use another one

mgrandis:
Minor discussion points, otherwise looks good to me!

mgrandis (approved):

mndeveci (approved):

tests/unit/local/docker/test_container.py
awood45: This was mainly an explicit over implicit thing, in case it's changed by the Docker SDK (which has made breaking changes in the past). Will keep this way, but fair callout.

samcli/local/docker/container.py
awood45: Adding this to the `delete` API as well.

samcli/local/docker/container.py
awood45: We already use this logic as-is in `delete` which works, so for this one inclined to not mess with it.

samcli/local/docker/container.py
```diff
@@ -201,6 +231,8 @@ def delete(self):
             # Skip such exceptions.
             if not removal_in_progress:
                 raise ex
+            # else debug log explaining that we had and skipped an exception
```
mgrandis, line 234: I think we should actually move the two lines of comment from 230:231 here, they describe what we are doing.

Also, is passing the exception to the logger better that using its string representation? Like so:
`LOG.debug("Container removal is in progress, skipping exception", exc_info=True)`

awood45: I just updated the comments above to explain the "else" condition

PR #2525, opened 2021-01-08
chore: Make methods that do not need to be instance methods static

aahung:
…static

There are many instance methods do not need to be instance methods since `self` is not being used. Making it static increases readability. 

Making them static. One exception, for `BuildStrategy`, the class is meant to be an abstract class, so I made its two instance methods abstract instead of static. 

no, 
> A static method can be called either on the class (such as C.f()) or on an instance (such as C().f()). https://docs.python.org/3/library/functions.html#staticmethod

We don't need to change how we invoke them.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/lib/build_module/test_build_strategy.py
+++ b/tests/unit/lib/build_module/test_build_strategy.py
@@ -39,14 +39,22 @@ class BuildStrategyBaseTest(TestCase):
         self.build_graph.put_layer_build_definition(self.layer_build_definition2, Mock())
 
 
+class _TestBuildStrategy(BuildStrategy):
+    def build_single_function_definition(self, build_definition):
+        return {}
+
+    def build_single_layer_definition(self, layer_definition):
+        return {}
+
+
 class BuildStrategyTest(BuildStrategyBaseTest):
     def test_build_functions_layers(self):
-        build_strategy = BuildStrategy(self.build_graph)
+        build_strategy = _TestBuildStrategy(self.build_graph)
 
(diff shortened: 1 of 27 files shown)


tests/unit/lib/build_module/test_build_strategy.py
```diff
@@ -39,14 +39,22 @@ def setUp(self):
...
+        return {}
+
+    def build_single_layer_definition(self, layer_definition):
+        return {}
+
+
 class BuildStrategyTest(BuildStrategyBaseTest):
     def test_build_functions_layers(self):
-        build_strategy = BuildStrategy(self.build_graph)
+        build_strategy = _TestBuildStrategy(self.build_graph)
```
elbayaaa, line 52: would you explain why we need this  "_TestBuildStrategy" class?

aahung: `BuildStrategy` is an abstract class so I need to create a subclass to instantiate an object.

aahung: (and adding prefix `_` to indicate it is a "private" class)

elbayaaa (approved):

aahung:
(manually rebase the latest develop to resolve the conflict)

aahung:
Added type hints

CoshUS (approved):

elbayaaa (approved):

PR #1501, opened 2019-11-05
Init User Flow Rework

awood45:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2529, opened 2021-01-12
refactor: Remove python-dateutil dependency

mvanbaak:
Need python-dateutil 2.8.1 in projects, and since it brings 0 breaking changes compared to 2.8.0, 2.8.1 should be allowed.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
I am skeptical on this. We originally had this but added the restriction here: https://github.com/aws/aws-sam-cli/pull/1501/files. There is no details but am assuming there was an explicit reason at the time.

SAM CLI is not really meant to be install along side your project. If you really want to use pip, I suggest you to install in an venv so conflicts with your packages do not happen.

mgrandis:
Couldn't we just drop this dependency?

It seems we are only using it in https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/time.py#L95:
`some_time = some_time.astimezone(tzutc())`

Could this be replaced by:
`some_time = some_time.astimezone(datetime.timezone.utc)`

mvanbaak:
Yes, we can:

```python
>>> import datetime
>>> from dateutil.tz import tzutc
>>> foo = datetime.datetime.now()
>>> print(foo)
2021-01-14 14:23:00.009815
>>> bar = foo.astimezone(datetime.timezone.utc)
>>> print(bar)
2021-01-14 13:23:00.009815+00:00
>>> baz = foo.astimezone(tzutc())
>>> print(baz)
2021-01-14 13:23:00.009815+00:00
>>>
```

I'll update the PR.

jfuss (changes requested):
This generally looks good. Since we are dropping the dependency, we will need to generate a new https://github.com/aws/aws-sam-cli/blob/develop/requirements/reproducible-linux.txt file. This can be done through `make update-reproducible-reqs`. It looks like dateutil is pulled in from other dependencies though, so it still needs to be in the ATTRIBUTION.txt file for our build images.

mgrandis:
@mvanbaak Could you address @jfuss comments please?
Generate the `reproducible-linux.txt` file and restore the `ATTRIBUTION.txt` file section.
Thanks!

mvanbaak:
Yes, sorry for the delay. Will look at it this week.

mvanbaak:
Done.

mgrandis (approved):

jfuss (approved):

mgrandis:
Thanks @mvanbaak !

Issue #2494, opened 2020-12-21
Regression: Event is not UTF-8 encoded when testing lambda locally


birtles:
### Description:

Passing a non-Latin body to `sam local invoke` produces a `UnicodeEncodeError`.

My guess is the changes from #2281 might be involved (cc @awood45).

### Steps to reproduce:

Unfortunately I haven't yet had time to produce a minimal repo but the command line I am using is:

```bash
sam local invoke "FunctionName" -e <path to below event json file>.json -n <path to test env>.json
```

Event JSON file:

```json
{
  "Records": [
    {
      "messageId": "19dd0b57-b21e-4ac1-bd88-01bbb068cb78",
      "receiptHandle": "MessageReceiptHandle",
[...]

PR #2533, opened 2021-01-16
fix: local invoke encoding

hoffa:

#2494

Allow events made of UTF-8 bytes.

Encode `str` to `utf-8` before passing to `requests.post`, otherwise it silently encodes to `latin-1` and fails on unsupported code points.

None.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


sriram-mv (approved):

CoshUS (approved):

tests/integration/testdata/invoke/event.json
```diff
@@ -1,4 +1,5 @@
 {
+    "utf8": "안녕",
```
jfuss, line 2: This works though, I would ask for a separate integ test for this. Reason being as a contributor, I want to know what my test actually broke. If this regresses, we will see some test fail with name "test_wait_for_result_no_error" but doesn't really tell me why or what I really broke at a glance. This is more of a raise the bar request as what you have works.

hoffa: The reason I kept it here is that UTF-8 support is inherently coupled to nearly all text representation (and more so for arbitrary user input) in our codebase, i.e. UTF-8 support is _the_ default case.

hoffa: Separated the tests; in the general case I don't think separating is necessary, but in this case it seems reasonable as there was a regression.

Issue #1645, opened 2019-12-06
Global AllowCredentials is not working with localhost / local server


dz902:
```
Globals:
  Api:
    Cors:
      AllowOrigin: "'http://localhost:9000'"
      AllowCredentials: "'true'"
      AllowMethods: "'GET,POST'"
      AllowHeaders: "'*'"
```

I have seen this: https://github.com/awslabs/serverless-application-model/issues/1166

But `AllowCredentials` is definitely not working for either `true`, `"true"` or `"'true'"`. Since this is a very important setting, not sure why no one is raising this up.

All other cors settings work without problem, just `AllowCredentials`.

CGreenburg:
I'm also running into this issue. I set cookies in my API but the cookies won't persist in the browser since the `Access-Control-Allow-Credentials` header doesn't return true from the OPTIONS request. I think this is a necessary addition.

sanathkr:
WIP PR #1648 is approved. We need someone to complete unit and integration tests for the change before it can be merged/released

walkerlangley:
Just wanted to follow up here and see if anyone has a workaround for this. Currently running aws-sam-cli v0.40.0 and can't get the AllowCredentials: true header to be provided in the response no matter what I try.

dz902:
It seems the PR was reverted somehow. As of v0.47.0, this problem persists.

Not sure how others are using SAM and API Gateway, but without this almost no Web applications can be developed and tested in browser locally.

Why isn't this issue a high priority? Not many folks doing some serious Serverless Web business...? 🤔

CGreenburg:
@dz902 I hadn't noticed it was reverted. I reached out on the revert pull request (#1664) to see why.

yhengyeepin:
Just ran into this issue recently. It is blocking local testing.

PR #2058, opened 2020-06-19
fix: cors support for "Access-Control-Allow-Credentials" header in local api

evan-richard:
*Issue #, if available:*
fix for #1645 

*Why is this change necessary?*
AllowCredentials is not working for any of True, true, or "'true'" while using the local start-api command. This setting is necessary for making requests to the local api while using the withCredentials flag in the request.

*How does it address the issue?*
While parsing the other cors options, we will now also pars AllowCredentials and add it to the "Access-Control-Allow-Credentials". AllowCredentials should accept the values True, true, and "'true'" (note the required nested single quotes).

*What side effects does this change have?*
No known side effects of this change.

*Did you change a dependency in `requirements/base.txt`?*
    *If so, did you run `make update-reproducible-reqs`*
No change.

*Checklist:*

- [x] Write Design Document ([Do I need to write a design document?](https://github.com/awslabs/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.rst#design-document))
- [x] Write unit tests
- [x] Write/update functional tests
- [x] Write/update integration tests
- [x] `make pr` passes
- [x] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Title: Design document
====================================

Use this as a template to write a design document when adding new
commands or major features to SAM CLI. It helps other developers
understand the scope of the project, validate technical complexity and
feasibility. It also serves as a public documentation of how the feature
actually works.

**Process:** 

1. Copy this template to another file in the `designs` folder.
2. Fill out the sections in the template.
3. Send a "Work In Progress" Pull Request with your design document. We can discuss the
designs in more detail and iterate on the requirements. Feel free to
start implementing a prototype if you think it will help flush out
design.
4. Once the PR is approved, create Github Issues for each task
listed in the document and start implementing them.

What is the problem?
--------------------

AllowCredentials is not working for any of True, true, or "'true'" while using the local start-api command. This setting is necessary for making requests to the local api while using the withCredentials flag in the request.

Current behavior: Log from the Chrome developer tools console is "Access to XMLHttpRequest at [local-api-url] from origin [clien-url] has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute." 

Expected behavior: The request should complete successfully when using the withCredentials attribute.

[...]

PR #2536, opened 2021-01-17
fix: honour the AllowCredentials CORS option in SAM templates

discorev:
#1645

This change is needed to complete an incomplete feature in the local API CORS implementation. Currently the `AllowCredentials` CORS option is ignored in the local API mock and blocks any local development that has a CORS requirement and needs the `Access-Control-Allow-Credentials` header to be set.

Whilst parsing CORS options for a REST API, the `AllowCredentials` option is parsed allowing for a boolean or quoted string value. If and only if this value is a representation of true, the header is set to the string `true`

Whilst parsing CORS options for a Http API the `AllowCredentials` options is parsed only allowing for a boolean value (following the documentation). If this is set to true, the header is set to the string `true`

The header is always output as a lower-cased string to be compliant with [section 5.6 of RFC7480](https://tools.ietf.org/html/rfc7480#section-5.6)

There are no new dependancies introduced and this does not impact any other areas of functionality.

There are no side effects outside of having the `Access-Control-Allow-Credentials` header set when the template demands it.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


hoffa:
Thanks for the PR!

Related PRs for reference: #1648 (reverted in #1664) and #2058.

samcli/lib/providers/cfn_base_api_provider.py
```diff
@@ -121,12 +127,23 @@ def _get_cors_prop(cors_dict, prop_name):
...
         Return
         ------
         A string with the extra quotes removed
         """
         prop = cors_dict.get(prop_name)
         if prop:
+            if allow_bool and isinstance(prop, bool):
+                # The only boolean header is allow credentials and the only valid value for that is 'true'
+                if not prop:
+                    return None
```
hoffa, line 145: Unreachable; `prop` is truthy due to L141.

samcli/lib/providers/cfn_base_api_provider.py
```diff
@@ -165,16 +182,26 @@ def extract_cors_http(self, cors_prop):
             allow_headers = self._get_cors_prop_http(cors_prop, "AllowHeaders", list)
             if isinstance(allow_headers, list):
                 allow_headers = ",".join(allow_headers)
+
+            # Read AllowCredentials but only output the header with the case-sensitive value of true
+            # (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials)
+            allow_credentials = "true" if self._get_cors_prop_http(cors_prop, "AllowCredentials", bool) else None
```
hoffa, line 188: Would getting rid of the `if` work?

If the prop exists and it's `True`, `_get_cors_prop` will return `true`, otherwise `None`. Seems identical here.

hoffa (changes requested):
Some minor comments; overall looks great!

samcli/lib/providers/cfn_base_api_provider.py
discorev: Due to the RFC requiring that Access-Control-Allow-Credentials be the case sensitive value `true` and a boolean python value being converted to the value `True` when output, this is required to ensure correct casing of the output

hoffa: Gotcha, just noticed this one call `_get_cors_prop_http`, which type checks and passes the value through.

hoffa (approved):

c2tarun (approved):

mgrandis (approved):

hoffa:
Thanks @discorev; merged!

PR #2549, opened 2021-01-22
Fix: Adjusted build command behaviour treating Serverless and Lambda functions using InlineCode

qingchm:
Because currently the build command builds the parent directory into a folder and insert that folder reference as CodeUri property into the template for functions using InlineCode. We want to skip builds when an InlineCode property is specified for serverless functions, similarly skip when a ZipFile property is specified for lambda functions.

This pr fixes the issue by adding an inlinecode property into the sam function builder. When this attribute is present, we do not add that function into the build list.

People who use InlineCode in their sam template will observe change in behaviour, some artifacts they used to build all the time will not appear after builds anymore.

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -218,10 +218,12 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
             properties = resource.setdefault("Properties", {})
 
             if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["CodeUri"] = store_path
+                if "InlineCode" not in properties or properties["InlineCode"] is None:
+                    properties["CodeUri"] = store_path
 
             if resource_type == SamBaseProvider.LAMBDA_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["Code"] = store_path
+                if "ZipFile" not in properties["Code"] or properties["Code"]["ZipFile"] is None:
```
mgrandis, line 225: We may have a condition where `Code` is not in `properties` (line 218), you may want to check it.

qingchm: makes sense, I will add the logic here

mgrandis: Also we could maybe get the `PackageType` value into a variable instead of calling get with a default 4 times.


```python
package_type = properties.get("PackageType", ZIP)

if package_type == ZIP:
    if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and ("InlineCode" not in properties or properties["InlineCode"] is None):
        ...

    if resource_type == SamBaseProvider.LAMBDA_FUNCTION and ("ZipFile" not in properties["Code"] or properties["Code"]["ZipFile"] is None):
         ...

if package_type == IMAGE:
     ...
```

mgrandis: just saw that it is done this way in the other files, `packagetype` is without a `_` though.

qingchm: good suggestion!

qingchm: @mgrandis the "packagetype" you see in other files is an attribute name in provider.py. I think package_type would be the better name here

mgrandis: I would go with `package_type`, always easier to read with underscores.

samcli/lib/providers/provider.py
```diff
@@ -57,10 +59,15 @@ def __init__(self):
...
 
     def add_function(self, function):
+        if function.inlinecode and function.inlinecode is not None:
+            return
         self.result.get("Function").append(function)
 
     def add_functions(self, functions):
-        self.result.get("Function").extend(functions)
+        for function in functions:
+            if function.inlinecode and function.inlinecode is not None:
```
mgrandis, line 68: You could maybe call the `add_function` above here

qingchm: makes sense!

samcli/lib/build/app_builder.py
```diff
@@ -216,20 +216,26 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
...
-            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["CodeUri"] = store_path
+            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and package_type == ZIP:
+                if "InlineCode" not in properties or properties["InlineCode"] is None:
+                    properties["CodeUri"] = store_path
 
-            if resource_type == SamBaseProvider.LAMBDA_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["Code"] = store_path
+            if resource_type == SamBaseProvider.LAMBDA_FUNCTION and package_type == ZIP:
+                if "Code" in properties:
```
mgrandis, line 226: I think you should be able to do it all in one condition:
```python
if "Code" not in properties or "ZipFile" not in properties["Code"] or properties["Code"]["ZipFile"] is None:
```

samcli/lib/build/app_builder.py
```diff
@@ -216,20 +216,26 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
 
             resource_type = resource.get("Type")
             properties = resource.setdefault("Properties", {})
+            package_type = properties.get("PackageType", ZIP)
 
-            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["CodeUri"] = store_path
+            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and package_type == ZIP:
```
mgrandis, line 221: You should do it like in other functions:
```python
if package_type == ZIP:
    if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and ("InlineCode" not in properties or properties["InlineCode"] is None):
        properties["CodeUri"] = store_path
    elif resource_type == SamBaseProvider.LAMBDA_FUNCTION and ("Code" not in properties or "ZipFile" not in properties["Code"] or properties["Code"]["ZipFile"] is None):
        properties["Code"] = store_path
elif package_type == IMAGE:
    if resource_type == SamBaseProvider.LAMBDA_FUNCTION:
        properties["Code"] = built_artifacts[logical_id]
    elif resource_type == SamBaseProvider.SERVERLESS_FUNCTION:
        properties["ImageUri"] = built_artifacts[logical_id]
```

qingchm: will probably not refactor it this way because we have another resource type check for layers, would probably make all conditions look parallel!

samcli/lib/build/app_builder.py
qingchm: thanks ;) separated the conditions because they are two different cases, but I guess merging them is easy to understand as well

samcli/lib/build/app_builder.py
```diff
@@ -216,20 +216,27 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
 
             resource_type = resource.get("Type")
             properties = resource.setdefault("Properties", {})
+            package_type = properties.get("PackageType", ZIP)
```
jfuss, line 219: Let's move this back to be inline. I understand this is duplicated throughout but package_type is not on all resources we update here. The way this is written suggests it could also be applied to Layers (which is does not).

samcli/lib/build/app_builder.py
```diff
@@ -216,20 +216,27 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
...
 
-            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["CodeUri"] = store_path
+            if resource_type == SamBaseProvider.SERVERLESS_FUNCTION and package_type == ZIP:
+                if "InlineCode" not in properties or properties["InlineCode"] is None:
```
jfuss, line 222: Why do we need to check for InlineCode here? On line 201, we check to see if the function was built, so assuming that we don't build any functions with InlineCode, this is would never execute right?

samcli/lib/build/app_builder.py
```diff
@@ -216,20 +216,27 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
...
+                    properties["CodeUri"] = store_path
 
-            if resource_type == SamBaseProvider.LAMBDA_FUNCTION and properties.get("PackageType", ZIP) == ZIP:
-                properties["Code"] = store_path
+            if resource_type == SamBaseProvider.LAMBDA_FUNCTION and package_type == ZIP:
+                if (
+                    "Code" not in properties
+                    or "ZipFile" not in properties["Code"]
+                    or properties["Code"]["ZipFile"] is None
```
jfuss, line 229: Same here

samcli/lib/providers/provider.py
```diff
@@ -57,10 +59,13 @@ def __init__(self):
         self.result = {"Function": [], "Layer": []}
 
     def add_function(self, function):
+        if function.inlinecode and function.inlinecode is not None:
+            return
         self.result.get("Function").append(function)
```
jfuss, line 64: This is used for much more that build. Why are we adjusting this, vs when we create the list of build only?

samcli/lib/providers/provider.py
```diff
@@ -57,10 +59,13 @@ def __init__(self):
...
 
     def add_functions(self, functions):
-        self.result.get("Function").extend(functions)
+        for function in functions:
```
jfuss, line 67: Why was this changed?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -160,20 +160,29 @@ def _convert_sam_function_resource(name, resource_properties, layers, ignore_cod
...
-
         codeuri = SamFunctionProvider.DEFAULT_CODEURI
+        inlinecode = None
         imageuri = None
         packagetype = resource_properties.get("PackageType", ZIP)
         if packagetype == ZIP:
-            codeuri = SamFunctionProvider._extract_sam_function_codeuri(
-                name, resource_properties, "CodeUri", ignore_code_extraction_warnings=ignore_code_extraction_warnings
-            )
-            LOG.debug("Found Serverless function with name='%s' and CodeUri='%s'", name, codeuri)
```
jfuss, line 171: We loose this debug statement for Functions with CodeUri.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -160,20 +160,29 @@ def _convert_sam_function_resource(name, resource_properties, layers, ignore_cod
...
+            if "InlineCode" in resource_properties and resource_properties["InlineCode"] is not None:
```
jfuss, line 168: It would be much simplier to just attempt to get InlineCode. If the dict doesn't have InlineCode, it will automatically be set to `None`. Which reduces the complexity here.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -198,20 +207,27 @@ def _convert_lambda_function_resource(name, resource_properties, layers):  # pyl
...
             LOG.debug("Found Lambda function with name='%s' and Imageuri='%s'", name, imageuri)
 
-        return SamFunctionProvider._build_function_configuration(name, codeuri, resource_properties, layers, imageuri)
+        return SamFunctionProvider._build_function_configuration(
+            name, codeuri, resource_properties, layers, inlinecode, imageuri
+        )
 
     @staticmethod
-    def _build_function_configuration(name, codeuri, resource_properties, layers, imageuri=None):
+    def _build_function_configuration(name, codeuri, resource_properties, layers, inlinecode=None, imageuri=None):
```
jfuss, line 230: If `inlinecode` is optional, should `codeuri` be optional as well?

Not sure why we are inconsistent with our naming here but I suggest to us `incline_code` rather than `inclinecode`. It is easier to read with underscores.

jfuss (changes requested):
Can you add integration tests for these cases? This will help keep our integrations tests up to date and avoid regressions related to building with InlineCode.

samcli/lib/build/app_builder.py
mgrandis: We should maybe move the layers code before this line then:
```python
    if resource_type in [SamBaseProvider.SERVERLESS_LAYER, SamBaseProvider.LAMBDA_LAYER]:
        properties["ContentUri"] = store_path
        continue
```
it's weird that it is in the middle of the others

samcli/lib/providers/sam_function_provider.py
qingchm: @jfuss this was named without underscore to match the other property names in provider.py

samcli/lib/build/app_builder.py
qingchm: @jfuss you're right, other changes are sufficient, verified that this check is unnecessary

samcli/lib/build/app_builder.py
qingchm: same here

samcli/lib/providers/sam_function_provider.py
qingchm: Added this back

qingchm:
> Can you add integration tests for these cases? This will help keep our integrations tests up to date and avoid regressions related to building with InlineCode.

Hey I added another commit with new integration test added

mgrandis (approved):

samcli/lib/providers/sam_function_provider.py
```diff
@@ -198,20 +208,31 @@ def _convert_lambda_function_resource(name, resource_properties, layers):  # pyl
...
-
         if packagetype == ZIP:
-            codeuri = SamFunctionProvider._extract_lambda_function_code(resource_properties, "Code")
-            LOG.debug("Found Lambda function with name='%s' and CodeUri='%s'", name, codeuri)
+            if (
+                "Code" in resource_properties
+                and isinstance(resource_properties["Code"], dict)
+                and resource_properties["Code"].get("ZipFile") is not None
+            ):
+                inlinecode = resource_properties["Code"]["ZipFile"]
```
aahung, line 220: It is interesting `InlineCode` can also be a zip file, just learned that from the doc

samcli/lib/providers/sam_function_provider.py
```diff
@@ -160,20 +160,30 @@ def _convert_sam_function_resource(name, resource_properties, layers, ignore_cod
...
         codeuri = SamFunctionProvider.DEFAULT_CODEURI
+        inlinecode = None
         imageuri = None
         packagetype = resource_properties.get("PackageType", ZIP)
         if packagetype == ZIP:
-            codeuri = SamFunctionProvider._extract_sam_function_codeuri(
-                name, resource_properties, "CodeUri", ignore_code_extraction_warnings=ignore_code_extraction_warnings
-            )
-            LOG.debug("Found Serverless function with name='%s' and CodeUri='%s'", name, codeuri)
+            if resource_properties.get("InlineCode") is not None:
```
aahung, line 168: nit: I don't think we need to explicitly compare to `is not None` here

```suggestion
            if resource_properties.get("InlineCode"):
```

samcli/lib/providers/sam_function_provider.py
aahung: Yeah I also agree we should keep these two consistent. We can in fact remove all "=None" from the definition because there is no use of `_build_function_configuration()` that do not provide all arguments.
```suggestion
    def _build_function_configuration(name, codeuri, resource_properties, layers, inlinecode, imageuri):
```

And in the future (if you don't want to do it now), we can add types to them to indicate whether they are optional or not:

```py
def _build_function_configuration(name: str, codeuri: Optional[str], resource_properties: Dict, layers: List[<sometype>], inlinecode: Optional[str], imageuri: Optional[str]):
```

qingchm: Changed it to behave the same way for all three fields

samcli/lib/providers/provider.py
qingchm: Changed the place for this check

jfuss (approved):
LGTM

PR #2560, opened 2021-01-28
feat: use Lambda base images for build images

hoffa:
### Which issue(s) does this change fix?

N/A

### Why is this change necessary?

Follows https://github.com/aws/aws-sam-cli/pull/2561:

- Smaller: faster pulls and less storage. Significant savings at scale.
- More maintainabile: no multi-layer/YUM hacks, base image consistent with `PackageType: Image`.
- Gets rid of dependency on emulation images in DockerHub.

### How does it address the issue?

Uses [Lambda base images](https://github.com/aws/aws-lambda-base-images).

### Comparison (January 2021)

Sizes in GB:

| Old | New      | Old Size | New Size | Change |
|-----------|---|---------------|---------------|--------|
| `amazon/aws-sam-cli-emulation-image-nodejs10.x` | `public.ecr.aws/lambda/nodejs:10` | 1.89          | 1.27          | -33%   |
| `amazon/aws-sam-cli-emulation-image-nodejs12.x` | `public.ecr.aws/lambda/nodejs:12` | 1.89          | 1.31          | -31%   |
| `amazon/aws-sam-cli-emulation-image-java8`  | `public.ecr.aws/lambda/java:8`     | 1.68          | 1.51          | -10%   |
| `amazon/aws-sam-cli-emulation-image-java8.al2` | `public.ecr.aws/lambda/java:8.al2`  | 2.22          | 1.64          | -26%   |
| `amazon/aws-sam-cli-emulation-image-java11` | `public.ecr.aws/lambda/java:11`    | 2.11          | 1.44          | -32%   |
| `amazon/aws-sam-cli-emulation-image-python2.7` | `public.ecr.aws/lambda/python:2.7`  | 1.47          | 1.42          | -3%    |
| `amazon/aws-sam-cli-emulation-image-python3.6` | `public.ecr.aws/lambda/python:3.6`  | 1.59          | 1.54          | -3%    |
| `amazon/aws-sam-cli-emulation-image-python3.7` | `public.ecr.aws/lambda/python:3.7`  | 1.63          | 1.58          | -3%    |
| `amazon/aws-sam-cli-emulation-image-python3.8` | `public.ecr.aws/lambda/python:3.8`  | 2.17          | 1.41          | -35%   |
| `amazon/aws-sam-cli-emulation-image-ruby2.5` | `public.ecr.aws/lambda/ruby:2.5`   | 1.65          | 1.61          | -2%    |
[...]

PR #2561, opened 2021-01-28
feat: improve build images

hoffa:

N/A

Similar to #2560, but only makes changes not affecting base images.

- Use fewer layers where sensible (shaves off about 50 MB from every image).
- Fix Python 2.7 build image.
- Fail build script if any build fails.
- Remove `curl -k`s.

None.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


build-image-src/Dockerfile-python27
```diff
@@ -56,6 +55,6 @@ ENV LANG=en_US.UTF-8
 # Python for it to be picked up during `sam build`
 RUN pip3 install wheel
 
-RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && rm get-pip.py
```
awood45, line 59: Has this been tested today? Making sure this isn't impacted by the recent `pip` deprecation (which frankly might be a broader issue for building Python 2.7 going forward beyond our scope).

awood45 (approved):
This generally looks great! One testing question

build-image-src/Dockerfile-java8-al2
```diff
@@ -5,8 +5,8 @@ FROM amazonlinux:2
 COPY --from=0 / /rootfs
 
 # Installing by yum at copied location
-RUN yum groupinstall -y development --installroot=/rootfs && yum clean all
```
sriram-mv, line 8: why are we removing `yum clean all` ? is it because there is not any cache to clear?

sriram-mv (approved):
Looks good to me, I had one question.

build-image-src/Dockerfile-python27
hoffa: I've tested it builds, `pip` links to the Python 2 `pip`, and that a dummy `pip install` works fine.

build-image-src/Dockerfile-java8-al2
hoffa: It's redundant as it's run L27 in the same layer.

build-image-src/Dockerfile-python27
hoffa: And for reference, that was added fairly recently: https://github.com/pypa/get-pip/pull/81

The homepage hasn't yet been updated, but refers to the same method for Python 2.6: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py

Also https://stackoverflow.com/a/65866547

jfuss (approved):
LGTM

PR #2562, opened 2021-01-28
Fix python command for installing pip and requirements in Makefile-co…

hawflau:
It fixes failing integration test `TestBuildCommand_ProvidedFunctions` for the container builds.
In `Makefile-container`, it was using `python` command, which is Python2, to install pip. As [pip has dropped support for Python2.7](https://pip.pypa.io/en/stable/news/#id1), the installation doesn't work anymore.

change `python` to `python3`

Not that I'm aware of

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/testdata/buildcmd/Provided/Makefile-container
+++ b/tests/integration/testdata/buildcmd/Provided/Makefile-container
@@ -2,6 +2,6 @@ build-Function:
 	cp *.py $(ARTIFACTS_DIR)
 	cp requirements.txt $(ARTIFACTS_DIR)
 	curl https://bootstrap.pypa.io/get-pip.py > /tmp/get-pip.py
-	python /tmp/get-pip.py
-	python -m pip install -r requirements.txt -t $(ARTIFACTS_DIR)
+	python3 /tmp/get-pip.py
+	python3 -m pip install -r requirements.txt -t $(ARTIFACTS_DIR)
 	rm -rf $(ARTIFACTS_DIR)/bin


sriram-mv (approved):
LGTM

aahung (approved):

tests/integration/testdata/buildcmd/Provided/Makefile-container
```diff
@@ -2,6 +2,6 @@ build-Function:
 	cp *.py $(ARTIFACTS_DIR)
 	cp requirements.txt $(ARTIFACTS_DIR)
 	curl https://bootstrap.pypa.io/get-pip.py > /tmp/get-pip.py
```
hoffa, line 4: Shouldn't this be changed to `https://bootstrap.pypa.io/2.7/get-pip.py` instead?

See https://github.com/pypa/get-pip/pull/81 and the similar instruction for Python 2.6: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py

hawflau: @aahung and I had a discussion about this. We think it's safer to assume `python3` exists than to assume `python` always runs the Python2 Interpreter. 
If we change to https://bootstrap.pypa.io/2.7/get-pip.py, we need to assume `python` is always running Python2

hawflau: in fact, we verified that both work

hoffa: I see, if this isn't testing anything related to Python 2 then it seems fine.

hoffa (approved):

tests/integration/testdata/buildcmd/Provided/Makefile-container
hawflau: yeah, this one is not related to Python2.
We also need to fix py2.7 build image though.

hoffa: ^ #2561

PR #2563, opened 2021-01-29
feat: new version notifier

mndeveci:
This feature will inform users if there is a newer version of SAM CLI available.

As of now, there is no way to inform customers if there is a newer version of SAM CLI available.

This compares current installed version with up to date one, and informs user if there is a newer version available. This check only runs once in a week for all commands except `sam --version` which will check everytime.

Whenever it runs, it might add slight delay to the command (expected less than a second). Even with the failures during check, it will update last verification time, so that it won't run every command in case host has no internet access.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+    func: function reference
+        Actual function (command) which will be executed
+
+    Returns
+    -------
+    function reference:
+        A wrapped function reference which executes original function and checks newer version of SAM CLI
+    """
+
+    def wrapped(*args, **kwargs):
```
CoshUS, line 37: Nit: Maybe add functools.wraps decorator.

CoshUS (approved):

samcli/cli/main.py
```diff
@@ -58,6 +59,18 @@ def print_info(ctx, param, value):
...
 
 
+def print_version_option(ctx, param, value):
+    """
+    Since default implementation for click.version_option doesn't support callback's we need to create a simple version
+    of it, which will print version information, and check for new version and exit the cli
+    """
+    if not value or ctx.resilient_parsing:
+        return
+    click.echo("SAM CLI, version %(version)s" % {"version": __version__}, color=ctx.color)
```
mgrandis, line 69: minor: Use `f"SAM CLI, version {__version__}"` formatting?

wchengru (approved):

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+        actual_result = func(*args, **kwargs)
+        # check and inform newer version if it is available
+        inform_newer_version()
+
+        return actual_result
+
+    return wrapped
+
+
+def inform_newer_version(check_all_always=False):
```
mgrandis, line 48: I think we usually use the word `force` for this kind of parameter.
Something like: `force_check`

samcli/cli/main.py
```diff
@@ -58,6 +59,18 @@ def print_info(ctx, param, value):
...
+    inform_newer_version(True)
```
mgrandis, line 70: minor: You should add the name of the parameter to make it more obvious what is `True`

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Parameters
+    ----------
+    check_all_always: bool
+        When it is True, it will trigger checking new version of SAM CLI. Default value is False
+
+    """
+    # run everything else in try-except block
+    global_config = None
+    need_to_update_last_check_time = True
```
mgrandis, line 64: I would start with `False` here and then update it to `True` after the call to `compare_current_version` and remove the `else`.

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+        GlobalConfig object that have been read
+    """
+    try:
+        if global_config:
+            global_config.last_version_check = datetime.utcnow().timestamp()
+    except Exception as e:
+        LOG.debug("Updating last version check time was failed", exc_info=e)
+
+
+def is_last_check_older_then_week(last_version_check):
```
mgrandis, line 108: Should we rename to `is_last_check_older_than_delta` or `is_version_check_overdue`?
The function does not reference a week directly but `DELTA_DAYS` which could be any amount of days.

samcli/lib/utils/version_checker.py
mndeveci: Oops that is a type there, will update the param name, thanks!

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            need_to_update_last_check_time = False
+    except Exception as e:
+        LOG.debug("New version check failed", exc_info=e)
+    finally:
+        if need_to_update_last_check_time:
+            update_last_check_time(global_config)
+
+
+def compare_current_version():
```
mgrandis, line 80: I would rename this to `get_and_compare_versions`

mndeveci: I would agree with this one, if we were returning something from this command, but it only outputs the difference.
Updating it as `fetch_and_compare_versions` as we discussed offline

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    Compare current up to date version with the installed one, and inform if a newer version available
+    """
+    response = get(AWS_SAM_CLI_PYPI_ENDPOINT, timeout=PYPI_CALL_TIMEOUT_IN_SECONDS)
+    result = response.json()
+    current_version = result.get("info", {}).get("version", None)
```
mgrandis, line 86: I would rename `current_version` to `latest_version`

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+    Update last_check_time in GlobalConfig
+    Parameters
+    ----------
+    global_config: GlobalConfig
+        GlobalConfig object that have been read
+    """
+    try:
+        if global_config:
+            global_config.last_version_check = datetime.utcnow().timestamp()
+    except Exception as e:
```
hoffa, line 106: When does this fail?

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Compare current up to date version with the installed one, and inform if a newer version available
+    """
+    response = get(AWS_SAM_CLI_PYPI_ENDPOINT, timeout=PYPI_CALL_TIMEOUT_IN_SECONDS)
+    result = response.json()
+    latest_version = result.get("info", {}).get("version", None)
+    LOG.debug("Installed version %s, current version %s", installed_version, latest_version)
+    if latest_version and installed_version != latest_version:
+        click.secho(f"\nSAM CLI update available ({latest_version}); ({installed_version} installed)", fg="green")
+        click.echo(f"To download: {AWS_SAM_CLI_INSTALL_DOCS}")
```
hoffa, line 92: UI: first line has a blank newline; add a `\n` to the end as well so the message is visually symmetrical? (Assuming this currently means there's only one blank newline at the top)

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+        GlobalConfig object that have been read
+    """
+    try:
+        if global_config:
+            global_config.last_version_check = datetime.utcnow().timestamp()
+    except Exception as e:
+        LOG.debug("Updating last version check time was failed", exc_info=e)
+
+
+def is_last_check_older_than_delta(last_version_check):
```
hoffa, line 110: Can we add type hints?

samcli/cli/main.py
```diff
@@ -58,6 +59,18 @@ def print_info(ctx, param, value):
...
 
+def print_version_option(ctx, param, value):
+    """
+    Since default implementation for click.version_option doesn't support callback's we need to create a simple version
+    of it, which will print version information, and check for new version and exit the cli
+    """
+    if not value or ctx.resilient_parsing:
+        return
+    click.echo(f"SAM CLI, version {__version__}", color=ctx.color)
+    inform_newer_version(force_check=True)
```
hoffa, line 70: One thing to keep in mind is that `--version` (for pretty much every command) is often used as a dead-simple, immediate sanity check (e.g. in CI/CD installing SAM CLI and running `sam --version` to make sure it doesn't return a zero exit code). I don't think that forcing network operations at every call is sensible, as people will often except immediate feedback, and the implicit assumption seems to be that `--version` barely prints a constant and exits.

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+    ----------
+    last_version_check: epoch time
+        last_version_check epoch time read from GlobalConfig
+
+    Returns
+    -------
+    bool:
+        True if last_version_check is None or older then a week, False otherwise
+    """
+    if last_version_check is None or type(last_version_check) not in [int, float]:
```
hoffa, line 124: When would the type be an `int` versus a `float`?

samcli/lib/utils/version_checker.py
mndeveci: It updates value and writes it to the config file which is kept under `~/.aws-sam/metadata.json`. For any reason, I don't want this to fail original command with a non-zero exit code.

samcli/lib/utils/version_checker.py
mndeveci: I added that new line to separate this output from the command's output. I couldn't understand what you suggested, you mean adding new line at the end of line 92 or line 91?

samcli/cli/main.py
mndeveci: This was coming from UX team as a suggestion, if we know that is been used in this way, I agree to remove this force check for `--version`.

samcli/lib/utils/version_checker.py
mndeveci: Again just been defensive, in case somebody edits file and puts some invalid data (or somehow file is corrupted), I don't want original command to fail.

samcli/lib/utils/version_checker.py
hoffa: Meant end of L92. But if this message is always shown at the very end, then no need. If there's output above and below, then it looks nicer if the block of text has same spacing on both sides.

samcli/lib/utils/version_checker.py
hoffa: Oh okay I see, interesting. So what looks like a simple assignment actually does I/O.

mndeveci: Yeah, the setter is overridden inside `GlobalConfig` and it is doing all this magic there

samcli/lib/utils/version_checker.py
mndeveci: Yeah this is the last output from the command, and it exits afterwards.

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+def fetch_and_compare_versions():
+    """
+    Compare current up to date version with the installed one, and inform if a newer version available
+    """
+    response = get(AWS_SAM_CLI_PYPI_ENDPOINT, timeout=PYPI_CALL_TIMEOUT_IN_SECONDS)
+    result = response.json()
+    latest_version = result.get("info", {}).get("version", None)
+    LOG.debug("Installed version %s, current version %s", installed_version, latest_version)
+    if latest_version and installed_version != latest_version:
+        click.secho(f"\nSAM CLI update available ({latest_version}); ({installed_version} installed)", fg="green")
```
sriram-mv, line 91: something to think here, should this go to stderr or stdout? Reasoning for this: users might be scraping the output directly today.

sriram-mv (changes requested):
Relatively minor comments, overall this looks good!

samcli/cli/main.py
```diff
@@ -19,6 +19,7 @@
 from .context import Context
 from .command import BaseCommand
 from .global_config import GlobalConfig
+from ..lib.utils.version_checker import inform_newer_version
```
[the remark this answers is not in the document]
mndeveci: I checked other files but it is imported with relative path on this one, will update it.

samcli/lib/utils/version_checker.py
```diff
@@ -0,0 +1,128 @@
...
+from functools import wraps
+
+import click
+from requests import get
+from samcli import __version__ as installed_version
+from samcli.cli.global_config import GlobalConfig
+
+LOG = logging.getLogger(__name__)
+
+AWS_SAM_CLI_PYPI_ENDPOINT = "https://pypi.org/pypi/aws-sam-cli/json"
```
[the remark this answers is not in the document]
mndeveci: I think they should stay wherever they are used. Otherwise it will be a huge file with lots of constants in it.

samcli/lib/utils/version_checker.py
mndeveci: We are using `click` for formatting purposes, and I verified that it is printing into `stdout`. I don't know if there is standard for this but it feels like it should be printed into `stdout`. `stderr` should be used actual error situations as the name suggest.

tests/unit/lib/utils/test_version_checker.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def test_last_check_time_none_should_return_true(self):
+        self.assertTrue(is_version_check_overdue(None))
+
+    def test_last_check_time_week_older_should_return_true(self):
+        eight_days_ago = datetime.utcnow() - timedelta(days=8)
+        self.assertTrue(is_version_check_overdue(eight_days_ago))
+
+    def test_last_check_time_week_earlier_should_return_false(self):
+        eight_days_ago = datetime.utcnow() - timedelta(days=6)
```
mgrandis, line 140: Looks like 6 days ago to me :)

mndeveci: Damn copy paste, will update it :)

sriram-mv (approved):

mgrandis (approved):

samcli/cli/main.py
mndeveci: Agreed that it will create too much burden for a command which is usually used as sanity check.

hoffa (approved):

PR #2564, opened 2021-01-29
chore: Add type hint to build command related modules

aahung:

Add full type hints to selected modules, which brings the following advantages:
- type safe
- help code comprehension
- better IDE experience (e.g., auto-completion)

0, it does not change the code behavior, type check is enforced in CI only, no runtime overheads.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -25,20 +26,20 @@ class BuildContext:
...
+        build_dir: str,
+        cache_dir: str,
+        cached: bool,
+        mode: Optional[str],
+        manifest_path: Optional[str],
+        clean: bool,
+        use_container: bool,
+        parameter_overrides: dict,
+        docker_network: Optional[str],
+        skip_pull_image: bool,
```
aahung, line 41: The main motivation to remove them is because `parameter_overrides`'s default value here is not correct. The default value is `{}`:
https://github.com/aws/aws-sam-cli/blob/eefcd0744fd90323da31232504b87449629e10eb/samcli/commands/_utils/options.py#L251-L260

So I remove them all default args here because the default value are defined in `@click...` somewhere already, defining here requires us to maintain two sources of truth. Also we always provide full arguments when invoking this funtion, so there is no need to have default arguments.

samcli/commands/build/build_context.py
```diff
@@ -25,20 +26,23 @@ class BuildContext:
...
+        cache_dir: str,
+        cached: bool,
+        mode: Optional[str],
+        manifest_path: Optional[str] = None,
+        clean: bool = False,
+        use_container: bool = False,
+        # pylint: disable=fixme
+        # FIXME: parameter_overrides is never None, we should change this to "dict" from Optional[dict]
+        # See samcli/commands/_utils/options.py:251 for its all possible values
+        parameter_overrides: Optional[dict] = None,
```
elbayaaa, line 42: what is the reason for not applying the fix now?

samcli/commands/build/build_context.py
```diff
@@ -110,62 +113,67 @@ def _setup_build_dir(build_dir, clean):
...
-        return self._container_manager
+    def container_manager(self) -> ContainerManager:
+        # Note(xinhol): self._container_manager will be assigned with a value if it is None in __enter__()
+        return self._container_manager  # type: ignore
 
     @property
-    def function_provider(self):
-        return self._function_provider
+    def function_provider(self) -> SamFunctionProvider:
+        # Note(xinhol): self._container_manager will be assigned with a value if it is None in __enter__()
```
elbayaaa, line 122: typo in the comment? _function_provider instead of _container_manager?

elbayaaa: Also, what is the value of the comment in general?

samcli/commands/build/build_context.py
```diff
@@ -110,62 +113,67 @@ def _setup_build_dir(build_dir, clean):
...
+        return self._function_provider  # type: ignore
 
     @property
-    def layer_provider(self):
-        return self._layer_provider
+    def layer_provider(self) -> SamLayerProvider:
+        # Note(xinhol): self._container_manager will be assigned with a value if it is None in __enter__()
```
elbayaaa, line 127: same

samcli/commands/build/build_context.py
aahung: Oh we want to keep this PR small and make the code review easier. This PR only focuses on adding types without behavior change. So we can merge it as quick as possible. I will create a follow up for those fixmes.

samcli/commands/build/build_context.py
aahung: good catch!

samcli/lib/build/app_builder.py
```diff
@@ -253,8 +264,8 @@ def _build_lambda_image(self, function_name, metadata):
 
         LOG.info("Building image for %s function", function_name)
 
-        dockerfile = metadata.get("Dockerfile")
-        docker_context = metadata.get("DockerContext")
+        dockerfile = cast(str, metadata.get("Dockerfile"))
```
elbayaaa, line 267: rename to `docker_file` for consistency?

samcli/lib/build/workflow_config.py
```diff
@@ -116,31 +121,28 @@ def get_selector(selector_list, identifiers, specified_workflow=None):
...
     try:
         # Find first non-None selector.
         # Return the first selector with a match.
-        selector = next(_selector for _selector in selectors if _selector)
+        return next(_selector for _selector in selectors if _selector)
     except StopIteration:
         pass
 
-    return selector
+    return None
```
elbayaaa, line 142: I prefer the previous version (return selector), is there a reason to change it?

samcli/lib/build/workflow_config.py
```diff
@@ -251,17 +255,18 @@ def get_workflow_config(runtime, code_dir, project_dir, specified_workflow=None)
             identifiers=[specified_workflow, runtime],
             specified_workflow=specified_workflow,
         )
+        assert selector
```
elbayaaa, line 258: Is this a common practice to use assert inside a production code?

samcli/lib/build/workflow_config.py
aahung: two reasons:
1. I don't need to annotate `selector = None` to `selector: Optional[.....] = None`
2. return inside try matches the comment "Return the first selector with a match."

samcli/lib/build/workflow_config.py
```diff
@@ -305,13 +310,13 @@ class BasicWorkflowSelector:
...
-    def __init__(self, configs):
-        if not isinstance(configs, list):
-            configs = [configs]
-
-        self.configs = configs
+    def __init__(self, configs: Union[CONFIG, List[CONFIG]]) -> None:
+        if isinstance(configs, list):
+            self.configs = configs
+        else:
+            self.configs = [configs]
```
elbayaaa, line 317: [Minor] `self.configs = configs if isinstance(configs, list) else [config]` 

Also, is it `list` of `List` ?

samcli/lib/build/app_builder.py
aahung: I think `docker_context` was named `docker_context` instead of `dockercontext` because the `C` was uppercase, which does not apply to `Dockerfile`, which seems to be a whole word.

samcli/lib/build/workflow_config.py
aahung: absolutely not, good catch!

samcli/lib/build/workflow_config.py
aahung: `list` and `List` both work, to keep it consistent, let's use `List` 👍 

```py
isinstance([], typing.List)
True
isinstance([], list)
True
```

aahung: FYI: `List = _alias(list, T, inst=False)`

elbayaaa:
Would you add details about how the changes are tested and what will be the behavior if the value assigned to a variable at runtime is different than the value specified in the type hint?

aahung:
> Would you add details about how the changes are tested 

I use `make pr` to test. There is no behavior changes mostly, except for a few new exception raise. So testing is mostly sanity checking.

> and what will be the behavior if the value assigned to a variable at runtime is different than the value specified in the type hint?

Type hints have no effect during the runtime. type check is done by 3rd party tools like mypy (we are using it).

> The Python runtime does not enforce function and variable type annotations. They can be used by third party tools such as type checkers, IDEs, linters, etc. https://docs.python.org/3/library/typing.html

samcli/lib/build/workflow_config.py
aahung: I had to revert this 😢 pylint cannot recognize this alias

> W1116: Second argument of isinstance is not a type (isinstance-second-argument-not-valid-type)

samcli/lib/build/workflow_config.py
```diff
@@ -251,17 +255,19 @@ def get_workflow_config(runtime, code_dir, project_dir, specified_workflow=None)
...
             specified_workflow=specified_workflow,
         )
+        if not selector:
+            raise RuntimeError(f"Cannot find workflow selector for {specified_workflow} or {runtime}")
 
         # Identify workflow configuration from the workflow selector.
         config = selector.get_config(code_dir, project_dir)
         return config
-    except ValueError as ex:
+    except (ValueError, AssertionError) as ex:
```
elbayaaa, line 264: need to remove the "AssertionError" exception?

elbayaaa (approved):

samcli/commands/build/build_context.py
```diff
@@ -110,62 +113,67 @@ def _setup_build_dir(build_dir, clean):
...
-        return self._function_provider
+    def function_provider(self) -> SamFunctionProvider:
+        # Note(xinhol): self._function_provider will be assigned with a value if it is None in __enter__()
+        return self._function_provider  # type: ignore
 
     @property
-    def layer_provider(self):
-        return self._layer_provider
+    def layer_provider(self) -> SamLayerProvider:
+        # Note(xinhol): self._layer_provider will be assigned with a value if it is None in __enter__()
```
hoffa, line 127: What value? Where is this documented?

samcli/lib/build/app_builder.py
```diff
@@ -383,6 +403,10 @@ def _build_function(  # pylint: disable=R1710
             Path to the location where built artifacts are available
         """
         if packagetype == IMAGE:
+            if not metadata:
+                raise MissingMetadataForImageFunctionException(
+                    f"{function_name} has Image package type but Metadata is missing."
+                )
```
hoffa, line 409: Why does this have to be in this PR? Would make reviewing easier if there was no logic changes.

samcli/lib/build/app_builder.py
```diff
@@ -572,4 +611,4 @@ def _parse_builder_response(stdout_data, image_name):
             LOG.debug("Builder crashed")
             raise ValueError(msg)
 
-        return response
+        return cast(dict, response)
```
hoffa, line 614: nit: `Dict`

samcli/lib/build/build_graph.py
```diff
@@ -28,7 +30,9 @@
 LAYER_FIELD = "layer"
 
 
-def _function_build_definition_to_toml_table(function_build_definition):
+def _function_build_definition_to_toml_table(
+    function_build_definition: "FunctionBuildDefinition",
+) -> "tomlkit.items.Table":
```
hoffa, line 35: Can use class.

samcli/lib/build/build_graph.py
```diff
@@ -28,7 +30,9 @@
 LAYER_FIELD = "layer"
 
 
-def _function_build_definition_to_toml_table(function_build_definition):
+def _function_build_definition_to_toml_table(
+    function_build_definition: "FunctionBuildDefinition",
```
hoffa, line 34: Any reason to write type as string when there's no forward reference?

samcli/lib/build/workflow_config.py
```diff
@@ -116,31 +121,28 @@ def get_selector(selector_list, identifiers, specified_workflow=None):
     """
 
     # Create a combined view of all the selectors
-    all_selectors = {}
+    all_selectors: Dict[str, Union["BasicWorkflowSelector", "ManifestWorkflowSelector"]] = dict()
```
hoffa, line 124: Would it make sense to use type aliases for repeated types? e.g. `Union["BasicWorkflowSelector", "ManifestWorkflowSelector"]`?

samcli/lib/build/workflow_config.py
hoffa: Why the logic change?

hoffa (approved):

PR #2567, opened 2021-01-29
ci(package):  NamedTemporaryFile Being Opened Twice

CoshUS:
In Windows NT or later, NamedTemporaryFile cannot be opened twice.
https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/package/test_package_command_image.py
```diff
@@ -131,6 +131,10 @@ def test_package_template_with_image_function_in_nested_application(self, templa
         # when image function is not in main template, erc_repo_name does not show up in stdout
         # here we download the nested application template file and verify its content
         with tempfile.NamedTemporaryFile() as packaged_file, tempfile.TemporaryFile() as packaged_nested_file:
+            # https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
+            # Closes the NamedTemporaryFile as on Windows NT or later, NamedTemporaryFile cannot be opened twice.
+            packaged_file.close()
```
jfuss, line 136: Why not use `TemporaryFile()` for `package_file`?

hawflau (approved):
LGTM!

tests/integration/package/test_package_command_image.py
CoshUS: TemporaryFile does not guarantee the file is visible in the file system on Unix. We need to read the file later in order to validate the output.

jfuss (approved):

tests/integration/package/test_package_command_image.py
hawflau: We need a visible name for `package_file` in the file system, which is used for `output_template_file` in the command. 

`TemporaryFile()` does not guarantee a visible name in the file system.
We don't need the file to actually exist though. Since the command will create the file again

PR #2569, opened 2021-01-29
chore: Raise InvalidLocationError when init's --location is not valid

aahung:

This provides more helpful message when user run commands like

```
# URL exists but not a valid repo
$ sam init --location http://example.com
Error: The template location specified is not valid: http://example.com
# URL does not exist
$ sam init --location http://example.com
Error: The template location specified is not valid: http://example.comaa
```

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/init/init_generator.py
```diff
@@ -4,11 +4,11 @@
...
 from samcli.commands.exceptions import UserException
 from samcli.lib.init import generate_project
-from samcli.lib.init.exceptions import GenerateProjectFailedError, ArbitraryProjectDownloadFailed
+from samcli.lib.init.exceptions import InitErrorException
 
 
 def do_generate(location, package_type, runtime, dependency_manager, output_dir, name, no_input, extra_context):
     try:
         generate_project(location, package_type, runtime, dependency_manager, output_dir, name, no_input, extra_context)
-    except (GenerateProjectFailedError, ArbitraryProjectDownloadFailed) as e:
```
aahung, line 13: `InitErrorException` only has two subclasses, GenerateProjectFailedError, ArbitraryProjectDownloadFailed

hoffa: Is this related to the `InvalidLocationError`?

aahung: `InvalidLocationError` inherents `InitErrorException` as well. So I intended to add `InvalidLocationError` to the list, which is equivalent to a single  `InitErrorException`

samcli/lib/init/exceptions.py
```diff
@@ -16,5 +16,9 @@ class GenerateProjectFailedError(InitErrorException):
     fmt = "An error occurred while generating this project {project}: {provider_error}"
 
 
+class InvalidLocationError(InitErrorException):
+    fmt = "The template location specified is not valid: {template}"
```
sriram-mv, line 20: nit: since the class name has invalid, we could use the same in the message too.

```
The template location specified is invalid
```

sriram-mv (approved):

hoffa (approved):

c2tarun (approved):

Issue #2195, opened 2020-09-01
Support --acl parameter for sam package


kowalski:
It would be very helpful if `sam package` command accepted `--acl` parameter. It would be passed on to uploads to S3 done by the client (similarly how `--metadata` works).
This would allow to work with S3 buckets that require specific ACL settings. Currently we need to resort to some tedious workarounds (like allowing more permissive policy, iterating over s3 key under prefix and chaning permissions with aws s3api put-object-acl.

Would be so great not to have to do it this way.

faboulaye:
Hi
Same issue I'm facing, it will be helpful to have this feature so we can package our resources and send it to our production buckets as artifacts.

celestialorb:
Created a basic PR to try and address this as this is a feature we require as well.

kowalski:
@celestialorb nice! I really hope this gets accepted!

PR #2577, opened 2021-02-01
add ACL parameter for package artifact uploads

celestialorb:
#2195 

There are certain environments where being able to set the ACL of the uploaded objects in S3 on package would be beneficial. In my exact scenario I am uploading to another account and publishing from that account, which requires me to set the artifact's object's ACL to the `bucket-owner-full-control` canned ACL policy.

This PR adds a new flag, very similar to the `--metadata` flag that will set the ACL header appropriately when uploading the artifacts to S3.

Should be none, as the new flag should be optional and will only have additional behavior if specified.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


tests/unit/lib/package/test_s3_uploader.py
```diff
@@ -52,6 +53,18 @@ def test_s3_uploader_artifact_metadata(self):
...
+            s3_client=self.s3,
+            bucket_name=self.bucket_name,
+            prefix=self.prefix,
+            kms_key_id=self.kms_key_id,
+            force_upload=self.force_upload,
+            no_progressbar=self.no_progressbar,
+        )
+        s3_uploader.artifact_acl = "bucket-owner-full-control"
+        self.assertEqual(s3_uploader.artifact_acl, "bucket-owner-full-control")
+
```
hawflau, line 67: I believe we also need a test case to test `s3_uploader.upload` with `acl` specified

hawflau (changes requested):
Thanks for your contribution! and sorry for the delayed response. We are reviewing the changes as there's some conflicts with the changes made from Accelerate. We will get back shortly.

samcli/commands/deploy/command.py
```diff
@@ -155,6 +155,12 @@
     help="Automatically resolve s3 bucket for non-guided deployments."
     "Do not use --s3-guided parameter with this option.",
 )
+@click.option(
+    "--acl",
+    default=None,
+    required=False,
+    help="Sets the specified canned ACL policy on uploaded artifact S3 objects."
+)
```
qingchm, line 163: Since this is shared between package and deploy have this option refactored as a shared option imported into both classes for best practice purposes! Thanks!

tests/unit/lib/package/test_s3_uploader.py
qingchm: @celestialorb Can we address this please? Thanks!

qingchm (changes requested):
Left some small comments to address, please also resolve the merge conflicts with current develop branch thanks!

jfuss:
Closing as inactive.

Happy to review an updated PR from the community, if someone wants to revive the PR

Issue #2382, opened 2020-11-15
Allow samconfig properties like `stack_name` to apply to all commands


mhart:
### Description:

EDIT: I had previously misunderstood how samconfig.toml works. I thought by providing the stack name in `sam deploy --guided`, it should use that stack name for all commands (like `sam logs`). Turns out that only applies to the `sam deploy` command specifically.

You currently have to specify `stack_name=mystack` for each command you want it to apply to – it would be nice if there would be a way for it to apply to all commands.

I also think that `sam deploy --guided` should apply settings like stack, region, etc to all commands that need them (unless they've already been set)

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS
2. `sam --version`: 1.10.0

awood45:
This actually works for me, can you share your samconfig.toml file, specifically the section you have for the “logs” command?

mhart:
What do you mean the section I have for the "logs" command?

mhart:
I did a `sam deploy --guided` to create the `samconfig.toml`, it looks like this:

```
version = 0.1
[default]
[default.deploy]
[default.deploy.parameters]
stack_name = "mystack"
s3_bucket = "mybucket"
s3_prefix = "mystack"
region = "us-east-1"
capabilities = "CAPABILITY_IAM"
```

awood45:
This is definitely something we can document in more detail - the samconfig.toml file acts as “default arguments” per method. For a farther afield example, you can create a “build” command section to turn on “use_container” for all builds.

For stack_name to be picked up for logs, you’d want to add a “logs” section to your samconfig.toml file and try again. This currently requires a manual edit.

awood45:
I’ll post my samconfig in a moment when I get to my laptop. The key is that this will only provide parameters for the “deploy” command.

mhart:
Ah. Will `stack_name` apply to all commands if I put it under `[default]`? Duplicating it for each command that needs it seems wasteful.

I can update the issue title if that's not currently possible

awood45:
I can test in a bit, but I believe at best you'd run into an issue where now every command will expect that parameter to exist and will fail if it does not, which would be a problem for build, etc.

I do think that making that a bit smarter to allow for a cross-command default is a good feature request.

mhart:
Updated title and description to make it a feature request 👍

awood45:
[...]

PR #2580, opened 2021-02-02
feat: allow setting samconfig common parameters in [<env>.global.parameters]

wchengru:
#2382 

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/cli/main.py
```diff
@@ -58,6 +59,27 @@ def print_info(ctx, param, value):
     ctx.exit()
 
 
+def log_cmdline_info(func):
+    def wrapper(*args, **kwargs):
+        if kwargs["config_file"] and kwargs["config_env"]:
```
mgrandis, line 64: Shall we test if `config_file` and `config_env` are in `kwargs` before testing their value?

We risk getting a `KeyError`.

```
if "config_file" in kwargs and "config_env" in kwargs and kwargs["config_file"] and kwargs["config_env"]
```

samcli/lib/config/samconfig.py
```diff
@@ -73,7 +74,14 @@ def get_all(self, cmd_names, section, env=DEFAULT_ENV):
         env = env or DEFAULT_ENV
 
         self._read()
-        return self.document[env][self._to_key(cmd_names)][section]
+        if isinstance(self.document, dict):
+            params = self.document.get(env, {}).get(self._to_key(cmd_names), {}).get(section, {})
```
mgrandis, line 78: You should put `self.document.get(env, {})` in a variable as you are accessing it 3 times.

samcli/lib/config/samconfig.py
```diff
@@ -106,19 +114,21 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
-            # in parent layer will override the whole child layer
-            if self.document.get(env, None):
-                if self.document[env].get(self._to_key(cmd_names), None):
-                    if not self.document[env][self._to_key(cmd_names)].get(section, None):
-                        self.document[env][self._to_key(cmd_names)].update({section: {key: value}})
+        # Empty document prepare the initial structure.
+        # self.document is a nested dict, we need to check each layer and add new tables, otherwise duplicated key
+        # in parent layer will override the whole child layer
+        if self.document.get(env, None):
+            if self.document[env].get(self._to_key(cmd_names), None):
```
mgrandis, line 121: Same here, I would put `self.document[env].get(self._to_key(cmd_names), None)` in a variable

samcli/lib/config/samconfig.py
```diff
@@ -190,6 +200,50 @@ def _write(self):
     def _version(self):
         return self.document.get(VERSION_KEY, None)
 
+    def _deduplicate_global_parameters(self, cmd_names, section, key, value, env=DEFAULT_ENV):
```
mgrandis, line 203: Are we sure we want to do this?
Potentially erasing customer values.
I think that what is expected is to be able to define global values but also be able to supersede the global value by defining one in a more specific section.

samcli/lib/config/samconfig.py
wchengru: Rewrite this part to make the if-else branches cleaner.

samcli/lib/config/samconfig.py
wchengru: done

samcli/cli/main.py
wchengru: Fixed this, thanks!

samcli/lib/config/samconfig.py
```diff
@@ -106,19 +115,26 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
+        # self.document is a nested dict, we need to check each layer and add new tables, otherwise duplicated key
+        # in parent layer will override the whole child layer
+        cmd_name_key = self._to_key(cmd_names)
+        env_content = self.document.get(env, {})
+        cmd_content = env_content.get(cmd_name_key, {})
+        param_content = cmd_content.get(section, {})
+        if env_content:
+            if cmd_content:
+                if param_content:
+                    self.document[env][cmd_name_key][section].update({key: value})
```
mgrandis, line 128: You can use `param_content` here.

samcli/lib/config/samconfig.py
```diff
@@ -106,19 +115,26 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
                 else:
-                    self.document[env].update({self._to_key(cmd_names): {section: {key: value}}})
+                    self.document[env][cmd_name_key].update({section: {key: value}})
```
mgrandis, line 130: `cmd_content` here

samcli/lib/config/samconfig.py
```diff
@@ -106,19 +115,26 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
             else:
-                self.document.update({env: {self._to_key(cmd_names): {section: {key: value}}}})
-        # Only update appropriate key value pairs within a section
-        self.document[env][self._to_key(cmd_names)][section].update({key: value})
+                self.document[env].update({cmd_name_key: {section: {key: value}}})
```
mgrandis, line 132: `env_content` here

samcli/lib/config/samconfig.py
mgrandis: I think the global section should set default values that can be overwritten in specific sections.

samcli/lib/config/samconfig.py
wchengru: updated this one and the other two below.

mgrandis (approved):

qingchm (commented):
Other then the docstring improvement everything looks good to me! Nice to see those helper functions all being tested

samcli/lib/config/samconfig.py
```diff
@@ -190,6 +204,49 @@ def _write(self):
...
+        In case the global parameters contains the same key-value with command parameters,
+        we remove the entry in command parameters
+
+        Parameters
+        ----------
+        cmd_name_key : str
+            key of command name
+
+        section : str
+            Specific section within the command to look into Ex: `parameters`
```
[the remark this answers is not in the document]
wchengru: Updated here and the other two usages. Thanks.

qingchm (approved):

samcli/lib/config/samconfig.py
```diff
@@ -99,19 +108,24 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
-                if self.document[env].get(self._to_key(cmd_names), None):
-                    if not self.document[env][self._to_key(cmd_names)].get(section, None):
-                        self.document[env][self._to_key(cmd_names)].update({section: {key: value}})
-                else:
-                    self.document[env].update({self._to_key(cmd_names): {section: {key: value}}})
-            else:
-                self.document.update({env: {self._to_key(cmd_names): {section: {key: value}}}})
-        # Only update appropriate key value pairs within a section
-        self.document[env][self._to_key(cmd_names)][section].update({key: value})
+        # Empty document prepare the initial structure.
```
hoffa, line 111: Are the `if` changes a refactor?

samcli/cli/main.py
```diff
@@ -58,6 +59,27 @@ def print_info(ctx, param, value):
     ctx.exit()
 
 
+def log_cmdline_info(func):
```
hoffa, line 62: Why not simply print `kwargs` as-is? Seems very complex for what's essentially a simulation of what the command could look like.

samcli/lib/config/samconfig.py
```diff
@@ -183,6 +197,46 @@ def _write(self):
...
+            save_global_message = (
+                f'\n\tParameter "{key}={value}" in [{env}.{cmd_name_key}.{section}] is defined as a global '
+                f"parameter [{env}.{DEFAULT_GLOBAL_CMDNAME}.{section}].\n\tThis parameter will be only saved "
+                f"under [{env}.{DEFAULT_GLOBAL_CMDNAME}.{section}] in {self.filepath}."
+            )
+            LOG.info(save_global_message)
+            # Only keep the global parameter
+            items = self.document[env][cmd_name_key][section].copy()
+            items.pop(key)
+            self.document[env][cmd_name_key][section] = items
```
hoffa, line 238: Why the copy? If this is just deleting `key`, `del self.document[env][cmd_name_key][section][key]` should be enough.

samcli/lib/config/samconfig.py
wchengru: Yes.

samcli/lib/config/samconfig.py
wchengru: No special preference on these two options.

samcli/cli/main.py
wchengru: Just want to make the format better.

samcli/lib/config/samconfig.py
hoffa: Then `del` makes more sense as it's obvious what's going on and avoids unnecessary copy.

samcli/cli/main.py
hoffa: I'd argue it's overfitted to a very specific use case (and without docstring `log_cmdline_info` suggests it's generic). If `value` is a `list` or `dict` or `set` then the output will be weird.

samcli/lib/config/samconfig.py
wchengru: updated with del.

samcli/cli/main.py
wchengru: Thanks for mentioning this, the docstring was missed. It's added.

PR #2583, opened 2021-02-03
fix: explicit event encoding

hoffa:

https://github.com/aws/aws-sam-cli/pull/2533 doesn't work on Windows as default encoding is used in:

https://github.com/aws/aws-sam-cli/blob/bb7dac0935a3664b845193d70270a464df17ec1f/samcli/commands/local/invoke/cli.py#L201-L202

This defaults to `cp1252` (same as Latin-1) on Windows (and `utf-8` on Linux), which is a single-byte encoding, and fails on the UTF-8 bytes in the `event_utf8.json` of the PR above.

##### Side note

We should check every usage of `open` and `click.open_file`. Anything that doesn't specify `encoding` defaults to the [default encoding](https://docs.python.org/3/library/functions.html#open), meaning all those situations will fail with UTF-8 input on Windows.

Explicitly read as UTF-8.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/invoke/cli.py
```diff
@@ -198,5 +198,5 @@ def _get_event(event_file_name):
 
     # click.open_file knows to open stdin when filename is '-'. This is safer than manually opening streams, and
     # accidentally closing a standard stream
-    with click.open_file(event_file_name, "r") as fp:
+    with click.open_file(event_file_name, "r", encoding="utf-8") as fp:
```
c2tarun, line 201: This out of scope for this PR, but do you have any idea on how we can detect this in Unit test and not integration test?

c2tarun (approved):

samcli/commands/local/invoke/cli.py
hoffa: Running the UTs on Windows is one way (but not necessarily sustainable). Another way is probably forcing the default encoding to be `cp1252` during UTs, so that it simulates the Windows case.

c2tarun (approved):

wchengru (approved):

samcli/commands/local/invoke/cli.py
hoffa: Setting `LC_ALL=C PYTHONUTF8=0` seems to be one way to force a single-byte default encoding.

hoffa:
Likely relevant: https://github.com/aws/aws-sam-cli-app-templates/issues/114

PR #2587, opened 2021-02-03
Making pytest stop running tests after 10 failures.

c2tarun:
Our tests when run on AppVeyor take too long to fail. This change will make it fail fast so that we do not have to wait long time to see the failure. Change is take from pytest documentation here: https://docs.pytest.org/en/2.1.0/customize.html

This change will wait for 10 failures and then fail the full test run and report the failures.

NA

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 log_cli = 1
 log_cli_level = INFO
+addopts = --maxfail=10 -rf
 #filterwarnings =
 #    error


hoffa (approved):

wchengru (approved):

hoffa:
Included in https://github.com/aws/aws-sam-cli/pull/2586

Issue #2144, opened 2020-07-29
SAM build container with environment variables


forzagreen:
### Description

When building lambda functions in Docker container ( `sam build --use-container` ), we can pass in a Docker network (with `--docker-network`), but you can't pass environment variables.

It would be very useful to be able to provide environment variables.

### Use Case

Python lambda function using packages from a private registry. You won't have to hardcode your credentials in `requirements.txt`

### Proposal

Add a new parameter to pass environment variables. Something like:
```sh
sam build --use-container --docker-env VAR1=value1 --docker-env VAR2=value2
```
or:
```sh
sam build --use-container --docker-env-file dev.env
```

willfish:
@awsjeffg What's the pm-review process? Looking at the container/CLI implementation it seems like this is pretty straightforward to achieve from a code perspective?

morefaey:
Thanks @willfish for your comment. The pm-review label is for us to track and review feature requests and rank the ones that we have to work on. you are right that this straightforward to achieve, and we are currently reviewing it to see when it gets implemented.

staenker:
So, is that backlog grooming meeting still going on? If so, I personally like a good beard and a good beard definitely deserves good grooming. So if the beard is the issue, you have my full understanding and emotional support. 
If not, could you maybe consider pushing this showstopper issue for using sam with containers for anything more than a hello world to a higher rank?

qingchm:
@staenker Hi there yes we are looking to support it, sorry for the wait but I will try to work on this as soon as possible! But please understand that there are other tasks with high ranks as well! Again apologies on the wait!

nickswiss:
I feel like this may be too specific to only env vars. Secrets are also a highly used feature in docker builds and I think something along the lines of `--docker-build-flags` might work even better, which would cover both use cases.

qingchm:
@forzagreen Sorry for all the wait, just a heads up this implementation is done and under review now, so you should be able to install private registries by passing in an environment variable json file as a build option very soon (if I'm correctly understanding your use case). Again thanks for being patient with our SAM CLI ;) hopefully this update will make you satisfied!

qingchm:
Released this feature with our v1.20.0, closing the issue now, let me know if you have any further questions!

Issue #2312, opened 2020-10-18
sam package & deploy without using an S3 bucket


augustoproiete:
### Describe your idea/feature/enhancement

Currently sam package & deploy require an S3 bucket to store the packaged app and that the deployment be done from that location. Developers that already have an established CI/CD pipeline already have a repository for build artifacts where they'd like store the packaged sam app, and deploy packages from that location

The workaround today is to push artifacts to the developer's repository and during deployment run `sam package` and `sam deploy` just to satisfy the S3 requirement, which is inefficient and has limitations such as having to manually set up S3 buckets with the appropriate permissions as described in #1701.
 
### Proposal

Allow developers to specify an output folder & filename for creating a packaged application as a zip file, and allow developers to deploy from that zip file.

e.g.
```
sam package --output-zip-file my-sam-app.zip
sam deploy --zip-file my-sam-app.zip
```

ajhool:
Previously mentioned in: #1516 

@prenx4x would you mind commenting on #1516 when you have some time? There is support for it but no comments from AWS team yet.

sriram-mv:
Closing this issue as aws sam cli sees packaging and deploying of built artifacts as a non local action. However one can zip up the `.aws-sam/build` directory after running `sam build` and store it as an artifact that can be used for deployments separately.

PR #2592, opened 2021-02-05
Feat: SAM CLI env vars option for containerized build

qingchm:
#2144 

Whenever we do a sam build, —-use-container is a convenient flag to use to containerize our build so that we don’t need to worry about matching the required environments on our machine. 
However, it does not perfectly simulate the capabilities of sam build command when the —use-container is off. One example is that when we build without —u, environment variables can be set in command line as part of the build, but we do not support this feature when we do a containerized build so far. A lot of customers expressed the need for this feature.

Customers are able to use this functionality to set function or layer specific env vars for the containers bused to build each function and layer. In their build containers, they can pass in credentials like GitHub token to pull packages from private registries as a use case example. They will input a json file containing environment variables, which we will pass to the build containers. In this way, they can customize build configurations inside build containers at a function/layer level. Before this we did not support passing environment variables into build containers. 

No side effects known.

#### Checklist

- [X] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write unit tests
- [X] `make pr` passes
- [X] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -611,3 +629,73 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
             raise ValueError(msg)
 
         return cast(Dict, response)
+
+    @staticmethod
+    def _get_env_vars_value(filename: Optional[str]) -> Any:
```
aahung, line 634: The docstring writes return value is a `dict`, can we use `Dict` here instead of `Any`?

> :return dict: Value of environment variables, if provided. None otherwise

samcli/lib/build/app_builder.py
```diff
@@ -611,3 +629,73 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+            with open(filename, "r") as fp:
+                return json.load(fp)
+
+        except Exception as ex:
+            raise IOError(
+                "Could not read environment variables overrides from file {}: {}".format(filename, str(ex))
+            ) from ex
+
+    @staticmethod
+    def _make_env_vars(function: Function, env_vars_values: dict) -> Any:
```
aahung, line 658: same here, `Any` -> `Dict`

samcli/commands/build/command.py
```diff
@@ -126,6 +126,13 @@
     "requests=1.x and the latest request module version changes from 1.1 to 1.2, "
     "SAM will not pull the latest version until you run a non-cached build.",
 )
+@click.option(
+    "--container-env-vars",
```
aahung, line 130: this option name feels a bit off. When I see it, I expect to pass the vars in the format the same as parameter overrides instead of a file.

aahung: update: existing similar option https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-invoke.html#serverless-sam-cli-using-invoke-environment-file

samcli/local/docker/lambda_build_container.py
```diff
@@ -46,6 +47,7 @@ def __init__(  # pylint: disable=too-many-locals
         source_dir = str(pathlib.Path(source_dir).resolve())
 
         container_dirs = LambdaBuildContainer._get_container_dirs(source_dir, manifest_dir)
+        env_vars = env_vars or {}
```
hoffa, line 50: nit: haven't seen this pattern much in Python, I'd just use a plain old `if`

samcli/lib/build/app_builder.py
```diff
@@ -332,7 +337,12 @@ def _stream_lambda_image_build_logs(self, build_logs: List[Dict[str, str]], func
                     self._stream_writer.flush()
 
     def _build_layer(
-        self, layer_name: str, codeuri: str, specified_workflow: str, compatible_runtimes: List[str]
+        self,
+        layer_name: str,
+        codeuri: str,
+        specified_workflow: str,
+        compatible_runtimes: List[str],
+        env_vars: Optional[Dict] = None,
```
hoffa, line 345: Are these specifically for container-build envvars? Envvars are by definition already in the environment and don't otherwise require special handling. In that case I'd make it clear it's only for containers.

samcli/lib/build/app_builder.py
hoffa, line 634: This does the same thing:

https://github.com/aws/aws-sam-cli/blob/b5cfc46aa9726b5cd006df8ecc08d1b4eedeb9ea/samcli/commands/local/cli_common/invoke_context.py#L400

I'm not convinced the duplication or special method is necessary, instead of a generic `json.load` wrapper.

samcli/lib/build/app_builder.py
```diff
@@ -611,3 +629,73 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+        :return dict: Value of environment variables, if provided. None otherwise
+        :raises IOError: If the file was not found or not a valid JSON
+        """
+        if not filename:
+            return None
+
+        # Try to read the file and parse it as JSON
+        try:
+
+            with open(filename, "r") as fp:
```
hoffa, line 649: This will break on Windows with UTF-8 bytes. See https://github.com/aws/aws-sam-cli/pull/2583, #2312, the linked issues, and likely many other unreported ones.

qingchm: Got it, will update this, thanks!

samcli/lib/build/app_builder.py
qingchm: Dict wouldn't work as it can return none as well

samcli/lib/build/app_builder.py
qingchm: But the type check would fail when we return none if we type hint this to be Dict

samcli/local/docker/lambda_build_container.py
qingchm: ah, I can change this to env_vars if env_vars else {}, but this pattern is used in our code in a couple of places for similar assignment, that's why I used this

samcli/lib/build/app_builder.py
aahung: `Optional[Dict]` can be used if it can be None

aahung: and `json.loads()` returns `Any`, since we know it is a JSON dictionary, we can do `cast(Dict, parsed_json)`

samcli/local/docker/lambda_build_container.py
hoffa: Or simply

```
if not env_vars:
    env_vars = {}
```

(which is also symmetrical to the pattern used in the file below)

hoffa: Actually I assume in these case the clearest would likely to default to an empty `dict`; would also get rid of the `Optional`.

samcli/lib/build/app_builder.py
qingchm: that is correct, are you suggesting a name change? or a comment would be sufficient? I would want to use env_vars name since all the other constructors use the field name env_vars

samcli/lib/build/app_builder.py
qingchm: sure let me try this change with tests!

samcli/local/docker/lambda_build_container.py
qingchm: our pylint doesn't allow the default value to be an empty dict unfortunately, so I changed it to env_vars if env_vars else {}, but without removing the Optional

samcli/lib/build/app_builder.py
qingchm: sounds good, removed this method for simplicity

qingchm: Changed the default return value from None to empty dict, got rid of the Any return type, otherwise it always requires it to be Any

samcli/lib/build/app_builder.py
```diff
@@ -361,7 +380,7 @@ def _build_layer(
                     build_runtime = compatible_runtimes[0]
             options = ApplicationBuilder._get_build_options(layer_name, config.language, None)
 
-            build_method(config, code_dir, artifacts_dir, scratch_dir, manifest_path, build_runtime, options)
+            build_method(config, code_dir, artifacts_dir, scratch_dir, manifest_path, build_runtime, options, env_vars)
```
aahung, line 383: What is gonna happen to `env_vars` when `container_manager` is `None`?

about the naming `env_vars`, it might cause a bit confusion here. I prefer to name it the full `container_env_vars` because `env_vars` in only used when `if self._container_manager:` is true.
Otherwise I was like "why env_vars is not used when we don't use container" if I don't have the full picture.

tests/unit/lib/build_module/test_build_strategy.py
```diff
@@ -156,6 +157,7 @@ def test_build_layers_and_functions(self, mock_copy_tree, mock_path):
                     self.function_build_definition2.get_handler_name(),
                     ANY,
                     self.function_build_definition2.metadata,
+                    self.function_build_definition1.env_vars,
```
aahung, line 160: double check here, why it is `definition1` here?

aahung (changes requested):
Did a full pass and here are two comments, thanks!

tests/unit/lib/build_module/test_build_strategy.py
qingchm: good catch! because they're both mock results, that's why tests don't fail

samcli/lib/build/app_builder.py
qingchm: actually we were thinking about removing the word container even, I'll leave this discussion with Mark and UX team and come back to this

samcli/commands/build/command.py
```diff
@@ -101,6 +107,21 @@
     help="If your functions depend on packages that have natively compiled dependencies, use this flag "
     "to build your function inside an AWS Lambda-like Docker container",
 )
+@click.option(
+    "--container-env-vars",
+    "-e",
+    default=None,
+    multiple=True,  # Can pass in multiple env vars
+    required=False,
+    help="Path to environment variable json file (ex: env_vars.json) to pass into build containers",
```
aahung, line 116: Why this help text looks exactly the same as `--container-env-vars-file?`

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+
+        name = function.name
+        result = {}
+
+        if file_env_vars and inline_env_vars:
+            for file_env_var in file_env_vars.values():
+                if not isinstance(file_env_var, dict):
+                    reason = """
+                                Environment variables in incorrect format
+                                """
```
aahung, line 739: This `""" ..."""` indentation looks off, can we double check this?

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+                    LOG.debug(reason)
+                    raise OverridesNotWellDefinedError(reason)
+
+            for inline_env_var in inline_env_vars.values():
+                if not isinstance(inline_env_var, dict):
+                    reason = """
+                                Environment variables in incorrect format
+                                """
```
aahung, line 747: same here, is this indentation intentional?

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+                    LOG.debug(reason)
+                    raise OverridesNotWellDefinedError(reason)
```
aahung, line 749: these two for loops look identical to me. Let's combine them?
```py
for env_var in file_env_vars.values() + isinstance(inline_env_var, dict):
    ...
```

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+
+            for inline_env_var in inline_env_vars.values():
+                if not isinstance(inline_env_var, dict):
+                    reason = """
+                                Environment variables in incorrect format
```
aahung, line 746: Maybe we can be more helpful here. like `Environment variables {inline_env_var} in incorrect format`

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+            LOG.debug("Environment variables data is being parsed")
+            # CloudFormation parameter file format
+            parameter_result = file_env_vars.get("Parameters", {})
+            result = parameter_result.copy()
+            inline_parameter_result = inline_env_vars.get("Parameters", {})
+            result.update(inline_parameter_result)
+            specific_result = file_env_vars.get(name, {})
+            result.update(specific_result)
+            inline_specific_result = inline_env_vars.get(name, {})
+            result.update(inline_specific_result)
```
aahung, line 760: question: for `inline_parameter_result` and `inline_specific_result`, who has higher precedence? Maybe worth to drop a note here.

samcli/lib/build/app_builder.py
```diff
@@ -663,3 +706,89 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
...
+                                """
+                    LOG.debug(reason)
+                    raise OverridesNotWellDefinedError(reason)
+
+            LOG.debug("Environment variables data is being parsed")
+            # CloudFormation parameter file format
+            parameter_result = inline_env_vars.get("Parameters", {})
+            result = parameter_result.copy()
+            specific_result = inline_env_vars.get(name, {})
+            result.update(specific_result)
```
aahung, line 792: this `if ... elif ... elif` can be simplified to avoid code duplication. 

You can initialize a empty `parameter_result` and go through `file_env_vars` and `inline_env_vars` if they exist.

aahung (changes requested):

samcli/lib/build/app_builder.py
qingchm: I don't really believe so, the priority needs to be written this way to be maintained, I don't see any ways to simplify this

samcli/lib/build/app_builder.py
qingchm: I am putting a comment for this function to specify the precedence

samcli/lib/build/app_builder.py
qingchm: makes sense

samcli/lib/build/app_builder.py
qingchm: I'll create a helper for them to use

samcli/lib/build/app_builder.py
qingchm: I'll check on this

samcli/lib/build/app_builder.py
qingchm: I'll check on this

samcli/commands/build/command.py
qingchm: My bad, pasted the wrong text here, will ship a fix soon

samcli/commands/build/command.py
```diff
@@ -335,3 +368,34 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
         raise click.UsageError("Invalid value for 'mode': invalid choice: {}. (choose from {})".format(mode, choices))
 
     return mode
+
+
+def _process_env_var(container_env_vars: list) -> Dict:
+    processed_env_vars: Dict = {}
+
+    for env_var in container_env_vars:
+        if "." in env_var:
```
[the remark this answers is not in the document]
qingchm: I can add some comments here to help understand

samcli/commands/build/command.py
```diff
@@ -335,3 +368,34 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
         raise click.UsageError("Invalid value for 'mode': invalid choice: {}. (choose from {})".format(mode, choices))
 
     return mode
+
+
+def _process_env_var(container_env_vars: list) -> Dict:
+    processed_env_vars: Dict = {}
```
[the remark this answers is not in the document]
qingchm: sure

samcli/lib/build/app_builder.py
aahung: This?

```py
@staticmethod
def _make_env_vars(function: Function, file_env_vars: Dict, inline_env_vars: Optional[Dict]) -> Dict:
    """Returns the environment variables configuration for this function
    Parameters
    ----------
    function : samcli.lib.providers.provider.Function
        Lambda function to generate the configuration for
    Returns
    -------
    dictionary
        Environment variable configuration for this function
    Raises
    ------
    samcli.commands.local.lib.exceptions.OverridesNotWellDefinedError
        If the environment dict is in the wrong format to process environment vars
    """

    name = function.name
    result = {}


    # validate and raise OverridesNotWellDefinedError
    for env_var in (file_env_vars or {}).values() + (inline_env_vars or {}).values():
        if not isinstance(env_var, dict):
            reason = """
                        Environment variables in incorrect format
                        """
            LOG.debug(reason)
            raise OverridesNotWellDefinedError(reason)

    if file_env_vars:
        parameter_result = file_env_vars.get("Parameters", {})
        result.update(parameter_result)
    
    if inline_env_vars:
        inline_parameter_result = inline_env_vars.get("Parameters", {})
        result.update(inline_parameter_result)
    
    if file_env_vars:
        specific_result = file_env_vars.get(name, {})
        result.update(specific_result)
    
    if inline_env_vars:
        inline_specific_result = inline_env_vars.get(name, {})
        result.update(inline_specific_result)

    return result
```

qingchm: sure I'm not sure if this would create any behaviour but thanks let me try to run it with this!

aahung: This proposed version only needs less than 30 lines of code and with better readability.

aahung: yeah, you have unit tests to verify it.

samcli/commands/build/command.py
```diff
@@ -113,7 +113,9 @@
     default=None,
     multiple=True,  # Can pass in multiple env vars
     required=False,
-    help="Path to environment variable json file (ex: env_vars.json) to pass into build containers",
+    help="Input environment variables through command line to pass into build containers, you can either"
```
aahung, line 116: seems to miss a space at the end of the line.

samcli/lib/build/app_builder.py
```diff
@@ -711,6 +711,12 @@ def _parse_builder_response(stdout_data: str, image_name: str) -> Dict:
     def _make_env_vars(function: Function, file_env_vars: Dict, inline_env_vars: Optional[Dict]) -> Dict:
         """Returns the environment variables configuration for this function
 
+        Priority order (high to low):
+        1. Function specific env vars from command line
+        2. Function specific env vars from json file
+        3. Global env vars from command line
+        4. Global env vars from file
```
aahung, line 718: "from json file?"

samcli/commands/build/command.py
```diff
@@ -335,3 +370,47 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
...
+            test_target = env_var.split("=")
+            if len(test_target) != 2 or not test_target[0].strip() or not test_target[1].strip():
+                LOG.error("Invalid command line --container-env-vars input %s, skipped", env_var)
+                continue
+            key, value = env_var.split("=")
+            if not processed_env_vars.get("Parameters"):
+                processed_env_vars["Parameters"] = {}
+            processed_env_vars["Parameters"][key] = value
+
+    return processed_env_vars
```
aahung, line 416: same as the `_make_env_vars()`, this can be simplified in a waterfall manner:

```py
def _process_env_var(container_env_vars: list) -> Dict:
    processed_env_vars: Dict = {}

    for env_var in container_env_vars:
        location_key = 'Parameters'
        pure_env_var = env_var
        if "." in env_var:
            test_target = env_var.split(".")
            if len(test_target) != 2 or not test_target[0].strip() or not test_target[1].strip():
                LOG.error("Invalid command line --container-env-vars input %s, skipped", env_var)
                continue
            location_key, pure_env_var = env_var.split(".")
        
        test_target = pure_env_var.split("=")
        if len(test_target) != 2 or not test_target[0].strip() or not test_target[1].strip():
            LOG.error("Invalid command line --container-env-vars input %s, skipped", env_var)
            continue
        key, value = pure_env_var.split("=")
        if not processed_env_vars.get(location_key):
            processed_env_vars[location_key] = {}
        processed_env_vars[location_key][key] = value
    return processed_env_vars
```

aahung (commented):
LGTM now, just some minor comments left

aahung:
thanks for addressing the comments.

samcli/commands/build/command.py
qingchm: good catch!

samcli/lib/build/app_builder.py
qingchm: yep I'll add it to keep it consistent

samcli/commands/build/command.py
qingchm: thanks! refactoring this way is tested and worked as expected

samcli/commands/build/command.py
```diff
@@ -149,6 +172,8 @@ def cli(
     parallel: bool,
     manifest: Optional[str],
     docker_network: Optional[str],
+    container_env_vars: list,
+    container_env_vars_file: str,
```
aahung, line 176: I am not sure about this one, if `required=False` for both of them in click, should we add `Optional` to them?

```suggestion
    container_env_vars: Optional[List[str]],
    container_env_vars_file: Optional[str],
```

samcli/commands/build/command.py
```diff
@@ -191,6 +218,8 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     skip_pull_image: bool,
     parameter_overrides: Dict,
     mode: Optional[str],
+    container_env_vars: list,
+    container_env_vars_file: str,
```
aahung, line 222: same here, both of them are optional

samcli/commands/build/command.py
```diff
@@ -216,6 +245,8 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     if use_container:
         LOG.info("Starting Build inside a container")
 
+    processed_env_vars = _process_env_var(container_env_vars)
```
aahung, line 248: when customers do not provide any env vars, is `container_env_vars` `None` or empty list?

samcli/commands/build/command.py
```diff
@@ -335,3 +370,41 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
         raise click.UsageError("Invalid value for 'mode': invalid choice: {}. (choose from {})".format(mode, choices))
 
     return mode
+
+
+def _process_env_var(container_env_vars: list) -> Dict:
```
aahung, line 375:
```suggestion
def _process_env_var(container_env_vars: List[str]) -> Dict:
```

aahung (commented):
Thx! Just did another round, some questions about whether the two arguments could be `None` or not.

samcli/commands/build/command.py
qingchm: It's going to be none, as specified in the click option. But after processing it will be an empty list

aahung: so `container_env_vars` could be `None`? `_process_env_var()` does not seem to handle the `None` situation.

qingchm: Adding a check to prevent iterating over None type

samcli/commands/build/command.py
qingchm: addressed

samcli/commands/build/command.py
qingchm: addressed

samcli/commands/build/command.py
qingchm: addressed

samcli/commands/build/command.py
aahung: Kind of strange our test suites did not catch this, anyway thanks for fixing

aahung (approved):
LGTM

samcli/commands/build/command.py
aahung: I think we do have unit tests to test when `container_env_vars` is `None` right?

samcli/commands/build/command.py
```diff
@@ -101,6 +107,23 @@
     help="If your functions depend on packages that have natively compiled dependencies, use this flag "
     "to build your function inside an AWS Lambda-like Docker container",
 )
+@click.option(
+    "--container-env-vars",
```
hoffa, line 111: Shouldn't this be `--container-env-var`?

samcli/commands/build/command.py
```diff
@@ -101,6 +107,23 @@
...
+    default=None,
+    multiple=True,  # Can pass in multiple env vars
+    required=False,
+    help="Input environment variables through command line to pass into build containers, you can either "
+    "input function specific format (FuncName.VarName=Value) or global format (VarName=Value). Example: "
+    "sam build --use-container --container-env-vars Func1.VAR1=value1 --container-env-vars VAR2=value2",
+)
+@click.option(
+    "--container-env-vars-file",
+    "-ef",
```
hoffa, line 122: Do we ever use two-character shorthands? Is it needed?

samcli/commands/build/command.py
```diff
@@ -101,6 +107,23 @@
...
+    default=None,
+    type=click.Path(),  # Must be a json file
+    help="Path to environment variable json file (ex: env_vars.json) to pass into build containers",
```
hoffa, line 125: nit: AFAIK `ex` is not standard for "example"; better to use `e.g. env_vars.json`

samcli/commands/build/command.py
```diff
@@ -335,3 +370,42 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
...
+        Processed command line environment variables
+    """
+    processed_env_vars: Dict = {}
+
+    if container_env_vars:
+        for env_var in container_env_vars:
+            location_key = "Parameters"
+            pure_env_var = env_var
+            if "." in env_var:
+                test_target = env_var.split(".")
```
hoffa, line 396: What's a test target?

samcli/commands/build/command.py
```diff
@@ -335,3 +370,42 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
         raise click.UsageError("Invalid value for 'mode': invalid choice: {}. (choose from {})".format(mode, choices))
 
     return mode
+
+
+def _process_env_var(container_env_vars: Optional[List[str]]) -> Dict:
```
hoffa, line 375: Do we use this format elsewhere? Is there any other command that has similar use case?

samcli/commands/build/command.py
```diff
@@ -335,3 +370,42 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
...
+    """
+    processed_env_vars: Dict = {}
+
+    if container_env_vars:
+        for env_var in container_env_vars:
+            location_key = "Parameters"
+            pure_env_var = env_var
+            if "." in env_var:
+                test_target = env_var.split(".")
+                if len(test_target) != 2 or not test_target[0].strip() or not test_target[1].strip():
```
hoffa, line 397: So what if I input `Function.SAM_CLI_VERSION=1.2.3`?

samcli/commands/build/command.py
```diff
@@ -335,3 +370,42 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
...
+                    LOG.error("Invalid command line --container-env-vars input %s, skipped", env_var)
+                    continue
+                location_key, pure_env_var = env_var.split(".")
+
+            test_target = pure_env_var.split("=")
+            if len(test_target) != 2 or not test_target[0].strip() or not test_target[1].strip():
```
hoffa, line 403: Very similar to above and can be refactored. They're doing essentially the same thing: getting a key-value pair from a string with arbitrary delimiter. And there's no need to enforce a 2 values; it can instead just limit the `split` to 1.

samcli/commands/build/command.py
```diff
@@ -101,6 +107,23 @@
...
+    "-e",
+    default=None,
+    multiple=True,  # Can pass in multiple env vars
+    required=False,
+    help="Input environment variables through command line to pass into build containers, you can either "
+    "input function specific format (FuncName.VarName=Value) or global format (VarName=Value). Example: "
+    "sam build --use-container --container-env-vars Func1.VAR1=value1 --container-env-vars VAR2=value2",
+)
+@click.option(
+    "--container-env-vars-file",
```
hoffa, line 121: `sam local invoke` has `--container-env-vars` as a path to a file. It would be more consistent to make this `--container-env-vars` as well, and use the singular form for direct input.

samcli/commands/build/command.py
qingchm: This is designed this way to be consistent with the image-repositories tag in deploy command. This has been finalized with Mark and UX team.

samcli/commands/build/command.py
qingchm: a target string to be tested on its format

samcli/commands/build/command.py
aahung: looks like we have some usage of `ex:` in our current codebase. 

image

and similar amount of `e.g.,`

image

samcli/commands/build/command.py
qingchm: nope, this is a unique pattern so far, but might have future use cases

samcli/commands/build/command.py
hoffa: In any case the standard notation for "for example" is "e.g.", see for example:
- https://english.stackexchange.com/a/16209
- https://blog.prepscholar.com/ie-vs-eg-vs-ex-definition

aahung: I think it is worth to go extra mile to make all `ex:` into `e.g.,`

qingchm: right, I have already changed the manifest one. we can create a separate pr to change all of them

hoffa (approved):
Shaped up really nicely, this will be super helpful! 👍

Approved, but I'd still strongly consider the naming of the arguments to avoid it being both misleading (plural when `--container-env-vars` takes a single pair) and inconsistent (`local invoke --container-env-vars` takes a file, `build --container-env-vars` takes a key-value string).

qingchm:
> Shaped up really nicely, this will be super helpful! 👍
> 
> Approved, but I'd still strongly consider the naming of the arguments to avoid it being both misleading (plural when `--container-env-vars` takes a single pair) and inconsistent (`local invoke --container-env-vars` takes a file, `build --container-env-vars` takes a key-value string).

Asking for Mark's feedback, will merge after that final discussion

aahung:
one last check, I have never seen `e.g.` used before `:`. I only see `e.g., ....`, can english export @hoffa double check this?

samcli/commands/build/command.py
qingchm: ah actually we don't, we only have a few tests that test on build command

qingchm:
> one last check, I have never seen `e.g.` used before `:`. I only see `e.g., ....`, can english export @hoffa double check this?

lmao export as in exporting env vars haha, that fits well with the topic

hoffa:
@aahung Haha you're right (sources: [1](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/), [2](https://english.stackexchange.com/a/16173), [3](https://www.warnell.uga.edu/sites/default/files/inline-files/ieVeg.pdf)), and even mentioned it half tongue-in-cheek privately; didn't dare to be as pedantic here.

aahung:
> @aahung Haha you're right (sources: [1](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/), [2](https://english.stackexchange.com/a/16173), [3](https://www.warnell.uga.edu/sites/default/files/inline-files/ieVeg.pdf)), and even mentioned it half tongue-in-cheek privately; didn't dare to be as pedantic here.

@qingchm can you change the `e.g.: ` to `e.g., `?

qingchm:
> > @aahung Haha you're right (sources: [1](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/), [2](https://english.stackexchange.com/a/16173), [3](https://www.warnell.uga.edu/sites/default/files/inline-files/ieVeg.pdf)), and even mentioned it half tongue-in-cheek privately; didn't dare to be as pedantic here.
> 
> @qingchm can you change the `e.g.: ` to `e.g., `?

Yep changed it, pushing a new version

samcli/commands/build/command.py
qingchm: Naming change confirmed. (Just to record this in conversation, I know you are aware already)

samcli/commands/build/command.py
qingchm: Redesigned parsing mechanism to support this type of input

qingchm: Also added unit test with weird inputs

samcli/commands/build/command.py
qingchm: Added a unit test for the None case

driverpt:
Has this fix bee nreleased ?

qingchm:
> Has this fix been released ?

Hey @driverpt this will be released very soon! Will notify you when the release happens!

driverpt:
Please define "very soon".

qingchm:
@driverpt it is released now with our latest version 1.20.0 ;)

PR #2594, opened 2021-02-05
chore: Change function/layer providers for multi-stack support

aahung:
#### Why is this change necessary?

This PR is a preparation for nested stack support. It changes `SamFunctionProvider` and `SamLayerProvider` to make them capable of extracting functions/layers in multiple stacks.

1. Add `stack_path` to `Function` and `LayerVersion` to make them aware of their location in stack hierarchy.
   - `""` means in root stack
   - `"xxx"` means in root->xxx stack
   - `"xxx/yyy"` means in root->xxx>yyy stack (grandchild of root)
2. Create new Class `Stack` to encapsulate `template_dict` and `parameters_overrides` with additional info like name, template file location and its place relative to root template.
   - Both providers' `get(name: str)` can accept either name or `stack_path/name`, but customers won't notice the difference until we enable nested stack.
3. Change their constructors from accepting `template_dict` and `parameters_overrides` to a list of `Stack`.

**Why split the PR?** To make PR review easier and smooth, nested stack feature will be divided into shippable parts with no customers impact except the final one. For this particular PR, despite `SamLocalStackProvider.get_stacks()` has the capability to extract all nested stacks, it does not do so unless in unit tests.

#### What side effects does this change have?

This PR has no customer impact

#### Checklist

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #2609, opened 2021-02-10
fix: SamFunctionProvider could miss functions with same name but in different stacks

aahung:

This is a follow up of #2594, 2nd PR to support nested stack.
*These two commits are taken out from https://github.com/aahung/aws-sam-cli/tree/nested-stack-2

two functions with the same name can appear in two different nested stacks, which causing one overriding the other. 

Before the dict's key is function name, which is not unique in a multi-stack setting. Now the key is a unique key "full_path" consisting of stack_path and function name.

N/A, nested stack support is not available and this PR has no customer impact.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -57,9 +57,13 @@ class Function(NamedTuple):
     @property
     def full_path(self) -> str:
         """
-        Return the unique path-like identifier in a multi-stack situation.
+        Return the path-like identifier of this Function. If it is in root stack, full_path = name.
+        This path is guaranteed to be unique in a multi-stack situation.
+        Example:
+            "HelloWorldFunction"
+            "ChildStackA/GrandChildStackB/AFunctionInNestedStack"
         """
-        return _get_full_path(self)
+        return get_full_path(self.stack_path, self.name)
 
 
 class ResourcesToBuildCollector:
@@ -250,9 +254,13 @@ class LayerVersion:
     @property
     def full_path(self) -> str:
         """
-        Return the unique path-like identifier in a multi-stack situation.
+        Return the path-like identifier of this Layer. If it is in root stack, full_path = name.
+        This path is guaranteed to be unique in a multi-stack situation.
+        Example:
+            "HelloWorldLayer"
+            "ChildStackA/GrandChildStackB/ALayerInNestedStack"
         """
-        return _get_full_path(self)
+        return get_full_path(self.stack_path, self.name)
 
     def __eq__(self, other):
         if isinstance(other, type(self)):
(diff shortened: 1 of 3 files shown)


samcli/lib/providers/sam_function_provider.py
```diff
@@ -108,7 +108,7 @@ def _extract_functions(stacks: List[Stack], ignore_code_extraction_warnings=Fals
             Function configuration object
         """
 
-        result = {}
+        result = {}  # dict: full_path -> Function
```
c2tarun, line 111: no types?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -126,13 +126,14 @@ def _extract_functions(stacks: List[Stack], ignore_code_extraction_warnings=Fals
                         stack.resources,
                         ignore_code_extraction_warnings=ignore_code_extraction_warnings,
                     )
-                    result[name] = SamFunctionProvider._convert_sam_function_resource(
+                    f = SamFunctionProvider._convert_sam_function_resource(
```
c2tarun, line 129: f -> function

c2tarun (changes requested):

samcli/lib/providers/sam_function_provider.py
aahung: Sorry about the confusion, I meant to explain what are the keys and values. Updated

samcli/lib/providers/sam_function_provider.py
aahung: renamed.

mndeveci (approved):
LGTM. Quick question, if I want to invoke a function from child stack, can I invoke it by running `sam local invoke ChildStack/Function`? Same question applies for `sam build` as well.

aahung:
> LGTM. Quick question, if I want to invoke a function from child stack, can I invoke it by running `sam local invoke ChildStack/Function`? Same question applies for `sam build` as well.

Yes, but not yet supported until implemented in next PR, we can either invoke by `local invoke Function` or `local invoke ChildStack/Function` (to avoid ambiguity).

c2tarun (approved):

samcli/lib/providers/provider.py
```diff
@@ -371,3 +393,11 @@ def resources(self) -> Dict:
...
         resources: Dict = processed_template_dict.get("Resources", {})
         return resources
+
+
+def get_full_path(stack_path: str, logical_id: str) -> str:
+    """
+    Return the unique posix path-like identifier
+    while will used for identify a resource from resources in a multi-stack situation
+    """
+    return posixpath.join(stack_path, logical_id)
```
sriram-mv, line 403: are there any concerns here, where posixpath might behave weirdly trying to look for absolute paths and discard anything else?

sriram-mv (changes requested):

samcli/lib/providers/provider.py
aahung: this `full_path` will not be used in the file system, it is a `posix path-like identifier` used in `sam build` and also in `sam local invoke <full_path>`

sriram-mv: ok sounds good, the reason I ask is sometimes Pathlib on windows tries to resolve paths before joining them, I wanted to make sure that posixpath does not run into the same issue.

aahung: Here is the source code of `posixpath.join`, we are pretty safe

https://github.com/python/cpython/blob/fd053fdd39fbdf114b4218ea4309666bafa95788/Lib/posixpath.py#L67-L92

sriram-mv (approved):

PR #2612, opened 2021-02-10
fix: Custom region is not passed properly in local invoke

aahung:

Fix local invoke with custom region. Somehow it was not caught in the canary run yesterday https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-canary-linux/builds/37696048/job/5vth4t3503scpn9a

Caused by a typo in #2594, `_parameter_overrides` should be `parameter_overrides`.
This also further refactors how property `parameter_overrides` works. Doing mutation in property getter is not a good practice. When I looked at the code I did not realize `parameter_overrides` could mutate `_parameter_overrides` and is different from `_parameter_overrides`. The mutation has been moved to the constructor.

N/A

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -132,7 +132,12 @@ class InvokeContext:
         self._debug_args = debug_args
         self._debugger_path = debugger_path
         self._container_env_vars_file = container_env_vars_file
+
         self._parameter_overrides = parameter_overrides or {}
+        # Override certain CloudFormation pseudo-parameters based on values provided by customer
+        if aws_region:
+            self._parameter_overrides["AWS::Region"] = aws_region
+
         self._layer_cache_basedir = layer_cache_basedir
         self._force_image_build = force_image_build
         self._aws_region = aws_region
@@ -372,10 +377,6 @@ class InvokeContext:
 
     @property
     def parameter_overrides(self):
-        # Override certain CloudFormation pseudo-parameters based on values provided by customer
-        if self._aws_region:
-            self._parameter_overrides["AWS::Region"] = self._aws_region
-
         return self._parameter_overrides
 
     @property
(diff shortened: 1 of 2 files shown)


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -132,7 +132,12 @@ def __init__(
         self._debug_args = debug_args
         self._debugger_path = debugger_path
         self._container_env_vars_file = container_env_vars_file
+
         self._parameter_overrides = parameter_overrides or {}
+        # Override certain CloudFormation pseudo-parameters based on values provided by customer
+        if self._aws_region:
```
jfuss, line 138: Why are we moving this out of the property? You don't set `self._aws_region` until line 143, so this is actually always `None`.

jfuss (changes requested):

samcli/commands/local/cli_common/invoke_context.py
aahung: Just noticed that too, fixed

aahung: From what my understanding of the code, this is a part of `self._parameter_overrides` initialization process so it should happen in the constructor instead of in the property.

hoffa (approved):

aahung:
Also changed the unit test to make line
```
self.assertEqual(invoke_context._parameter_overrides.......
``` 
more consistent with other 
```
self.assertEqual(invoke_context._...
```

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -44,7 +44,7 @@ def test_must_read_from_necessary_files(self, SamFunctionProviderMock, Container
         )
 
         template_dict = "template_dict"
-        stacks = [Stack("", "", template_file, invoke_context.parameter_overrides, template_dict)]
+        stacks = [Stack("", "", template_file, invoke_context._parameter_overrides, template_dict)]
```
jfuss, line 47: This should be accessing the property not the private variable.

jfuss (changes requested):

jfuss (approved):
Thanks for handling the feedback.

PR #2618, opened 2021-02-11
fix: Nested stack cannot be extracted if root template is not in working dir

aahung:

This is a follow up of #2594, to support nested stack.

If root template is not `./template.yaml` or in current working directory, local nested templates cannot be resolved properly.

Does a path derivation depending on whether nested template location is absolute path or not. If it is not, it is joined with parent's template directory.

N/A, nested stack support is not available and this PR has no customer impact.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/commands/local/lib/test_stack_provider.py
+++ b/tests/unit/commands/local/lib/test_stack_provider.py
@@ -29,8 +29,8 @@ class TestSamBuildableStackProvider(TestCase):
 
     @parameterized.expand(
         [
-            (AWS_SERVERLESS_APPLICATION, "Location", "./child.yaml", "./child.yaml"),
-            (AWS_CLOUDFORMATION_STACK, "TemplateURL", "./child.yaml", "./child.yaml"),
+            (AWS_SERVERLESS_APPLICATION, "Location", "./child.yaml", "child.yaml"),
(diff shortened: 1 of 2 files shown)


tests/unit/commands/local/lib/test_stack_provider.py
```diff
@@ -168,3 +168,43 @@ def test_remote_stack_is_skipped(self, resource_type, location_property_name):
...
+                "ChildStack": {
+                    "Type": resource_type,
+                    "Properties": {location_property_name: child_location},
+                }
+            }
+        }
+        self.get_template_data_mock.side_effect = lambda t: {
+            template_file: template,
+            child_location_path: LEAF_TEMPLATE,
+        }.get(t)
```
mndeveci, line 196: What kind of sorcery happening here? :)

mndeveci (approved):

tests/unit/commands/local/lib/test_stack_provider.py
aahung: I found this is the most simple way to "return Y if input is X" 😄

aahung:
updated the unit test to take windows path into consideration

hoffa (approved):

PR #2623, opened 2021-02-12
feat: Support build functions in nested stacks

aahung:

Following up #2594, #2609, and #2618, it change how the build artifacts are organized.
There is no customer impact because new behavior only makes a difference when `SAM_CLI_ENABLE_NESTED_STACK` is set to `1`.

##### About the integration test

- Test functions with same name but in different stacks (ZIP & Image)
- Test cache/parallel build (ZIP & Image)

##### Out of scope

This PR focuses on build with a small tweak on local invoke so that build integ test can pass (because build integ test uses invoke to verify built artifacts)
- thorough refactoring and testing on local invoke.

This PR has no customer impact.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


aahung:
Just ran a integration test on this PR https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-canary-linux/build/job/tubrr332pa239pum

Despite there are 5 integ test failures, they are caused by `The security token included in the request is expired`, which is fine because they are not reproducible locally.

image

samcli/lib/build/app_builder.py
```diff
@@ -184,16 +184,24 @@ def _get_build_graph(self) -> BuildGraph:
...
         folder
 
         Parameters
         ----------
-        template_dict
+        stack_path: str
+        template_dict: dict
         original_template_path : str
             Path where the template file will be written to
+        stack_output_template_path_by_stack_path: Dict[str, str]
```
mndeveci, line 204: Some doc will be helpful around this parameter

samcli/lib/build/app_builder.py
```diff
@@ -342,7 +368,7 @@ def _build_layer(
         subfolder = get_layer_subfolder(specified_workflow)
 
         # artifacts directory will be created by the builder
-        artifacts_dir = str(pathlib.Path(self._build_dir, layer_name, subfolder))
+        artifacts_dir = str(pathlib.Path(artifacts_dir, subfolder))
```
mndeveci, line 371: Why `layer_name` is removed from the path? Isn't this going to put built artifact directly in stack's build root folder?

aahung: The artifacts_dir contains layer name now. I added a new argument `artifacts_dir` to `_build_layer()` to make it consistent with 

```
def _build_function(  # pylint: disable=R1710
        self,
        function_name: str,
        codeuri: str,
        packagetype: str,
        runtime: str,
        handler: Optional[str],
        artifacts_dir: str,
        metadata: Optional[Dict] = None,
    )
```

And `artifacts_dir` for both function and layer are created by app_builder, for example for `_build_layer()`:

image

samcli/lib/build/app_builder.py
aahung: updated in https://github.com/aws/aws-sam-cli/pull/2623/commits/94ff67193081484c7ecd1e4e336f205d0389eab1

samcli/lib/build/app_builder.py
aahung: I updated the docstring of `_build_layer()` and hope it make it clearer

mndeveci: Oh I see, but if that is the case, I would rather name the variable as `artifact_dir` (singular) to point that out.

aahung: good idea, changed

mndeveci (approved):

samcli/commands/_utils/template.py
```diff
@@ -87,6 +87,9 @@ def move_template(src_template_path, dest_template_path, template_dict):
     # update any relative paths in the template to be relative to the new location.
     modified_template = _update_relative_paths(template_dict, original_root, new_root)
 
+    # if a stack only has image functions, the directory for that directory won't be created.
```
c2tarun, line 90: Why are we doing this? As in Image Functions are already working (however they are working) right?

samcli/commands/build/command.py
```diff
@@ -248,22 +249,34 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
...
 
-            move_template(ctx.original_template_path, ctx.output_template_path, modified_template)
+            stack_output_template_path_by_stack_path = {
+                stack.stack_path: stack.get_output_template_path(ctx.build_dir) for stack in ctx.stacks
+            }
+            for stack in ctx.stacks:
+                modified_template = builder.update_template(
+                    stack.stack_path,
+                    stack.template_dict,
+                    stack.location,
```
c2tarun, line 260: Just curious, why pass three stack properties individually? and not the whole object? It'll also make documentation bit easier, no?

samcli/lib/build/app_builder.py
```diff
@@ -208,39 +217,52 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
 
         for logical_id, resource in template_dict.get("Resources", {}).items():
 
-            if logical_id not in built_artifacts:
-                # this resource was not built. So skip it
+            full_path = get_full_path(stack_path, logical_id)
+            is_artifact = full_path in built_artifacts
```
c2tarun, line 221: what does `is_artifact` denote? Does this mean that stack is built and artifacts are present? If yes then in what case it won't be present?

samcli/lib/build/app_builder.py
```diff
@@ -208,39 +217,52 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
...
+            is_stack = full_path in stack_output_template_path_by_stack_path
+
+            if not is_artifact and not is_stack:
+                # this resource was not built or a nested stack. So skip it
```
c2tarun, line 225: TBH, I would prefer if we choose a better name of `is_artifact` rather than this overly obvious comment. 😋

Or at least provide a reason of why `skip it`?

samcli/lib/build/app_builder.py
```diff
@@ -208,39 +217,52 @@ def update_template(template_dict: Dict, original_template_path: str, built_arti
...
                 continue
 
-            artifact_dir = pathlib.Path(built_artifacts[logical_id]).resolve()
+            resource_type = resource.get("Type")
+            properties = resource.setdefault("Properties", {})
 
+            abs_out_path = pathlib.Path(
```
c2tarun, line 231: `abs_out_path` -> `absolute_output_path`

c2tarun (changes requested):
Some minor suggestions and questions.

samcli/commands/_utils/template.py
aahung: for image function, sam build won't generate any artifact files, instead it is a image tag (from docker).

samcli/lib/build/app_builder.py
aahung: let's change to `has_build_artifact`. It indicates whether "this resource" has a artifacts generated by sam build.

samcli/commands/build/command.py
aahung: good catch, simplified

samcli/lib/build/app_builder.py
aahung: done

samcli/lib/build/app_builder.py
aahung: comment updated!

c2tarun (approved):

PR #2624, opened 2021-02-12
fix: If image functions have same metadata, only 1st one is in the result

aahung:
…s built

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/build_strategy.py
+++ b/samcli/lib/build/build_strategy.py
@@ -125,16 +125,18 @@ class DefaultBuildStrategy(BuildStrategy):
         function_build_results[single_function_name] = result
 
         # copy results to other functions
-        for function in build_definition.functions:
-            if function.name != single_function_name:
-                if build_definition.packagetype == ZIP:
+        if build_definition.packagetype == ZIP:
+            for function in build_definition.functions:
+                if function.name != single_function_name:
                     # for zip function we need to copy over the artifacts
                     # artifacts directory will be created by the builder
                     artifacts_dir = str(pathlib.Path(self._build_dir, function.name))
                     LOG.debug("Copying artifacts from %s to %s", single_build_dir, artifacts_dir)
                     osutils.copytree(single_build_dir, artifacts_dir)
                     function_build_results[function.name] = artifacts_dir
-                elif build_definition.packagetype == IMAGE:
+        elif build_definition.packagetype == IMAGE:
+            for function in build_definition.functions:
+                if function.name != single_function_name:
                     # for image function, we just need to copy the image tag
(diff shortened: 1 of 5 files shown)


tests/unit/lib/build_module/test_build_strategy.py
```diff
@@ -193,6 +193,30 @@ def test_build_layers_and_functions(self, mock_copy_tree, mock_path):
             str(mock_path(given_build_dir, self.function1_2.name)),
         )
 
+    def test_build_single_function_definition_image_functions_with_same(self, mock_copy_tree, mock_path):
```
sriram-mv, line 196: should we also add an integ test for this?

sriram-mv (changes requested):

tests/unit/lib/build_module/test_build_strategy.py
aahung: sure

samcli/lib/build/build_strategy.py
```diff
@@ -125,14 +125,18 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
         function_build_results[single_function_name] = result
 
         # copy results to other functions
-        if build_definition.packagetype == ZIP:
-            for function in build_definition.functions:
-                if function.name is not single_function_name:
+        for function in build_definition.functions:
+            if function.name != single_function_name:
+                if build_definition.packagetype == ZIP:
+                    # for zip function we need to copy over the artifacts
```
mndeveci, line 131: Why we need to change the order here? This will check the `if` statement for every function in the build definition (no matter it is `ZIP` or `IMAGE`), but the previous order will have 2 separate flow for `IMAGE` and `ZIP` from the beginning.

mndeveci (changes requested):

samcli/lib/build/build_strategy.py
aahung: right, good catch! I optimized for more less code. This indeed introduced more checks. Reverted

samcli/lib/build/build_strategy.py
```diff
@@ -127,12 +127,18 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
         # copy results to other functions
         if build_definition.packagetype == ZIP:
             for function in build_definition.functions:
-                if function.name is not single_function_name:
+                if function.name != single_function_name:
```
aahung, line 130: we should use `!=` instead of `is not` to compare strings.

sriram-mv (approved):

mndeveci (approved):

PR #2631, opened 2021-02-16
chore: Provide future support for nested stack in local invoke/start-api/start-lambda

aahung:

In #2594, #2609, #2618 and #2623, `sam build`, `sam local invoke` and `sam local start-lambda` supports functions in nested stacks when `SAM_CLI_ENABLE_NESTED_STACK=1`.

How PR is a followup to do these things:
- add type hint to make following changes easier
- make start-api work with nested stacks.
  - add "stack_path" to `Route`. In previous PRs, we already added `stack_path` to `Function` and `LayerVersion`.
  - ApiProvider now goes through all nested stacks to find routes.
- add unit/integration tests to verify `local invoke` `local start-lambda` and `local start-api`. 
- fix a bug causing `AWS::Region` is not passed to child stacks when users invoke with `--region` arg (bug discovered during writing integration tests, therefore I put this fix in this PR).

#### What's left for nested stack?

- Remove the env var `SAM_CLI_ENABLE_NESTED_STACK` and make nested stack enabled by default.
- Add integration tests for `sam package` and `sam deploy`.

This PR has no customer impact.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -134,10 +134,11 @@ class InvokeContext:
         self._debugger_path = debugger_path
         self._container_env_vars_file = container_env_vars_file
 
-        self.parameter_overrides = parameter_overrides or {}
+        self._parameter_overrides = parameter_overrides
         # Override certain CloudFormation pseudo-parameters based on values provided by customer
+        self._pseudo_parameter_overrides: Optional[Dict] = None
         if aws_region:
-            self.parameter_overrides["AWS::Region"] = aws_region
+            self._pseudo_parameter_overrides = {"AWS::Region": aws_region}
 
         self._layer_cache_basedir = layer_cache_basedir
         self._force_image_build = force_image_build
@@ -397,7 +398,11 @@ class InvokeContext:
 
     def _get_stacks(self) -> List[Stack]:
         try:
-            return SamLocalStackProvider.get_stacks(self._template_file, parameter_overrides=self.parameter_overrides)
+            return SamLocalStackProvider.get_stacks(
+                self._template_file,
+                parameter_overrides=self._parameter_overrides,
+                pseudo_parameter_overrides=self._pseudo_parameter_overrides,
+            )
         except (TemplateNotFoundException, TemplateFailedParsingException) as ex:
             raise InvokeContextException(str(ex)) from ex
 
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -5,6 +5,7 @@ import errno
 import json
 import logging
 import os
+import posixpath
 from enum import Enum
 from pathlib import Path
 from typing import Dict, List, Optional, IO, cast, Tuple, Any
@@ -288,12 +289,12 @@ class InvokeContext:
             return all_functions[0].name
 
         # Get all the available function names to print helpful exception message
-        all_function_names = [f.name for f in all_functions]
+        all_function_full_paths = [posixpath.join(f.stack_path, f.name) for f in all_functions]
 
         # There are more functions in the template, and function identifier is not provided, hence raise.
         raise InvokeContextException(
-            "You must provide a function identifier (function's Logical ID in the template). "
-            "Possible options in your template: {}".format(all_function_names)
+            "You must provide a function logical ID when there are more than one functions in your template. "
+            "Possible options in your template: {}".format(all_function_full_paths)
         )
 
     @property
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -154,12 +154,13 @@ class InvokeContext:
 
         self._debug_function = debug_function
 
-        # Note(xinhol): despite self._template_dict and self._function_provider are initialized as None
+        # Note(xinhol): despite self._template_dict, self._function_provider and self._stacks are initialized as None
         # they will be assigned with a non-None value in __enter__() and
         # it is only used in the context (after __enter__ is called)
         # so we can assume they are not Optional here
         self._template_dict: Dict = None  # type: ignore
         self._function_provider: SamFunctionProvider = None  # type: ignore
+        self._stacks: List[Stack] = None  # type: ignore
         self._env_vars_value: Optional[Dict] = None
         self._container_env_vars_value: Optional[Dict] = None
         self._log_file_handle: Optional[IO] = None
@@ -177,9 +178,9 @@ class InvokeContext:
         :returns InvokeContext: Returns this object
         """
 
-        stacks = self._get_stacks()
-        self._template_dict = SamLocalStackProvider.find_root_stack(stacks).template_dict
-        self._function_provider = SamFunctionProvider(stacks)
+        self._stacks = self._get_stacks()
+        self._template_dict = SamLocalStackProvider.find_root_stack(self._stacks).template_dict
+        self._function_provider = SamFunctionProvider(self._stacks)
 
         self._env_vars_value = self._get_env_vars_value(self._env_vars_file)
         self._container_env_vars_value = self._get_env_vars_value(self._container_env_vars_file)
@@ -364,6 +365,15 @@ class InvokeContext:
         :return dict: Template data
         """
         return self._template_dict
+
+    @property
+    def stacks(self) -> List[Stack]:
+        """
+        Returns the list of stacks (including the root stack and all children stacks)
+
+        :return list: list of stacks
+        """
+        return self._stacks
 
     def get_cwd(self) -> str:
         """
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -136,9 +136,9 @@ class InvokeContext:
 
         self._parameter_overrides = parameter_overrides
         # Override certain CloudFormation pseudo-parameters based on values provided by customer
-        self._pseudo_parameter_overrides: Optional[Dict] = None
+        self._global_parameter_overrides: Optional[Dict] = None
         if aws_region:
-            self._pseudo_parameter_overrides = {"AWS::Region": aws_region}
+            self._global_parameter_overrides = {"AWS::Region": aws_region}
 
         self._layer_cache_basedir = layer_cache_basedir
         self._force_image_build = force_image_build
@@ -390,7 +390,7 @@ class InvokeContext:
             return SamLocalStackProvider.get_stacks(
                 self._template_file,
                 parameter_overrides=self._parameter_overrides,
-                pseudo_parameter_overrides=self._pseudo_parameter_overrides,
+                global_parameter_overrides=self._global_parameter_overrides,
             )
         except (TemplateNotFoundException, TemplateFailedParsingException) as ex:
             raise InvokeContextException(str(ex)) from ex
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -301,7 +301,7 @@ class InvokeContext:
     @property
     def lambda_runtime(self) -> LambdaRuntime:
         if not self._lambda_runtimes:
-            layer_downloader = LayerDownloader(self._layer_cache_basedir, self.get_cwd())
+            layer_downloader = LayerDownloader(self._layer_cache_basedir, self.get_cwd(), self._stacks)
             image_builder = LambdaImage(layer_downloader, self._skip_pull_image, self._force_image_build)
             self._lambda_runtimes = {
                 ContainersMode.WARM: WarmLambdaRuntime(self._container_manager, image_builder),
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -7,11 +7,15 @@ import logging
 import os
 from enum import Enum
 from pathlib import Path
+from typing import Dict, List, Optional, IO, cast, Tuple, Any
 
 import samcli.lib.utils.osutils as osutils
+from samcli.lib.providers.provider import Stack, Function
 from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
 from samcli.lib.utils.async_utils import AsyncContext
 from samcli.lib.utils.stream_writer import StreamWriter
+from samcli.commands.exceptions import ContainersInitializationException
+from samcli.commands.local.cli_common.user_exceptions import InvokeContextException, DebugContextException
 from samcli.commands.local.lib.local_lambda import LocalLambdaRunner
 from samcli.commands.local.lib.debug_context import DebugContext
 from samcli.local.lambdafn.runtime import LambdaRuntime, WarmLambdaRuntime
@@ -20,8 +24,6 @@ from samcli.local.docker.manager import ContainerManager
 from samcli.commands._utils.template import TemplateNotFoundException, TemplateFailedParsingException
 from samcli.local.layers.layer_downloader import LayerDownloader
 from samcli.lib.providers.sam_function_provider import SamFunctionProvider
-from .user_exceptions import InvokeContextException, DebugContextException
-from ...exceptions import ContainersInitializationException
 
 LOG = logging.getLogger(__name__)
 
@@ -54,26 +56,26 @@ class InvokeContext:
 
     def __init__(
         self,  # pylint: disable=R0914
-        template_file,
-        function_identifier=None,
-        env_vars_file=None,
-        docker_volume_basedir=None,
-        docker_network=None,
-        log_file=None,
-        skip_pull_image=None,
-        debug_ports=None,
-        debug_args=None,
-        debugger_path=None,
-        container_env_vars_file=None,
-        parameter_overrides=None,
-        layer_cache_basedir=None,
-        force_image_build=None,
-        aws_region=None,
-        aws_profile=None,
-        warm_container_initialization_mode=None,
-        debug_function=None,
-        shutdown=False,
-    ):
+        template_file: str,
+        function_identifier: Optional[str] = None,
+        env_vars_file: Optional[str] = None,
+        docker_volume_basedir: Optional[str] = None,
+        docker_network: Optional[str] = None,
+        log_file: Optional[str] = None,
+        skip_pull_image: Optional[bool] = None,
+        debug_ports: Optional[Tuple[int]] = None,
+        debug_args: Optional[str] = None,
+        debugger_path: Optional[str] = None,
+        container_env_vars_file: Optional[str] = None,
+        parameter_overrides: Optional[Dict] = None,
+        layer_cache_basedir: Optional[str] = None,
+        force_image_build: Optional[bool] = None,
+        aws_region: Optional[str] = None,
+        aws_profile: Optional[str] = None,
+        warm_container_initialization_mode: Optional[str] = None,
+        debug_function: Optional[str] = None,
+        shutdown: bool = False,
+    ) -> None:
         """
         Initialize the context
 
@@ -90,8 +92,7 @@ class InvokeContext:
         docker_network str
             Docker network identifier
         log_file str
-            Path to a file to send container output to. If the file does not exist, it will be
-            created
+            Path to a file to send container output to. If the file does not exist, it will be created
         skip_pull_image bool
             Should we skip pulling the Docker container image?
         aws_profile str
@@ -153,19 +154,23 @@ class InvokeContext:
 
         self._debug_function = debug_function
 
-        self._template_dict = None
-        self._function_provider = None
-        self._env_vars_value = None
-        self._container_env_vars_value = None
-        self._log_file_handle = None
-        self._debug_context = None
-        self._layers_downloader = None
-        self._container_manager = None
-        self._lambda_runtimes = None
+        # Note(xinhol): despite self._template_dict and self._function_provider are initialized as None
+        # they will be assigned with a non-None value in __enter__() and
+        # it is only used in the context (after __enter__ is called)
+        # so we can assume they are not Optional here
+        self._template_dict: Dict = None  # type: ignore
+        self._function_provider: SamFunctionProvider = None  # type: ignore
+        self._env_vars_value: Optional[Dict] = None
+        self._container_env_vars_value: Optional[Dict] = None
+        self._log_file_handle: Optional[IO] = None
+        self._debug_context: Optional[DebugContext] = None
+        self._layers_downloader: Optional[LayerDownloader] = None
+        self._container_manager: Optional[ContainerManager] = None
+        self._lambda_runtimes: Optional[Dict[ContainersMode, LambdaRuntime]] = None
 
-        self._local_lambda_runner = None
+        self._local_lambda_runner: Optional[LocalLambdaRunner] = None
 
-    def __enter__(self):
+    def __enter__(self) -> "InvokeContext":
         """
         Performs some basic checks and returns itself when everything is ready to invoke a Lambda function.
 
@@ -218,7 +223,7 @@ class InvokeContext:
 
         return self
 
-    def __exit__(self, *args):
+    def __exit__(self, *args: Any) -> None:
         """
         Cleanup any necessary opened resources
         """
@@ -230,13 +235,13 @@ class InvokeContext:
         if self._containers_mode == ContainersMode.WARM:
             self._clean_running_containers_and_related_resources()
 
-    def _initialize_all_functions_containers(self):
+    def _initialize_all_functions_containers(self) -> None:
         """
         Create and run a container for each available lambda function
         """
         LOG.info("Initializing the lambda functions containers.")
 
-        def initialize_function_container(function):
+        def initialize_function_container(function: Function) -> None:
             function_config = self.local_lambda_runner.get_invoke_config(function)
             self.lambda_runtime.run(None, function_config, self._debug_context)
 
@@ -256,14 +261,15 @@ class InvokeContext:
             self._clean_running_containers_and_related_resources()
             raise ContainersInitializationException("Lambda functions containers initialization failed") from ex
 
-    def _clean_running_containers_and_related_resources(self):
+    def _clean_running_containers_and_related_resources(self) -> None:
         """
-        Clean the running containers and any other related open resources
+        Clean the running containers and any other related open resources,
+        it is only used when self.lambda_runtime is a WarmLambdaRuntime
         """
-        self.lambda_runtime.clean_running_containers_and_related_resources()
+        cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
 
     @property
-    def function_name(self):
+    def function_name(self) -> str:
         """
         Returns name of the function to invoke. If no function identifier is provided, this method will return name of
         the only function from the template
@@ -291,7 +297,7 @@ class InvokeContext:
         )
 
     @property
-    def lambda_runtime(self):
+    def lambda_runtime(self) -> LambdaRuntime:
         if not self._lambda_runtimes:
             layer_downloader = LayerDownloader(self._layer_cache_basedir, self.get_cwd())
             image_builder = LambdaImage(layer_downloader, self._skip_pull_image, self._force_image_build)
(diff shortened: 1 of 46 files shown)


aahung:
Same as #2623, had canary build on this PR https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-canary-linux/builds/37934677/job/qy3qja68fih0d71m.

Despite there are 5 integ test failures, they are caused by `The security token included in the request is expired`, which is fine because I triggered canary in a non-traditional way and they are not reproducible locally.

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -54,26 +56,26 @@ class InvokeContext:
...
-        shutdown=False,
-    ):
+        template_file: str,
+        function_identifier: Optional[str] = None,
+        env_vars_file: Optional[str] = None,
+        docker_volume_basedir: Optional[str] = None,
+        docker_network: Optional[str] = None,
+        log_file: Optional[str] = None,
+        skip_pull_image: Optional[bool] = None,
+        debug_ports: Optional[Tuple[int]] = None,
```
mndeveci, line 66: Just wondering what is the difference between `Optional[Tuple[int]]` vs `Optional[List[int]]`?

samcli/commands/local/invoke/cli.py
```diff
@@ -162,7 +162,7 @@ def do_cli(  # pylint: disable=R0914
 
             # Invoke the function
             context.local_lambda_runner.invoke(
-                context.function_name, event=event_data, stdout=context.stdout, stderr=context.stderr
+                context.function_identifier, event=event_data, stdout=context.stdout, stderr=context.stderr
```
mndeveci, line 165: Is this the property that we are going to use from now on? And it will have the value like `Stack/FunctionName`?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -279,10 +287,10 @@ def function_name(self):
 
         all_functions = list(self._function_provider.get_all())
         if len(all_functions) == 1:
-            return all_functions[0].functionname
+            return all_functions[0].name
```
mndeveci, line 290: Why we are changing from `functionname` to `name`?

samcli/commands/local/lib/local_lambda.py
```diff
@@ -246,13 +260,18 @@ def _make_env_vars(self, function):
...
+    def _get_session_creds(self) -> Credentials:
         if self._boto3_session_creds is None:
             # to pass command line arguments for region & profile to setup boto3 default session
             LOG.debug("Loading AWS credentials from session with profile '%s'", self.aws_profile)
-            session = boto3.session.Session(profile_name=self.aws_profile, region_name=self.aws_region)
+            # The signature of boto3.session.Session defines the default values of profile_name and region_name
+            # so they should be Optional. But mypy follows its docstring which does not have "Optional."
+            # Here we trick mypy thinking they are both str rather than Optional[str].
+            session = boto3.session.Session(
+                profile_name=cast(str, self.aws_profile), region_name=cast(str, self.aws_region)
```
mndeveci, line 273: Why we are using `cast(str, self.aws_profile)` instead of `str(self.aws_profile)` here?

samcli/lib/providers/api_collector.py
```diff
@@ -145,12 +146,12 @@ def dedupe_function_routes(routes):
...
         Return
         -------
-        A list of routes without duplicate routes with the same function_name and method
+        A list of routes without duplicate routes with the same stack_path, function_name and method
         """
         grouped_routes = {}
 
         for route in routes:
-            key = "{}-{}".format(route.function_name, route.path)
+            key = "{}-{}-{}".format(route.stack_path, route.function_name, route.path)
```
mndeveci, line 154: Will this mean people need to prepend stack name when calling a lambda locally?

samcli/lib/providers/provider.py
```diff
@@ -38,7 +38,7 @@ class Function(NamedTuple):
     codeuri: Optional[str]
     # Environment variables. This is a dictionary with one key called Variables inside it.
     # This contains the definition of environment variables
-    environment: Optional[str]
+    environment: Optional[Dict]
```
mndeveci, line 41: Would this create any issue with backward compatibility? If it was `Dict` when we first introduced types, why it didn't fail at that time but we are updating it with `Dict` now?

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -39,12 +44,17 @@ def __init__(
         :param dict template_dict: SAM Template as a dictionary
         :param dict parameter_overrides: Optional dictionary of values for SAM template parameters that might want
             to get substituted within the template
+        :param dict pseudo_parameter_overrides: Optional dictionary of values for SAM template pseudo parameters
+            that might want to get substituted within the template and its child templates
```
mndeveci, line 48: The difference between these 2 params is, first one will only affect root stack but second one will affect all stacks (including root and its children)? If so why we don't use the first param for that purpose, since we never worked on child stacks and passed parameter overrides to the child stacks before.

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -214,3 +271,13 @@ def find_root_stack(stacks: List[Stack]) -> Stack:
             stacks_str = ", ".join([stack.stack_path for stack in stacks])
             raise ValueError(f"{stacks_str} does not contain a root stack")
         return candidates[0]
+
+    @staticmethod
+    def merge_parameter_overrides(
+        parameter_overrides: Optional[Dict], pseudo_parameter_overrides: Optional[Dict]
+    ) -> Optional[Dict]:
+        if not parameter_overrides and not pseudo_parameter_overrides:
+            return None
```
mndeveci, line 280: Would empty `Dict` be better for this one?

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -214,3 +271,13 @@ def find_root_stack(stacks: List[Stack]) -> Stack:
...
+        merged_parameter_overrides = parameter_overrides or {}
+        merged_parameter_overrides.update(pseudo_parameter_overrides or {})
```
mndeveci, line 282: Can we make sure in class parameters documentation about which parameter has the higher precedence?

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -970,3 +1001,43 @@ def test_unresolved_layer_due_to_bad_instrinsic(self):
...
         self.assertIn(expected_error_output, error_output)
+
+
+class TestInvokeWithFunctionFullPathToAvoidAmbiguity(InvokeIntegBase):
+    template = Path("template-deep-root.yaml")
+    nested_stack_enabled = True
+
+    @pytest.mark.flaky(reruns=3)
+    def test_invoke_with_function_name_will_call_functions_in_top_level_stacks(self):
+        command_list = self.get_command_list("FunctionA", template_path=self.template_path, event_path=self.event_path)
```
mndeveci, line 1012: What will happen if I want to invoke `FunctionB` and it is only defined in one of the child stacks, would that work?
If so can we add the test here?

samcli/commands/local/cli_common/invoke_context.py
aahung: The explanation is hidden in the commit message of f4f767d6a786e4b20705df85c24e13754abb608a:

> change the return value when user did not provide an identifier
from .functionname to .name. This is to make it consistent with
LocalLambda.invoke(), where in line 104
`all_functions = [f.name for f in self.provider.get_all()]`

samcli/commands/local/invoke/cli.py
aahung: yes, function_identifier can be either function logical ID, function name or full path like Stack/FunctionLogicalId

samcli/commands/local/lib/local_lambda.py
aahung: here we know `self.aws_profile` is `str`, `cast(str, self.aws_profile)` is effectively noop in runtime (return `self.aws_profile`) while `str(self.aws_profile)`, according to the doc

> Create a new string object from the given object.

samcli/lib/providers/api_collector.py
aahung: this key does not do anything customer facing. the key is used for dedup only.

samcli/lib/providers/provider.py
aahung: type annotations have no runtime impact and are only checked in CI. 
This was fixing my original type annotation. Should have been `Dict` in the first place 😓

samcli/lib/providers/sam_stack_provider.py
aahung: Each stack (including root and child stacks) has its own `parameter_overrides`. The `parameter_overrides` for nested stacks are defined like this:

```yaml
  SubApp:
    Type: AWS::Serverless::Application
    Properties:
      Location: ../template.yml
      Parameters:
        DefaultTimeout: ...
        MyRuntimeVersion: 'some string'
        StringValueTimeout: !Ref ABCD
```

So we cannot share `parameter_overrides` among all stacks, considering there might be parameter name clash between parent and child stacks.

samcli/lib/providers/sam_stack_provider.py
aahung: Good point, changed

samcli/lib/providers/sam_stack_provider.py
aahung: Good question, they are not supposed to have the same keys. `parameter_overrides` should only contain none-`AWS::*` parameters and `pseudo_parameter_overrides`, as today, can only have `AWS::Region`.

tests/integration/local/invoke/test_integrations_cli.py
aahung: Added two tests, one is `FunctionB` you mentioned. Another one is calling by function name instead of logical ID for function in a child stack.

samcli/commands/local/cli_common/invoke_context.py
aahung: both of them have index access `debug_ports[0]` and tuple is immutable. 

I set it to tuple due to its docstring

> debug_ports tuple(int)
            Ports to bind the debugger to

elbayaaa (dismissed):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -49,24 +57,28 @@ def __init__(
...
             and sorted(self.methods) == sorted(other.methods)
             and self.function_name == other.function_name
             and self.path == other.path
+            and self.stack_path == other.stack_path
         )
 
     def __hash__(self):
-        route_hash = hash(self.function_name) * hash(self.path)
+        # stack_path could be empty string, and hash('') = 0. To avoid that, we add a prefix to stack_path
+        route_hash = hash(self.function_name) * hash(self.path) * hash(f"stack_path:{self.stack_path}")
```
elbayaaa, line 81: I would make it like `hash(F"{self.function_name}{self.path}{self.stack_path}")`

tests/unit/commands/local/lib/swagger/test_parser.py
```diff
@@ -11,17 +11,20 @@
 
 
 class TestSwaggerParser_get_apis(TestCase):
+    def setUp(self) -> None:
+        self.stack_path = Mock()
```
elbayaaa, line 15: why do me mock a String?

samcli/commands/local/lib/local_api_service.py
```diff
@@ -34,9 +34,7 @@ def __init__(self, lambda_invoke_context, port, host, static_dir):
         self.static_dir = static_dir
 
         self.cwd = lambda_invoke_context.get_cwd()
-        self.api_provider = ApiProvider(
-            lambda_invoke_context.template, parameter_overrides=lambda_invoke_context.parameter_overrides, cwd=self.cwd
```
elbayaaa, line 38: why did we remove this `parameter_overrides`?

samcli/lib/providers/api_provider.py
```diff
@@ -51,44 +48,43 @@ def get_all(self):
...
         collector = ApiCollector()
-        provider = self.find_api_provider(resources)
-        provider.extract_resources(resources, collector, cwd=self.cwd)
+        provider = self.find_api_provider(self.stacks)
+        provider.extract_resources(self.stacks, collector, cwd=self.cwd)
         return collector.get_api()
 
     @staticmethod
-    def find_api_provider(resources):
+    def find_api_provider(stacks: List[Stack]):
```
elbayaaa, line 69: The logic here seems off. Different API resources my have different providers. Should this return a resource-provider map?

samcli/local/apigw/local_apigw_service.py
aahung: good call, changed

tests/unit/commands/local/lib/swagger/test_parser.py
aahung: I prefer to mock a string whenever possible because for a string it might "coincidentally get it right" in tests but for `Mock()` object it is guarantee to be unique

samcli/lib/providers/sam_api_provider.py
```diff
@@ -261,8 +279,12 @@ def merge_routes(collector):
...
 
         # By adding implicit APIs to the end of the list, they will be iterated last. If a configuration was already
-        # written by explicit API, it will be overriden by implicit API, just by virtue of order of iteration.
-        all_configs = explicit_routes + implicit_routes
+        # written by explicit API, it will be overridden by implicit API, just by virtue of order of iteration.
+        # Within the explicit/implicit APIs, one defined in top level stack has the higher precedence. Here we
+        # use length of stack_path to order APIs (desc) due to top level stacks always have shorter stack_path.
+        all_configs = sorted(explicit_routes, key=lambda route: -len(route.stack_path)) + sorted(
+            implicit_routes, key=lambda route: -len(route.stack_path)
+        )
```
elbayaaa: as discussed, let's depend on stack depth instead of path-length for consistency

samcli/commands/local/lib/local_api_service.py
aahung: `Stack` object encapsulates `template` and `parameter_overrides`

samcli/lib/providers/sam_api_provider.py
aahung: the commit is pushed

samcli/lib/providers/api_provider.py
aahung: I did not change much for this method. What seems off here?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -154,12 +154,13 @@ def __init__(
...
         self._debug_function = debug_function
 
-        # Note(xinhol): despite self._template_dict and self._function_provider are initialized as None
+        # Note(xinhol): despite self._template_dict, self._function_provider and self._stacks are initialized as None
         # they will be assigned with a non-None value in __enter__() and
         # it is only used in the context (after __enter__ is called)
         # so we can assume they are not Optional here
         self._template_dict: Dict = None  # type: ignore
         self._function_provider: SamFunctionProvider = None  # type: ignore
+        self._stacks: List[Stack] = None  # type: ignore
```
elbayaaa, line 164: I would prefer the `Stack` object to have a List[Stack] property and a getter method to get all stacks and just pass the root stack instead of passing them all

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -365,6 +366,15 @@ def template(self) -> Dict:
         """
         return self._template_dict
```
elbayaaa, line 368: do we want to get rid of this property and replace with the stack everywhere?

samcli/commands/local/cli_common/invoke_context.py
aahung: Storing it in a tree like structure makes sense for readability. Since this is a late stage PR, refactoring now would change a lot of work done in several previous PRs. I would hold this suggestion for now and it is a good future improvement. What do you think?

samcli/lib/providers/api_provider.py
```diff
@@ -25,20 +25,17 @@ def __init__(self, template_dict, parameter_overrides=None, cwd=None):
...
+            List of stacks apis are extracted from
 
         cwd : str
             Optional working directory with respect to which we will resolve relative path to Swagger file
         """
-        self.template_dict = SamBaseProvider.get_template(template_dict, parameter_overrides)
-        self.resources = self.template_dict.get("Resources", {})
-
-        LOG.debug("%d resources found in the template", len(self.resources))
+        self.stacks = stacks
```
elbayaaa, line 34: Did we miss the parameters value substitution in `stack.resources` https://github.com/aws/aws-sam-cli/blob/12efe15b682926b006ce2a846d7cc6c71f5f2d74/samcli/lib/providers/sam_base_provider.py#L162

elbayaaa (commented):
sure, no problem.

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -134,10 +134,11 @@ def __init__(
         self._debugger_path = debugger_path
         self._container_env_vars_file = container_env_vars_file
 
-        self.parameter_overrides = parameter_overrides or {}
+        self._parameter_overrides = parameter_overrides
         # Override certain CloudFormation pseudo-parameters based on values provided by customer
+        self._pseudo_parameter_overrides: Optional[Dict] = None
```
elbayaaa, line 139: why do we need this additional property? why not just adding the AWS::Region to self._parameter_overrides

samcli/lib/providers/api_provider.py
aahung: I did not forgot:

```
@property
def resources(self) -> Dict:
    """
    Return the resources dictionary where SAM plugins have been run
    and parameter values have been substituted.
    """
    processed_template_dict: Dict = SamBaseProvider.get_template(self.template_dict, self.parameters)
    resources: Dict = processed_template_dict.get("Resources", {})
    return resources
```

There are a few integ tests to verify it like `test_invoke_with_env_using_parameters_with_custom_region()`

samcli/commands/local/cli_common/invoke_context.py
aahung: @mndeveci asked the same question above: https://github.com/aws/aws-sam-cli/pull/2631#discussion_r582447560

samcli/commands/local/cli_common/invoke_context.py
aahung: Good catch! just verified it is no longer used

aahung: Also remove the same thing from `BuildContext`, thx!

samcli/lib/providers/sam_stack_provider.py
elbayaaa: Are you trying to say we don't share parameter_overrides between nested stacks but we share pseudo_parameter_overrides? If so, I'm still unclear why not just adding(replicating) AWS::Region to each parameter_overrides

aahung: > If so, I'm still unclear why not just adding(replicating) AWS::Region to each parameter_overrides

aahung: > If so, I'm still unclear why not just adding(replicating) AWS::Region to each parameter_overrides

Yes that's another option. I want to make this more future proof so that in the future, if we want to pass other `AWS::*` pseudo_parameter, we can easily do so.

aahung: Other pseudo_parameters like:

- `AWS::Partitions` ('aws' | 'aws-cn' | 'aws-us-gov')
- `AWS::URLSuffix`
- ... https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html

Right now I only find overrides for `AWS::Region` in sam-cli though.

samcli/lib/providers/api_provider.py
elbayaaa: I understand, I'm saying the whole logic here seems confusing. It seems like we don't really care which provider. Anyway, this is out of scope of this PR

samcli/lib/providers/sam_api_provider.py
```diff
@@ -285,3 +308,17 @@ def merge_routes(collector):
...
+        Returns negative stack depth, used for sorted(routes, _sort_key_stack_depth_desc).
+        Examples:
+            "" (root stack), -depth = 0
+            "A" (1-level nested stack), -depth = -1
+            "A/B/C" (3-level nested stack), -depth = -3
+        """
+
+        if not route.stack_path:
+            return 0
+        return -(len(re.findall(r"/", route.stack_path)) + 1)
```
elbayaaa, line 324: route.stack_path.count("/")

samcli/lib/providers/sam_api_provider.py
```diff
@@ -285,3 +308,17 @@ def merge_routes(collector):
             len(result),
         )
         return list(result)
+
+    @staticmethod
+    def _sort_key_stack_depth_desc(route: Route):
```
elbayaaa, line 313: use the `reverse` flag of `sorted` instead of having explicit descending order

samcli/lib/providers/api_provider.py
aahung: yeah, agree, it is confusing. 

The behavior is:
1. go through all resources until find an API.
2. use the API type to determine the whole start-api behavior
    - if the first found api is  SAM api, we are gonna ignore all CFN APIs
    - vice versa

samcli/lib/providers/sam_api_provider.py
aahung: Thanks, this looks nicer, changed

samcli/lib/providers/sam_api_provider.py
aahung: changed as in previous comment

samcli/local/layers/layer_downloader.py
```diff
@@ -91,7 +97,18 @@ def download(self, layer, force=False):
         """
         if layer.is_defined_within_template:
             LOG.info("%s is a local Layer in the template", layer.name)
-            layer.codeuri = resolve_code_path(self.cwd, layer.codeuri)
+            # the template file containing the layer might not be in the same directory as root template file
+            # therefore we need to join the path of template directory and codeuri in case codeuri is a relative path.
+            stacks = [stack for stack in self._stacks if stack.stack_path == layer.stack_path]
+            if not stacks:
+                raise RuntimeError(f"Cannot find stack that matches layer's stack_path {layer.stack_path}")
+            stack = stacks[0]
```
elbayaaa, line 105: we are sure there is at most one matching stack (no two stacks have the same stack path), so having a list of stacks is confusing, use the `next` method instead. `stack = next(stack for stack in self._stacks if stack.stack_path == layer.stack_path)`

aahung: Should know `next()` earlier, thx, changed.

mndeveci (commented):
This one is a little bit long, let's sync back on Monday to resolve items here.

samcli/commands/local/cli_common/invoke_context.py
aahung: types only checked during linting. So even this code is valid:

```py
list: str = [] # assigning a list to a string-type variable.
```

samcli/lib/providers/sam_stack_provider.py
aahung: Added!

samcli/lib/providers/sam_stack_provider.py
aahung: Now it is called `global_parameter_overrides` now, it is defintely more readable.

samcli/commands/local/cli_common/invoke_context.py
aahung: As discussed offline, we also changed the error message that list all available options to use full_path instead of just the function name

tests/integration/local/invoke/test_integrations_cli.py
aahung: As discussed offline, added one more test to 

> Also call by FunctionSomeLogicalID in test_invoke_with_function_name_will_call_functions_in_top_level_stacks

samcli/commands/local/lib/local_lambda.py
elbayaaa: Are we using cast here, because `self.aws_profile` and `self.aws_region` are **Optional**[str] while mypy wants them to be str? if so, will this work
```
aws_profile = self.aws_profile # May be adding "or None" ??
aws_region = self.aws_region  # May be adding "or None" ??
session = boto3.session.Session(profile_name=aws_profile, region_name=aws_region)
```

samcli/lib/providers/api_provider.py
```diff
@@ -51,44 +48,43 @@ def get_all(self):
...
-            The dictionary containing the different resources within the template
         Returns
         ---------
         An Api from the parsed template
         """
 
         collector = ApiCollector()
-        provider = self.find_api_provider(resources)
-        provider.extract_resources(resources, collector, cwd=self.cwd)
+        provider = self.find_api_provider(self.stacks)
```
elbayaaa, line 64: better to call the static method through the class instead of the instance, i.e., `ApiProvider.find_api_provider(self.stacks)`

aahung: I am not sure which it is better? I don't find doc online regarding which one is better. In official doc, it writes:

> A static method can be called either on the class (such as C.f()) or on an instance (such as C().f()).

In my own opinion, calling static methods in the same class with `self.` provides a hint that this method is in the same file, and it is shorter to write. But I think it does not matter too much, what do you think?

samcli/commands/local/lib/local_lambda.py
aahung: This won't work because it does not eliminate the "nullability," if we want to force `aws_profile` to be `str` instead of `Optional[str]`, there must be one of the below:

- `type: ignore`, like `aws_profile: str = self.aws_profile # type: ignore`
- `cast`, like `aws_profile = cast(str, self.aws_profile)`
- assertion, like `assert(self.aws_profile)` or `if not self.aws_profile: raise SomeException(.....)`

samcli/lib/providers/api_provider.py
elbayaaa: when I read this line at the first glance, I thought `find_api_provider` is an instance method and had the question: why do we send `self.stacks` as a parameter while we already have access to `self`. If it was called with the class name instead, I wouldn't need to go down to the method to realize that it is not an instance method.

elbayaaa (approved):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -49,24 +57,27 @@ def __init__(
...
             isinstance(other, Route)
             and sorted(self.methods) == sorted(other.methods)
             and self.function_name == other.function_name
             and self.path == other.path
+            and self.stack_path == other.stack_path
         )
 
     def __hash__(self):
-        route_hash = hash(self.function_name) * hash(self.path)
+        route_hash = hash(f"{self.stack_path}-{self.function_name}-{self.path}")
```
hoffa, line 80: Not entirely familiar with this, but would this cause redeployments?

aahung: A good question, I cannot find this hash is essentially used other than unit tests. 

First this Route's hash is used in `Api.__hash__` (samcli/lib/providers/provider.py:305), but then noone uses Api's hash.

aahung: And since `Api` or `Route` is not buildable resources (no artifacts), I don't think they would cause redeployment. But the hash seems not used anymore, we can have a followup to try to remove them.

mndeveci (approved):

samcli/lib/providers/api_provider.py
aahung: it makes sense, I will change them in the final PR, there are other places with this usage as well. Thx

aahung: done in https://github.com/aws/aws-sam-cli/pull/2662/commits/31d4d4ad1852d1020cfa67f9f52d5f29ef428136

PR #2638, opened 2021-02-19
fix: Pylint Import Check

CoshUS:
SAM CLI integration tests run with dev.txt requirements.
In the case of a dev library being imported into prod code, the tests will not raise an error. 

By isolating Pylint into its own file, we can create an env with only prod libraries and pylint which enables the ability to check dev imports while linting.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


setup.py
```diff
@@ -8,7 +8,8 @@
 
 def read(*filenames, **kwargs):
     encoding = kwargs.get("encoding", "utf-8")
-    sep = kwargs.get("sep", os.linesep)
```
aahung, line 11: I am not familiar with this file, when will this being executed>?

appveyor.yml
```diff
@@ -61,26 +61,47 @@ for:
 
       # Install AWS CLI Globally outside of a venv.
       - "pip install awscli"
+
```
aahung, line 64: I think an explanation of what envs will be created, what is the motivation, what each env is for here is good for people without context.

aahung (commented):
thanks for adding this check! small feedbacks and questions

setup.py
CoshUS: setup.py is used when pip install is called.
extras_require["pre-dev"] will be used with pip install -e ".[pre-dev]"

appveyor.yml
CoshUS: Added more comments

appveyor.yml
```diff
@@ -61,26 +61,48 @@ for:
 
       # Install AWS CLI Globally outside of a venv.
       - "pip install awscli"
+      
+      # Create pre-dev virtual env without installing dev depenencies
+      # This is used for import check for prod code
```
aahung, line 66:
```suggestion
      # This is used for import check for prod code to prevent importing packages from dev.txt
```

aahung (approved):
LGTM, thx!

appveyor.yml
```diff
@@ -61,26 +61,48 @@ for:
...
       - "python -m virtualenv venv"
       - "venv\\Scripts\\activate"
       - "python --version"
+      - "deactivate"
 
     build_script:
-      # Activate virtualenv again on windows
+      # Install pre-dev dependencies
+      - "pre-dev-venv\\Scripts\\activate"
+      - "python -c \"import sys; print(sys.executable)\""
```
sriram-mv, line 83: nit: we could specifically say which env this is and what is the executable too for when we are debugging builds.

setup.py
sriram-mv: can we verify os.linesep on windows is \n too? I remember some time ago there was a \r\n that crept into it. I could be wrong.

setup.py
```diff
@@ -18,7 +19,16 @@ def read(*filenames, **kwargs):
 
 def read_requirements(req="base.txt"):
     content = read(os.path.join("requirements", req))
-    return [line for line in content.split(os.linesep) if not line.strip().startswith("#")]
+    requirements = list()
+    for line in content.split("\n"):
```
sriram-mv, line 23: same here should we look at os.linesep?

sriram-mv (changes requested):

setup.py
CoshUS: This was changed to "\n" specifically as `io.open` in `read(*filenames, **kwargs)` defaults to "\n" as universal line ending no matter on what system. It was technically broken without an issue previously and has been confirmed on Windows.

CoshUS: https://docs.python.org/3/glossary.html#term-universal-newlines

appveyor.yml
CoshUS: Added comment on its current use.

sriram-mv (approved):

PR #2646, opened 2021-02-22
fix: Go 1.16 set GOFLAGS to allow change of go.mod and go.sum

sriram-mv:
* Go integration tests

* Go 1.16 introduced a breaking change with go build no longer being allowed to change/create go.mod and go.sum files by default.
* Initial Assessment was that setting go env will not fix this, but that been set incorrectly without an "=" sign.

* Set a go env variable to allow for changes.
* None, allows for build tests to start passing again.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-windows-build-go.yml
+++ b/appveyor-windows-build-go.yml
@@ -41,6 +41,11 @@ install:
     - "refreshenv"
     - "go version"
     - "go env"
+    # Set Go to allow modify go.mod and go.sum
+    # Breaking change introduced by https://golang.org/doc/go1.16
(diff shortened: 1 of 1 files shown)


aahung (changes requested):
Thanks for fixing. We also need to do the same for Canary

CoshUS (approved):

sriram-mv:
@aahung : https://github.com/aws/aws-sam-cli/blob/develop/appveyor.yml#L91 (There looks to be a version lock for go)

aahung (approved):

Issue #2613, opened 2021-02-10
`sam validate` fails when running against a SAM docker image


xirkus:
### Description:
The AWS quick start template for a docker image fails `sam validate`.

### Steps to reproduce:
1. Run `sam init`.
2. Select `2 - Image (artifact is an image uploaded to an ECR image repository)` for the package type.
3. Select a base image (I use `5 - amazon/python3.7-base` for my example).
4. Name your application
5. `cd` into your application directory and run `sam build`
6. After the build completes run `sam validate`

### Observed result:

```
% sam validate
Template provided at '/path/to/repo/sam/sam-app/template.yaml' was invalid SAM Template.
Error: [InvalidResourceException('HelloWorldFunction', "'ImageUri' must be set.")] ('HelloWorldFunction', "'ImageUri' must be set.")
```

### Expected result:
`sam validate` should be aware when a Docker image is being used as the deployment package type an validate cleanly. For example, the example project generated will successfully deploy with `sam build`.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOS 11.2.1
2. SAM CLI, version 1.17.0
3. AWS region: us-east-1

xirkus:
NVM. I just realized that the default `template.yaml` does not supply a ECR. Perhaps consider adding this resource:

```
FunctionName:
    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
    Properties:
      PackageType: Image
      ImageUri: !Ref FunctionECR
      Events:
        HelloWorld:
          Type: Api # More info about API Event Source: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#api
          Properties:
            Path: /function
            Method: get
    Metadata:
      DockerTag: python3.7-v1
      DockerContext: ./function
      Dockerfile: Dockerfile
```


It has the advantage of require less setup for starting devs and an ECR that is automatically named.

xirkus:
Hmmm. This seems to be a chicken-egg problem. If you specify an ECR resource in your `template.yaml` it seems that `sam deploy` is unable to recognize it and prompts for `Image Repository for FunctionName:`. I think this fundamentally describes a workflow where an ECR repo is treated like a bucket (which is unconventional as each Docker registry should be associated with a single versioned artifact instead of multiple resources, such as different lambdas, being referenced by tag.

[...]

PR #2647, opened 2021-02-22
fix: allow sam validate to look built template

sriram-mv:
https://github.com/aws/aws-sam-cli/issues/2613

* `sam validate` only looks for current template.yaml which maybe out of date, if a build has been run.

* Directly check for a built template.yaml if present, `sam build` generates this template.

* None
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/validate/validate.py
+++ b/samcli/commands/validate/validate.py
@@ -8,7 +8,7 @@ from botocore.exceptions import NoCredentialsError
 import click
 
 from samcli.cli.main import pass_context, common_options as cli_framework_options, aws_creds_options
-from samcli.commands._utils.options import template_option_without_build
+from samcli.commands._utils.options import template_click_option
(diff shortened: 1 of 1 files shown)


jfuss:
@sriram-mv Is there any kind of tests we need to include for this?

Rabadash8820:
Just came across this PR from aws/serverless-application-model#1907. Any updates? Also out of curiosity, what will be the intended functionality now if we run `sam validate` _before_ `sam build`? Will build be run automatically, or will validate just fail with a more informative error message?

jfuss:
This is solved by https://github.com/aws/aws-sam-cli/pull/2892, from my understanding. If this that is not the case, please open a new Issue, as the one linked by this PR is closed.

Closing

Issue #2578, opened 2021-02-02
feature: Allow specifying Docker build target


mikkelam:
### Description:
I'm trying to use the DockerBuildArgs option to use a specific target in my Dockerfile like so :

```
...
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: .
      DockerBuildArgs:
        target: prod
...
```
However it does not work. I'm getting multiple warnings from docker that build args were not consumed

### Steps to reproduce:
Dockerfile like so
```
FROM public.ecr.aws/lambda/python:3.8 as prod

COPY app.py requirements.txt ./

RUN python3.8 -m pip install -r requirements.txt -t .

# Command can be overwritten by providing a different command in the template directly.
CMD ["app.lambda_handler"]

FROM prod as dev
CMD []
```
along with `sam build`

### Observed result:
```
`[Warning] One or more build-args [target] were not consumed`
```

### Expected result:
I would be able to use a specific target with sam

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macos Big Sur
2. `sam --version`: SAM CLI, version 1.15.0

mgrandis:
It seems the `DockerBuildArgs` param is used to pass parameters to the DockerFile at build time.
That's why you get the warning for unused/not consumed parameter.
We are not supporting specifying a `target` for now but we'll review adding this feature.
This would basically mean allowing a `DockerBuildTarget` element under `Metadata`.

Docker python client reference: https://docker-py.readthedocs.io/en/stable/images.html

mgrandis:
Released in [1.20.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.20.0)

PR #2652, opened 2021-02-23
feat: Allow specifying Docker build target 

jonife:

[#2578](https://github.com/aws/aws-sam-cli/issues/2578)

- DockerBuildArgs param is used to pass parameters to the DockerFile at build time. and SAMCLI doesn't support specifying a target for now.

- Allow DockerBuildTarget element under Metadata in the DockerFile.

- For tempate.yaml with or without a specified Docker build Target, Docker image would build success without a warning.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


mgrandis:
@jonife 
Are you sure `make pr` passed?
The AppVeyor build has failed, can you run it again locally?

tests/unit/lib/build_module/test_app_builder.py
```diff
@@ -739,6 +739,34 @@ def test_can_build_image_function_under_debug(self, mock_os):
...
+        self.builder._build_lambda_image("Name", metadata)
+        self.assertEqual(
+            self.docker_client_mock.api.build.call_args,
+            call(
+                path=ANY,
+                dockerfile="Dockerfile",
+                tag="name:Tag-debug",
+                buildargs={"a": "b", "SAM_BUILD_MODE": "debug"},
+                decode=True,
+                DockerBuildTarget="target"
```
mgrandis, line 767: `DockerBuildTarget` doesn't seem like the right name here.

hoffa (approved):

tests/unit/lib/build_module/test_app_builder.py
```diff
@@ -739,6 +739,35 @@ def test_can_build_image_function_under_debug(self, mock_os):
...
+            "DockerBuildArgs": {"a": "b"},
+            "DockerBuildTarget": "stage",
+        }
+
+        self.docker_client_mock.api.build.return_value = []
+
+        result = self.builder._build_lambda_image("Name", metadata)
+
+        self.assertEqual(result, "name:Tag-debug")
+        self.builder._build_lambda_image("Name", metadata)
```
hoffa, line 758: Is the duplicate call necessary?

sriram-mv: I dont think its necessary.

sriram-mv (approved):
Small nit as pointed out by @hoffa could we address that, other than LGTM

tests/unit/lib/build_module/test_app_builder.py
sriram-mv: could we address this?

sriram-mv (changes requested):

tests/unit/lib/build_module/test_app_builder.py
wchengru: It also appears in another test case: https://github.com/aws/aws-sam-cli/blob/49851f58d5b64f65d48d127ef9a5fc4eccf96ed9/tests/unit/lib/build_module/test_app_builder.py#L729

sriram-mv (approved):

DanielLoyAugmedix:
Is this in the documentation anywhere?

aksel:
Are you considering adding this to the docs? Or is there maybe some way to contribute to the docs?

I was about to do some unnecessary work because this was not documented, but lucky for me I searched a bit first :smile:

Issue #2653, opened 2021-02-23
quiet mode for local invoke


HP4k1h5:
### Describe your idea/feature/enhancement

to cut down on noise when running `sam local invoke` one can redirect stderr like
```
sam local invoke TestLambda 2> /dev/null
```

which removes 
```
TestLayer is a local Layer in the template
Building image............
```
from the output. However if `sam` is not up to date you can still get this message. 

```
SAM CLI update available (1.18.2); (1.18.1 installed)
To download: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html
```
on shared runners etc, its not easy to update `sam`.

### Proposal

this problem can be solved a number of ways. 
1) provide a `--quiet` flag for `sam local invoke` to eliminate all non response/error text in the output
2) change https://github.com/aws/aws-sam-cli/blob/223999e6fec448484a8b24668f81ba7c2c214612/samcli/lib/utils/version_checker.py#L92 to use `err=True` parameter
```
click.echo('Hello World!', err=True)
```

moelasmar:
the PR resolving this issue got merged.

PR #2654, opened 2021-02-23
edit: print outdated version message to stderr

HP4k1h5:
#2653

in sam-based projects, certain tests depend on the output of `sam local invoke`. keeping the output to the function return enables reliable testing outputs

sam cli is using `click.echo` to print messages. `echo` has an `err` flag that  can be used to redirect its output to stderr. setting `err=True` on `echo` calls will allow for usage e.g.
```
sam local invoke Test 2> /dev/null
```
which should help keep the output to the function return value

could break tools that depend on the version_checker stdout

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


moelasmar (approved):

mndeveci (approved):
Thanks for your contribution 🎉
Can you also update the unit tests which is failing at the moment?

HP4k1h5:
> Thanks for your contribution 🎉
> Can you also update the unit tests which is failing at the moment?

thank you for getting to this. 🙏

PR #2675, opened 2021-03-02
Request for Comments: Auto Create ECR Repos in Guided Deploy

CoshUS:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+
+3. Specify image repo: `Image Repository for HelloWorldFunction []: 12345.dkr.ecr.us-west-2.amazonaws.com/helloworldfunctionrepo`
+
+What will be changed?
+---------------------
+
+When deploying with guided, SAM CLI will prompt the option to auto create ECR repos for image based functions.
+The auto created ECR repos will reside in a companion stack that gets deployed along with the actual stack. 
+
+During each guided deploy, the functions and repos will be synced. New repos will be created for new functions and repos without an associating function will be prompt for deletion.
```
sriram-mv, line 24: can we showcase the sync part with more detail?

designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+
+Success criteria for the change
+-------------------------------
+
+* No context switching from SAM CLI to creating resources using other tools.
+
+Out-of-Scope
+------------
+
+* SAM CLI will not manage lifecycles of the created resources outside of SAM CLI, i.e. customer deleting a function from console.
```
sriram-mv, line 36: Lets define what a lifecycle is.

designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+Confirm changes before deploy [y/N]: y
+#SAM needs permission to be able to create roles to connect to the resources in your template
+Allow SAM CLI IAM role creation [Y/n]: y
+Save arguments to configuration file [Y/n]: y
+SAM configuration file [samconfig.toml]:
+SAM configuration environment [default]:
+Looking for resources needed for deployment:
+ S3 bucket: Found! (aws-sam-cli-managed-default-samclisourcebucket-abcdef)
+ Image repositories: Not found.
+ #Managed repositories will be deleted when their functions are removed from the template and deployed
```
sriram-mv, line 65: +1 about being explicit here.

designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+  None:
+    Value: !Sub ${AWS::AccountId}.dkr.ecr.${AWS::Region}.${AWS::URLSuffix}/${TestFunction0150919004Repo}
+
+  None:
+    Value: !Sub ${AWS::AccountId}.dkr.ecr.${AWS::Region}.${AWS::URLSuffix}/${AnotherTestFunction025c2cfd8cRepo}
+```
+
+Documentation Changes
+=====================
+* New option `--resolve-image-repos`. This option will auto create repos without the needs of going through guided experience.
```
sriram-mv, line 291: should we showcase how non interactive usage looks like when specifying `--resolve-image-repos`. I'm assuming it does no destructive actions as opposed to sam deploy --guided which prompts to delete repos when functions dont exist.

sriram-mv (changes requested):

designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+-------------------------------
+
+* No context switching from SAM CLI to creating resources using other tools.
+
+Out-of-Scope
+------------
+
+* SAM CLI will not manage lifecycles of the created resources outside of SAM CLI, i.e. customer deleting a function from console.
+* Auto create repo only concerns about guided experience. Repos are assumed to be provided in CI/CD situations. However the option --resolve-image-repos will be added for auto creating repos without going through guided.
+
```
metaskills, line 38: FWIW, I agree that "Repos are assumed to be provided in CI/CD situations." and I've been doing the same in GitHub Actions for us.

designs/auto-create-ecr.md
```diff
@@ -0,0 +1,306 @@
...
+What will be changed?
+---------------------
+
+When deploying with guided, SAM CLI will prompt the option to auto create ECR repos for image based functions.
+The auto created ECR repos will reside in a companion stack that gets deployed along with the actual stack. 
+
+During each guided deploy, the functions and repos will be synced. New repos will be created for new functions and repos without an associating function will be prompt for deletion.
+
+There will be an escape hatch to use non SAM CLI managed repos by specifying `--image-repositories` or change `samconfig.toml`.
```
metaskills, line 26: Would `--image-repository` also be an escape hatch or is that interface deprecated now?

metaskills (commented):
Thanks for asking and really appreciate the work. I've certainly moved more to the side of solving this with CI/CD vs in SAM, but also think this proposal is needed for new adopters to avoid the manual shake and bake :)

designs/auto-create-ecr.md
CoshUS: Added more details.

designs/auto-create-ecr.md
CoshUS: Updated.

designs/auto-create-ecr.md
CoshUS: --resolve-image-repos will be the same as guided experience including destructive actions.

1. This will keep a consistent behavior as the guided. In guided, SAM CLI will abort deployment if deletion of auto created image repos is denied.
2. For UX, this will avoid image repos and functions mapping to an invalid state where orphaned image repos exist. For this case, we will also need to track which repos should be kept and makes the sync less robust.
3. From a security perspective, keeping old image repos will increase the impact radius of an information leakage. A customer might expect a sensitive image repo will be deleted as soon as the function itself is also removed like the guided experience.

designs/auto-create-ecr.md
CoshUS: `--image-repository` can still be used as an escape hatch.

sriram-mv (approved):
Overall looks good!

jfuss (approved):

PR #2677, opened 2021-03-03
chore: Add docstring to all functions with 30+ LOC

aahung:

We often forgot to write docstring for a new function/method, which is very helpful for other people to understand the code.

This PR re-enables pylint rule to check functions/methods with 30+ LOC and enforce them to have a docstring. I set it to 30 because there are too many short functions/methods lacking docstring and I couldn't add them all. So let's start with 30. 

No code change. 

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -80,7 +80,7 @@ disable=
     # We will remove them one by one.
     # W9005, # "%s" has constructor parameters documented in class and __init__
     W9006, # "%s" not documented as being raised
-    W9008, # Redundant returns documentation
+    # W9008, # Redundant returns documentation
     # W9010, # Redundant yields documentation
     W9011, # Missing return documentation
     W9012, # Missing return type documentation
--- a/.pylintrc
+++ b/.pylintrc
@@ -84,8 +84,8 @@ disable=
     # W9010, # Redundant yields documentation
     W9011, # Missing return documentation
     W9012, # Missing return type documentation
-    W9013, # Missing yield documentation
-    W9014, # Missing yield type documentation
+    # W9013, # Missing yield documentation
+    # W9014, # Missing yield type documentation
     # W9015, # "%s" missing in parameter documentation
     W9016, # "%s" missing in parameter type documentation
     # W9017, # "%s" differing in parameter documentation
--- a/.pylintrc
+++ b/.pylintrc
@@ -74,6 +74,7 @@ disable=
     R0914, # Too many local variables (%s/%s)
     R0915, # Too many statements
     C0415, # Import outside toplevel (%s) Used when an import statement is used anywhere other than the module toplevel. Move this import to the top of the file.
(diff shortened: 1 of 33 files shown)


samcli/commands/deploy/deploy_context.py
```diff
@@ -174,6 +177,42 @@ def deploy(
...
+        tags : List[str]
+            List of tags passed to CloudFormation
+        region : str
+            AWS region to deploy the stack to
+        fail_on_empty_changeset : bool
+            Should fail when changeset is empty
+        confirm_changeset : bool
+            Should wait for customer's confirm before executing the changeset
+
+        Returns
```
mgrandis, line 212: Not sure this is necessary if the method doesn't return anything (VSCode doesn't add it)

aahung: Good question. I don't find standard about this one in the official doc. https://numpydoc.readthedocs.io/en/latest/format.html

In PyCharm, Returns is added even it returns `None`. And this stackoverflow post suggests to leave a None here with 42 upvotes

https://stackoverflow.com/questions/27784179/docstrings-when-nothing-is-returned

> You should use None, as that is what your function actually returns:

aahung: update, I think we should remove it, I looked into the numpy source code and numpy itself does not add "Returns None" 

https://github.com/numpy/numpy/blob/5cae51e794d69dd553104099305e9f92db237c53/numpy/lib/format.py#L455-L468

```py
def write_array_header_2_0(fp, d):
    """ Write the header for an array using the 2.0 format.
        The 2.0 format allows storing very large structured arrays.
    .. versionadded:: 1.9.0
    Parameters
    ----------
    fp : filelike object
    d : dict
        This has the appropriate entries for writing its string
        representation to the header of the file.
    """
    _write_array_header(fp, d, (2, 0))
```

.pylintrc
```diff
@@ -74,6 +74,7 @@ disable=
     R0914, # Too many local variables (%s/%s)
     R0915, # Too many statements
     C0415, # Import outside toplevel (%s) Used when an import statement is used anywhere other than the module toplevel. Move this import to the top of the file.
+    C0115, # missing-class-docstring
```
hoffa, line 77: Why does this have to be disabled?

aahung: I won't be able to add docstring to all the classes

hoffa (approved):
LGTM, thanks! 🙌

samcli/commands/deploy/deploy_context.py
aahung: There is actually a pylint rule for this "redundant-returns-doc"

All fixed

mgrandis (approved):

PR #2679, opened 2021-03-03
[Draft] Use-case driven init flow

sapessi:
**do not merge [until the new use case field in the templates manifest is merged](https://github.com/aws/aws-sam-cli-app-templates/pull/90)**

Customers approach the SAM CLI with a use-case in mind first and foremost. Today, we may have init templates for their use-case only for a subset of the runtimes and customers won't be able to discover them unless they are lucky enough to select the correct runtime in the first option.

This changes the `init` flow for the SAM CLI to first present a list of use-cases and then give access to the templates we have for each one. The entire init menu is dynamically generated from the `manifest.json` file in the app templates repository

We rely on GitHub to be available for the `init` flow to work

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from typing import Dict
 
 import click
+import requests
 
 from samcli.cli.main import global_cfg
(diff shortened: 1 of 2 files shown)


jfuss:
Closing in favor of #2958

PR #2680, opened 2021-03-03
WIP ci: Make tests run in parallel

aahung:

CI check is slow, especially canary.

image

Run tests in parallel (4 workers), for integration tests, 2 workers according to [AppVeyor resources](https://www.appveyor.com/docs/build-environment/) (2 or 4 cores)

Tests in the same module are run in sequential via `--dist loadfile` to avoid situation like stack name reuse in deployment tests. This is based on the assumption tests in two files won't affect each other, which at the first scan, I did not find any. (For example, all deployment integ tests are in the same file)

> --dist loadfile: Tests are grouped by their containing file. Groups are distributed to available workers as whole units. This guarantees that all tests in a file run in the same worker. https://pypi.org/project/pytest-xdist/

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -169,7 +169,11 @@ for:
         if [[ -n $BY_CANARY ]] && [[ -n $DOCKER_USER ]] && [[ -n $DOCKER_PASS ]];
           then echo Logging in Docker Hub; echo $DOCKER_PASS | docker login --username $DOCKER_USER --password-stdin;
         fi"
-      - sh: "pytest -vv -n 2 --dist loadfile tests/integration"
+      # warm container integration tests might be interfered by other tests using docker.
+      # To grant exclusive access to docker, here we run warm container integration tests in sequential,
+      # then run everything else in parallel.
+      - sh: "pytest -vv -k 'TestWarmContainers' tests/integration"
+      - sh: "pytest -vv -n 2 --dist loadfile -k 'not TestWarmContainers' tests/integration"
       - sh: "pytest -vv -n 4 --dist loadfile tests/regression"
       - sh: "black --check setup.py tests samcli"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -99,7 +99,7 @@ for:
 
       # Run tests with dev env
       - "venv\\Scripts\\activate"
-      - "pytest --cov samcli --cov-report term-missing --cov-fail-under 94 -n 4 --dist loadscope tests/unit"
+      - "pytest --cov samcli --cov-report term-missing --cov-fail-under 94 -n 4 --dist loadfile tests/unit"
       - "pylint --rcfile .pylintrc samcli"
       - "mypy setup.py samcli tests"
       - "pytest -n 4 tests/functional"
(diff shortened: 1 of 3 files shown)


tests/unit/local/docker/test_lambda_container.py
```diff
@@ -511,7 +511,8 @@ def test_must_provide_container_env_vars(self, runtime):
 
         self.assertIsNotNone(container_env_vars)
 
-    @parameterized.expand([param(r) for r in set(RUNTIMES_WITH_BOOTSTRAP_ENTRYPOINT)])
+    # sort the parameters is essential making unit test names deterministic, which is required by pytest-xdist
+    @parameterized.expand([param(r) for r in sorted(set(RUNTIMES_WITH_BOOTSTRAP_ENTRYPOINT))])
```
aahung, line 515: @mndeveci this is the fix I mentioned.

`pytest-xdist` needs to collect tests in each worker and raise an exception if the collection is not deterministic. Here we sort to ensure each worker can collect the same collection of tests.

mndeveci: Do we need this kind of sorts anywhere in the codebase? Like the ones in `integration/deployment`?

aahung: only where we have `set(...)`, list is fine as long as it has an order

appveyor.yml
```diff
@@ -169,8 +169,12 @@ for:
         if [[ -n $BY_CANARY ]] && [[ -n $DOCKER_USER ]] && [[ -n $DOCKER_PASS ]];
           then echo Logging in Docker Hub; echo $DOCKER_PASS | docker login --username $DOCKER_USER --password-stdin;
         fi"
-      - sh: "pytest -vv tests/integration"
-      - sh: "pytest -vv tests/regression"
+      # warm container integration tests might be interfered by other tests using docker.
+      # To grant exclusive access to docker, here we run warm container integration tests in sequential,
+      # then run everything else in parallel.
+      - sh: "pytest -vv -k 'TestWarmContainers' tests/integration"
+      - sh: "pytest -vv -n 2 --dist loadfile -k 'not TestWarmContainers' tests/integration"
```
jfuss, line 176: Why isn't this in the makefile?

Are we sure this works as we expect? I remember long ago we ran into problems with parallel tests because build verifies the containers were removed.

How are we handling throttle from services like CFN in our tests?

jfuss (changes requested):

appveyor.yml
aahung: > Why isn't this in the makefile?

- [ ] Make makefile and appveyor.yaml consistent

> Are we sure this works as we expect? I remember long ago we ran into problems with parallel tests because build verifies the containers were removed.

good call out, I excluded build integration tests when I tested it locally. 

> How are we handling throttle from services like CFN in our tests?

2 workers should not hit the throttle limit


**I am gonna convert this back to draft, looks like there are concerns over build integ tests. The canary test run is scheduled later today, will update later.**

aahung:
Experienced build integ issue as @jfuss mentioned, need to find a solution

image

aahung:
Closing for now, this approach doesn't work out of the box, we need a solution to find tests that can be parallelized and only parallelized them

PR #2690, opened 2021-03-05
fix: CodeUri is not resolvable in nested stacks not located with root template

aahung:

#2692

Fix a bug causing relative resource CodeUri is not resolvable when the resources is defined in a nested stack whose template file is located in a different directory from root template.

For example, 

```
            root stack                -> template.yaml
            child stack               -> folder/template.yaml
            a resource in child stack -> folder/resource
```

The CodeUri of the resource is "./resource" (or without `./`). When sam tries to build it, it assume every codeuri is relative to the root template, which results in build failure. 

When creating `Function` and `LayerVersion`, the `CodeUri` is replaced using `SamLocalStackProvider.normalize_resource_path()`.

This change applies to `Function` and `LayerVersion`'s `codeuri` and `Function`'s `metadata.DockerContext`.

Originally, in `local_labmda` and `layer_download` handles this problem themselves, now we have a universal solution.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/sam_function_provider.py
+++ b/samcli/lib/providers/sam_function_provider.py
@@ -290,7 +290,9 @@ class SamFunctionProvider(SamBaseProvider):
         """
         metadata = resource_properties.get("Metadata", None)
         if metadata and "DockerContext" in metadata:
-            metadata["DockerContext"] = SamLocalStackProvider.normalize_resource_path(stack, metadata["DockerContext"])
+            metadata["DockerContext"] = SamLocalStackProvider.normalize_resource_path(
+                stack.location, metadata["DockerContext"]
+            )
 
         return Function(
             stack_path=stack.stack_path,
@@ -301,7 +303,7 @@ class SamFunctionProvider(SamBaseProvider):
             memory=resource_properties.get("MemorySize"),
             timeout=resource_properties.get("Timeout"),
             handler=resource_properties.get("Handler"),
-            codeuri=SamLocalStackProvider.normalize_resource_path(stack, codeuri) if codeuri else None,
+            codeuri=SamLocalStackProvider.normalize_resource_path(stack.location, codeuri) if codeuri else None,
             imageuri=imageuri if imageuri else resource_properties.get("ImageUri"),
             imageconfig=resource_properties.get("ImageConfig"),
             environment=resource_properties.get("Environment"),
@@ -388,7 +390,7 @@ class SamFunctionProvider(SamBaseProvider):
                 layers.append(
                     LayerVersion(
                         layer_logical_id,
-                        SamLocalStackProvider.normalize_resource_path(stack, codeuri) if codeuri else None,
+                        SamLocalStackProvider.normalize_resource_path(stack.location, codeuri) if codeuri else None,
                         compatible_runtimes,
                         layer_resource.get("Metadata", None),
                         stack_path=stack.stack_path,
--- a/samcli/lib/providers/sam_function_provider.py
+++ b/samcli/lib/providers/sam_function_provider.py
@@ -10,6 +10,7 @@ from samcli.lib.utils.colors import Colored
 from samcli.lib.utils.packagetype import ZIP, IMAGE
 from .provider import Function, LayerVersion, Stack
 from .sam_base_provider import SamBaseProvider
+from .sam_stack_provider import SamLocalStackProvider
 
 LOG = logging.getLogger(__name__)
 
@@ -121,13 +122,12 @@ class SamFunctionProvider(SamBaseProvider):
 
                 if resource_type == SamFunctionProvider.SERVERLESS_FUNCTION:
                     layers = SamFunctionProvider._parse_layer_info(
-                        stack.stack_path,
+                        stack,
                         resource_properties.get("Layers", []),
-                        stack.resources,
                         ignore_code_extraction_warnings=ignore_code_extraction_warnings,
                     )
                     function = SamFunctionProvider._convert_sam_function_resource(
-                        stack.stack_path,
+                        stack,
                         name,
                         resource_properties,
                         layers,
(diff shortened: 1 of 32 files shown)


samcli/lib/providers/sam_stack_provider.py
```diff
@@ -290,3 +290,27 @@ def merge_parameter_overrides(
...
+        stack
+        path
+
+        Returns
+        -------
+        str
+            the normalized path relative to root stack
+
+        """
+        return path if os.path.isabs(path) else os.path.normpath(os.path.join(os.path.dirname(stack.location), path))
```
hoffa, line 316: 1. A bit hard to follow with everything on one line.
2. Can we use `pathlib`? It should normalize paths by default which should simplify this.

samcli/lib/providers/sam_layer_provider.py
```diff
@@ -105,8 +104,8 @@ def _convert_lambda_layer_resource(
 
         return LayerVersion(
             layer_logical_id,
-            codeuri,
+            codeuri and SamLocalStackProvider.normalize_resource_path(stack, codeuri),
```
hoffa, line 107: Why is the `codeuri and ...` needed?

samcli/lib/providers/sam_stack_provider.py
aahung: `Path()` does not always normalize, for example, `Path('./a', '..')` would give `a/..` while `os.path.normpath(os.path.join('./a', '..'))` gives `.`

aahung: made it 2 line!

samcli/lib/providers/sam_layer_provider.py
aahung: `codeuri` could be `None`, so I need to check it first.

hoffa: Would it make sense to check that in `normalize_resource_path` instead?

aahung: I made it consistent with built in path functions, such as `os.path.join(a, b)` or `Path(a, b)`, functions with "path" in their name do not accept `None` as an argument.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -277,32 +288,35 @@ def _build_function_configuration(
...
+            stack_path=stack.stack_path,
             name=name,
             functionname=resource_properties.get("FunctionName", name),
             packagetype=resource_properties.get("PackageType", ZIP),
             runtime=resource_properties.get("Runtime"),
             memory=resource_properties.get("MemorySize"),
             timeout=resource_properties.get("Timeout"),
             handler=resource_properties.get("Handler"),
-            codeuri=codeuri,
+            codeuri=codeuri and SamLocalStackProvider.normalize_resource_path(stack, codeuri),
```
sriram-mv, line 304:
```
SamLocalStackProvider.normalize_resource_path(stack, codeuri) if codeuri else None
```

would that be more descriptive? is that what you are going for?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -374,10 +388,10 @@ def _parse_layer_info(
                 layers.append(
                     LayerVersion(
                         layer_logical_id,
-                        codeuri,
+                        codeuri and SamLocalStackProvider.normalize_resource_path(stack, codeuri),
```
sriram-mv, line 391: same here

sriram-mv (changes requested):

samcli/lib/providers/sam_function_provider.py
aahung: ok, makes sense, consistent with `imageuri=imageuri if imageuri else resource_properties.get("ImageUri"),`

hoffa (approved):

kornicameister:
hello everyone,
I am playing around with this PR because it will make some (or rather a lot of things) simpler on my end. Works like a charm apart from one thing. The way our templates are organized is via symlinks resembling some like

```
stack_1
stack_2
stack_3
  stacks:
    - stack_1
    - stack_2
```

Reason for that is that we can reference supplemental/extra sources with the same S3 object prefix and reuse those in templates. However even if I can do `Location: ../stack_1/template.yml` I cannot do `Location: ./stack_1/template.yml` (i.e. symlink).  

Perhaps this PR could be enhanced with resolving symlinks?

aahung:
> hello everyone,
> I am playing around with this PR because it will make some (or rather a lot of things) simpler on my end. Works like a charm apart from one thing. The way our templates are organized is via symlinks resembling some like
> 
> ```
> stack_1
> stack_2
> stack_3
>   stacks:
>     - stack_1
>     - stack_2
> ```
> 
> Reason for that is that we can reference supplemental/extra sources with the same S3 object prefix and reuse those in templates. However even if I can do `Location: ../stack_1/template.yml` I cannot do `Location: ./stack_1/template.yml` (i.e. symlink).
> 
> Perhaps this PR could be enhanced with resolving symlinks?

Thanks for reporting, it will be fixed by https://github.com/aws/aws-sam-cli/pull/2697

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1585,8 +1588,59 @@ def test_nested_build(self, use_container, cached, parallel):
...
+            - FunctionB
+            - ChildStackY
+                - template.yaml
+                - FunctionA
+                - MyLayerVersion
+    """
+
+    template = os.path.join("deep-nested", "template.yaml")
+
+    # @pytest.mark.flaky(reruns=3)
```
sriram-mv, line 1608: was this test supposed to be flaky? if not lets remove this comment.

sriram-mv (approved):
minor nit: otherwise looks good.

tests/integration/buildcmd/test_build_cmd.py
aahung: good catch! I commented out it for local testing purpose, let me bring it back! thx

aahung:
Accidentally put a commit in #2697 here, just reverted.

aahung:
@kornicameister update: the symbolic link is supported in #2697, will be out in the next release.

kornicameister:
Awesome. Thanks a lot.

Issue #2689, opened 2021-03-05
Deploying nested application gives false error message


lvlattanzio:
### Description:
Prior to updating my aws-sam-cli to version 1.20.0 I had a stack with several nested stacks from the local filesystem.  This was done primarily to break up my template into more manageable chunks.  Now that I upgraded to v1.20.0 I am receiving the following error upon running `sam deploy`:
`Nested application 'Application' has specified S3 location for Location which is unsupported. Skipping resources inside this nested application.`

### Steps to reproduce:
1. `sam init`
```
Which template source would you like to use?
        1 - AWS Quick Start Templates
        2 - Custom Template Location
Choice: 1
What package type would you like to use?
        1 - Zip (artifact is a zip uploaded to S3)
        2 - Image (artifact is an image uploaded to an ECR image repository)
Package type: 1

Which runtime would you like to use?
        1 - nodejs14.x
        2 - python3.8
        3 - ruby2.7
        4 - go1.x
        5 - java11
        6 - dotnetcore3.1
        7 - nodejs12.x
        8 - nodejs10.x
        9 - python3.7
        10 - python3.6
        11 - python2.7
        12 - ruby2.5
        13 - java8.al2
        14 - java8
        15 - dotnetcore2.1
Runtime: 7

Project name [sam-app]: 

Cloning app templates from https://github.com/aws/aws-sam-cli-app-templates

AWS quick start application templates:
        1 - Hello World Example
        2 - Step Functions Sample App (Stock Trader)
        3 - Quick Start: From Scratch
        4 - Quick Start: Scheduled Events
        5 - Quick Start: S3
        6 - Quick Start: SNS
        7 - Quick Start: SQS
        8 - Quick Start: Web Backend
Template selection: 1

    -----------------------
    Generating application:
    -----------------------
    Name: sam-app
    Runtime: nodejs12.x
    Dependency Manager: npm
    Application Template: hello-world
    Output Directory: .
    
    Next steps can be found in the README file at ./sam-app/README.md
```
2. clone template.yaml (e.g., template.1.yaml) and point to it in template.yaml:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  sam-app

  Sample SAM Template for sam-app

Resources:
  Application:
    Type: AWS::Serverless::Application
    Properties:
      Location: ./template.1.yaml
```
3. ```sam build && sam deploy --debug```

### Observed result:
```
Initiating deployment
[...]

PR #2691, opened 2021-03-05
fix: Only outputing remote stack warning during building

aahung:

#2689

Error message `Nested application 'Application' has specified S3 location for Location which is unsupported. Skipping resources inside this nested application.` is printed when deploying, which is misleading. 

Add `ignore_code_extraction_warnings` to stack provider like the one in function provider.

No 

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/sam_stack_provider.py
+++ b/samcli/lib/providers/sam_stack_provider.py
@@ -3,11 +3,12 @@ Class that provides all nested stacks from a given SAM template
 """
 import logging
 import os
-from typing import Optional, Dict, cast, List, Iterator
+from typing import Optional, Dict, cast, List, Iterator, Tuple
 from urllib.parse import unquote, urlparse
 
 from samcli.commands._utils.template import get_template_data
-from samcli.lib.providers.provider import Stack
+from samcli.lib.providers.exceptions import RemoteStackLocationNotSupported
+from samcli.lib.providers.provider import Stack, get_full_path
 from samcli.lib.providers.sam_base_provider import SamBaseProvider
 
 LOG = logging.getLogger(__name__)
@@ -56,8 +57,11 @@ class SamLocalStackProvider(SamBaseProvider):
 
         LOG.debug("%d stacks found in the template", len(self._resources))
 
-        # Store a map of stack name to stack information for quick reference
-        self._stacks = self._extract_stacks()
+        # Store a map of stack name to stack information for quick reference -> self._stacks
+        # and detect remote stacks -> self._remote_stack_full_paths
+        self._stacks: Dict[str, Stack] = {}
+        self.remote_stack_full_paths: List[str] = []
+        self._extract_stacks()
 
     def get(self, name: str) -> Optional[Stack]:
         """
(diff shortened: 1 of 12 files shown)


samcli/lib/providers/sam_stack_provider.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         template_dict: Dict,
         parameter_overrides: Optional[Dict] = None,
         global_parameter_overrides: Optional[Dict] = None,
+        ignore_code_extraction_warnings=False,
```
qingchm, line 30: Are we missing the type check here? How did make pr pass haha

qingchm:
Is it possible to update the test cases so we have a better coverage since this new parameter creates new branches? Is is possible to have a test case to verify that we do not log the warning when we have this field on?

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -107,11 +110,21 @@ def _extract_stacks(self) -> Dict[str, Stack]:
...
+                    self._ignore_code_extraction_warnings,
                 )
             if resource_type == SamLocalStackProvider.CLOUDFORMATION_STACK:
                 stack = SamLocalStackProvider._convert_cfn_stack_resource(
-                    self._template_directory, self._stack_path, name, resource_properties
+                    self._template_directory,
+                    self._stack_path,
+                    name,
+                    resource_properties,
+                    self._global_parameter_overrides,
```
aahung, line 126: this was previously `None` for `global_parameter_overrides`. It should have been `self._global_parameter_overrides`. I added it just for convenience. 

The `global_parameter_overrides` does not make a difference in our code because all `Stack` objects that are actually used by other classes are created in `get_stacks()`. I will simplify this class in the future. For now, it does not matter whether `global_parameter_overrides` is provided or not.

aahung:
> Is it possible to update the test cases so we have a better coverage since this new parameter creates new branches? Is is possible to have a test case to verify that we do not log the warning when we have this field on?

good idea, just added! thx

qingchm (approved):

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -170,15 +186,17 @@ def _convert_cfn_stack_resource(
...
     ) -> Optional[Stack]:
         template_url = resource_properties.get("TemplateURL", "")
 
         if SamLocalStackProvider.is_remote_url(template_url):
-            LOG.warning(
-                "Nested stack '%s' has specified S3 location for Location which is unsupported. "
-                "Skipping resources inside this nested stack.",
-                name,
-            )
+            if not ignore_code_extraction_warnings:
```
hoffa, line 194: Feels weird that the method that's dealing with all the core logic has to leak its abstractions and ask callers to tell it whether there's something to warn about.

aahung: I changed the implementation to not let stack provider aware of any warning. now its `get_stacks()` returns a tuple instead of a single list. The 2nd item of the tuple indicates the list of remote stacks detected, so that build context can warn accordingly.

hoffa (approved):

aahung:
updated error message:

```
Below nested stacks(s) specify non-local URL(s), which is unsupported:
- StackRemote
Skipping building resources inside this these nested stacks.
```

PR #2698, opened 2021-03-10
chore: Add typehint to samcli.lib.providers.*

aahung:

Add type hint to all files in module "samcli.lib.providers"

No behavior change. 

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/cfn_api_provider.py
+++ b/samcli/lib/providers/cfn_api_provider.py
@@ -1,12 +1,13 @@
 """Parses SAM given a template"""
 import logging
-from typing import List
+from typing import Any, Dict, Optional, Tuple, List, cast
 
 from samcli.commands.local.lib.swagger.integration_uri import LambdaUri
 from samcli.lib.providers.provider import Stack
 from samcli.local.apigw.local_apigw_service import Route
 from samcli.commands.local.cli_common.user_exceptions import InvalidSamTemplateException
 from samcli.lib.providers.cfn_base_api_provider import CfnBaseApiProvider
+from samcli.lib.providers.api_collector import ApiCollector
 
 LOG = logging.getLogger(__name__)
 
@@ -33,7 +34,7 @@ class CfnApiProvider(CfnBaseApiProvider):
         APIGATEWAY_V2_STAGE,
     ]
 
-    def extract_resources(self, stacks: List[Stack], collector, cwd=None):
+    def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: Optional[str] = None) -> None:
         """
         Extract the Route Object from a given resource and adds it to the RouteCollector.
 
@@ -47,10 +48,6 @@ class CfnApiProvider(CfnBaseApiProvider):
 
         cwd : str
             Optional working directory with respect to which we will resolve relative path to Swagger file
-
-        Return
-        -------
-        Returns a list of routes
         """
 
         for stack in stacks:
@@ -75,13 +72,14 @@ class CfnApiProvider(CfnBaseApiProvider):
                 if resource_type == CfnApiProvider.APIGATEWAY_V2_STAGE:
                     self._extract_cfn_gateway_v2_stage(resources, resource, collector)
 
-        all_apis = []
-        for _, apis in collector:
-            all_apis.extend(apis)
-        return all_apis
-
     @staticmethod
-    def _extract_cloud_formation_route(stack_path: str, logical_id, api_resource, collector, cwd=None):
+    def _extract_cloud_formation_route(
+        stack_path: str,
+        logical_id: str,
+        api_resource: Dict[str, Any],
+        collector: ApiCollector,
+        cwd: Optional[str] = None,
+    ) -> None:
         """
         Extract APIs from AWS::ApiGateway::RestApi resource by reading and parsing Swagger documents. The result is
         added to the collector.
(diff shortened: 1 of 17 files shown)


samcli/lib/providers/cfn_api_provider.py
```diff
@@ -75,13 +72,14 @@ def extract_resources(self, stacks: List[Stack], collector, cwd=None):
                 if resource_type == CfnApiProvider.APIGATEWAY_V2_STAGE:
                     self._extract_cfn_gateway_v2_stage(resources, resource, collector)
 
-        all_apis = []
-        for _, apis in collector:
-            all_apis.extend(apis)
-        return all_apis
```
aahung, line 81: we never used the return value of `extract_resources()`, removing the return value makes this method consistent with the ones in the super class and another subclass.

samcli/lib/providers/provider.py
```diff
@@ -169,14 +178,12 @@ def _compute_layer_version(is_defined_within_template, arn):
 
         try:
             _, layer_version = arn.rsplit(":", 1)
-            layer_version = int(layer_version)
+            return int(layer_version)
```
aahung, line 181: explain: avoid changing the type of `layer_version` after initialization

samcli/lib/providers/sam_api_provider.py
```diff
@@ -294,12 +301,9 @@ def merge_routes(collector):
             # method on explicit route.
             for normalized_method in config.methods:
                 key = config.path + normalized_method
-                if (
-                    all_routes.get(key)
-                    and all_routes.get(key).payload_format_version
```
aahung, line 299: explain: mypy cannot remember the nullability of `all_routes.get(key)`

samcli/lib/providers/sam_base_provider.py
```diff
@@ -64,12 +65,12 @@ def _extract_lambda_function_code(resource_properties, code_property_key):
         codeuri = resource_properties.get(code_property_key, SamBaseProvider.DEFAULT_CODEURI)
 
         if isinstance(codeuri, dict):
-            codeuri = SamBaseProvider.DEFAULT_CODEURI
+            return SamBaseProvider.DEFAULT_CODEURI
```
aahung, line 68: explain: avoid changing the type of `codeuri` during re-assignment

samcli/lib/providers/sam_base_provider.py
```diff
@@ -132,18 +136,18 @@ def _extract_sam_function_codeuri(
...
                 LOG.warning(
                     "Lambda function '%s' has specified S3 location for CodeUri which is unsupported. "
                     "Using default value of '%s' instead",
                     name,
-                    codeuri,
+                    SamBaseProvider.DEFAULT_CODEURI,
                 )
-        return codeuri
+            return SamBaseProvider.DEFAULT_CODEURI
+        return cast(str, codeuri)
```
aahung, line 147: explain: avoid changing the type of `codeuri` during re-assignment

hoffa (approved):
LGTM! Thanks for working on adding these!

samcli/lib/providers/cfn_api_provider.py
elbayaaa: returning the extracted resources here and in the other methods as well makes more sense to me, but it is fine if you are sure this won't break anything

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -382,12 +412,15 @@ def _get_integration_function_name(integration):
 
         if integration and isinstance(integration, dict):
             # Integration must be "aws_proxy" otherwise we don't care about it
-            return LambdaUri.get_function_name(integration.get("Uri"))
+            uri: str = cast(str, integration.get("Uri"))
+            return LambdaUri.get_function_name(uri)
```
elbayaaa, line 416: why did we split this in to two lines? did mypy complain?

samcli/lib/providers/sam_api_provider.py
```diff
@@ -207,15 +215,14 @@ def _convert_event_route(stack_path: str, lambda_logical_id, event_properties, e
...
-        path = event_properties.get(SamApiProvider._EVENT_PATH)
-        method = event_properties.get(SamApiProvider._EVENT_METHOD)
+        path = cast(str, event_properties.get(SamApiProvider._EVENT_PATH))
+        method = cast(str, event_properties.get(SamApiProvider._EVENT_METHOD))
 
         # An RESTAPI (HTTPAPI) Event, can have RestApiId (ApiId) property which designates the resource that owns this
         # API. If omitted, the API is owned by Implicit API resource. This could either be a direct resource logical ID
         # or a "Ref" of the logicalID
 
-        api_resource_id = None
```
elbayaaa, line 217: why did we remove this?

samcli/lib/providers/cfn_api_provider.py
aahung: Yeah, I think so too. Especially in function provider and layer provider, `extract_*()` returns the extraction results. 
In this PR, I don't want to change too much behaviors so I only remove unused code. 

It does not break anything:

https://github.com/aws/aws-sam-cli/search?q=extract_resources

image

samcli/lib/providers/cfn_api_provider.py
aahung: Yes mypy complaint, `dict.get()` returns `Any`, I made it into two lines, one for casting and one for usage so that the lines are more readable.

elbayaaa (approved):

samcli/lib/providers/sam_api_provider.py
aahung: Declaring `api_resource_id` as None is not necessary since it is overridden in the next few lines. Making it None confused mypy making it thinking it is an Optional object

Issue #2492, opened 2020-12-21
Testing Api Gateway endpoint locally returning `No response from invoke container` and `Invalid lambda response received` in Dockerised sam app


alliesground:
### Description:
I'm dockerising sam app. When accessing api endpoint for HelloWorld locally, it responds with `No response from invoke container for HelloWorldFunction` and `Invalid lambda response received: Lambda response must be valid json`.

It works fine whithout dockerisation.

### Steps to reproduce:
**Dockerfile**

```Dockerfile
FROM ruby:2.7.0-alpine

RUN apk add --update --no-cache \
    build-base \
    postgresql-dev \
    postgresql-client \
    python3 \
    py3-pip \
    util-linux \
    python3-dev

RUN pip3 install aws-sam-cli

ARG USER
ARG HOME
ARG UID

RUN apk add --update \
    sudo

RUN echo "Welcome home: $USER => $UID"

RUN adduser -S -D -G users -u $UID $USER
RUN addgroup -S sudo
RUN echo "%sudo ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/sudo 
RUN adduser $USER sudo

RUN echo "Welcome home: $USER"

WORKDIR ${HOME}

EXPOSE 3001

ENTRYPOINT ["sh", "./entrypoint.sh"]
```

**docker-compose.yml**

```yaml
version: '3.8'
services:
  sam_app:
    build:
      context: ./sam-app
      args:
        - HOME
        - USER
        - UID
    user: "${UID}:100"
    command: ["$PWD"]
    ports:
      - "3001:3001"
    volumes:
      - ./sam-app:$HOME
      - /var/run/docker.sock:/var/run/docker.sock
```

**Entrypoint.sh**
```shell
BASEDIR="$1"

echo "Basedir => ${BASEDIR}"

sudo sam local start-api \
  --template ./template.yaml \
  --host 0.0.0.0 \
  --port 3001 \
  --docker-volume-basedir "${BASEDIR}/sam-app/" \
  --docker-network drink_default \
  --debug
```

**hello_world/app.rb**
```ruby
def lambda_handler(event:, context:)
  {
    statusCode: 200,
    body: {
      message: "Hello World!",
      # location: response.body
    }.to_json
  }
end
```

### Observed result:
```
[...]

Issue #2436, opened 2020-12-02
1.13.1: Invalid lambda response received: Lambda response must be valid json


BR0kEN-:
### Description:

`sam local start-api` stopped triggering Lambda by HTTP request. Downgrading to `1.12.0` gives the expected result.

### Steps to reproduce:

```bash
sam build --use-container --debug
sam local start-api --port 4004 --docker-network host --skip-pull-image --debug &
curl -X POST http://127.0.0.1:4004/api/v1/app/add
```

### Observed result:

```
[...]

PR #2700, opened 2021-03-10
feat: add option --container-host to commands local start-api, local start-lambda and local invoke

xazhao:
Add an option `--container-host` to commands `local start-api`, `local start-lambda` and `local invoke`. Customers will be able to specify the host of local emulated containers to support more use cases.

#2492 #2436 

Some customers are having issues to run sam local commands within containers. 

After some investigation, [this line](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/container.py#L41) is the root cause that SAM CLI can not be run within containers. The `localhost` can not be recognized by the system when running in a container. With the new option `--container-host`, customers will be able to specify host of the local emulated container (e.g. host.docker.internal in Docker).
If the new option is not set, `localhost` will be used as default so it won't have impact on current use cases.

#### Checklist

- [x] Write unit tests
I didn't write integration tests for this feature. Basically the new option only make a small change to the URL of the container and unit tests show the correct URL is being called. It's very complicated to write integration test to simulate real use case since the feature will only be used when running in a container. Therefore I believe unit tests are sufficient for this change.
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/docker/container.py
```diff
@@ -266,8 +271,13 @@ def wait_for_http_response(self, name, event, stdout):
         # TODO(sriram-mv): `aws-lambda-rie` is in a mode where the function_name is always "function"
         # NOTE(sriram-mv): There is a connection timeout set on the http call to `aws-lambda-rie`, however there is not
         # a read time out for the response received from the server.
+
+        # if localhost is not set, we assume it's running on the local machine.
+        if not self._localhost:
+            self._localhost = "localhost"
```
hoffa, line 277: The default should be set at the CLI level (where the `click.option` is), so that it shows up with `--help`.

samcli/commands/local/start_api/cli.py
```diff
@@ -47,6 +47,7 @@
     default="public",
     help="Any static assets (e.g. CSS/Javascript/HTML) files located in this directory " "will be presented at /",
 )
+@click.option("--localhost", help="help set localhost running in Docker")
```
hoffa, line 50: I don't think `localhost` is the correct term here. It has a specific meaning as the loopback address, but Docker containers can be started anywhere (as all communication happens through a socket).

hoffa (changes requested):

samcli/commands/local/start_api/cli.py
xazhao: How about `container_host`? And also I'm just thinking this option can also be used in local invoke. So should we put it in the common options?

hoffa: If there's no ambiguity as to what container it applies to, then I think that's reasonable. As for common options, sure if it makes sense to be used by other subcommands.

aahung:
I have a hard time coming up with a reasonable cli option name for this. 

I wonder if this is the approach we should go. What are the possible values customers might specify using `--localhost`? If there are only 1 possibility other than `localhost` (`host.docker.internal`), maybe we can get away with a boolean flag instead of a string cli argument, something like `--listen-to-docker-internal-localhost-hostname` (way too verbose of course). 

Or even further, we can detect whether `host.docker.internal` is resolvable and use it by default (we need to know what's the side effect).

xazhao:
As far as I know, `docker.for.mac.localhost` is another possible value. I think `host.docker.internal` would cover most cases though. It's a trade off between ease of use vs universal. I'm just thinking if there is another value that we don't know currently, we need to make this change again. So letting user specify the container host seems better to me. Let me know your opinion.

hoffa:
All communication with the Docker daemon happens through socket, and technically the listening end (hence the containers started by the daemon) could be anywhere, so the address the containers are listening on could also be anywhere.

vitorfec:
> All communication with the Docker daemon happens through socket, and technically the listening end (hence the containers started by the daemon) could be anywhere, so the address the containers are listening on could also be anywhere.

This is exactly my problem. I have the daemon on a different machine and need to specify that as the host

tests/unit/local/lambdafn/test_runtime.py
```diff
@@ -81,6 +81,7 @@ def test_must_create_lambda_container(self, LambdaContainerMock):
             debug_options=debug_options,
             env_vars=self.env_var_value,
             memory_mb=self.DEFAULT_MEMORY,
+            container_host=None,
```
hoffa, line 84: Isn't `container_host` already `None` by default?

sriram-mv (changes requested):

tests/unit/local/lambdafn/test_runtime.py
xazhao: Yes lambda container is called with `container_host = None` by default. However if we don't set the None here, the call assertion will fail.

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,12 @@ def local_common_options(f):
             default=False,
             help="If set, will emulate a shutdown event after the invoke completes, "
             "in order to test extension handling of shutdown behavior.",
-        )
+        ),
+        click.option(
+            "--container-host",
+            default="localhost",
+            help="Host of locally emulated Lambda container",
```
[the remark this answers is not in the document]
xazhao: Sure! I'm thinking `Host of locally emulated Lambda container. This option can be used for running SAM CLI in containers e.g. If you want to run SAM CLI in a Docker container, use this option with host.docker.internal. For SAM CLI running on your local machine, this option is not needed.` This should be clear for customers who want to use this option. Let me know your thought.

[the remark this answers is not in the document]
hoffa: `host.docker.internal` works in some situations such as for macOS, but doesn't unconditionally resolve to the actual machine running Docker out of the box. See https://github.com/docker/for-linux/issues/264

Even if SAM CLI is running locally (not inside a container), the Docker daemon could be running remotely (see e.g. https://github.com/aws/aws-sam-cli/pull/2700#issuecomment-797077597), requiring this argument.

Mentioning `host.docker.internal` is probably fine, but the main point is that this option is useful when the container ports aren't exposed on SAM CLI's `localhost`, regardless of where SAM CLI may run.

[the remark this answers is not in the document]
xazhao: Yes you're right. The above message could be a little misleading for some use cases. I will try to make the help text more general with a detailed example like `host.docker.internal`.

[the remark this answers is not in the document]
xazhao: `Host of locally emulated Lambda container. This option can be used when the container ports aren't exposed on SAM CLI's localhost. e.g. If you want to run SAM CLI in a Docker container on Mac OS, use this option with host.docker.internal.`
How is this help text?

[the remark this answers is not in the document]
hoffa: I like it! I'd maybe amend a little bit to something like this:

> Host of locally emulated Lambda container. This option is useful when the container ports aren't exposed on the same host SAM CLI runs on. For example, if you want to run SAM CLI in a Docker container on macOS, use this option with `host.docker.internal`.

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,12 @@ def local_common_options(f):
             default=False,
             help="If set, will emulate a shutdown event after the invoke completes, "
             "in order to test extension handling of shutdown behavior.",
-        )
+        ),
+        click.option(
+            "--container-host",
+            default="localhost",
```
hoffa, line 54: Since it seems we don't show defaults... by default, should probably include `show_default=True` here.

samcli/commands/local/cli_common/options.py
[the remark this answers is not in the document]
xazhao: Talked to Chris and we feel this looks better for customers to understand and it's pretty clear too:

> Host of locally emulated Lambda container. This option is useful when the containers run on a different host than SAM CLI. For example, if you want to run SAM CLI in a Docker container on macOS, use this option with `host.docker.internal`

samcli/commands/local/cli_common/options.py
xazhao: +1 I like this idea.

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,16 @@ def local_common_options(f):
...
+            show_default=True,
+            help="Host of locally emulated Lambda container. "
+            "This option is useful when the containers run on a different host than SAM CLI. "
+            "For example, if you want to run SAM CLI in a Docker container on macOS, "
+            "use this option with host.docker.internal",
```
hoffa, line 59: This is me being pedantic, but all help descriptions seem to be full sentences. I wonder if this looks more coherent...

```suggestion
            "use this option with host.docker.internal.",
```

hoffa: Eh never mind, we have some inconsistency elsewhere. I can do a full pass of all help messages later to satisfy my pedantic whims.

tests/unit/local/docker/test_container.py
```diff
@@ -567,17 +615,17 @@ def test_wait_for_result_error_retried(self, mock_requests):
             calls,
             [
                 call(
-                    "http://localhost:7077/2015-03-31/functions/function/invocations",
+                    "http://None:7077/2015-03-31/functions/function/invocations",
```
hoffa, line 618: What is this testing?

xazhao: It's to test whether the URL can be replaced by the new option. None is the default value of container_host in container class. So if we don't assign the value here, the URL should be replaced to this.

hoffa: I see. I'd probably pass `localhost` explicitly then as a principle of least surprise (seeing `None` in an URL is a little weird).

xazhao: Setting `localhost` here is fine. The only thing I worry about is that it will override the previous unit test (the URL will only be replaced when setting `container_host`). Is it okay?
If it's okay, I will remove the new unit test(using `host.docker.internal` as container_host) as well since it basically does the same thing.

hoffa: If they're both testing the same thing, then sure there's no point in having two tests.

samcli/local/docker/container.py
```diff
@@ -71,6 +72,7 @@ def __init__(
         :param docker_client: Optional, a docker client to replace the default one loaded from env
         :param container_opts: Optional, a dictionary containing the container options
         :param additional_volumes: Optional list of additional volumes
+        :param string container_host: Optional. Host of locally emulated Lambda container
```
aahung, line 75: this cannot be optional anymore, because if customers do not specify a value, click will use the default value `localhost`.

samcli/local/docker/container.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         docker_client=None,
         container_opts=None,
         additional_volumes=None,
+        container_host=None,
```
aahung, line 58: same here, this is not optional

aahung (commented):
It seems `container_host` can never be `None` because it has a default value in click. Therefore, the `Optional` for this might be misleading.

samcli/local/docker/container.py
xazhao: `_init_` here is also used in another its child class `LambdaBuildContainer`. The new option is not used in that class so I make it as optional.

samcli/local/docker/container.py
xazhao: Same reason as my another comment below.

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,16 @@ def local_common_options(f):
...
             help="If set, will emulate a shutdown event after the invoke completes, "
             "in order to test extension handling of shutdown behavior.",
-        )
+        ),
+        click.option(
+            "--container-host",
+            default="localhost",
+            show_default=True,
+            help="Host of locally emulated Lambda container. "
+            "This option is useful when the containers run on a different host than SAM CLI. "
```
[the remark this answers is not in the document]
xazhao: Oh it's a typo. Will change it.

samcli/local/docker/container.py
sriram-mv: if this is used by other classes (i'm assuming those other classes are instantiated outside of sam local suite, is it sam build?) , should we set a default value? this can get overridden when the actual option is used.

sriram-mv (changes requested):

samcli/local/docker/container.py
xazhao: As far as I know, `container_host` is only used in SAM LOCAL suite when one of its child class is instantiated. So whether or not setting the default value here makes no difference. If we could potentially expand the use of new option outside of sam local suite, then yes we should set default to `localhost` here. Considering there is no negative impact setting the default value, I will add the default value to the option here.

samcli/commands/local/cli_common/options.py
```diff
@@ -54,7 +54,7 @@ def local_common_options(f):
             default="localhost",
             show_default=True,
             help="Host of locally emulated Lambda container. "
-            "This option is useful when the containers run on a different host than SAM CLI. "
+            "This option is useful when the container run on a different host than SAM CLI. "
```
aahung, line 57:
```suggestion
            "This option is useful when the container runs on a different host than SAM CLI. "
```

aahung (approved):
Although we store `localhost` (the default value) in two different places, it is acceptable to me due to it is very unlikely to change. Thanks for addressing the comments.

sriram-mv (approved):

tests/unit/commands/local/lib/test_local_lambda.py
```diff
@@ -625,6 +625,45 @@ def test_must_raise_if_imageuri_not_found(self):
...
+class TestLocalLambda_invoke_with_container_host_option(TestCase):
+    def setUp(self):
+        self.runtime_mock = Mock()
+        self.function_provider_mock = Mock()
+        self.cwd = "/my/current/working/directory"
+        self.debug_context = None
+        self.aws_profile = "myprofile"
+        self.aws_region = "region"
+        self.env_vars_values = {}
+        self.container_host = "localhost"
```
hoffa, line 637: This should be testing a different value from the default.

xazhao: The default value of `container_host` is `None` in this class. Since it's tested in isolation, the default value of the option won't be passed into here.

hoffa (approved):
LGTM! 👍

samcli/commands/local/lib/local_lambda.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         aws_region: Optional[str] = None,
         env_vars_values: Optional[Dict[Any, Any]] = None,
         debug_context: Optional[DebugContext] = None,
+        container_host: Optional[str] = None,
```
mndeveci, line 46: Since there is a default value for the parameter, would this param be `None` anytime?

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,16 @@ def local_common_options(f):
             default=False,
             help="If set, will emulate a shutdown event after the invoke completes, "
             "in order to test extension handling of shutdown behavior.",
-        )
+        ),
+        click.option(
+            "--container-host",
```
mndeveci, line 53: For any security reasons, do we need to validate if this option valid?

mndeveci (commented):
Great feature, thanks for the implementation 🎉

Can we also add some integration test cases to validate this new parameter as well?

samcli/commands/local/lib/local_lambda.py
xazhao: This param won't be `None` in the current use case. The reasons I set it to `None`:
- I'd like to have minimum impact on the current use case. If this class is initialized somewhere else, we shouldn't assume it's `localhost` unless it's passed from cli.
- Match to the overall style.

samcli/commands/local/cli_common/options.py
xazhao: This is a good point. Let me check on this and get back to you.

xazhao:
For the integration test, please see my explanation in the summary.

> I didn't write integration tests for this feature. Basically the new option only make a small change to the URL of the container and unit tests show the correct URL is being called. It's very complicated to write integration test to simulate real use case since the feature will only be used when running in a container. Therefore I believe unit tests are sufficient for this change.

samcli/commands/local/cli_common/options.py
xazhao: For security reasons, there is nothing we can validate. There is a use case that docker daemon is running on a different host and the customer wants to use that host. We won't be able to understand if it's a good host provided by customer or just a bad actor.

mndeveci (approved):

hanneswidrig:
Why did this get reverted in master?

PR #2708, opened 2021-03-12
feat(Public ECR): Changed Build Image Registry to Public ECR

CoshUS:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_build_container.py
+++ b/samcli/local/docker/lambda_build_container.py
@@ -18,9 +18,10 @@ class LambdaBuildContainer(Container):
     and if the build was successful, copies back artifacts to the host filesystem
     """
 
-    _IMAGE_OWNER = "amazon"
-    _IMAGE_NAME_PREFIX = "aws-sam-cli-build-image"
-    _IMAGE_NAME_FORMAT = "{}/{}-{}"  # amazon/aws-sam-cli-build-image-nodejs10.x
+    _IMAGE_OWNER = "sam"
+    _IMAGE_NAME_PREFIX = "build"
+    _IMAGE_REGISTRY = "public.ecr.aws"
+    _IMAGE_NAME_FORMAT = "{}/{}/{}-{}"  # public.ecr.aws/sam/build-java11
     _BUILDERS_EXECUTABLE = "lambda-builders"
 
     def __init__(  # pylint: disable=too-many-locals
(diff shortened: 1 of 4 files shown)


hoffa:
Were some sanity tests done?

samcli/local/docker/lambda_build_container.py
```diff
@@ -237,5 +238,8 @@ def _convert_to_container_dirs(host_paths_to_convert, host_to_container_path_map
     @staticmethod
     def _get_image(runtime):
         return LambdaBuildContainer._IMAGE_NAME_FORMAT.format(
```
jfuss, line 240: This would be easier to understand/read if we combined what we could into one string instead of splitting them. Especially since `_IMAGE_OWNER`, `_IMAGE_NAME_PREFIX`, and `_IMAGE_REGISTRY` are only even used to construct the image uri. 

To make this more readable, I would suggest:
```
__IMAGE_URI_PREFIX = "public.ecr.aws/sam/build-"
return f"{_IMAGE_URI_PREFIX}-{runtime}"
```

samcli/local/docker/manager.py
```diff
@@ -126,7 +126,7 @@ def stop(self, container: Container) -> None:
             container.stop()
         container.delete()
 
-    def pull_image(self, image_name, stream=None):
+    def pull_image(self, image_name, tag="latest", stream=None):
```
jfuss, line 129: Should we add any validation that we only pull latest in some manner?

jfuss (changes requested):
We might need to update our build integ tests: https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/buildcmd/build_integ_base.py#L118 which seems like it is very much out of date :(

samcli/local/docker/lambda_build_container.py
CoshUS: Updated.

samcli/local/docker/manager.py
CoshUS: Tests will be added in another PR. This PR is against the feat/public-ecr feature branch. Trying out using feature branches to see if there are issues.

CoshUS:
> Were some sanity tests done?

Simple manual runs are done but tests will be added in another PR. This PR is against the feat/public-ecr instead of the develop branch.

jfuss (approved):
LGTM

PR #2709, opened 2021-03-12
cookiecutter template workflow

elbayaaa:
This PR defines a generic workflow for cookiecutter project generation. The work flow contains the following steps:
1. Ask the user a series of questions to fulfill the values of the keys of the cookiecutter.json (cookiecutter context)
2. Run a series of pre-processors to manipulate the context and do any required bootstrapping before generating the cookiecutter project.
3. generate the cookiecutter project
4. Run a series of post-processors to do any required cleanup, printing messages or whatever actions required after successfully creating the project

First step in supporting a new SAM feature.

No side effect, All of this code is new and is not referenced anywhere yet.
#### Checklist

- [√] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [√] Write unit tests
- [√] `make pr` passes
- [ ] Write documentation


samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,126 @@
...
+        The key of the cookiecutter config to associate the answer with.
+    _text: str
+        The text to prompt to the user
+    _options: Optional[List[str]]
+        if this is a mcq question, this attribute specifies the possible choices
+    _required: bool
+        Whether the user must provide an answer for this question or not.
+    _default_answer: Optional[str]
+        A default answer that is suggested to the user
+    _next_question_map: Optional[Dict[str, str]]
```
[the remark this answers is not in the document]
aahung: and also what `*` means (default)

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+        answer: Any = None
+        while question:
+            if question.is_info():
+                click.echo(question.text)
+            elif question.is_mcq():
+                answer = InteractiveFlow._ask_a_multiple_choice_question(question)
+            elif question.is_yes_no():
+                answer = InteractiveFlow._ask_a_yes_no_question(question)
+            else:
+                answer = InteractiveFlow._ask_a_question(question)
```
aahung, line 89: `_ask_a_question` feels like a generic method covering the previous 2, maybe `_ask_a_text_question`?

aahung (commented):
gonna do another round later

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+            text="Choice",
+            default=question.default_answer,
+            show_choices=False,
+            type=click.Choice(question.get_choices_indexes(base=1)),
+        )
+        return question.get_option(int(choice) - 1)
+
+    def get_next_question(self, current_answer: Optional[str] = None) -> Optional[Question]:
+        """
+        Based on the answer of the current question, what is the next question to ask.
```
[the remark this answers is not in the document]
elbayaaa: agree, will rename to advance_to_next_question

samcli/lib/cookiecutter/question.py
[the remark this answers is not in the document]
elbayaaa: * rethinking about the string casting, I think I will change the map to be `_next_question_map: Optional[Dict[Any, str]]` instead of casting
* the `*` means match any answer, I think this is a common behavior, right?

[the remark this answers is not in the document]
aahung: > rethinking about the string casting, I think I will change the map to be _next_question_map: Optional[Dict[Any, str]] instead of casting

samcli/lib/cookiecutter/interactive_flow.py
elbayaaa: It is common to try the custom handlers first then if nothing match, go for the generic one. So, yes `_ask_a_question` is a generic handler that can handle any question, that is why it is in an `else` clause not a `elif`.

samcli/lib/cookiecutter/question.py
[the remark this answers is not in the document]
aahung: > rethinking about the string casting, I think I will change the map to be _next_question_map: Optional[Dict[Any, str]] instead of casting

how it affects the load question from json? json key should be string right?

> the * means match any answer, I think this is a common behavior, right?

not until wildcard came to my mind... `*` could also be a possible user input. I would rather have a dedicate attribute `_default_next_question`

samcli/lib/cookiecutter/interactive_flow.py
aahung: I see, 👍

samcli/lib/cookiecutter/question.py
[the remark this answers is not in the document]
elbayaaa: * `Any` is the type of the answer not the key, _next_question_map = {answerX: qX, answerY: qY: *(every other answer): qZ}
* make sense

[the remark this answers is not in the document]
elbayaaa: discard my previous comment, answer has to be a string as well as it is also provided by the use in the questions.yaml file. I will add a comment why we do the casting.

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
+"""A flow of questions to be asked to the user in an interactive way."""
+from typing import Any, Dict, Optional
+import click
+from .question import Question
+
+
+class InteractiveFlow:
```
sriram-mv, line 7: Question: **_Can we make the interactive part of sam init a library that can then be injected_**? does this have any external dependencies so is everything self contained within samcli/lib/cookiecutter

This way it be can like using `aws-lambda-builders` from `aws-sam-cli` through sam build. We can have clearly defined interfaces elsewhere and we only need to instantiate the appropriate interactive workflow.

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+        The entry-point of the flow, it tells which question to ask the user first
+    """
+
+    def __init__(self, questions: Dict[str, Question], first_question_key: str):
+        self._questions: Dict[str, Question] = questions
+        self._first_question_key: str = first_question_key
+        self._current_question: Optional[Question] = None
+
+    @staticmethod
+    def _ask_a_question(question: Question) -> Any:
```
sriram-mv, line 25: suggestion: `prompt` instead of `_ask_a_question`?

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+        return click.prompt(text=question.text, default=question.default_answer)
+
+    @staticmethod
+    def _ask_a_yes_no_question(question: Question) -> bool:
```
sriram-mv, line 29: suggestion: `confirmation` instead of `_ask_a_yes_no_question`?

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+        return click.confirm(text=question.text, default=question.default_answer)
+
+    @staticmethod
+    def _ask_a_multiple_choice_question(question: Question) -> Any:
```
sriram-mv, line 33: `prompt_mulitple_choice` maybe?

samcli/lib/cookiecutter/interactive_flow.py
sriram-mv: I like the idea of separating the flow of questions into its own class, instead of relying on multiple if, else statements. Nice!

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+            The cookiecutter context before prompting this flow's questions
+
+        Returns: A new copy of the context with user's answers added to the copy such that each answer is
+                 associated to the key of the corresponding question
+        """
+        context = context.copy()
+        question = self.advance_to_next_question()
+        answer: Any = None
+        while question:
+            if question.is_info():
```
sriram-mv, line 82: We definitely need a way to do actions based on the questions in play, instead of multiple if, else. Have you thought about a dictionary dispatch pattern?

https://www.oreilly.com/library/view/python-cookbook/0596001673/ch01s07.html

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -0,0 +1,92 @@
...
+                is the start of the flow.
+        """
+        if not self._current_question:
+            self._current_question = self._questions.get(self._first_question_key)
+        else:
+            next_question_key = self._current_question.get_next_question_key(current_answer)
+            self._current_question = self._questions.get(next_question_key) if next_question_key else None
+        return self._current_question
+
+    def run(self, context: Dict) -> Dict:
```
sriram-mv, line 65: is there any pre-processing we need to do before? Such as

* if a user has passed in prompts via cli flags, do those questions need to be skipped?

samcli/lib/cookiecutter/template.py
```diff
@@ -0,0 +1,158 @@
...
+            for flow in self._interactive_flows:
+                context = flow.run(context)
+            return context
+        except Exception as e:
+            raise UserException(str(e), wrapped_from=e.__class__.__name__) from e
+
+    def generate_project(self, context: Dict):
+        """
+        Generates a project based on this cookiecutter template and the given context. The context is first
+        processed and manipulated by series of preprocessors(if any) then the project is generated and finally
```
sriram-mv, line 123: Q: Where are the pre-processors and post-processors defined, and what kind of actions are done? are they controlled by us or by users?

samcli/lib/cookiecutter/template.py
```diff
@@ -0,0 +1,158 @@
...
+            the cookiecutter context to fulfill the values of cookiecutter.json keys
+
+        Raise:
+        ------
+        InvalidLocationError: if the given location is not from a known repo type
+        GenerateProjectFailedError: if something else went wrong
+        """
+        try:
+            LOG.debug("preprocessing the cookiecutter context")
+            for processor in self._preprocessors:
```
sriram-mv, line 138: are there any specific processor based exceptions we can catch?

sriram-mv (changes requested):
Did a 1st round review, have not looked at tests yet.

samcli/lib/cookiecutter/interactive_flow.py
elbayaaa: The way I looked to it was that this is an human interactive flow,  I wanted it to be an English-like conversation. We are literally asking a human being a question. Make sense?

samcli/lib/cookiecutter/template.py
elbayaaa: This is the workflow. Any template/plugin uses this workflow will define its own preprocessing and postprocessing logic, if needed. The next PR will have an example plugin using it.

samcli/lib/cookiecutter/template.py
elbayaaa: The workflow doesn't define any processing logic on its own, it gives the template-creator the ability to define a processing step before/after generating the cookiecutter project, i.e., the workflow itself doesn't know about specific processor-based exceptions.
We can of course define and raise a particular exception in case if any exception occurred while doing the preprocessing/postprocessing but I'm quite unsure of its benefits at the moment.

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+        click on how to prompt the user for the answer; click.prompt, click.confirm, click.choice ...etc
+    """
+
+    def __init__(
+        self,
+        key: str,
+        text: str,
+        options: Optional[List[str]] = None,
+        default: Optional[str] = None,
+        is_required: Optional[bool] = False,
```
hoffa, line 57: nit: doesn't seem to be `Optional`

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+    @property
+    def default_next_question_key(self):
+        return self._default_next_question_key
+
+    @property
+    def kind(self):
+        return self._kind
+
+    def get_choices_indexes(self, base: int = 0) -> List[str]:
+        if self._options:
```
hoffa, line 107: Does "choices" refer to "options"? Consistent terminology will reduce ambiguity.

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+            choices = list(range(base, len(self._options) + base))
+            return list(map(str, choices))
+        return list()
+
+    def get_option(self, index) -> Any:
```
hoffa, line 112: Should this be `-> str`?

samcli/lib/cookiecutter/interactive_flow.py
sriram-mv: Sure, but the contents of the question are present within the question right? as in prompt(question). all of the contents are questions, that is why I asked regarding that.

samcli/lib/cookiecutter/interactive_flow.py
sriram-mv: Yeah. this will eventually be wired up to a command right, `sam xyz` which will be an interactive experience, I want that part to be pluggable and driven by a library, and then we can possibly port over the same logic back to sam init.

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+        return self._kind == QuestionKind.confirm
+
+    def is_info(self) -> bool:
+        return self._kind == QuestionKind.info
+
+    def get_next_question_key(self, answer: Any) -> Optional[str]:
+        # _next_question_map is a Dict[str(answer), str(next question key)]
+        # so we need to parse the passed argument(answer) to str
+        answer = str(answer)
+        return self._next_question_map.get(answer, self._default_next_question_key)
```
hoffa, line 128: Is there any non-confirm-type question where the next question is based on the previous answer?

In other words, if we know what the possible branches are, then a MCQ already covers it.

samcli/lib/cookiecutter/interactive_flow.py
elbayaaa: yeah, that is the intention of this effort.

samcli/lib/cookiecutter/question.py
elbayaaa: tried a lot with no success. mypy keep complaining that `self._options[index]` is of type `Any` and I have no idea why.

samcli/lib/cookiecutter/question.py
elbayaaa: not sure whether I understood your point or not. A Yes/No questions and a MCQ questions are the obvious type of questions where branching is valid.

hoffa: Okay I see, so this is used for MCQs as well; makes sense!

samcli/lib/cookiecutter/interactive_flow.py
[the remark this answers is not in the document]
aahung: thanks!

samcli/lib/cookiecutter/interactive_flow.py
aahung: I think @elbayaaa tries to make these 3 methods consistent in naming (all ending with a `question`)

samcli/lib/cookiecutter/plugin.py
```diff
@@ -0,0 +1,40 @@
...
+    def interactive_flow(self):
+        return self._interactive_flow
+
+    @property
+    def preprocessor(self):
+        return self._preprocessor
+
+    @property
+    def postprocessor(self):
+        return self._postprocessor
```
aahung, line 40: This class does not seem to do anything, I would suggest to make it simpler like this:

```py
class Plugin(NamedTuple):
  interactive_flow: InteractiveFlow
  preprocessor: Processor
  postprocessor: Processor
```

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+        self._text = text
+        self._options = options or []
+        self._required = is_required
+        self._default_answer = default
+        # if it is an optional question, set an empty default answer to prevent click from keep asking for an answer
+        if not self._required and self._default_answer is None:
+            self._default_answer = ""
+        self._next_question_map = next_question_map or {}
+        self._default_next_question_key = default_next_question_key
+        self._kind = kind or (QuestionKind.confirm if options else QuestionKind.default)
```
aahung, line 72: Why is it `confirm` by default when `options` is not empty?

samcli/lib/cookiecutter/processor.py
```diff
@@ -0,0 +1,22 @@
...
+        the processing logic of this processor
+
+        Parameters
+        ----------
+        context: Dict
+            The cookiecutter context to process
+
+        Returns: A processed copy of the cookiecutter context
+        """
+        raise NotImplementedError("not implemented")
```
aahung, line 22: Adding a `@abstractmethod` instead of raise an exception here? https://docs.python.org/3/library/abc.html and https://softwareengineering.stackexchange.com/questions/326207/using-notimplementederror-instead-of-abstract-classes

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,80 @@
...
+    @staticmethod
+    def _create_question_from_json(question_json: Dict) -> Question:
+        key = question_json["key"]
+        text = question_json["question"]
+        options = question_json.get("options")
+        default = question_json.get("default")
+        is_required = question_json.get("isRequired")
+        next_question_map = question_json.get("nextQuestion")
+        kind_str = question_json.get("kind")
+        kind = QuestionKind[kind_str] if kind_str else None
```
aahung, line 71: nit: I think here you want to do `QuestionKind(kind_str)` instead of `[kind_str]`. The former is finding the enum by value, the latter is by name. I feel by value makes more sense. Since your enum's name and value are the same, it does not matter too much here.

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        questions, first_question_key = InteractiveFlowCreator._load_questions(flow_definition_path)
+        return InteractiveFlow(questions=questions, first_question_key=first_question_key)
+
+    @staticmethod
+    def _load_questions(flow_definition_path: str) -> Tuple[Dict[str, Question], str]:
+        previous_question: Optional[Question] = None
+        first_question_key: str = ""
+        questions: Dict[str, Question] = {}
+        template_data = get_template_data(flow_definition_path)
```
aahung, line 50: Although `get_template_data()` works here, it might be worth just to use its implementation `yaml_parse(fp.read())` here. Otherwise customers might see error message like this `Template file not found at ...` when anything goes wrong, which is misleading.

samcli/lib/cookiecutter/question.py
aahung: Change this to this in your `__init__()` should resolve it

```
self._options = options or []
->
self._options: List[str] = options or []
```

aahung: `options or []` returns `Union[List[str], List[Any]]`, maybe that's why you keep getting that error.

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,134 @@
...
+    ):
+        self._key = key
+        self._text = text
+        self._options = options or []
+        self._required = is_required
+        self._default_answer = default
+        # if it is an optional question, set an empty default answer to prevent click from keep asking for an answer
+        if not self._required and self._default_answer is None:
+            self._default_answer = ""
+        self._next_question_map = next_question_map or {}
```
aahung, line 70: See https://github.com/aws/aws-sam-cli/pull/2709/files#r594590917

```suggestion
        self._next_question_map: Dict[str, str] = next_question_map or {}
```

samcli/lib/cookiecutter/exceptions.py
```diff
@@ -0,0 +1,18 @@
+""" Exceptions raised from the cookiecutter workflow"""
+
+
+class CookiecutterErrorException(Exception):
```
aahung, line 4: I wonder should we make this class private (by prefixing it with `_`) so people won't mistakenly use it.

samcli/lib/cookiecutter/question.py
elbayaaa: good catch, I meant `choice`

tests/unit/lib/cookiecutter/test_question.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_creating_questions(self):
+        q = Question(text=self._ANY_TEXT, key=self._ANY_KEY)
+        assert q.text == self._ANY_TEXT
+        assert q.key == self._ANY_KEY
+        assert q.options == []
+        assert q.default_answer == ""
+        assert q.required is False
+        assert q.next_question_map == {}
+        assert q.default_next_question_key is None
+        assert q.kind == QuestionKind.default
```
aahung, line 40: We should use `self.assertEqual()` and other `self.assert*()` instead of `assert`. The former will output more helpful information in case of a failure. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

samcli/lib/cookiecutter/question.py
elbayaaa: > Change this to this in your `__init__()` should resolve it
> 
> ```
> self._options = options or []
> ->
> self._options: List[str] = options or []
> ```

I tried this yesterday, didn't work, let me give it a 2nd try.

samcli/lib/cookiecutter/exceptions.py
elbayaaa: but people may need to use it to capture `GenerateProjectFailedError` and `InvalidLocationError` and take an action.

samcli/lib/cookiecutter/interactive_flow_creator.py
elbayaaa: how is this possible? the method raise "Template file not found" only if `if not pathlib.Path(template_file).exists():` it raises "Failed to parse template" if anything else went wrong. Anyway, if there is a misleading behavior in the method, we can fix it, instead of defining a new method

aahung: if you take a look at the implementation of `get_template_data()`, it is actually just open the file and pass to `yaml_parse()`. The reason I am calling this out is because `get_template_data()`'s name does not match what it is doing here.

We don't need to define a new method, we just need to do `yaml_parse()`

samcli/lib/cookiecutter/interactive_flow.py
```diff
@@ -76,17 +82,16 @@ def run(self, context: Dict) -> Dict:
                  associated to the key of the corresponding question
         """
         context = context.copy()
+        question_handlers = {
+            QuestionKind.info: InteractiveFlow._echo,
+            QuestionKind.choice: InteractiveFlow._ask_a_multiple_choice_question,
+            QuestionKind.confirm: InteractiveFlow._ask_a_yes_no_question,
+            QuestionKind.default: InteractiveFlow._ask_a_question,
+        }
```
aahung, line 90: nit:

`question_handlers` can be put in out scope. And can we add one more layer of safety in case in the future we have more kinds of questions with this unit test?

```py
for kind in QuestionKind:
    self.assertIn(kind, InteractiveFlow.question_handlers)
```

elbayaaa: will add the unit test. Not sure about the scope of `question_handlers` though, it is only required here

aahung: we can also leave it in there, then we just need to unit test the whole method.

samcli/lib/cookiecutter/exceptions.py
aahung: Oh I see. There seems to have no other way

tests/unit/lib/cookiecutter/questions.yaml
```diff
@@ -0,0 +1,22 @@
...
+      }
+    },
+    {
+      "key": "3rd",
+      "question": "any text",
+      "isRequired": true,
+      "kind": "confirm"
+    }
+  ]
+}
```
aahung, line 22: this test file is a JSON file, why it has ".yaml" as extension?

tests/unit/lib/cookiecutter/test_interactive_flow_creator.py
```diff
@@ -0,0 +1,70 @@
...
+            InteractiveFlowCreator.create_flow(flow_definition_path=questions_path)
+
+    def assert_equal(self, q1: Question, q2: Question):
+        self.assertEqual(q1.key, q2.key)
+        self.assertEqual(q1.text, q2.text)
+        self.assertEqual(q1.options, q2.options)
+        self.assertEqual(q1.default_answer, q2.default_answer)
+        self.assertEqual(q1.required, q2.required)
+        self.assertEqual(q1.next_question_map, q2.next_question_map)
+        self.assertEqual(q1.kind, q2.kind)
```
aahung, line 70: this can be simplified using `__dict__`. See an example in our codebase:

https://github.com/aws/aws-sam-cli/blob/01207ede573fe91ee53bfab9f48e6f23276d35ce/samcli/lib/providers/provider.py#L109-L111

samcli/lib/cookiecutter/question.py
aahung: I think now it can be a `-> str`

aahung (commented):
Overall it looks good to me. thx for address the comment. Is there a way for me to quickly try an interactive flow myself?

tests/unit/lib/cookiecutter/questions.yaml
elbayaaa: good catch. Fixed

tests/unit/lib/cookiecutter/test_interactive_flow_creator.py
elbayaaa: Nice. Fixed

elbayaaa:
> Overall it looks good to me. thx for address the comment. Is there a way for me to quickly try an interactive flow myself?

This should be available in the next PRs

samcli/lib/cookiecutter/question.py
aahung:
```suggestion
    def get_option(self, index) -> str:
```

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,123 @@
...
+    def key(self):
+        return self._key
+
+    @property
+    def text(self):
+        return self._text
+
+    @property
+    def options(self):
+        return self._options
```
aahung, line 84: for these properties, are they all used? I cannot find usage of this "options()"

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,123 @@
...
+                         _next_question_map={"True": "r2", "False": "r1-creation"}
+                Question(key="r1-creation" text="What do you like to name resource1?"}
+                Question(key="r2" text="do you already have resource2?",
+                         _next_question_map={"True": "r3", "False": "r2-creation"}
+    _default_next_question_key: str
+        The key of the next question that is not based on user's answer, i.e., if there is no matching in
+        the _next_question_map
+    _kind: QuestionKind
+        The kind of the question, it can be one of the values of the QuestionKind enum. This basically directs
+        click on how to prompt the user for the answer; click.prompt, click.confirm, click.choice ...etc
```
aahung, line 48: minor: some of the types in docstring does not match the ones in `__init__`. In my opinion, since we already have type hints, we can remove the types in docstring to make maintenance easier.

aahung (approved):
Other than small comments, the whole PR LGTM. thx for addressing the comments.

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,125 @@
...
+    pass
+
+
+class QuestionsFailedParsingException(UserException):
+    pass
+
+
+class InteractiveFlowCreator:
+    @staticmethod
+    def create_flow(flow_definition_path: str, extra_context: Optional[Dict] = None):
```
sriram-mv, line 20: One thing to consider for the future. this flow definition path is static, however there are parts of `sam init` specifically those deal with event bridge, where the questions are generated dynamically based on the previous option selected.

Can we chain together flows in the future?

tests/unit/lib/cookiecutter/questions.yaml
hoffa: Technically JSON is valid YAML :^)

samcli/lib/cookiecutter/template.py
sriram-mv: GenerateProjectFailedError is the generic exception message, a more specific one will help us deep dive on issues.

sriram-mv (commented):
This mostly looks good, Going to do a final pass on tests.

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,125 @@
...
+    @staticmethod
+    def _load_questions(
+        flow_definition_path: str, extra_context: Optional[Dict] = None
+    ) -> Tuple[Dict[str, Question], str]:
+        previous_question: Optional[Question] = None
+        first_question_key: str = ""
+        questions: Dict[str, Question] = {}
+        questions_definition = InteractiveFlowCreator._parse_questions_definition(flow_definition_path, extra_context)
+
+        for question in questions_definition.get("Questions"):
```
hoffa, line 66: UX: `Questions` isn't consistent with other capitalization. Better to keep it as `questions` so that we don't have to remember which key is the exception.

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,125 @@
...
+        try:
+            return parse_yaml_file(file_path=file_path, extra_context=extra_context)
+        except FileNotFoundError as ex:
+            raise QuestionsNotFoundException(f"questions definition file not found at {file_path}") from ex
+        except (KeyError, ValueError, yaml.YAMLError) as ex:
+            raise QuestionsFailedParsingException(f"Failed to parse questions: {str(ex)}") from ex
+
+    @staticmethod
+    def _create_question_from_json(question_json: Dict) -> Question:
+        key = question_json["key"]
```
hoffa, line 107: Is question mandatory even if it's never used?

samcli/lib/cookiecutter/question.py
elbayaaa: I kept some of them, like options(), for future usage, if any.

samcli/lib/cookiecutter/question.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+
+    def __init__(
+        self,
+        key: str,
+        text: str,
+        options: Optional[List[str]] = None,
+        default: Optional[str] = None,
+        is_required: Optional[bool] = None,
```
hoffa, line 57: We can make this non-optional:

```suggestion
        is_required: bool = False,
```

hoffa (approved):
Added small suggestion, otherwise LGTM! 👍

samcli/lib/cookiecutter/template.py
```diff
@@ -0,0 +1,164 @@
...
+            templates, it is better to be extracted to a plugin that can then be plugged in to each of these templates.
+        """
+        self._location = location
+        self._interactive_flows = interactive_flows or []
+        self._preprocessors = preprocessors or []
+        self._postprocessors = postprocessors or []
+        self._plugins = plugins or []
+        for plugin in self._plugins:
+            if plugin.interactive_flow:
+                self._interactive_flows.append(plugin.interactive_flow)
```
sriram-mv, line 93: does this control order of execution? should plugin interactive flow be the first thing? similairly for pre and post processors? or does it not matter?

tests/unit/lib/cookiecutter/test_template.py
```diff
@@ -0,0 +1,155 @@
...
+    _PLUGIN_MOCK = Mock()
+    _ANY_INTERACTIVE_FLOW_CONTEXT = Mock()
+    _ANY_PLUGIN_INTERACTIVE_FLOW_CONTEXT = Mock()
+    _ANY_PROCESSOR_CONTEXT = Mock()
+
+    @patch("samcli.lib.cookiecutter.interactive_flow")
+    @patch("samcli.lib.cookiecutter.processor")
+    @patch("samcli.lib.cookiecutter.processor")
+    @patch("samcli.lib.cookiecutter.plugin")
+    def test_creating_a_template(self, mock_plugin, mock_preprocessor, mock_postprocessor, mock_interactive_flow):
```
sriram-mv, line 28: Could we add one more test where the supplied plugin already has pre-processors of its own?

samcli/lib/cookiecutter/template.py
elbayaaa: A standard flow is as following:
1. InteractiveFlow: interactive questionnaire that gather the user's answers and create the cookiecutter context 
2. Preprocessing: a template-defined logic to process the context, for example the preprocessing of the pipeline template(coming next) is to create the required AWS resources if the user didn't provide it
3. Cookiecutter: generating the cookiecutter project form the template's files
4. postprocessing: a template-defined logic. for example the pipeline templates will use the post-processing to print instructions on what resources it created on behalf of the users.

So a template defines an interactive-flow, pre-processors and post-processors.

Now, multiple templates may have a similar flow and processing logic, and this is where the **plugin** concept come into place. We can extract the common interactive-flow, preprocessor and postprocessor to a plugin, then plug it into each of these template which will just append these three components to the corresponding template's components.

Regarding the order; it will be the template's own component then the plugins' components in the same order they are plugged in.

Example:
t = Template(interactive_flows=[f1, f2], preprocessors=[p1, p2,], postprocessors=[p3, p4], plugins=[g1,g2])
the order of execution of each component will be as following:
1. interactive_flows: f1, f2, g1.interactive_flow, g2.interactive_flow
2. preprocesspors: p1, p2, g1.preprocesspor, g2.preprocesspor
3. postprocesspors: p3, p4, g1.postprocesspor, g2.postprocesspor

tests/unit/lib/cookiecutter/test_template.py
elbayaaa: I already made this test for the interactive-flow(Template with direct interactive flow and a plugin's interactive flow) preprocessing and postprocessing are just the same concept, but no worries let me add them

elbayaaa: Oh, I actually have them [here](https://github.com/aws/aws-sam-cli/pull/2709/files#diff-3e916e76ac8ee3165271ce7efc4cf30d72710051b5987971c987e92de2f1c551R119-R123)

sriram-mv: Yep, understand the concept being the same, just want to ensure that both processors on the plugin and template are run.

samcli/lib/cookiecutter/template.py
sriram-mv: Thanks for the detailed explanation!

sriram-mv (approved):

samcli/lib/cookiecutter/interactive_flow.py
moelasmar: We have different type of questions, and each question has a different way to be asked. I feel like we should move the ask method to be part of the question itself, and we can have different types of Question class, and the asking logic can be included in each type .. so we can be open for adding new question type without needs to add new method here

samcli/lib/cookiecutter/interactive_flow_creator.py
```diff
@@ -0,0 +1,125 @@
...
+        first_question_key: str = ""
+        questions: Dict[str, Question] = {}
+        questions_definition = InteractiveFlowCreator._parse_questions_definition(flow_definition_path, extra_context)
+
+        for question in questions_definition.get("questions"):
+            q = InteractiveFlowCreator._create_question_from_json(question)
+            if not first_question_key:
+                first_question_key = q.key
+            elif previous_question and not previous_question.default_next_question_key:
+                previous_question.set_default_next_question_key(q.key)
```
moelasmar, line 71: I think it is better that the default next question should be defined in the questions file itself, as by this way, how can we differentiate if some question is the last question of the follow or not.

example:
```
[
     {
         "key": "q1",
         "nextQuestion": {
                  "True": "q2",
                  "False": "q3",
         }
     },
     {
         "key": "q2"
     },
     {
         "key": "q3"
     }
]
```
my assumption in the above questions is q2 or q3 are the end of the interactive flow, but using the code above I think you will add q3 as the default new question of q2

samcli/lib/cookiecutter/interactive_flow.py
elbayaaa: I thought about this too, but wanted to make the Question class unaware of "click". I might revise this later if I found this pattern is repeated again.

samcli/lib/cookiecutter/interactive_flow_creator.py
elbayaaa: by default a next question for a question is the one defined after it in the questions' array. We don't need to define this`nextQuestion` map unless we have the need to do branching. I understand that this will need a last question that all branches terminates to but this is usually the case in the interactive-flows, you end the questionnaire with a message. For example, in the case you provided it will be something like this:

```
[
     {
         "key": "q1",
         "nextQuestion": {
                  "True": "q2",
                  "False": "q3",
         }
     },
     {
         "key": "q2"
         "defaultNextQuestion": "q4"
     },
     {
         "key": "q3"
     },
     {
         "key": "q4"
         "kind": "info"
         "question": "End of Text 👯👯 "
     }
]
```

moelasmar (approved):

PR #2725, opened 2021-03-17
feat(Public ECR): Added --build-image Options

CoshUS:
## Notes
Integration tests will be added in another PR.

### DIFF
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -4,7 +4,7 @@ CLI command for "build" command
 
 import os
 import logging
-from typing import List, Optional, Dict
+from typing import List, Optional, Dict, Tuple
 import click
 
 from samcli.cli.context import Context
@@ -181,9 +181,9 @@ def cli(
     parallel: bool,
     manifest: Optional[str],
     docker_network: Optional[str],
-    container_env_var: Optional[List[str]],
+    container_env_var: Optional[Tuple[str]],
     container_env_var_file: Optional[str],
-    build_image: Optional[str],
+    build_image: Optional[Tuple[str]],
     skip_pull_image: bool,
     parameter_overrides: dict,
     config_file: str,
@@ -232,9 +232,9 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     skip_pull_image: bool,
     parameter_overrides: Dict,
     mode: Optional[str],
-    container_env_var: Optional[List[str]],
+    container_env_var: Optional[Tuple[str]],
     container_env_var_file: Optional[str],
-    build_image: Optional[List[str]],
+    build_image: Optional[Tuple[str]],
 ) -> None:
     """
     Implementation of the ``cli`` method
@@ -390,12 +390,12 @@ def _get_mode_value_from_envvar(name: str, choices: List[str]) -> Optional[str]:
     return mode
 
 
-def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
+def _process_env_var(container_env_var: Optional[Tuple[str]]) -> Dict:
     """
     Parameters
     ----------
-    container_env_var : list
-        the list of command line env vars received from --container-env-var flag
+    container_env_var : Tuple
+        the tuple of command line env vars received from --container-env-var flag
         Each input format needs to be either function specific format (FuncName.VarName=Value)
         or global format (VarName=Value)
 
@@ -434,14 +434,14 @@ def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
     return processed_env_vars
 
 
-def _process_image_options(image_args: Optional[List[str]]) -> Dict:
+def _process_image_options(image_args: Optional[Tuple[str]]) -> Dict:
     """
     Parameters
     ----------
-    image_args : list
-        List of command line image options in the format of
-        Function1=public.ecr.aws/abc/abc:latest or
-        public.ecr.aws/abc/abc:latest
+    image_args : Tuple
+        Tuple of command line image options in the format of
+        "Function1=public.ecr.aws/abc/abc:latest" or
+        "public.ecr.aws/abc/abc:latest"
 
     Returns
     -------
--- a/samcli/commands/build/command.py
+++ b/samcli/commands/build/command.py
@@ -4,7 +4,7 @@ CLI command for "build" command
 
 import os
 import logging
-from typing import List, Optional, Dict, Tuple, Union
+from typing import List, Optional, Dict, Tuple
 import click
 
 from samcli.cli.context import Context
@@ -130,13 +130,13 @@ $ sam build MyFunction
     default=None,
     multiple=True,  # Can pass in multiple build images
     required=False,
-    help="Docker image URIs for building functions."
-    "You can specify for each individual function with "
-    "(--build-image FunctionLogicalID=public.ecr.aws/sam/build-nodejs14.x:latest)."
+    help="Container image URIs for building functions. "
     "You can specify for all functions with just the image URI "
-    "(--build-image public.ecr.aws/sam/build-nodejs14.x:latest)."
-    "A combination of the two can be used. If a function does not have build image specified or"
-    "a image uri for all functions, the default SAM CLI build images will be used.",
+    "(--build-image public.ecr.aws/sam/build-nodejs14.x:latest). "
+    "You can specify for each individual function with "
+    "(--build-image FunctionLogicalID=public.ecr.aws/sam/build-nodejs14.x:latest). "
+    "A combination of the two can be used. If a function does not have build image specified or "
+    "an image URI for all functions, the default SAM CLI build images will be used.",
 )
 @click.option(
     "--parallel",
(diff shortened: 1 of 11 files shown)


samcli/commands/build/command.py
```diff
@@ -418,3 +438,34 @@ def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
...
+        Function as key and the corresponding image URI as value.
+        Global default image URI is contained in the None key.
+    """
+    build_images: Dict[Union[str, None], str] = dict()
+    if image_args:
+        for build_image_string in image_args:
+            function_name: Union[str, None] = None
+            if "=" in build_image_string:
+                parts = build_image_string.split("=", 1)
+                function_name = parts[0]
```
qingchm, line 464: Are we adding any checks here so that function_name and image_uri won't be empty?

samcli/lib/build/app_builder.py
```diff
@@ -428,28 +432,23 @@ def _build_layer(
...
-            options = ApplicationBuilder._get_build_options(layer_name, config.language, None)
+                self._build_function_on_container(
+                    config, code_dir, artifact_subdir, manifest_path, build_runtime, options, container_env_vars
+                )
+            else:
+                self._build_function_in_process(
+                    config, code_dir, artifact_subdir, scratch_dir, manifest_path, build_runtime, options
+                )
 
-            build_method(
```
qingchm, line 443: The refactoring makes sense here, since we are adding more and more container unique features, just wondering do we have make pr passing so far? I think this change might affect many test cases?

samcli/lib/build/app_builder.py
```diff
@@ -520,21 +519,28 @@ def _build_function(  # pylint: disable=R1710
...
                 # By default prefer to build in-process for speed
-                build_method = self._build_function_in_process
                 if self._container_manager:
-                    build_method = self._build_function_on_container
-                    return build_method(
+                    image = None
+                    if self._build_images is not None:
+                        if function_name in self._build_images:
+                            image = self._build_images[function_name]
+                        elif None in self._build_images:
```
qingchm, line 527: Can we give this a name or something and specify in the doc? I don't really support having None as a key

CoshUS: We can but I don't see a reason to not use `None` as key for this use case. Having a special string for default would mean we need to make sure all references are matching with an extra constant.

samcli/lib/build/app_builder.py
CoshUS: Unit tests are passing. Integration tests have not been modified and not in the scope of this PR.

samcli/commands/build/command.py
```diff
@@ -124,6 +124,20 @@
...
+    default=None,
+    multiple=True,  # Can pass in multiple build images
+    required=False,
+    help="Docker image URIs for building functions."
+    "You can specify for each individual function with "
+    "(--build-image FunctionLogicalID=public.ecr.aws/sam/build-nodejs14.x:latest)."
+    "You can specify for all functions with just the image URI "
+    "(--build-image public.ecr.aws/sam/build-nodejs14.x:latest)."
+    "A combination of the two can be used. If a function does not have build image specified or"
+    "a image uri for all functions, the default SAM CLI build images will be used.",
```
hoffa, line 139: - Needs space after `or`. Same for L133, L135 and L137.
- typo: image URI

samcli/commands/build/command.py
```diff
@@ -124,6 +124,20 @@
     type=click.Path(),  # Must be a json file
     help="Path to environment variable json file (e.g., env_vars.json) to pass into build containers",
 )
+@click.option(
+    "--build-image",
+    "-bi",
+    default=None,
+    multiple=True,  # Can pass in multiple build images
+    required=False,
+    help="Docker image URIs for building functions."
```
hoffa, line 133: I wouldn't overfit to Docker unless it's a Docker-specific technology. These are OCI containers and any runtime can run them. "Container image" should be enough.

samcli/commands/build/command.py
```diff
@@ -418,3 +438,34 @@ def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
...
+    Returns
+    -------
+    dictionary
+        Function as key and the corresponding image URI as value.
+        Global default image URI is contained in the None key.
+    """
+    build_images: Dict[Union[str, None], str] = dict()
+    if image_args:
+        for build_image_string in image_args:
+            function_name: Union[str, None] = None
```
hoffa, line 461: More idiomatic is `Optional[str]`

samcli/commands/build/command.py
```diff
@@ -418,3 +438,34 @@ def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
...
+        "public.ecr.aws/abc/abc:latest"
+
+    Returns
+    -------
+    dictionary
+        Function as key and the corresponding image URI as value.
+        Global default image URI is contained in the None key.
+    """
+    build_images: Dict[Union[str, None], str] = dict()
+    if image_args:
```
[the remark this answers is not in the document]
CoshUS: Added _parse_key_value_pair()

samcli/commands/build/command.py
CoshUS: Added empty checks

samcli/commands/build/command.py
CoshUS: Fixed.

samcli/commands/build/command.py
CoshUS: Updated.

samcli/commands/build/command.py
CoshUS: Updated.

hoffa:
More of a general comment; we should test the following cases:

1. `--build-image a:b` pulls only `a:b` and not `a:latest`
2. `--build-image a` pulls only `a:latest` and not all tags (see e.g. https://github.com/docker/docker-py/issues/2693)

samcli/commands/build/command.py
```diff
@@ -124,6 +124,20 @@
...
 )
+@click.option(
+    "--build-image",
+    "-bi",
+    default=None,
+    multiple=True,  # Can pass in multiple build images
+    required=False,
+    help="Container image URIs for building functions. "
+    "You can specify for all functions with just the image URI "
+    "(--build-image public.ecr.aws/sam/build-nodejs14.x:latest). "
```
jfuss, line 135: Why not have two options for this? We have seen in the past overriding options for different cases can become complex quickly. Seems like having one option for all and one option to specify per function might be safer to have. 

Side question: Did we consider putting this data on the function in the template Metadata like we do for other build information?

samcli/commands/build/command.py
```diff
@@ -418,3 +433,53 @@ def _process_env_var(container_env_var: Optional[List[str]]) -> Dict:
...
+    dictionary
+        Function as key and the corresponding image URI as value.
+        Global default image URI is contained in the None key.
+    """
+    build_images: Dict[Optional[str], str] = dict()
+    if image_args:
+        for build_image_string in image_args:
+            function_name, image_uri = _parse_key_value_pair(build_image_string)
+            if not image_uri:
+                LOG.error("Invalid command line --build-image input %s, skipped.", build_image_string)
```
jfuss, line 458: Why should we be skipping on invalid input rather than failing the command?

samcli/lib/build/app_builder.py
jfuss: > Integration tests have not been modified and not in the scope of this PR.

I am concerned that we are splitting up prs on testing. This will make it much harder to understand changes. I know this PR is going into a feature branch but we lose sight of what is actually changing and what needs to be added later. I would really discourage this behavior and push us to have the context within the same pr. There can be times for this, but it should not be the typical process we follow. If this was a contribution from a community member, we would be asking them to add the integ tests into this pr. Why are we treating ourselves in a different manner?

CoshUS: > > Integration tests have not been modified and not in the scope of this PR.
> 
> I am concerned that we are splitting up prs on testing. This will make it much harder to understand changes. I know this PR is going into a feature branch but we lose sight of what is actually changing and what needs to be added later. I would really discourage this behavior and push us to have the context within the same pr. There can be times for this, but it should not be the typical process we follow. If this was a contribution from a community member, we would be asking them to add the integ tests into this pr. Why are we treating ourselves in a different manner?

This was done for this specific PR because of capacity reasons and trying to make implementation and tests parallel between Qingchuan and me.

samcli/commands/build/command.py
qingchm: Yep I would agree on failing the command here. This behaviour here is to align with how we parse the env vars. But I would agree that the building should not happen if the image cannot be parsed properly

samcli/commands/build/command.py
CoshUS: This was done for consistency with the env var option as both of the have the same format and similar effect. We can bring this into UX review for a further discussion. 
Metadata in the template is not in scope of this change, but we can explore it in the future.

samcli/commands/build/command.py
CoshUS: Updated.

samcli/commands/build/command.py
qingchm: I believe right now this is safe to have them in one option. Previously the container env var option have the the same overriding and precedence pattern which we feel that is safe and would not be too complex to cause any issues. Anyways we have a ux review set up on Friday as well, @jfuss if you can join to discuss this it would be nice as well!

qingchm (approved):

samcli/lib/build/app_builder.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         docker_client: Optional[docker.DockerClient] = None,
         container_env_var: Optional[Dict] = None,
         container_env_var_file: Optional[str] = None,
+        build_images: Optional[dict] = None,
```
qingchm, line 70: Missed this during review, will change it to Dict in next pr for consistency

samcli/commands/build/command.py
hoffa: > Did we consider putting this data on the function in the template Metadata like we do for other build information?

I think this is a good call-out. We should have a consistent story about how parameters are passed. How do we choose between the template or CLI arguments? Why are CLI arguments are preferred here versus the template?

samcli/local/docker/manager.py
```diff
@@ -142,6 +142,8 @@ def pull_image(self, image_name, tag="latest", stream=None):
         DockerImagePullFailedException
             If the Docker image was not available in the server
         """
+        if tag is None:
+            tag = image_name.split(":")[1] if ":" in image_name else "latest"
```
hoffa, line 146: Which tag does it use if `image_name` has a tag and `tag` is set?

PR #2735, opened 2021-03-18
Native Installer script/configs for nightly builds

wchengru:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


installer/assets/install_nightly_build
```diff
@@ -0,0 +1,145 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
```
sriram-mv, line 13: what is different about this file and https://github.com/aws/aws-sam-cli/blob/develop/installer/assets/install . If there are differences, can they be parameterized instead?

installer/pyinstaller/samcli-nightly.spec
```diff
@@ -0,0 +1,29 @@
+# -*- mode: python -*-
+block_cipher = None
+exe_name = 'sam-nightly'
```
sriram-mv, line 3: can the exe_name be parameterized too?

sriram-mv (changes requested):

installer/assets/install_nightly_build
wchengru: This script is for customer to install the package. We don't want to introduce extra parameters to make the installation more complate. So I just copy that script and do necessary.

installer/pyinstaller/build-linux.sh
```diff
@@ -47,13 +55,17 @@ echo "Installing PyInstaller"
...
+../venv/bin/python -m PyInstaller -D --clean installer/pyinstaller/${sam_cli_spec_filename}
 
 
 mkdir pyinstaller-output
 
-mv dist/sam pyinstaller-output/dist
+mv dist/${binary_name} pyinstaller-output/dist
 cp installer/assets/* pyinstaller-output
+if ! [ "$nightly_build" = "" ]; then
+    echo "Renaming install script name"
```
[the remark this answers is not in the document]
wchengru: Done

installer/pyinstaller/samcli-nightly.spec
wchengru: These files are load as parameter of pyinstaller, so I feel that loading different file is a better way and not easy to make mistake.

installer/pyinstaller/build-linux.sh
```diff
@@ -47,15 +48,32 @@ echo "Installing PyInstaller"
...
 mkdir pyinstaller-output
-
-mv dist/sam pyinstaller-output/dist
+dist_folder="sam"
+if ! [ "$nightly_build" = "" ]; then
+    echo "using dist_folder with nightly build"
+    dist_folder="sam-nightly"
+fi
+echo "dist_folder=$dist_folder"
+mv dist/$dist_folder pyinstaller-output/dist
```
hoffa, line 67: Please `shellcheck` the script. This will break if `dist_folder` includes a space.

installer/assets/install_nightly_build
wchengru: The only differences are the default install path, and the binary name. The install script can replace them in place now.

installer/pyinstaller/samcli-nightly.spec
wchengru: I have updated the install script, the name here can be replaced by script now.

installer/pyinstaller/build-linux.sh
wchengru: Done. And checked with `shellcheck`.

installer/pyinstaller/build-linux.sh
```diff
@@ -47,15 +48,32 @@ echo "Installing PyInstaller"
 
 echo "Building Binary"
 cd src
+if ! [ "$nightly_build" = "" ]; then
+    echo "Updating samcli.spec with nightly build"
+    sed -i "s/'sam'/'sam-nightly'/g" installer/pyinstaller/samcli.spec
```
hoffa, line 53: `-i` as-is won't work on BSD `sed` AFAIK. If it doesn't matter if we create files, `-i.bak` should work on both BSD and GNU.

installer/pyinstaller/build-linux.sh
```diff
@@ -47,15 +48,32 @@ echo "Installing PyInstaller"
 
 echo "Building Binary"
 cd src
+if ! [ "$nightly_build" = "" ]; then
```
mndeveci, line 51: Can we move nightly build check into a variable and use it in all places here?

installer/pyinstaller/build-linux.sh
wchengru: done.

installer/pyinstaller/build-linux.sh
wchengru: done.

installer/pyinstaller/build-linux.sh
```diff
@@ -11,6 +12,13 @@ if [ "$python_version" = "" ]; then
     python_version="3.7.9";
 fi
 
+if ! [ "$nightly_build" = "" ]; then
+    echo "Building native installer with nightly build"
+    is_nightly="true"; else
```
mndeveci, line 17: nit: is this how `else` written in bash files? If we can move this into new line it will be better.

mndeveci (approved):
One small comment on the styling

installer/pyinstaller/build-linux.sh
wchengru: Thanks! Fixed this.

installer/pyinstaller/build-linux.sh
```diff
@@ -47,15 +57,34 @@ echo "Installing PyInstaller"
 
 echo "Building Binary"
 cd src
+if [ "$is_nightly" = "true" ]; then
+    echo "Updating samcli.spec with nightly build"
+    sed -i.bak "s/'sam'/'sam-nightly'/g" installer/pyinstaller/samcli.spec
+    rm installer/pyinstaller/samcli.spec.bak
```
hoffa, line 63: I'd put this in a separate function to make it clear what it does and prevent human error on the names. e.g.

```shell
sed_inplace() {
    sed -i.bak "$1" "$2"
    rm "$2.bak"
}
```

hoffa (approved):

installer/pyinstaller/build-linux.sh
wchengru: Thanks, I agree that making the script into function is a good idea. Since this is a common usage to remove backup. I will keep this to save testing effort. Will use this as a function if we need more in-place changes in the future.

Issue #2703, opened 2021-03-11
`sam build` handles when some functions/layers already built by other methods (with CodeUri pointing to local build artifact ZIPs)


mfriesen:
### Describe your idea/feature/enhancement

Previous to sam cli 1.20 you could specify a Lambda zip file in AWS::Serverless::Function CodeUri and sam would build and create an S3 link for the lambda zip file. (https://github.com/aws/aws-sam-cli/issues/2693)

As of 1.20, sam only supports local directories where it try and build the software or S3 urls. This really limits the use cases for sam for building large enterprise systems or being able to reuse code without copy and pasting it into every project that uses the same "module".

### Proposal

The AWS::Serverless::Function CodeUri parameter should support local Zip files. When building it should upload the zip file to the SAM cli S3 bucket and replace the CodeUri with the S3 link.

### Edit [by @aahung] 
The issue description does not reflect the root issue. For clarification, please see the 1st comment below.

aahung:
AWS::Serverless::Function always support CodeUri parameter as a ZIP file, before or after 1.20.0. 

@mfriesen let me put some clarification here, following the discussion from #2693:

**First, there are two type of ZIP files, source code used by SAM to build into artifact and build artifacts** Here the ZIP file referred by @mfriesen is the latter. 

Some background of @mfriesen's template: it is a template with nested stacks in it:

- root template does not have buildable resources (i.e., local AWS::Serverless::Function or AWS::Serverless::LayerVersion).
- nested stacks contains functions, but @mfriesen already build them using other approach than `sam build`, that's why a zip file is used.
- the current workflow is
    1. `sam build`
    2. `sam package` or `sam deploy`

Before 1.20.0, the step 1 (`sam build`) actually does not build anything excepting moving all the root template file to `.aws-sam/build`. now in 1.20.0, with nested stack support, `sam build` tries to build functions in nested stacks, that are already built, which caused error. 

_In short, 1.20.0 does not change the behavior on ZIP file handling, it just tries to build resources which was left behind before. `sam build` always supports CodeUri to be a zip file, just not local artifact-type zip file._

Since the root template does not have buildable resources, the original suggestion given to @mfriesen was to skip step 1, and the issue was resolved.

As a follow up discussin in ##2693, @mfriesen came up with a new scenario, what if in the template, some functions use artifact-type ZIP as `CodeUri` and some functions use source code `CodeUri`? How SAM CLI should handle this scenario?

aahung:
[...]

PR #2737, opened 2021-03-18
fix: Skip building functions/layers if codeuri is a zip file

aahung:

#2703

currently (tested with 1.10.0, 1.19.0 and 1.21.1), when codeuri is a zip file, `sam build` will try to build using the zip file's path as source directory and `LambdaBuilder` will fail with error like this:
```
uilding codeuri: Archive.zip runtime: python3.7 metadata: {} functions: ['Function2']
Running PythonPipBuilder:ResolveDependencies

Build Failed
Error: PythonPipBuilder:ResolveDependencies - Requirements file not found: /Users/xinhol/Downloads/test-20210317/Archive.zip/requirements.txt
``` 

Some debugging context:

image

This issue becomes more impacting because before 1.20.0 was released, customers need to build functions in nested stacks themselves. Setting `codeuri` to a artifact zip was one of the practices to package/deploy nested stacks. With 1.20.0, `sam build` will try to build those zip files, which results in build error. With this patch, zip files will not be built again.

Skip building functions or layers when their codeuri is a zip file. 
**It is excluded in BuildContext instead of providers, because we still want these functions to be local invoke-able.**

no side effect.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -265,3 +265,27 @@ def _collect_single_buildable_layer(
...
         resource_collector.add_layer(layer)
+
+    @staticmethod
+    def _is_function_buildable(function: Function):
+        # no need to build inline functions
+        if function.inlinecode:
+            LOG.debug("Skip building inline function: %s", function.full_path)
+            return False
+        # no need to build functions that are already packaged as a zip file
+        if isinstance(function.codeuri, str) and function.codeuri.endswith(".zip"):
```
mndeveci, line 276: Are we going to consider an S3 location as well?

aahung: I am still checking details why we handled the S3 this way (this code exists before our git history) with @jfuss. I am gonna handle s3 in a separate PR

```py
        # CodeUri can be a dictionary of S3 Bucket/Key or a S3 URI, neither of which are supported
        if isinstance(codeuri, dict) or (isinstance(codeuri, str) and codeuri.startswith("s3://")):
            if not ignore_code_extraction_warnings:
                LOG.warning(
                    "Lambda function '%s' has specified S3 location for CodeUri which is unsupported. "
                    "Using default value of '%s' instead",
                    name,
                    SamBaseProvider.DEFAULT_CODEURI,
                )
            return SamBaseProvider.DEFAULT_CODEURI
        return cast(str, codeuri)
```

sriram-mv (commented):
Overall, this looks good, should we add an integ test as well?

sriram-mv (approved):

samcli/commands/build/build_context.py
qingchm, line 276: should we add a strip() before checking the endswith?

samcli/commands/build/build_context.py
```diff
@@ -265,3 +265,27 @@ def _collect_single_buildable_layer(
...
+        return True
+
+    @staticmethod
+    def _is_layer_buildable(layer: LayerVersion):
+        # if build method is not specified, it is not buildable
+        if not layer.build_method:
+            LOG.debug("Skip building layer without a build method: %s", layer.full_path)
+            return False
+        # no need to build layers that are already packaged as a zip file
+        if isinstance(layer.codeuri, str) and layer.codeuri.endswith(".zip"):
```
qingchm, line 288: Same question here

samcli/commands/build/build_context.py
aahung: the `codeuri` is from yaml, so we don't need to.

qingchm (approved):

Issue #2551, opened 2021-01-25
sam build fails with "AWS quick start -> Java 11 -> Docker"


vilunet:
### Description:

Build fails with docker build at step 8  - File Not Found

### Steps to reproduce:

Just run AWS quick start to create a sample app and run sam build.

### Observed result:

Step 7/16 : COPY gradle/lambda-build-init.gradle ./build
 ---> Using cache
 ---> bfaaeb63805a
Step 8/16 : RUN ./gradlew --project-cache-dir $SCRATCH_DIR/gradle-cache -Dsoftware.amazon.aws.lambdabuilders.scratch-dir=$SCRATCH_DIR --init-script $SCRATCH_DIR/lambda-build-init.gradle build
 ---> Running in 37994a9fdc4b
←[0m←[91m: No such file or directory
←[0m
Build Failed
Error: HelloWorldFunction failed to build: The command '/bin/sh -c ./gradlew --project-cache-dir $SCRATCH_DIR/gradle-cache -Dsoftware.amazon.aws.lambdabuilders.scratch-dir=$SCRATCH_DIR --init-script $SCRATCH_DIR/lambda-build-init.gradle build' returned a non-zero code: 127

### Expected result:

build should return no error

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10 - Powershell
2. `sam --version`: 1.15.0

aahung:
Update: cannot reproduce it on macOS with hello world (zip) template. Will try it on windows

CoshUS:
This issue is due to git pulling the quick start repo with `\r\n` as the line ending on Windows.
A workaround is to modify `gradlew` and replace all `\r\n` to `\n`

CoshUS:
Fixed in 1.22.0 release.

PR #2738, opened 2021-03-18
fix: Added CR LF to LF Conversion for gradlew

CoshUS:
https://github.com/aws/aws-sam-cli/issues/2551

SAM CLI quick start Java gradle projects fail to build on Windows with --use-container or with image type.
Windows git clone will change gradlew line endings to \r\n which will not be recognized by Unix sh.

Replace all \r\n in gradlew to \n after cookiecutter.
Simply replacing when the repo is cloned will not work as cookiecutter also changes line endings.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hoffa:
> Windows git clone will change gradlew line endings to \r\n

Is this documented behavior?

CoshUS:
> > Windows git clone will change gradlew line endings to \r\n
> 
> Is this documented behavior?

git config --global core.autocrlf true
https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#:~:text=text%20eol%3Dcrlf%20Git%20will,LF%20endings%2C%20even%20on%20Windows.

The Windows git installer also prompts the users during installation on how newlines should be handled. The default is commit with \n and pull with \r\n.

samcli/lib/init/__init__.py
```diff
@@ -90,6 +92,9 @@ def generate_project(
     try:
         LOG.debug("Baking a new template with cookiecutter with all parameters")
         cookiecutter(**params)
+        # Fix gradlew line ending issue caused by Windows git
+        # Putting it after cookiecutter as it will change the line ending
+        _fix_gradlew_line_ending(output_dir)
```
hoffa, line 97: So if I'm understanding this, the problem here is that by default `git` on Windows converts LF to CRLF. But when running that inside a (Linux) container with `--use-container`, the CRLF endings mess up the parsing?

Doesn't this mean `gradlew` is essentially unreadable on Windows afterwards, as the line endings are LF?

CoshUS: gradlew is a shell script and Windows can read both CRLF and LF.

hoffa: Won't it show up weird in Notepad for example?

I might not have full context here, but it feels this is fixing a symptom and not the root cause. Looking at the [Gradle docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html), it seems there's supposed to be a `gradlew.bat` as well?

hoffa: Actually never mind if it's a shell script it makes sense. Can you add a comment explaining that? Since it's what makes this essential to understand.

samcli/lib/init/__init__.py
```diff
@@ -104,3 +109,22 @@ def generate_project(
         raise InvalidLocationError(template=params["template"]) from e
     except CookiecutterException as e:
         raise GenerateProjectFailedError(project=name, provider_error=e) from e
+
+
+def _fix_gradlew_line_ending(path):
```
hoffa, line 114: Should we make this more generic so it can accept multiple paths? In case there are other Unix-only files that might be messed up by Windows.

samcli/lib/init/__init__.py
hoffa: If Windows can handle LF, can we disable `core.autocrlf`?

CoshUS: > If Windows can handle LF, can we disable `core.autocrlf`?

It's a global user preference, we shouldn't be changing that.

hoffa: Can't it be disabled for the pull only?

samcli/lib/init/__init__.py
CoshUS: Updated.

samcli/lib/init/__init__.py
CoshUS: We can add specific file .gitattributes in sam-app-template repo, but cookiecutter will still mess up the line endings as the processing is done after git. Related PR: https://github.com/cookiecutter/cookiecutter/pull/1407.

samcli/lib/utils/osutils.py
```diff
@@ -154,3 +155,21 @@ def copytree(source, destination, ignore=None):
             copytree(new_source, new_destination, ignore=ignore)
         else:
             shutil.copy2(new_source, new_destination)
+
+
+def convert_files_to_unix_line_endings(path: str, target_files: Optional[List[str]] = None) -> None:
```
sriram-mv, line 160: There is a test within sam build for java which uses gradlew, which could use this helper function. Right now I believe, it changes the line endings before starting the test, could you change that test to use this helper function as well?

sriram-mv (dismissed):

samcli/lib/utils/osutils.py
CoshUS: Updated.

hoffa (approved):

moelasmar (approved):

PR #2740, opened 2021-03-19
Extract the logic of managing SAM cloudformation stack from the the S…

elbayaaa:
Extract the logic of managing SAM CloudFormation stack to a utility module

NA

SAM bootstrap creates/manages a S3 bucket via a CFN stack("aws-sam-cli-managed-default"). This change is to decoupl the logic of managing a SAM CFN stack from this particular "aws-sam-cli-managed-default" stack so that the logic can be reused to manage any other SAM CFN stacks.

Nothing, this refactor doesn't cause any behavior change.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -0,0 +1,130 @@
...
+
+SAM_CLI_STACK_PREFIX = "aws-sam-cli-managed-"
+LOG = logging.getLogger(__name__)
+
+
+def manage_stack(profile, region, stack_name, template_body):
+    try:
+        if profile:
+            session = boto3.Session(profile_name=profile, region_name=region if region else None)
+            cloudformation_client = session.client("cloudformation")
```
elbayaaa, line 26: this is a new logic but doesn't not result in any behavior change as the method caller(see bootstrap.py) always pass profile=None (which how it used to be)

samcli/lib/bootstrap/bootstrap.py
```diff
@@ -4,129 +4,30 @@
...
-    if tags is None or outputs is None:
-        stack_state = stack.get("StackName", None)
-        msg = (
-            f"Stack {SAM_CLI_STACK_NAME} is missing Tags and/or Outputs information and therefore not in a "
-            f"healthy state (Current state:{stack_state}). Failing as the stack was likely not created "
-            f"by the AWS SAM CLI"
-        )
-        raise UserException(msg)
+    outputs = manage_cloudformation_stack(
+        profile=None, region=region, stack_name=SAM_CLI_STACK_NAME, template_body=_get_stack_template()
```
elbayaaa, line 18: setting profile to None as the existing code didn't use this param anywhere

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -0,0 +1,130 @@
...
+        else:
+            cloudformation_client = boto3.client(
+                "cloudformation", config=Config(region_name=region if region else None)
+            )
+    except ProfileNotFound as ex:
+        raise CredentialsError(
+            f"Error Setting Up Managed Stack Client: the provided AWS name profile '{profile}' is not found. "
+            "please check the documentation for setting up a named profile: "
+            "https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html"
+        ) from ex
```
elbayaaa, line 36: this is a new logic but doesn't not result in any behavior change as the method caller(see bootstrap.py) always pass profile=None (which how it used to be) so this `ProfileNotFound`exception will never be raised

samcli/lib/utils/managed_cloudformation_stack.py
moelasmar: I think we should not create a new session, the session is handled here

https://github.com/aws/aws-sam-cli/blob/f084cf2d84709e25fe2b7381bdca0d20d452d317/samcli/cli/context.py#L169-L186

so whenever we need to apply any changes to the AWS services session we can do it from this location only, like if we need to use specific services endpoints, so we can change in the default session, and we are sure that it will be applied to all AWS services connections

elbayaaa: This use case is different. The one you are referring to is for the --profile option where we execute a SAM CLI command against **one** AWS account. But for this use case we want to deploy a CFN template to **multiple** AWS accounts so we don't manage the AWS account through this --profile option. 

Also we don't

moelasmar (approved):

CoshUS (approved):

PR #2741, opened 2021-03-19
fix the flaky warm containers integration test cases

moelasmar:
try to fix the flaky warm containers integration test cases by doing the following:
- increase the waiting time in case of local test command and warm-containers option is eager, as the command may take more time at the beginning to build the required images, and start the lambda functions containers)
- remove the retry form the warm/lazy options test cases, as every time the test case got retried the number of containers will increase and this will cause that the test case will fail any way.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


wchengru (approved):

tests/integration/local/start_api/start_api_integ_base.py
```diff
@@ -58,20 +59,23 @@ def build(cls):
...
+        wait_time = 5
         if os.getenv("SAM_CLI_DEV"):
             command = "samdev"
 
         command_list = [command, "local", "start-api", "-t", cls.template, "-p", cls.port]
 
         if cls.container_mode:
             command_list += ["--warm-containers", cls.container_mode]
+            if cls.container_mode == ContainersInitializationMode.EAGER.value:
+                wait_time = 25
```
mndeveci, line 71: Is this related to `docker pull` time? And if so can we pull the images upfront and validate they are ready locally?

moelasmar: yes, mainly the image creation time

mndeveci: My worry is, if there will be any network issue, this 25 seconds might not be enough. If we pull and validate their existence in local, that would be safer option.

sriram-mv: is there some kind of callback we could add here, which periodically polls (with some backoff) the process to check if the containers are created? and we could add an overall timeout as well.

sriram-mv (changes requested):

Issue #2671, opened 2021-03-02
An error occurred (Throttling) when calling the DescribeStacks operation (reached max retries: 4)


smurphnturf:
### Relates to: 
No response to comments on:
https://github.com/aws/aws-sam-cli/issues/1610

### Description:
When running sam deploy with of two multiple stacks in parallel, Cloud Formation throttles SAM for calling describe-stacks too many times. Cloud Formation rate caps describe-stacks to 5 calls per second and this cannot be increased. In particular, this happens when two different provisioned lambdas are deployed with sam deploy as each deployment can take a few minutes 

### Steps to reproduce:
Deploy two provisioned lambdas in parallel at the same time

### Observed result:
```
2021-02-15 22:49:02 - Waiting for stack create/update to complete

CloudFormation events from changeset
-------------------------------------------------------------------------------------------------
ResourceStatus           ResourceType             LogicalResourceId        ResourceStatusReason   
-------------------------------------------------------------------------------------------------
CREATE_IN_PROGRESS       AWS::Lambda::LayerVers   RenderedPromotionReade   -                      
                         ion                      rDependenciesLayerOne9                          
                                                  4c42c52c2                                       
CREATE_IN_PROGRESS       AWS::Lambda::LayerVers   PromotionSubscriptionH   -                      
                         ion                      andlerDependenciesLaye                          
                                                  rOnefb4f0ee29e                                  
CREATE_COMPLETE          AWS::Lambda::LayerVers   RenderedPromotionReade   -                      
                         ion                      rDependenciesLayerOne9                          
                                                  4c42c52c2                                       
CREATE_IN_PROGRESS       AWS::Lambda::LayerVers   RenderedPromotionReade   Resource creation      
                         ion                      rDependenciesLayerOne9   Initiated              
                                                  4c42c52c2                                       
CREATE_COMPLETE          AWS::Lambda::LayerVers   PromotionSubscriptionH   -                      
                         ion                      andlerDependenciesLaye                          
                                                  rOnefb4f0ee29e                                  
CREATE_IN_PROGRESS       AWS::Lambda::LayerVers   PromotionSubscriptionH   Resource creation      
                         ion                      andlerDependenciesLaye   Initiated              
                                                  rOnefb4f0ee29e                                  
UPDATE_IN_PROGRESS       AWS::Lambda::Function    RenderedPromotionReade   -                      
                                                  rFunction                                       
[...]

PR #2748, opened 2021-03-22
feat: add --initial-reporting-sleep for sam deploy to fix (Throttling) when calling the DescribeStacks

smurphnturf:
#2671 

Enables sam deployments to be run in parallel without fear of AWS throttling and killing the deployment which dramatically decreases deployment times.

Allows for a configurable waiting period between the deployments Describe-Stack call that reports the state of the deployment. This will allow more than 1 same deploy command to run in parallel without Cloud Formation throttling Describe-Stacks at 5 calls/second

Default behaviour remains the same. A bad actor could supply a really large --initial-reporting-sleep making the deployment seemed like it just stalled. 

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/commands/_utils/options.py
```diff
@@ -279,6 +279,20 @@ def no_progressbar_option(f):
     return no_progressbar_click_option()(f)
 
 
+def initial_reporting_sleep_click_option():
+    return click.option(
+        "--initial-reporting-sleep",
```
sriram-mv, line 284: We probably will need a more general option name, I'm also wondering if this could be an env var instead to control behavior in context of where this is being run, for example, you are less likely to run into this (throttling) when running on a personal account, but will be probably run into it more often on accounts which already have high activity.

sriram-mv (changes requested):

samcli/commands/_utils/options.py
fschroder-slyp: If you want to go that way maybe https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure can be used as a template?

smurphnturf: Yeah that would be nice to be consistent with the aws cli. @sriram-mv would that be preferable? Though if you mean a context for enterprise vs personal or isolated might have far great implications on the SAM cli other than just the deployment reporting throttling.
 Otherwise very happy to change the name maybe something like "reporting-delay", "report-wait", "output-waiting", "output-sleep"

fschroder-slyp: hey @sriram-mv, can we clarify what is needed to move this PR forward?

smurphnturf: Hey @sriram-mv, our team stil hits this most days...

softprops: we're also getting hit with this more and more often I'd love to see this feature ship if even in some beta (changable) state

sriram-mv:
Apologies on having this stall, I was looking into this some more and I think the right approach would definitely be natively leveraging boto3's retry methodology as referenced in https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure

Looks like the line of code to be changed, would be passing a custom retry to the boto3 config at https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/deploy/deploy_context.py#L128

Happy to fully support this and getting PR reviews going again.

pjungermann:
> Apologies on having this stall, I was looking into this some more and I think the right approach would definitely be natively leveraging boto3's retry methodology as referenced in https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure
> 
> Looks like the line of code to be changed, would be passing a custom retry to the boto3 config at https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/deploy/deploy_context.py#L128
> 
> Happy to fully support this and getting PR reviews going again.

@sriram-mv as I commented at issue #2671 it is possible to use the retry config for boto3 via env vars or config file.

I'm wondering a bit if we need 2 retry logics, one for boto3 and one for sam-cli when describe stack events is called.

Of course, the boto3 config applies to all calls not just describe stack events. Maybe there is a use case for having additional/more retries for describe stack events.

softprops:
> Maybe there is a use case for having additional/more retries for describe stack events.

For what it's worth, describe stack events is the _only_ request that has ever failed for me with sam cli. I'd like to see the conversation focus in on this so we don't get distracted on how to solve retries for other calls that, in practice, are never effected by this issue because they aren't called repeatedly in the same way describe stack events is.

pjungermann:
> describe stack events is the _only_ request that has ever failed for me with sam cli.

I encountered issues for the StackCreateComplete waiter, too, which uses describe-stacks.
```
botocore.exceptions.WaiterError: Waiter StackCreateComplete failed: An error occurred (Throttling): Rate exceeded
```

In general, if only the describe-stack-events operation fails due to throttling, the low-level retry provided by botocore will also apply to it.

Do we need a different logic for this specific operation? Is there anything we want to do differently compared to the logic applied at botocore already? (exponential backoff by factor 2, considering various errors and exceptions)
And maybe the answer is yes.

If yes, supporting env vars (too) could be a good approach. E.g., `SAM_CLI_MAX_ATTEMPTS` (to align with existing ones like `SAM_CLI_TELEMETRY`).

sriram-mv:
@pjungermann The primary use of SAM CLI's own retry format is for showcasing stack events during sam deploy. I like the idea of usage an env var to enforce this behavior.

@softprops Does the usage of a custom boto config help in any manner? or do you still see the same issues?

softprops:
> Does the usage of a custom boto config help in any manner? or do you still see the same issues?

I'm a sam cli customer. Boto and external boto configurations feel like implementation details a sam cli customer shouldn't have to know or care about.

That said given the lack of explicit sam cli options, what would be the suggestion for a sam cli user using GitHub actions be? I just use the sam cli that comes standard with those build environments. For deployments that hit the throttle limit, they fail the action workflow, even when they might eventually succeed. This is the concrete use problem I'm hoping can be addressed

sriram-mv:
@softprops Thanks for the additional context, so the first I can think of is to classify the exceptions into hard fail vs soft fail cases. Describing the stack events is meant more for tracking the flow of the deployment rather than an indicator of success. SAM CLI can soft-fail with an exit code 0 in throttling cases, but still allow for the exception string to propagate through.

How does that sound as a way forward in the immediate case?

softprops:
That would solve part of the problem.

A new problem that would introduce is that I have my GitHub actions pipelines set up as a queue so only one deploy can happen at a time. When one deployment workflow completes with even with a soft fail, and the next kicks off, it creates the scenario with the second can fail with a stack update in progress (or rollback in progress) error from the previous deployment.

I'd really like the first deployment to await a cf completion state before giving up _or_ potentially provide a hook that cancels cf the update right after it gives up.

The ability to monitor a deployment with progress updates is my favorite feature of sam cli. Especially compared to cf deploy where there is no feedback cycle. Understandability it's tricky to deal with rate limits. I'd be happy to compromise on a longer delay between progress update checks if it would increase the likelyhood I could monitor to a completion state without getting rate limited

pjungermann:
@softprops your perspective makes sense. Users of sam-cli shouldn't be required to dig up config options of internals.

Maybe it could help though to add the viable options as part of sam-cli's documentation. Env vars like AWS_MAX_ATTEMPTS don't sound "boto-specific". They could apply to aws-sam-cli, too.

On the other hand, having more explicit options - maybe env vars as well as command arguments - for sam-cli seems like a good idea, too. Those could easily be translated/passed on to boto.

..or only used for the DescribeStacks calls as pointed out in some comments above. Personally, I don't really see much issue with having retries for all. Only cases like throttling, etc. are retried anyways. Means: cases which can be retried, cases which likely should be retried as part of any automation.

@sriram-mv I would agree with @softprops here. For us, it wouldn't be enough to just exit with code 0 if we get throttled. It is also about knowing that the stack creation was successful and not rolled back as well as that the stack creation was completed and isn't still in progress.

BourgoisMickael:
I'm having the same problem of throttling.

I have 50+ repositories, they have Github Actions that run on every PR to deploy stacks. We have a lot of PRs in parallel for different stacks. A lot of stacks takes 10+ minutes to deploy. We keep getting throttled and have to rerun the workflows multiple times to deploy everything.

Increasing retries won't fix anything as it will just send more and more requests and keep blocking the rate limit.

As it's in a pipeline I don't care about having realtime updates of status at the second.

Therefore it could be nice to have an option to send the describe stack request only once every minutes.

If I could just add something like `SAM_CLI_DEPLOY_DESCRIBE_STACK_INTERVAL=60000` in my pipeline env it would be perfect

jbernales5:
Hello @sriram-mv, do you believe the PR #3500 would be better suited to fix this issue? It looks like it implements all the requirements described above.

jfuss:
This PR seems to be abandoned by the original author. We have another PR open (though it's using an env var): https://github.com/aws/aws-sam-cli/pull/3500

I responded on https://github.com/aws/aws-sam-cli/pull/3500 but it seems like there is mix messaging between when each was reviewed and working with the team to drive to a decision. Will handle this fix through #3500.

Closing.

PR #2756, opened 2021-03-22
fix: Skip functions/layers pointing to s3 location in providers

aahung:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/sam_function_provider.py
```diff
@@ -387,43 +388,51 @@ def _parse_layer_info(
...
                 )
+                if found_layer:
+                    layers.append(found_layer)
 
         return layers
 
+    @staticmethod
+    def _locate_layer_from_ref(
+        stack: Stack, layer: Dict, use_raw_codeuri: bool = False, ignore_code_extraction_warnings: bool = False
+    ) -> Optional[LayerVersion]:
```
aahung, line 402: pylint complains about `_parse_layer_info()` being too complex, here I extract some logic to `_locate_layer_from_ref()`

samcli/lib/providers/sam_base_provider.py
```diff
@@ -107,45 +128,6 @@ def _extract_sam_function_imageuri(resource_properties: Dict, code_property_key:
...
-        # CodeUri can be a dictionary of S3 Bucket/Key or a S3 URI, neither of which are supported
-        if isinstance(codeuri, dict) or (isinstance(codeuri, str) and codeuri.startswith("s3://")):
-            if not ignore_code_extraction_warnings:
-                LOG.warning(
-                    "Lambda function '%s' has specified S3 location for CodeUri which is unsupported. "
-                    "Using default value of '%s' instead",
-                    name,
-                    SamBaseProvider.DEFAULT_CODEURI,
-                )
-            return SamBaseProvider.DEFAULT_CODEURI
```
aahung, line 146: explain: now the warning is being handled by `SamFunctionProvider` and `SamLayerProviders`.

samcli/lib/providers/sam_base_provider.py
```diff
@@ -27,6 +27,13 @@ class SamBaseProvider:
     CLOUDFORMATION_STACK = AWS_CLOUDFORMATION_STACK
     DEFAULT_CODEURI = "."
 
+    CODE_PROPERTY_KEYS = {
```
sriram-mv, line 30: `sam package` already has a bunch of these constants for package-able code paths, would it make sense to refactor that, so thats its usable across the code base?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -387,43 +388,51 @@ def _parse_layer_info(
...
+            code_property_key = SamBaseProvider.CODE_PROPERTY_KEYS[resource_type]
+            if SamBaseProvider._is_code_uri_s3(layer_properties.get(code_property_key)):
+                # Content can be a dictionary of S3 Bucket/Key or a S3 URI, neither of which are supported
+                if not ignore_code_extraction_warnings:
+                    SamFunctionProvider._warn_code_extraction(resource_type, layer_logical_id, code_property_key)
+                return None
+            codeuri = SamBaseProvider._extract_codeuri(layer_properties, code_property_key)
+
+        if codeuri and not use_raw_codeuri:
+            LOG.debug("--base-dir is presented not, adjusting uri %s relative to %s", codeuri, stack.location)
```
sriram-mv, line 425: There's some grammar issue on this sentence.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1812,3 +1812,54 @@ def test_nested_build(self, use_container, cached, parallel):
...
+    template = "template-with-s3-code.yaml"
+    EXPECTED_FILES_PROJECT_MANIFEST = {
+        "__init__.py",
+        "main.py",
+        "numpy",
+        # 'cryptography',
+        "requirements.txt",
+    }
+
+    # @pytest.mark.flaky(reruns=3)
```
sriram-mv, line 1833: comment that might need to be removed?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1812,3 +1812,54 @@ def test_nested_build(self, use_container, cached, parallel):
...
+
+        if not SKIP_DOCKER_TESTS:
+            self._verify_build(
+                ["ServerlessFunction", "LambdaFunction"],
+                [""],  # there is only one stack
+                command_result,
+            )
+
+            overrides = self._make_parameter_override_arg({})
+            self._verify_invoke_built_functions(
```
sriram-mv, line 1858: The template is buildable because s3 based locations are explicitly skipped, what does the local invoke do here?

sriram-mv (changes requested):

tests/integration/buildcmd/test_build_cmd.py
aahung: again 🤦

samcli/lib/providers/sam_base_provider.py
aahung: good idea

samcli/lib/providers/sam_function_provider.py
aahung: good catch

tests/integration/buildcmd/test_build_cmd.py
aahung: good point. I don't think functions containing these remote layers should be local invocable at all. What do you think, disallow their invocation? @sriram-mv

aahung: I implemented this integ test base in #2760, will rebase this PR once #2760 is merged.

aahung: Correct: these two functions are still invocable

samcli/lib/providers/sam_base_provider.py
aahung: I have another PR ready for this purpose

samcli/lib/providers/sam_function_provider.py
```diff
@@ -387,40 +388,11 @@ def _parse_layer_info(
...
-                    codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
-
-                layers.append(
-                    LayerVersion(
-                        layer_logical_id,
-                        codeuri,
-                        compatible_runtimes,
-                        layer_resource.get("Metadata", None),
-                        stack_path=stack.stack_path,
-                    )
```
aahung, line 422: explain: pylint complains about `_parse_layer_info()` being too complex, here I extract some logic to `_locate_layer_from_ref()`

hoffa: We shouldn't have to make these large (in terms of diff size) unrelated refactors to get around tooling, the tooling should work for us. Makes it difficult to review and focus on the relevant changes. Could the new code be refactored in a way that doesn't require moving existing code around?

qingchm (commented):
Open question: should we add any special handling in local invoke/start api if the template contains s3 locations?

samcli/lib/providers/sam_function_provider.py
hoffa: The commits were reorganized to isolate pure refactors. Thanks a lot @aahung, that's very helpful! 🙌

aahung:
> Open question: should we add any special handling in local invoke/start api if the template contains s3 locations? I mean, any warnings?

good question. I think we should. As right now, customers will see a universal warning "AWS::Serverless::LayerVersion XXX has specified S3 location for ContentUri which is unsupported." I think it is good enough to raise the concern.

But for other cases (like #2760 pass-up example), we don't have the warning yet, but we have `LOG.debug` in place. https://github.com/aws/aws-sam-cli/blob/a8be3ae428b0fc9fb77aff9c4d6f562c3468aaee/samcli/lib/providers/sam_function_provider.py#L424-L428

samcli/lib/providers/sam_base_provider.py
```diff
@@ -61,14 +68,28 @@ def _extract_lambda_function_code(resource_properties: Dict, code_property_key:
...
-
         codeuri = resource_properties.get(code_property_key, SamBaseProvider.DEFAULT_CODEURI)
 
         if isinstance(codeuri, dict):
             return SamBaseProvider.DEFAULT_CODEURI
 
         return cast(str, codeuri)
 
+    @staticmethod
+    def _is_code_uri_s3(codeuri: Optional[Any]) -> bool:
```
hoffa, line 79: nit: isn't `Optional[Any] == Any`?

qingchm:
> #2760

To me the easier solution here would be during the build time if we have the is s3 function returned true, we can log a message warning about local invoke might fail. The more complex approach would be to add the message at invoke time so that they know why the invoke didn't work. For me personally I prefer the first option, but as I said this is an open question so more opinions are strongly welcomed

samcli/lib/providers/sam_base_provider.py
```diff
@@ -61,14 +68,28 @@ def _extract_lambda_function_code(resource_properties: Dict, code_property_key:
...
+        return (isinstance(codeuri, dict) and ("S3Bucket" in codeuri or "Bucket" in codeuri)) or (
+            isinstance(codeuri, str) and codeuri.startswith("s3://")
+        )
+
+    @staticmethod
+    def _warn_code_extraction(resource_type: str, resource_name: str, code_property: str) -> None:
```
hoffa, line 85: Feels slightly weird to have the logging specifically refactored out. Doesn't seem very meaningful on its own.

Would it make sense to (e.g.) instead have a method whose sole purpose is to check whether a resource uses an S3-based `CodeUri`? Something like that should also allow simplifying the 3 very similar use cases in this PR.

samcli/lib/providers/sam_base_provider.py
qingchm: Can codeuri be anything other than None, dict, str? Shouldn't we use something like Optional[str, Dict, None]?

aahung:
> > #2760
> 
> To me the easier solution here would be during the build time if we have the is_s3 function returned true, we can log a message warning about local invoke might fail. The more complex approach would be to add the message at invoke time so that they know why the invoke didn't work. For me personally I prefer the first option (kind of aligns with the layer behaviour), but as I said this is an open question so more opinions are strongly welcomed

That's a good point. For this PR, it is a bit out of the scope. I created a ticket (P45519700) to keep track of this issue.

samcli/lib/providers/sam_base_provider.py
```diff
@@ -61,14 +68,28 @@ def _extract_lambda_function_code(resource_properties: Dict, code_property_key:
...
+        LOG.warning(
+            "%s '%s' has specified S3 location for %s which is unsupported.",
```
qingchm, line 87: This warning feels a bit ambiguous here. Shouldn't we specify the scope of this "unsupported"?

samcli/lib/providers/sam_base_provider.py
aahung: I think we should do something like we did in https://github.com/aws/aws-sam-cli/pull/2691 (we had similar discussion before).

samcli/lib/providers/sam_base_provider.py
aahung: > nit: isn't Optional[Any] == Any?
yeah right.

I changed to `Optional[Union[str, Dict]]`

samcli/lib/providers/sam_base_provider.py
hoffa: Yeah so I think the whole flow in these methods feels weird as we're essentially dealing with two very similar abstractions (`AWS::Serverless::Function` and `AWS::Lambda::Function`) with vaguely-defined operations (e.g. methods that deal with both abstractions) but still keep repeating similar logic for both.

Not in this scope so I'm happy with this for now, but ideally I'm thinking we would have a well-defined abstraction to deal with functions, e.g. but wrapping them or performing a translation beforehand so we only deal with a subset of well-defined resource types with well-defined operations, so that we don't _need_ all this `if` and `isinstance` ceremony scattered across the codebase.

hoffa (approved):

tests/integration/buildcmd/test_build_cmd.py
qingchm: maybe only test building? I feel that testing invoke might be unnecessary here

aahung: no strong opinion here, the invoke is added based on the current sam-cli behavior, we might not need this in the future.  here removed

sriram-mv: does the invoke just skip? what is our current behavior? could we add a comment on what the behavior we are trying to test. This will be useful for future deep dives.

qingchm (approved):

sriram-mv (approved):

Issue #442, opened 2018-05-29
AttributeError in arn_generator.py


Kansuler:
**Description:**
Trying to run the command `sam validate` and get an error that is somewhat hard to understand. May this be related to python version? (Currently `2.7.12`) or something else?

**Observed result:**
```
2018-05-29 14:05:38 Starting new HTTPS connection (1): sts.amazonaws.com
2018-05-29 14:05:39 Starting new HTTPS connection (1): iam.amazonaws.com
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 11, in <module>
    sys.exit(cli())
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/home/kansuler/.local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/samcli/commands/validate/validate.py", line 28, in cli
    do_cli(ctx, template)  # pragma: no cover
  File "/usr/local/lib/python2.7/dist-packages/samcli/commands/validate/validate.py", line 42, in do_cli
    validator.is_valid()
  File "/usr/local/lib/python2.7/dist-packages/samcli/commands/validate/lib/sam_template_validator.py", line 60, in is_valid
    parameter_values={})
  File "/home/kansuler/.local/lib/python2.7/site-packages/samtranslator/translator/translator.py", line 57, in translate
    deployment_preference_collection = DeploymentPreferenceCollection()
[...]

PR #2759, opened 2021-03-23
fix(bug): Pass boto_session to SAM Translator library

jfuss:
When validating the sam template, SAM CLI requires credentials to get
the IAM Manged Policies. SAM Translator also requires the region in order
to figure out the parition. Previously, SAM Translator assumed this to be
on the Env but SAM CLI could get this information from a command line argument
or a profile. This commit passes the boto_session into the SAM Translator lib
(v1.35.0 or later), so that SAM Translator can figure out the partition from the
information passed to SAM CLI.

#442

SAM will through an AttributeError if a region is None (it assumes there will be a region). Locally, there was no way to pass the region/profile into the SAM Translator before v1.35.0. With v1.35.0 and above, there is an optional parameter for passing in a `boto_session` which allows SAM CLI to forward information from the CLI into the SAM Translator Lib.

Passes Boto Session to the SAM Translator lib.

None

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hoffa (approved):

mndeveci (approved):

samcli/commands/validate/validate.py
```diff
@@ -49,13 +51,20 @@ def do_cli(ctx, template):
...
+    )
 
     try:
         validator.is_valid()
     except InvalidSamDocumentException as e:
         click.secho("Template provided at '{}' was invalid SAM Template.".format(template), bg="red")
         raise InvalidSamTemplateException(str(e)) from e
+    except NoRegionFound as no_region_found_e:
+        raise UserException(
+            "AWS Region was not found. Please configure your region through a profile or --region option",
```
sriram-mv, line 65: should we add AWS_REGION env var too?

sriram-mv (changes requested):

samcli/commands/validate/validate.py
jfuss: I am not sold on adding [`AWS_DEFAULT_REGION`](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#guide-configuration). Yes customers can set this as it is resolved by boto3, but I don't see the need to add every single option here. Most customers are going to use `--region` or define them through a profile locally.

Is there a specific reason why you think we should add it?

sriram-mv: does sam validate automatically pick up the env vars? just trying to see if we can reduce friction where we can.

sriram-mv (approved):

samcli/commands/validate/validate.py
jfuss: Yes. We create a `boto_session` like we do in other places throughout. When `None` is passed, boto will resolve for us.

PR #2760, opened 2021-03-23
fix: sam build fails when layer logicalID shared across stacks

### LINKED ISSUE #2724: Layers shared in multiple stacks cause build/deploy to fail
The issue I am facing seems to be because of a recent update to SAM cli. Since my SAM templates did not change recently and my deploy was running fine up until a few days ago. Possibly a bug introduced in 1.21.x or 1.20.0.

### Description:

`sam build` errors out with following message:
`Error: L is an Invalid Layer Arn.`

_The following steps will help replicate the issue, but please refer to the "debug" output at the bottom. Because I think the underlying problem is exposed there._

I create a nested application in this case. In the parent application's template there are 2 layers defined. I pass the layer ARNs from parent to child template using Parameters. The error origins where the layer ARNs are specified inside a lambda function resource definition.

### Steps to reproduce:

Parent template:
```
Transform: AWS::Serverless-2016-10-31

Parameters:
  AWSAccountId:
    Type: String
    Default: "123"
  IAMRoleName:
    Type: String
    Default: "role"

Resources:
  OneLayer:
    Properties:
      LayerName: "one-layer"
      ContentUri: ../layers/one/
      CompatibleRuntimes:
        - python3.8
      RetentionPolicy: Delete

  AnotherLayer:
    Properties:
      LayerName: "another-layer"
      ContentUri: ../layers/another/
      CompatibleRuntimes:
        - python3.8
      RetentionPolicy: Delete

  SomeLambdaFunction:
    Properties:
      CodeUri: ../lambdas/example/
      FunctionName: "SomeLambdaName"
      Handler: test.lambda_handler
      Role: !Ref IAMRole
      Layers:
        - !Ref OneLayer
        - !Ref AnotherLayer

  TestApp:
    Type: AWS::Serverless::Application
    Properties:
      Location: apps/test.yaml
      Parameters:
        IAMRole: !Join ['', ["arn:aws:iam::", !Ref AWSAccountId, ":role/", !Ref IAMRoleName]]
        Layers: !Join [",", [!Ref OneLayer, !Ref AnotherLayer]]
```
Nested App template:
```
Transform: AWS::Serverless-2016-10-31

Parameters:
  IAMRole:
    Type: String
  Layers:
    Type: CommaDelimitedList

Resources:
  NestedLambda:
    Properties:
      CodeUri: ../lambdas/example/
      FunctionName: "NestedLambdaName"
      Handler: nested.lambda_handler
      Role: !Ref IAMRole
      Layers: !Ref Layers
```

### Observed result:

```
Error: L is an Invalid Layer Arn.
```
### Expected result:

```
Build Succeeded

Built Artifacts  ...
.
.
```

1. OS: MacOS/Ubuntu
2. `sam --version`: 1.21.1/1.21.0
3. AWS region: us-east-1

With the `--debug` flag the log statement in output that stands out is:
`Unable to resolve property Layers: {'Fn::Join': [',', [{'Ref': 'OneLayer'}, {'Ref': 'AnotherLayer'}]]}. Leaving as is.`

In my actual template I can see more statements like this for other parameters which use `Join` function in value.

aahung:

fix #2724

sam-cli does not support resolving intrinsic functions in parameters/outputs, which results in incorrect layer reference when the layer is passed from some other stacks using intrinsic functions.

Before we have a great support of intrinsic functions, customers should still be able to build their stacks. Incorrect resolution of layers should only affect `local *`.

Delay the validation process and only raise exception when certain layer properties are used.
I added two integration tests to cover both using parameters (pass down) and outputs (pass up). 

*pass up: template-pass-up.yaml, pass down: template-pass-down.yaml
- pass up before: 
  - `sam build` works, 
  - `sam local invoke` works
- pass up now: 
  - `sam build` works, 
  - `sam local invoke`: ~`Error: {'Fn::GetAtt': ['ChildApp', 'Outputs.LayerVersion']} is an Unsupported Intrinsic`~ works
- pass down before: 
  - `sam build` failed with `invalid layer arn` error
  - `sam local invoke` failed with `invalid layer arn` error
- pass down now: 
  - `sam build` works
  - `sam local invoke` failed with `invalid layer arn` error

~for the "pass up" case, customers won't be able to local invoke anymore. This should be the expected behavior.~

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -145,6 +145,8 @@ class LayerVersion:
             compatible_runtimes = []
         if metadata is None:
             metadata = {}
+        if not isinstance(arn, str):
+            raise UnsupportedIntrinsic("{} is an Unsupported Intrinsic".format(arn))
 
         self._stack_path = stack_path
         self._arn = arn
@@ -222,11 +224,6 @@ class LayerVersion:
 
     @property
     def arn(self) -> str:
-        # because self.arn is only used in local invoke.
-        # here we delay the validation process for self._arn, rather than in __init__() to ensure
-        # customers still have a smooth build experience.
-        if not isinstance(self._arn, str):
-            raise UnsupportedIntrinsic("{} is an Unsupported Intrinsic".format(self._arn))
         return self._arn
 
     @property
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -120,6 +120,9 @@ class LayerVersion:
 
     LAYER_NAME_DELIMETER = "-"
 
+    _name: Optional[str] = None
+    _version: Optional[int] = None
+
     def __init__(
         self,
         arn: str,
@@ -142,15 +145,11 @@ class LayerVersion:
             compatible_runtimes = []
         if metadata is None:
             metadata = {}
-        if not isinstance(arn, str):
-            raise UnsupportedIntrinsic("{} is an Unsupported Intrinsic".format(arn))
 
         self._stack_path = stack_path
         self._arn = arn
         self._codeuri = codeuri
         self.is_defined_within_template = bool(codeuri)
-        self._name = LayerVersion._compute_layer_name(self.is_defined_within_template, arn)
-        self._version = LayerVersion._compute_layer_version(self.is_defined_within_template, arn)
         self._build_method = cast(Optional[str], metadata.get("BuildMethod", None))
         self._compatible_runtimes = compatible_runtimes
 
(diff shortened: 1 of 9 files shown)


hoffa (approved):

CoshUS (approved):

samcli/lib/providers/provider.py
```diff
@@ -145,6 +145,8 @@ def __init__(
             compatible_runtimes = []
         if metadata is None:
             metadata = {}
+        if not isinstance(arn, str):
+            raise UnsupportedIntrinsic("{} is an Unsupported Intrinsic".format(arn))
```
aahung: this was originally here, just reverting

aahung:
just discussed with @mndeveci, for the change that prevents local invoking a function containing unresolvable layer, the risk is unknown and the benefit does not justify the risk. I just reverted that part and use a `LOG.debug` instead.

mndeveci (approved):

PR #2766, opened 2021-03-25
feat: Add CICD platform to metrics

aahung:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/cicd.py
+++ b/samcli/lib/telemetry/cicd.py
@@ -1,0 +1,120 @@
+"""
+Module used for detecting whether SAMCLI is executed in a CI/CD environment.
+"""
+
+import os
+from enum import Enum, auto
+from typing import Mapping, Optional, Dict, Union, Callable
+
+
+class CICDPlatform(Enum):
+    Jenkins = auto()
+    GitLab = auto()
+    GitHubAction = auto()
+    TravisCI = auto()
+    CircleCI = auto()
+    AWSCodeBuild = auto()
+    TeamCity = auto()
+    Bamboo = auto()
+    Buddy = auto()
+    CodeShip = auto()
+    Semaphore = auto()
+    Appveyor = auto()
+    # make sure Unknown is at the bottom, it is the fallback.
+    Unknown = auto()
+
+
+def _is_codeship(environ: Mapping) -> bool:
+    """
+    Use environ to determine whether it is running in CodeShip.
+    According to the doc,
+    https://docs.cloudbees.com/docs/cloudbees-codeship/latest/basic-builds-and-configuration/set-environment-variables
+    > CI_NAME                # Always CodeShip. Ex: codeship
+
+    to handle both "CodeShip" and "codeship," here the string is converted to lower case first.
+
+    Parameters
+    ----------
+    environ
+
+    Returns
+    -------
+    bool
+        whether the env is CodeShip
+    """
+    ci_name: str = environ.get("CI_NAME", "").lower()
+    return ci_name == "codeship"
+
+
+_ENV_VAR_OR_CALLABLE_BY_PLATFORM: Dict[CICDPlatform, Union[str, Callable[[Mapping], bool]]] = {
+    # https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
+    CICDPlatform.Jenkins: "JENKINS_URL",
+    # https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
+    CICDPlatform.GitLab: "GITLAB_CI",
+    # https://docs.github.com/en/actions/reference/environment-variables
+    CICDPlatform.GitHubAction: "GITHUB_ACTION",
+    # https://docs.travis-ci.com/user/environment-variables/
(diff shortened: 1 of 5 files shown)


CoshUS (approved):

samcli/lib/telemetry/cicd.py
```diff
@@ -0,0 +1,120 @@
...
+    # https://docs.github.com/en/actions/reference/environment-variables
+    CICDPlatform.GitHubAction: "GITHUB_ACTION",
+    # https://docs.travis-ci.com/user/environment-variables/
+    CICDPlatform.TravisCI: "TRAVIS",
+    # https://circleci.com/docs/2.0/env-vars/
+    CICDPlatform.CircleCI: "CIRCLECI",
+    # https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+    CICDPlatform.AWSCodeBuild: "CODEBUILD_BUILD_ID",
+    # https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html
+    CICDPlatform.TeamCity: "CODEBUILD_BUILD_ID",
```
wchengru, line 63: Should this also be CODEBUILD_BUILD_ID?

samcli/lib/telemetry/cicd.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # it is a callable, use the return value
+    return env_var_or_callable(environ)
+
+
+# detect CI/CD platform. Here we only calculate this value once so platform() can return this value
+# without re-calculating because CICD platform won't change once sam-cli starts.
+try:
+    _cicd_platform: Optional[CICDPlatform] = next(
+        cicd_platform for cicd_platform in CICDPlatform if _is_cicd_platform(cicd_platform, os.environ)
```
wchengru, line 105: Any order consideration in it? For example, if one env var appears in two platform. It will be fall in the first platform.

samcli/lib/telemetry/cicd.py
```diff
@@ -0,0 +1,120 @@
...
+    CircleCI = auto()
+    AWSCodeBuild = auto()
+    TeamCity = auto()
+    Bamboo = auto()
+    Buddy = auto()
+    CodeShip = auto()
+    Semaphore = auto()
+    Appveyor = auto()
+    # make sure Unknown is at the bottom, it is the fallback.
+    Unknown = auto()
```
hoffa, line 24: Maybe rename to `Other`? `Unknown` sounds like we don't know anything, but we do know it's in CI.

samcli/lib/telemetry/cicd.py
hoffa: And related to this, `dict` order isn't guaranteed on Python 3.6, so we shouldn't depend on it.

> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

samcli/lib/telemetry/metric.py
```diff
@@ -349,15 +351,17 @@ def _get_execution_environment() -> str:
...
+        other CICD platform name    - SAM CLI was executed in CICD
 
         Returns
         -------
         str
             Name of the environment where SAM CLI is executed in.
         """
+        cicd_platform = cicd.platform()
+        if cicd_platform:
+            return cicd_platform.name
```
hoffa, line 364: Where will this show up?

samcli/lib/telemetry/cicd.py
```diff
@@ -0,0 +1,120 @@
...
+    if isinstance(env_var_or_callable, str):
+        return env_var_or_callable in environ
+
+    # it is a callable, use the return value
+    return env_var_or_callable(environ)
+
+
+# detect CI/CD platform. Here we only calculate this value once so platform() can return this value
+# without re-calculating because CICD platform won't change once sam-cli starts.
+try:
```
hoffa, line 103: I'm not sure whether we should have code at the module scope, as this means the code is executed immediately when imported (not the usual expectation). Doesn't seem to be a problem here but could become a footgun later on.

samcli/lib/telemetry/cicd.py
aahung: Good idea, renamed

samcli/lib/telemetry/cicd.py
aahung: yes, the environment variables are very platform-specific if you look at them. Except for the last one `CI`, which is placed at the bottom as a "catch-all"

samcli/lib/telemetry/cicd.py
aahung: nice catch!

samcli/lib/telemetry/cicd.py
aahung: Great idea, I moved them to `Metric`'s `__init__()`

samcli/lib/telemetry/metric.py
aahung: where? it is in `executionEnvironment` metric

wchengru (approved):

hoffa (approved):

PR #2772, opened 2021-03-26
Feat: Added integration tests for new build image option

qingchm:
N.A.

To add integration tests for new option --build-image

N.A.

N.A.

#### Checklist

- [X] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


installer/pyinstaller/build-linux.sh
```diff
@@ -2,6 +2,7 @@
 binary_zip_filename=$1
 python_library_zip_filename=$2
 python_version=$3
+nightly_build=$4
```
hoffa, line 5: Why is this here?

qingchm: The this due to when I pulled from the feature branch, my branch was already ahead of the feature branch. So it included extra commits. Not sure if any good solutions here. Will discuss about this as a part of the release improvements.

PR #2780, opened 2021-03-30
Feat: added integration tests, container tag check and layer support 

qingchm:
N.A.

To add integration tests for new option --build-image

N.A.

N.A.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         docker_client: Optional[docker.DockerClient] = None,
         container_env_var: Optional[Dict] = None,
         container_env_var_file: Optional[str] = None,
-        build_images: Optional[dict] = None,
+        build_images: Optional[Dict] = None,
```
jfuss, line 70: Curious: What is the difference between `dict` and `Dict` types?

samcli/lib/build/app_builder.py
```diff
@@ -441,8 +441,14 @@ def _build_layer(
                     # Only set to this value if specified workflow is makefile
                     # which will result in config language as provided
                     build_runtime = compatible_runtimes[0]
+                image = None
+                if self._build_images is not None:
+                    if layer_name in self._build_images:
+                        image = self._build_images[layer_name]
+                    elif None in self._build_images:
+                        image = self._build_images[None]
                 self._build_function_on_container(
```
jfuss, line 450: I am slightly confused on this. Which I think is mainly due to the nesting. 

A couple questions:
1) Can `self._build_images` be `None`? If so, is there a way we can safely default this to an empty dict instead? 
2) Why are we checking if the `layer_name` is in `self._build_images`?
3) I am not familiar with the `build_images` property and how it gets set. Could you describe the contents?

Something to think about for the future:
When you are writing deeply nested code, consider ways you can simplify it. There are usually ways to write the same logic with less control flows or splitting the nesting into simpler private functions. So in this case, if we could default `self._build_images` to an empty dict in `__init__` then we safe ourselves the need to do `None` checks (we do this throughout the code base for this reason). 

When acting on dictionaries in python, try to avoid using `[]` to access elements and instead use `.get()` method. The `.get()` method does the checks you want for you. So you can simplify both inner if statements by doing `image = self._build_images.get(layer_name)`. By default, this will return `None` if the element (`layer_name`) is not in the dict.

I think I actually answered my above questions while writing the last part but will leave the in the comment to help explain my thought process/questions I would ask.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1837,3 +1838,56 @@ def test_nested_build(self, use_container, cached, parallel):
...
+    ((IS_WINDOWS and RUNNING_ON_CI) and not CI_OVERRIDE),
+    "Skip build tests on windows when running in CI unless overridden",
+)
+class TestBuildWithCustomBuildImage(BuildIntegBase):
+    template = "provided_image_function.yaml"
+
+    @parameterized.expand(
+        [
+            ("use_container", None),
+            ("use_container", "amazon/aws-sam-cli-build-image-nodejs10.x"),
```
jfuss, line 1853: Can we keep the languages our integ tests use the same as others (python I think)? This helps us in keeping things up to date across all our tests.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1837,3 +1838,56 @@ def test_nested_build(self, use_container, cached, parallel):
...
+            process.kill()
+            raise
+        process_stderr = stderr.strip()
+
+        self._verify_right_image_pulled(build_image, process_stderr)
+        self._verify_build_succeeds(self.default_build_dir)
+
+        self.verify_docker_container_cleanedup("nodejs10.x")
+
+    def _verify_right_image_pulled(self, build_image, process_stderr):
```
jfuss, line 1881: I thought we have helper functions that do this already?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1837,3 +1838,56 @@ def test_nested_build(self, use_container, cached, parallel):
...
+        image_name = build_image if build_image is not None else "public.ecr.aws/sam/build-nodejs10.x"
```
jfuss, line 1882: If you really want to stay with node (which I prefer we keep all our integ tests the same language unless really testing something lang specific), we should be using the latest node.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1837,3 +1838,56 @@ def test_nested_build(self, use_container, cached, parallel):
...
+        if use_container and SKIP_DOCKER_TESTS:
+            self.skipTest(SKIP_DOCKER_MESSAGE)
+
+        cmdlist = self.get_command_list(use_container=use_container, build_image=build_image)
+
+        LOG.info("Running Command: {}".format(cmdlist))
+        process = Popen(cmdlist, cwd=self.working_dir, stdout=PIPE, stderr=PIPE)
+        try:
+            stdout, stderr = process.communicate(timeout=TIMEOUT)
+            LOG.info(f"Stdout: {stdout.decode('utf-8')}")
```
jfuss, line 1867: Do we need this info always displayed in our test output?

Same goes for other `LOG.` statements, mainly those who will output logs of data (stderr and stdout).

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -122,7 +122,7 @@ def test_build(self, do_cli_mock):
 
             LOG.debug(Path(config_path).read_text())
             runner = CliRunner()
-            result = runner.invoke(cli, [])
+            result = runner.invoke(cli, ["--use-container"])
```
jfuss, line 125: Why are we adding `--use-container` to all of these? It's unclear why we need to update this.

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
+"""
+Module to check container based cli parameters
+"""
+import click
+
+
+class Container(click.Option):
```
jfuss, line 7: Should we name this something more specific like `ContainerOptions`? We have other Container classes and might be easier to read/understand the difference at a glance.

tests/integration/buildcmd/test_build_cmd.py
qingchm: We do this same logging behaviour for all of our integration tests, when we run the command @jfuss

tests/integration/buildcmd/test_build_cmd.py
qingchm: We have already added the support to default the tags to be latest if we don't input the tag, so this is exactly the same thing. But for the testing we can test if latest tag is included

tests/unit/commands/samconfig/test_samconfig.py
qingchm: Because otherwise as the CliRunner doesn't see this tag, it fails the invocation and showcases the ContainerTag check error

samcli/commands/build/click_container.py
qingchm: Sure I will change the name

samcli/lib/build/app_builder.py
qingchm: "There is no real difference between using a plain typing.Dict and dict, no.
However, typing.Dict is a Generic type * that lets you specify the type of the keys and values too, making it more flexible" e.g. Dict[str, str]
In general we are just using Dict since it has more functionality. As we already imported Dict and Dict is used everywhere else in the class. Changing this for consistency.

samcli/lib/build/app_builder.py
qingchm: 1. Yes it is defaulted to be none. Type hinting doesn't allow any parameter defaults to be {}, so no.
2. self._build_images has function names and layer names as keys and image uri's as value.
3. Details in https://github.com/aws/aws-sam-cli/pull/2725, basically we take the command line argument and make a dict that stores the URI's and their related logicalId's

samcli/lib/build/app_builder.py
aahung: Yes, it is just for consistency

tests/integration/buildcmd/test_build_cmd.py
qingchm: Sure I can change the test case to python then

tests/integration/buildcmd/test_build_cmd.py
qingchm: this is checking the name of the image, I don't think we have anything similar

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -174,7 +174,7 @@ def test_build_with_container_env_vars(self, do_cli_mock):
...
```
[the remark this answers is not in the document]
qingchm: tests without this fails due to missing --use-container flag

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Preprocessing checks for presence of --use-container flag for container based options.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.require_container: bool = kwargs.pop("require_container", None)
```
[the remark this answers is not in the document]
qingchm: Hahaha my bad, typo here

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx, opts, args):
+        if self.require_container:
+            if "use_container" not in opts and opts.get(self.name) is not None:
+                msg = f"""
+Missing required parameter, with --{self.name.replace("_", "-")} set.
+
+Must provide the --use-container flag in order to use --{self.name.replace("_", "-")} flag.
+                """
```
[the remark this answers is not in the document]
qingchm: Okay, this is following the pattern used in Mutex class, can you check that to see which one we should use?

samcli/lib/build/app_builder.py
qingchm: +1 on using get(), this is following the previous pr's pattern, will change both to get()

qingchm: And we can set self._build_images to {} during init, I agree totally.

qingchm:
> The logic seems generally good and I like how you adding the `Container` option class!
> 
> Can we make sure we have unit tests for any code we added? It looks like we don't have tests for the `Container` class or the additional logic you added into the `app_builder`. Pytest should give you lines/files that are missing coverage and you can use that as a starting place to have the lines/files you added are covered.

The unit tests are already included in the previous pr. As for Container class, I don't think we have tests on click_mutex class either. Are we looking to add tests for click related code?

samcli/commands/build/click_container.py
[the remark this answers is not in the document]
aahung: Looks like we have a mix of with new line (3) vs without new line (2 in the mutex file). I would just go with the "slightly" majority, plus error messages do not need to start with a new line anyway.

tests/unit/commands/samconfig/test_samconfig.py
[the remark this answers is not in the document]
aahung: Previously `test_build_with_container_env_vars` can pass, I am a bit confused, does it mean we cannot use `sam` without `--use-container`?

samcli/commands/build/click_container.py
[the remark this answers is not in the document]
qingchm: Sure, we can make them consistent

tests/unit/commands/samconfig/test_samconfig.py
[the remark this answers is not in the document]
qingchm: We didn't add in the Container clx before, but we have this check newly added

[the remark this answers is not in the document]
qingchm: For any container related flags, we will fail the command if the --use container flag is not present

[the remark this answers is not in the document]
aahung: oh I see, so the test before was not correct?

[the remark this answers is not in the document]
qingchm: Before we allowed specifying container specific options without --use-container flag, we just ignore them. Now we decided to fail them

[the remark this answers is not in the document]
aahung: got it, thx

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx, opts, args):
+        if self.require_container:
+            if "use_container" not in opts and opts.get(self.name) is not None:
+                msg = f"""\
+Missing required parameter, with --{self.name.replace("_", "-")} set.
+
+Must provide the --use-container flag in order to use --{self.name.replace("_", "-")} flag.\
+                """
```
aahung, line 24: My previous suggestion still left a few space behind

```suggestion
Must provide the --use-container flag in order to use --{self.name.replace("_", "-")} flag."""
```

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+                raise click.UsageError(msg)
+            self.prompt = None
```
aahung, line 26: what does this `self.prompt = None` do?

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+import click
+
+
+class ContainerOptions(click.Option):
+    """
+    Preprocessing checks for presence of --use-container flag for container based options.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.require_container: bool = kwargs.pop("require_container", False)
```
aahung, line 13: I am not sure why we need this `require_container`, as far as I can search from this PR change, `require_container` is always `True` when `ContainerOptions` is used. If we remove it, would it still work?

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+
+        super().__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx, opts, args):
+        if self.require_container:
+            if "use_container" not in opts and opts.get(self.name) is not None:
```
aahung, line 19: we are using two forms of "check if X is in opts"
- `X in opts`
- `opts.get(X) is not None`

Is there a reason?

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,27 @@
...
+                msg = f"""\
+Missing required parameter, with --{self.name.replace("_", "-")} set.
```
aahung, line 21: This seems redundant, the next sentence basically says the same thing.

samcli/commands/build/click_container.py
qingchm: in the future if we get more complex with conditions and dependencies we can continue using this class adding other variables, that's the main reason why we design it this way

samcli/commands/build/click_container.py
qingchm: There is a possibility that it is in opts but it is None, that's what our mock test does. In practice this doesn't occur. But they do check different things

samcli/commands/build/click_container.py
qingchm: this is telling which parameter requires the -u flag, e.g., --build-image

samcli/commands/build/click_container.py
qingchm: this is an attribute inherited from click.Options

samcli/commands/build/click_container.py
qingchm: actually it kind of makes sense to remove this completely, for additional logic they would still require container as a prereq

tests/integration/buildcmd/test_build_cmd.py
jfuss: I am not talking about the image tag here. You are using Nodejs10.x, which is the oldest of the 2 Node versions Lambda supports. My comment is to not use the oldest and instead us the latest version of Node (or any language we us in our integ tests).

tests/integration/buildcmd/test_build_cmd.py
jfuss: In other build tests we do not log stdout and stderr. This comment is about removing the large outputs in our tests logs. Leaving what commands and other small information is fine, logging 30+ lines seems like overkill. If you think this info is important, then we should only output it on test failure instead of every single run.

tests/unit/commands/samconfig/test_samconfig.py
jfuss: I might be missing something but my initial feeling is that seems wrong. Wouldn't this mean we are requiring `--use-container` now? The command shouldn't fail if we don't provided `--use-container` to the cli.

samcli/lib/build/app_builder.py
```diff
@@ -441,8 +441,13 @@ def _build_layer(
                     # Only set to this value if specified workflow is makefile
                     # which will result in config language as provided
                     build_runtime = compatible_runtimes[0]
+                image = None
+                if layer_name in self._build_images:
+                    image = self._build_images.get(layer_name)
+                elif None in self._build_images:
```
jfuss, line 447: What does `None` represent in `self._build_images`? This seems like an odd line to me. Is `None` some signal meaning the default image to use? Is so, we would change this to improve the readability. It is super unclear why we would do a look up in a dictionary with a key of `None`.

tests/integration/buildcmd/test_build_cmd.py
jfuss: I guess we do but it is done through the `run_command`: https://github.com/aws/aws-sam-cli/blob/4543732bf3c0da3b57fe1e5aa43ce3f41d2bd0ba/tests/testing_utils.py#L25 Why are we not using this here?

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,23 @@
+"""
+Module to check container based cli parameters
+"""
+import click
+
+
+class ContainerOptions(click.Option):
```
jfuss, line 7: Do we have unit tests for this class?

jfuss (changes requested):

samcli/lib/build/app_builder.py
qingchm: This was discussed in the previous PR, #2725, None key represents the global image for all functions/layers. How do you suggest to change it? Involving @weichw into this discussion too

tests/unit/commands/samconfig/test_samconfig.py
qingchm: It should now, we are going to fail any use of container related tags if we don't provide --use-container to the cli instead of ignoring them now

tests/integration/buildcmd/test_build_cmd.py
qingchm: because we need to do extra work that run_command does not provide, which is checking the content in the stderr

samcli/commands/build/click_container.py
qingchm: We don't, I don't see any tests related to https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/local/cli_common/click_mutex.py either, let me know your thoughts regarding testing on click.Option classes! One of the ways that I can think of testing it would be to create a mock CliRunner and verify that command fails when we provide container related tags without -u

samcli/commands/build/click_container.py
aahung: why we set prompt to `None` here, the intention is not clear here, maybe a comment would help.

samcli/commands/build/click_container.py
aahung: 👍 thx for addressing

samcli/commands/build/click_container.py
aahung: If in practice it never occurs, our unit test should not test it that way then. This inconsistency makes me confused. I am not sure if it is within the scope of this PR, if will be nice if we can unify them

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Preprocessing checks for presence of --use-container flag for container based options.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
aahung, line 13: This constructor does not seem to do anything, I think we can remove it

samcli/commands/build/click_container.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def handle_parse_result(self, ctx, opts, args):
+        if "use_container" not in opts and opts.get(self.name) is not None:
+            msg = f"""\
+Missing required parameter, with --{self.name.replace("_", "-")} set.
+
+Must provide the --use-container flag in order to use --{self.name.replace("_", "-")} flag."""
+            raise click.UsageError(msg)
+        self.prompt = None
```
aahung, line 22: worth to explain this line since its purpose is not straightforward

samcli/lib/build/app_builder.py
aahung: this makes sense, but I would suggest to add a comment here explaining this because `None` representing the global image is quite implicit

samcli/lib/build/app_builder.py
```diff
@@ -521,11 +526,10 @@ def _build_function(  # pylint: disable=R1710
...
                     image = None
-                    if self._build_images is not None:
-                        if function_name in self._build_images:
-                            image = self._build_images[function_name]
-                        elif None in self._build_images:
-                            image = self._build_images[None]
+                    if function_name in self._build_images:
+                        image = self._build_images.get(function_name)
+                    elif None in self._build_images:
+                        image = self._build_images.get(None)
```
aahung, line 532: the entire code block (including the `image = None`) can be simplified as

```py
image = self._build_images.get(function_name, self._build_images.get(None))
```

samcli/lib/build/app_builder.py
aahung: the entire code block (including the `image = None`) can be simplified as

```py
image = self._build_images.get(function_name, self._build_images.get(None))
```

samcli/commands/build/click_container.py
qingchm: Because click options have a prompt field which can be set to generate user input prompts, this is just to make sure no prompting happens

samcli/lib/build/app_builder.py
qingchm: I feel that this logic here is simple and I don't think the code is over-complicated, I would argue for keeping it this way? For better readability

samcli/lib/build/app_builder.py
qingchm: I would still argue for the original logic, a get inside a get doesn't seem as straight forward

samcli/commands/build/click_container.py
aahung: exactly, it totally make sense after you explain it. It is not clear just by looking at the code, that's why I think a comment here would greatly help.

samcli/lib/build/app_builder.py
aahung: I have an alternative for you then:
```py
global_image = self._build_images.get(None)
image = self._build_images.get(layer_name, global_image)
```

samcli/lib/build/app_builder.py
aahung: duplicate from the one above, resolving

samcli/lib/build/app_builder.py
qingchm: updated

samcli/lib/build/app_builder.py
```diff
@@ -441,8 +441,12 @@ def _build_layer(
                     # Only set to this value if specified workflow is makefile
                     # which will result in config language as provided
                     build_runtime = compatible_runtimes[0]
+                image = None
```
aahung, line 444: this line is not needed since it will be immediately overridden at line 447

tests/integration/buildcmd/test_build_cmd.py
aahung: We do support checking `stderr` using `run_command`

https://github.com/aws/aws-sam-cli/blob/4543732bf3c0da3b57fe1e5aa43ce3f41d2bd0ba/tests/integration/buildcmd/test_build_cmd.py#L1548-L1551

samcli/lib/build/app_builder.py
```diff
@@ -521,11 +525,9 @@ def _build_function(  # pylint: disable=R1710
                 # By default prefer to build in-process for speed
                 if self._container_manager:
                     image = None
```
aahung, line 527: same, this line is not needed

samcli/commands/build/click_container.py
aahung: I am not sure I understand, when this error message is printed, it looks like this:
```
Missing required parameter, with --build-image set.
Must provide the --use-container flag in order to use --build-image flag.
```

I think the 2nd sentence contains all the information in the first one, what do you think?

aahung (approved):
Overall LGTM, left some small comments/questions

samcli/lib/build/app_builder.py
aahung: cool, thx

tests/integration/buildcmd/test_build_cmd.py
aahung: +1, this can save us some work (or delay the work) when we deprecate node 10

tests/integration/buildcmd/test_build_cmd.py
qingchm: I can try something like this, trying to use the nametuple returned

qingchm: Before I had several differences between this function and run command, but seems like this run_command will return what I need

qingchm: Updated to use run_command()

samcli/commands/build/click_container.py
qingchm: Sounds fair, I'll combine them

samcli/lib/build/app_builder.py
jfuss: > This was discussed in the previous PR, #2725, None key represents the global image for all functions/layers. How do you suggest to change it? Involving @CoshUS into this discussion too

I don't really know what has been discussed in previous prs. I am just looking at what is in this review and what we are adding/changing.

My suggestion is to make this more readable through being explicit. Getting `None` from a dictionary is very strange to me, especially because I assumed this is a `str:str` dictionary instead of mixing types. Someone should be able to understand the code by looking/read the function/class. Jumping between files to see where things were set makes it very difficult to follow code. In this case, I assumed we were getting the build image that corresponds to a specific layer or function. `None` in this case, seems "random" and doesn't indicate what/why we are getting this key. 

Ways to solve this and improve the readability:
1. pass a default image to use if one is not within `self._build_images`
2. Create `self._build_images` in a way that is consistent. So every function/layer is a key in `self._build_image` and whoever is creating this dictionary is responsible for setting it up that way.
3. Abstract this away into some class or function that hides the detail. So instead of interacting with the dictionary, we can create an indirection through a method that anyone can call. This moves the `None` logic into one place and provides a much more readable solution.

If you stay with the accessing of `None` (which I recommend we don't), you should be adding a comment that clearly explains why we are accessing `None` in a `build_images` dictionary.

samcli/commands/build/click_container.py
jfuss: Just because we do not have tests for another class, doesn't mean we should not be adding tests for new code. We always want to improve the code base through adding tests. With that logic one could say, we shouldn't be adding any tests because we have a class that doesn't have tests at all.

Please add unit tests for code we add, so we have some basic checks in place on that expectation of that code.

We shouldn't need to mock a CliRunner here or verify commands (kind of out of scope of a unit tests in my opinion). What we want to verify is that the logic you added is correct. So you can mock out methods, input, etc to help in validating what you expect to happen does happen. At the very least we should have tests to cover:

```
if "use_container" not in opts and self.name in opts:
            opt_name = self.name.replace("_", "-")
            msg = f"Missing required parameter, need the --use-container flag in order to use --{opt_name} flag."
            raise click.UsageError(msg)
```
An alternative way to do this is by creating a sub method and testing that, if mocking the rest of things becomes a pain.

qingchm: adding two branches of tests

samcli/lib/build/app_builder.py
qingchm: We have a default value for build image, the None key's value is a customer override of the default one, while the function/layer specific key value is the override of the None key one. This structure will not change. Adding a comment to explain

tests/unit/commands/buildcmd/test_command.py
```diff
@@ -284,3 +285,29 @@ def test_two_functions_with_default_image(self):
             ["Function1=image1", "Function2=image2", "image3"],
             {"Function1": "image1", "Function2": "image2", None: "image3"},
         )
+
+
+@patch("samcli.commands.build.click_container.ContainerOptions")
+class TestContainerOptionsSucceeds(TestCase):
```
jfuss, line 291: Nit: Since this is in its own file, lets match that with our unit tests as well. This is will make it consistent with all our other tests in the repo and make it easier to know where to update tests in the future when things change.

qingchm: Sounds good

jfuss (approved):
Thanks for handling all the feedback through this PR.

Issue #2779, opened 2021-03-30
sam deploy Issue - S3 TemplateURLs transformed to local file path


kirstenkissmeyer:
This issue occurs when running sam deploy.

There is a very similar issue that was fixed approx a year ago when running sam build:
https://github.com/aws/aws-sam-cli/issues/1839

I did place a comment on that closed issue

### Description:

sam deploy transforms TemplateURL from S3 https link to local file path. It prepends the path from which the sam deploy command is run to the https link. Then, of course, it says the resulting TemplateURL is invalid.

 The example provided is a basic CFN template to show the output but have seen this on full SAM templates with SAM resources as well.

### Steps to reproduce:

---
AWSTemplateFormatVersion: '2010-09-09'
Transform: myTransform
Description: A test CFN template

Resources:
  ExampleNestedStack:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://mybucket.s3.amazonaws.com/20210326_222039_m2/templates/Template/exampletemplate.y\
aml

* note that mybucket is not actual name of the bucket 

sam deploy --role-arn my-admin-role --capabilities CAPABILITY_IAM --template-file Template/sample-template.yaml --stack-name my-sample-stack

### Observed result:

Error: Unable to upload artifact https://mybucket.s3.amazonaws.com/20210326_222039_m2/templates/Template/exampletemplate.yaml referenced by TemplateURL parameter of ExampleNestedStack resource.
TemplateURL parameter of ExampleNestedStack resource is invalid. It must be a S3 URL or path to CloudFormation template file. \
Actual: /home/ec2-user/project/myproj/Template/https:/mybucket.s3.amazonaws.com/20210326_222039_m2/templates/Template/exampletemplate.yaml

### Expected result:

successful interpretation and deploy processing of TemplateURL as the provided https path, with no prepending:
https:/mybucket.s3.amazonaws.com/20210326_222039_m2/templates/Template/exampletemplate.yaml

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Amazon Linux 2 on t3.small instance
2. `sam --version`:   1.21.1
3. AWS region: us-west-2

`Add --debug flag to command you are running`

sam deploy --debug --role-arn my-admin-role --capabilities CAPABILITY_IAM --template-file Template/sample-template.yaml --stack-name my-sample-stack

[...]

PR #2785, opened 2021-04-01
fix: package/deploy failure when Location/TemplateURL is virtual host S3 URL

aahung:

#2779

The `CloudFormationStackResource` only check a URL against a limited s3 url format (to decide what artifact not to upload), causing package/deploy to fail.

Add a new `is_s3_url()` function to check all kinds of s3 urls and skip the upload when detecting so.

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/package/artifact_exporter.py
+++ b/samcli/lib/package/artifact_exporter.py
@@ -36,7 +36,7 @@ from samcli.lib.package.packageable_resources import (
 )
 from samcli.lib.package.s3_uploader import S3Uploader
 from samcli.lib.package.uploaders import Uploaders
-from samcli.lib.package.utils import is_local_folder, make_abs_path, is_s3_url, is_local_file, mktempfile
+from samcli.lib.package.utils import is_local_folder, make_abs_path, is_s3_protocol_url, is_local_file, mktempfile
 from samcli.lib.utils.packagetype import ZIP
 from samcli.yamlhelper import yaml_parse, yaml_dump
 
@@ -64,7 +64,7 @@ class CloudFormationStackResource(ResourceZip):
 
         if (
             template_path is None
-            or is_s3_url(template_path)
+            or is_s3_protocol_url(template_path)
             or template_path.startswith(self.uploader.s3.meta.endpoint_url)
             or template_path.startswith("https://s3.amazonaws.com/")
         ):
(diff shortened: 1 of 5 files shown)


samcli/lib/package/artifact_exporter.py
```diff
@@ -62,12 +69,7 @@ def do_export(self, resource_id, resource_dict, parent_dir):
 
         template_path = resource_dict.get(self.PROPERTY_NAME, None)
 
-        if (
-            template_path is None
-            or is_s3_url(template_path)
-            or template_path.startswith(self.uploader.s3.meta.endpoint_url)
```
aahung, line 68: explain: `self.uploader.s3.meta.endpoint_url` looks like `https://s3.us-west-2.amazonaws.com`. I think originally @sriram-mv 's intention is to make sure the url is in the same region. (@sriram-mv correct me if wrong) I think we don't need to handle it here because if the url lacks region info (like `s3://bucket-name/key`), we let it pass through. So we can have a consistent behavior here.

samcli/lib/package/utils.py
```diff
@@ -33,14 +34,49 @@ def make_abs_path(directory, path):
...
+    """
     try:
         S3Uploader.parse_s3_url(url)
         return True
     except ValueError:
         return False
 
 
+# https://docs.aws.amazon.com/AmazonS3/latest/dev-retired/UsingBucket.html
+_REGION_PATTERN = r"[a-zA-Z0-9-]+"
```
sriram-mv, line 49: Lets move these to the top of file, if they are constants, after all the imports.

sriram-mv (changes requested):

samcli/lib/package/utils.py
aahung: updated

sriram-mv (approved):

ssenchenko (approved):

samcli/lib/package/utils.py
```diff
@@ -22,6 +23,29 @@
 
 LOG = logging.getLogger(__name__)
 
+# https://docs.aws.amazon.com/AmazonS3/latest/dev-retired/UsingBucket.html
+_REGION_PATTERN = r"[a-zA-Z0-9-]+"
+_DOT_AMAZONAWS_COM_PATTERN = r"\.amazonaws\.com"
```
narasimman-elu, line 28: Will this work for Chinese domain?

aahung: Good catch, #3001

aahung: #3001 merged

PR #2799, opened 2021-04-09
fix: Cookiecutter template generation failed when related dir exists

aahung:

The output directory might exist already (very possible for pipeline generation, e.g., adding a new gh action workflow when `.github/workflows` contains other workflow files), not having `overwrite_if_exists` will cause the `cookiecutter()` to fail

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/cookiecutter/template.py
+++ b/samcli/lib/cookiecutter/template.py
@@ -144,7 +144,9 @@ class Template:
 
         try:
             LOG.debug("Baking a new template with cookiecutter with all parameters")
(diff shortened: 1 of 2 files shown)


elbayaaa:
@aahung I'm quite confused why we are pushing this, we agreed that the correct behavior is to fail if the directory already exist, right?

aahung:
Yeah that's why I closed this PR @elbayaaa

PR #2805, opened 2021-04-13
chore: Use BUILD_TAG instead of JENKINS_URL to identify Jenkins env

aahung:

According to the doc https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#working-with-your-jenkinsfile

`JENKINS_URL` is only available if Jenkins URL set in "System Configuration." So here we use `BUILD_TAG` to identify Jenkins instead to ensure no false negatives

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/cicd.py
+++ b/samcli/lib/telemetry/cicd.py
@@ -53,8 +53,14 @@ def _is_jenkins(environ: Mapping) -> bool:
     https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#working-with-your-jenkinsfile
     > BUILD_TAG
     >   String of jenkins-${JOB_NAME}-${BUILD_NUMBER}.
+    > ...
+    > JENKINS_URL
+    >   Full URL of Jenkins, such as https://example.com:port/jenkins/
+    >   (NOTE: only available if Jenkins URL set in "System Configuration")
+
+    Here firstly check JENKINS_URL's presence, if not, then fallback to check BUILD_TAG starts with "jenkins"
     """
-    return bool(environ.get("BUILD_TAG", "").startswith("jenkins-"))
+    return "JENKINS_URL" in environ or environ.get("BUILD_TAG", "").startswith("jenkins-")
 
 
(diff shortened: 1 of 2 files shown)


samcli/lib/telemetry/cicd.py
```diff
@@ -46,9 +46,19 @@ def _is_codeship(environ: Mapping) -> bool:
...
+    > BUILD_TAG
+    >   String of jenkins-${JOB_NAME}-${BUILD_NUMBER}.
+    """
+    return bool(environ.get("BUILD_TAG", "").startswith("jenkins-"))
+
+
 _ENV_VAR_OR_CALLABLE_BY_PLATFORM: Dict[CICDPlatform, Union[str, Callable[[Mapping], bool]]] = {
-    # https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
-    CICDPlatform.Jenkins: "JENKINS_URL",
+    CICDPlatform.Jenkins: _is_jenkins,
```
sriram-mv, line 61: should we do an or condition here? or is build_tag always true and present with the jenkins prefix?

sriram-mv (changes requested):

samcli/lib/telemetry/cicd.py
aahung: According to the doc, it does not have a condition. 
I tested it in my own jenkins instance and can see the variable value did starts with `jenkins-`

sriram-mv: ok, but should we look for JENKINS_URL or BUILD_TAG with jenkins-, is there a condition under which either one may be present or not, we would want to include both.

aahung: `BUILD_TAG` itself should suffice, but I don't mind adding a `JENKINS_URL` since it works today already, having it is safer

aahung: added

sriram-mv (approved):

hoffa (approved):

PR #2806, opened 2021-04-13
feat: add option --container-host and --container-host-interface to sam local commands

xazhao:
#2492 #2436

We found there is an issue of warm container with #2700 before release and remote Docker host is not support in that PR. So we decided to revert that change.
This PR includes following changes:

- #2700 which was reverted last week
- fix warm container issue when using `--container-host` option
- add remote Docker support by using a new option `--container-host-interface`. Customers can specify what interface they want to bind ports to with this option.

For Docker API call, we was binding ports to `127.0.0.1` which makes it can not be accessed from remote. With the new change, customers can bind ports to a specific host. In this way, using remote Docker host will be available.
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -58,6 +58,14 @@ def local_common_options(f):
             "For example, if you want to run SAM CLI in a Docker container on macOS, "
             "use this option with host.docker.internal",
         ),
+        click.option(
+            "--container-host-interface",
+            default="127.0.0.1",
+            show_default=True,
+            help="Interface that Docker host binds ports to. "
+            "This option is useful when users want to bind ports to a different interface. "
+            "Note: the input should be an IP address. ",
+        ),
     ]
 
     # Reverse the list to maintain ordering of options in help text printed with --help
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -63,7 +63,7 @@ def local_common_options(f):
             default="127.0.0.1",
             show_default=True,
             help="IP address of the host network interface that container ports should bind to. "
-            "Use 0.0.0.0 to bind to all interfaces."
+            "Use 0.0.0.0 to bind to all interfaces.",
         ),
     ]
 
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -62,11 +62,8 @@ def local_common_options(f):
             "--container-host-interface",
             default="127.0.0.1",
             show_default=True,
-            help="Interface that Docker host binds ports to. "
-            "This option is useful when users want to bind ports to a different interface. "
-            "It will bind ports to loopback interface by default. "
-            "For example, using 0.0.0.0 will bind ports to all interfaces. "
-            "Note: the input should be an IP address. ",
+            help="IP address of the host network interface that container ports should bind to. "
+            "Use 0.0.0.0 to bind to all interfaces."
         ),
     ]
 
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -48,7 +48,16 @@ def local_common_options(f):
             default=False,
             help="If set, will emulate a shutdown event after the invoke completes, "
             "in order to test extension handling of shutdown behavior.",
-        )
+        ),
+        click.option(
+            "--container-host",
+            default="localhost",
+            show_default=True,
+            help="Host of locally emulated Lambda container. "
+            "This option is useful when the container runs on a different host than SAM CLI. "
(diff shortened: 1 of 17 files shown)


samcli/local/docker/container.py
```diff
@@ -150,11 +155,14 @@ def create(self):
...
             kwargs["environment"] = self._env_vars
 
-        kwargs["ports"] = {self.RAPID_PORT_CONTAINER: ("127.0.0.1", self.rapid_port_host)}
+        kwargs["ports"] = {self.RAPID_PORT_CONTAINER: (self._container_host, self.rapid_port_host)}
 
         if self._exposed_ports:
             kwargs["ports"].update(
-                {container_port: ("127.0.0.1", host_port) for container_port, host_port in self._exposed_ports.items()}
+                {
+                    container_port: (self._container_host, host_port)
```
hoffa, line 163: Should we resolve the (potential) domain name before? I can't see domain names being mentioned in the [docs](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run).

xazhao: Yes we should resolve the domain name before passing into ports binding. Trying to figure out what is the right ip it should bind to when running in a container. Resolved `host.docker.internal` doesn't work in this case.

xazhao: Haven't figured out why resolved `host.docker.internal` doesn't work here. An idea just came to my mind. Whenever we use local Docker host, it should bind to `127.0.0.1` while the host name could be `localhost`, `host.docker.internal` etc. When using a remote Docker host, we just use the ip of Docker host. So maybe we should use 2 options?

sriram-mv: Overall, this looks good, what are the scenarios we have tested?

xazhao: Tested 3 scenarios and all work well:
1. Run `sam local` in local
2. Run `sam local` in a container with Docker socket binding.
3. Run `sam local` in a container with Docker running in the other container.

However I had a talk with Chris and we are thinking to add another option `--container-host-interface`. Because docker port binding can only use ip address. And the interface and container host don't match sometimes.(For example, when we use `host.docker.internal` as  container host, we would like bind ports to `127.0.0.1`) So ideally we need 2 options to cover all use case.

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,24 @@ def local_common_options(f):
...
+            "For example, if you want to run SAM CLI in a Docker container on macOS, "
+            "use this option with host.docker.internal",
+        ),
+        click.option(
+            "--container-host-interface",
+            default="127.0.0.1",
+            show_default=True,
+            help="Interface that Docker host binds ports to. "
+            "This option is useful when users want to bind ports to a different interface. "
+            "Note: the input should be an IP address. ",
```
hoffa, line 67: Also mention that `0.0.0.0` binds to all interfaces?

xazhao: Sure!

samcli/commands/local/cli_common/options.py
```diff
@@ -48,7 +48,26 @@ def local_common_options(f):
...
+        ),
+        click.option(
+            "--container-host-interface",
+            default="127.0.0.1",
+            show_default=True,
+            help="Interface that Docker host binds ports to. "
+            "This option is useful when users want to bind ports to a different interface. "
+            "It will bind ports to loopback interface by default. "
+            "For example, using 0.0.0.0 will bind ports to all interfaces. "
+            "Note: the input should be an IP address. ",
```
hoffa, line 69: More succinct?
```suggestion
            help="IP address of the host network interface that container ports should bind to. "
            "Use 0.0.0.0 to bind to all interfaces."
```

sriram-mv: should we add a differentiation in the help between the two options?

sriram-mv (approved):
I added a comment, to showcase the help text which differentiates between --container-host and --container-host-interface.

both --container--host and --container--host-interface are new options which have defaults baked in, which are the same what is was before, So I dont see any regression concerns. @xazhao is my understanding correct?

xazhao:
@sriram-mv Right! Default values will be same as what they were before. Customers will have the same experience if they don't use these 2 options.

samcli/commands/local/cli_common/options.py
xazhao: I think the name and help text are enough to explain the difference between the two options. And from comments of #2492, customers who needs to use these 2 options know clearly the difference. For now, adding additional differentiation seems a little verbose for me.

hoffa (approved):

Issue #1713, opened 2020-01-05
Feature Request: CI / CD pipelines (sam pipeline init)


a-h:
I wish that SAM made it easy to setup a CI /CD deployment pipeline with AWS tools such as a CodeBuild so I don't have to do it myself.

It would be especially useful if it took the tedium out of setting up the cross-account deployments that are commonly used to segregate access to development and test environments from production access.

It would be in the form of a CLI wizard that asks for key information as per the new AWS ECS CLI (https://github.com/aws/amazon-ecs-cli-v2). It has an `ecs pipeline init` feature which seems to be in the sprit of what I mean (I haven't played around with it much).

* Set up a CodeBuild job in a build account that builds the master branch when it changes (`make build` as per the quickstart template, but also `make test`), packages it up and publishes it to an S3 artifact bucket.
* Setup CodeDeploy jobs in each environment (e.g. dev, uat, prod) to deploy specific versions of the SAM package from the bucket.
* Set up a CodePipeline job to be able to deploy the package across environments. Ideally, the first environment would be automated, with subsequent environments needing to be triggered manually, or manually approved.

### Proposal

* Add a `sam pipeline init` command that starts a wizard to create the required CloudFormation templates.
* Add a `sam pipeline deploy` command that deploys the various CodeBuild, CodeDeploy and CodePipeline things.

It shouldn't mean any differences to the SAM specification, just improved developer experience.

### Additional Details

Perhaps this subcommand could be written in Go as per the new ECS CLI to reuse some of the code.

a-h:
It looks like it could be a guided walkthrough of https://aws.amazon.com/quickstart/architecture/serverless-cicd-for-enterprise/

bigkraig:
You may want to take a look at [Projen](https://github.com/projen/projen) and see how close it is to supporting SAM. I don't think its an official AWS project but it has AWS engineers all over it and it's got a channel on their CDK Slack.

hoffa:
Closing since this has been released as [AWS SAM Pipelines](https://aws.amazon.com/blogs/compute/introducing-aws-sam-pipelines-automatically-generate-deployment-pipelines-for-serverless-applications):

- [`sam pipeline bootstrap`](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-pipeline-bootstrap.html)
[...]

PR #2811, opened 2021-04-14
sam pipeline bootstrap

elbayaaa:
#1713 

Creating "sam pipeline bootstrap" which creates the required pipeline infrastructure resources for a given stage

Executes a CFN template with the required resources

N/A it is merging to non-develop branch
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/cli/command.py
```diff
@@ -21,6 +21,7 @@
     "samcli.commands.deploy",
     "samcli.commands.logs",
     "samcli.commands.publish",
+    "samcli.commands.pipeline.pipeline",
```
hoffa, line 24: Can we use [`hidden=True`](https://click.palletsprojects.com/en/7.x/api/?highlight=hidden#click.Option) instead so the command can be tested?

c2tarun:
No integration tests?

samcli/cli/command.py
elbayaaa: Are you suggesting releasing the command but not showing it during `sam --help` ?

elbayaaa:
> No integration tests?

Waiting for `sam pipeline init` to be ready so as to have a e2e integration testing

samcli/cli/command.py
hoffa: Right, that way we can test the feature without changing code.

c2tarun:
> > No integration tests?
> 
> Waiting for `sam pipeline init` to be ready so as to have a e2e integration testing

This is bootstrap right? This is not coupled with `pipeline init` command. Can't we integration test this separately?

samcli/cli/command.py
c2tarun: I like this, until we have `pipeline init` ready it makes sense to keep this hidden so that we don't block any releases until we are fully ready.

elbayaaa:
> > > No integration tests?
> > 
> > 
> > Waiting for `sam pipeline init` to be ready so as to have a e2e integration testing
> 
> This is bootstrap right? This is not coupled with `pipeline init` command. Can't we integration test this separately?

Yeah, that is right, they are technically not coupled with each other, but from usage perspective, they usually are.

samcli/cli/command.py
elbayaaa: This PR is merging to non-develop branch, it won't be released. Having it released and hidden won't prevent anyone from using it which what we need to avoid at the moment as this is still in develop.

c2tarun:
> > > > No integration tests?
> > > 
> > > 
> > > Waiting for `sam pipeline init` to be ready so as to have a e2e integration testing
> > 
> > 
> > This is bootstrap right? This is not coupled with `pipeline init` command. Can't we integration test this separately?
> 
> Yeah, that is right, they are technically not coupled with each other, but from usage perspective, they usually are.

Sorry I'll push back on this. In current integration tests for aws-sam-cli user is supposed to run build before deploy or build before local invoke almost every time. But we have separate integration tests for build, deploy and local-invoke.
1. Build integration tests: https://github.com/aws/aws-sam-cli/tree/develop/tests/integration/buildcmd
2. Deploy integration tests: https://github.com/aws/aws-sam-cli/tree/develop/tests/integration/deploy
3. Local invoke integration tests: https://github.com/aws/aws-sam-cli/tree/develop/tests/integration/local/invoke

Plus there is a possibility that after setting up with our initial template user will only run bootstrap for further stages and not run `pipeline init` again. Without dedicated integration tests we'll miss issues impacting those users.

samcli/cli/command.py
hoffa: Why not keep in a non-develop branch and the command hidden if needed?

I want to be able to test this code, and it's trivial if I don't need to make code changes.

hoffa: Oh never mind, synced offline. I thought it wasn't enabled. I'm fine with either in that case.

mypy.ini
```diff
@@ -59,6 +59,6 @@ ignore_missing_imports=True
 ignore_missing_imports=True
 
 # progressive add typechecks and these modules already complete the process, let's keep them clean
-[mypy-samcli.commands.build,samcli.lib.build.*,samcli.commands.local.cli_common.invoke_context,samcli.commands.local.lib.local_lambda,samcli.lib.providers.*]
+[mypy-samcli.commands.build,samcli.lib.build.*,samcli.commands.local.cli_common.invoke_context,samcli.commands.local.lib.local_lambda,samcli.lib.providers.*,samcli.lib.cookiecutter.*,samcli.lib.pipeline.*,samcli.commands.pipeline.*]
```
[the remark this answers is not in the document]
elbayaaa: these are the enabled ones.

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,274 @@
...
+        missing_resources: str = self._get_non_user_provided_resources_msg()
+        click.echo(f"This will create the following required resources for the {self.name} stage: {missing_resources}")
+        if confirm_changeset:
+            confirmed: bool = click.confirm("Should we proceed with the creation?")
+            if not confirmed:
+                return
+
+        stage_name_suffix: str = re.sub("[^0-9a-zA-Z]+", "-", self.name)
+        stack_name: str = f"{STACK_NAME_PREFIX}-{stage_name_suffix}-{STAGE_RESOURCES_STACK_NAME_SUFFIX}"
+        stage_resources_template_body = Stage._read_template("stage_resources.yaml")
```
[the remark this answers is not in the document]
elbayaaa: no, any thoughts?

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
+""" Represents AWS resource"""
+from typing import List, Optional
+
+
+class ARNParts:
```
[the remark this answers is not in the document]
elbayaaa: do you mean add link to https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html ??

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,274 @@
...
+            click.secho(
+                "\nYou provided the following resources. Please make sure it has the required permissions as shown at "
+                "https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/pipeline/bootstrap/stage_resources.yaml",
+                fg="green",
+            )
+            for resource in provided_resources:
+                click.secho(f"\t{resource.arn}", fg="green")
+
+        if not self.pipeline_user.is_user_provided:
+            click.secho("Please configure your CICD project with the following pipeline-user credentials:", fg="green")
```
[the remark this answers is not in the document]
elbayaaa: I chose green for outputs and important messages.

Here, it is important to tell the user to share the credentials with the CI/CD provider.
There, I used the green for showing the ARNs of the resources we've created

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
hoffa: Perfect, yes exactly.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+from samcli.lib.config.samconfig import SamConfig
+from samcli.lib.pipeline.bootstrap.stage import Stage
+from samcli.lib.telemetry.metric import track_command
+from samcli.lib.utils.version_checker import check_newer_version
+from .guided_context import GuidedContext
+
+SHORT_HELP = "Sets up infrastructure resources for AWS SAM CI/CD pipelines."
+
+HELP_TEXT = """Sets up the following infrastructure resources for AWS SAM CI/CD pipelines:
+\n\t - Pipeline user(IAM User) with AccessKeyId and SecretAccessKey credentials to be shared with the CI/CD provider
```
aahung, line 21:
```suggestion
\n\t - Pipeline user (IAM User) with AccessKeyId and SecretAccessKey credentials to be shared with the CI/CD provider
```

aahung, line 21: writing: we should use "access key ID" and "secret access key" here instead of variable-like names.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+from samcli.cli.main import pass_context, common_options, aws_creds_options, print_cmdline_args
+from samcli.lib.config.samconfig import SamConfig
+from samcli.lib.pipeline.bootstrap.stage import Stage
+from samcli.lib.telemetry.metric import track_command
+from samcli.lib.utils.version_checker import check_newer_version
+from .guided_context import GuidedContext
+
+SHORT_HELP = "Sets up infrastructure resources for AWS SAM CI/CD pipelines."
+
+HELP_TEXT = """Sets up the following infrastructure resources for AWS SAM CI/CD pipelines:
```
aahung, line 20: image

Looks like there is one empty line between each items, is it intended? It feel too "sparse"

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+\n\t - Pipeline user(IAM User) with AccessKeyId and SecretAccessKey credentials to be shared with the CI/CD provider
+\n\t - Pipeline Execution Role(IAM Role) that is assumed by the Pipeline user to obtain access to the AWS account
```
aahung, line 22:
```suggestion
\n\t - Pipeline Execution Role (IAM Role) that is assumed by the Pipeline user to obtain access to the AWS account
```

aahung, line 22: to be consistent with the first item "to be shared with the CI/CD provider"

```suggestion
\n\t - Pipeline Execution Role(IAM Role) to be assumed by the Pipeline user to obtain access to the AWS account
```

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+\n\t - CloudFormation Execution Role(IAM Role) that is assumed by CloudFormation to deploy the SAM application
```
aahung, line 23:
```suggestion
\n\t - CloudFormation Execution Role (IAM Role) to be assumed by CloudFormation to deploy the SAM application
```

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+\n\t - Artifacts bucket(S3 bucket) to store the sam build artifacts
```
aahung, line 24:
```suggestion
\n\t - Artifacts bucket (S3 bucket) to store the sam build artifacts
```

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+\n\t - ECR repo for the container images of Lambda functions having PackageType property set to Image
```
aahung, line 25:
```suggestion
\n\t - ECR repository for the container images of Lambda functions having PackageType property set to Image
```

aahung: Since this is optional, do we want to add "optional" to it?

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+"""
+
+PIPELINE_CONFIG_DIR = os.path.join(".aws-sam", "pipeline")
+PIPELINE_CONFIG_FILENAME = "pipelineconfig.toml"
+
+
+@click.command("bootstrap", short_help=SHORT_HELP, help=HELP_TEXT, context_settings=dict(max_content_width=120))
+@configuration_option(provider=TomlProvider(section="parameters"))
+@click.option(
+    "--interactive/--no-interactive",
```
aahung, line 35: is interactive the default? Why we have both `--interactive` and `--no-interactive` here? It looks conflicting

Reference to our existing code: https://github.com/aws/aws-sam-cli/blob/5f87fb2c762ef920e9dea3cfe3027c1aa2e1b184/samcli/commands/init/__init__.py#L121-L125

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    help="Disable interactive prompting for bootstrap parameters, and fail if any required arguments are missing.",
+)
+@click.option(
+    "--stage-name",
+    help="The name of the corresponding pipeline stage. It is used as a suffix for the created resources.",
+    required=False,
+)
+@click.option(
+    "--pipeline-user",
+    help="The ARN of the IAM user having its AccessKeyId and SecretAccessKey shared with the CI/CD provider."
```
aahung, line 47: same as above, AccessKeyId and SecretAccessKey

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    "It is used to grant this IAM user the permissions to access the corresponding AWS account."
+    "If not provided, the command will create one along with AccessKeyId and SecretAccessKey credentials.",
+    required=False,
+)
+@click.option(
+    "--pipeline-execution-role",
+    help="The ARN of an IAM Role to be assumed by the pipeline-user to operate on this stage. "
```
aahung, line 54: "pipeline-user" -> "pipeline user"

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    "Provide it only if you want to user your own role, otherwise, the command will create one",
+    required=False,
+)
+@click.option(
+    "--cloudformation-execution-role",
+    help="The ARN of an IAM Role to be assumed by the CloudFormation service while deploying the application's stack "
```
aahung, line 60:
```suggestion
    help="The ARN of an IAM Role to be assumed by the CloudFormation service while deploying the application's stack. "
```

image

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    "Provide it only if you want to user your own role, otherwise, the command will create one",
+    required=False,
+)
+@click.option(
+    "--artifacts-bucket",
+    help="The ARN of a S3 bucket to hold the sam build artifacts. "
```
aahung, line 66: we used "store" above

```suggestion
    help="The ARN of a S3 bucket to store the sam build artifacts. "
```

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    "Provide it only if you want to user your own S3 bucket, otherwise, the command will create one",
+    required=False,
+)
+@click.option(
+    "--create-ecr-repo/--no-create-ecr-repo",
```
aahung, line 71: same as `--interactive`, if we have a default, why we still need two flags.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    is_flag=True,
+    default=False,
+    help="If set to true and no ecr-repo is provided this command will create an ECR repo to hold the image container "
+    "of the lambda functions having Image package type.",
+)
+@click.option(
+    "--ecr-repo",
+    help="The ARN of an ECR repo to hold the image containers of the lambda functions of image package type. "
```
aahung, line 79:
```suggestion
    help="The ARN of an ECR repository to store the image containers of the lambda functions of image package type. "
```

aahung: Above we wrote:

> ECR repo for the container images of Lambda functions having PackageType property set to Image

I think we need to find a consistent way to describe `PackageType=Image`

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    "If Provided, the create-ecr-repo argument is ignored. If not provided and create-ecr-repo is set to true "
+    "The command will create one.",
```
aahung, line 81:
```suggestion
    "If Provided, the create-ecr-repo argument is ignored. If not provided and create-ecr-repo is set to true, "
    "the command will create one.",
```

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    required=False,
+)
+@click.option(
+    "--pipeline-ip-range",
+    help="If provided, all requests coming from outside of the given range(s) are denied.",
```
aahung, line 86: It is not apparent what format customers should use here. Maybe provide some examples

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    required=False,
+)
+@click.option(
+    "--confirm-changeset/--no-confirm-changeset",
```
aahung, line 90: same as `--interactive`

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+
+    stage.bootstrap(confirm_changeset=confirm_changeset)
+
+    stage.print_resources_summary()
+
+    try:
+        stage.save_config(
+            config_dir=PIPELINE_CONFIG_DIR, filename=PIPELINE_CONFIG_FILENAME, cmd_names=_get_command_name()
+        )
+    except Exception:
```
aahung, line 204: What are the possible exception types here?

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,221 @@
...
+    if not samconfig.exists():
+        return None
+    config: Dict[str, str] = samconfig.get_all(cmd_names=_get_command_name(), section="parameters")
+    return config.get("pipeline_user")
+
+
+def _get_command_name() -> List[str]:
+    ctx = click.get_current_context()
+    cmd_names: List[str] = cast(List[str], get_cmd_names(ctx.info_name, ctx))  # ["pipeline", "bootstrap"]
+    return cmd_names
```
aahung, line 221: what is this function's purpose, looks like it is only used in `samconfig.get_all(cmd_names=_get_command_name(), section="parameters")`

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    Decompose a given ARN into its parts
+
+    Attributes
+    ----------
+    partition: str
```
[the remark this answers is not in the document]
elbayaaa: I followed [this](https://realpython.com/documenting-python-code/) format assuming it is standard.

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: It is intended(henece the "\n") that is my own preference but it may change after the UX review

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: I got this format from the `deploy` command https://github.com/aws/aws-sam-cli/blob/c99f1eee1ef4b23ffffde85e6c5f669b3970f083/samcli/commands/deploy/command.py#L139

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: I'm quite unsure what do you mean; the example in the official docs has a default value and still define both of them `@click.option('--shout/--no-shout', default=False)` https://click.palletsprojects.com/en/7.x/options/#boolean-flags

samcli/lib/pipeline/bootstrap/stage.py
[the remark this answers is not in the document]
hoffa: I'm fine not doing so, just wasn't clear where this file is.

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
hoffa: You can see some examples here without types: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

In any case this is error-prone, as shown by e.g. `ARNParts` or `Stage` where the types are incorrect.

samcli/commands/pipeline/bootstrap/cli.py
hoffa: I'd agree, unless there's a logical grouping between parts of text, we should avoid extra newlines (this applies especially to terminals, where real estate can be very limited).

elbayaaa: Let's ignore the UX issues for now; there is a UX review next week.

CoshUS:
FYI `continuous-integration/appveyor/pr/windows-build-java-ruby-inprocess` is broken. The fix is in develop now.

samcli/commands/pipeline/bootstrap/cli.py
hoffa, line 25: Terminology consistent with AWS docs, and added IAM/S3 to the term itself as they're newly introduced in this context:

```suggestion
\n\t - Pipeline IAM user with access key ID and secret access key credentials to be shared with the CI/CD provider
\n\t - Pipeline execution IAM role assumed by the pipeline user to obtain access to the AWS account
\n\t - CloudFormation execution IAM role assumed by CloudFormation to deploy the AWS SAM application
\n\t - Artifacts S3 bucket to store the AWS SAM build artifacts
\n\t - Optionally, an ECR repository to store container image Lambda deployment packages
```

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: The intention is to capture all kind of exceptions(including unexpected ones); At this point the resources are already bootstrapped, we don't prompt/confirm with the users that we are going to save these configs, i.e. they are unaware of that. So, the bootstrap command has already succeeded and the users should face any failure messages.

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: this returns the name of the sub-section of the toml file where we save the created infrastructure

samcli/lib/pipeline/bootstrap/resource.py
aahung, line 5: this class can be converted into a `NamedTuple` because we don't need to mutate any attributes after it is created.

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+        if not isinstance(arn, str):
+            raise ValueError(f"Invalid ARN ({arn}) is not a String")
+
+        parts = arn.split(":")
+        try:
+            self.partition: str = parts[1]
+            self.service: str = parts[2]
+            self.region: str = parts[3]
+            self.account_id: str = parts[4]
+            self.resource_id: str = parts[5]
```
aahung, line 35: This looks simplier

```suggestion
            [self.partition, self.service, self.region......] = parts[1]
```

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+            self.resource_type: Optional[str] = None
+        except IndexError as ex:
+            raise ValueError(f"Invalid ARN ({arn})") from ex
+
+        if "/" in self.resource_id:
+            resource_type_and_id: List[str] = self.resource_id.split("/")
+            self.resource_type = resource_type_and_id[0]
+            self.resource_id = resource_type_and_id[1]
```
aahung, line 43: simpler version: 

```suggestion
            [self.resource_type, self.resource_id] = self.resource_id.split("/")
```

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_uri(self) -> Optional[str]:
+        """
+        extracts and returns the URI of the given ECR repo from its ARN
+        Raises
+        ------
+        ValueError if the ARN is invalid
+        """
+        arn_parts: Optional[ARNParts] = self._get_arn_parts()
+        return (
```
aahung, line 129: can you add this link as a comment here https://docs.aws.amazon.com/general/latest/gr/ecr.html about why the link is constructed this way?

samcli/lib/pipeline/bootstrap/stage.py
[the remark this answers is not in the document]
aahung: we should have a constant list at the top for the list of files

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,276 @@
...
+                "CloudFormationExecutionRoleArn": self.cloudformation_execution_role.arn or "",
+                "ArtifactsBucketArn": self.artifacts_bucket.arn or "",
+                "CreateEcrRepo": "true" if self.create_ecr_repo else "false",
+                "EcrRepoArn": self.ecr_repo.arn or "",
+            },
+        )
+
+        self.pipeline_user.arn = output.get("PipelineUser")
+        self.pipeline_user.access_key_id = output.get("PipelineUserAccessKeyId")
+        self.pipeline_user.secret_access_key = output.get("PipelineUserSecretAccessKey")
```
aahung, line 164: Can we replace the entire `self.pipeline_user` instead of assigning attributes individually? I prefer to make `pipeline_user` an immutable object so that we know it is "updating the entire user object instead of mutating attributes." And it also ensures we never update some of the attributes and forget the others (out of sync)

aahung: same goes to `self.artifacts_bucket` below

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,276 @@
...
+        samconfig: SamConfig = SamConfig(config_dir=config_dir, filename=filename)
+
+        if self.pipeline_user.arn:
+            samconfig.put(cmd_names=cmd_names, section="parameters", key="pipeline_user", value=self.pipeline_user.arn)
+
+        if self.pipeline_execution_role.arn:
+            samconfig.put(
+                cmd_names=cmd_names,
+                section="parameters",
+                key="pipeline_execution_role",
```
aahung, line 207: are we using the keys in the future? May be worth to put them into constants

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -0,0 +1,94 @@
...
+            click.echo("\t2 - Yes, I need a help creating one")
+            click.echo("\t3 - I already have an ECR repo")
+            choice = click.prompt(text="Choice", show_choices=False, type=click.Choice(["1", "2", "3"]))
+            if choice == "1":
+                self.create_ecr_repo = False
+            elif choice == "2":
+                self.create_ecr_repo = True
+            else:  # choice == "3"
+                self.create_ecr_repo = False
+                self.ecr_repo_arn = click.prompt("ECR repo", type=click.STRING)
```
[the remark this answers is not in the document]
elbayaaa: No, user has to choose one of them

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: simpler like better readability?

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,276 @@
...
+            be prompted for confirmation
+
+        Returns True if bootstrapped, otherwise False
+        """
+
+        if self.did_user_provide_all_required_resources():
+            click.secho(f"\nAll required resources for the {self.name} stage exist, skipping creation.", fg="yellow")
+            return True
+
+        missing_resources: str = self._get_non_user_provided_resources_msg()
```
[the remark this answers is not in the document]
elbayaaa: yeah, this should be missing_resources_message

samcli/lib/pipeline/bootstrap/resource.py
aahung: this is python's way to assign elements in a list to different variables. Yeah, it helps with the readability and it is less prune to typos because there are no need to maintain the index

samcli/commands/pipeline/bootstrap/guided_context.py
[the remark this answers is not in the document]
aahung: if user types "4" will we treat it as "3" or raise an exception?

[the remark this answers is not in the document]
elbayaaa: Click guarantees that the choice will be one of ["1", "2" or "3"].

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: IMO, the current assignment is more readable in this case especially with the type-hint inplace. Regarding index being error prone, it is the same issue if we swapped two variables.

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: actually, I'm reverting this, figured out resource_id can have multiple slashes (resource_path)

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: seems like an invalid NamedTupel case; "Cannot overwrite NamedTuple attribute __init__"

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
elbayaaa: Where is the incorrect types in `ARNParts` and `Stage`, I didn't find it?

I'm trying to see what is the standards; I see in our code base we set the parameter's type in the docs. Also, even without type, it is still error-prone if you added/removed/renamed a parameter without modifying the docs accordingly. So the question is whether fixing the discrepancy is the correct way or not.

samcli/lib/pipeline/bootstrap/stage.py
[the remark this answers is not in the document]
elbayaaa: @aahung would you elaborate more in this? I didn't get your point

samcli/lib/pipeline/bootstrap/stage.py
elbayaaa: We actually update some resources and leave others here; A resource has an attribute `is_user_provided` that is set to true during `init` only if the user provided the ARN during the interactive questionnaire.

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
hoffa: In `ARNParts` the names are wrong (e.g. `account-id` instead of `account_id`), and in `Stage` the types are wrong (e.g. `aws_profile: str` in docstring and `aws_profile: Optional[str]` in type hint).

Ensuring type hints and docstring have same arguments is one thing; ensuring types are correct is yet another thing on top of that.

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -0,0 +1,94 @@
...
+        if not self.artifacts_bucket_arn:
+            self.artifacts_bucket_arn = click.prompt(
+                "\nArtifacts bucket(S3 bucket to hold the sam build artifacts) " "[leave blank to create one]",
+                default="",
+                type=click.STRING,
+            )
+        if not self.ecr_repo_arn:
+            click.echo(
+                "\nIf your SAM template will include lambda functions of Image package-type, "
+                "then an ECR repo is required, should we create one?"
```
hoffa, line 76:
```suggestion
                "\nIf your SAM template will include Lambda functions of Image package type, "
                "then an ECR repository is required. Should we create one?"
```

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+            click.echo("\t1 - No, My SAM Template won't include lambda functions of Image package-type")
+            click.echo("\t2 - Yes, I need a help creating one")
+            click.echo("\t3 - I already have an ECR repo")
```
hoffa, line 80:
```suggestion
            click.echo("\t1 - No, My SAM template won't include Lambda functions of Image package type")
            click.echo("\t2 - Yes, I need help creating one")
            click.echo("\t3 - I already have an ECR repository")
```

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -0,0 +1,94 @@
...
+            if choice == "1":
+                self.create_ecr_repo = False
+            elif choice == "2":
+                self.create_ecr_repo = True
+            else:  # choice == "3"
+                self.create_ecr_repo = False
+                self.ecr_repo_arn = click.prompt("ECR repo", type=click.STRING)
+
+        if not self.pipeline_ip_range:
+            click.echo("\nWe can deny requests if not coming from a recognized IP address.")
```
hoffa, line 91:
```suggestion
            click.echo("\nWe can deny requests not coming from a recognized IP address range.")
```

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -0,0 +1,94 @@
...
+            self.pipeline_ip_range = click.prompt(
+                "Pipeline IP address range [leave blank if you don't know]", default="", type=click.STRING
```
hoffa, line 93:
```suggestion
                "Pipeline IP address range (using CIDR notation) [leave blank if you don't know]", default="", type=click.STRING
```

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,282 @@
...
+            click.secho(
+                "\nYou provided the following resources. Please make sure it has the required permissions as shown at "
+                "https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/pipeline/bootstrap/stage_resources.yaml",
+                fg="green",
+            )
+            for resource in provided_resources:
+                click.secho(f"\t{resource.arn}", fg="green")
+
+        if not self.pipeline_user.is_user_provided:
+            click.secho("Please configure your CICD project with the following pipeline-user credentials:", fg="green")
```
hoffa, line 280:
```suggestion
            click.secho("Please configure your CI/CD project with the following pipeline user credentials:", fg="green")
```

samcli/lib/pipeline/bootstrap/stage.py
aahung: I mean when we update a resource, we update all attributes of that resource right? If that's the case, doing a single `self.pipeline_user = ...` (with `pipeline_user` being an immutable object like Tuple) is less error prune than having three lines `self.pipeline_user.* = ` to update individual attributes.

samcli/lib/pipeline/bootstrap/resource.py
aahung: another downside of doing it using index: we cannot verify the length of `parts`. Take this as an example, if the format is unexpected and making `parts` have 7 elements, this code will still run without exception.

samcli/commands/pipeline/bootstrap/cli.py
aahung: what are the possible return values of this `_get_command_name()`?

aahung (commented):
Thanks for addressing the comments!

samcli/commands/pipeline/bootstrap/cli.py
elbayaaa: it is always ["pipeline", "bootstrap"] but I don't want to hardcode it to be consistent with the other places in code

samcli/commands/pipeline/bootstrap/guided_context.py
elbayaaa: I used the future tense to let the users think about the future; Instead of thinking "no my templates doesn't contain this" they will think "will my template contain this".

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
elbayaaa: Why do you think they are different? it is the same exact issue; discrepancies between the docs and the actual arguments.

samcli/lib/pipeline/bootstrap/stage.py
elbayaaa: "when we update a resource, we update all attributes of that resource right?" Nope, we don't, that is the point.

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
hoffa: Sure, they're both a discrepancy. But duplicated types requires ensuring (1) the names match, and (2) the types match, while not specifying type in docstring only requires (1).

samcli/commands/pipeline/bootstrap/guided_context.py
aahung: but it is not how "if" is used in English as far as I know. You can check the article I posted above, they cover all scenarios and none of them use future tense. (specifically about the "likely")

aahung: more reading: https://www.grammarly.com/blog/will-would-in-if-clause/

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
aahung: @hoffa and I had this chat before about whether or not to include types in docstring. The conclusion was just to not "have" to because it will increase the maintenance cost and it does not bring any values in modern IDEs (IDE will pick up types from signatures).

samcli/lib/pipeline/bootstrap/resource.py
aahung:
```py
class ARNParts:

    partition: str
    service: str
    region: str
    account_id: str
    resource_id: str

    def __init__(self, arn: str) -> None:
        if not isinstance(arn, str):
            raise ValueError(f"Invalid ARN ({arn}) is not a String")

        parts = arn.split(":")
        try:
            [self.partition, self.service, self.region, self.account_id, self.resource_id] = parts
        except IndexError as ex:
            raise ValueError(f"Invalid ARN ({arn})") from ex
```

Here is a more concrete example why this notation is better: it can catch fires when `parts` has more elements than expected:

image

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
elbayaaa: then while reading the docs (running `help(X)`) of a class or method with many attributes, we will need to keep jumping over to the top of the docs to know the type of an argument that is documented all the way down in the list.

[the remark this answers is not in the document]
aahung: Generally we would avoid to have too many attributes (7 is the limit by pylint), but that's a fair point.

samcli/commands/pipeline/bootstrap/cli.py
aahung: that makes sense. `_get_command_name()` is a generic function right? If so I think it is better to place in a more common file (also let's add an example to its docstring, I was confused about what values it would generate)

aahung (commented):
Overall it LGTM, a few suggestions and questions, I need to go through the template yaml

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,222 @@
...
+    samconfig: SamConfig = SamConfig(config_dir=PIPELINE_CONFIG_DIR, filename=PIPELINE_CONFIG_FILENAME)
+    if not samconfig.exists():
+        return None
+    config: Dict[str, str] = samconfig.get_all(cmd_names=_get_command_name(), section="parameters")
+    return config.get("pipeline_user")
+
+
+def _get_command_name() -> List[str]:
+    ctx = click.get_current_context()
+    cmd_names: List[str] = cast(List[str], get_cmd_names(ctx.info_name, ctx))  # ["pipeline", "bootstrap"]
```
hoffa, line 221: Instead of `cast`s would be better to update the function type hints.

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
hoffa: In any case the types in the `Stage` docstring are wrong, which is my point. Too easy to make a mistake.

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+        if not arn_parts:
+            return None
+        # ECR's resource_id contains the resource-type("resource") which is excluded from the URL
+        # from docs: https://docs.aws.amazon.com/AmazonECR/latest/userguide/security_iam_service-with-iam.html
+        # ECR's ARN: arn:${Partition}:ecr:${Region}:${Account}:repository/${Repository-name}
+        if not arn_parts.resource_id.startswith("repository/"):
+            raise ValueError(f"Invalid ECR ARN ({self.arn}), can't extract the URL from it.")
+        i = len("repository/")
+        repo_name = arn_parts.resource_id[i:]
+        return f"{arn_parts.account_id}.dkr.ecr.{arn_parts.region}.amazonaws.com/{repo_name}"
```
hoffa, line 133: Is this format documented?

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,280 @@
...
+        if self.did_user_provide_all_required_resources():
+            click.secho(f"\nAll required resources for the {self.name} stage exist, skipping creation.", fg="yellow")
+            return True
+
+        missing_resources_msg: str = self._get_non_user_provided_resources_msg()
+        click.echo(
+            "This will create the following required resources for the {self.name} stage: " f"{missing_resources_msg}"
+        )
+        if confirm_changeset:
+            confirmed: bool = click.confirm("Should we proceed with the creation?")
```
hoffa, line 147: Is this option related to CloudFormation ChangeSets? (as is the case for `sam deploy`)

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,280 @@
...
+        self.cloudformation_execution_role.arn = output.get("CloudFormationExecutionRole")
+        self.artifacts_bucket.arn = output.get("ArtifactsBucket")
+        self.artifacts_bucket.kms_key_arn = output.get("ArtifactsBucketKMS")
+        self.ecr_repo.arn = output.get("EcrRepo")
+        return True
+
+    @staticmethod
+    def _read_template(template_file_name: str) -> str:
+        template_path: str = os.path.join(CFN_TEMPLATE_PATH, template_file_name)
+        with open(template_path, "r") as fp:
```
hoffa, line 183: This will fail on Windows on UTF-8 bytes. See e.g. https://github.com/aws/aws-sam-cli/pull/2583

I'm surprised we don't have a generic method for reading a file as a string.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,222 @@
...
+
+def _load_saved_pipeline_user_arn() -> Optional[str]:
+    samconfig: SamConfig = SamConfig(config_dir=PIPELINE_CONFIG_DIR, filename=PIPELINE_CONFIG_FILENAME)
+    if not samconfig.exists():
+        return None
+    config: Dict[str, str] = samconfig.get_all(cmd_names=_get_command_name(), section="parameters")
+    return config.get("pipeline_user")
+
+
+def _get_command_name() -> List[str]:
```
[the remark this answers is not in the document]
elbayaaa: why do you want to add docstring to a private method? 
The method that do the actual logic, i.e. `get_cmd_names` is already documented https://github.com/aws/aws-sam-cli/blob/d4796787e00e8a6abe8a3aef22fb090c2beb9212/samcli/cli/context.py#L190-L204

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
elbayaaa: let me fix the wrong types; Removing all of the types should be a team decision and applied in a separate command as it is:
1. Will affect everywhere; most of the docstrong across the package include type-hint.
2. Seems to be not the standard way.
2.

samcli/commands/pipeline/bootstrap/cli.py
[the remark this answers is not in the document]
aahung: > why do you want to add docstring to a private method?

Because it is not intuitively clear what this function will output. If I just look at it, it involves `click` and another function. Without the click context, docstring can help to understand what it does.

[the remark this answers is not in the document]
aahung: Besides, `get_cmd_names()` is a confusing function itself.

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    partition: str
+    service: str
+    region: str
+    account_id: str
+    resource_id: str
+
+    def __init__(self, arn: str) -> None:
+        if not isinstance(arn, str):
+            raise ValueError(f"Invalid ARN ({arn}) is not a String")
```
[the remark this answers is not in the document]
elbayaaa: ARN is a user input so I made it that way just in case if the user entered non-string input (numbers for example) and we didn't cast to string. So basically a user can enter a number and python don't enforce types.

samcli/commands/pipeline/bootstrap/cli.py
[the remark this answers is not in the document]
elbayaaa: Ummmm, are you saying the function name `_get_command_name` doesn't clearly indicates that this function returns the name of the  command?

this function just combine two lines of code together instead of duplicating them every time.

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: Yes, see the URL included in the docstring

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -0,0 +1,280 @@
...
+        Returns True if bootstrapped, otherwise False
+        """
+
+        if self.did_user_provide_all_required_resources():
+            click.secho(f"\nAll required resources for the {self.name} stage exist, skipping creation.", fg="yellow")
+            return True
+
+        missing_resources_msg: str = self._get_non_user_provided_resources_msg()
+        click.echo(
+            "This will create the following required resources for the {self.name} stage: " f"{missing_resources_msg}"
```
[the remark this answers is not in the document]
elbayaaa: Most probably this came from black

samcli/commands/pipeline/bootstrap/cli.py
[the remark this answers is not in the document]
aahung: yeah.. the confusing part is, which of the following is the output?
- `["sam", "pipeline", "bootstrap"]`
- `["sam", "pipeline"]`
- `["pipeline", "bootstrap"]`
If I cannot get the answer by looking at the function code, I think it deserves a docstring (comment)

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
aahung: That makes sense. However, it is confusing here and I don't think it is `Resource`'s job to deal with it. The user input for `ARN` should be converted to string when we receive the inputs, maybe in `GuidedContext`. 

I did a test and looks like `click.prompt()` convert the input to string itself, see:
image Do we really need this check at all?

samcli/lib/pipeline/bootstrap/stage.py
elbayaaa: yes, used like here https://github.com/aws/aws-sam-cli/blob/d965ee4d3228c8f7fd76e2cb083f22a0d4841ec7/samcli/commands/deploy/deploy_context.py#L237

samcli/lib/pipeline/bootstrap/stage.py
elbayaaa: Are you saying python's builtin `open` function working on windows? the issue you are linking uses `click.open`

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -25,10 +22,45 @@ def __init__(self, ex):
...
+        self._stack_output: List[Dict[str, str]] = stack_output
+
+    def get(self, key) -> Optional[str]:
+        try:
+            return next(o for o in self._stack_output if o.get("OutputKey") == key).get("OutputValue")
+        except StopIteration:
+            return None
+
+
+def manage_stack(
```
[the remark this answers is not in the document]
elbayaaa: Not 100% sure why it is named `manage_stack` but let's not change it in this PR.

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -25,10 +22,45 @@ def __init__(self, ex):
...
+        AWS named profile for the AWS account
+    parameter_overrides: Optional[Dict[str, Union[str, List[str]]]]
+        Values of template parameters, if any.
+
+    Returns: Stack output section(list of OutputKey, OutputValue pairs)
+    """
     try:
         if profile:
-            session = boto3.Session(profile_name=profile, region_name=region if region else None)
+            session = boto3.Session(profile_name=profile, region_name=region if region else None)  # type: ignore
```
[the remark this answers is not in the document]
elbayaaa: I didn't change this logic but maybe an empty-string VS None thing.

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -134,3 +183,15 @@ def _create_stack(cloudformation_client, stack_name, template_body):
...
+
+
+def _generate_stack_parameters(
+    parameter_overrides: Optional[Dict[str, Union[str, List[str]]]] = None
+) -> List[Dict[str, str]]:
+    parameters = []
+    if parameter_overrides:
+        for key, value in parameter_overrides.items():
+            if isinstance(value, Collection) and not isinstance(value, str):
+                value = ",".join(value)
```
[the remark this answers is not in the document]
elbayaaa: Not sure how cloudformation handles this, we don't have a List parameter in our case.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -0,0 +1,222 @@
...
+        stage.print_resources_summary()
+
+        try:
+            stage.save_config(
+                config_dir=PIPELINE_CONFIG_DIR, filename=PIPELINE_CONFIG_FILENAME, cmd_names=_get_command_name()
+            )
+        except Exception:
+            # Swallow saving exceptions, if any, as the resources are already bootstrapped and the ARNs are already
+            # printed out in the screen.
+            pass
```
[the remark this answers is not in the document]
elbayaaa: it is a common approach to have a safe operation that never fail. What about defining a method `save_config_safe` and moving this logic to there?

samcli/commands/pipeline/bootstrap/cli.py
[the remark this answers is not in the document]
elbayaaa: then this should be fixed in the public method `get_cmd_names()` itself, right?

samcli/lib/pipeline/bootstrap/stage.py
hoffa: It uses `open` under the hood. If `encoding` isn't specified, it defaults to whatever's the system default (typically Latin-1 on Windows).

samcli/lib/pipeline/bootstrap/resource.py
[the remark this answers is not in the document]
elbayaaa: this is because we specifically set the type to `click.STRING`. I'm just saying what if we forgot to make such a casting. This check just adds one more layer of safety in this case, otherwise, it will raise an AttributeError, and every call to this method should handle both ValueError and AttributeError instead of ValueError only.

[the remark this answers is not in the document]
aahung: > I'm just saying what if we forgot to make such a casting. 

For now we don't need to worry about it. Even if we forgot to make such a casting happens in the future, it is our code bug that we should fix. Besides from a code comprehension aspect, it creates this extra mental burden: "if the type does not match the hint, how about others, should we do this everywhere?" I don't think we want to do this

hoffa (approved):
LGTM! Thanks for addressing the comments! 🙌

aahung (approved):

aahung:
Thanks for addressing all the comments!

samcli/lib/pipeline/bootstrap/stage_resources.yaml
```diff
@@ -0,0 +1,359 @@
...
+      Tags:
+        - Key: ManagedStackSource
+          Value: AwsSamCli
+      VersioningConfiguration:
+        Status: Enabled
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+          - ServerSideEncryptionByDefault:
+              KMSMasterKeyID: !GetAtt ArtifactsBucketKey.Arn
+              SSEAlgorithm: aws:kms
```
hoffa, line 178: How about using S3-SSE so there's no key management?

```yaml

BucketEncryption:
  ServerSideEncryptionConfiguration:
    - ServerSideEncryptionByDefault:
        SSEAlgorithm: AES256
```

samcli/lib/pipeline/bootstrap/stage_resources.yaml
```diff
@@ -0,0 +1,359 @@
...
+              Resource:
+                Fn::If:
+                  - MissingEcrRepo
+                  - !GetAtt EcrRepo.Arn
+                  - !Ref EcrRepoArn
+            - !Ref AWS::NoValue
+      Roles:
+        - !Ref PipelineExecutionRole
+
+  EcrRepo:
```
sriram-mv, line 280: https://github.com/aws/aws-sam-cli/pull/2675 creates an ECR repo such that its per function. @CoshUS can you add your comments here? My concern is we do conflicting things via ci/cd vs guided deploy experience.

CoshUS: We had a discussion on the overlapping parts of these two features. They have enough differences between them (command executor permission, image per function, and stack per account) that keeping them separate is the current option we are going for. However, we will make sure the ECR repo permissions are consistent and the documentation states clearly which one should be used for what use-case.

elbayaaa: That is a good call out. We will make sure of the consistency between the two functionality. But both are needed as they have different use-cases; in the non CI/CD case, `sam deploy` is ran by the user who has an administrative permissions to the account while in case of CI/CD `sam deploy` is ran through an IAM entity that has limited permissions. So, creating the ECR repo should be part of the `sam pipeline bootstrap` which is ran by the normal user(the one with the administrative access)

samcli/lib/pipeline/bootstrap/stage_resources.yaml
elbayaaa: Done

samcli/lib/pipeline/bootstrap/resource.py
```diff
@@ -0,0 +1,131 @@
...
+    kms_key_arn: Optional[str]
+        The ARN of the KMS key used in encrypting this S3Bucket, if any.
+    """
+
+    def __init__(self, arn: Optional[str], kms_key_arn: Optional[str] = None) -> None:
+        self.kms_key_arn: Optional[str] = kms_key_arn
+        super().__init__(arn=arn)
+
+
+class EcrRepo(Resource):
```
CoshUS, line 107: Nit: ECRRepo instead of EcrRepo. Same goes for other class names.
> Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError.

samcli/lib/pipeline/bootstrap/stage_resources.yaml
```diff
@@ -0,0 +1,307 @@
...
+    Type: AWS::ECR::Repository
+    Condition: MissingEcrRepo
+    Properties:
+      RepositoryPolicyText:
+        Version: "2012-10-17"
+        Statement:
+          - Sid: AllowPushPull
+            Effect: Allow
+            Principal:
```
CoshUS, line 242: How does Lambda get permission to pull and use the images?

CoshUS (changes requested):

samcli/lib/pipeline/bootstrap/stage_resources.yaml
elbayaaa: To use image-based lambdas, it is the IAM user/role that requires ECR permissions, not the function itself https://docs.aws.amazon.com/lambda/latest/dg/configuration-images.html#configuration-images-permissions

samcli/lib/pipeline/bootstrap/resource.py
elbayaaa: I agree on that but we are already use the CamelCase with the abbreviations in many other places like for example:
AWS::Serverless::Api and AWS::Serverless::HttpApi

CoshUS: This is for Python naming convention instead of CFN. https://legacy.python.org/dev/peps/pep-0008/#descriptive-naming-styles

elbayaaa: I see your point, fixed.

CoshUS (approved):

PR #2815, opened 2021-04-16
feat(Auto ECR): Added Option to Auto Create ECR Repos During Deployment

CoshUS:
Added options to auto create ECR repos when deploying.
RFC: https://github.com/aws/aws-sam-cli/pull/2675

Also fixes Windows ANSI output during image pushes.
`samcli/lib/package/stream_cursor_utils.py`

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/deploy/command.py
```diff
@@ -150,11 +151,14 @@
     "the output AWS CloudFormation template. YAML is used by default.",
 )
 @click.option(
-    "--resolve-s3",
+    "--resolve-s3", required=False, is_flag=True, help="Automatically resolve s3 bucket for non-guided deployments."
```
hoffa, line 154:
```suggestion
    "--resolve-s3", required=False, is_flag=True, help="Automatically resolve S3 bucket for non-guided deployments."
```

samcli/commands/deploy/command.py
```diff
@@ -150,11 +151,14 @@
...
+)
+@click.option(
+    "--resolve-image-repos",
     required=False,
     is_flag=True,
-    help="Automatically resolve s3 bucket for non-guided deployments."
-    "Do not use --s3-guided parameter with this option.",
+    help="Automatically create and delete ECR repos for image based functions in non-guided deployments."
+    "Auto created image repos will be deleted if the corresponding functions are removed.",
```
hoffa, line 161: I haven't seen "repo" being used in any AWS docs (compare `"repo" site:docs.aws.amazon.com` versus `"repository" site:docs.aws.amazon.com`).

```suggestion
    help="Automatically create and delete ECR repositories for image-based functions in non-guided deployments. "
    "Automatically created image repositories will be deleted if the corresponding functions are removed.",
```

samcli/commands/deploy/guided_context.py
```diff
@@ -169,9 +167,14 @@ def guided_prompts(self, parameter_override_keys):
                 type=click.STRING,
             )
 
+        click.echo("\n\tLooking for resources needed for deployment:")
         s3_bucket = manage_stack(profile=self.profile, region=region)
-        click.echo(f"\n\t\tManaged S3 bucket: {s3_bucket}")
-        click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
+        click.echo(f"\t Managed S3 bucket: {s3_bucket}")
+        click.echo("\t A different default S3 bucket can be set in samconfig.toml")
```
hoffa, line 173: Applies to most prompts/outputs, there's an extra space on top of the tab:
```suggestion
        click.echo(f"\tManaged S3 bucket: {s3_bucket}")
        click.echo("\tA different default S3 bucket can be set in samconfig.toml")
```

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+AWSTemplateFormatVersion : '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: AWS SAM CLI Managed ECR Repo Stack
+Metadata:
+  SamCliInfo: $sam_cli_version
+  CompanionStackname:  $companion_stack_name
+
+Resources:
+$resources
```
hoffa, line 22: Using string interpolation is fragile to changes, as it's both indentation-sensitive and cannot be easily syntax-checked. Would using Python objects and serializing to JSON help?

samcli/lib/bootstrap/companion_stack/data_types.py
```diff
@@ -0,0 +1,127 @@
...
+    _parent_stack_name: str
+    _escaped_parent_stack_name: str
+    _parent_stack_hash: str
+    _stack_name: str
+
+    def __init__(self, parent_stack_name: str) -> None:
+        self._parent_stack_name = parent_stack_name
+        self._escaped_parent_stack_name = re.sub(r"[^a-z0-9]", "", self._parent_stack_name.lower())
+        self._parent_stack_hash = str_checksum(self._parent_stack_name)
+        self._stack_name = f"{self._parent_stack_name[:104]}-{self._parent_stack_hash[:8]}-CompanionStack"
```
hoffa, line 24: Why escaped? Why `104`? `8`?

samcli/lib/bootstrap/companion_stack/data_types.py
```diff
@@ -0,0 +1,127 @@
...
+        self._companion_stack = companion_stack
+
+        self._logical_id = logical_id
+        self._physical_id = physical_id
+        self._output_logical_id = output_logical_id
+
+    @property
+    def logical_id(self) -> Optional[str]:
+        if self._logical_id is None and self._function_logical_id and self._function_md5:
+            self._logical_id = self._function_logical_id[:52] + self._function_md5[:8] + "Repo"
```
hoffa, line 99: Why `52` and `8`?

samcli/commands/deploy/guided_context.py
CoshUS: The spaces were added on purpose as required by UX.

samcli/lib/bootstrap/companion_stack/data_types.py
CoshUS: Escaped stack name is used for naming ECR repos.
There is max 128 characters limit on the length of stack name.
```
repo_name = 
    escaped_stack_name + stack_md5[:8] + "/" + escaped_lambda_logical_id + function_md5[:8] + "repo"
    #128        +         8      +  1  +     64      +        8      +      4 = 213 max char
```
104 for parent stack name, 8 for hashes in case of duplication, 16 for two dashes and "CompanionStack" suffix.

samcli/lib/bootstrap/companion_stack/data_types.py
CoshUS:
```
#MD5 is used to avoid two having the same escaped name with different Lambda Functions
#For example: Helloworld and HELLO-WORLD
repo_logical_id =
    lambda_logical_id [:52] + function_md5 [:8] + "Repo"
    #52 + 8 + 4 = 64 max char
```

hoffa: I see. Please add a comment to some reference or refactor in a way that makes these operations well-defined, otherwise anyone new to the code will have no idea why things are this way.

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
CoshUS: Good point! Updated to use dict + json.dumps.

samcli/commands/deploy/command.py
```diff
@@ -150,11 +151,14 @@
...
-    "--resolve-s3",
+    "--resolve-s3", required=False, is_flag=True, help="Automatically resolve S3 bucket for non-guided deployments."
+)
+@click.option(
+    "--resolve-image-repos",
     required=False,
     is_flag=True,
-    help="Automatically resolve s3 bucket for non-guided deployments."
-    "Do not use --s3-guided parameter with this option.",
+    help="Automatically create and delete ECR repositories for image-based functions in non-guided deployments. "
```
sriram-mv, line 160: should we talk about the mapping here? on new ecr repos per function resource in the template.

samcli/commands/deploy/command.py
```diff
@@ -287,13 +294,23 @@ def do_cli(
...
-    elif resolve_s3:
-        s3_bucket = manage_stack(profile=profile, region=region)
-        click.echo(f"\n\t\tManaged S3 bucket: {s3_bucket}")
-        click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
-        click.echo("\t\tOr by specifying --s3-bucket explicitly.")
+    else:
+        if resolve_s3:
+            if bool(s3_bucket):
+                raise DeployResolveS3AndS3SetError()
+            s3_bucket = manage_stack(profile=profile, region=region)
```
sriram-mv, line 301: I wonder if we should encapsulate creation of other resources other than template to its own class, that way addition of new resources becomes easier in the future.

samcli/commands/deploy/guided_config.py
```diff
@@ -102,7 +102,7 @@ def _save_parameter_overrides(self, cmd_names, config_env, parameter_overrides,
                 samconfig.put(cmd_names, self.section, "parameter_overrides", " ".join(_params), env=config_env)
 
     def _save_image_repositories(self, cmd_names, config_env, samconfig, image_repositories):
-        if image_repositories:
+        if image_repositories is not None:
```
sriram-mv, line 105: is this a bug we are fixing? or better validation, we may want to add typing to the function signature in that case too.

samcli/commands/deploy/guided_context.py
```diff
@@ -285,52 +288,226 @@ def prompt_parameters(
...
+        image_repositories = image_repositories.copy() if image_repositories is not None else {}
+        self.function_provider = SamFunctionProvider(stacks, ignore_code_extraction_warnings=True)
+
+        manager_helper = CompanionStackManagerHelper(
+            stack_name, region, s3_bucket, s3_prefix, self.template_file, image_repositories
+        )
+
+        create_all_repos = self.prompt_create_all_repos(manager_helper)
+        if create_all_repos:
+            image_repositories.update(manager_helper.manager.get_repository_mapping())
```
sriram-mv, line 338:
```
manager_helper.manager
```

Seems a bit off to me, abstraction wise. should it just be CompanionStack? or something else to showcase manager is contained within an entity that could contain it. Thoughts? I'm not sold on the naming that I proposed either.

samcli/commands/deploy/guided_context.py
```diff
@@ -285,52 +288,226 @@ def prompt_parameters(
...
+                type=click.STRING,
+            )
+            if not is_ecr_url(image_uri):
+                raise GuidedDeployFailedError(f"Invalid Image Repository ECR URI: {image_uri}")
+
+            image_repositories[function_logical_id] = image_uri
+
+        return image_repositories
+
+    def prompt_create_all_repos(self, manager_helper: CompanionStackManagerHelper) -> bool:
```
sriram-mv, line 388: I feel we are mixing prompt logic with computing if we need repos to be created/deleted, can we compute that outside this method and pass in the repos to be created/deleted instead? Thoughts?

samcli/commands/deploy/guided_context.py
```diff
@@ -285,52 +288,226 @@ def prompt_parameters(
...
+            click.echo("\t Deployment aborted!")
+            click.echo(
+                "\t #The deployment was aborted to prevent "
+                "unreferenced managed ECR repositories from being deleted.\n"
+                "\t #You may remove repositories from the SAMCLI "
+                "managed stack to retain them and resolve this unreferenced check."
+            )
+            raise GuidedDeployFailedError("Unreferenced Auto Created ECR Repos Must Be Deleted.")
+
+        return manager_helper.remove_unreferenced_repos_from_mapping(image_repositories)
```
sriram-mv, line 485: the prompts and the actions taken from the prompts seem to be tightly coupled here, maybe we could break them apart.

samcli/commands/deploy/guided_context.py
```diff
@@ -285,52 +288,226 @@ def prompt_parameters(
...
+
+    @staticmethod
+    def verify_images_exist_locally(functions: Dict[str, Function], image_repositories: Dict[str, str]) -> None:
```
sriram-mv, line 488: Nice way to breakout common functionality 👍🏽

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
```diff
@@ -0,0 +1,130 @@
+"""
```
sriram-mv, line 1: I wonder if we should move the s3 bucket stack to be of this format as well, makes it more consistent experience for us to reason about.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,259 @@
...
+        self._companion_stack = CompanionStack(stack_name)
+        self._builder = CompanionStackBuilder(self._companion_stack)
+        self._boto_config = Config(region_name=region if region else None)
+        self._s3_bucket = s3_bucket
+        self._s3_prefix = s3_prefix
+        try:
+            self._cfn_client = boto3.client("cloudformation", config=self._boto_config)
+            self._ecr_client = boto3.client("ecr", config=self._boto_config)
+            self._s3_client = boto3.client("s3", config=self._boto_config)
+            self._account_id = boto3.client("sts").get_caller_identity().get("Account")
```
sriram-mv, line 50: This is interesting, we are now using additional permissions, do we really need the account id? I see its used for fetching the ecr uri, is there no direct api call that will help us there?

Customers may or may not have sts permissions, especially as this does not directly map with the resources they are trying to deploy.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,259 @@
...
+            s3_uploader = S3Uploader(
+                self._s3_client, bucket_name=self._s3_bucket, prefix=self._s3_prefix, no_progressbar=True
+            )
+            # TemplateUrl property requires S3 URL to be in path-style format
+            parts = S3Uploader.parse_s3_url(
+                s3_uploader.upload_with_dedup(temporary_file.name, "template"), version_property="Version"
+            )
+
+        template_url = s3_uploader.to_path_style_s3_url(parts["Key"], parts.get("Version", None))
+        waiter_config = {"Delay": 10, "MaxAttempts": 120}
```
sriram-mv, line 99: let's make common config like this as part of the class or an ENUM like thing to be referenceable.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,259 @@
...
+                pass
+
+    def sync_repos(self) -> None:
+        """
+        Blocking call to sync companion stack with the following actions
+        Create/Update companion stack.
+        Deletes unreferenced repos.
+        Deletes companion stack if there isn't any repo left.
+        """
+        exists = self.does_companion_stack_exist()
```
sriram-mv, line 190: Q: what do we consider hard failure vs soft failure during sync of repos. Lets mention that in the docstring to document behavior.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,259 @@
...
+        Parameters
+        ----------
+        repo: ECRRepo
+
+        Returns
+        -------
+        str
+            ECR repo URI based on account ID and region.
+        """
+        return repo.get_repo_uri(self._account_id, self._region_name)
```
sriram-mv, line 240: Q: why do we need this to determine repo URI? is there no other way to determine it?

samcli/lib/bootstrap/companion_stack/data_types.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @property
+    def physical_id(self) -> Optional[str]:
+        if (
+            self._physical_id is None
+            and self._companion_stack
+            and self._function_md5
+            and self._escaped_function_logical_id
+        ):
+            # 128 + 8 + 1 + 64 + 8 + 4 = 213 max char
```
sriram-mv, line 116: if these cases are referenced in the design doc, please add those references here too.

samcli/lib/package/stream_cursor_utils.py
```diff
@@ -1,11 +1,19 @@
 """
 Stream cursor utilities for moving cursor in the terminal.
 """
+import os
 
 # NOTE: ANSI escape codes.
 # NOTE: Still needs investigation on non terminal environments.
 ESC = "\u001B["
 
+# Enables ANSI escape codes on Windows
```
sriram-mv, line 10: 🎉  Thanks for looking into this!

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -59,11 +59,10 @@ def _create_or_get_stack(cloudformation_client, stack_name, template_body):
         ds_resp = cloudformation_client.describe_stacks(StackName=stack_name)
         stacks = ds_resp["Stacks"]
         stack = stacks[0]
-        click.echo("\n\tLooking for resources needed for deployment: Found!")
         _check_sanity_of_stack(stack, stack_name)
         return stack["Outputs"]
     except ClientError:
-        click.echo("\n\tLooking for resources needed for deployment: Not found.")
+        pass
```
sriram-mv, line 65: Is this something we still need to have? or do we have additional statements which already reflect it and it was a duplicate.

sriram-mv (changes requested):
Took a first pass, did not go through tests yet.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: GetCallerIdentity requires no permission and it's the recommended way of getting account ID.
https://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html
https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

samcli/lib/utils/managed_cloudformation_stack.py
CoshUS: This is an UX change that moves "Found"/"Not Found" to the individual managed resources.
```
Looking for resources needed for deployment:
 Managed S3 bucket: aws-sam-cli-managed-default-samclisourcebucket
 Image repositories: Not found.
```

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Generating repo URI locally will enable the comparison between deployed and to-be-deployed repos. The comparison then can prompt the user which repos will be removed.
This is the same as CFN: `!Sub ${AWS::AccountId}.dkr.ecr.${AWS::Region}.${AWS::URLSuffix}/repo_logical_id` but without going through CFN.

samcli/commands/deploy/command.py
CoshUS: Added more details on mapping.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: I'm not too sure on what you mean by soft failures in this case. The hard failures here are assume to be errors from API calls.

samcli/commands/deploy/guided_context.py
CoshUS: Updated prompt methods (prompt_specify_repos, prompt_create_all_repos, and prompt_delete_unreferenced_repos) to be less coupled.

samcli/commands/deploy/guided_context.py
CoshUS: Updated prompt methods (prompt_specify_repos, prompt_create_all_repos, and prompt_delete_unreferenced_repos) to be less coupled.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Move out as instance variables.

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
CoshUS: Updated S3 template to be using dict.

samcli/commands/deploy/guided_context.py
CoshUS: This is a helper class to bridge things that do not necessarily fit into CompanionStackManager but are needed for CLI purposes. I'm not sure about the naming of it either, CompanionStackManagerHelper just came to mind due to its purpose.

samcli/commands/deploy/command.py
CoshUS: This is a good idea, but current the difference in behavior between package and deploy regarding having --resolve-image-repos will make it less beneficial. We can include this refactoring in the follow up implementation of adding --resolve-image-repos to package.

samcli/commands/deploy/guided_config.py
CoshUS: This is an edge case where all companion stack repos are deleted which leaves image-repositories empty. With the original check, empty dict will not be saved to config.
Added comments explaining this.

samcli/commands/deploy/guided_context.py
```diff
@@ -285,52 +288,210 @@ def prompt_parameters(
...
-                    try:
-                        tag = tag_translation(image)
-                    except NonLocalImageException:
-                        pass
-                    except NoImageFoundException as ex:
-                        raise GuidedDeployFailedError("No images found to deploy, try running sam build") from ex
-                    else:
-                        click.secho(f"\t  {image} to be pushed to {image_repositories.get(resource_id)}:{tag}")
-            click.secho(nl=True)
+        image_repositories = image_repositories.copy() if image_repositories is not None else {}
```
mgrandis, line 329: Overwriting a parameter value is usually not a good idea.
You could maybe rename the local variable to `updated_repositories` or something like that.

samcli/lib/bootstrap/companion_stack/companion_stack_manager_helper.py
```diff
@@ -0,0 +1,80 @@
...
+        self.function_logical_ids = get_template_function_resource_ids(template_file=template_file, artifact=IMAGE)
+        self.missing_repo_functions = list()
+        self.auto_ecr_repo_functions = list()
+        self.manager = CompanionStackManager(stack_name, region, s3_bucket, s3_prefix)
+        self.deployed_repos = self.manager.list_deployed_repos()
+        self.deployed_repo_uris = [self.manager.get_repo_uri(repo) for repo in self.deployed_repos]
+        self.update_sepcified_image_repos(specified_image_repos)
+        self.unreferenced_repos = self.manager.get_unreferenced_repos()
+
+    def update_sepcified_image_repos(self, specified_image_repos: Dict[str, str]) -> None:
```
mgrandis, line 38: Little typo: `specified`

samcli/lib/package/stream_cursor_utils.py
```diff
@@ -1,11 +1,19 @@
...
+if os.name == "nt":
```
mgrandis, line 11: We usually use `platform.system().lower() == "windows"` for this, any reason for the change?

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,262 @@
...
+
+        template_url = s3_uploader.to_path_style_s3_url(parts["Key"], parts.get("Version", None))
+
+        exists = self.does_companion_stack_exist()
+        if exists:
+            self._cfn_client.update_stack(
+                StackName=stack_name, TemplateURL=template_url, Capabilities=["CAPABILITY_AUTO_EXPAND"]
+            )
+            waiter = self._cfn_client.get_waiter("stack_update_complete")
+        elif self._builder.repo_mapping:
```
mgrandis, line 110: is this `if` condition necessary? We have no `else` here, waiter at line 116 could be `None`.

samcli/commands/deploy/command.py
```diff
@@ -287,13 +295,23 @@ def do_cli(
...
+            click.echo(f"\n\t\tManaged S3 bucket: {s3_bucket}")
+            click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
+            click.echo("\t\tOr by specifying --s3-bucket explicitly.")
+
+        if resolve_image_repos:
+            if image_repositories is None:
+                image_repositories = {}
+            manager_helper = CompanionStackManagerHelper(
```
mgrandis, line 310: Feels weird to instantiate a `CompanionStackManagerHelper` to only use its `manager` functions.
Can't we instantiate a `CompanionStackManager` instead?
Or could we merge `CompanionStackManagerHelper` into `CompanionStackManager`?


(Following the [discussion with Sriram](https://github.com/aws/aws-sam-cli/pull/2815/files#r620576568))

samcli/lib/bootstrap/companion_stack/companion_stack_manager_helper.py
CoshUS: Fixed.

samcli/commands/deploy/guided_context.py
CoshUS: Renamed to `updated_repositories`.

samcli/lib/package/stream_cursor_utils.py
CoshUS: The initial thought was that `os.name` and `platform.system().lower()` are the same for this use case and `os` module is already imported for `os.system` call. However, I'm in favor of keeping the consistency with rest of the code. Updated to use `platform` instead.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Removed `elif` and added `repo_mapping` check at the beginning.

samcli/commands/deploy/command.py
CoshUS: `CompanionStackManagerHelper` is used here to gather function logical IDs and resolve cases where some functions have image repos already specified.
As for merging with `CompanionStackManager`, the helper class was created with the intention to not bloat `CompanionStackManager` and separate out functionalities that are not directly related to managing the stack.

samcli/commands/deploy/command.py
sriram-mv: could we add a todo here.

samcli/commands/deploy/guided_context.py
```diff
@@ -285,51 +288,211 @@ def prompt_parameters(
...
+                "\t #Managed repositories will be deleted when "
+                "their functions are removed from the template and deployed"
+            )
+            return confirm(
+                f"\t {self.start_bold}Create managed ECR repositories for all functions?{self.end_bold}", default=True
+            )
+
+        functions_with_repo_count = len(functions) - len(functions_without_repo)
+        click.echo(
+            "\t Image repositories: "
```
sriram-mv, line 422: Should we showcase the mapping of function to repos?

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Build Companion stack template dictionary with Resources and Outputs not filled
+        Returns
+        -------
+        dict
+            Companion stack template dictionary
+        """
+        template = {
+            "AWSTemplateFormatVersion": "2010-09-09",
+            "Transform": "AWS::Serverless-2016-10-31",
```
sriram-mv, line 62: Technically you dont need a transform, since there are no SAM resources.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,265 @@
...
+        -------
+        List[ECRRepo]
+            List of ECR repos deployed for this companion stack
+            Returns empty list if companion stack does not exist
+        """
+        if not self.does_companion_stack_exist():
+            return []
+        repos: List[ECRRepo] = list()
+        stack = boto3.resource("cloudformation", config=self._boto_config).Stack(self._companion_stack.stack_name)
+        resources = stack.resource_summaries.all()
```
sriram-mv, line 145: can we just grok outputs?

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,265 @@
...
+        Returns
+        -------
+        List[ECRRepo]
+            List of ECR repos deployed for this companion stack
+            Returns empty list if companion stack does not exist
+        """
+        if not self.does_companion_stack_exist():
+            return []
+        repos: List[ECRRepo] = list()
+        stack = boto3.resource("cloudformation", config=self._boto_config).Stack(self._companion_stack.stack_name)
```
sriram-mv, line 144: why are we re-creating clients?

sriram-mv (commented):
Last pass, will approve once these comments are addressed.

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
mgrandis: I think this is only the base template, resources are added after at L46

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: This is using boto3 resource abstraction instead of doing manual client calls and parsing.

samcli/commands/deploy/guided_context.py
CoshUS: Added

samcli/commands/deploy/guided_context.py
```diff
@@ -289,51 +292,244 @@ def prompt_parameters(
...
+            Dictionary of functions in the stack to be deployed with key as their logical ID.
+        """
+        for _, function_prop in functions.items():
+            if function_prop.packagetype != IMAGE:
+                continue
+            image = function_prop.imageuri
+            try:
+                tag_translation(image)
+            except NonLocalImageException:
+                pass
```
mgrandis, line 530: We should add a log debug here

samcli/lib/utils/managed_cloudformation_stack.py
mgrandis: We should maybe have a log.debug here as well.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        Blocking call to delete all deployed ECR repos that are unreferenced by a function
+        If repo does not exist, this will simply skip it.
+        """
+        repos = self.get_unreferenced_repos()
+        for repo in repos:
+            try:
+                self._ecr_client.delete_repository(repositoryName=repo.physical_id, force=True)
+            except self._ecr_client.exceptions.RepositoryNotFoundException:
+                pass
```
mgrandis, line 196: Here as well, a log debug.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        exists = self.does_companion_stack_exist()
+        has_repo = bool(self.get_repository_mapping())
+        if exists:
+            self.delete_unreferenced_repos()
+            if has_repo:
+                self.update_companion_stack()
+            else:
+                self.delete_companion_stack()
+        elif not exists and has_repo:
```
mgrandis, line 213: Not sure the `not exists` is necessary here.
Which means you can probably directly `if self.does_companion_stack_exist():`

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,314 @@
...
+        Returns
+        -------
+        bool
+            Returns True if companion stack exists
+        """
+        try:
+            self._cfn_client.describe_stacks(StackName=self._companion_stack.stack_name)
+            return True
+        except ClientError:
+            return False
```
mgrandis, line 229: Is there a way here to distinguish between a connection error or timeout (for which I think we could use a log debug) and the stack not found error (for which returning False is fine)?

tests/unit/lib/bootstrap/companion_stack/test_data_types.py
```diff
@@ -0,0 +1,63 @@
...
+from samcli.lib.bootstrap.companion_stack.companion_stack_builder import CompanionStackBuilder
+from unittest import TestCase
+from unittest.mock import Mock, patch
+
+
+class TestCompanionStack(TestCase):
+    def setUp(self):
+        self.check_sum = "checksum"
+        self.parent_stack_name = "Parent-Stack"
+        self.check_sum_patch = patch("samcli.lib.bootstrap.companion_stack.data_types.str_checksum")
```
mgrandis, line 11: Not sure this should be mocked.
If we were to change the checksum algorithm (`str_checksum` return a different value), I think we would like to see these tests fail to surface the change of name of our stacks.

Also, could you add some tests to verify the max-length cutting of logical id and checksum?

mgrandis (commented):
Looks good! Just a few minor changes

tests/unit/lib/bootstrap/companion_stack/test_data_types.py
CoshUS: Added tests for str_checksum and cutoff.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Added check for stack does not exist message. Throws the error out if it does not match.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Removed not exists and uses if self.does_companion_stack_exist():

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        Blocking call to delete all deployed ECR repos that are unreferenced by a function
+        If repo does not exist, this will simply skip it.
+        """
+        repos = self.get_unreferenced_repos()
+        for repo in repos:
+            try:
+                self._ecr_client.delete_repository(repositoryName=repo.physical_id, force=True)
+            except self._ecr_client.exceptions.RepositoryNotFoundException:
+                LOG.debug("Image repo not found in companion stack. Skipping deletion.")
```
mgrandis, line 199: We may want to put the repo physical_id in the debug message

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -97,12 +97,11 @@ def _create_or_get_stack(
         ds_resp = cloudformation_client.describe_stacks(StackName=stack_name)
         stacks = ds_resp["Stacks"]
         stack = stacks[0]
-        click.echo("\n\tLooking for resources needed for deployment: Found!")
         _check_sanity_of_stack(stack)
         stack_outputs = cast(List[Dict[str, str]], stack["Outputs"])
         return StackOutput(stack_outputs)
     except ClientError:
-        click.echo("\n\tLooking for resources needed for deployment: Not found.")
+        LOG.debug("Managed S3 stack not found. Creating a new one.")
```
mgrandis, line 104: Here also, maybe add the name of the stack?

mgrandis (commented):
Just 2 more minor changes

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
CoshUS: Added.

samcli/lib/utils/managed_cloudformation_stack.py
CoshUS: Added.

mgrandis (approved):

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,319 @@
...
+
+        waiter.wait(StackName=stack_name, WaiterConfig=self._update_stack_waiter_config)  # type: ignore
+
+    def delete_companion_stack(self):
+        """
+        Blocking call to delete the companion stack
+        """
+        stack_name = self._companion_stack.stack_name
+        waiter = self._cfn_client.get_waiter("stack_delete_complete")
+        self._cfn_client.delete_stack(StackName=stack_name)
```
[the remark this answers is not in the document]
CoshUS: The logic for handling deleting ECR repos are in sync_repos()

samcli/commands/deploy/command.py
```diff
@@ -289,13 +301,21 @@ def do_cli(
...
+            click.echo(f"\n\t\tManaged S3 bucket: {s3_bucket}")
+            click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
+            click.echo("\t\tOr by specifying --s3-bucket explicitly.")
+
+        # TODO Refactor resolve-s3 and resolve-image-repos into one place
+        # after we figure out how to enable resolve-images-repos in package
+        if resolve_image_repos:
+            image_repositories = sync_ecr_stack(
+                template_file, stack_name, region, s3_bucket, s3_prefix, image_repositories
+            )
```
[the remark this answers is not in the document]
CoshUS: Updated validation to include "image_repositories" as well in image_repository_validation.py

samcli/lib/bootstrap/companion_stack/companion_stack_builder.py
```diff
@@ -0,0 +1,130 @@
...
+
+    _parent_stack_name: str
+    _companion_stack: CompanionStack
+    _repo_mapping: Dict[str, ECRRepo]
+
+    def __init__(self, companion_stack: CompanionStack) -> None:
+        self._companion_stack = companion_stack
+        self._repo_mapping: Dict[str, ECRRepo] = dict()
+
+    def add_function(self, function_logical_id: str) -> None:
```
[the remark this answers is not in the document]
CoshUS: Function Logical IDs are from function.full_path which will include stack path.
Ex: "ChildStackA/GrandChildStackB/AFunctionInNestedStack"

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,319 @@
...
+        -------
+        bool
+            Returns True if companion stack exists
+        """
+        try:
+            self._cfn_client.describe_stacks(StackName=self._companion_stack.stack_name)
+            return True
+        except ClientError as e:
+            error_message = e.response.get("Error", {}).get("Message")
+            if error_message == f"Stack with id {self._companion_stack.stack_name} does not exist":
```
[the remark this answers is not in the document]
CoshUS: The error code is ValidationError which can be too generic and might be caused by other issues instead of stack not found.
All the data we have from ClientError is:
```
{
	"response": {
		"Error": {
			"Type": "Sender",
			"Code": "ValidationError",
			"Message": "Stack with id abcdefg-be64f26c-CompanionStack does not exist"
		},
		"ResponseMetadata": {
			"RequestId": "111111-2acf-45a7-98ff-11111111",
			"HTTPStatusCode": 400,
			"HTTPHeaders": {
				"x-amzn-requestid": "111111-2acf-45a7-98ff-11111111",
				"content-type": "text/xml",
				"content-length": "320",
				"date": "Tue, 10 Aug 2021 11:11:11 GMT",
				"connection": "close"
			},
			"RetryAttempts": 0
		}
	},
	"operation_name": "DescribeStacks"
}
```

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -0,0 +1,319 @@
...
+        repo: ECRRepo
+
+        Returns
+        -------
+        str
+            ECR repo URI based on account ID and region.
+        """
+        return repo.get_repo_uri(self._account_id, self._region_name)
+
+    def is_repo_uri(self, repo_uri: Optional[str], function_logical_id: str) -> bool:
```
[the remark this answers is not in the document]
CoshUS: Since these are methods of CompanionStackManager class, this context should be enough for it to not be used for generic ECR repo URIs.
My concern on adding companion stack to the method name is that we will need to add this for all other methods referring repos for consistency which can be too verbose.
```
comapnion_stack_manager.is_companion_repo_uri()
manager.get_companion_repos()
manager.get_unreferenced_companion_repos()
```

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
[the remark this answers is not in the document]
moelasmar: ok, it seems not possible .. thanks

moelasmar (approved):

samcli/lib/bootstrap/bootstrap.py
```diff
@@ -51,54 +51,47 @@ def get_current_account_id(profile: Optional[str] = None):
...
-      SourceBucket:
-        Value: !Ref SamCliSourceBucket
-    """
-
-    return template.format(info=json.dumps(info))
+    template = {
+        "AWSTemplateFormatVersion": "2010-09-09",
+        "Transform": "AWS::Serverless-2016-10-31",
+        "Description": "Managed Stack for AWS SAM CLI",
+        "Metadata": {
```
sriram-mv, line 58: Should we add the metadata ourselves in code, and move this entire template to its own file. We can then run cfn-lint as part of CI process on this template.

PR #2821, opened 2021-04-22
refactor: Extract git-clone functionality out of InitTemplates class

elbayaaa:
#1713 

This refactors allows the reuse of `git clone` functionality in both `sam init` and `sam pipeline init` 

It extracts the `git clone`logic to its own class

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -4,23 +4,18 @@ Manages the set of application templates.
 
 import itertools
 import json
-import os
 import logging
-import platform
-import shutil
-import subprocess
-
+import os
 from pathlib import Path
 from typing import Dict
 
 import click
 
 from samcli.cli.main import global_cfg
-from samcli.commands.exceptions import UserException, AppTemplateUpdateException
-from samcli.lib.utils import osutils
-from samcli.lib.utils.osutils import rmtree_callback
-from samcli.local.common.runtime_template import RUNTIME_DEP_TEMPLATE_MAPPING, get_local_lambda_images_location
+from samcli.commands.exceptions import UserException
+from samcli.lib.utils.git_repo import GitRepo, CloneRepoException, CloneRepoUnstableStateException
 from samcli.lib.utils.packagetype import IMAGE
+from samcli.local.common.runtime_template import RUNTIME_DEP_TEMPLATE_MAPPING, get_local_lambda_images_location
 
 LOG = logging.getLogger(__name__)
 
@@ -31,13 +26,12 @@ class InvalidInitTemplateError(UserException):
 
 class InitTemplates:
     def __init__(self, no_interactive=False, auto_clone=True):
-        self._repo_url = "https://github.com/aws/aws-sam-cli-app-templates"
-        self._repo_name = "aws-sam-cli-app-templates"
-        self._temp_repo_name = "TEMP-aws-sam-cli-app-templates"
-        self.repo_path = None
-        self.clone_attempted = False
         self._no_interactive = no_interactive
-        self._auto_clone = auto_clone
+        self._git_repo: GitRepo = GitRepo(
+            url="https://github.com/aws/aws-sam-cli-app-templates",
+            name="aws-sam-cli-app-templates",
+            auto_clone=auto_clone,
+        )
 
     def prompt_for_location(self, package_type, runtime, base_image, dependency_manager):
         """
@@ -89,7 +83,7 @@ class InitTemplates:
         if template_md.get("init_location") is not None:
             return (template_md["init_location"], template_md["appTemplate"])
         if template_md.get("directory") is not None:
-            return (os.path.join(self.repo_path, template_md["directory"]), template_md["appTemplate"])
+            return (os.path.join(self._git_repo.local_path, template_md["directory"]), template_md["appTemplate"])
         raise InvalidInitTemplateError("Invalid template. This should not be possible, please raise an issue.")
 
     def location_from_app_template(self, package_type, runtime, base_image, dependency_manager, app_template):
@@ -99,7 +93,7 @@ class InitTemplates:
             if template.get("init_location") is not None:
                 return template["init_location"]
             if template.get("directory") is not None:
-                return os.path.join(self.repo_path, template["directory"])
+                return os.path.join(self._git_repo.local_path, template["directory"])
             raise InvalidInitTemplateError("Invalid template. This should not be possible, please raise an issue.")
         except StopIteration as ex:
             msg = "Can't find application template " + app_template + " - check valid values in interactive init."
@@ -112,14 +106,23 @@ class InitTemplates:
         return bool(entry["appTemplate"] == app_template)
 
     def init_options(self, package_type, runtime, base_image, dependency_manager):
-        if not self.clone_attempted:
-            self._clone_repo()
-        if self.repo_path is None:
+        if not self._git_repo.clone_attempted:
+            shared_dir: Path = global_cfg.config_dir
+            try:
+                self._git_repo.clone(clone_dir=shared_dir, replace_existing=True)
+            except CloneRepoUnstableStateException as ex:
+                raise UserException(str(ex)) from ex
+            except (OSError, CloneRepoException):
+                # If can't clone, try using an old clone from a previous run if already exist
+                expected_previous_clone_local_path: Path = shared_dir.joinpath(self._git_repo.name)
+                if expected_previous_clone_local_path.exists():
+                    self._git_repo.local_path = expected_previous_clone_local_path
+        if self._git_repo.local_path is None:
             return self._init_options_from_bundle(package_type, runtime, dependency_manager)
         return self._init_options_from_manifest(package_type, runtime, base_image, dependency_manager)
 
     def _init_options_from_manifest(self, package_type, runtime, base_image, dependency_manager):
-        manifest_path = os.path.join(self.repo_path, "manifest.json")
+        manifest_path = os.path.join(self._git_repo.local_path, "manifest.json")
         with open(str(manifest_path)) as fp:
             body = fp.read()
             manifest_body = json.loads(body)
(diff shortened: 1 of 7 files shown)


samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,188 @@
...
+        OSError:
+            when file management errors like unable to mkdir, copytree, rmtree ...etc
+        CloneRepoException:
+            General errors like for example; if an error occurred while running `git clone`
+            or if the local_clone already exists and replace_existing is not set
+        CloneRepoUnstableStateException:
+            when reaching unstable state, for example with replace_existing flag set, unstable state can happen
+            if removed the current local clone but failed to copy the new one from the temp location to the destination
+        """
+        if not self._auto_clone:  # Unit test escape hatch
```
mgrandis, line 140: If I initialize the GitRepo with `auto_clone` at `False`, I can never clone?

mgrandis: Not sure the `auto_clone` property is meant to be inside the class.
If so, I would expect it to control if the `__init__` method is launching a clone.

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,188 @@
...
+                raise
+            except subprocess.CalledProcessError as clone_error:
+                output = clone_error.output.decode("utf-8")
+                if "not found" in output.lower():
+                    LOG.warning("WARN: Could not clone repo %s", self._url, exc_info=clone_error)
+                raise CloneRepoException from clone_error
+            finally:
+                self._clone_attempted = True
+
+    @staticmethod
```
mgrandis, line 168: Is this really static?
It's private and only called in a non static way from the clone method which is not static

samcli/commands/init/init_templates.py
```diff
@@ -112,14 +106,23 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
...
+            shared_dir: Path = global_cfg.config_dir
+            try:
+                self._git_repo.clone(clone_dir=shared_dir, replace_existing=True)
+            except CloneRepoUnstableStateException as ex:
+                raise UserException(str(ex)) from ex
+            except (OSError, CloneRepoException):
+                # If can't clone, try using an old clone from a previous run if already exist
+                expected_previous_clone_local_path: Path = shared_dir.joinpath(self._git_repo.name)
+                if expected_previous_clone_local_path.exists():
+                    self._git_repo.local_path = expected_previous_clone_local_path
```
hoffa, line 119: Where is the corresponding logic in the old code?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,188 @@
...
+    def _persist_local_repo(temp_path: str, dest_dir: Path, dest_name: str, replace_existing: bool) -> Path:
+        dest_path = os.path.normpath(dest_dir.joinpath(dest_name))
+        try:
+            if Path(dest_path).exists():
```
mgrandis, line 172: Could we optimize this by checking this before cloning and directly cloning into the right directory instead of copying every time?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,188 @@
...
+            LOG.debug("Copying from %s to %s", temp_path, dest_path)
+            shutil.copytree(temp_path, dest_path, ignore=shutil.ignore_patterns("*.git"))
+            return Path(dest_path)
+        except (OSError, shutil.Error) as ex:
+            # UNSTABLE STATE
+            # it's difficult to see how this scenario could happen except weird permissions, user will need to debug
+            raise CloneRepoUnstableStateException(
+                "Unstable state when updating repo. "
+                f"Check that you have permissions to create/delete files in {dest_dir} directory "
+                "or file an issue at https://github.com/awslabs/aws-sam-cli/issues"
```
mgrandis, line 187: We may want to remove this line as we are not going to put it into all our error messages

samcli/lib/utils/git_repo.py
mgrandis: Ok, we may want to keep the old clone if the new clone fails I guess

samcli/commands/init/init_templates.py
elbayaaa: https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/init/init_templates.py#L183

samcli/lib/utils/git_repo.py
elbayaaa: As I understood from the code, `auto_clone` is just a flag used in the unit-testing to turn-off cloning. check in `tests/unit/commands/init/test_cli.py` for example [this line](https://github.com/aws/aws-sam-cli/blob/13a4250c137ce1f9e0087b7a17cca434b8c32bff/tests/unit/commands/init/test_cli.py#L66)  

in reality the `cli` method of `sam init` always call `do_cli` with `auto_clone` set to true see [here](https://github.com/aws/aws-sam-cli/blob/c99f1eee1ef4b23ffffde85e6c5f669b3970f083/samcli/commands/init/__init__.py#L259)

samcli/lib/utils/git_repo.py
elbayaaa: yes it is static; it has no dependency on *self*, and logically, it just moves a directory from one location to another(with replace existing) so seems to be really static.

what do you mean by "called in a non static way"? does this mean `self. _persist_local_repo(...)` instead of `GitRepo._persist_local_repo(...)` ?

samcli/lib/utils/git_repo.py
elbayaaa: This is the same as before, I didn't change it.

hoffa (approved):

samcli/lib/utils/git_repo.py
elbayaaa: This one is re-raised as a UserException (click.ClickException) and the message is displayed to the user

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+            try:
+                subprocess.Popen([executable], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                # No exception. Let's pick this
+                return executable
+            except OSError as ex:
+                LOG.debug("Unable to find executable %s", executable, exc_info=ex)
+
+        raise OSError("Cannot find git, was looking at executables: {}".format(executables))
+
+    def clone(self, clone_dir: Path, clone_name: Optional[str] = None, replace_existing: bool = False) -> Path:
```
hoffa, line 105: If the sole purpose is to clone a Git repository, what is the reasoning used for which arguments are in the constructor, and which are in `clone`?

samcli/commands/init/init_templates.py
```diff
@@ -112,14 +105,23 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
...
+            shared_dir: Path = global_cfg.config_dir
+            try:
+                self._git_repo.clone(clone_dir=shared_dir, replace_existing=True)
+            except CloneRepoUnstableStateException as ex:
+                raise AppTemplateUpdateException(str(ex)) from ex
+            except (OSError, CloneRepoException):
+                # If can't clone, try using an old clone from a previous run if already exist
+                expected_previous_clone_local_path: Path = shared_dir.joinpath(self._git_repo.name)
+                if expected_previous_clone_local_path.exists():
+                    self._git_repo.local_path = expected_previous_clone_local_path
```
aahung, line 118: this code code should be moved into `git_repo.py` because it does not seem to depend on anything in this class.

samcli/lib/utils/git_repo.py
elbayaaa: my intention is to make `clone` the initial functionality but not the only one. that is why in the docstring I added the word "currently":

> Class for managing a Git repo, **currently** it has a clone functionality only

aahung: Why there is a `clone_name`, the only usage is `self._git_repo.clone(clone_dir=shared_dir, replace_existing=True)` and it does not give `clone_name` any value.

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+        clone_dir: Path
+            The directory to create the local clone inside
+        clone_name: Optional[str]
+            The dirname of the local clone, if not provided, repository's name (self._name) will be used
+        replace_existing: bool
+            Whether to replace the current local clone directory if already exists or not
+
+        Returns
+        -------
+            The path of the created local clone
```
aahung, line 120: from the usage, it does not need to return anything from this method

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+                self._local_path = self._persist_local_repo(temp_path, clone_dir, clone_name, replace_existing)
+                return self._local_path
+            except OSError as ex:
+                LOG.warning("WARN: Could not clone repo %s", self._url, exc_info=ex)
+                raise
+            except subprocess.CalledProcessError as clone_error:
+                output = clone_error.output.decode("utf-8")
+                if "not found" in output.lower():
+                    LOG.warning("WARN: Could not clone repo %s", self._url, exc_info=clone_error)
+                raise CloneRepoException from clone_error
```
aahung, line 156: if we handle OSError and CloneRepoException in the same way, we should just raise CloneRepoException in both `except`

samcli/commands/init/init_templates.py
```diff
@@ -112,14 +105,23 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
         return bool(entry["appTemplate"] == app_template)
 
     def init_options(self, package_type, runtime, base_image, dependency_manager):
-        if not self.clone_attempted:
-            self._clone_repo()
-        if self.repo_path is None:
+        if not self._git_repo.clone_attempted:
+            shared_dir: Path = global_cfg.config_dir
```
aahung, line 109: you use `clone_dir` in `git_repo`, maybe we should use `clone_dir` to make it consistent?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+            LOG.debug("Copying from %s to %s", temp_path, dest_path)
+            shutil.copytree(temp_path, dest_path, ignore=shutil.ignore_patterns("*.git"))
+            return Path(dest_path)
+        except (OSError, shutil.Error) as ex:
+            # UNSTABLE STATE
+            # it's difficult to see how this scenario could happen except weird permissions, user will need to debug
+            raise CloneRepoUnstableStateException(
+                "Unstable state when updating repo. "
+                f"Check that you have permissions to create/delete files in {dest_dir} directory "
+                "or file an issue at https://github.com/awslabs/aws-sam-cli/issues"
```
aahung, line 179: the github link is outdated

samcli/commands/init/init_templates.py
```diff
@@ -112,14 +105,23 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
...
+            try:
+                self._git_repo.clone(clone_dir=shared_dir, replace_existing=True)
+            except CloneRepoUnstableStateException as ex:
+                raise AppTemplateUpdateException(str(ex)) from ex
+            except (OSError, CloneRepoException):
```
aahung, line 114: `CloneRepoException` can be also raised if `replace_existing=False` and `Path(dest_path).exists()`, which does not seem to match the error handling behavior here.

samcli/lib/utils/git_repo.py
aahung: `clone_name` is unnecessary. It is being used at all.

samcli/commands/init/init_templates.py
```diff
@@ -31,13 +26,12 @@ class InvalidInitTemplateError(UserException):
 
 class InitTemplates:
     def __init__(self, no_interactive=False, auto_clone=True):
-        self._repo_url = "https://github.com/aws/aws-sam-cli-app-templates"
-        self._repo_name = "aws-sam-cli-app-templates"
-        self._temp_repo_name = "TEMP-aws-sam-cli-app-templates"
```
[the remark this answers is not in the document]
elbayaaa: Apparently this wasn't used anywhere, but the clone method itself has the parameter `clone_name` that do the same logic behind this field; cloning the repository under a different name

samcli/commands/init/init_templates.py
elbayaaa: it has a different meaning in each context:
clone-method (abstract meaning): the location to clone into
here: we are cloning SAM's templates into SAM's shared directory, a directory that is shared between all of the generated SAM applications.

samcli/commands/init/init_templates.py
elbayaaa: this call has replace_existing set to True

samcli/commands/init/init_templates.py
elbayaaa: No, this is an `InitTemplates` behavior to use existing old clone if failed to make a new clone. Another caller may choose to raise or handle it however

samcli/lib/utils/git_repo.py
hoffa: I still don't understand what value all those names bring. A Git directory (URL) is cloned to a directory (name). What other "name" is needed?

samcli/commands/init/init_templates.py
aahung: Oh I see, when is `replace_existing` set to `False`?

samcli/lib/utils/git_repo.py
elbayaaa: It make sense to return this path. 

why in `InitTemplates` we read from `local_path` instead? because the behavior of `InitTemplates` is to not attempt clone if already did before

but generally:
```
local_clone = git_repo.clone(....)
```

is better than
```
git_repo.clone(...)
local_clone = git_repo.local_path
```

samcli/commands/init/init_templates.py
elbayaaa: GitRepo doesn't care about this. it is a case by case thing

samcli/lib/utils/git_repo.py
hoffa: Actually it's more confusing on second look. There's `_name`, `_local_path`, `clone_name` and `clone_dir`.

`git clone` takes a single optional name as argument (the clone directory); why is there so much complexity here? I understand this is mostly existing logic, but it feels like a great potential win to significantly simplify this.

The whole `_persist_local_repo` logic seems unnecessary as well. Why not directly clone to the target directory?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+class CloneRepoException(Exception):
+    """
+    Exception class when clone repo fails.
+    """
+
+
+class CloneRepoUnstableStateException(CloneRepoException):
+    """
+    Exception class when clone repo enters an unstable state.
+    """
```
aahung, line 26: If we want to make this `GitRepo` generic, we should provide more info about exceptions.
What does "unstable" mean here? what scenarios it cover? Does it mean "unexpected?" I think we need to add more details in this docstring.

samcli/lib/utils/git_repo.py
aahung: If that's the case, why we need to have `git_repo.local_path`?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+    def local_path(self, value: Path) -> None:
+        self._local_path = value
+
+    @property
+    def clone_attempted(self) -> bool:
+        return self._clone_attempted
+
+    @clone_attempted.setter
+    def clone_attempted(self, value: bool) -> None:
+        self._clone_attempted = value
```
aahung, line 78: this getter and setter do not do anything special, we can remove then and make `_clone_attempted` -> `clone_attempted`

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def local_path(self) -> Optional[Path]:
+        return self._local_path
+
+    @local_path.setter
+    def local_path(self, value: Path) -> None:
+        self._local_path = value
```
aahung, line 70: this getter and setter do not do anything special, we can remove then and make _local_path -> local_path

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+    _local_path: Path
+        The path of the local clone of this Git repository
+    _clone_attempted: bool
+        whether an attempt to clone this Git repository took place or not
+
+    Methods
+    -------
+    clone(self, clone_dir: Path, clone_name=None, replace_existing=False) -> Path:
+        creates a local clone of this Git repository
+    """
```
aahung, line 48: This class is pretty hard to understand, can we add more details in docstring about what it does and why we need `_local_path` and `_clone_attempted`. I think I sort of get the idea of "avoid cloning multiple times" but it is not clear about what roles these two attributes are playing.

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+    def clone_attempted(self, value: bool) -> None:
+        self._clone_attempted = value
+
+    @staticmethod
+    def _ensure_clone_directory_exists(clone_dir: Path) -> None:
+        try:
+            clone_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
+        except OSError as ex:
+            LOG.warning("WARN: Unable to create clone directory.", exc_info=ex)
+            raise
```
aahung, line 86: Is this a unstable error?

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+
+        for executable in executables:
+            try:
+                subprocess.Popen([executable], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                # No exception. Let's pick this
+                return executable
+            except OSError as ex:
+                LOG.debug("Unable to find executable %s", executable, exc_info=ex)
+
+        raise OSError("Cannot find git, was looking at executables: {}".format(executables))
```
aahung, line 103: `CloneRepoException` feels more suitable here compared to `OSError`.

samcli/lib/utils/git_repo.py
```diff
@@ -0,0 +1,180 @@
...
+            Whether to replace the current local clone directory if already exists or not
+
+        Returns
+        -------
+            The path of the created local clone
+
+        Raises
+        ------
+        OSError:
+            when file management errors like unable to mkdir, copytree, rmtree ...etc
```
aahung, line 125: In `_persist_local_repo` you package `OSError` (copytree?) into `CloneRepoUnstableStateException`. Can we make exception raising consistent?

samcli/lib/utils/git_repo.py
elbayaaa: `persist_local_repo` is to move the created clone from temp to its final destination. It should be clear why it is created at temp in the first place, right?


why are you mixing method parameters; `clone_name` and `clone_dir` with class attributes?
For the class attributes:
* A GitRepo **has a** _url  # example: "https://github.com/aws/aws-sam-cli-app-pipeline-templates"
* A GitRepo **has a** _name # example "aws-sam-cli-app-pipeline-templates"
* A GitRepo **has a** _local_path # example "~/.aws-sam/aws-sam-cli-app-pipeline-templates"

* _local_path is set only after clone as it is derived from clone's destination `{clone_dir}/{clone_name}`

samcli/commands/init/init_templates.py
```diff
@@ -112,14 +105,23 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
         return bool(entry["appTemplate"] == app_template)
 
     def init_options(self, package_type, runtime, base_image, dependency_manager):
-        if not self.clone_attempted:
-            self._clone_repo()
-        if self.repo_path is None:
+        if not self._git_repo.clone_attempted:
```
aahung, line 108: Looks like `clone_attempted` is never used inside `GitRepo`, I think it should be moved out and stored here instead

aahung (approved):

mgrandis (approved):

PR #2826, opened 2021-04-23
docs: add tiny tutorial

hoffa:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -21,6 +21,41 @@ This will start a new Gitpod workspace, and immediately kick off a build of the 
 
 Gitpod is free for 50 hours per month - make sure to stop your workspace when you're done (you can always resume it later, and it won't need to run the build again).
 
+5-Minute Tutorial
+-----------------
+
+This tiny tutorial will show you how to make a small change to SAM CLI.
+
+See the rest of this document for a more thorough introduction.
+
+1. Clone SAM CLI:
+
+    ```bash
+    git clone git@github.com:aws/aws-sam-cli.git
+    cd aws-sam-cli
+    ```
+
+2. Create a Python [virtual environment](https://docs.python.org/3/library/venv.html):
+
+    ```bash
+    python3 -m venv .venv
+    source .venv/bin/activate
+    ```
+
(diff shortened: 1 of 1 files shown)


DEVELOPMENT_GUIDE.md
```diff
@@ -21,6 +21,41 @@ This will start a new Gitpod workspace, and immediately kick off a build of the
...
+-----------------
+
+This tiny tutorial will show you how to make a small change to SAM CLI.
+
+See the rest of this document for a more thorough introduction.
+
+1. Clone SAM CLI:
+
+    ```bash
+    git clone git@github.com:aws/aws-sam-cli.git
```
xazhao, line 34: Should we do git clone using `https` instead of `ssh`?

DEVELOPMENT_GUIDE.md
```diff
@@ -21,6 +21,41 @@ This will start a new Gitpod workspace, and immediately kick off a build of the
...
+    cd aws-sam-cli
+    ```
+
+2. Create a Python [virtual environment](https://docs.python.org/3/library/venv.html):
+
+    ```bash
+    python3 -m venv .venv
```
xazhao, line 41: I feel it should be consistent using `venv` or `pyenv`. If we want to use `venv` here, we might also want to change all `pyenv` to `venv` in this doc.

DEVELOPMENT_GUIDE.md
hoffa: It's using the Git protocol, which IIRC authenticates properly if the credentials are available. It also work without any write permissions.

prenx4x: i prefer ssh instead of https because of convenience of not putting passwords everytime.

DEVELOPMENT_GUIDE.md
hoffa: `pyenv` requires additional dependencies and setup for more complex use cases (e.g. quickly switching between Python versions), which are unnecessary to quickly get started on any system.

DEVELOPMENT_GUIDE.md
xazhao: Interesting... I don't need to put password for https and it works out of box. But for ssh, I need to set up a key. Not sure if it's related to my setup. As long as it's new hire friendly, I'm good either way.

hoffa:
Discussed with @aahung, whole development guide should be overhauled as it's quite unclear and could be better organized. Closing for now.

PR #2831, opened 2021-04-27
sam pipeline init command

elbayaaa:
#1713 

part of sam-pipelines feature

by implementing new command`sam pipeline init` that generates the pipeline configuration files

N/A - new feature that doesn't affect existing code

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/pipeline/init/cli.py
+++ b/samcli/commands/pipeline/init/cli.py
@@ -10,9 +10,10 @@ from samcli.cli.main import pass_context, common_options as cli_framework_option
 from samcli.commands.pipeline.init.interactive_init_flow import do_interactive
 from samcli.lib.telemetry.metric import track_command
 
-SHORT_HELP = "Generates a CI/CD pipeline configuration file."
+SHORT_HELP = "Generates CI/CD pipeline configuration files."
 HELP_TEXT = """
-Generates a CI/CD pipeline configuration file for a chosen CI/CD provider like Jenkins, Gitlab and Github Actions
+Generates CI/CD pipeline configuration files for a chosen CI/CD provider such as Jenkins, 
+GitLab CI/CD or GitHub Actions
 """
 
 
--- a/samcli/commands/pipeline/init/cli.py
+++ b/samcli/commands/pipeline/init/cli.py
@@ -1,0 +1,43 @@
+"""
+CLI command for "pipeline init" command
+"""
+from typing import Any, Optional
+
+import click
+
+from samcli.cli.cli_config_file import configuration_option, TomlProvider
+from samcli.cli.main import pass_context, common_options as cli_framework_options
+from samcli.commands.pipeline.init.interactive_init_flow import do_interactive
+from samcli.lib.telemetry.metric import track_command
+
+SHORT_HELP = "Generates a CI/CD pipeline configuration file."
+HELP_TEXT = """
+Generates a CI/CD pipeline configuration file for a chosen CI/CD provider like Jenkins, Gitlab and Github Actions
+"""
+
+
+@click.command("init", help=HELP_TEXT, short_help=SHORT_HELP)
+@configuration_option(provider=TomlProvider(section="parameters"))
+@cli_framework_options
+@pass_context
+@track_command  # pylint: disable=R0914
+def cli(
+    ctx: Any,
+    config_env: Optional[str],
+    config_file: Optional[str],
+) -> None:
+    """
+    `sam pipeline init` command entry point
+    """
+
+    # Currently we support interactive mode only, i.e. the user doesn't provide the required arguments during the call
+    # so we call do_cli without any arguments. This will change after supporting the non interactive mode.
+    do_cli()
+
+
+def do_cli() -> None:
+    """
+    implementation of `sam pipeline init` command
+    """
+    # TODO non-interactive mode
+    do_interactive()
--- a/samcli/commands/pipeline/init/interactive_init_flow.py
+++ b/samcli/commands/pipeline/init/interactive_init_flow.py
@@ -1,0 +1,244 @@
+"""
+Interactive flow that prompts that users for pipeline template (cookiecutter template) and used it to generate
+pipeline configuration file
+"""
+import logging
+import os
+import shutil
+from pathlib import Path
+from typing import Dict, List
+
+import click
+
+from samcli.cli.main import global_cfg
+from samcli.commands.exceptions import PipelineTemplateCloneException
+from samcli.lib.cookiecutter.interactive_flow import InteractiveFlow
+from samcli.lib.cookiecutter.interactive_flow_creator import InteractiveFlowCreator
+from samcli.lib.cookiecutter.template import Template
+from samcli.lib.utils.git_repo import GitRepo, CloneRepoException
+from samcli.lib.utils.osutils import rmtree_callback
+from .pipeline_templates_manifest import PipelineTemplateManifest, PipelineTemplatesManifest
+
+LOG = logging.getLogger(__name__)
+shared_path: Path = global_cfg.config_dir
+APP_PIPELINE_TEMPLATES_REPO_URL = "https://github.com/aws/aws-sam-cli-pipeline-init-templates.git"
+APP_PIPELINE_TEMPLATES_REPO_LOCAL_NAME = "aws-sam-cli-app-pipeline-templates"
+CUSTOM_PIPELINE_TEMPLATE_REPO_LOCAL_NAME = "custom-pipeline-template"
+
+
+def do_interactive() -> None:
+    """
+    An interactive flow that prompts the user for pipeline template (cookiecutter template) location, downloads it,
+    runs its specific questionnaire then generates the pipeline config file based on the template and user's responses
+    """
+    click.echo("Which pipeline template source would you like to use?")
+    click.echo("\t1 - AWS Quick Start Pipeline Templates\n\t2 - Custom Pipeline Template Location")
+    location_choice = click.prompt("Choice", type=click.Choice(["1", "2"]), show_choices=False)
+    if location_choice == "2":
+        _generate_from_custom_location()
+    else:
+        _generate_from_app_pipeline_templates()
+
+
+def _generate_from_app_pipeline_templates() -> None:
+    """
+    Prompts the user to choose a pipeline template from SAM predefined set of pipeline templates hosted in the git
+    repository: aws/aws-sam-cli-pipeline-init-templates.git
+    downloads locally, then generates the pipeline config file from the selected pipeline template.
+    """
+    pipeline_templates_local_dir: Path = _clone_app_pipeline_templates()
+
+    pipeline_templates_manifest: PipelineTemplatesManifest = _read_app_pipeline_templates_manifest(
+        pipeline_templates_dir=pipeline_templates_local_dir
+    )
+    # The manifest contains multiple pipeline-templates so select one
+    selected_pipeline_template_manifest: PipelineTemplateManifest = _select_pipeline_template(
+        pipeline_templates_manifest=pipeline_templates_manifest
+    )
+    selected_pipeline_template_dir: Path = pipeline_templates_local_dir.joinpath(
+        selected_pipeline_template_manifest.location
+    )
+    _generate_from_pipeline_template(pipeline_template_dir=selected_pipeline_template_dir)
+
+
+def _generate_from_custom_location() -> None:
+    """
+    Prompts the user for a custom pipeline template location, downloads locally, then generates the pipeline config file
+    """
+    pipeline_template_repo_url: str = click.prompt(text="template Git location")
+    pipeline_template_local_dir: Path = _clone_custom_pipeline_template(repo_url=pipeline_template_repo_url)
+    _generate_from_pipeline_template(pipeline_template_dir=pipeline_template_local_dir)
+    # Unlike app pipeline templates, custom pipeline templates are not shared between different SAM applications
+    # and should be cleaned up from users' machines after generating the pipeline config files.
+    shutil.rmtree(pipeline_template_local_dir, onerror=rmtree_callback)
+
+
+def _generate_from_pipeline_template(pipeline_template_dir: Path) -> None:
+    """
+    Generates a pipeline config file from a given pipeline template local location
+    """
+    pipeline_template: Template = _initialize_pipeline_template(pipeline_template_dir=pipeline_template_dir)
+    context: Dict = pipeline_template.run_interactive_flows()
+    pipeline_template.generate_project(context)
+
+
+def _clone_app_pipeline_templates() -> Path:
+    """
+    clone aws/aws-sam-cli-pipeline-init-templates.git Git repo to the local machine in SAM shared directory.
+    Returns:
+        the local directory path where the repo is cloned.
+    """
+    try:
+        return _clone_pipeline_templates(
+            repo_url=APP_PIPELINE_TEMPLATES_REPO_URL, clone_name=APP_PIPELINE_TEMPLATES_REPO_LOCAL_NAME
+        )
+    except PipelineTemplateCloneException:
+        # If can't clone app pipeline templates, try using an old clone from a previous run if already exist
+        expected_previous_clone_local_path: Path = shared_path.joinpath(APP_PIPELINE_TEMPLATES_REPO_LOCAL_NAME)
+        if expected_previous_clone_local_path.exists():
+            click.echo("Unable to download updated app pipeline templates, using existing ones")
+            return expected_previous_clone_local_path
+        raise
+
+
+def _clone_custom_pipeline_template(repo_url: str) -> Path:
+    """
+    clone a given Git pipeline template's Git repo to the user machine in SAM shared directory.
+    Returns: the local directory path where the repo is cloned to.
+
+    Parameters:
+        repo_url: the URL of the Git repo to clone
+
+    Returns:
+        the local directory path where the repo is cloned.
+    """
+    return _clone_pipeline_templates(repo_url=repo_url, clone_name=CUSTOM_PIPELINE_TEMPLATE_REPO_LOCAL_NAME)
+
+
+def _clone_pipeline_templates(repo_url: str, clone_name: str) -> Path:
+    """
+    clone a given pipeline templates' Git repo to the user machine inside the SAM shared directory(default: ~/.aws-sam)
+    under the given clone name. For example, if clone_name is "custom-pipeline-template" then the location to clone
+    to is "~/.aws-sam/custom-pipeline-template/"
+
+    Parameters:
+        repo_url: the URL of the Git repo to clone
+        clone_name: The folder name to give to the created clone
+
+    Returns:
+        Path to the local clone
+    """
+    try:
+        repo: GitRepo = GitRepo(url=repo_url)
+        clone_path: Path = repo.clone(clone_dir=shared_path, clone_name=clone_name, replace_existing=True)
+        return clone_path
+    except (OSError, CloneRepoException) as ex:
+        raise PipelineTemplateCloneException(str(ex)) from ex
+
+
+def _read_app_pipeline_templates_manifest(pipeline_templates_dir: Path) -> PipelineTemplatesManifest:
+    """
+    parse and return the manifest yaml file located in the root directory of the SAM pipeline templates folder:
+
+    Parameters:
+        pipeline_templates_dir: local directory of SAM pipeline templates
+
+    Raises:
+        AppPipelineTemplateManifestException if the manifest is not found, ill-formatted or missing required keys
+
+    Returns:
+        The manifest of the pipeline templates
+    """
+    manifest_path: Path = Path(os.path.normpath(pipeline_templates_dir.joinpath("manifest.yaml")))
+    return PipelineTemplatesManifest(manifest_path=manifest_path)
+
+
+def _select_pipeline_template(pipeline_templates_manifest: PipelineTemplatesManifest) -> PipelineTemplateManifest:
+    """
+    Prompts the user a list of the available CI/CD providers along with associated app pipeline templates to choose
+    one of them
+
+    Parameters:
+        pipeline_templates_manifest: A manifest file lists the available providers and the associated pipeline templates
+
+    Returns:
+         The manifest (A section in the pipeline_templates_manifest) of the chosen pipeline template;
+    """
+    provider = _prompt_for_cicd_provider(pipeline_templates_manifest.providers)
+    provider_pipeline_templates: List[PipelineTemplateManifest] = list(
+        filter(lambda t: t.provider == provider, pipeline_templates_manifest.templates)
+    )
+    selected_template_manifest: PipelineTemplateManifest = _prompt_for_pipeline_template(provider_pipeline_templates)
+    return selected_template_manifest
+
+
+def _prompt_for_cicd_provider(available_providers: List[str]) -> str:
+    """
+    Prompts the user a list of the available CI/CD providers to choose from
+
+    Parameters:
+        available_providers: List of available CI/CD providers like Jenkins, Gitlab and CircleCI
+
+    Returns:
+        The chosen provider
+    """
+    choices = list(map(str, range(1, len(available_providers) + 1)))
+    click.echo("CICD provider")
+    for index, provider in enumerate(available_providers):
+        click.echo(message=f"\t{index + 1} - {provider}")
+    choice = click.prompt(text="Choice", show_choices=False, type=click.Choice(choices))
+    return available_providers[int(choice) - 1]
+
+
+def _prompt_for_pipeline_template(
+    available_pipeline_templates_manifests: List[PipelineTemplateManifest],
+) -> PipelineTemplateManifest:
+    """
+    Prompts the user a list of the available pipeline templates to choose from
+
+    Parameters:
+        available_pipeline_templates_manifests: List of available pipeline templates manifests
+
+    Returns:
+        The chosen pipeline template manifest
+    """
+    choices = list(map(str, range(1, len(available_pipeline_templates_manifests) + 1)))
+    click.echo("Which pipeline template would you like to use?")
+    for index, template in enumerate(available_pipeline_templates_manifests):
+        click.echo(f"\t{index + 1} - {template.name}")
+    choice = click.prompt(text="Choice", show_choices=False, type=click.Choice(choices))
+    return available_pipeline_templates_manifests[int(choice) - 1]
+
+
+def _initialize_pipeline_template(pipeline_template_dir: Path) -> Template:
+    """
+    Initialize a pipeline template from a given pipeline template (cookiecutter template) location
+
+    Parameters:
+        pipeline_template_dir: The local location of the pipeline cookiecutter template
+
+    Returns:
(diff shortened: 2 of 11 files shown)


samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+    An interactive flow that prompts the user for pipeline template (cookiecutter template) location, downloads it,
+    runs its specific questionnaire then generates the pipeline config file based on the template and user's responses
+    """
+    click.echo("Which pipeline template source would you like to use?")
+    click.echo("\t1 - AWS Quick Start Pipeline Templates\n\t2 - Custom Pipeline Template Location")
+    location_choice = click.prompt("Choice", type=click.Choice(["1", "2"]), show_choices=False)
+    if location_choice == "2":
+        _generate_from_custom_location()
+    else:
+        _generate_from_app_pipeline_templates()
```
hoffa, line 40: Can we abstract this away? Showing a bunch of choices and calling corresponding callbacks is generic and well-defined. I noticed this is also being used in #2811, but it's fragile (need to ensure the choices and prompts match) and hard to read (coupled stuff like the choice value, expected response and callback are tightly coupled but scattered).

samcli/commands/pipeline/init/cli.py
```diff
@@ -0,0 +1,43 @@
...
+def cli(
+    ctx: Any,
+    config_env: Optional[str],
+    config_file: Optional[str],
+) -> None:
+    """
+    `sam pipeline init` command entry point
+    """
+
+    # Currently we support interactive mode only, i.e. the user doesn't provide the required arguments during the call
```
hoffa, line 33: Are there any challenges to supporting non-interactive mode?

elbayaaa: We discussed before and agreed that it won't be included in the first run. Actually, we are quite unsure about the use-case. We believe this command will mainly be used in an interactive way. One use-case for the non-interactive mode is the IDEs if we are integrating with it, but this integration is not discussed.

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: do you mean not depending on the fact that click is guaranteed to return either "1" or "2" in this case? if so, yes we can but this just adds one more unnecessarily branch.

hoffa: I mean it's too verbose and error-prone.

We can observe that the multiple-choice prompts always (1) print the choices in a specific way (ordered list with tabs and `-`) (2) using specific inputs (a numbered list). Having to read and write this every time and ensuring it's consistent is too much cognitive complexity.

Instead we can refactor into something like this:

```python
def prompt_choices(choices: List[str]) -> int:
    """Prompt choice and return its index."""
    for i, choice in enumerate(choices):
        click.echo(f"\t{i + 1} - {choice}")
    inputs = [str(i + 1) for i in range(len(choices))]
    input = click.prompt("Choice", type=click.Choice(inputs), show_choices=False)
    return int(input) - 1
```

This way,

```python
    click.echo("\t1 - AWS Quick Start Pipeline Templates\n\t2 - Custom Pipeline Template Location")
    location_choice = click.prompt("Choice", type=click.Choice(["1", "2"]), show_choices=False)
    if location_choice == "2":
        _generate_from_custom_location()
    else:
        _generate_from_app_pipeline_templates()
```
 
becomes

```python
    location_choice = prompt_choice([
        "AWS Quick Start Pipeline Templates",
        "Custom Pipeline Template Location",
    ])
    if location_choice == 0:
        _generate_from_app_pipeline_templates()
    else:
        _generate_from_custom_location()
```

And this (from below)

```python
choices = list(map(str, range(1, len(available_pipeline_templates_manifests) + 1)))
for index, template in enumerate(available_pipeline_templates_manifests):
    click.echo(f"\t{index + 1} - {template.name}")
choice = click.prompt(text="Choice", show_choices=False, type=click.Choice(choices))
```

becomes

```python
choice = prompt_choice(available_pipeline_templates_manifests)
```

samcli/commands/pipeline/init/cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+from samcli.cli.cli_config_file import configuration_option, TomlProvider
+from samcli.cli.main import pass_context, common_options as cli_framework_options
+from samcli.commands.pipeline.init.interactive_init_flow import do_interactive
+from samcli.lib.telemetry.metric import track_command
+
+SHORT_HELP = "Generates a CI/CD pipeline configuration file."
+HELP_TEXT = """
+Generates a CI/CD pipeline configuration file for a chosen CI/CD provider like Jenkins, Gitlab and Github Actions
+"""
```
hoffa, line 16: Could be multiple files:

```suggestion
SHORT_HELP = "Generates CI/CD pipeline configuration files."
HELP_TEXT = """
Generates CI/CD pipeline configuration files for a chosen CI/CD provider such as Jenkins, GitLab CI/CD or GitHub Actions.
"""
```

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _generate_from_app_pipeline_templates() -> None:
+    """
+    Prompts the user to choose a pipeline template from SAM predefined set of pipeline templates hosted in the git
+    repository: aws/aws-sam-cli-pipeline-init-templates.git
+    downloads locally, then generates the pipeline config file from the selected pipeline template.
+    """
+    pipeline_templates_local_dir: Path = _clone_app_pipeline_templates()
```
hoffa, line 49: nit: no need for type hints if they're in the function definition (in fact might be better, to avoid accidental casting)

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+        selected_pipeline_template_manifest.location
+    )
+    _generate_from_pipeline_template(pipeline_template_dir=selected_pipeline_template_dir)
+
+
+def _generate_from_custom_location() -> None:
+    """
+    Prompts the user for a custom pipeline template location, downloads locally, then generates the pipeline config file
+    """
+    pipeline_template_repo_url: str = click.prompt(text="template Git location")
```
hoffa, line 68: How does this show up in the terminal? Should it be:

```suggestion
    pipeline_template_repo_url: str = click.prompt(text="Template Git location: ")
```

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+def _generate_from_app_pipeline_templates() -> None:
+    """
+    Prompts the user to choose a pipeline template from SAM predefined set of pipeline templates hosted in the git
+    repository: aws/aws-sam-cli-pipeline-init-templates.git
+    downloads locally, then generates the pipeline config file from the selected pipeline template.
+    """
+    pipeline_templates_local_dir: Path = _clone_app_pipeline_templates()
+
+    pipeline_templates_manifest: PipelineTemplatesManifest = _read_app_pipeline_templates_manifest(
+        pipeline_templates_dir=pipeline_templates_local_dir
```
hoffa, line 52: A little confusing since `pipeline_templates_dir` isn't an optional argument. This should be reserved for `kwargs`. The value should be a regular argument. Applies to all arguments in PR.

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+def _generate_from_custom_location() -> None:
+    """
+    Prompts the user for a custom pipeline template location, downloads locally, then generates the pipeline config file
+    """
+    pipeline_template_repo_url: str = click.prompt(text="template Git location")
+    pipeline_template_local_dir: Path = _clone_custom_pipeline_template(repo_url=pipeline_template_repo_url)
+    _generate_from_pipeline_template(pipeline_template_dir=pipeline_template_local_dir)
+    # Unlike app pipeline templates, custom pipeline templates are not shared between different SAM applications
+    # and should be cleaned up from users' machines after generating the pipeline config files.
+    shutil.rmtree(pipeline_template_local_dir, onerror=rmtree_callback)
```
hoffa, line 73: Is `pipeline_template_local_dir` outside of the SAM app root directory? What's different about custom templates?

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+    clone aws/aws-sam-cli-pipeline-init-templates.git Git repo to the local machine in SAM shared directory.
+    Returns:
+        the local directory path where the repo is cloned.
+    """
+    try:
+        return _clone_pipeline_templates(
+            repo_url=APP_PIPELINE_TEMPLATES_REPO_URL, clone_name=APP_PIPELINE_TEMPLATES_REPO_LOCAL_NAME
+        )
+    except PipelineTemplateCloneException:
+        # If can't clone app pipeline templates, try using an old clone from a previous run if already exist
```
hoffa, line 96: Is this logic necessary? Do we really expect it to fail enough times to justify this (potentially fragile, as no guarantee what's in that path) extra logic?

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+    Prompts the user a list of the available CI/CD providers to choose from
+
+    Parameters:
+        available_providers: List of available CI/CD providers like Jenkins, Gitlab and CircleCI
+
+    Returns:
+        The chosen provider
+    """
+    choices = list(map(str, range(1, len(available_providers) + 1)))
+    click.echo("CICD provider")
```
hoffa, line 186:
```suggestion
    click.echo("CI/CD provider")
```

samcli/commands/pipeline/init/pipeline_templates_manifest.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import Dict, List
+
+import yaml
+
+from samcli.commands.exceptions import AppPipelineTemplateManifestException
+from samcli.yamlhelper import parse_yaml_file
+
+
+class PipelineTemplateManifest:
+    """ The metadata of a Given pipeline template"""
```
hoffa, line 29: nit: if it's metadata, shouldn't we call it `PipelineTemplateMetadata`?

samcli/commands/pipeline/init/pipeline_templates_manifest.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from typing import Dict, List
+
+import yaml
+
+from samcli.commands.exceptions import AppPipelineTemplateManifestException
+from samcli.yamlhelper import parse_yaml_file
+
+
+class PipelineTemplateManifest:
```
hoffa, line 28: I thought a "manifest" is for the whole file? Here it suggests it's `templates` value?

samcli/commands/pipeline/init/pipeline_templates_manifest.py
```diff
@@ -0,0 +1,50 @@
+"""
+Represents a manifest that lists the available SAM pipeline templates.
+Example:
+    providers:
```
hoffa, line 4: `templates` already has identical `provider`s. Why is another `providers` necessary?

samcli/commands/pipeline/init/pipeline_templates_manifest.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, manifest_path: Path) -> None:
+        try:
+            manifest: Dict = parse_yaml_file(file_path=str(manifest_path))
+            self.providers: List[str] = manifest["providers"]
+            self.templates: List[PipelineTemplateManifest] = list(map(PipelineTemplateManifest, manifest["templates"]))
+        except (FileNotFoundError, KeyError, yaml.YAMLError) as ex:
+            raise AppPipelineTemplateManifestException(
+                "SAM pipeline templates manifest file is not found or ill-formatted. This could happen if the file "
+                f"{manifest_path} got deleted or manipulated."
```
hoffa, line 48: >or manipulated

"Manipulated" is similar to "your software might be compromised". The first sentence seems good enough, not sure if the URL is even necessary (should every error include the GitHub URL?).

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: it is outside the SAM app; in a shared location (default ~/.aws-sam/). 
[App pipeline templates](https://github.com/aws/aws-sam-cli-pipeline-init-templates), are the ones owned by SAM and can be shared between different SAM projects.

custom ones are provided by the user and the assumption is that it is not shared between different SAM projects.

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: this is just to be consistent with the `sam init` experience

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: I see. Why do we store them in a shared directory?

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: Consistency is great when a common pattern is made similar so that customers can reuse previous knowledge. This however seems like an edge case with questionable logic, so I see no value that consistency would provide here.

samcli/commands/pipeline/init/pipeline_templates_manifest.py
elbayaaa: We can extract the unique providers from the templates metadata, but defining them separately comes with two benefits:
1. readability; you instantly figure out which providers we support
2. consistent user experience; In this approach, this list of providers will be displayed to the user in this same order every time they run the command. The other approach doesn't guarantee this.

samcli/commands/pipeline/init/pipeline_templates_manifest.py
elbayaaa: so this is the local clone of the template (found in users' machine at ~/.aws-sam directory) so we mean the users themselves deleted or manipulated the file, therefore we add the next message "If you believe this is not the case ..."

What about using "modified" instead of "manipulated" ?

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: I understand there is no need to for passing a required parameter by name instead of position but I just see this more readable. Is the convention to not use named parameters for required ones?

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: We are sure the user has read/write permissions to this shared directory (otherwise we wouldn't be able to create the SAM project). But actually it make more sense to store in `/tmp`

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: It's a convention and they're technically different (see `*args` vs `**kwargs`).

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: Why not store in `.aws-sam` in the SAM project? All projects should be well isolated.

samcli/commands/pipeline/init/pipeline_templates_manifest.py
hoffa: For (2), `templates` is a list so each `provider` will also be ordered, no?

elbayaaa: it depends on how we order the templates. Let's say we order them first by number of stages (single-stage-pipelines, two-stages-pipelines...) then by provider: Jenkins, Gitlab, GithubActions

For example, Initially we may have:
```
templates:
  - single-stage-gitlab
  - single-stage-github-actions
  - two-stages-jenkins
  - two-stages-gitlab
  - two-stages-github-actions
```
which gave `["gitlab", "github-actions", "jenkins"]`
Now the single-stage-jenkins pipeline template became ready and we added it to its order on top of the list, then the providers list will change accordingly to ` [ "jenkins", "gitlab", "github-actions"]`

samcli/commands/pipeline/init/pipeline_templates_manifest.py
hoffa: Okay I see. And yes "modified" is great!

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+    Parameters:
+        pipeline_templates_dir: local directory of SAM pipeline templates
+
+    Raises:
+        AppPipelineTemplateManifestException if the manifest is not found, ill-formatted or missing required keys
+
+    Returns:
+        The manifest of the pipeline templates
+    """
+    manifest_path: Path = Path(os.path.normpath(pipeline_templates_dir.joinpath("manifest.yaml")))
```
aahung, line 152: why we need to do `normpath`?

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Parameters:
+        pipeline_templates_manifest: A manifest file lists the available providers and the associated pipeline templates
+
+    Returns:
+         The manifest (A section in the pipeline_templates_manifest) of the chosen pipeline template;
+    """
+    provider = _prompt_for_cicd_provider(pipeline_templates_manifest.providers)
+    provider_pipeline_templates: List[PipelineTemplateManifest] = list(
+        filter(lambda t: t.provider == provider, pipeline_templates_manifest.templates)
```
aahung, line 169: It is better to use this form rather than `filter` and lambda function: `[something for something in some_list if something.xyz == abc]`

See this post: https://stackoverflow.com/questions/1247486/list-comprehension-vs-map/6407222#6407222

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: the colon is automatically added by click

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: storing in /tmp will automatically clean it up rather thank manually remove it.

samcli/commands/pipeline/init/interactive_init_flow.py
elbayaaa: The main reason to do **clone** here is to get the up to date pipeline templates (in case if we updated [our pipeline templates](https://github.com/aws/aws-sam-cli-pipeline-init-templates)) other than this, we want to reused those already existing templates (that is why it is in a shared directory).

The pipeline templates are not expected to be actively updated, so this approach looks reasonable, no?

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,244 @@
...
+    clone a given Git pipeline template's Git repo to the user machine in SAM shared directory.
+    Returns: the local directory path where the repo is cloned to.
+
+    Parameters:
+        repo_url: the URL of the Git repo to clone
+
+    Returns:
+        the local directory path where the repo is cloned.
+    """
+    return _clone_pipeline_templates(repo_url=repo_url, clone_name=CUSTOM_PIPELINE_TEMPLATE_REPO_LOCAL_NAME)
```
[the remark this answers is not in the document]
elbayaaa: We discussed this before; Our Git repository cloning functiona accepts the destination as two parts:
1. clone_dir: parent directory where the local clone is going to be created at
2. clone_name: the name of the local clone

for example if we want to clone to ~/.aws-sam/templates/pipelines/my-custom-template 
the we pass the destination as:
* clone_dir: ~/.aws-sam/templates/pipelines/
* clone_name: my-custom-template 

**The reason behind that is that the function ensures the existence of the clone_dir(~/.aws-sam/templates/pipelines/) i.e. it creates it if not already exists**.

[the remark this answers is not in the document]
aahung: however it violates the separation of concern at line 73.

The discussion we had before was based on the usage at that time. let's have a call on

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: Isn't the `.aws-sam` directory already to store all kinds of project-related semi-temporary files? What value does another directory add?

elbayaaa: I don't have a strong opinion here. The only difference between this case and others is that we cleanup the pipeline template after usage. So, I think it is fine to clone to the project's `.aws-sam` folder and keep it without cleanup.

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -149,11 +146,11 @@ def _read_app_pipeline_templates_manifest(pipeline_templates_dir: Path) -> Pipel
     Returns:
         The manifest of the pipeline templates
     """
-    manifest_path: Path = Path(os.path.normpath(pipeline_templates_dir.joinpath("manifest.yaml")))
-    return PipelineTemplatesManifest(manifest_path=manifest_path)
+    manifest_path: Path = Path(pipeline_templates_dir.joinpath("manifest.yaml"))
```
aahung, line 149: `joinpath()` itself returns a Path, I don't think we need to wrap the return value in `Path(...)` again.
```suggestion
    return PipelineTemplatesManifest(pipeline_templates_dir.joinpath("manifest.yaml")
```

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,239 @@
...
+from samcli.lib.cookiecutter.question import Choice
+from samcli.lib.cookiecutter.template import Template
+from samcli.lib.utils.git_repo import GitRepo, CloneRepoException
+from samcli.lib.utils import osutils
+from .pipeline_templates_manifest import PipelineTemplateMetadata, PipelineTemplatesManifest
+
+LOG = logging.getLogger(__name__)
+shared_path: Path = global_cfg.config_dir
+APP_PIPELINE_TEMPLATES_REPO_URL = "https://github.com/aws/aws-sam-cli-pipeline-init-templates.git"
+APP_PIPELINE_TEMPLATES_REPO_URL = "https://github.com/elbayaaa/pipeline-templates.git"
```
aahung, line 24: Is it intentional?

aahung (commented):
I tried the `pipeline init` command (seems only works for gitlab for now), can we have a final message before exiting that list all the files are being generated?

The command ends with this could make customers confused and don't know what happened. Especially for gitlab, the generated file is a hidden file (.gitlab.yml). If we can print out a message like this, it would be great. (It is tricky considering only hooks know them, maybe we should formalize the hooks? or a manifest file in each template containing the list of generated files)

```
These files are created:
- .gitlab.yml
- cicd/123
- cicd/456
```

```
What is the production S3 bucket name for artifacts?: 4
What is the production ECR repository URI? []: 5
What is the production AWS region? [us-east-2]:
(end)
```

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa, line 24: Typo?

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,239 @@
...
+    pipeline_template_source_question = Choice(
+        key="pipeline-template-source",
+        text="Which pipeline template source would you like to use?",
+        options=[SAM_PIPELINE_TEMPLATE_SOURCE, CUSTOM_PIPELINE_TEMPLATE_SOURCE],
+    )
+    source = pipeline_template_source_question.ask()
+    if source == CUSTOM_PIPELINE_TEMPLATE_SOURCE:
+        _generate_from_custom_location()
+    else:
+        _generate_from_app_pipeline_templates()
```
hoffa, line 45: Why are they treated differently? They both clone from an URI.

samcli/commands/pipeline/init/pipeline_templates_manifest.py
hoffa: Okay I see, so it's essentially to ensure prompt ordering.

Is `provider`(`s`) internal identifiers or human-readable names? If the latter, they should be `GitLab CI/CD` and `GitHub Actions`.

hoffa (approved):

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,250 @@
...
+    click.echo("Successfully created the pipeline configuration file(s)")
```
aahung, line 46:
```suggestion
    click.echo("Successfully created the pipeline configuration file(s).")
```

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+    Prompts the user to choose a pipeline template from SAM predefined set of pipeline templates hosted in the git
+    repository: aws/aws-sam-cli-pipeline-init-templates.git
+    downloads locally, then generates the pipeline config file from the selected pipeline template.
+    """
+    pipeline_templates_local_dir: Path = _clone_app_pipeline_templates()
+    pipeline_templates_manifest: PipelineTemplatesManifest = _read_app_pipeline_templates_manifest(
+        pipeline_templates_local_dir
+    )
+    # The manifest contains multiple pipeline-templates so select one
```
aahung, line 59:
```suggestion
    # The manifest contains multiple pipeline-templates so let user select one
```

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -0,0 +1,250 @@
...
+    _generate_from_pipeline_template(selected_pipeline_template_dir)
+
+
+def _generate_from_custom_location() -> None:
+    """
+    Prompts the user for a custom pipeline template location, downloads locally, then generates the pipeline config file
+    """
+    pipeline_template_git_location: str = click.prompt("Template Git location")
+    if os.path.exists(pipeline_template_git_location):
+        _generate_from_pipeline_template(Path(pipeline_template_git_location))
```
aahung, line 75: A comment here can help understand why we check a git location exists or not, is to check whether it is a local path right?

aahung (approved):
lgtm

samcli/commands/pipeline/init/interactive_init_flow.py
hoffa: Okay so this behavior is useful only if a clone has been done previously, and a new clone fails (e.g. due to no Internet connection).

I'm fine keeping it as-is since we're already doing something similar, but I do think it has negatives:

1. Requires extra care to ensure the target is clean (since we don't have that guarantee)
2. Adds a precedent where we're using both local and global state, which means future developers will have to spend those extra seconds considering why and their caveats

PR #2841, opened 2021-04-29
Integration test for pipeline init and pipeline bootstrap

aahung:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -290,3 +289,8 @@ def print_resources_summary(self) -> None:
             )
             click.secho(f"\tACCESS_KEY_ID: {self.pipeline_user.access_key_id}", fg="green")
             click.secho(f"\tSECRET_ACCESS_KEY: {self.pipeline_user.secret_access_key}", fg="green")
+
+    @staticmethod
```
elbayaaa, line 293: should we make it a non static method as we user `self.name` as part of the stack-name by design

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+
+class PipelineBase(TestCase):
+    def base_command(self):
+        command = "sam"
+        if os.getenv("SAM_CLI_DEV"):
+            command = "samdev"
+
+        return command
+
+    def run_command_with_inputs(self, command_list, inputs: List[str]):
```
elbayaaa, line 21: [nit] type hint for command_list

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+        return command_list
+
+
+class BootstrapIntegBase(PipelineBase):
+    stack_names: List[str]
+
+    @classmethod
+    def setUpClass(cls):
+        cls.cf_client = boto3.client("cloudformation")
+        pass
```
elbayaaa, line 59: typo?

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def setUp(self):
+        self.stack_names = []
+        super().setUp()
+
+    def tearDown(self):
+        if self.stack_names:
```
elbayaaa, line 66: unnecessary check?

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+            for stack_name in self.stack_names:
+                self.cf_client.delete_stack(StackName=stack_name)
+        shutil.rmtree(os.path.join(os.getcwd(), ".aws-sam", "pipeline"), ignore_errors=True)
```
elbayaaa, line 69: This line is dangerous; what if I ran `sam pipeline bootstrap` (for real use-case not for integ-tests) then later on I ran the integration tests, I will lose the real data that I don't want to lose.

Also, if we later on added additional files to `.aws-sam/pipeline` folder it will be lost as well.

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+        super().tearDown()
+
+    def base_command(self):
+        command = "sam"
+        if os.getenv("SAM_CLI_DEV"):
+            command = "samdev"
+
+        return command
```
elbayaaa, line 77: should be removed? already defined in PipelineBase ?

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def get_bootstrap_command_list(
+        self,
+        # right now we only support interactive mode
```
elbayaaa, line 81: not really. bootstrap supports both modes. It is actually more convenient to use it in non-interactive mode of you are going to let the command generate all of the resources for you.

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+        interactive=True,
+    ):
+        command_list = [self.base_command(), "pipeline", "bootstrap"]
+
+        if interactive:
+            command_list = command_list + ["--interactive"]
```
elbayaaa, line 87: why do we need this? the default is the interactive mode

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+        if len(tokens) == 1:
+            # no resources created
+            return []
+
+        created_resources_section = tokens[1]
+        # after created resource section, it might be provided resource section,
+        # or configure credential section.
+        # we use two split to find where it ends.
+        created_resources_section = re.split(configure_start, created_resources_section)[0]
+        created_resources_section = re.split(provided_start, created_resources_section)[0]
```
elbayaaa, line 113: typo

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,119 @@
...
+from tests.testing_utils import run_command_with_input
+
+
+class PipelineBase(TestCase):
+    def base_command(self):
+        command = "sam"
+        if os.getenv("SAM_CLI_DEV"):
+            command = "samdev"
+
+        return command
```
elbayaaa, line 19: [out of scope of this PR] This method along with other many methods are duplicated in all of the integrations test. We need to have a Base class for all of them.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,163 @@
...
+        ]
+
+        bootstrap_process_execute = self.run_command_with_inputs(bootstrap_command_list, inputs)
+
+        self.assertEqual(bootstrap_process_execute.process.returncode, 0)
+        stdout = bootstrap_process_execute.stdout.decode()
+        self.assertIn("We have created the following resources", stdout)
+        # make sure pipeline user's credential is printed
+        self.assertIn("ACCESS_KEY_ID", stdout)
+        self.assertIn("SECRET_ACCESS_KEY", stdout)
```
elbayaaa, line 41: flaky test? if `sam pipeline bootstrap` was ran before, then this will not be shown. It is only shown the first time.

elbayaaa: Nevermind, we already set pipeline user to empty at line 25. so it will be created and the credentials will be shown.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,163 @@
...
+    @parameterized.expand([("create_ecr_repo",), (False,)])
+    def test_interactive_with_no_resources_provided(self, create_ecr_repo: bool):
+        stage_name = self._method_to_stage_name(self.id())
+        self.stack_names = [Stage._get_stack_name(stage_name)]
+
+        bootstrap_command_list = self.get_bootstrap_command_list(interactive=True)
+
+        inputs = [
+            stage_name,
+            "",  # pipeline user
```
elbayaaa, line 25: just want to make sure about the behavior when there is a default value from pipelineconfig.toml file and we provide an empty string here. Will the final value be empty string or the one provided in pipelineconfig.toml ?

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,163 @@
...
+        bootstrap_command_list = self.get_bootstrap_command_list(interactive=True)
+
+        inputs = [
+            stage_name,
+            "arn:aws:iam::123:user/user-name",  # pipeline user
+            "arn:aws:iam::123:role/role-name",  # Pipeline execution role
+            "arn:aws:iam::123:role/role-name",  # CloudFormation execution role
+            "arn:aws:s3:::bucket-name",  # Artifacts bucket
+            "3",  # Should we create ECR repo, 3 - specify one
+            "arn:aws:ecr:::repository/repo-name",  # ecr repo
```
elbayaaa, line 62: [nit] it  worth adding an AWS region, i.e. "arn:aws:ecr:us-east-2::repository/repo-name" as unlike other resources above, ECR is region specific.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,163 @@
...
+            "y",  # proceed
+        ]
+
+        bootstrap_process_execute = self.run_command_with_inputs(bootstrap_command_list, inputs)
+
+        self.assertEqual(bootstrap_process_execute.process.returncode, 0)
+        stdout = bootstrap_process_execute.stdout.decode()
+        self.assertIn("Successfully created!", stdout)
+        # make sure the not provided resource is the only resource created.
+        self.assertIn("aws-sam-cli-managed-test-CloudFormationExecution", stdout)
```
elbayaaa, line 120: I'm confused about this line, how does this mean CloudFormation execution role is the only resource that got created?

samcli/lib/pipeline/bootstrap/stage.py
aahung: changed

tests/integration/pipeline/base.py
aahung: added

tests/integration/pipeline/base.py
aahung: removed `pass`

tests/integration/pipeline/base.py
aahung: removed

tests/integration/pipeline/base.py
aahung: It is our practice to do that, see tests/integration/deploy/test_deploy_command.py:56

Also it is unlikely to run `sam pipeline bootstrap` and integration test in the same directory.

tests/integration/pipeline/base.py
aahung: removed

tests/integration/pipeline/base.py
aahung: Added a few test for no-interactive

tests/integration/pipeline/base.py
aahung: updated

tests/integration/pipeline/base.py
aahung: why it is a typo?

tests/integration/pipeline/test_bootstrap_command.py
aahung: why `sam pipeline bootstrap` was ran before?

tests/integration/pipeline/test_bootstrap_command.py
aahung: good question, I think it should be the one in pipelineconfig.toml.

tests/integration/pipeline/test_bootstrap_command.py
aahung: it does not matter actually, because we don't have logics to check region and it is just a random string to check stdout again

tests/integration/pipeline/test_bootstrap_command.py
aahung: Check the next line, the next line ensures that

tests/integration/pipeline/base.py
elbayaaa: yeah, I agree it is unlikely. So, this should be fine.

tests/integration/pipeline/base.py
elbayaaa: Honestly, I feel like this is a very fragile way to determine created resources; Any changes in the printed messages will break these tests. It is a high maintenance effort. Can we instead use CloudFormation's describe-stack-resources operation ?

tests/integration/pipeline/test_bootstrap_command.py
elbayaaa: well, I asked this because if there is a value in pipelineconfig.toml then this test case will fail as the pipeline-user is provided not created. However, as you said, it is highly unlikely to run the integration tests from a directory that already has a `pipelineconfig.toml` file. Let's not handle this case but we need to add a comment about it.

tests/integration/pipeline/test_bootstrap_command.py
elbayaaa: yeah it doesn't matter in our case (that's why it is bit). I just see you use complete ARNs for that other resources so pointed this out. It is actually missing both the region and accountId.

tests/integration/pipeline/test_bootstrap_command.py
aahung: Yeah it is unlikely. I don't think we need comment about the environment. It is a reasonable assumption that the test environment is clean. Otherwise we need to have comments in a lot of places, like samconfig.toml will affect package or deploy.

tests/integration/pipeline/base.py
aahung: Yeah I like the idea of using describe-stack-resources

tests/integration/pipeline/test_bootstrap_command.py
elbayaaa: oh, I see it now:
* "aws-sam-cli-managed" is the stack prefix
* "test" is the only remaining part of stage-name (test_interactive_with_some_required_resources_provided) in this case
* CloudFormationExecution is what we actually looking for.

This is quite confusing. Hopfully this will be fixed with using descibe-stack-resources

aahung: yeah, switching to use `descibe-stack-resources`

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+@skipIf(SKIP_DEPLOY_TESTS, "Skip bootstrap tests in CI/CD only")
+class TestBootstrap(BootstrapIntegBase):
+    @parameterized.expand([("create_ecr_repo",), (False,)])
+    def test_interactive_with_no_resources_provided(self, create_ecr_repo: bool):
+        stage_name = self._method_to_stage_name(self.id())
+        mock_stage = Mock()
+        mock_stage.name = stage_name
+        self.stack_names = [Stage._get_stack_name(mock_stage)]
```
elbayaaa, line 28: this logic is repeated in many place. Can we have something like
`stage_name, stack_name = self._method_to_stage_and_stack_names()` ? 
Also, we can create real Stage instead of mock? the only required attribute is stage_name

aahung: yeah good idea, updated

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,122 @@
...
+        if artifacts_bucket:
+            command_list += ["--artifacts-bucket", artifacts_bucket]
+        if create_ecr_repo:
+            command_list += ["--create-ecr-repo"]
+        if ecr_repo:
+            command_list += ["--ecr-repo", ecr_repo]
+        if pipeline_ip_range:
+            command_list += ["--pipeline-ip-range", pipeline_ip_range]
+        if no_confirm_changeset:
+            command_list += ["--no-confirm-changeset"]
```
elbayaaa, line 105: I'm fine with that for the purpose of integration test but it would be great if we refactored it.

tests/integration/pipeline/base.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return command_list
+
+    def _extract_created_resource_logical_ids(self, stack_name: str) -> Set[str]:
+        response = self.cf_client.describe_stack_resources(StackName=stack_name)
+        return {resource["LogicalResourceId"] for resource in response["StackResources"]}
```
elbayaaa, line 111: much better 👍

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+from tests.testing_utils import (
+    run_command_with_input,
+    RUNNING_ON_CI,
+    RUNNING_TEST_FOR_MASTER_ON_CI,
+    RUN_BY_CANARY,
+    run_command,
+)
+
+# bootstrap tests require credentials and CI/CD will only add credentials to the env if the PR is from the same repo.
+# This is to restrict tests to run outside of CI/CD, when the branch is not master or tests are not run by Canary
```
elbayaaa, line 15: I didn't understand this comment

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+SKIP_DEPLOY_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
```
elbayaaa, line 16: should it be `SKIP_PIPELINE_BOOTSTRAP_TESTS` ?

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+            "PipelineUser",
+            "PipelineUserAccessKey",
+            "CloudFormationExecutionRole",
+            "PipelineExecutionRole",
+            "ArtifactsBucket",
+            "ArtifactsBucketPolicy",
+            "PipelineExecutionRolePermissionPolicy",
+        }
+        if create_ecr_repo:
+            self.assertIn("arn:aws:ecr:", stdout)
```
elbayaaa, line 58: I think let's not have this assertion. Having "ECRRepo" in the `created_resource_logical_ids` is sufficient.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+                    *common_resources,
+                    "ECRRepo",
+                },
+                self._extract_created_resource_logical_ids(stack_name),
+            )
+        else:
+            self.assertNotIn("arn:aws:ecr:", stdout)
+            self.assertSetEqual(common_resources, self._extract_created_resource_logical_ids(stack_name))
+
+    @parameterized.expand([("create_ecr_repo",), (False,)])
```
elbayaaa, line 70: I'm quite confused about the "create_ecr_repo" string here, is this a way to pass a True value for the parameter (a non empty string is evaluated to True)? why not just passing True?

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+        stage_name, stack_name = self._get_stage_and_stack_name()
+        self.stack_names = [stack_name]
+
+        bootstrap_command_list = self.get_bootstrap_command_list(
+            no_interactive=True, create_ecr_repo=create_ecr_repo, no_confirm_changeset=True
+        )
+
+        bootstrap_process_execute = run_command(bootstrap_command_list)
+
+        self.assertEqual(bootstrap_process_execute.process.returncode, 2)
```
elbayaaa, line 81: please also test the non-interactive happy case.

elbayaaa: Never mind, found it below.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+        bootstrap_command_list = self.get_bootstrap_command_list(
+            no_interactive=True,
+            stage_name=stage_name,
+            pipeline_user="arn:aws:iam::123:user/user-name",  # pipeline user
+            pipeline_execution_role="arn:aws:iam::123:role/role-name",  # Pipeline execution role
+            cloudformation_execution_role="arn:aws:iam::123:role/role-name",  # CloudFormation execution role
+            artifacts_bucket="arn:aws:s3:::bucket-name",  # Artifacts bucket
+            ecr_repo="arn:aws:ecr:::repository/repo-name",  # ecr repo
+            pipeline_ip_range="1.2.3.4/24",  # Pipeline IP address range
+            no_confirm_changeset=confirm_changeset,
```
elbayaaa, line 123: typo?

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+
+        inputs = [
+            "y",  # proceed
+        ]
```
elbayaaa, line 128: can we make this something like `inputs = ["y"] if confirm_changeset else []` to make sure we don't pass "y" when there is no confrimation required.

elbayaaa: no worries, it is handled right below.

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,227 @@
...
+                "arn:aws:iam::123:role/role-name",  # Pipeline execution role
+                "arn:aws:iam::123:role/role-name",  # CloudFormation execution role
+                "arn:aws:s3:::bucket-name",  # Artifacts bucket
+                "1",  # Should we create ECR repo, 1 - No, 2 - Yes
+                "",  # Pipeline IP address range
+                "y",  # proceed
+            ]
+
+            bootstrap_process_execute = run_command_with_input(
+                bootstrap_command_list, ("\n".join(inputs) + "\n").encode()
```
elbayaaa, line 213: why are we joining in "\n" ? this is not the case with elsewhere

tests/integration/pipeline/test_bootstrap_command.py
aahung: fixed and move no-interactive test to a new one

tests/integration/pipeline/test_bootstrap_command.py
aahung: We don't want to run these tests in PR because it will expose our AWS creds.

The comment is the same as in 

https://github.com/aws/aws-sam-cli/blob/5cc7680068c820e972d6165a0cccd21677e2a428/tests/integration/deploy/test_deploy_command.py#L19-L21

tests/integration/pipeline/test_bootstrap_command.py
aahung: good catch, fixed

tests/integration/pipeline/test_bootstrap_command.py
aahung: makes sense, removed

tests/integration/pipeline/test_bootstrap_command.py
aahung: It is our parameterized convention. By giving a string (which will resolved to `True` eventually` instead of a boolean will make the test name generated by pytest more readable (so that `create_ecr_repo` shows up in test name).

See how we did it in the past:

https://github.com/aws/aws-sam-cli/blob/5cc7680068c820e972d6165a0cccd21677e2a428/tests/integration/buildcmd/test_build_cmd.py#L90-L103

tests/integration/pipeline/test_bootstrap_command.py
aahung: This is how we used to do with `run_command_with_input`, see https://github.com/aws/aws-sam-cli/blob/5cc7680068c820e972d6165a0cccd21677e2a428/tests/integration/deploy/test_deploy_command.py#L535-L537

tests/integration/pipeline/base.py
aahung: We used to this "batch" if..., like in https://github.com/aws/aws-sam-cli/blob/5cc7680068c820e972d6165a0cccd21677e2a428/tests/integration/deploy/deploy_integ_base.py#L49-L64

Do you have improvement suggestions?

tests/integration/pipeline/test_bootstrap_command.py
```diff
@@ -0,0 +1,247 @@
...
+            "1",  # Should we create ECR repo, 1 - No
+            "",  # Pipeline IP address range
+            "N",  # cancel
+        ]
+
+        bootstrap_process_execute = self.run_command_with_inputs(bootstrap_command_list, inputs)
+
+        self.assertEqual(bootstrap_process_execute.process.returncode, 0)
+        stdout = bootstrap_process_execute.stdout.decode()
+        self.assertTrue(stdout.strip().endswith("Should we proceed with the creation? [y/N]:"))
```
elbayaaa, line 179: Let's also assert no stack is created

tests/integration/pipeline/test_bootstrap_command.py
elbayaaa: I was trying to understand what is the difference between this one and every other place in this PR. I believe you just need to use `run_command_with_inputs` instead of `run_command_with_input`.

Also, consider moving `run_command_with_inputs` to testing_utils instead of defining it here only.

elbayaaa (approved):
logic LGTM. Please fix the failing AppVeyor integration-tests on Windows.

tests/integration/pipeline/test_bootstrap_command.py
aahung: added

tests/integration/pipeline/test_bootstrap_command.py
aahung: good idea, moved to utils

aahung:
Other CI check failure is related to do with pips and should be resolved if `sam-pipelines` is up to date with `develop` branch

aahung:
Did a merge from branch sam-pipelines

hoffa (approved):

aahung:
@hoffa @elbayaaa did a rebase to change stage -> env, nothing else changed. It it passes, I am gonna merge it

Issue #2355, opened 2020-11-05
Automatic template.json detection


tsndr:
### Describe your idea/feature/enhancement
I'd like to see SAM CLI detect `template.json` automatially just like it does with `template.yaml`/`template.yml`.

#### Example:
If I run `sam validate` with a `template.yaml` file it works.

If I use the JSON syntax I get an error: **SAM Template Not Found**
So I have to type: `sam validate -t template.json` to get it to work.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)
No!

mndeveci:
Resolving as this have been released with [SAM CLI v1.24.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.24.0)

PR #2855, opened 2021-05-05
Validate default template.json

ssenchenko:

#2355

Allows users to validate template.json (in json instead of yaml format) without specifying `--template` argument with a path to the file. Improves user experience

Adding `template.json` to the list of default file names when looking up a template file.

None

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -14,7 +14,7 @@ from samcli.cli.types import CfnParameterOverridesType, CfnMetadataType, CfnTags
 from samcli.commands._utils.custom_options.option_nargs import OptionNargs
 from samcli.commands._utils.template import get_template_artifacts_format
 
-_TEMPLATE_OPTION_DEFAULT_VALUE = "template.[yaml|yml]"
+_TEMPLATE_OPTION_DEFAULT_VALUE = "template.[yaml|yml|json]"
 DEFAULT_STACK_NAME = "sam-app"
 
 LOG = logging.getLogger(__name__)
@@ -35,7 +35,7 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
 
     original_template_path = os.path.abspath(provided_value)
 
-    search_paths = ["template.yaml", "template.yml"]
+    search_paths = ["template.yaml", "template.yml", "template.json"]
 
     if include_build:
         search_paths.insert(0, os.path.join(".aws-sam", "build", "template.yaml"))
@@ -43,7 +43,7 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
     if provided_value == _TEMPLATE_OPTION_DEFAULT_VALUE:
         # "--template" is an alias of "--template-file", however, only the first option name "--template-file" in
         # ctx.default_map is used as default value of provided value. Here we add "--template"'s value as second
-        # default value in this option, so that the command line paramerters from config file can load it.
+        # default value in this option, so that the command line parameters from config file can load it.
         if ctx and ctx.default_map.get("template", None):
             provided_value = ctx.default_map.get("template")
         else:
--- a/tests/integration/validate/test_validate_command.py
+++ b/tests/integration/validate/test_validate_command.py
@@ -62,7 +62,7 @@ class TestValidate(TestCase):
     def test_default_template(self, relative_folder: str, expected_file: TemplateFileTypes):
         cwd = f"tests/integration/testdata/validate/{relative_folder}"
         command_result = run_command(self.command_list(), cwd=cwd)
-        pattern = self.patterns[expected_file]
+        pattern = self.patterns[expected_file]  # type: ignore
         output = command_result.stdout.decode("utf-8")
         self.assertEqual(command_result.process.returncode, 0)
         self.assertIsNotNone(pattern.match(output))
--- a/tests/integration/validate/test_validate_command.py
+++ b/tests/integration/validate/test_validate_command.py
@@ -23,7 +23,7 @@ class TemplateFileTypes(Enum):
     YAML = auto()
 
 
-@skipIf(SKIP_VALIDATE_TESTS, "Skip deploy tests in CI/CD only")
+@skipIf(SKIP_VALIDATE_TESTS, "Skip validate tests in CI/CD only")
 class TestValidate(TestCase):
     @classmethod
     def setUpClass(cls):
--- a/tests/integration/validate/test_validate_command.py
+++ b/tests/integration/validate/test_validate_command.py
@@ -1,0 +1,68 @@
+"""
+Integration tests for sam validate
+"""
+
+import os
+import re
+from enum import Enum, auto
+from pathlib import Path
+from typing import List, Optional
+from unittest import TestCase
+
+from parameterized import parameterized
+from tests.testing_utils import run_command
+
+
+class TemplateFileTypes(Enum):
+    JSON = auto()
+    YAML = auto()
+
+
+class TestValidate(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.patterns = {
+            TemplateFileTypes.JSON: re.compile(r"^/.+/template[.]json is a valid SAM Template$"),
+            TemplateFileTypes.YAML: re.compile(r"^/.+/template[.]yaml is a valid SAM Template$"),
+        }
+
+    @staticmethod
+    def base_command() -> str:
+        return "samdev" if os.getenv("SAM_CLI_DEV") else "sam"
+
+    def command_list(
+        self,
+        template_file: Optional[Path] = None,
+        profile: Optional[str] = None,
+        region: Optional[str] = None,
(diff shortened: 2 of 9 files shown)


samcli/commands/_utils/options.py
```diff
@@ -35,15 +35,15 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
 
     original_template_path = os.path.abspath(provided_value)
 
-    search_paths = ["template.yaml", "template.yml"]
+    search_paths = ["template.yaml", "template.yml", "template.json"]
```
sriram-mv, line 38: something to check, if build is included, we always the convert the built template back to yaml (IIRC) and that is referenced template if one is not provided during the deploy command, can we check this behavior?

tests/unit/commands/_utils/test_options.py
```diff
@@ -53,7 +53,18 @@ def test_must_return_yml_extension(self, os_mock):
     def test_must_return_yaml_extension(self, os_mock):
         expected = "template.yaml"
 
-        os_mock.path.exists.return_value = True
+        os_mock.path.exists.side_effect = lambda file_name: file_name == expected
```
sriram-mv, line 56: Lets add an integration test too.

sriram-mv (changes requested):

samcli/commands/_utils/options.py
aahung: > something to check, if build is included, we always the convert the built template back to yaml (IIRC)

If `sam build` is executed, `template.json` will be rewritten into `.aws-sam/build/template.yaml`. And if `sam package/deploy` is later executed, `.aws-sam/build/template.yaml` should be picked up. Nice to double check this behavior. 

@sriram-mv so we want the priority in `sam local` and `sam package/deploy` to be:
1. .aws-sam/build/template.yaml
2. ./template.yaml
3. ./template.yml
4. ./template.json
is that accurate?

ssenchenko: @sriram-mv Yes, when we run build, `template.json` is converted to `template.yaml` in `.aws-sam/build` after `sam build` execution. It works without explicit `--template` parameter

tests/unit/commands/_utils/test_options.py
ssenchenko: Sure, will do

tests/integration/validate/test_validate_command.py
```diff
@@ -0,0 +1,68 @@
...
+    ) -> List[str]:
+        command_list = [self.base_command(), "validate"]
+        if template_file:
+            command_list += ["--template-file", str(template_file)]
+        if profile:
+            command_list += ["--profile", profile]
+        if region:
+            command_list += ["--region", region]
+        if config_file:
+            command_list = ["--config_file", str(config_file)]
```
aahung, line 48: Looks like you are missing a `+=` here?

tests/integration/validate/test_validate_command.py
```diff
@@ -0,0 +1,68 @@
...
+            ("default_json", TemplateFileTypes.JSON),  # project with template.json
+            ("multiple_files", TemplateFileTypes.YAML),  # project with both template.yaml and template.json
+            (
+                "with_build",
+                TemplateFileTypes.JSON,
+            ),  # project with template.json and standard build directory .aws-sam/build/template.yaml
+        ]
+    )
+    def test_default_template(self, relative_folder: str, expected_file: TemplateFileTypes):
+        cwd = f"tests/integration/testdata/validate/{relative_folder}"
```
aahung, line 63: Can we use this way to define the testdata path? So that integration tests can be started from any directory (also using `/` might not work on Windows)

https://github.com/aws/aws-sam-cli/blob/283e546097f7520da740560de8a963ea387098dd/tests/integration/buildcmd/test_build_cmd.py#L1467

tests/integration/validate/test_validate_command.py
```diff
@@ -0,0 +1,68 @@
...
+        command_result = run_command(self.command_list(), cwd=cwd)
+        pattern = self.patterns[expected_file]  # type: ignore
+        output = command_result.stdout.decode("utf-8")
+        self.assertEqual(command_result.process.returncode, 0)
+        self.assertIsNotNone(pattern.match(output))
```
aahung, line 68: You can use `self.assertRegex` here

aahung (commented):
Thanks for adding the tests!

tests/integration/validate/test_validate_command.py
ssenchenko: yeah thanks

tests/integration/validate/test_validate_command.py
ssenchenko: Agree, that's more bullet proof. Thanks! Though direct slash `/` works for windows path as well

tests/integration/validate/test_validate_command.py
ssenchenko: I didn't know about that one. Thank you

tests/integration/validate/test_validate_command.py
aahung: oh really? that's cool, does it work in normal CMD as well, apart from WSL?

ssenchenko: it works in PowerShell not sure about an archaic cmd though 😄 I haven't been using cmd for a long time. But in this case it will work because python supports both `\\` and `/` for windows paths

tests/integration/validate/test_validate_command.py
```diff
@@ -8,16 +8,22 @@
...
+# This is to restrict package tests to run outside of CI/CD, when the branch is not master or tests are not run by Canary
+SKIP_VALIDATE_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
 
 
 class TemplateFileTypes(Enum):
     JSON = auto()
     YAML = auto()
 
 
+@skipIf(SKIP_VALIDATE_TESTS, "Skip deploy tests in CI/CD only")
```
aahung, line 26:
```suggestion
@skipIf(SKIP_VALIDATE_TESTS, "Skip validate tests in CI/CD only")
```

tests/integration/validate/test_validate_command.py
```diff
@@ -59,10 +65,11 @@ def command_list(
             ),  # project with template.json and standard build directory .aws-sam/build/template.yaml
         ]
     )
-    def test_default_template(self, relative_folder: str, expected_file: TemplateFileTypes):
-        cwd = f"tests/integration/testdata/validate/{relative_folder}"
-        command_result = run_command(self.command_list(), cwd=cwd)
+    def test_default_template_file_choice(self, relative_folder: str, expected_file: TemplateFileTypes):
+        test_data_path = Path(__file__).resolve().parents[2] / "integration" / "testdata" / "validate"
```
aahung, line 69: I didn't know `/` operator can be used in `Path`, wow!

aahung (approved):
LGTM! 🚢

tests/integration/validate/test_validate_command.py
ssenchenko: yeah, it's nice to have operator overloading 😄

sriram-mv: is the "/" similar to using pathlib.join ?

samcli/commands/_utils/options.py
sriram-mv: ok sounds good, it (template.json) should be the last thing to look up in order to avoid any breaking changes or unwanted behavior from this.

tests/integration/validate/test_validate_command.py
sriram-mv: 👍🏽

tests/integration/validate/test_validate_command.py
aahung: yeah I tested, it is

samcli/commands/_utils/options.py
sriram-mv: Something I just realised. the parser is still a yaml_parser.

yaml_helper.py

```
def yaml_parse(yamlstr):
    """Parse a yaml string"""
    try:
        # PyYAML doesn't support json as well as it should, so if the input
        # is actually just json it is better to parse it with the standard
        # json parser.
        return json.loads(yamlstr, object_pairs_hook=OrderedDict)
    except ValueError:
        yaml.SafeLoader.add_constructor(yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, _dict_constructor)
        yaml.SafeLoader.add_multi_constructor("!", intrinsics_multi_constructor)
        return yaml.safe_load(yamlstr)
```

if you look at the comment, it suggest use a json parser instead.

sriram-mv (changes requested):

sriram-mv (approved):
I didnt realize we are calling json.loads() first, I focused on the docstring instead.

PR #2857, opened 2021-05-05
fix: Catch more errors when building an image

valerena:
In certain scenarios, the error from the Docker build is not
raised as an exception, but it's returned as part of the command's
output.

In the current situation, when invoking a function locally, if for some reason building the rapid image fails, then the error shown to the user is raised later in the process, hiding the actual build error.

The response of the build command is now taken into account, checking if there's an error, and throwing a Build Exception when that's the case.

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


sriram-mv (approved):

tests/unit/local/docker/test_lambda_image.py
```diff
@@ -292,7 +297,60 @@ def test_build_image_fails_with_BuildError(
...
+            decode=True,
+        )
+
+        docker_full_path_mock.unlink.assert_not_called()
+
+    @patch("samcli.local.docker.lambda_image.create_tarball")
+    @patch("samcli.local.docker.lambda_image.uuid")
+    @patch("samcli.local.docker.lambda_image.Path")
+    @patch("samcli.local.docker.lambda_image.LambdaImage._generate_dockerfile")
+    def test_build_image_fails_with_BuildError_from_output(
```
sriram-mv, line 314: could we add an integration test oo?

sriram-mv (changes requested):

samcli/lib/build/build_graph.py
```diff
@@ -351,7 +351,7 @@ def __init__(
     def __str__(self) -> str:
         return (
             f"LayerBuildDefinition({self.name}, {self.codeuri}, {self.source_md5}, {self.uuid}, "
-            f"{self.build_method}, {self.compatible_runtimes}, {self.env_vars}, {self.layer.name})"
+            f"{self.build_method}, {self.compatible_runtimes}, {self.env_vars})"
```
valerena, line 354: This change comes from this commit, that's already merged into `aws:develop`: https://github.com/aws/aws-sam-cli/commit/199ded1901964ed17023e9d5ed5682ff77ae3dfd I'm not sure why it appears here

sriram-mv (approved):

mgrandis (approved):

PR #2862, opened 2021-05-08
refactor: refactor logs command library

mndeveci:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/observability/cw_logs/cw_log_formatters.py
```diff
@@ -0,0 +1,87 @@
...
+
+        return event
+
+
+class CWColorizeErrorsFormatter(ObservabilityEventMapper[CWLogEvent]):
+    """
+    Mapper implementation which will colorize some pre-defined error messages
+    """
+
+    NODEJS_CRASH_MESSAGE = "Process exited before completing request"
```
sriram-mv, line 42: Seems to be a typo in the ENUM name.

samcli/lib/observability/cw_logs/cw_log_puller.py
```diff
@@ -0,0 +1,111 @@
...
+    Puller implementation that can pull events from CloudWatch log group
+    """
+
+    def __init__(
+        self,
+        logs_client: Any,
+        consumer: ObservabilityEventConsumer,
+        cw_log_group: str,
+        resource_name: Optional[str] = None,
+        max_retries: int = 1000,
```
sriram-mv, line 27: This is a large number of retries, is there any reason for it?

samcli/lib/observability/observability_info_puller.py
```diff
@@ -0,0 +1,142 @@
...
+        ----------
+        mappers : List[ObservabilityEventMapper]
+            List of event mappers which will be used to process events before passing to consumer
+        consumer : ObservabilityEventConsumer
+            Actual consumer which will handle the events after they are processed by mappers
+        """
+        self._mappers = mappers
+        self._consumer = consumer
+
+    def consume(self, event: ObservabilityEvent):
```
sriram-mv, line 134: 🎉

samcli/lib/observability/cw_logs/cw_log_puller.py
mndeveci: Not sure, I just refactored the current value from here; https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/logs/fetcher.py#L81

AFAIK, this tracks the latest log messages, so if there are no new log events for 1000 retry, the command will exit. I agree this is bad for both cases, whether you get an error or when you don't get any new logging message. I will check with UX to see if we need to tie this to a parameter to make it more flexible.

samcli/lib/observability/cw_logs/cw_log_formatters.py
mndeveci: I think I need another coffee ☕  Can you point the type in the name?

jfuss (commented):
A couple small comments.

samcli/lib/observability/observability_info_puller.py
```diff
@@ -0,0 +1,142 @@
...
+        resource_name : Optional[str]
+            Resource name related to this event. This is optional since not all events is connected to a single resource
+        """
+        self.event = event
+        self.timestamp = timestamp
+        self.resource_name = resource_name
+
+
+# Generic type for identifying different ObservabilityEvent
+ObservabilityEventType = TypeVar("ObservabilityEventType", bound=ObservabilityEvent)
```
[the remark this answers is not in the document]
mndeveci: I tried to do this while developing, but since it is using `ObservabilityEvent` class, it needs to be declared after that :(

samcli/lib/observability/cw_logs/cw_log_formatters.py
```diff
@@ -0,0 +1,87 @@
...
+            event.message = self._colored.red(event.message)
+        return event
+
+
+class CWJsonFormatter(ObservabilityEventMapper[CWLogEvent]):
+    """
+    Mapper implementation which will auto indent the input if the input is a JSON object
+    """
+
+    # pylint: disable=R0201
```
jfuss, line 62: Can you add a comment on why this is disabled? I can guess but would be good to document.

samcli/lib/observability/cw_logs/cw_log_formatters.py
```diff
@@ -0,0 +1,87 @@
...
+    def map(self, event: CWLogEvent) -> CWLogEvent:
+        try:
+            if event.message.startswith("{"):
+                msg_dict = json.loads(event.message)
+                event.message = json.dumps(msg_dict, indent=2)
+        except Exception:
+            pass
```
jfuss, line 69: This can easily hide unexpected errors since this will catch any python error raised. Can you comment why we need to catch everything here? 

My generally suggestion is to scope this and let unhandled exceptions bubble up instead of masking them. Though I know sometimes this is needed.

samcli/lib/observability/cw_logs/cw_log_formatters.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Mapper implementation which will format given CloudWatch log event into string with coloring
+    log stream name and timestamp
+    """
+
+    def __init__(self, colored: Colored):
+        self._colored = colored
+
+    # pylint: disable=R0201
+    def map(self, event: CWLogEvent) -> str:
```
jfuss, line 84: I just noticed this function returns `str` but all others return `CWLogEvent`. My one concern is if the callers do not need to care about the formatters, it could be confusing to use/understand. Not a huge deal but figured I would call it out in case there is something that might fit better (another object?).

jfuss (changes requested):
Just a couple small comments. Love the refactor!

samcli/lib/observability/cw_logs/cw_log_formatters.py
mndeveci: I understand your concern, this was coming from the old code and I assume that is here in case string is not a valid json, so we don't want to fail. Let me see if I can tighten the exception a little bit.

jfuss (approved):

samcli/lib/observability/cw_logs/cw_log_formatters.py
sriram-mv:
```
NODEJS_CRASH_MESSAGE
```

sriram-mv (approved):
Overall looks good!

Issue #2846, opened 2021-04-30
Multiple lambda image application templates found. This should not be possible, please raise an issue


vsimon:
### Description:

`Error: Multiple lambda image application templates found. This should not be possible, please raise an issue.`

### Steps to reproduce:

`brew install aws/tap/aws-sam-cli`

```
sam --version
SAM CLI, version 1.23.0
```

`sam init --name foo --package-type Image --base-image amazon/python3.8-base`

### Observed result:

```
sam init --name foo --package-type Image --base-image amazon/python3.8-base

Cloning app templates from https://github.com/aws/aws-sam-cli-app-templates
Error: Multiple lambda image application templates found. This should not be possible, please raise an issue.
```

### Expected result:

Not an error?

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: OSX
2. `sam --version`: 1.23.0
3. AWS region: us-east-1

`Add --debug flag to command you are running`

```
sam init --name foo --package-type Image --base-image amazon/python3.8-base --debug
2021-04-30 15:26:45,011 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2021-04-30 15:26:45,012 | Using config file: samconfig.toml, config environment: default
2021-04-30 15:26:45,012 | Expand command line arguments to:
2021-04-30 15:26:45,012 | --name=foo --package_type=Image --base_image=amazon/python3.8-base --output_dir=.
2021-04-30 15:26:45,013 |
Cloning app templates from https://github.com/aws/aws-sam-cli-app-templates
2021-04-30 15:26:47,252 | Removing old templates from /Users/foo/.aws-sam/aws-sam-cli-app-templates
2021-04-30 15:26:47,461 | Copying templates from /var/folders/1f/h901r68j4q7_j55p0njzl6hw0000gn/T/tmpb3u1gjat/aws-sam-cli-app-templates to /Users/foo/.aws-sam/aws-sam-cli-app-templates
[...]

PR #2864, opened 2021-05-10
sam init - Enable --app-template argument for Image package-type

elbayaaa:
#2846 and aws-sam-cli-app-templates#95

To fix a known bug where `sam init` raises an exception if the user is trying to generate a new SAM project from the managed project templates for Image package-type of a base-image having multiple templates (like; [python3.8](https://github.com/aws/aws-sam-cli-app-templates/blob/master/manifest.json#L578-L614) and [Java11](https://github.com/aws/aws-sam-cli-app-templates/blob/master/manifest.json#L633-L648))

Although the `sam init` command already has an --app-template argument that is used to select on of the available project templates, there is an explicit logic to raise an error if there are multiple project templates in case of Image package-type. **This change is to not raise this explicit error if  the --app-template argument is provided.**

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/init/__init__.py
```diff
@@ -277,13 +277,14 @@ def do_cli(
         if package_type == IMAGE and image_bool:
             base_image, runtime = _get_runtime_from_image(base_image)
```
sriram-mv, line 278: are we now adding a requirement on both base image and app template always needing to be specified. Am I summarizing that correctly.

sriram-mv (approved):

hoffa (approved):

PR #2867, opened 2021-05-10
Generic Metric Tracker

hawflau:
This PR implements a generic metric tracker which makes tracking metric easier and more standardized.
Also enforces that only registered metric name (via `MetricName` class) can be tracked.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/build/command.py
```diff
@@ -205,6 +205,14 @@ def cli(
     """
     # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
 
+    track_metric(
```
sriram-mv, line 208: can we move telemetry related code to its own decorator?

samcli/lib/telemetry/metric.py
```diff
@@ -29,16 +30,120 @@
...
+
+
+class MetricName(Enum):
+    # Metrics that can be reported only once in a session
+    installed = ("installed", True)
+    commandRun = ("commandRun", True)
+
+    # Metrics that can be reported any number of times during a session
+    # e.g. docker_image_download_status
+    templateWarning = ("templateWarning", False)
```
sriram-mv, line 43: is the idea to continuously track certain things throughout the session, I'm assuming these are no longer anything to do with the command that was used to run SAM CLI then, since those options and flags are static once they are invoked.

samcli/lib/telemetry/metric.py
```diff
@@ -29,16 +30,120 @@
...
+
+def flush():
+    """
+    Emits all metrics to telemetry and flush the metric sink
+    """
+    telemetry = Telemetry()
+    metric_sink = _get_global_metric_sink()
+    for metric_list in metric_sink.values():
+        while metric_list:
+            metric = metric_list.pop(0)
```
sriram-mv, line 69: is this stack based implementation because we inherently value the latest information more than what was first tracked within the session? if that is the case, lets reason it out and add a comment, could help shape our design philosphy.

samcli/lib/telemetry/metric.py
```diff
@@ -29,16 +30,120 @@
...
+    All metrics collected in the global metric sink will be emitted at program exit
+    (by calling flush(), see samcli/cli/main.py)
+
+    Examples
+    --------
+    tracking metric with metric-specific attributes
+    >>> track_metric(MetricName.commandRun, {"foo": "bar"})
+    >>> track_metric(MetricName.commandRun, metric_specific_attrs={"foo": "bar"})
+
+    overwriting exisitng metric-specific attributes
```
sriram-mv, line 92: Would we ever over-write, what are the conditions under which this would be used, I notice we also mention a note of caution while using these APIs, how destructive are they?

samcli/lib/telemetry/metric.py
```diff
@@ -29,16 +30,120 @@
...
 
 
 def send_installed_metric():
     LOG.debug("Sending Installed Metric")
 
     telemetry = Telemetry()
-    metric = Metric("installed")
-    metric.add_data("osPlatform", platform.system())
-    metric.add_data("telemetryEnabled", bool(GlobalConfig().telemetry_enabled))
+    # NOTE: not using track_metric here because we want to emit the metric right away
```
sriram-mv, line 143: we should note that the concept of a session does not apply here. Good call out.

samcli/lib/telemetry/metric.py
```diff
@@ -138,15 +242,21 @@ def wrapped(*args, **kwargs):
...
+                    "debugFlagProvided": bool(ctx.debug),
+                    "region": ctx.region or "",
+                    "commandName": ctx.command_path,
+                    "duration": duration_fn(),
+                    "exitReason": exit_reason,
+                    "exitCode": exit_code,
+                },
+                # NOTE: setting overwrite_existing_top_level_attrs to True since this is run after the commd logic runs
+                # the top_level_attrs tracked here takes precedence
+                overwrite_existing_top_level_attrs=True,
```
sriram-mv, line 258: I still dont fully understand this, is there a order in which the track_metric gets called, where the overwriting is safe?

samcli/lib/telemetry/metric.py
```diff
@@ -304,33 +312,26 @@ def __init__(self, metric_name, should_add_common_attributes=True):
...
         self._data["requestId"] = str(uuid.uuid4())
         self._data["installationId"] = self._gc.installation_id
         self._data["sessionId"] = self._session_id
         self._data["executionEnvironment"] = self._get_execution_environment()
         self._data["ci"] = bool(self._cicd_detector.platform())
         self._data["pyversion"] = platform.python_version()
         self._data["samcliVersion"] = samcli_version
+        # metricSpecificAttributes is for holding attributes that are specific for that metric
+        # For example, if we want to track usage of warm containers, the corresponding attribute should go into here
+        self._data["metricSpecificAttributes"] = dict()
```
sriram-mv, line 334: Tracking additional parameters this way makes sense to me. 👍🏽

samcli/local/lambdafn/runtime.py
```diff
@@ -170,6 +169,12 @@ def invoke(
         timer = None
         container = None
         try:
+            # Track runtime
```
sriram-mv, line 172: Do you think a decorator still makes sense here?

sriram-mv (changes requested):
Have not gone through the tests yet, I like the direction this is taking! I have a few questions and concerns on the implementation.

samcli/lib/telemetry/metric.py
hawflau: The `track_metric` here is called after all command behaviors complete. It shouldn't happen but we could mistakenly write some of the attributes (e.g. `region`) in previous `track_metric` calls throughout the session.

Thinking out loud - maybe we should have two `track_metric` methods instead:
* `track_metric` - safe and doesn't come with the capabilitiy of overwriting existing attributes. This should be used as a generic method used throughout the session
* `_track_metric_unsafe` - unsafe in a sense that it can overwrite existing attributes. It should only be used here, where overwriting is necessary

samcli/lib/telemetry/metric.py
hawflau: In the `track_command` decorator, certain attributes are added (e.g. `region`, `duration`) after the main command behaviors complete. Current `track_metric` method allows the possibility that these attributes to be added prematurely (e.g. when `track_metric` is called somewhere during the command session). 

Thinking a bit deeper, I think we should have a list of "protected attributes" that can be added by `track_metric`

samcli/lib/telemetry/metric.py
hawflau: There are two cases here:
1. metrics that can be reported at most once per session (e.g. `commandRun`) - in this case, the stack will always be of length one. 
2. metrics that can be reported more than once per session (e.g. `templateWarning` and `runtimeMetric`) - in this case, the `metric_list` works more like a queue than a stack (i.e. FIFO). This will preserve the chronicle order when they are emitted to the telemetry api.

I'll add a comment explaining this explicitly

samcli/lib/telemetry/metric.py
hawflau: Options and flags are static and can be accessed via the context object. But sometimes we want to track things that are created within the command actions. One example is for `runtimeMetric`, we want to track the runtime used for functions during a `local invoke` run.

samcli/commands/build/command.py
hawflau: yeah, I'll revert this part, and do another PR to extend `track_command` to track options

samcli/local/lambdafn/runtime.py
hawflau: I don't have strong preference towards decorator or just inline tracking. 
My main concern was mainly that the `capture_parameter` decorator was not very intuitive to read and use. 
I think we can think more about how the decorators for tracking metric should be like. Meanwhile, it looks more intuitive to me to use `track_metric` inline here

PR #2877, opened 2021-05-12
feat: Supports uncompression local layer zips in sam local

aahung:

#947

Support passing a local archive file in "AWS::Serverless::LayerVersion" in local invoke

We already handle when function's codeuri points to a zip file, this PR extend that to `ContentUri` of LayerVersion

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -254,7 +254,7 @@ class LambdaRuntime:
 
     def _get_code_dir(self, code_path: str) -> str:
         """
-        Method to get a path to a directory where the Lambda function code is available. This directory will
+        Method to get a path to a directory where the function/layer code is available. This directory will
         be mounted directly inside the Docker container.
 
         This method handles a few different cases for ``code_path``:
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -1,20 +1,21 @@
 """
 Classes representing a local Lambda runtime
 """
-
+import copy
 import os
 import shutil
 import tempfile
 import signal
 import logging
 import threading
-from typing import Optional
+from typing import Optional, Union, Dict
 
 from samcli.local.docker.lambda_container import LambdaContainer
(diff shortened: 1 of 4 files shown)


samcli/local/lambdafn/runtime.py
```diff
@@ -274,13 +276,34 @@ def _get_code_dir(self, code_path):
...
+            decompressed_dir: str = _unzip_file(code_path)
             self._temp_uncompressed_paths_to_be_cleaned += [decompressed_dir]
             return decompressed_dir
 
         LOG.debug("Code %s is not a zip/jar file", code_path)
         return code_path
 
+    def _unarchived_layer(self, layer: Union[str, Dict, LayerVersion]) -> Union[str, Dict, LayerVersion]:
+        """
+        If the layer's content uri points to a supported local archive file, use self._get_code_dir() to
```
c2tarun, line 288: Just wanted to confirm, this only works if layer zip is present on local?

aahung: yes, same as how we handle `Function` today

aahung: if customers provide a S3 zip url, it is not gonna work

samcli/local/lambdafn/runtime.py
```diff
@@ -250,7 +252,7 @@ def signal_handler(sig, frame):
         timer.start()
         return timer
 
-    def _get_code_dir(self, code_path):
+    def _get_code_dir(self, code_path: str) -> str:
         """
         Method to get a path to a directory where the Lambda function code is available. This directory will
```
hawflau, line 257: nit: update the docstring to tell that this is also used for Layer

hawflau (approved):

samcli/local/lambdafn/runtime.py
aahung: updated 👍

c2tarun (approved):

Issue #2859, opened 2021-05-06
sam build cache does not seem to work with CodeBuild


mateogianolio:
### Description:

Cache works locally on my computer (MacOS 11.3), but never in CodeBuild (aws/codebuild/amazonlinux2-x86_64-standard:3.0).

I have verified that the CodeBuild cache path is exactly the same for two subsequent builds, but `sam build` still says the cache is invalid.

I think it might have something to do with paths or symlinks, but not sure.

Sorry if this is the wrong place to report this.

### Steps to reproduce:

CodeBuild set to custom local cache (also tried with s3 cache, still does not work).

**buildspec.yml (modified for brevity)**

```yaml
version: 0.2

cache:
  paths:
    - '.aws-sam/**/*' # Cache `.aws-sam`
phases:
  build:
    commands:
      - ./build.sh <stack_name>
```

**build.sh (modified for brevity)**

```sh
stack_name=$1
region=eu-west-1

sam build \
  --region "$region" \
  --template-file "$stack_name/template.yaml" \
  --build-dir ".aws-sam/$stack_name/build" \
  --cache-dir ".aws-sam/$stack_name/cache" \
  --use-container \
  --cached \
  --parallel;
```

### Observed result:

CodeBuild mounts the cache path correctly (`<id>` is the cache id that I verified stays the same for subsequent builds):

```
[Container] Expanded cache path .aws-sam/**/*
[Container] MkdirAll: /codebuild/local-cache/custom/<id>/.aws-sam
[Container] Symlinking: /codebuild/output/srcXXXXXXXXX/src/github.com/<repo>/.aws-sam => /codebuild/local-cache/custom/<id>/.aws-sam
```

```
Unique function build definition found, adding as new (Function Build Definition: BuildDefinition(nodejs14.x, /codebuild/output/srcXXXXXXXXX/src/github.com/<repo>/<stack_name>/<codeuri>, Zip, , <uuid>, {}, {}, []), Function: Function(name='<redacted>', functionname='<redacted>', runtime='nodejs14.x', memory=None, timeout=5, handler='index.handler', imageuri=None, packagetype='Zip', imageconfig=None, codeuri='/codebuild/output/srcXXXXXXXXX/src/github.com/<repo>/<stack_name>/<codeuri>', environment=None, rolearn='<redacted>', layers=[], events=None, metadata=None, inlinecode=None, codesign_config_arn=None, stack_path=''))

...

[...]

PR #2878, opened 2021-05-12
fix: Make sam build cache robust by always using relative paths

aahung:

#2859

currently sam build stores absolute paths of function code in build.toml file. It needs both path and source code md5 to match to treat the cache as valid. However, in CI environment, the code might have different absolute path in different CI run. 

In this PR, all paths in `FunctionBuildDefinition` is converted to relative path (and this path will be saved to build.toml after each build) to make sam build cache portable.

- When customers upgrade SAM CLI, the cache might be invalidated once.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -191,8 +191,18 @@ class ApplicationBuilder:
 
         for function in functions:
             container_env_vars = self._make_env_vars(function, file_env_vars, inline_env_vars)
+
+            # In case of CI env, the project might have inconsistent absolute path,
+            # which will cause build cache to be invalidated.
+            # Here relative paths are always used in FunctionBuildDefinition (#2872).
+            codeuri = (
+                os.path.relpath(function.codeuri)
+                # use os.path.isabs() to opt-out URLs like file:///...
+                if isinstance(function.codeuri, str) and os.path.isabs(function.codeuri)
+                else function.codeuri
+            )
             function_build_details = FunctionBuildDefinition(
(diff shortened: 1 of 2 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -191,8 +191,18 @@ def _get_build_graph(
...
+            # Here relative paths are always used in FunctionBuildDefinition (#2872).
+            codeuri = (
+                os.path.relpath(function.codeuri)
+                # use os.path.isabs() to opt-out URLs like file:///...
+                if isinstance(function.codeuri, str) and os.path.isabs(function.codeuri)
+                else function.codeuri
+            )
             function_build_details = FunctionBuildDefinition(
-                function.runtime, function.codeuri, function.packagetype, function.metadata, env_vars=container_env_vars
+                function.runtime, codeuri, function.packagetype, function.metadata, env_vars=container_env_vars
```
mndeveci, line 205: Not totally sure, but having absolute path vs relative path might affect the build process. Can we store another field for relative path and use that one for validating the cache?

aahung: I forgot to bring that up. Before nested stack feature was introduced, the codeuri here was relative paths. (verified with 1.19.0)

In nested stack feature, codeuri needs to be resolved relative against its template file, and `click` always convert template file path to absolute paths, which led to the current behavior

aahung:
@xazhao is working on another bug fix that makes this PR obsolete, closing

Issue #2832, opened 2021-04-27
Stack error AttributeError: 'collections.OrderedDict' object has no attribute 'startswith' for cdk generated template


stmfcgcg:
### Description:
I have large project made up of 50 lambas in nested stacks. I have generated the template.yml from a cdk project (see attached template.yml). When I run 'sam local start-api' the application does not run and displays a stack trace (see attachment)

### Steps to reproduce:
run 'sam local start-api' with the attached template.yml

### Observed result:
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/samcli/lib/telemetry/metric.py", line 153, in wrapped
[...]

PR #2879, opened 2021-05-12
fix: Fix crash when nested CFN stack has dict TemplateURL (unresolved intrinsics)

aahung:

#2832

This PR does not fix the use case in #2832, but rather make SAM CLI handle well when nested CFN stack's `TemplateURL` is a dictionary (pointing to a stack on SAR). This makes it consistent with how we handle nested serverless applications:
https://github.com/aws/aws-sam-cli/blob/9e03633dc6f8c5bdaef3b88be67314f00c28d10d/samcli/lib/providers/sam_stack_provider.py#L128-L145

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/sam_stack_provider.py
```diff
@@ -163,8 +163,14 @@ def _convert_cfn_stack_resource(
         resource_properties: Dict,
         global_parameter_overrides: Optional[Dict] = None,
     ) -> Optional[Stack]:
-        template_url = resource_properties.get("TemplateURL", "")
+        template_url = resource_properties.get("TemplateURL")
 
+        if isinstance(template_url, dict):
+            # This happens when TemplateURL has unresolvable intrinsic functions
+            # and it usually happens in CDK generated template files (#2832).
```
hawflau, line 170: wondering if we can show a message to tell customers to try `sam-beta-cdk` ([link](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-cdk.html)

aahung: Would that solve this problem?

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -163,8 +163,14 @@ def _convert_cfn_stack_resource(
         resource_properties: Dict,
         global_parameter_overrides: Optional[Dict] = None,
     ) -> Optional[Stack]:
-        template_url = resource_properties.get("TemplateURL", "")
+        template_url = resource_properties.get("TemplateURL")
 
+        if isinstance(template_url, dict):
+            # This happens when TemplateURL has unresolvable intrinsic functions
```
hoffa, line 169: > This happens when TemplateURL has unresolvable intrinsic functions

Why?

aahung: unresolved intrinsic functions have type "dict"

hoffa (approved):

hawflau (approved):

Issue #2747, opened 2021-03-21
samconfig.toml is ignored when running with a custom build directory


eyalroth:
### Description:
`samconfig.toml` is ignored when running with a custom build directory, probably due to its path being relative to the custom build directory instead of the root directory.

### Steps to reproduce:
1. Create a new project using `sam init` with the `Hello World Example`:
```
$ sam init
1 - AWS Quick Start Templates
1 - Zip (artifact is a zip uploaded to S3)
8 - nodejs10.x
Project name [sam-app]: sam-test
1 - Hello World Example
```
2. Add a simple `console.log(process.env.Foo);` to `hello-world/app.js`.
3. Add an environment variable `Foo` which points to parameter `Bar` in the function in the `template.yml`:
```yml
Environment:
  Variables:
    Foo: !Ref Bar
```
4. Create the following `samconfig.toml`:
```toml
version=0.1
[default.local_invoke.parameters]
parameter_overrides="Bar=456"
```
5. Build with a custom build directory (`sam build -b out/build`).
6. Local invoke (`sam local invoke -t out/build/template.yaml`).

### Observed result:
Environment variable read by the lambda is the default (`Bar`) instead of the set parameter (`456`).

This is not a specific problem with environment variables nor override parameters. This also happened with `--log-file` and I would assume with other command parameters as well. It's simply a bit more visible with this setup.

This does actually work in the following cases:
1. `--config-file` is set to the absolute path of the configuration file or a path relative to the custom build directory.
2. `local invoke` is executed without `-t out/build/template.yaml` (which differs from the instruction outputted by the `build` command).

```
sam build -b out/build --debug
2021-03-21 16:58:50,484 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2021-03-21 16:58:50,541 | 'build' command is called
2021-03-21 16:58:50,548 | No Parameters detected in the template
2021-03-21 16:58:50,574 | 2 stacks found in the template
2021-03-21 16:58:50,574 | No Parameters detected in the template
2021-03-21 16:58:50,595 | 2 resources found in the stack
2021-03-21 16:58:50,595 | No Parameters detected in the template
2021-03-21 16:58:50,630 | No Parameters detected in the template
2021-03-21 16:58:50,651 | Found Serverless function with name='HelloWorldFunction' and CodeUri='./hello-world/'
2021-03-21 16:58:50,651 | No Parameters detected in the template
[...]

PR #2891, opened 2021-05-19
samconfig debug level logging fixed; documentation updated

ssenchenko:

#2747

Current SAM CLI behaviour while looking up the `samconfig.toml` is confusing for customers.

- printing out configuration file location or a warning if `samconfig.toml` hasn't been found for each command which involves configuration file parsing
- fixing debug level logging to provide accurate information on the configuration file parameters
- updating sam cli help messages
- updating configuration file doc

None found

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/cli/cli_config_file.py
```diff
@@ -57,14 +57,12 @@ def __call__(self, config_path, config_env, cmd_names):
 
         samconfig = SamConfig(config_file_dir, config_file_name)
 
-        # Enable debug level logging by environment variable "SAM_DEBUG"
-        if os.environ.get("SAM_DEBUG", "").lower() == "true":
```
ssenchenko, line 61: It never worked unless SAM_DEBUG was set manually apart from --debug flag because click doesn't create/set environment variables

samcli/cli/main.py
```diff
@@ -74,10 +74,6 @@ def print_cmdline_args(func):
     """
 
     def wrapper(*args, **kwargs):
-        if kwargs.get("config_file") and kwargs.get("config_env"):
```
ssenchenko, line 77: It was inaccurate. When the template file was used outside of the project source directory, instead of giving a warning that the config was not found, it output `samconfig.toml` which created an illusion that the file was found.

samcli/cli/cli_config_file.py
```diff
@@ -57,14 +57,12 @@ def __call__(self, config_path, config_env, cmd_names):
...
-        LOG.debug("Config file location: %s", samconfig.path())
-
-        if not samconfig.exists():
-            LOG.debug("Config file '%s' does not exist", samconfig.path())
+        # bringing samconfig file location up to info level,
+        # to improve UX and make it clear where we're looking for samconfig file
+        if samconfig.exists():
+            click.echo(f"Config file location: {samconfig.path()}")
+        else:
+            click.secho(f"Config file '{samconfig.path()}' does not exist", fg="yellow", err=True)
```
wchengru, line 65: The samconfig is not necessary in the project folder. I would prefer not to put it in error level.

samcli/cli/main.py
wchengru: I agree with your point that it gives an impression that the file is found. Here it gives the debugging information about the config file and config env. If the file is not found, the debug will show up later. I suggest changing the log to better describe this. Or combine it into the expanded cli parameters.

samcli/cli/cli_config_file.py
ssenchenko: agree

samcli/cli/options.py
```diff
@@ -20,8 +20,14 @@ def callback(ctx, param, value):
...
         return value
 
+    # NOTE: --debug option should be eager to be evaluated before other parameters and to set log level to DEBUG
+    # before any other option/parameter processing will require to outbut debug info.
+    # Otherwise parameters are evaluated according to their order and if --debug is specified at the end of the command
+    # some debug output can be lost
+    # https://click.palletsprojects.com/en/7.x/advanced/#callback-evaluation-order
     return click.option(
         "--debug",
+        is_eager=True,
```
wchengru, line 30: We are using is_eager in some other options. Can you please check the callback order is correct if more than one `is_eager` options?

ssenchenko: Yes. The order is okay. There is no problem in this case because `--debug` doesn't do anything but sets the log level.
Here is the list of eager options:
- help, version (by default)
- info
- [config-file](https://github.com/aws/aws-sam-cli/blob/4e1a9796d9538e37dacc170d550c4f5561e10f63/samcli/cli/cli_config_file.py#L244), [config-env](https://github.com/aws/aws-sam-cli/blob/4e1a9796d9538e37dacc170d550c4f5561e10f63/samcli/cli/cli_config_file.py#L265), and [the full block](https://github.com/aws/aws-sam-cli/blob/4e1a9796d9538e37dacc170d550c4f5561e10f63/samcli/cli/cli_config_file.py#L200)
- template-file
- guided

samcli/cli/cli_config_file.py
wchengru: `--debug` also loads from this envvar. In case this callback is executed before callback of `--debug`, the log level will not be properly set and here the debug log will not be shown.

samcli/cli/cli_config_file.py
ssenchenko: updated

samcli/cli/main.py
ssenchenko: I deleted this code because I added file directory echo when config file is parsed and because the config env name was [already in debug](https://github.com/aws/aws-sam-cli/blob/4e1a9796d9538e37dacc170d550c4f5561e10f63/samcli/cli/cli_config_file.py#L71). It didn't get to output without `SAM_DEBUG` but now it does.

samcli/cli/cli_config_file.py
ssenchenko: It used to be the case but now it is not. Now `--debug` callback is executed before. I kept [SAM_DEBUG](https://github.com/aws/aws-sam-cli/blob/4e1a9796d9538e37dacc170d550c4f5561e10f63/samcli/cli/options.py#L27) as a back door though. Hopefully, we won't ever need it because it makes UX confusing

sriram-mv:
Overall this looks good, my only concern is if showcasing a warning on configuration file not being found might confuse users who are not using configuration files already. But overall, this is a good improvement.

samcli/cli/main.py
```diff
@@ -115,6 +111,15 @@ def cli(ctx):
 
     The AWS Serverless Application Model extends AWS CloudFormation to provide a simplified way of defining the
     Amazon API Gateway APIs, AWS Lambda functions, and Amazon DynamoDB tables needed by your serverless application.
+
+    SAM CLI commands run in the project root directory which is the directory with SAM template file
+    (template.{yml|yaml|json}). If no template file is specified explicitely, SAM CLI looks it up in the current
```
sriram-mv, line 116: explicitly

samcli/commands/_utils/options.py
```diff
@@ -217,7 +217,12 @@ def template_click_option(include_build=True):
         callback=partial(get_or_default_template_file_name, include_build=include_build),
         show_default=True,
         is_eager=True,
-        help="AWS SAM template which references built artifacts for resources in the template. (if applicable)"
+        help="AWS SAM template which references built artifacts for resources in the template. "
+        "Template file defines the root directory of the project and allows to point SAM CLI to the directory "
+        "for build, local invocation etc. If template file is not specified explicitly SAM CLI expects it to be "
+        "in the current working directory (where it is running). When you use config file and specify --template-file "
+        "SAM CLI expects samconfig.toml and the template file to be in the same directory."
```
sriram-mv, line 224: "When you use the default config file and specify --template-file SAM CLI expects samconfig.toml and the template file to be in the same directory. Alternatively, if --config-file is explicitly specified, it can point to a custom samconfig.toml location"

How does that sound?

samcli/cli/cli_config_file.py
```diff
@@ -236,8 +234,10 @@ def decorator_customize_config_file(f):
     config_file_param_decls = ("--config-file",)
     config_file_attrs["help"] = (
         "The path and file name of the configuration file containing default parameter values to use. "
-        "Its default value is 'samconfig.toml' in project directory. For more information about configuration files, "
-        "see: "
+        "Its default value is 'samconfig.toml' in project root directory. Project root directory is defined by the "
+        "template file location. When you use config file and specify --template-file SAM CLI expects samconfig.toml "
+        "and the template file to be in the same directory. Alternatively, use --config-file to point to samconfig.toml"
```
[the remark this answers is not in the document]
ssenchenko: Sounds better indeed. Thanks!

samcli/cli/main.py
ssenchenko: Thanks!

samcli/commands/_utils/options.py
```diff
@@ -217,7 +217,12 @@ def template_click_option(include_build=True):
         callback=partial(get_or_default_template_file_name, include_build=include_build),
         show_default=True,
         is_eager=True,
-        help="AWS SAM template which references built artifacts for resources in the template. (if applicable)"
+        help="AWS SAM template which references built artifacts for resources in the template. "
```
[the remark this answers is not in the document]
ssenchenko: okay

samcli/commands/_utils/options.py
ssenchenko: Thanks, that's better 😃

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -172,6 +172,8 @@ def _verify_invoke_built_function(self, template_path, function_logical_id, over
         process_execute.process.wait()
 
         process_stdout = process_execute.stdout.decode("utf-8")
+        if process_stdout.startswith("Config file"):
+            *_, process_stdout = process_stdout.partition("\n")
```
aahung, line 176: Do we need "if" here since "Config file" is always printed?

aahung (approved):
LGTM!

tests/integration/buildcmd/build_integ_base.py
ssenchenko: Well it's not exactly every time.  There are commands which don't use it so it's better be on the safe side

samcli/cli/main.py
```diff
@@ -115,6 +111,15 @@ def cli(ctx):
...
+    SAM CLI commands run in the project root directory which is the directory with SAM template file
+    (template.{yml|yaml|json}). If no template file is specified explicitly, SAM CLI looks it up in the current
+    working directory (where SAM CLI is running).
+
+    SAM CLI options can be either passed directly to the commands and/or stored in the config file (samconfig.toml),
+    which is expected to be in the project root directory by default. It is also possible to specify a custom directory
+    for the config file if necessary.
+
     You can find more in-depth guide about the SAM specification here:
     https://github.com/awslabs/serverless-application-model.
```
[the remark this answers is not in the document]
ssenchenko: changed

samcli/cli/cli_config_file.py
```diff
@@ -236,8 +234,10 @@ def decorator_customize_config_file(f):
     config_file_param_decls = ("--config-file",)
     config_file_attrs["help"] = (
         "The path and file name of the configuration file containing default parameter values to use. "
-        "Its default value is 'samconfig.toml' in project directory. For more information about configuration files, "
-        "see: "
+        "Its default value is 'samconfig.toml' in project root directory. Project root directory is defined by the "
```
sriram-mv, line 237: Final tweak, I think the help text being in 3rd person makes it sound more uniform, apologies, I didnt note this on my first pass. But I'm approving the PR, feel free to change the help text to be in 3rd person format.

sriram-mv (approved):

samcli/cli/cli_config_file.py
ssenchenko: Changing it. Thanks

samcli/commands/_utils/options.py
```diff
@@ -217,7 +217,12 @@ def template_click_option(include_build=True):
         callback=partial(get_or_default_template_file_name, include_build=include_build),
         show_default=True,
         is_eager=True,
-        help="AWS SAM template which references built artifacts for resources in the template. (if applicable)"
+        help="AWS SAM template which references built artifacts for resources in the template (if applicable). "
+        "Template file defines the root directory of the project and allows to point SAM CLI to the directory "
+        "for build, local invocation etc. If template file is not specified explicitly SAM CLI expects it to be "
```
aahung, line 222:
```suggestion
        "for build, local invocation, etc. If template file is not specified explicitly SAM CLI expects it to be "
```

samcli/commands/_utils/options.py
```diff
@@ -217,7 +217,12 @@ def template_click_option(include_build=True):
...
+        "in the current working directory (where it is running). When you use config file and specify --template-file "
+        "SAM CLI expects samconfig.toml and the template file to be in the same directory."
+        "Alternatively, if --config-file is explicitly specified, it can point to a custom samconfig.toml location."
```
aahung, line 225: "explicitly" seems unnecessary because `--config-file`, the CLI option, is either specified or not specified. There is no implicitly specified. 

```suggestion
        "Alternatively, if --config-file is specified, it can point to a custom samconfig.toml location."
```

aahung (approved):

PR #2892, opened 2021-05-20
fix: `sam validate` with local image

hoffa:

https://github.com/aws/aws-sam-cli/issues/2893

Create fake `ImageUri` when the image is referenced locally.

Tested with:

```bash
SAM_CLI_TELEMETRY=0 make func-test
```

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -120,6 +121,23 @@ def _replace_local_codeuri(self):
...
+        """
+        Adds fake ImageUri to AWS::Serverless::Functions that reference a local image using Metadata.
+        This ensures sam validate works without having to package the app or use ImageUri.
+        """
+        resources = self.sam_template.get("Resources", {})
+        for _, resource in resources.items():
+            resource_type = resource.get("Type")
+            properties = resource.get("Properties", {})
+
+            is_image_function = resource_type == "AWS::Serverless::Function" and properties.get("PackageType") == IMAGE
```
sriram-mv, line 134: should this be an enum we can look at? Otherwise this looks great 🎉

sriram-mv (changes requested):

samcli/commands/validate/lib/sam_template_validator.py
aahung: `AWS_SERVERLESS_FUNCTION` from samcli/commands/_utils/resources.py:8

aahung (approved):
LGTM

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -120,6 +122,23 @@ def _replace_local_codeuri(self):
...
+        """
+        Adds fake ImageUri to AWS::Serverless::Functions that reference a local image using Metadata.
+        This ensures sam validate works without having to package the app or use ImageUri.
+        """
+        resources = self.sam_template.get("Resources", {})
+        for _, resource in resources.items():
+            resource_type = resource.get("Type")
+            properties = resource.get("Properties", {})
+
+            is_image_function = resource_type == AWS_SERVERLESS_FUNCTION and properties.get("PackageType") == IMAGE
```
sriram-mv, line 135: Sorry, I missed this before, we need to consider AWS_LAMBDA_FUNCTION too?

sriram-mv (changes requested):

samcli/commands/validate/lib/sam_template_validator.py
hoffa: Not AFAIK since `AWS::Lambda::Function` doesn't have a notion of local images (and any pure CFN resources should pass validation).

sriram-mv: agree on pure CFN, however we do still build Image based AWS::Lambda::Function though IIRC.

hoffa: `sam validate` already works with any `AWS::Lambda::Function` such as:

```yaml
Transform: AWS::Serverless-2016-10-31
Resources:
  Function:
    Type: AWS::Lambda::Function
    Properties: Something
```

sriram-mv (approved):

PR #2894, opened 2021-05-20
Let 'sam pipeline init'  prefills pipeline's infrastructure resources…

elbayaaa:
#1713 

Better user experience for "sam pipeline init" by pre-loading the values of the infrastructure resources instead of having the user manually entering it

reading the saved values at pipeline.config.toml file 

N/A
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/config/samconfig.py
+++ b/samcli/lib/config/samconfig.py
@@ -44,7 +44,7 @@ class SamConfig:
     def get_env_names(self):
         self._read()
         if isinstance(self.document, dict):
-            return [env for env in self.document.keys() if env != VERSION_KEY]
+            return [env for env, value in self.document.items() if isinstance(value, dict)]
         return []
 
     def get_all(self, cmd_names, section, env=DEFAULT_ENV):
--- a/samcli/lib/config/samconfig.py
+++ b/samcli/lib/config/samconfig.py
@@ -41,6 +41,12 @@ class SamConfig:
         """
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
 
+    def get_env_names(self):
+        self._read()
+        if isinstance(self.document, dict):
+            return [env for env in self.document.keys() if env != VERSION_KEY]
+        return []
+
     def get_all(self, cmd_names, section, env=DEFAULT_ENV):
         """
         Gets a value from the configuration file for the given environment, command and section
--- a/tests/unit/lib/samconfig/test_samconfig.py
+++ b/tests/unit/lib/samconfig/test_samconfig.py
@@ -1,10 +1,9 @@
 import os
-import shutil
 from pathlib import Path
 from unittest import TestCase
 
 from samcli.lib.config.exceptions import SamConfigVersionException
-from samcli.lib.config.samconfig import SamConfig, DEFAULT_CONFIG_FILE_NAME, DEFAULT_GLOBAL_CMDNAME
+from samcli.lib.config.samconfig import SamConfig, DEFAULT_CONFIG_FILE_NAME, DEFAULT_GLOBAL_CMDNAME, DEFAULT_ENV
 from samcli.lib.config.version import VERSION_KEY, SAM_CONFIG_VERSION
 from samcli.lib.utils import osutils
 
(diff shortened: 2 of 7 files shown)


tests/unit/lib/samconfig/test_samconfig.py
```diff
@@ -28,14 +27,25 @@ def _check_config_file(self):
         self.assertTrue(self.samconfig.sanity_check())
         self.assertEqual(SAM_CONFIG_VERSION, self.samconfig.document.get(VERSION_KEY))
 
-    def _update_samconfig(self, cmd_names, section, key, value, env):
-        self.samconfig.put(cmd_names=cmd_names, section=section, key=key, value=value, env=env)
+    def _update_samconfig(self, cmd_names, section, key, value, env=None):
+        if env:
+            self.samconfig.put(cmd_names=cmd_names, section=section, key=key, value=value, env=env)
+        else:
+            self.samconfig.put(cmd_names=cmd_names, section=section, key=key, value=value)
```
aahung, line 34: Does this suffice?

```suggestion
        self.samconfig.put(cmd_names=cmd_names, section=section, key=key, value=value, env=env)
```

aahung (approved):
LGTM. Can we have a unit test for `_load_pipeline_bootstrap_context()`?

samcli/lib/config/samconfig.py
```diff
@@ -41,6 +41,12 @@ def __init__(self, config_dir, filename=None):
         """
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
 
+    def get_env_names(self):
+        self._read()
+        if isinstance(self.document, dict):
+            return [env for env in self.document.keys() if env != VERSION_KEY]
```
hoffa, line 47: Shouldn't it instead check that `env` is a dictionary (or whatever represents a TOML table)? Otherwise this will break if there's another top-level key.

elbayaaa:
> LGTM. Can we have a unit test for `_load_pipeline_bootstrap_context()`?

It is already there https://github.com/elbayaaa/aws-sam-cli/blob/sam-pipelines-init-load-bootstrapped-values/tests/unit/commands/pipeline/init/test_initeractive_init_flow.py#L161-L166

samcli/lib/config/samconfig.py
elbayaaa: TOML is like YAML; A hierarchy-based document. It is SAM who defined its own hierarchy to be: `[env / command_names / section / key]`. i.e. all the first-level keys in the toml file represents names of envs. (with one exception, the version).

hoffa: If tomorrow `samconfig.toml` gets a new top-level value `foo = "bar"`, this code breaks as it will include `foo` in the return values, no?

tests/unit/lib/samconfig/test_samconfig.py
elbayaaa: this will pass a `env=None` to `samconfig.put` while we want to not pass a value at all so that the default value ([DEFAULT_ENV](https://github.com/aws/aws-sam-cli/blob/e4853b39bf8011132d08a12fdbb8c2e18437b9d9/samcli/lib/config/samconfig.py#L19)) is used. see [here](https://github.com/aws/aws-sam-cli/blob/e4853b39bf8011132d08a12fdbb8c2e18437b9d9/samcli/lib/config/samconfig.py#L90)

aahung: Oh I see

samcli/lib/config/samconfig.py
elbayaaa: If [SamConfig.put](https://github.com/aws/aws-sam-cli/blob/e4853b39bf8011132d08a12fdbb8c2e18437b9d9/samcli/lib/config/samconfig.py#L90) method changed they [way it persists the configuration](https://github.com/aws/aws-sam-cli/blob/e4853b39bf8011132d08a12fdbb8c2e18437b9d9/samcli/lib/config/samconfig.py#L120-L131), this method should change accordingly.

hoffa: Right, so `env` is a top-level dictionary. Something like this should be robust to changes:

```python
[k for k, v in self.document.items() if isinstance(v, dict)]
```

elbayaaa: Ah, you are talking about how to exclude the `Version` key. I thought you are questioning the whole approach.

hoffa (approved):

PR #2911, opened 2021-05-28
feat: Support creating pipeline files directly into . without hooks

aahung:

Need to replace `APP_PIPELINE_TEMPLATES_REPO_URL` to run it.

```
APP_PIPELINE_TEMPLATES_REPO_URL = "https://github.com/aahung/aws-sam-cli-pipeline-init-templates-1.git"
```

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/pipeline/init/interactive_init_flow.py
+++ b/samcli/commands/pipeline/init/interactive_init_flow.py
@@ -17,6 +17,7 @@ from samcli.lib.cookiecutter.interactive_flow_creator import InteractiveFlowCrea
 from samcli.lib.cookiecutter.question import Choice
 from samcli.lib.cookiecutter.template import Template
 from samcli.lib.utils import osutils
+from samcli.lib.utils.colors import Colored
 from samcli.lib.utils.git_repo import GitRepo, CloneRepoException
 from samcli.lib.utils.osutils import copytree
 from .pipeline_templates_manifest import Provider, PipelineTemplateMetadata, PipelineTemplatesManifest
@@ -46,9 +47,9 @@ def do_interactive() -> None:
         generated_files = _generate_from_custom_location()
     else:
         generated_files = _generate_from_app_pipeline_templates()
-    click.echo("Successfully created the pipeline configuration file(s):")
+    click.secho(Colored().green("Successfully created the pipeline configuration file(s):"))
     for file in generated_files:
-        click.echo(f"  - {file}")
+        click.secho(Colored().green(f"  - {file}"))
 
 
 def _generate_from_app_pipeline_templates() -> List[str]:
--- a/samcli/commands/pipeline/init/interactive_init_flow.py
+++ b/samcli/commands/pipeline/init/interactive_init_flow.py
@@ -4,7 +4,6 @@ pipeline configuration file
 """
 import logging
 import os
-import tempfile
 from pathlib import Path
 from typing import Dict, List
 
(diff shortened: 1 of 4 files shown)


samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -18,6 +19,7 @@
 from samcli.lib.cookiecutter.template import Template
 from samcli.lib.utils import osutils
 from samcli.lib.utils.git_repo import GitRepo, CloneRepoException
+from samcli.lib.utils.osutils import copytree
```
elbayaaa, line 22: we have already imported `osutils` above and used it like `osutils.mkdir_temp`, can we unify the behavior? I vote for using `osutils.copytree`.

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -4,6 +4,7 @@
 """
 import logging
 import os
+import tempfile
```
elbayaaa, line 7: let's use [osutils.mkdir_temp](https://github.com/aws/aws-sam-cli/blob/aa8061481bf3a5238f0047fd6978702111a9ac36/samcli/lib/utils/osutils.py#L17)

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -107,7 +109,23 @@ def _generate_from_pipeline_template(pipeline_template_dir: Path) -> None:
     pipeline_template: Template = _initialize_pipeline_template(pipeline_template_dir)
     bootstrap_context: Dict = _load_pipeline_bootstrap_context()
     context: Dict = pipeline_template.run_interactive_flows(bootstrap_context)
-    pipeline_template.generate_project(context)
+    with tempfile.TemporaryDirectory() as tempdir:
+        LOG.debug("Generating pipeline files into %s", tempdir)
+        context["outputDir"] = "."  # prevent cookiecutter from generating a sub-folder
```
elbayaaa, line 114: we always set the "outputDir" to "." https://github.com/aws/aws-sam-cli/blob/ae4b8b64fbeb5c87611d64eb0786a702a224fdf7/samcli/lib/cookiecutter/template.py#L149

samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -107,7 +109,23 @@ def _generate_from_pipeline_template(pipeline_template_dir: Path) -> None:
...
+        pipeline_template.generate_project(context, tempdir)
+        _copy_dir_contents_fail_on_exist(tempdir, ".")
```
elbayaaa, line 116: better to use ` os.getcwd()` ?

samcli/commands/pipeline/init/interactive_init_flow.py
aahung: these are two different things. cookiecutter has an `output_dir` and then our template has an `outputDir`.

The former is where the latter is created in.

samcli/commands/pipeline/init/interactive_init_flow.py
aahung: changed

samcli/commands/pipeline/init/interactive_init_flow.py
aahung: No idea whether either of which is better. I changed the method name and use cwd

aahung: changing back to "."

I added the feature to print out the list of generated files. `os.getcwd` will print out absolute paths and it looks verbose on a list of file paths.

aahung: image

hoffa (approved):

elbayaaa (approved):

Issue #2923, opened 2021-06-02
Global parameters together with confirm_changeset results in AttributeError 


lobbin:
### Description:

Adding a global parameters section, such as "[default.global.parameters]" to samconfig.toml makes the sam-cli break with AttributeError if confirm_changeset is present in deploy configuration.

### Steps to reproduce:

sam init, choice: 1, 1, 7, 1

Provide default samconfig.toml with the following content
```
version = 0.1

[default.global.parameters]

[default.deploy.parameters]
stack_name = "Test"
confirm_changeset = true
```

### Observed result:

```
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1257, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 700, in make_context
    self.parse_args(ctx, args)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1048, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1623, in handle_parse_result
    value = self.full_process_value(ctx, value)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1965, in full_process_value
    return Parameter.full_process_value(self, ctx, value)
  File "/usr/local/Cellar/aws-sam-cli/1.23.0/libexec/lib/python3.8/site-packages/click/core.py", line 1589, in full_process_value
    value = self.process_value(ctx, value)
[...]

PR #2929, opened 2021-06-03
chore: Update requests to 2.25.1 to remove the locking on urllib3 to 1.25

mgrandis:
#### Checklist

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #2939, opened 2021-06-07
Updating tomlkit version as we need fix of the dataloss bug during copy() method use on Table object

c2tarun:

#2923 

We use copy() method when trying to parse tomlconfig. The updated version has a fix which prevents values during copy operation.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


aahung (approved):

requirements/reproducible-linux.txt
```diff
@@ -296,9 +296,9 @@ text-unidecode==1.3 \
     --hash=sha256:1311f10e8b895935241623731c2ba64f4c455287888b18189350b67134a822e8 \
     --hash=sha256:bad6603bb14d279193107714b288be206cac565dfa49aa5b105294dd5c4aab93
     # via python-slugify
-tomlkit==0.7.0 \
-    --hash=sha256:6babbd33b17d5c9691896b0e68159215a9387ebfa938aa3ac42f4a4beeb2b831 \
-    --hash=sha256:ac57f29693fab3e309ea789252fcce3061e19110085aa31af5446ca749325618
+tomlkit==0.7.2 \
```
jfuss, line 299: We need to make sure we update our THIRD-PARTY-LICENSES file: https://github.com/aws/aws-sam-cli/blob/develop/installer/assets/THIRD-PARTY-LICENSES#L2490

jfuss (changes requested):
The changes seem reasonable. Should we have some additional integ tests so we can make sure this behavior isn't regressed in the future?

c2tarun:
> The changes seem reasonable. Should we have some additional integ tests so we can make sure this behavior isn't regressed in the future?

Sure, added integ test.

> We need to make sure we update our THIRD-PARTY-LICENSES file: https://github.com/aws/aws-sam-cli/blob/develop/installer/assets/THIRD-PARTY-LICENSES#L2490

I think this file wasn't updated in some time. Updating it now, several updates in this file.

jfuss:
We recently updated the THRID-PARTY-LICENSES in https://github.com/aws/aws-sam-cli/pull/2929/files. Looks like you might not have had an updated version of `develop` given the conflicts.

jfuss (approved):

sriram-mv (approved):

PR #2941, opened 2021-06-10
fix: Allow the base64Encoded field in REST Api, skip validation of unknown fields and validate missing statusCode for Http Api

moelasmar:
fix API Gateway emulator: 
 - skip validating the non allowed fields for Http Api Gateway, as it always skip the unknown fields
 - add base64Encoded as an allowed field for Rest Api gateway
 - base64 decoding will be always done for Http API gateway if the lambda response isBase64Encoded is true regardless the content-type
 - validate if statusCode is missing in case of Http API, and payload version 1.0
 
#### Checklist

- [x] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -405,7 +409,14 @@ def _parse_v1_payload_format_lambda_output(lambda_output: str, binary_types, fla
         body = json_output.get("body")
         if body is None:
             LOG.warning("Lambda returned empty body!")
-        is_base_64_encoded = json_output.get("isBase64Encoded") or False
+
+        # Http API gateway checks lambda response for isBase64Encoded field, and ignore base64Encoded
+        # Rest API gateway checks first the field base64Encoded field, if not exist, it checks isBase64Encoded field
```
jfuss, line 414: Is this true? Docs suggests isBase64Encoded is only supported: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

jfuss (changes requested):

samcli/local/apigw/local_apigw_service.py
moelasmar: based on my trials that I found the following (as I could not find in the documentation anything related if APi gateway will fail if the lambda response contains any unknown field or not)
- for HTTP API Gateway just skip any un known fields 
- for Rest API, it check also for base64Encoded and fail if there is any other unknown fields .. and for doing the decoding it depends on the value of base64Encoded first, then isBase64Encoded ... which means if both fields are there and base64Encoded = True, and isBase64Encoded = False .. Api gateway will try to decode the response body

samcli/local/apigw/local_apigw_service.py
```diff
@@ -405,7 +409,14 @@ def _parse_v1_payload_format_lambda_output(lambda_output: str, binary_types, fla
...
+        is_base_64_encoded = (
+            json_output.get("base64Encoded")
```
jfuss, line 416: Do we need to guard against if `base64Encoded` or `isBase64Encoded` is not a`bool`? The if will catch if `base64Encoded` is set to a non `None` object like "true" or "1". Super edge case-y but figured I would at least ask.

tests/integration/testdata/start_api/main.py
```diff
@@ -80,6 +80,17 @@ def base64_response(event, context):
...
+
+    return {
+        "statusCode": 200,
+        "body": gifImageBase64,
+        "isBase64Encoded": False,
```
jfuss, line 89: Should we have a test for the `base64Encoded` case in the response?

samcli/local/apigw/local_apigw_service.py
moelasmar: Done

tests/integration/testdata/start_api/main.py
moelasmar: done

samcli/local/apigw/local_apigw_service.py
moelasmar: done

jfuss (approved):
🕺

wchengru (approved):

tests/integration/testdata/start_api/main.py
```diff
@@ -80,6 +80,40 @@ def base64_response(event, context):
...
...Xp1rWlMc6awJjiAAd2fm4ogXjz56aypOoIde4OE5u/F9x199dlXnnGiHZWEYbGpsAEA3QXYnHwEFliKAgswgJ8LPeiUXGwedCAKABACCN+EA1pYIIYaFlcDhytd51sGAJbo3onOpajiihlO92KHGaUXGwWjUBChjSPiWJuOO/LYIm4v1tXfE6J4gCSJEZ7YgRYUNrkji9P55sF/ogxw5ZkSqIDaZBV6aSGYq/lGZplndkckZ98xoICbTcIJGQAZcNmdmUc210hs35nCyJ58fgmIKX5RQGOZowxaZwYA+JaoKQwswGijBV4C6SiTUmpphMspJx9unX4KaimjDv9aaXOEBteBqmuuxgEHoLX6Kqx+yXqqBANsgCtit4FWQAEkrNbpq7HSOmtwag5w57GrmlJBASEU18ADjUYb3ADTinIttsgSB1oJFfA63bduimuqKB1keqwUhoCSK374wbujvOSu4QG6UvxBRydcpKsav++Ca6G8A6Pr1x2kVMyHwsVxUALDq/krnrhPSOzXG1lUTIoffqGR7Goi2MAxbv6O2kEG56I7CSlRsEFKFVyovDJoIRTg7sugNRDGqCJzJgcKE0ywc0ELm6KBCCJo8DIPFeCWNGcyqNFE06ToAfV0HBRgxsvLThHn1oddQMrXj5DyAQgjEHSAJMWZwS3HPxT/QMbabI/iBCliMLEJKX2EEkomBAUCxRi42VDADxyTYDVogV+wSChqmKxEKCDAYFDFj4OmwbY7bDGdBhtrnTQYOigeChUmc1K3QTnAUfEgGFgAWt88hKA6aCRIXhxnQ1yg3BCayK44EWdkUQcBByEQChFXfCB776aQsG0BIlQgQgE8qO26X1h8cEUep8ngRBnOy74E9QgRgEAC8SvOfQkh7FDBDmS43PmGoIiKUUEGkMEC/PJHgxw0xH74yx/3XnaYRJgMB8obxQW6kL9QYEJ0FIFgByfIL7/IQAlvQwEpnAC7DtLNJCKUoO/w45c44GwCXiAFB/OXAATQryUxdN4LfFiwgjCNYg+kYMIEFkCKDs6PKAIJouyGWMS1FSKJOMRB/BoIxYJIUXFUxNwoIkEKPAgCBZSQHQ1A2EWDfDEUVLyADj5AChSIQW6gu10bE/JG2VnCZGfo4R4d0sdQoBAHhPjhIB94v/wRoRKQWGRHgrhGSQJxCS+0pCZbEhAAOw=="  # NOQA
```
mgrandis, line 84: You may want to put this in a constant to avoid repetition.
Also maybe use a smaller input file, like a 1x1 white px image

moelasmar: done

mgrandis (approved):

Issue #2885, opened 2021-05-17
Cache is invalid when using containers environment variables


wnkz:
### Description:

Cache is always invalid when building with containers environment variables (`-e`, `--container-env-var`)

### Steps to reproduce:

- Create a any sample app with `sam init`
- Build multiple times with `sam build -cu -e FOO=BAR`

### Observed result:

- "Cache is invalid" message
- Function's hash is changing

```
❯ sam build -cu -e FOO=BAR
Starting Build use cache
Starting Build inside a container
Cache is invalid, running build and copying resources to function build definition of b89956ad-7a89-401a-98b5-68dad76c9c2d
Building codeuri: /private/var/folders/bf/f_z6y6n959193x3csh_pcb0m0000gn/T/tmp.Zjp9bgzU/sam-app/hello_world runtime: python3.8 metadata: {} functions: ['HelloWorldFunction']

Fetching public.ecr.aws/sam/build-python3.8:latest Docker container image......
Mounting /private/var/folders/bf/f_z6y6n959193x3csh_pcb0m0000gn/T/tmp.Zjp9bgzU/sam-app/hello_world as /tmp/samcli/source:ro,delegated inside runtime container

Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Invoke Function: sam local invoke
[*] Deploy: sam deploy --guided

Running PythonPipBuilder:ResolveDependencies
Running PythonPipBuilder:CopySource
❯ sam build -cu -e FOO=BAR
Starting Build use cache
Starting Build inside a container
Cache is invalid, running build and copying resources to function build definition of fb423804-b45f-431a-9eca-e5d58909cbe6
Building codeuri: /private/var/folders/bf/f_z6y6n959193x3csh_pcb0m0000gn/T/tmp.Zjp9bgzU/sam-app/hello_world runtime: python3.8 metadata: {} functions: ['HelloWorldFunction']

Fetching public.ecr.aws/sam/build-python3.8:latest Docker container image......
Mounting /private/var/folders/bf/f_z6y6n959193x3csh_pcb0m0000gn/T/tmp.Zjp9bgzU/sam-app/hello_world as /tmp/samcli/source:ro,delegated inside runtime container

Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Invoke Function: sam local invoke
[...]

PR #2943, opened 2021-06-10
fix: pass copy of environment variables for keeping cache valid

mndeveci:
https://github.com/aws/aws-sam-cli/issues/2885

When building inside the container, we add additional environment variables like `LAMBDA_BUILDERS_LOG_LEVEL` which is causing cache to be invalid all the time. 

We are sending a copy of the dictionary instead of reference to original one, so additional changes won't affect function build definition and cache information.

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss (changes requested):
Integ tests to strengthen our testing suite against future regressions?

mndeveci:
> Integ tests to strengthen our testing suite against future regressions?

@jfuss good point, added integ tests 👍🏼

jfuss (approved):

qingchm (approved):
LGTM

Issue #2665, opened 2021-02-26
Lambda local artifact not re-uploaded after file permission only change


kftsehk:
### Description:
When lambda function encounter `EACCES: permission denied` error
```
    "Error: EACCES: permission denied, open '/var/task/index.js'",
    "    at Object.openSync (fs.js:462:3)",
    "    at Object.readFileSync (fs.js:364:35)",
    "    at Object.Module._extensions..js 
```

The fix is to change the file permission `chmod a+r * -R`, this can lead to a permission-only change to the lambda source code.

However, the current hashing method of lambda source code directory cannot detect permission-only change, thus will skip re-upload code artifact

The deployment will then continue using previously uploaded artifact having incorrect file permissions and still report successful. This leads to unexpected behavior that deployed lambda package have file permission different from local files.

### Steps to reproduce:
1. Create a lambda function
2. `sam deploy --guided`
3. Update _only_ file permissions of the lambda's code
4. `sam deploy --debug`

### Observed result:
```
2021-02-26 19:49:19,778 | File with same data is already exists at stackname/d3a727670c47aa78e08d049c72647445. Skipping upload
```

### Expected result:
A new package with updated permission is uploaded and deployed.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: RHEL8
2. `sam --version`: 1.18.2 (pypi)
3. AWS region: ap-northeast-1

### Suggested fix

A straight forward fix is to calculate checksum _after_ zipping.
Zipping will record file permissions, hashing the zip can thus detect file permission change.

https://github.com/aws/aws-sam-cli/blob/5cbe2db23e6327f545a4e328b35bf17a3446e3d0/samcli/lib/package/utils.py#L159-L175

### Supplementary information

`dir_checksum` does not reflect permission change in files

```
>>> import hash
>>> hash.dir_checksum(".")
'3e19f2ac12731fdc3d7321b3fff0fcba'

(base) [kftse@DESKTOP-J8ITE1P utils]$ chmod og-r * 

>>> import hash
>>> hash.dir_checksum(".")
'3e19f2ac12731fdc3d7321b3fff0fcba'
```

Hash after zip seems to work

```

(base) [kftse@DESKTOP-J8ITE1P lib]$ zip utils_ur.zip utils/ -r
  adding: utils/ (stored 0%)
  ...
  adding: utils/version_checker.py (deflated 64%)
  adding: utils/__pycache__/ (stored 0%)
[...]

PR #2946, opened 2021-06-11
feat: Increase awareness of same file warning during package

mndeveci:
https://github.com/aws/aws-sam-cli/issues/2665

If same package is tried to be packaged and/or deployed, same file exists message is printed as `debug` where it makes harder to be visible.

By changing the `debug` into `info`

N/A

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/package/s3_uploader.py
+++ b/samcli/lib/package/s3_uploader.py
@@ -85,7 +85,7 @@ class S3Uploader:
 
         # Check if a file with same data exists
         if not self.force_upload and self.file_exists(remote_path):
-            LOG.info("File with same data is already exists at %s. " "Skipping upload", remote_path)
+            LOG.info("File with same data already exists at %s, skipping upload", remote_path)
             return self.make_url(remote_path)
 
         try:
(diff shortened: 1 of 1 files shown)


wchengru (approved):

samcli/lib/package/s3_uploader.py
```diff
@@ -85,7 +85,7 @@ def upload(self, file_name: str, remote_path: str) -> str:
 
         # Check if a file with same data exists
         if not self.force_upload and self.file_exists(remote_path):
-            LOG.debug("File with same data is already exists at %s. " "Skipping upload", remote_path)
+            LOG.info("File with same data is already exists at %s. " "Skipping upload", remote_path)
```
mgrandis, line 88: Let's fix the message too: `A file with the same name already exists at %s`
- It seems we are checking the file path, not its content/data
- Remove the final `.`
- Fix the grammar

mndeveci: Thanks @mgrandis ;
- The filename is generated md5 from its contents at line 140, so checking its name is basically checking its contents
- Done
- Done

mgrandis (approved):

PR #2958, opened 2021-06-15
SAM Init Command interactive flow Redesign

jonife:
Improve the user experience on the sam init interactive experience. Today we "hide" the types of app we support through a layer of questions (custom or pre made examples, Package type for lambda, what runtime, then show apps). While this works, we are moving to the idea of surfacing what examples we have first. 

The new flow is something like: 

custom or pre-made ---> here are the types of examples we have ----> from that list here are the package types ---> from that here are the runtimes. 

Improve the user experience on the sam init interactive experience

Changed the entire cli interactive flow

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes


samcli/commands/init/init_templates.py
```diff
@@ -171,3 +173,46 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+        """
+        This method get the manifest cloned from the git repo and preprocessed it.
+
+        Raises:
+            InvalidInitTemplateError: Exception raise when template is invalid or missing a parameter
+
+        Returns:
+            dict: This is preprocessed manifest with the use_case as key
+        """
+        manifest_dump = requests.get(self._manifest_url)
```
sriram-mv, line 187: we could for check for response object instead 

```
resp = requests.get(self._manifest_url)

if not resp.ok:
    ....
```

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +173,46 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+            raise InvalidInitTemplateError(
+                "Can't retrieve templates manifest from {url}".format(url=self._manifest_url)
+            )
+        preprocessed_manifest = {}
+        manifest_body = manifest_dump.json()
+
+        for template_runtime in manifest_body:
+            template_list = manifest_body[template_runtime]
+            for template in template_list:
+                if "useCaseName" not in template:
```
sriram-mv, line 198: useCaseName is an additional field to existing dictionary within a list for a given runtime, so backwards compat should be okay here.

Could we add a comment on how the expected manifest structure needs to be like for this function?

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +173,46 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+                    raise InvalidInitTemplateError(
+                        "Template {name} missing use case descriptor".format(name=template["displayName"])
+                    )
+                use_case = template["useCaseName"]
+                if use_case not in preprocessed_manifest:
```
sriram-mv, line 203: Instead of multiple if statements could we adopt a matcher pattern, so that we can supply a list of matchers (eg: predicates) that would short circuit the error process, makes the code easier to read. Agree on the fact the sam init codebase has some tech debt, we can fix it a little bit at a time. :)

tests/integration/init/schemas/test_init_with_schemas_command.py
```diff
@@ -55,29 +56,29 @@ def test_init_interactive_with_event_bridge_app_partner_registry(self):
...
 eb-app-maven
-3
 Y
 3
 1
         """
         with tempfile.TemporaryDirectory() as temp:
             runner = CliRunner()
             result = runner.invoke(init_cmd, ["--output-dir", temp], input=user_input)
+            breakpoint()
```
sriram-mv, line 81: Looks to have snuck in.

tests/unit/commands/init/test_cli.py
```diff
@@ -17,7 +17,12 @@
 
 
 class MockInitTemplates:
-    def __init__(self, no_interactive=False):
+    def __init__(self, no_interactive=False, auto_clone=True):
+        self._manifest_url = "https://raw.githubusercontent.com/sapessi/aws-sam-cli-app-templates/master/manifest.json"
```
sriram-mv, line 21: Will there be a PR out to aws-sam-cli-app-templates with the proposed changes to manifest.json?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -149,57 +153,36 @@ def _generate_from_app_template(
...
-            click.echo(msg)
-            choice_num = choice_num + 1
-        choice = click.prompt("Base image", type=click.Choice(choices), show_choices=False)
-        image = LAMBDA_IMAGES_RUNTIMES[int(choice) - 1]  # zero index
-
     runtime = image[image.find("/") + 1 : image.find("-")]
-    return image, runtime
-
-
-def _get_dependency_manager(dependency_manager, runtime):
```
sriram-mv, line 185: Was this dead code? just trying to understand its removal.

sriram-mv (changes requested):

samcli/commands/init/init_templates.py
jonife: Also please note that we just started reviewing the tech doc and some iteration might be expected for this pr

tests/unit/commands/init/test_cli.py
jonife: there is a pr already this already link here: https://github.com/aws/aws-sam-cli-app-templates/pull/90

samcli/commands/init/interactive_init_flow.py
jonife: Yes are was not need because 

1. First the method was confusing, inspired good coding practice it should return only the runtime derived from the the image name. 
2. The list of images supported can be generated from the combination of use-cases and package-type selected by the customer. Therefore all this method `_get_runtime_from_image` needs to only get the runtime from the image name.
3. Currently `_get_dependency_manager` depends on `RUNTIME_TO_DEPENDENCY_MANAGERS` in the runtime.py file which is a mapping of runtimes and its dependencies. However in this design, this is not the case as the dependency manager can be easily gotten from the template choosen by the customer. This promotes, readability, ease of implementation and consistency

samcli/commands/init/init_templates.py
jonife: manifest structure? 
Do you mean the manifest from the aws-sam-cli-template repo or the preprocessed manifest that function creates?

sriram-mv: The manifest structure itself as referenced in https://github.com/aws/aws-sam-cli-app-templates/pull/90/files, the thing I have noticed is that we are using manifest structure that's in a different repo, but we dont have a comment explaining what exactly the structure should look like. Adding that will help with the readability of the code without having to jump to  a different repository.

samcli/commands/init/init_templates.py
jonife: I don't understand what you mean by `matcher pattern` but I have extracted the `if statements` into a different method to handle validation of the properties.

samcli/commands/init/interactive_init_flow.py
```diff
@@ -190,25 +283,20 @@ def _get_dependency_manager(dependency_manager, runtime):
...
 
 
-def _get_schema_template_details(schemas_api_caller):
-    try:
-        return get_schema_template_details(schemas_api_caller)
-    except ClientError as e:
-        raise SchemasApiException(
-            "Exception occurs while getting Schemas template parameter. %s" % e.response["Error"]["Message"]
-        ) from e
+def validate_runtime_dependancy_manager(dependency_manager, runtime):
```
mgrandis, line 293: Small typo: `dependancy` -> `dependency`

samcli/commands/init/__init__.py
```diff
@@ -273,9 +273,9 @@ def do_cli(
     image_bool = name and pt_explicit and base_image
     if location or zip_bool or image_bool:
         # need to turn app_template into a location before we generate
-        templates = InitTemplates(no_interactive)
+        templates = InitTemplates()
```
sriram-mv, line 276: Curious about these changes, --no-interactive ensures that we do not launch into a interactive experience at all, was this redundant code or something that has changed?

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
             if option.get("appTemplate") == app_template:
                 return option.get("isDynamicTemplate", False)
         return False
+
+    def get_preprocessed_manifest(self, filter_value=None):
+        """
+        This method get the manifest cloned from the git repo and preprocessed it.
+        The structure of the manifest is shown below:
+
+        {
```
sriram-mv, line 193: Lets create a link to the actual manifest as well, we do not want two percieved sources of truth in the future in the form of a docstring and the actual contents within a different file. This can lead to drift.

samcli/commands/init/init_templates.py
sriram-mv: Think of the validations you are performing to be of thier objects, that way you could do something like

```
for validator in validators:
      validator.validate(template)
```

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+                if package_type not in preprocessed_manifest[use_case][runtime]:
+                    preprocessed_manifest[use_case][runtime][package_type] = []
+                preprocessed_manifest[use_case][runtime][package_type].append(template)
+        return preprocessed_manifest
+
+    def get_bundle_option(self, package_type, runtime, dependency_manager):
+        return self._init_options_from_bundle(package_type, runtime, dependency_manager)
+
+
+def validate_value_in_template(value, template):
```
sriram-mv, line 250: This is in an example of a validator function, collate your functions together like this one so that if conditions can be cleaned up.

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+    if value not in template:
+        raise InvalidInitTemplateError(
+            "Template is missing the value for {property} in manifest file. Please raise a ticket".format(
+                property=value
+            )
+        )
+
+
+def get_runtime(package_type, template_runtime):
+    if package_type == "Image":
```
sriram-mv, line 260: you can use utils for "Image" string. I believe there is a sam cli enum for it.

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+        template_runtime = template_runtime[template_runtime.find("/") + 1 : template_runtime.find("-")]
```
sriram-mv, line 261: I wonder if a regex helps here.

samcli/commands/exceptions.py
```diff
@@ -78,3 +78,9 @@ class AppPipelineTemplateMetadataException(UserException):
     """
     Exception class when SAM is not able to parse the "metadata.json" file located in the SAM pipeline templates
     """
+
+
+class InvalidInitOptionException(UserException):
```
sriram-mv, line 83: Where do we use this exception?

sriram-mv (changes requested):

samcli/commands/init/init_templates.py
mgrandis: I think you may simply be able to do a:
```
use_case = preprocessed_manifest.get(use_case_name, {})
package_types = use_case.get(runtime, {}).get(package_type, [])
package_types.append(template)
preprocessed_manifest[use_case_name] = use_case
```

You may also want to rename `validate_value_in_template` to `get_template_value` and have it return the value directly

mgrandis (changes requested):
Few comments

samcli/commands/init/__init__.py
mgrandis: It looks like the value was passed but was never used

samcli/commands/exceptions.py
mgrandis: Here: https://github.com/aws/aws-sam-cli/pull/2958/files#diff-3ec414e5c1ea9746347f7336f0308441a1031df7076194c96ab9ab541f7660d3R103 (diff was collapsed)

samcli/commands/init/interactive_init_flow.py
```diff
@@ -149,57 +220,94 @@ def _generate_from_app_template(
...
+    templates = []
+    for template in templates_options:
+        if template["dependencyManager"] == dependency_manager:
+            templates.append(template)
+    return templates
+
+
+def _get_choice_from_options(choosen, options, question, msg):
+
+    if not choosen:
```
mgrandis, line 247: You may want to reverse this condition to short circuit the block and reduce indentation:
```
if chosen is not None:
   return chosen
```

samcli/commands/init/interactive_init_flow.py
```diff
@@ -149,57 +220,94 @@ def _generate_from_app_template(
...
+        click_choices = []
+
+        options_list = options if isinstance(options, list) else list(options.keys())
+
+        if len(options_list) == 1:
```
mgrandis, line 252: Same here, you can `click.echo(...)` then directly `return options_list[0]`

samcli/commands/init/interactive_init_flow.py
```diff
@@ -40,29 +45,90 @@ def do_interactive(
...
+            dependency_manager,
+            output_dir,
+            name,
+            app_template,
+            no_input,
+            location_opt_choice,
+        )
+
+
+def get_start_options(
```
mgrandis, line 69: You could rename this to `generate_application`

samcli/commands/init/interactive_init_flow.py
```diff
@@ -40,29 +45,90 @@ def do_interactive(
...
+
+    no_input = True
+    summary_msg = generate_summary_message(
+        package_type, runtime, base_image, dependency_manager, output_dir, name, app_template
+    )
+    click.echo(summary_msg)
+    do_generate(location, package_type, runtime, dependency_manager, output_dir, name, no_input, extra_context)
 
 
 def _generate_from_location(
```
mgrandis, line 134: We could also use the `generate_summary_message` here.
Add a `location=None` parameter at the end and return this message if it's set.

samcli/commands/init/interactive_init_flow.py
```diff
@@ -224,3 +360,34 @@ def _package_schemas_code(runtime, schemas_api_caller, schema_template_details,
         ) from e
     finally:
         remove(download_location.name)
+
+
+def generate_summary_message(package_type, runtime, base_image, dependency_manager, output_dir, name, app_template):
```
mgrandis, line 365: You may be able to group some common parts (like initializing the `summary_msg` with the Generating application header for example.

samcli/commands/init/init_templates.py
mgrandis: Or maybe use split

mgrandis (changes requested):
A few additional comments on the `interactive_init_flow` file

samcli/commands/init/__init__.py
jonife: it was passed but never used

mgrandis:
We need unit tests for `InitTemplates`:
- `get_runtime`
- `validate_value_in_template`
- `get_preprocessed_manifest`

samcli/commands/init/init_templates.py
jonife: ok

samcli/commands/init/interactive_init_flow.py
```diff
@@ -40,29 +45,90 @@ def do_interactive(
...
+    runtime,
+    base_image,
+    dependency_manager,
+    output_dir,
+    name,
+    app_template,
+    no_input,
+    location_opt_choice,
+):
+    if location_opt_choice == "1":
```
sriram-mv, line 82: I'm not a big fan of using if, else for control structure. It promotes more if else, if there are new options introduced.

Could we do something like dictionary dispatch? https://alysivji.github.io/quick-hit-dictionary-dispatch.html

samcli/commands/init/interactive_init_flow.py
```diff
@@ -40,29 +45,90 @@ def do_interactive(
...
-        _generate_from_app_template(
-            location, package_type, runtime, base_image, dependency_manager, output_dir, name, app_template
+
+
+# pylint: disable=too-many-statements
+def _generate_simple_application(
+    location, pt_explicit, runtime, base_image, dependency_manager, output_dir, name, app_template, package_type=ZIP
+):
+    if pt_explicit and package_type == IMAGE:
+        raise InvalidInitOptionException("Hello World Application only supports ZIP package type")
```
sriram-mv, line 103: should we add something about, what to do for IMAGE as well?

sriram-mv (changes requested):

samcli/commands/init/interactive_init_flow.py
mgrandis: I think we're good here, not a big fan of this type of dictionary dispatch, we also have few conditions and I don't think it will grow to become unreadable/difficult to follow.

sriram-mv: After a discussion with @mgrandis, I'm okay with this as long as its a few choices. Can we document that we would need a refactor if the number of options increase, I'm concerned about possible bugs being introduced in the control flow with more options and using if else for managing them.

samcli/commands/init/interactive_init_flow.py
jonife: we current do not support Images for the hello application. Customers can use image in Quick start template. During the last UX review this was agreed on this but if we need to pass a message about images we might need to reachout to Molly again.

mgrandis: I think what Sriram meant was that we may need a message to point customers in the right direction to use an image instead of just telling them "Not supported". We will see this with UX.

samcli/commands/init/init_templates.py
```diff
@@ -105,20 +110,28 @@ def _check_app_template(entry: Dict, app_template: str) -> bool:
...
+            self._git_repo.clone(clone_dir=shared_dir, clone_name=APP_TEMPLATES_REPO_NAME, replace_existing=True)
+        except CloneRepoUnstableStateException as ex:
+            raise AppTemplateUpdateException(str(ex)) from ex
+        except (OSError, CloneRepoException):
+            # If can't clone, try using an old clone from a previous run if already exist
+            expected_previous_clone_local_path: Path = shared_dir.joinpath(APP_TEMPLATES_REPO_NAME)
+            if expected_previous_clone_local_path.exists():
+                self._git_repo.local_path = expected_previous_clone_local_path
+
+    def _get_local_path(self):
```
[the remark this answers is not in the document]
jonife: yes it is redundant since we a resulting to cloning the repo

samcli/commands/init/init_templates.py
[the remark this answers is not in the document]
jonife: do you mean `get_bundle_option`? if so it can't be private as it call outside this class

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+
+        Raises:
+            InvalidInitTemplateError: Exception raise when template is invalid or missing a parameter
+
+        Returns:
+            dict: This is preprocessed manifest with the use_case as key
+        """
+        if not self._git_repo.clone_attempted:
+            self.git_cloner()
+        manifest_path = os.path.join(self._git_repo.local_path, "manifest.json")
```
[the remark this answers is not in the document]
jonife: The decision to use `path` was made because this would fail pr test as module `os` has no `Path` member. if you inspect the code base you would notice that `path` was used in place of `Path`

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +184,79 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+        with open(str(manifest_path)) as fp:
+            body = fp.read()
+            manifest_body = json.loads(body)
+
+        preprocessed_manifest = {}
+
+        for template_runtime in manifest_body:
+            if filter_value and filter_value == template_runtime:
```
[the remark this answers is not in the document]
jonife: sorry i should have cleaned it up, I wrote this during my brainstorming session on how to filter the manifest

samcli/commands/init/interactive_init_flow.py
jonife: `location` would always be set before calling `do_generate` and location is part of the message return.
The only way to add this `generate_summary_message` is add a flag, which i truly think shouldn't be used and all the number of argument for the `generate_summary_message` is already too much.

samcli/commands/init/interactive_init_flow.py
jonife: The common part are alternating with the uncommon part and since strings are immutable, I don't think this will be possible.

samcli/commands/init/interactive_init_flow.py
jonife: Do you mean information telling customer about Images?

samcli/commands/init/init_templates.py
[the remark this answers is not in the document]
mgrandis: `Path` is not part of `os` and is already imported in this file: `from pathlib import Path`

samcli/commands/init/init_templates.py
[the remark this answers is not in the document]
mgrandis: `validate_value_in_template` and `get_runtime`

samcli/commands/init/interactive_init_flow.py
mgrandis: Strings being immutable only means that under the hood Python creates a new String when you concatenate two occurrences, which is generally a performance issue only when doing it millions of times.

samcli/commands/init/interactive_init_flow.py
jonife: I am not a big fan of if/else statements either, i believe it can easily explode to make more conditions, but thought it fitting for the case since they were fewer condition and it helped with readability in this case.

samcli/commands/init/init_templates.py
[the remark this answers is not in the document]
jonife: this would fail pr as the keyword `self` was not used in the function.

mgrandis (changes requested):
A few more comments

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -4,10 +4,10 @@
 import logging
 import functools
 
+from boto3.session import Session
```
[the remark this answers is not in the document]
jonife: This is change was a recommendation by pylint after make pr failed, it wasn't a new addition but moved from line 10 to line 7.

samcli/commands/init/init_templates.py
```diff
@@ -2,23 +2,30 @@
...
-from samcli.local.common.runtime_template import RUNTIME_DEP_TEMPLATE_MAPPING, get_local_lambda_images_location
+from samcli.local.common.runtime_template import (
+    RUNTIME_DEP_TEMPLATE_MAPPING,
+    get_local_lambda_images_location,
+)
 
 LOG = logging.getLogger(__name__)
-APP_TEMPLATES_REPO_URL = "https://github.com/aws/aws-sam-cli-app-templates"
+# Note:: This would be changed when after aws-sam-cli-app-templates pr
+# https://github.com/aws/aws-sam-cli-app-templates/pull/127 is merged
```
sriram-mv, line 27: Is there anything preventing merge of https://github.com/aws/aws-sam-cli-app-templates/pull/127

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +180,75 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+        filter_value : string, optional
+            This could be a runtime or a base-image, by default None
+
+        Returns
+        -------
+        [dict]
+            This is preprocessed manifest with the use_case as key
+        """
+        self.clone_templates_repo()
+        manifest_path = Path(self._git_repo.local_path, "manifest.json")
```
sriram-mv, line 215: we should have any string constants such as "manifest.json" as class level constants.

samcli/commands/init/interactive_init_flow.py
sriram-mv: @jonife Yes, as @mgrandis pointed out, that is where I was going at.

sriram-mv (commented):
Getting much closer.

samcli/commands/init/init_templates.py
mgrandis: @jonife Have you tested that the new manifest also works with the current/develop version of the SAM CLI?

jonife: Yes I have test and the changes to the manifest file can't affect the how current/develop version of the SAM CLI works as it does not depend on the manifest for it interactive flow but only to spool the application after cloning.

samcli/commands/init/interactive_init_flow.py
jonife: Molly is currently away, she should be back on monday.

Issue #2360, opened 2020-11-07
SELinux prevents volume mount on Fedora 32


mbacchi:
### Description:

Rather than reopening #676, I'm opening a new issue for the most recent version of aws-sam-cli and docker. Issue #676 was prematurely closed in my opinion with the explanation that the user wasn't using the "current version of sam cli".

On Fedora 32, `sam local invoke` fails to mount a volume due to the fact that selinux prevents access to the host directory from within the docker container unless the `:z` or `:Z` mount option is added to add the [selinux label to the directory](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label).

If I run the following `sam local invoke` command using SAM to locally execute the Lambda function in my test repository [here](https://gitlab.com/mbacchi/lambda-env-vars) which uses a volume mount I get the error `[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'app'`:

```
sam local invoke LambdaEnvVarsFunction --event events/event.json
Invoking app.lambda_handler (python3.7)
Skip pulling image and use local one: amazon/aws-sam-cli-emulation-image-python3.7:rapid-1.7.0.

Mounting /home/user/lambda-env-vars/.aws-sam/build/LambdaEnvVarsFunction as /var/task:ro,delegated inside runtime container
START RequestId: afdda989-cb02-1c82-2916-f3e9927bce7d Version: $LATEST
[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'app'
END RequestId: afdda989-cb02-1c82-2916-f3e9927bce7d
REPORT RequestId: afdda989-cb02-1c82-2916-f3e9927bce7d	Init Duration: 78.38 ms	Duration: 2.04 ms	Billed Duration: 100 ms	Memory Size: 128 MB	Max Memory Used: 23 MB	

{"errorType":"Runtime.ImportModuleError","errorMessage":"Unable to import module 'app': No module named 'app'"}
```

As discussed in #676, if I try to run Docker without the `:z` mount option I cannot read the contents of the directory:

```
docker run --rm -it -v /home/user/lambda-env-vars/.aws-sam/build/LambdaEnvVarsFunction:/var/task amazon/aws-sam-cli-emulation-image-python3.7 /bin/bash
bash-4.2# ls
ls: cannot open directory .: Permission denied
bash-4.2# exit
```

[...]

PR #2970, opened 2021-06-18
Use :Z label to mount host dir for SELinux

mbacchi:
SELinux by default prevents mounting volumes from the host OS in
a docker container. To allow it to mount the host volume use the
:Z mount option.

#2360 

To allow Linux users with SELinux enabled to mount host volumes in SAM local invoke docker containers.

If SELinux is not enabled, nothing is changed for the end user. If SELinux is enabled we add the `:Z` mount option to allow the host volume mount to proceed with the proper context.

N/A

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -2,9 +2,6 @@
 Representation of a generic Docker container
 """
 import logging
-import os
-import subprocess
-import sys
 import tarfile
 import tempfile
 import threading
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -16,7 +16,7 @@ from docker.errors import NotFound as DockerNetworkNotFound
 from samcli.lib.utils.retry import retry
 from .exceptions import ContainerNotStartableException
 
-from .utils import to_posix_path, find_free_port, NoFreePortsError
+from .utils import to_posix_path, find_free_port, NoFreePortsError, is_selinux_enabled
 
 LOG = logging.getLogger(__name__)
 
@@ -132,17 +132,12 @@ class Container:
 
         if self._host_dir:
 
-            # When SELinux is enabled add the z label to mount the host volume
+            # When SELinux is enabled add the Z label to mount the host volume
             # inside the container.
-            if (
-                sys.platform == "linux"
-                and os.path.isfile("/usr/sbin/sestatus")
-                and "enabled"
-                in subprocess.Popen("/usr/sbin/sestatus", stdout=subprocess.PIPE).stdout.readline().decode("UTF-8")
-            ):
+            if is_selinux_enabled():
 
-                mount_mode = "z," + mount_mode
-                mount_options_msg = "Mounting %s as %s:z,ro,delegated inside runtime container"
+                mount_mode = "Z," + mount_mode
+                mount_options_msg = "Mounting %s as %s:Z,ro,delegated inside runtime container"
 
             LOG.info(mount_options_msg, self._host_dir, self._working_dir)
 
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -13,7 +13,7 @@ from docker.errors import NotFound as DockerNetworkNotFound
 from samcli.lib.utils.retry import retry
 from .exceptions import ContainerNotStartableException
 
-from .utils import to_posix_path, find_free_port, NoFreePortsError, is_selinux_enabled
+from .utils import to_posix_path, find_free_port, NoFreePortsError
 
 LOG = logging.getLogger(__name__)
 
@@ -129,12 +129,10 @@ class Container:
 
         if self._host_dir:
 
-            # When SELinux is enabled add the Z label to mount the host volume
-            # inside the container.
-            if is_selinux_enabled():
-
-                mount_mode = "Z," + mount_mode
-                mount_options_msg = "Mounting %s as %s:Z,ro,delegated inside runtime container"
+            # Add the SELinux Z label to mount the host volume inside the
+            # container, if SELinux is not present Docker will ignore the label.
+            mount_mode = "Z," + mount_mode
+            mount_options_msg = "Mounting %s as %s:Z,ro,delegated inside runtime container"
 
             LOG.info(mount_options_msg, self._host_dir, self._working_dir)
 
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -2,6 +2,9 @@
 Representation of a generic Docker container
 """
 import logging
+import os
+import subprocess
+import sys
 import tarfile
 import tempfile
 import threading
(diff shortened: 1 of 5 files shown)


samcli/local/docker/container.py
```diff
@@ -123,16 +126,33 @@ def create(self):
 
         _volumes = {}
 
+        # Default volume mount mode and message.
```
hawflau, line 129: minor: could be a better idea to refactor the whole `mount_mode` part to a `_get_mount_mode` method

hawflau (changes requested):
Thanks for the PR! 
Overall looks good! Just added a tiny comment about refactoring. Also, please add some unit tests for the changes.

samcli/local/docker/utils.py
```diff
@@ -154,3 +156,14 @@ def get_docker_platform(architecture: str) -> str:
...
+
+def is_selinux_enabled():
+    """Return True if SELinux is enabled, False otherwise."""
+    if sys.platform == "linux":
+        if os.path.isfile("/usr/sbin/sestatus") and "enabled" in subprocess.Popen(
+            "/usr/sbin/sestatus", stdout=subprocess.PIPE
+        ).stdout.readline().decode("UTF-8"):
+            return True
+        else:
+            return False
```
CoshUS, line 169: nit: Return the condition directly instead of returning True and False.
```suggestion
        return os.path.isfile("/usr/sbin/sestatus") and "enabled" in subprocess.Popen(
            "/usr/sbin/sestatus", stdout=subprocess.PIPE
        ).stdout.readline().decode("UTF-8")
```

CoshUS (changes requested):
Thank you for the PR! 
Just one small change and I'll approve after.

samcli/local/docker/utils.py
mbacchi: Thanks, I didn't run the pylint checks locally before the commit earlier. That is fixed with this most recent commit.

samcli/local/docker/container.py
mbacchi: Do you have a good rubric for understanding when these would be "Z,ro,delegated" and when delegated would be absent? I don't know how to write a `_get_mount_mode` function unless I know how and why those mount options are applied in each situation used here.

CoshUS (approved):

dmarinuswoodwing:
Please have a look at #2984 

`:Z` can always be supplied as it's ignored on systems not using selinux

hawflau (changes requested):
A couple [test cases](https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/42321810) failed. Could you please check them?

On the other hand, I'm a bit concerned when I read through the [docs on docker](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label):
> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

I don't have a good solution just yet, but I think we should have some guards to prevent or warn there could be unwanted consequence.

mbacchi:
> A couple [test cases](https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/42321810) failed. Could you please check them?
> 
> On the other hand, I'm a bit concerned when I read through the [docs on docker](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label):
> 
> > Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.
> 
> I don't have a good solution just yet, but I think we should have some guards to prevent or warn there could be unwanted consequence.

I noticed the Appveyor tests failing yesterday, and as someone else suggested removed the SELinux `sestatus` check, as it wasn't installed on those containers. I added a new commit which adds the Z label in all cases which will be ignored by Docker if the OS doesn't support it.

Regarding the bind mount on /home or /usr, I believe that means if the command is run in one of those directories, not a subdir of /home (which is where I run SAM.) I don't know whether that's a likely problem or not, but interested to hear your thoughts.

hawflau:
Hi @mbacchi, sorry for the delayed response. I think it makes sense to remove the SELinux check as applying `:Z` has no effect on non-SELinux systems. 

However, I'm still concerned with the risk of rendering the host inoperable due to bind-mounting a system directory such as `/home` or `/usr`. Currently there's no way preventing users from setting a property (e.g. `CodeUri`) to point to a system directory. Even though it's unlikely someone would do something like this, it sounds scary enough. So I think a better approach is to allow users to opt-in using `:Z` label through a flag. For example:
```
sam local invoke --selinux-host
sam build --use-container --selinux-host
```

To add the flag, you can look at [here](https://github.com/aws/aws-sam-cli/blob/ff2baf81245d13788d74c5b183b05a0ef152241e/samcli/commands/local/cli_common/options.py#L135) for `local` commands and [here](https://github.com/aws/aws-sam-cli/blob/ff2baf81245d13788d74c5b183b05a0ef152241e/samcli/commands/build/command.py#L80-L129) for `build` command. Thanks!

mbacchi:
Thanks @hawflau for getting back to me. I don't love the idea of a CLI flag, and I'm a little surprised at the concern over bind mounting in bare system directories (i.e. `/home` or `/usr`). This seems slightly paranoid but I guess that's a safe position too.

I wonder if instead of adding a flag as you suggest, we could globally refuse to perform any operations if the user is initiating the `sam local invoke` command from one of those directories? I have a suspicion there would be problems if you ran `sam local invoke` in the `/usr/bin` directory as well. Maybe the sam-cli has dodged a bullet by not being run in those directories all along, and this SELinux realization gives us a chance to rectify this by identifying a set of directories the sam-cli will refuse to run in, in order to protect users.

In the end, SELinux is a security tool, and other products/tools should make it easy to use their product with security tools, as opposed to magnifying the difficulty of using them in tandem. That's why I'm lobbying for another way than a command line flag to support SELinux. In effect, if you add a flag to support SELinux, users won't know they need to provide the flag unless they read pretty deep into the documentation. It will effectively break these users `sam local invoke` commands (as it is currently broken.)

The old adage goes: if you add a feature (or flag) you can't remove it later, as someone will be reliant on it. (I'm obviously paraphrasing here lol.) Maybe we could brainstorm workarounds first, before jumping to implement this flag. I'm willing to help however I can, and will reach out to other SAM users to help crowd source ideas.

How does that sound to you?

hawflau:
Thanks for your response @mbacchi 

> we could globally refuse to perform any operations if the user is initiating the `sam local invoke` command from one of those directories? I have a suspicion there would be problems if you ran `sam local invoke` in the /usr/bin directory as well

Yes, that works as well. From this [comment](https://github.com/moby/moby/issues/30934#issuecomment-403004060), it seems like we can block some common system paths like `/var`, `/usr` and `/`. But it's likely that we will still miss other errors.

And you are right, adding a flag sounds like a irreversible change. Applying `:z` or `:Z` label modifies user's host (as [here](https://projectatomic.io/blog/2016/03/dwalsh_selinux_containers/), it's basically equivalent to running `chcon -R -t svirt_sandbox_file_t /SOURCEDIR`). So I think it's important for a user to be aware what change that SAM CLI is doing to the user's host. Using a flag to opt in is one way to confirm the awareness.

From the links I quoted above, It seems like an alternative is to apply `--security-opt label=disabled` in the underlying docker call. Maybe we can try this too.

jfuss:
@mbacchi Are you still interested in contributing this? If so, can you handle the feedback? If not, I totally understand but would then like to close the PR.

mbacchi:
Yes, I'll try to look again soon, this is still an issue but I wanted to try out the suggestion to use `--security-opt label=disabled` when executing the docker call.

jfuss:
@mbacchi No problem at all. I have been trying to get through our PR backlog in our repos, so that was the reason for asking. Feel free to tag me when you get back to it.

The one thing we will need to make sure is compatibility with other "docker" like tools (like Rancher). But I think first verifying with docker, is best. 

I would love to get this through, I know it's a long standing issue.

jfuss:
@mbacchi No rush on getting back here but I am going to move this to draft. When you get back to things, please update the PR out of draft and we will take a look again.

jfuss:
I am going to close this. @mbacchi or anyone from the community: if you get back to this, feel free to reopen or create a new one and we will get a review out for you. It would be useful to link back to this one so we don't go back through the same discussions (assuming a new one is created).

PR #2981, opened 2021-06-24
feat: Delete methods for CF stacks and S3 files

hnnasit:
Methods and classes in this PR are necessary to delete the resources SAM deploy created

Comments:
Testing for command.py and delete_context.py will be done in a separate PR as I will be updating these files for the next tasks.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/delete/cf_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+            if "Stack with id {0} does not exist".format(stack_name) in str(e):
+                LOG.debug("Stack with id %s does not exist", stack_name)
+                return False
+            raise DeleteFailedError(stack_name=stack_name, msg=str(e)) from e
+        except BotoCoreError as e:
+            # If there are credentials, environment errors,
+            # catch that and throw a delete failed error.
+
+            LOG.debug("Botocore Exception : %s", str(e))
```
moelasmar, line 45: i think it is better to use log.error instead of debug in case of raising exceptions

samcli/lib/delete/utils.py
```diff
@@ -0,0 +1,17 @@
+"""
+Utilities for Delete
+"""
+
+from samcli.lib.utils.hash import file_checksum
+from samcli.lib.package.artifact_exporter import mktempfile
+
+
+def get_cf_template_name(template_str, extension):
```
moelasmar, line 9: do you think we do not have a method to generate the template name .. I believe we should maintain this logic in one place.

samcli/lib/package/s3_uploader.py
```diff
@@ -144,6 +145,43 @@ def upload_with_dedup(
 
         return self.upload(file_name, remote_path)
 
+    def delete_artifact(self, remote_path: str, is_key=False):
```
moelasmar, line 148: add some debug && error log messages

samcli/lib/package/s3_uploader.py
```diff
@@ -144,6 +145,43 @@ def upload_with_dedup(
...
+        """
+        Deletes a given file from S3
+        :param remote_path: Path to the file that will be deleted
```
moelasmar, line 151: question: the remote path should be determined by the children classes of the S3Uploader .. am I correct?

moelasmar (changes requested):

samcli/lib/package/s3_uploader.py
hnnasit: Yes the remote_path is basically the key when we parse S3 artifacts from the template file and if only the file_name is provided then we create the key using s3_prefix + remote_path.

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,131 @@
...
+                    self.s3_uploader.delete_artifact(self.cf_template_file_name)
+
+                # Delete the folder of artifacts if s3_bucket and s3_prefix provided
+                elif self.delete_artifacts_folder:
+                    self.s3_uploader.delete_prefix_artifacts()
+
+                # Delete the ECR companion stack
+
+                if self.cf_template_file_name:
+                    click.echo(f"- deleting template file {self.cf_template_file_name}")
```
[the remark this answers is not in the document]
hnnasit: So I have added a waiter in another branch when delete_stack is called and while delete is in progress, the user will see the text deleting it.

samcli/lib/delete/utils.py
hnnasit: I agree, I will modify the code so that the logic to get the template file name when the template file is deployed and deleted remains the same.

samcli/commands/delete/command.py
```diff
@@ -0,0 +1,71 @@
...
+LOG = logging.getLogger(__name__)
+
+
+@click.command(
+    "delete",
+    short_help=SHORT_HELP,
+    context_settings={"ignore_unknown_options": False, "allow_interspersed_args": True, "allow_extra_args": True},
+    help=HELP_TEXT,
+)
+@configuration_option(provider=TomlProvider(section=CONFIG_SECTION, cmd_names=[CONFIG_COMMAND]))
```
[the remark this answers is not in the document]
hnnasit: That makes sense, I will create a separate method to read the toml file if it exists and assign the parameters accordingly.

samcli/commands/delete/command.py
```diff
@@ -0,0 +1,88 @@
...
+
+import click
+from samcli.cli.main import aws_creds_options, common_options, pass_context, print_cmdline_args
+
+from samcli.lib.utils.version_checker import check_newer_version
+
+SHORT_HELP = "Delete an AWS SAM application and the artifacts created by sam deploy."
+
+HELP_TEXT = """The sam delete command deletes the Cloudformation
+Stack and all the artifacts which were created using sam deploy.
```
hoffa, line 15:
```suggestion
HELP_TEXT = """The sam delete command deletes the CloudFormation
stack and all the artifacts which were created using sam deploy.
```

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+        self.stack_name = stack_name
+        self.region = region
+        self.profile = profile
+        self.config_file = config_file
+        self.config_env = config_env
+        self.s3_bucket = None  # s3_bucket
+        self.s3_prefix = None  # s3_prefix
+        self.cf_utils = None
+        self.start_bold = "\033[1m"
+        self.end_bold = "\033[0m"
```
hoffa, line 40: ANSI escape characters aren't compatible on Windows AFAIK; we should be using higher-level libraries. I think `click` supports this.

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+
+class DeleteContext:
+    def __init__(self, stack_name, region, profile, config_file, config_env):
+        self.stack_name = stack_name
+        self.region = region
+        self.profile = profile
+        self.config_file = config_file
+        self.config_env = config_env
+        self.s3_bucket = None  # s3_bucket
+        self.s3_prefix = None  # s3_prefix
```
hoffa, line 37: nit: value of comment?

samcli/commands/delete/command.py
```diff
@@ -0,0 +1,88 @@
...
+):
+    """
+    `sam delete` command entry point
+    """
+
+    # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
+    do_cli(stack_name, ctx.region, config_file, config_env, ctx.profile)  # pragma: no cover
+
+
+def do_cli(stack_name, region, config_file, config_env, profile):
```
hoffa, line 79: Please add type hints to the PR.

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+                self.s3_bucket = config_options.get("s3_bucket", None)
+                self.s3_prefix = config_options.get("s3_prefix", None)
+
+    def delete(self):
+        """
+        Delete method calls for Cloudformation stacks and S3 and ECR artifacts
+        """
+        template_str = self.cf_utils.get_stack_template(self.stack_name, TEMPLATE_STAGE)
+
+        # template_dict = yaml_parse(template_str)
```
hoffa, line 88: Remove this if unused.

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+                )
+
+        click.echo("\n")
+        # Delete the primary stack
+        self.cf_utils.delete_stack(self.stack_name)
+
+        click.echo("- deleting Cloudformation stack {0}".format(self.stack_name))
+
+        # Delete the artifacts
+        # Intentionally commented
```
hoffa, line 112: Why?

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Delete the CF template file in S3
+        if self.delete_cf_template_file:
+            self.s3_uploader.delete_artifact(self.cf_template_file_name)
+
+        # Delete the folder of artifacts if s3_bucket and s3_prefix provided
+        elif self.delete_artifacts_folder:
+            self.s3_uploader.delete_prefix_artifacts()
+
+        # Delete the ECR companion stack
```
hoffa, line 125: Missing code?

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,163 @@
...
+                    f"\t{self.start_bold}Do you want to delete the template file"
+                    + f" {self.cf_template_file_name} in S3?{self.end_bold}",
+                    default=False,
+                )
+
+        click.echo("\n")
+        # Delete the primary stack
+        self.cf_utils.delete_stack(self.stack_name)
+
+        click.echo("- deleting Cloudformation stack {0}".format(self.stack_name))
```
hoffa, line 109: nit: the convention seems to be to use f-strings throughout

samcli/lib/delete/cf_utils.py
```diff
@@ -0,0 +1,102 @@
...
+        :param stack_name: Name or ID of the stack
+        :return: True if stack exists. False otherwise
+        """
+        try:
+            resp = self._client.describe_stacks(StackName=stack_name)
+            if not resp["Stacks"]:
+                return False
+
+            stack = resp["Stacks"][0]
+            return stack["StackStatus"] != "REVIEW_IN_PROGRESS"
```
hoffa, line 30: Why is this a special case?

samcli/lib/delete/cf_utils.py
```diff
@@ -0,0 +1,102 @@
...
+                return None
+
+            return resp["TemplateBody"]
+
+        except (ClientError, BotoCoreError) as e:
+            # If there are credentials, environment errors,
+            # catch that and throw a delete failed error.
+
+            LOG.error("Failed to delete stack : %s", str(e))
+            raise DeleteFailedError(stack_name=stack_name, msg=str(e)) from e
```
hoffa, line 74: `DeleteFailedError` from a read-only method is confusing.

hoffa (changes requested):

samcli/commands/delete/delete_context.py
hnnasit: Good catch! I forgot to remove it.

samcli/commands/delete/delete_context.py
hnnasit: I had kept it as the starting code for the next task which is now on a different branch, but I can remove the commented code in this branch to keep it clean.

samcli/lib/delete/cf_utils.py
hnnasit: My understanding was that when a stack is in review on CF, the change_set is not executed yet and so if a user wants to delete this stack, we will have to handle this separately(deleting change_set and not stack). Just to avoid that for now, I check the status. Correct me if my understanding is wrong.

hoffa: Not sure. I'd assume any stack that exists regardless of status should be deletable. In any case, whenever the _why_ isn't clear in the code, please add a comment for explanation.

samcli/lib/delete/cf_utils.py
hnnasit: I agree, does raising a ReadFailedError or FetchFailedError instead make sense?

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,141 @@
...
+                        "\tDo you want to delete the template file" + f" {self.cf_template_file_name} in S3?", bold=True
+                    ),
+                    default=False,
+                )
+
+        click.echo("\n")
+        # Delete the primary stack
+        self.cf_utils.delete_stack(stack_name=self.stack_name)
+
+        click.echo(f"- deleting Cloudformation stack {self.stack_name}")
```
hoffa, line 98: Is this consistent with the UI conventions in SAM CLI? (colors, indentation, capitalization) I know it's a bit of a mess currently, but until we have SAM CLI-wide UI guidelines, we should strive for consistency so the experience feels coherent.

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,137 @@
...
+            config_path=self.config_file, config_env=self.config_env, cmd_names=[CONFIG_COMMAND]
+        )
+        if config_options:
+            if not self.stack_name:
+                self.stack_name = config_options.get("stack_name", None)
+            if self.stack_name == config_options["stack_name"]:
+                if not self.region:
+                    self.region = config_options.get("region", None)
+                if not self.profile:
+                    self.profile = config_options.get("profile", None)
```
moelasmar, line 63: for region && profile also update SAM context, to refresh the default boto session

moelasmar (approved):

samcli/lib/package/utils.py
```diff
@@ -284,3 +284,13 @@ def copy_to_temp_dir(filepath):
     dst = os.path.join(tmp_dir, os.path.basename(filepath))
     shutil.copyfile(filepath, dst)
     return tmp_dir
+
+
+def get_cf_template_name(temp_file, template_str, extension):
```
hoffa, line 289: Should probably be something like (using `typing.TextIO` for opened text-based file descriptors):

```suggestion
def get_cf_template_name(temp_file: TextIO, template_str: str, extension: str) -> str:
```

samcli/commands/delete/delete_context.py
```diff
@@ -0,0 +1,141 @@
...
+        Read the provided config file if it exists and assign the options values.
+        """
+        toml_provider = TomlProvider(CONFIG_SECTION, [CONFIG_COMMAND])
+        config_options = toml_provider(
+            config_path=self.config_file, config_env=self.config_env, cmd_names=[CONFIG_COMMAND]
+        )
+        if config_options:
+            if not self.stack_name:
+                self.stack_name = config_options.get("stack_name", None)
+            if self.stack_name == config_options["stack_name"]:
```
hoffa, line 59: Is `stack_name` optional or not? L58 suggests it is, L59 asserts it isn't.

hnnasit: Good catch, stack_name is required. I have a user prompt on line 39 which should be placed before line 59.

hnnasit: But this is already checked on line 39 and if the customer does not have a stack_name present in samconfig.toml, line 58 will assign stack_name to None, we do not use any other information from samconfig.toml. So this case should be handled on line 39

hoffa (approved):

PR #2984, opened 2021-06-25
Add :Z to fix selinux labels

meeuw:
Might be duplicate of #2970 

Issue #2360 

When docker is used in an SELinux environment the container can't access files from a bind mounted volume without the proper labels. 

By adding :Z to the volume options docker automatically sets the right labels.

None that I know of, this is compatible with docker machine on MacOS.

#### Checklist

- ~Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods~
- ~Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))~
- ~Write unit tests~
- [x] `make pr` passes
- ~`make update-reproducible-reqs` if dependencies were changed~
- ~Write documentation~

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -124,7 +124,7 @@ class Container:
         _volumes = {}
 
         if self._host_dir:
-            LOG.info("Mounting %s as %s:ro,delegated inside runtime container", self._host_dir, self._working_dir)
+            LOG.info("Mounting %s as %s:ro,delegated,Z inside runtime container", self._host_dir, self._working_dir)
 
             _volumes = {
                 self._host_dir: {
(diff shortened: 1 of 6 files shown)


meeuw:
As I think of it, wouldn't it be a better idea to copy-in the required files instead of using bind mounts?

jfuss:
I am going to close this as a duplicate of https://github.com/aws/aws-sam-cli/pull/2970. We are reviewing and working with that author. @meeuw feel free to ping on that one or contribute to it's discussion.

Thank you for submitting the PR and spending you time out it.

PR #2997, opened 2021-06-29
Add region to pipeline bootstrap interactive flow

aahung:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/pipeline/bootstrap/guided_context.py
+++ b/samcli/commands/pipeline/bootstrap/guided_context.py
@@ -18,6 +18,7 @@ class GuidedContext:
         create_image_repository: bool = False,
         image_repository_arn: Optional[str] = None,
         pipeline_ip_range: Optional[str] = None,
+        region: Optional[str] = None,
     ) -> None:
         self.environment_name = environment_name
         self.pipeline_user_arn = pipeline_user_arn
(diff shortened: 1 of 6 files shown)


samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -40,6 +42,13 @@ def run(self) -> None:
                 type=click.STRING,
             )
 
+        if not self.region:
+            self.region = click.prompt(
+                "\nAWS region (the AWS region where the environment infrastructure resources will be deployed to)",
+                type=click.STRING,
+                default="us-east-1",
```
c2tarun, line 49: should we default to `us-east-1` or boto.session.region?

c2tarun (changes requested):

hoffa (approved):
LGTM. Not sure about the newline but we can revisit.

samcli/commands/pipeline/bootstrap/guided_context.py
aahung: updated, good catch

aahung:
> LGTM. Not sure about the newline but we can revisit.

It was to be consistent with others, we should revisit

We don't need to adjust our templates because `region` is already configured to read from the toml file

image

c2tarun (approved):

PR #3022, opened 2021-07-07
Delete template artifacts

hnnasit:
This PR contains methods for deleting artifacts for resources in the Cloudformation template
### Comments
Integration tests and the missing unit tests will be pushed in a separate PR.

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/ecr_uploader.py
```diff
@@ -83,6 +90,40 @@ def upload(self, image, resource_name):
...
+        image_uri
+
+        :param image_uri: image_uri of the image to be deleted
+        :param resource_id: id of the resource for which the image is deleted
+        :param property_name: provided property_name for the resource
+        """
+        try:
+            repo_image_tag = image_uri.split("/")[1].split(":")
+            repository = repo_image_tag[0]
+            image_tag = repo_image_tag[1]
```
moelasmar, line 105: is it possible that the image tag is in Format `repository:tag` and not in format `registry/repository:tag`. I think also the tag is optional, which means that the tag value is `latest`.

I think you need also to guard the code against wrong formats, I know that this code depends on the template you downloaded from CFN which means that it will be always a correct format, but think from code reusability point of view, so this method should be self contained, and does not depend on the caller.

samcli/lib/package/ecr_uploader.py
```diff
@@ -83,6 +90,40 @@ def upload(self, image, resource_name):
...
+            resp = self.ecr_client.batch_delete_image(
+                repositoryName=repository,
+                imageIds=[
+                    {"imageTag": image_tag},
+                ],
+            )
+            if resp["failures"]:
+                # Image not found
+                image_details = resp["failures"][0]
+                if image_details["failureCode"] == "ImageNotFound":
```
moelasmar, line 115: what should we do in else .. it seems that the deleting failed, but we did not mention that to the customer

samcli/lib/package/ecr_uploader.py
```diff
@@ -83,6 +90,40 @@ def upload(self, image, resource_name):
...
+                    LOG.error("ImageNotFound Exception : ")
+                    raise ImageNotFoundError(resource_id, property_name)
```
moelasmar, line 117: This exception will show this  message `Image not found for {property_name} parameter of {resource_id} resource. \n` to the customer, I prefer something like `Could delete Image {} used for {} resource because it is not exist`

tests/integration/local/start_api/test_start_api.py
```diff
@@ -382,14 +383,14 @@ def test_valid_v2_lambda_integer_response(self):
 
     @pytest.mark.flaky(reruns=3)
     @pytest.mark.timeout(timeout=600, method="thread")
-    def test_invalid_v2_lambda_response(self):
+    def test_v2_lambda_response_skip_unexpected_fields(self):
```
moelasmar, line 386: I believe this change is not your changes, you got them when you merged from develop branch.

I prefer to split these commits in a separate PR, so we can merge them directly. It will be easy for us to review your changes only.

samcli/commands/delete/command.py
```diff
@@ -57,34 +57,46 @@
     default="default",
     show_default=True,
 )
+@click.option(
+    "--force",
+    help=("Specify this flag to allow SAM CLI to skip through the guided prompts" ""),
+    is_flag=True,
+    type=click.BOOL,
+    required=False,
```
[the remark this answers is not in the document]
hnnasit: So the default value will be False since it's a flag.

samcli/commands/delete/delete_context.py
```diff
@@ -1,9 +1,11 @@
 """
 Delete a SAM stack
 """
-
+import logging
 import boto3
 
+
+import docker
```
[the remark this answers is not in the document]
hnnasit: Yes, docker is not required here.

samcli/lib/package/artifact_exporter.py
```diff
@@ -237,3 +238,31 @@ def export(self) -> Dict:
...
+
+        for resource_id, resource in self.template_dict["Resources"].items():
+
+            resource_type = resource.get("Type", None)
+            resource_dict = resource.get("Properties", {})
+            resource_deletion_policy = resource.get("DeletionPolicy", None)
+            # If the deletion policy is set to Retain,
+            # do not delete the artifact for the resource.
+            if resource_deletion_policy != "Retain":
+                for exporter_class in self.resources_to_export:
```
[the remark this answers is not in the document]
hnnasit: Good point, we can use a dict to make get calls faster.

samcli/lib/package/packageable_resources.py
```diff
@@ -197,6 +220,21 @@ def do_export(self, resource_id, resource_dict, parent_dir):
...
         set_value_from_jmespath(resource_dict, self.PROPERTY_NAME, {self.EXPORT_PROPERTY_CODE_KEY: uploaded_url})
 
+    def delete(self, resource_id, resource_dict):
+        """
+        Delete the ECR artifact using ECR url in PROPERTY_NAME referenced by EXPORT_PROPERTY_CODE_KEY
+        """
+        if resource_dict is None:
+            return
+
+        remote_path = resource_dict[self.PROPERTY_NAME][self.EXPORT_PROPERTY_CODE_KEY]
```
hawflau, line 230: This will raise a `KeyError` if `self.PROPERTY_NAME` is not in `resource_dict`, or if `self.EXPORT_PROPERTY_CODE_KEY]` is not in `resource_dict[self.PROPERTY_NAME]`. Are we sure both dictionaries contain the specified keys?

hnnasit: Thats a good point. I was thinking of replacing that with jmespath.search() which would return None if no value is found at 'self.PROPERTY_NAME.self.EXPORT_PROPERTY_CODE_KEY'

moelasmar: I think .get is enough, no need to use jmespath.search() .. You can make it 
`remote_path = resource_dict.get(self.PROPERTY_NAME, {}).get(self.EXPORT_PROPERTY_CODE_KEY)`

hnnasit: okay

samcli/lib/package/s3_uploader.py
```diff
@@ -263,6 +263,28 @@ def parse_s3_url(
...
+            parsed = urlparse(url)
+            result = dict()
+            # path would point to /bucket/key
+            s3_bucket_key = parsed.path.split("/", 2)[1:]
+
+            result[bucket_name_property] = s3_bucket_key[0]
+            result[object_key_property] = s3_bucket_key[1]
+
+            return result
+        raise ValueError("URL given to the parse method is not a valid path-style S3 url " "{0}".format(url))
```
hawflau, line 286: looks like the two-part strings can just be concatenated into one?

samcli/lib/package/s3_uploader.py
```diff
@@ -263,6 +263,28 @@ def parse_s3_url(
 
         raise ValueError("URL given to the parse method is not a valid S3 url " "{0}".format(url))
 
+    @staticmethod
+    def parse_path_style_s3_url(
```
hawflau, line 267: Should we just combine `parse_s3_url` and `parse_path_style_s3_url` into one method to handle both `s3://` and `https://s3` cases? We can still have the two logic separated in two protected methods, but just expose one common method to handle any style of s3 url

samcli/lib/package/artifact_exporter.py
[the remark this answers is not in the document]
hnnasit: As discussed in our meeting, this change would change the way package command runs as well and this does not affect efficiency much so no change required here.

samcli/lib/package/s3_uploader.py
hnnasit: Yes that makes sense, Mohamed suggested the same thing to me.

moelasmar (approved):

hawflau (approved):

PR #3030, opened 2021-07-08
chore: removed unused code which was using pre-defined managed policy…

mndeveci:
… list and used in a sam translator wrapper, but the code path is not used.

N/A

The file that is removed contains pre-defined policy list which has never used.

By removing unused code path + the related configuration files

N/A

#### Checklist

- [ ] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


jfuss (approved):
🕺

hawflau (approved):
LGTM!

PR #3035, opened 2021-07-12
sam check command creation

jordon-ogo:
#### Checklist

- [x] Added sam check command
- [x] Template file is read and translated
- [x] Moved some functionality from sam_template_validator to another class
- [x] Added unit tests for main check command file

#### TODO
 - [x] Fix failing test_command.py test
 - [x] Looking into why AppVeyor test fails

Other than translating the template, nothing else is done with it as of now.

### DIFF
--- a/samcli/commands/check/command.py
+++ b/samcli/commands/check/command.py
@@ -25,6 +25,7 @@ from samcli.yamlhelper import yaml_dump
 from samcli.lib.utils.packagetype import ZIP
 
 from samcli.lib.replace_uri.replace_uri import ReplaceLocalCodeUri
+from .exceptions import InvalidSamDocumentException
 
 
 SHORT_HELP = "Checks template for bottle necks."
@@ -83,17 +84,33 @@ def cli(
     do_cli(ctx, template_file)  # pragma: no cover
 
 
-def do_cli(ctx, template):
+def do_cli(ctx, template_path):
     """
     Implementation of the ``cli`` method
 
     Translate template into CloudFormation yaml format
     """
 
-    # acquire template and policies
-    sam_template = _read_sam_file(template)
+    converted_template = transform_template(ctx, template_path)
+
+    click.echo("... analyzing application template")
+
+
+def load_policies():
     iam_client = boto3.client("iam")
-    managed_policy_map = ManagedPolicyLoader(iam_client).load()
+    return ManagedPolicyLoader(iam_client).load()
+
+
+def replace_code_uri(template):
+    uri_replace = ReplaceLocalCodeUri(template)
+    return uri_replace._replace_local_codeuri()
+
+
+def transform_template(ctx, template_path):
+    managed_policy_map = load_policies()
+    original_template = _read_sam_file(template_path)
+
+    updated_template = replace_code_uri(original_template)
 
     sam_translator = Translator(
         managed_policy_map=managed_policy_map,
@@ -102,19 +119,15 @@ def do_cli(ctx, template):
         boto_session=Session(profile_name=ctx.profile, region_name=ctx.region),
     )
 
-    # Convert uri's
-    uri_replace = ReplaceLocalCodeUri(sam_template)
-    sam_template = uri_replace._replace_local_codeuri()
-
     # Translate template
     try:
-        template = sam_translator.translate(sam_template=sam_template, parameter_values={})
+        converted_template = sam_translator.translate(sam_template=updated_template, parameter_values={})
     except InvalidDocumentException as e:
         raise InvalidSamDocumentException(
             functools.reduce(lambda message, error: message + " " + str(error), e.causes, str(e))
         ) from e
 
-    click.echo("... analyzing application template")
+    return converted_template
 
 
 def _read_sam_file(template):
--- a/samcli/commands/check/command.py
+++ b/samcli/commands/check/command.py
@@ -97,16 +97,25 @@ def do_cli(ctx, template_path):
 
 
 def load_policies():
+    """
+    Load user policies from iam account
+    """
     iam_client = boto3.client("iam")
     return ManagedPolicyLoader(iam_client).load()
 
 
 def replace_code_uri(template):
+    """
+    Replaces the uri's in the template into a format that sam translate can understand
+    """
     uri_replace = ReplaceLocalCodeUri(template)
     return uri_replace._replace_local_codeuri()
 
 
 def transform_template(ctx, template_path):
+    """
+    Takes a sam template or a CFN json template and converts it into a CFN yaml template
+    """
     managed_policy_map = load_policies()
     original_template = _read_sam_file(template_path)
 
--- a/samcli/commands/check/command.py
+++ b/samcli/commands/check/command.py
@@ -1,0 +1,141 @@
+"""
+CLI command for "deploy" command
+"""
+import os
+
+import logging
+import functools
+
+import click
+
+from samcli.cli.main import pass_context, common_options as cli_framework_options, aws_creds_options, print_cmdline_args
+from samcli.cli.cli_config_file import TomlProvider, configuration_option
+from samcli.lib.telemetry.metric import track_command
+from samcli.lib.utils.version_checker import check_newer_version
+from samcli.commands._utils.options import template_option_without_build
+
+from samtranslator.translator.translator import Translator
+from samtranslator.public.exceptions import InvalidDocumentException
+from .exceptions import InvalidSamDocumentException
+
+import boto3
+from samtranslator.translator.managed_policy_translator import ManagedPolicyLoader
+from samtranslator.parser import parser
+from boto3.session import Session
+from samcli.yamlhelper import yaml_dump
+from samcli.lib.utils.packagetype import ZIP
+
+from samcli.lib.replace_uri.replace_uri import ReplaceLocalCodeUri
+
+
+
+SHORT_HELP = "Checks template for bottle necks."
+
+
+HELP_TEXT = """
+Check your application to determine if any endpoints will not be able to
+provide the expected arival rate of data. You will need to provide the
+expected duration of each lambda function, as well as the expected 
+per-second arrival rate. You will then be informed of the expected cost
+of running this application, as well as any bottle necks that may exist.
+
+This command must be run in the main directory of your application. 
+This command will work on any SAM application. It can also run 
+on a CloudFormation template.
+
+Connections between resources can be made after all required data is
+provided
+"""
+
+CONFIG_SECTION = "parameters"
+LOG = logging.getLogger(__name__)
+
+
+@click.command(
+    "check",
+    short_help=SHORT_HELP,
+    help=HELP_TEXT,
+)
+@configuration_option(provider=TomlProvider(section="parameters"))
+@template_option_without_build
+# options go here with this format
+# @click.option(
+#     "--test",
+#     "-t",
+#     required=False,
+#     is_flag=True,
+#     help="Test number 1",
+# )
+@aws_creds_options
+@cli_framework_options
+@pass_context
+@track_command
+@check_newer_version
+@print_cmdline_args
+def cli(
+    ctx,
+    template_file,
+    config_file,
+    config_env,
+):
+    """
+    `sam deploy` command entry point
+    """
+    # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
+    do_cli(ctx, template_file)  # pragma: no cover
+
+
+def do_cli(ctx, template):
+    """
+    Implementation of the ``cli`` method
+
+    Translate template into CloudFormation yaml format
+    """
+
+    # acquire template and policies
+    sam_template = _read_sam_file(template)
+    iam_client = boto3.client("iam")
+    managed_policy_map = ManagedPolicyLoader(iam_client).load()
+
+    sam_translator = Translator(
+        managed_policy_map=managed_policy_map,
+        sam_parser=parser.Parser(),
+        plugins=[],
+        boto_session=Session(profile_name=ctx.profile, region_name=ctx.region),
+    )
+
+    # Convert uri's
+    uri_replace = ReplaceLocalCodeUri(sam_template)
+    sam_template = uri_replace._replace_local_codeuri()
+
+    # Translate template
+    try:
+        template = sam_translator.translate(sam_template=sam_template, parameter_values={})
+    except InvalidDocumentException as e:
+        raise InvalidSamDocumentException(
+            functools.reduce(lambda message, error: message + " " + str(error), e.causes, str(e))
+        ) from e
+
+    click.echo("... analyzing application template")
+
+
+def _read_sam_file(template):
+    """
+    Reads the file (json and yaml supported) provided and returns the dictionary representation of the file.
+
+    :param str template: Path to the template file
+    :return dict: Dictionary representing the SAM Template
+    :raises: SamTemplateNotFoundException when the template file does not exist
+    """
+
+    from samcli.commands.local.cli_common.user_exceptions import SamTemplateNotFoundException
+    from samcli.yamlhelper import yaml_parse
+
+    if not os.path.exists(template):
+        click.secho("SAM Template Not Found", bg="red")
+        raise SamTemplateNotFoundException("Template at {} is not found".format(template))
+
+    with click.open_file(template, "r", encoding="utf-8") as sam_template:
+        sam_template = yaml_parse(sam_template.read())
+
+    return sam_template
--- a/samcli/commands/check/command.py
+++ b/samcli/commands/check/command.py
@@ -12,8 +12,6 @@ from samcli.lib.telemetry.metric import track_command
 from samcli.lib.utils.version_checker import check_newer_version
 from samcli.commands._utils.options import template_option_without_build
 
-from .lib.command_context import CheckContext
-
 
 SHORT_HELP = "Checks template for bottle necks."
 
@@ -57,7 +55,7 @@ def cli(
     config_env,
 ):
     """
-    `sam deploy` command entry point
+    `sam check` command entry point
     """
     # All logic must be implemented in the ``do_cli`` method. This helps with easy unit testing
     do_cli(ctx, template_file)  # pragma: no cover
@@ -68,5 +66,7 @@ def do_cli(ctx, template_path):
     Implementation of the ``cli`` method
     """
 
-    context = CheckContext(ctx, template_path)
+    from samcli.commands.check.lib.command_context import CheckContext
+
+    context = CheckContext(ctx.region, ctx.profile, template_path)
     context.run()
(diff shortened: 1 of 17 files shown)


samcli/commands/check/command.py
```diff
@@ -0,0 +1,161 @@
+"""
+CLI command for "deploy" command
```
mndeveci, line 2: nit: copy paste error :), it says `deploy command`

samcli/commands/check/command.py
```diff
@@ -0,0 +1,161 @@
...
+@configuration_option(provider=TomlProvider(section="parameters"))
+@template_option_without_build
+# options go here with this format
+# @click.option(
+#     "--test",
+#     "-t",
+#     required=False,
+#     is_flag=True,
+#     help="Test number 1",
+# )
```
mndeveci, line 67: Please remove the testing code once you try to merge your changes.

samcli/commands/check/command.py
```diff
@@ -0,0 +1,161 @@
...
+from samcli.cli.cli_config_file import TomlProvider, configuration_option
+from samcli.lib.telemetry.metric import track_command
+from samcli.lib.utils.version_checker import check_newer_version
+from samcli.commands._utils.options import template_option_without_build
+
+from samtranslator.translator.translator import Translator
+from samtranslator.public.exceptions import InvalidDocumentException
+
+import boto3
+from samtranslator.translator.managed_policy_translator import ManagedPolicyLoader
```
mndeveci, line 21: Please move possible imports into cli implementation. Keeping imports here affects the latency of `--help`. Take this one for example; https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/deploy/command.py#L267

samcli/lib/replace_uri/replace_uri.py
```diff
@@ -0,0 +1,100 @@
+from samcli.lib.utils.packagetype import ZIP
+
+
+
+
+class ReplaceLocalCodeUri:
```
mndeveci, line 6: Great job extracting this into separate file. Looking at the class, there is only one instance with single method with no parameters. I believe this could be a single function instead of a class. Then you can directly call this from the `sam check` command as well.

mndeveci: Plus I didn't see unit tests for this new file. When you extract these please don't forget to update existing tests + add new ones to cover new lines.

samcli/commands/check/command.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Translate template into CloudFormation yaml format
+    """
+
+    converted_template = transform_template(ctx, template_path)
+
+    click.echo("... analyzing application template")
+
+
+def load_policies():
```
mndeveci, line 99: I see that you put your functionality into this `command.py` file/module. I would rather move them into [`samcli/lib`](https://github.com/aws/aws-sam-cli/tree/develop/samcli/lib) with its own module. Then you can keep this file just a bridge between the user inputs and the actual implementation details.

mndeveci (changes requested):

samcli/lib/replace_uri/replace_uri.py
jordon-ogo: The code for the class ReplaceLocalCodeUri (I will change it to a single function) is pulled from sam_template_validator. There is already test code for that, as the code previously existed there. Those tests test the code in the class ReplaceLocalCodeUri. If I were to write tests for the class ReplaceLocalCodeUri, they would be 100% identical to the tests for sam_template_validator, as that is where the code use to exist. Even though the tests would be 100% identical, is it still proper to have them written in their own file? It seems unnecessary to have the same code written in two different places. Should I instead "point" the current tests for sam_template_validator to instead test the class ReplaceLocalCodeUri directly, since that is where the logic is now found?

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -136,7 +94,7 @@ def is_s3_uri(uri):
             Returns True if the uri given is an S3 uri, otherwise False
 
         """
-        return isinstance(uri, str) and uri.startswith("s3://")
+        return external_is_s3_uri(uri)
```
hawflau, line 97: will there be any chance that the input uri is a path style s3 uri? e.g. `https://s3.amazonaws.com/bucket/key` or `https://s3-us-east-2.amazonaws.com/bucket/key`?

samcli/lib/replace_uri/replace_uri.py
```diff
@@ -0,0 +1,94 @@
...
+                + [_properties.get("PackageType", ZIP) == ZIP for _, _properties in global_settings.items()]
+            ):
+                _update_to_s3_uri("CodeUri", properties)
+
+    for _, resource in all_resources.items():
+
+        resource_type = resource.get("Type")
+        resource_dict = resource.get("Properties", {})
+
+        if resource_type == "AWS::Serverless::Function" and resource_dict.get("PackageType", ZIP) == ZIP:
```
hawflau, line 32: Are these (L32 - L50) all the resources that we need to substitute local path with fake s3 uri? If you take a look `samcli/commands/_utils/resources.py`, you will see the list of all resources that can contain local paths. Do they need to be handled as well?

samcli/lib/replace_uri/replace_uri.py
```diff
@@ -0,0 +1,94 @@
+from samcli.lib.utils.packagetype import ZIP
+
+
+def _replace_local_codeuri(template):
```
hawflau, line 4: despite what the docstring explains, I feel the name of the function is a bit ambiguous. Can it be sth like `_replace_local_path_with_fake_s3_uri` that is less ambiguous about what it exactly does?

samcli/commands/validate/lib/sam_template_validator.py
jordon-ogo: I'm not entirely sure. The only changes I made to sam_template_validator is I extracted the code logic into another class so that I could use it in `sam check`.

samcli/lib/replace_uri/replace_uri.py
jordon-ogo: I'm not sure. This code was extracted and used from the already existing sam_template_validator. If there are resources that you believe are missing, then they were not implemented before I extracted this code.

samcli/lib/replace_uri/replace_uri.py
jordon-ogo: Since all of the code in this file was copied from sam_template_validator, I wanted to keep it as consistent and similar to what it originally was. If you still think that updating the function name would improve its readability, I can change it.

samcli/commands/check/command.py
```diff
@@ -0,0 +1,72 @@
...
+@check_newer_version
+@print_cmdline_args
+def cli(
+    ctx,
+    template_file,
+    config_file,
+    config_env,
+):
+    """
+    `sam deploy` command entry point
```
mndeveci, line 60: still says "deploy" :)

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,79 @@
+import os
```
mndeveci, line 1: Please add documentation about the file/module

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,79 @@
...
+from boto3.session import Session
+
+from samcli.commands.local.cli_common.user_exceptions import SamTemplateNotFoundException
+from samcli.yamlhelper import yaml_parse
+
+from samcli.lib.replace_uri.replace_uri import _replace_local_codeuri as external_replace_local_codeuri
+from ..exceptions import InvalidSamDocumentException
+
+
+class CheckContext:
```
mndeveci, line 20: Please add documentation about class. Why we need it, what it does etc.

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, ctx, template_path):
```
mndeveci, line 21: 1. Please add typing to the variables.
2. If that `ctx` is coming from `click` library, please keep it outside of this class but pass attributes which you will be using in this one

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,79 @@
...
+        :param str template: Path to the template file
+        :return dict: Dictionary representing the SAM Template
+        :raises: SamTemplateNotFoundException when the template file does not exist
+        """
+
+        if not os.path.exists(self.template_path):
+            click.secho("SAM Template Not Found", bg="red")
+            raise SamTemplateNotFoundException("Template at {} is not found".format(self.template_path))
+
+        with click.open_file(self.template_path, "r", encoding="utf-8") as sam_template:
```
mndeveci, line 76: Why we are using `click.open` but not `open`?

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Reads the file (json and yaml supported) provided and returns the dictionary representation of the file.
+
+        :param str template: Path to the template file
+        :return dict: Dictionary representing the SAM Template
+        :raises: SamTemplateNotFoundException when the template file does not exist
+        """
+
+        if not os.path.exists(self.template_path):
+            click.secho("SAM Template Not Found", bg="red")
```
mndeveci, line 73: Don't use `click` in `lib` folder (See; https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#code-conventions). You can log this as error, or you can raise an exception and catch it in the command and print there.

samcli/lib/replace_uri/replace_uri.py
```diff
@@ -0,0 +1,94 @@
+from samcli.lib.utils.packagetype import ZIP
```
mndeveci, line 1: Please add documentation for module.

mndeveci (changes requested):
Added some comments, please also check why some validations are failing, did you run `make pr` before pushing changes?

See your failure here; https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/39980013

samcli/commands/check/lib/command_context.py
jordon-ogo: Added a brief description of the file used in the documentation for "_read_sam_file"

samcli/commands/check/lib/command_context.py
jordon-ogo: 1. Do you mean in the parameters themselves (i.e. ctx: str, template_path: str), or in the documentation for the class like here: "https://github.com/aws/aws-sam-cli/blob/59c85768356089edb265c2ea7f53bce2412f9e19/samcli/commands/deploy/deploy_context.py#L190"?
2. Ok.

samcli/commands/check/lib/command_context.py
jordon-ogo: This is the same way other functions in sam check open the template file

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,89 @@
+import os
+import functools
+
+import click
+import boto3
+from boto3.session import Session
+
+from samcli.commands.check.command import LOG
```
mndeveci: Please create a new logger instead of using the one from the check command.

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, region, profile, template_path):
+        self.region = region
+        self.profile = profile
+        self.template_path = template_path
+
+    def run(self):
+        converted_template = self.transform_template()
+
+        click.echo("... analyzing application template")
```
mndeveci, line 39: Let's don't use click as much as we can in `lib` folder. I would prefer to use `LOG.info` for now.

samcli/commands/check/lib/command_context.py
mndeveci: But there is builtin `open` function which works exactly same, we should be using builtin one not the one from dependency.

```
with open(self.template_path, "r", encoding="utf-8") as sam_template:
```

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -9,7 +9,9 @@
 from samtranslator.translator.translator import Translator
 from boto3.session import Session
 
-from samcli.lib.utils.packagetype import ZIP
+from samcli.lib.replace_uri.replace_uri import _replace_local_codeuri as external__replace_local_codeuri
```
mndeveci, line 12: nit: import has 2 underscores `external__replace_local_codeuri`

mndeveci (changes requested):

hawflau (approved):

samcli/lib/replace_uri/replace_uri.py
moelasmar: I believe this comment is a valid comment.
Also I think we need to handle the Image Uri for Lambda functions of PackageType = Image

samcli/commands/check/command.py
```diff
@@ -0,0 +1,72 @@
...
+@cli_framework_options
+@pass_context
+@track_command
+@check_newer_version
+@print_cmdline_args
+def cli(
+    ctx,
+    template_file,
+    config_file,
+    config_env,
```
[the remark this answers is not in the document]
jordon-ogo: Currently I do not use them, but I am planning to make modifications to the config file. Just not in this branch. I will remove them until they become necessary.

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,92 @@
...
+        self.template_path = template_path
+
+    def run(self):
+        self.transform_template()
+
+        LOG.info("... analyzing application template")
+
+    def transform_template(self):
+        """
+        Takes a sam template or a CFN json template and converts it into a CFN yaml template
```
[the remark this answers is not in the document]
jordon-ogo: We technically do transform a CFN json to a CFN yaml. If a user wants to use a CFN json template, they can do so and `sam check` will still be able to read it, because it is transformed. CFN yaml is the end goal, regardless of whether the translator starts with a SAM or CFN json template

samcli/commands/check/lib/command_context.py
```diff
@@ -0,0 +1,92 @@
...
+            LOG.error("SAM Template Not Found")
+            raise SamTemplateNotFoundException("Template at {} is not found".format(self.template_path))
+
+        with open(self.template_path, "r", encoding="utf-8") as sam_template:
+            sam_template = yaml_parse(sam_template.read())
+
+        return sam_template
+
+
+def load_policies():
```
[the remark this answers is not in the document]
jordon-ogo: Correct, but while I use `sam_template_validator`, I do not run `sam validate`, therefore it does not get run twice if that was your concern. From what I can see there is not a common method to sue for both `sam validate` and `sam check`. This is also used in `wrapper.py`. I don't believe extracting this into its own method is necessary, given how small it is.

samcli/commands/check/command.py
```diff
@@ -0,0 +1,72 @@
...
+LOG = logging.getLogger(__name__)
+
+
+@click.command(
+    "check",
+    short_help=SHORT_HELP,
+    help=HELP_TEXT,
+)
+@configuration_option(provider=TomlProvider(section="parameters"))
+@template_option_without_build
```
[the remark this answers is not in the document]
jordon-ogo: This is not necessary. It has been removed.

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -9,7 +9,9 @@
 from samtranslator.translator.translator import Translator
 from boto3.session import Session
 
-from samcli.lib.utils.packagetype import ZIP
+from samcli.lib.replace_uri.replace_uri import _replace_local_codeuri as external_replace_local_codeuri
```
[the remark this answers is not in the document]
jordon-ogo: Fixed

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -73,52 +75,7 @@ def is_valid(self):
...
-                if "DefinitionUri" in resource_dict:
-                    SamTemplateValidator._update_to_s3_uri("DefinitionUri", resource_dict)
-
-            if resource_type == "AWS::Serverless::StateMachine":
-                if "DefinitionUri" in resource_dict:
-                    SamTemplateValidator._update_to_s3_uri("DefinitionUri", resource_dict)
+        return external_replace_local_codeuri(self.sam_template)
 
     @staticmethod
     def is_s3_uri(uri):
```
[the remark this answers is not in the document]
jordon-ogo: Same comment as above

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -73,52 +75,7 @@ def is_valid(self):
             ) from e
 
     def _replace_local_codeuri(self):
```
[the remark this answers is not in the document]
jordon-ogo: My intention for leaving them as is is because I wanted to leave `sam_template_validator` as close to how it was originally formatted before I made my changes to it. My intention was not to replace those methods, but just to extract them so I could get the data from them. If you still believe that they should be removed, let me know and I will do that.

samcli/commands/validate/lib/sam_template_validator.py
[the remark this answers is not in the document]
jordon-ogo: For some reason I cannot comment on your last comment below: "I believe this comment is a valid comment..."
Everything in `replace_uri.py` came from `sam_template_validator`. It is out of scope for this project for me to implement that feature.

samcli/commands/check/lib/command_context.py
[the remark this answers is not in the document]
moelasmar: my understanding is mainly our target here is to transform the SAM resources to CFN resources, and then we can do our checks on the CFN resources.

samcli/commands/check/lib/command_context.py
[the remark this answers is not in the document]
moelasmar: my concern is not related to be run twice, but mainly that the same logic should be maintained in one place even if it is a small.

samcli/commands/validate/lib/sam_template_validator.py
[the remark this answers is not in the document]
moelasmar: you already moved the method to be used from a new module so I think you should remove the old one, and use the new method instead. If the old method is used anywhere else in the code, I think it should be refactored to use the new method.

samcli/lib/replace_uri/replace_uri.py
moelasmar: > For some reason I cannot comment on your last comment below: "I believe this comment is a valid comment..."
> Everything in replace_uri.py came from sam_template_validator. It is out of scope for this project for me to implement that feature.

Ok

samcli/commands/check/command.py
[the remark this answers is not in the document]
jordon-ogo: Correction. Without it the `def cli` in `sam check` fails to retrieve the template_file, which is required.

moelasmar (approved):

Issue #1362, opened 2019-08-22
sam local generate-event stepfunctions error generates empty payload


makstaks:
### Description

When I try the command `sam local generate-event stepfunctions error` I get an empty payload.  

### Steps to reproduce

1. Type the following command in your terminal:
`sam local generate-event stepfunctions error` 

### Observed result

```
2019-08-22 14:17:38 Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
{}
2019-08-22 14:17:38 Sending Telemetry: {'metrics': [{'commandRun': {'awsProfileProvided': False, 'debugFlagProvided': True, 'region': '', 'commandName': 'sam local generate-event stepfunctions error', 'duration': 0, 'exitReason': 'success', 'exitCode': 0, 'requestId': '89f5019a-6b1d-40b1-8108-1a47cffea09c', 'installationId': 'd9166b69-80c5-4d1f-9551-a8d32fa4e6a4', 'sessionId': '932b92a5-a442-4f85-a907-4926a9a94c2b', 'executionEnvironment': 'CLI', 'pyversion': '3.7.4', 'samcliVersion': '0.19.0'}}]}
2019-08-22 14:17:38 Starting new HTTPS connection (1): aws-serverless-tools-telemetry.us-west-2.amazonaws.com:443
2019-08-22 14:17:38 https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com:443 "POST /metrics HTTP/1.1" 200 2
2019-08-22 14:17:38 Telemetry response: **200**

```
### Expected result

Payload with content mocking an error.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS Mojave 10.14.6 (18G87)
2. `sam --version`: SAM CLI, version 0.19.0

jfuss:
Removing bug tag. 

This could be improved further but this was an empty dictionary on purpose and to match what is in the AWS Console for this event type.

torresxb1:
Closing because the PR to add the sample payload has been merged

PR #3043, opened 2021-07-13
fix: add sample payload for 'sam local generate-event stepfunctions error'

torresxb1:
#1362

We currently provide an empty object as the sample payload for this command. The event payload from an error caught in stepfunctions does have some structure (see "Error output" section [here]( https://states-language.net/spec.html#errors)). Specifically, it should include an `Error` and `Cause`. 

This PR adds a sample payload with an `Error` and `Cause`, and allows the user to customize their own values for those properties. 

None

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/generated_sample_events/event-mapping.json
```diff
@@ -992,7 +992,16 @@
     "error": {
       "filename": "StepFunctionsError",
       "help": "Generates an AWS StepFunctions Error Event",
-      "tags": {}
+      "tags": {
+        "error": {
+          "type": "string",
+          "default": "SomeError"
```
mgrandis, line 998: nit: Maybe `ErrorName` would be just tad better than `SomeError`

mgrandis (approved):

mndeveci (approved):

wchengru (approved):

PR #3050, opened 2021-07-14
Get s3 info cf template

hnnasit:
sam delete currently gets s3 information from samconfig.toml file. This change helps get the s3 info when local config file is not present.
This solves the issue when there is a prefix with artifacts present for a deployed stack that needs to be deleted.
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/artifact_exporter.py
```diff
@@ -18,6 +18,7 @@
 import os
 from typing import Dict, Optional
 
+# import logging
```
hawflau, line 21: If we don't need it, why not just remove it?

samcli/lib/package/artifact_exporter.py
```diff
@@ -265,3 +266,47 @@ def delete(self):
...
                     exporter = exporter_class(self.uploaders, None)
                     exporter.delete(resource_id, resource_dict)
+
+    def get_s3_info(self):
+        """
+        Iterates the template_dict resources with S3 EXPORT_DESTINATION to get the
+        s3_bucket and s3_prefix information for the purpose of deletion.
+        Method finds the first resource with s3 information, extracts the information
+        and then terminates.
+        """
```
hawflau, line 276: I think we should also mention it is safe to assume all packaged files are in the same S3 bucket and with the same prefix, because this is how `sam package` and `sam deploy` works now.

samcli/lib/package/packageable_resources.py
```diff
@@ -322,13 +330,22 @@ def delete(self, resource_id, resource_dict):
         """
         if resource_dict is None:
             return
+
+        s3_info = self.get_s3_info(resource_dict)
+        self.uploader.delete_artifact(remote_path=s3_info["Key"], is_key=True)
+
+    def get_s3_info(self, resource_dict):
```
moelasmar, line 337: this class is to handle the resources that publish their own artifacts to ECR, so I am not able to understand the S3 part on this class

samcli/lib/package/artifact_exporter.py
hnnasit: That is a good point, this logic would only work for stacks which are created by sam. In the case we get bucket information which is not correct, SAM will not be able to tell that it is wrong and the template file wont be deleted.

samcli/commands/delete/delete_context.py
```diff
@@ -149,6 +149,20 @@ def delete(self):
         with mktempfile() as temp_file:
             self.cf_template_file_name = get_cf_template_name(temp_file, template_str, "template")
 
+        template = Template(
+            template_path=None, parent_dir=None, uploaders=self.uploaders, code_signer=None, template_str=template_str
+        )
+
+        # If s3 info is not available, try to obtain it from CF
```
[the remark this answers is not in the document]
hnnasit: Yes correct we get the s3 info from the cloudformation stack template and not cloudformation.

samcli/lib/package/packageable_resources.py
hnnasit: So the classes ResourceWithS3UrlDict and ResourceZip upload their artifacts to S3 and so they have the methods get_s3_info which extracts the s3_bucket and key from the template file for theses resources.

hawflau (approved):

moelasmar (approved):

PR #3060, opened 2021-07-15
ci: Speed up unit test by caching the git clone

aahung:

**To make `make pr` great again**

Speed up the `make pr` (specifically `make test`) and reduce its time by 65%

```
2346 passed, 12 skipped, 20 warnings in 103.91s (0:01:43)
to
2346 passed, 12 skipped, 20 warnings in 35.22s
```

Avoid making `git clone` on the same URL over and over again.

N/A

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/git_repo.py
+++ b/samcli/lib/utils/git_repo.py
@@ -6,6 +6,9 @@ import platform
 import shutil
 import subprocess
 from pathlib import Path
+
+# import check_output alone so that it can be patched without affecting
+# other parts of subprocess.
 from subprocess import check_output
 from typing import Optional
 
--- a/samcli/lib/utils/git_repo.py
+++ b/samcli/lib/utils/git_repo.py
@@ -6,6 +6,7 @@ import platform
 import shutil
 import subprocess
 from pathlib import Path
+from subprocess import check_output
 from typing import Optional
 
 from samcli.lib.utils import osutils
@@ -118,7 +119,7 @@ class GitRepo:
                 temp_path = os.path.normpath(os.path.join(tempdir, clone_name))
                 git_executable: str = GitRepo._git_executable()
                 LOG.info("\nCloning from %s", self.url)
-                subprocess.check_output(
+                check_output(
                     [git_executable, "clone", self.url, clone_name],
                     cwd=tempdir,
                     stderr=subprocess.STDOUT,
--- a/tests/unit/commands/init/test_cli.py
+++ b/tests/unit/commands/init/test_cli.py
@@ -1,5 +1,9 @@
 import os
+import shutil
+import subprocess
+import tempfile
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


samcli/lib/utils/git_repo.py
```diff
@@ -6,6 +6,7 @@
 import shutil
 import subprocess
 from pathlib import Path
+from subprocess import check_output
```
hoffa, line 9: A "why" comment here will be helpful.

hoffa (commented):
Great initiative 👍

samcli/lib/utils/git_repo.py
```diff
@@ -118,7 +119,7 @@ def clone(self, clone_dir: Path, clone_name: str, replace_existing: bool = False
                 temp_path = os.path.normpath(os.path.join(tempdir, clone_name))
                 git_executable: str = GitRepo._git_executable()
                 LOG.info("\nCloning from %s", self.url)
-                subprocess.check_output(
+                check_output(
                     [git_executable, "clone", self.url, clone_name],
```
hoffa, line 123: What if we shallow-clone instead? So that everyone benefits from `sam init` and there's no risk to the patched test code diverging from prod code.

aahung: Added, and it alone produced time ranging from 37s to 67s.

samcli/lib/utils/git_repo.py
aahung: added

samcli/lib/utils/git_repo.py
```diff
@@ -118,8 +122,8 @@ def clone(self, clone_dir: Path, clone_name: str, replace_existing: bool = False
                 temp_path = os.path.normpath(os.path.join(tempdir, clone_name))
                 git_executable: str = GitRepo._git_executable()
                 LOG.info("\nCloning from %s", self.url)
-                subprocess.check_output(
-                    [git_executable, "clone", self.url, clone_name],
+                check_output(
+                    [git_executable, "clone", "--shallow", self.url, clone_name],
```
hoffa, line 126: What's `--shallow`? Never seen or used it, nor can I find it in the docs: https://git-scm.com/docs/git-clone

aahung: removed, I was using a non-existing flag..

samcli/lib/utils/git_repo.py
aahung: reverted, wrong testing data

mndeveci (approved):
LGTM! And great improvement, thanks for taking this initiative! 🎉

c2tarun (approved):
LGTM!

hoffa (approved):

PR #3077, opened 2021-07-19
Add integ tests for package for cdk

hawflau:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/package/package_integ_base.py
```diff
@@ -129,3 +150,83 @@ def get_command_list(
...
+class CdkPackageIntegPythonBase(CdkPackageIntegBase):
+    def setUp(self):
+        super().setUp()
+        self._create_virtual_env()
+
+    def tearDown(self):
+        super().tearDown()
+
+    def _create_virtual_env(self):
+        create_venv_command = ["python3", "-m", "venv", ".venv"]
```
wchengru, line 218: any python3 version is fine, or do we want to specify a python3 version?
Also, the integration test runs with venv https://github.com/aws/aws-sam-cli/blob/1c42df379b442bd8fa16fcc6090035bdc884aa56/appveyor.yml#L122
We should recover the venv back after this test case.

tests/integration/package/package_integ_base.py
```diff
@@ -129,3 +150,83 @@ def get_command_list(
...
+        return command_list
+
+
+class CdkPackageIntegPythonBase(CdkPackageIntegBase):
+    def setUp(self):
+        super().setUp()
+        self._create_virtual_env()
+
+    def tearDown(self):
+        super().tearDown()
```
wchengru, line 215: Probably delete the .venv folder?

wchengru (commented):
LGTM, only a couple of small comments.

tests/integration/package/package_integ_base.py
hawflau: any python3 version is fine. The venv is used only to install cdk dependencies that are required for synthesize. 
And since the command is run with `cwd` set as `self.working_dir` (which is a temp folder created in each `setUp()` call), the venv actually lives under `self.working_dir`

tests/integration/package/package_integ_base.py
hawflau: Since `self.working_dir` is removed during each `tearDown` call, the venv is removed as well.

wchengru (approved):

mildaniel (approved):

PR #3080, opened 2021-07-20
Auto ECR Companion Stack Deletion

hnnasit:
Allows users to delete auto ecr companion stacks
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/delete/delete_context.py
```diff
@@ -137,6 +155,85 @@ def guided_prompts(self):
...
+    def ecr_companion_stack_prompts(self):
+        """
+        User prompt to delete the ECR companion stack.
+        """
+        click.echo(f"\tFound ECR Companion Stack {self.companion_stack_name}")
+        if not self.no_prompts:
+            delete_ecr_companion_stack_prompt = confirm(
+                click.style(
+                    "\tDo you you want to delete the ECR companion stack"
+                    + f" {self.companion_stack_name} in the region {self.region} ?",
```
hawflau, line 167: bit: `+` is not needed here. Two strings can just concatenate without `+`

samcli/lib/delete/cf_utils.py
```diff
@@ -16,6 +16,7 @@
 class CfUtils:
     def __init__(self, cloudformation_client):
         self._client = cloudformation_client
+        # self._resource_client = cloudformation_resource_client
```
hawflau, line 19: Please remove the line if it is not needed

samcli/lib/package/ecr_uploader.py
```diff
@@ -114,35 +113,48 @@ def delete_artifact(self, image_uri: str, resource_id: str, property_name: str):
                 # Image not found
                 image_details = resp["failures"][0]
                 if image_details["failureCode"] == "ImageNotFound":
-                    LOG.error("ImageNotFound Exception")
-                    message_fmt = (
-                        "Could not delete image for {property_name}"
-                        " parameter of {resource_id} resource as it does not exist. \n"
+                    LOG.debug(
+                        "Could not delete image for %s" " parameter of %s resource as it does not exist. \n",
```
hawflau, line 117: nit: the two strings can be merged into one

samcli/lib/package/packageable_resources.py
```diff
@@ -517,6 +518,27 @@ class CloudFormationResourceVersionSchemaHandlerPackage(ResourceZip):
     PROPERTY_NAME = RESOURCES_WITH_LOCAL_PATHS[AWS_CLOUDFORMATION_RESOURCEVERSION][0]
 
 
+class ECRResource(Resource):
```
hawflau, line 521: Can you add a docstring explaining what this class is for?
Seems like this class is not for exporting at all. Only for deleting

samcli/lib/package/ecr_uploader.py
```diff
@@ -114,35 +113,48 @@ def delete_artifact(self, image_uri: str, resource_id: str, property_name: str):
...
+        :param: physical_id of the repository to be deleted
+        """
+        try:
+            click.echo(f"\t- Deleting ECR repository {physical_id}")
+            self.ecr_client.delete_repository(repositoryName=physical_id, force=True)
+        except self.ecr_client.exceptions.RepositoryNotFoundException:
+            # If the repository is empty, cloudformation automatically deletes
+            # the repository when cf_client.delete_stack is called.
+            LOG.debug("Could not find repository %s", physical_id)
+            pass
```
hawflau, line 152: This seems to be unnecessary

samcli/commands/delete/delete_context.py
```diff
@@ -27,6 +35,9 @@
 
 
 class DeleteContext:
+
+    ecr_repos: Dict[str, Dict[str, str]]
```
moelasmar, line 39: why this variable is a class level variable?

samcli/commands/delete/delete_context.py
```diff
@@ -69,22 +82,28 @@ def parse_config_file(self):
...
                 self.stack_name = config_options.get("stack_name", None)
             # If the stack_name is same as the one present in samconfig file,
-            # get the information about parameters if not specified by customer.
+            # get the information about parameters if not specified by user.
             if self.stack_name and self.stack_name == config_options.get("stack_name", None):
                 LOG.debug("Local config present and using the defined options")
                 if not self.region:
                     self.region = config_options.get("region", None)
-                    click.get_current_context().region = self.region
+                    Context.get_current_context().region = self.region
```
moelasmar, line 90: do we need to double check that the region we get it from the config file is not none before updating the context ?

samcli/commands/delete/delete_context.py
```diff
@@ -69,22 +82,28 @@ def parse_config_file(self):
...
                 if not self.profile:
                     self.profile = config_options.get("profile", None)
-                    click.get_current_context().profile = self.profile
+                    Context.get_current_context().profile = self.profile
```
moelasmar, line 93: same as the previous comment

samcli/commands/delete/delete_context.py
hnnasit: That makes sense, context should be updated after profile and region are assigned values other than None.

moelasmar (approved):

hawflau (approved):

Issue #1893, opened 2020-03-25
Support ssh git dependencies in pip requirements.txt when building with use-container


johnc44:
### Describe your idea/feature/enhancement

In a requirements.txt file I can have something like this:

git+ssh://git@bitbucket.org/path/to/my.git

If I do a sam build on this, it works.  But if I add --use-container it does not.

  Host key verification failed.
  fatal: Could not read from remote repository.

I presume this is a combination of my SSH key plus known_hosts not being taken from the host.

### Proposal

The only 2 solutions I can think of are either to copy the keys onto the container (hmm) or to clone them on the host, copy them onto the container then run pip install on them.

Neither seems ideal to me (and I can see why you wouldn't want to do the first one), but I'm hoping that there is maybe something I've not thought of.

I'd like to continue to use --use-container but we also need a way to share code between lambdas and this seems to me to be the easiest.  I don't know much about hosting a private pypi yet but I presume I'll run into a similar problem there.

johnc44:
After a bit more rummaging around I found this: https://medium.com/@tonistiigi/build-secrets-and-ssh-forwarding-in-docker-18-09-ae8161d066

which sounds like it's at least theoretically possible to proxy ssh through the host.

nickswiss:
Any update here? Seems like this should be supported? I am installing python packages which may build differently based on the operating system, so I use `--use-container` command to avoid this issue.

```
sam build --use-container
```

I have private python packages (hosted on enterprise github) which I need install via ssh. I feel like secrets are a very common necessity in docker builds so I'm either surprised or cannot find the documentation explaining how to do this. 

The article posted by @johnc44 requires secrets so really if you guys added something which allowed passing secrets, or any docker flags alongsied the `--use-container` command, I'd greatly appreciate it.

cmurtaugh:
I can build an image outside of SAM using a command like: 
```
docker build --ssh build_ssh_key='/path/to/my/key' .
```
and then referencing that key in my Dockerfile like: 
```
RUN --mount=type=ssh,id=build_ssh_key pip install -r requirements.txt 
```

But if I try to build the same app using SAM, adding this to my function resource:
```
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./my_function
      DockerBuildArgs:
        ssh: "build_ssh_key='/path/to/my/key'"
```
the build fails with a publickey authentication error from GitHub.

Macok:
I implemented this in #3084.
[...]

PR #3084, opened 2021-07-20
Feature: Support dependencies accessed over SSH in containerized builds

Macok:
#1893

This change is needed to support git dependencies accessed over SSH, for builds executed inside docker containers (`--use-container`, or `-u` option).

This PR adds `--container-dir-mount` option to SAM interface, allowing the user to mount arbitrary directories into container where the build is executed.
Having this option available, the user can forward host's SSH configuration into the container by mounting `$SSH_AUTH_SOCK` and `known_hosts` files:

```
sam build -u --container-dir-mount $SSH_AUTH_SOCK:/ssh-agent -e SSH_AUTH_SOCK=/ssh-agent --container-dir-mount ~/.ssh/known_hosts:/etc/ssh/ssh_known_hosts
```

**This functionality integrates easily with Github Actions.** You can configure SSH on the GA worker using this plugin: [webfactory/ssh-agent](https://github.com/webfactory/ssh-agent). Then forward worker's SSH configuration to the container as shown above.

None, fully backward compatible

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/build_strategy.py
```diff
@@ -118,6 +118,7 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
         # we should create a copy and pass it down, otherwise additional env vars like LAMBDA_BUILDERS_LOG_LEVEL
         # will make cache invalid all the time
         container_env_vars = deepcopy(build_definition.env_vars)
+        container_dir_mounts = deepcopy(build_definition.dir_mounts)
```
Macok, line 121: Here in `build_single_function_definition` a copy of `build_definition.env_vars` is made.
Below however, in `build_single_layer_definition` there is no copy. Is this intended, or should it be fixed?

aidan-plenert-macdonald:
@jfuss Can we get this reviewed? It would be useful for us. Thanks!

pkspks:
This is useful for building with git repo dependencies in `requirements.txt`. Can not build in a container without this.

braaibander:
I would really like to see this functionality implemented
We have private repos and varing python versions, this is a real problem for us

samcli/commands/build/command.py
```diff
@@ -6,6 +6,7 @@
 import logging
 from typing import List, Optional, Dict, Tuple
 import click
+from pathvalidate import ValidationError, validate_filepath
```
qingchm, line 9: This seems like a new library licensed from MIT, I would recommend to have our own logic of path verification. To use this library would require some security discussions and at least some extra actions like pinning down a dependency version in the requirements file. If possible let's have a simple method for path validation and catch logic!

samcli/commands/build/command.py
```diff
@@ -128,6 +129,16 @@
     help="Path to environment variable json file (e.g., env_vars.json) to pass into build containers",
     cls=ContainerOptions,
 )
+@click.option(
+    "--container-dir-mount",
```
qingchm, line 133: For this option name and also the name in the build.toml file I will briefly consult our UX team on it! dir-mount doesn't sound exactly intuitive to me. Sorry but naming such options requires some extra attention!

samcli/commands/build/command.py
```diff
@@ -128,6 +129,16 @@
...
+    "-v",
+    default=None,
+    multiple=True,
+    required=False,
+    help="Directories to be mounted into docker container."
+    " E.g. sam build --use-container --container-dir-mount /host/path:/container/path",
```
qingchm, line 139: Since our builds support customizing different images for each build function, does it make more sense to also support function specific mount dir like --container-dir-mount Function1=/host/path:/container/path?

qingchm:
Generally I really like how detailed your changes are, it really respects some of the conventions used in our codebase! Design-wise left some comments above! Excited to help you push this through

Macok:
@qingchm 
Thanks for your review. I spent some time to resolve conflicts that arose between my branch and `develop`.
Regarding your comments:
- Using pathvalidate library:
I removed this lib and used simple regexes as you suggested.
- Choosing different name for `--container-dir-mount` option:
Did you talk to the UX team? I'm not a native speaker so I will just take what they suggest.
- Supporting function specific mounts like `Function1=/host/path:/container/path`:
For now I just put a TODO in `app_builder.py`.
I think this use case would be very rare and I'd rather keep this PR small to increase chances of merging it.
But if you think this is a must have I will try to find some time to work on it.

Macok:
@qingchm 
Did you have some time to look into this?

Macok:
> Thanks for this PR @Macok, I've updated your branch with latest from develop. I will check with @qingchm to see UX feedback.

Thanks @mndeveci , let me know once there are any news!

tniemann:
any process here?

moelasmar:
Thanks @Macok for your contribution. I updated your branch with the latest changes from the develop branch.

We need to ask our security team feedback about this change. We will initiate the review process, and update this PR with the updates.

NancyMakar:
can we get updates on this please?

jfuss:
Providing an update:

We are still working with our security team. Given re:invent and holidays, it's a tough time. This is something we are still trying to drive, but the process is just taking longer than anticipated. Unfortunately, I don't have anything more concrete to share at the moment or an ETA. We will share once we can. Thanks for understanding.

sriram-mv:
After looking into this PR a bit more as we started on the security review.  One way to go about this without having to adding new options and flags to AWS SAM CLI is to use the powerful `--build-image` option with sam build. A Dockerfile can be created and built with `--ssh` flag and the local image can then be supplied to `sam build --build-image mybuildimage:latest`

sriram-mv:
Closing this PR based on the latest comment. Will look into adding an example here.

Jonathan-Landeed:
@sriram-mv isn't `--ssh` for exposing ssh keys during the `docker buildx build` process? The keys will not be available when `sam build` actually runs in the resulting container. I suppose it might be possible by installing all requirements into the docker image and then using `--cached`, but that would be a pain.

Most networking or volume mounting tricks I can think of would be done when starting the container (which sam controls), not building the image.

The only thing I can think of is to bundle the ssh keys like so:
```
FROM public.ecr.aws/sam/build-python3.10:latest-x86_64

RUN ssh-keyscan github.com > /etc/ssh/ssh_known_hosts

COPY deploy_key /root/.ssh/deploy_key

RUN chmod 600 /root/.ssh/deploy_key && \
    chmod 700 /root/.ssh

RUN echo "    IdentityFile /root/.ssh/deploy_key" >> /etc/ssh/ssh_config
```

Not ideal from a security standpoint...

Issue #2250, opened 2020-09-24
New versions of SAM should clean up old docker images


abrooksv:
### Description
Now that a rapid layer is built and version specific, old versions should be cleaned up.

### Steps to reproduce
Upgrade and run a local invoke

### Observed result

```
amazon/aws-sam-cli-emulation-image-dotnetcore2.1                                                    rapid-1.0.0                                              9e9ef66ff35e        2 months ago        989MB
amazon/aws-sam-cli-emulation-image-java8                                                            rapid-1.0.0                                              05211eee97a6        2 months ago        662MB
amazon/aws-sam-cli-emulation-image-java8                                                            rapid-1.1.0                                              05211eee97a6        2 months ago        662MB
amazon/aws-sam-cli-emulation-image-java11                                                           rapid-1.0.0                                              8de94beb46f4        2 months ago        417MB
amazon/aws-sam-cli-emulation-image-java11                                                           rapid-1.1.0                                              8de94beb46f4        2 months ago        417MB
amazon/aws-sam-cli-emulation-image-nodejs10.x                                                       rapid-1.0.0                                              57fb8d5be66c        2 months ago        361MB
amazon/aws-sam-cli-emulation-image-nodejs12.x                                                       rapid-1.0.0                                              b64e2bfa4562        2 months ago        366MB
amazon/aws-sam-cli-emulation-image-python3.7                                                        rapid-1.2.0                                              95a05d06fb87        5 days ago          889MB
amazon/aws-sam-cli-emulation-image-python3.6                                                        rapid-1.2.0                                              9531da666779        5 days ago          849MB
```

awood45:
This came from a conversation I had with Austin. An additional facet to this is that these image tags sometimes mask updates to the underlying emulation image. A feature request could also be for us to occasionally detect and help force updates when `--skip-pull-image` is not set.

johnnatan-jarac:
Hello, 

I am getting the error: **"amcli.local.docker.manager.DockerImagePullFailedException: Could not find amazon/aws-sam-cli-emulation-image-java8:rapid-1.15.0 image locally and failed to pull it from docker"**, In fact the tag **"rapid-1.15.0"** doesn't exist in Docker Hub. How can I fix that?

Thanks in advance!!

mildaniel:
Closing as this was resolved by #3092.

PR #3092, opened 2021-07-21
feat: Remove old images when building new one

torresxb1:
https://github.com/aws/aws-sam-cli/issues/2250

When the SAM CLI version is bumped, a new rapid image (with the new version in the tag) is created on the next build. The old versions just sit unused, taking up space. 

When building a new rapid image, this removes old rapid images for the repo. 

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/local/docker/lambda_image.py
```diff
@@ -287,3 +289,18 @@ def _generate_dockerfile(base_image, layers):
...
+        Parameters
+        ----------
+        repo_name str
+            Repo for which all images will be removed
+        """
+        try:
+            for image in self.docker_client.images.list(name=repo_name):
+                self.docker_client.images.remove(image.id)
+        except docker.errors.APIError:
+            LOG.exception("Failed to remove images")
```
torresxb1, line 306: I wasn't sure if I should raise a user exception here, because I don't think the building of the new image should fail if deleting old images fail. So this would just add a log message, and the old images would be deleted (or at least tried to be deleted again) the next time a new image is built.

samcli/local/docker/lambda_image.py
```diff
@@ -287,3 +289,18 @@ def _generate_dockerfile(base_image, layers):
...
+        Remove all images for given repo
+
+        Parameters
+        ----------
+        repo_name str
+            Repo for which all images will be removed
+        """
+        try:
+            for image in self.docker_client.images.list(name=repo_name):
+                self.docker_client.images.remove(image.id)
```
torresxb1, line 304: This fails if there are other tags referencing this image (if there are multiple tags with the same image id). Should I use the `force` flag or should we never have multiple tags referencing the same image?
https://docs.docker.com/engine/reference/commandline/rmi/#examples

mgrandis (changes requested):
Added a few comments.
You may also want to add at least one integration test, updates to the Docker client/client library can easily break us.

tests/unit/local/docker/test_lambda_image.py
```diff
@@ -400,3 +404,21 @@ def test_build_image_fails_with_ApiError(
...
+        setattr(layer_downloader_mock, "layer_cache", self.layer_cache_dir)
+
+        lambda_image = LambdaImage(layer_downloader_mock, False, False, docker_client=docker_client_mock)
+
+        self.assertEqual(
+            lambda_image.build("python3.6", ZIP, None, []),
+            f"amazon/aws-sam-cli-emulation-image-python3.6:rapid-{version}",
+        )
+
+        docker_client_mock.images.remove.assert_has_calls([call("old1"), call("old2")])
```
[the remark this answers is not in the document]
torresxb1: For your first point, I believe this is the only test where the mock would have old images. For the other tests that went into this path, I added a mock return of an empty list when checking for old images. Do you want me to add old images to those and assert this? Or keep those as is and just add the assert not called for the tests that don't go into that path?

samcli/local/docker/lambda_image.py
```diff
@@ -287,3 +292,22 @@ def _generate_dockerfile(base_image, layers):
...
+        Parameters
+        ----------
+        repo str
+            Repo for which rapid images will be removed
+        """
+        LOG.info("Removing rapid images for repo %s", repo)
+        try:
+            for image in self.docker_client.images.list(name=repo):
+                for tag in image.tags:
+                    if ":rapid-" in tag:
```
torresxb1, line 309: not sure if this is enough or if I should do some regex like https://github.com/aws/aws-sam-cli/blob/a77d9e09e3739be6d802efa1394f2c8fe30ef13d/samcli/local/docker/manager.py#L200

mgrandis: I guess we could maybe take this functionality out in its own function, in `samcli/local/docker/utils.py` I guess

samcli/local/docker/lambda_image.py
```diff
@@ -288,3 +294,22 @@ def _generate_dockerfile(base_image, layers):
...
+        ----------
+        repo st
+            Repo for which rapid images will be removed
+        """
+        LOG.info("Removing rapid images for repo %s", repo)
+        try:
+            for image in self.docker_client.images.list(name=repo):
+                for tag in image.tags:
+                    if ContainerManager.is_rapid_image(tag):
+                        self.docker_client.images.remove(image=image.id, force=True)
```
wchengru, line 312: Is force=True safe here?

tests/integration/local/invoke/test_integration_cli_images.py
```diff
@@ -399,3 +399,88 @@ def test_invoke_with_error_during_image_build(self):
...
+
+    def tearDown(self):
+        for tag in self.old_rapid_image_tags + [self.new_rapid_image_tag]:
+            try:
+                self.client.api.remove_image(tag)
+            except APIError:
+                pass
+
+    @pytest.mark.flaky(reruns=3)
+    def test_building_new_rapid_image_removes_old_rapid_images(self):
```
wchengru, line 444: Suggest to test a case that it doesn't remove the `:rapid-` tagged images from the other repos.

wchengru (approved):

samcli/local/docker/lambda_image.py
torresxb1: I think it might not be necessary anymore. I'll remove it. @mgrandis let me know if you think this is still needed

tests/integration/local/invoke/test_integration_cli_images.py
torresxb1: Thanks for the suggestion, I'll add an integration test for this

mgrandis (approved):

samcli/local/docker/lambda_image.py
```diff
@@ -114,7 +115,8 @@ def build(self, runtime, packagetype, image, layers, stream=None):
         # Default image tag to be the base image with a tag of 'rapid' instead of latest.
         # If the image name had a digest, removing the @ so that a valid image name can be constructed
         # to use for the local invoke image name.
-        image_tag = f"{image_name.split(':')[0].replace('@', '')}:rapid-{version}"
+        image_repo = image_name.split(":")[0].replace("@", "")
+        image_tag = f"{image_repo}:rapid-{version}"
```
sriram-mv, line 119: More a nit, than anything. Lets not refer to "rapid" in static string anywhere directly, what do you think about making it a constant that we can refer to in f strings?

sriram-mv (changes requested):

samcli/local/docker/lambda_image.py
sriram-mv: Other than that, looks good!

samcli/local/docker/lambda_image.py
```diff
@@ -308,11 +310,24 @@ def _remove_rapid_images(self, repo: str) -> None:
...
                         try:
                             self.docker_client.images.remove(image.id)
                         except docker.errors.APIError as ex:
                             LOG.warning("Failed to remove rapid image with ID: %s", image.id, exc_info=ex)
                         break
         except docker.errors.APIError as ex:
             LOG.warning("Failed getting images from repo %s", repo, exc_info=ex)
+
+    @staticmethod
+    def is_rapid_image(image_name: str) -> bool:
```
torresxb1, line 323: had to move this out of ContainerManager and into here because the linter complained that there was a cyclical import if I had the constant on this file and this method in ContainerManager (since container manager used the constant but this LambdaImage used the method). The options were moving the constant to ContainerManager or moving the method here, so I thought the latter was better. I think the method might make more sense here anyway.

sriram-mv (approved):

PR #3109, opened 2021-07-22
feat(SAM Check): Bottle Neck creation

jordon-ogo:
#### Checklist

- [x] Template is parsed for lambda functions
- [x] Lambda function objects are created
- [x] Lambda functions can act as entry points
- [x] Bottle neck questions are asked for lambda functions and data is stored, but no calculations are done yet

#### TODO
- [x] type-hinting
- [x] Use @property


samcli/commands/check/graph_context.py
```diff
@@ -0,0 +1,22 @@
...
+        self.lambda_functions = lambda_functions
+
+    def generate(self):
+        graph = Graph()
+
+        # Find all entry points
+        for function in self.lambda_functions:
+            if function.get_parents() == []:  # No parent resourecs, so this is an entry point
+                graph.add_entry_point(function)
+            else:
```
CoshUS, line 19: Are we planning to add something else here in the future?
If not, let's remove the else and pass.

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,73 @@
+"""
+Bottle neck questions are asked here. Data is saved in graph, but not calcualted here.
+"""
+
+import click
+
+
+class BottleNecks:
+    def __init__(self, graph):
+        self.graph = graph
```
CoshUS, line 10: Does `self.graph` need to be public member?
If not, let's make it protected by changing its name to `self._graph`

One underscore > protected
Two underscore > private

samcli/commands/check/graph_context.py
```diff
@@ -0,0 +1,22 @@
+"""
+This class will generate the graph based on the template data.
+Connections between nodes will occur here.
+"""
+from .resources.Graph import Graph
+
+
+class GraphContext:
+    def __init__(self, lambda_functions):
+        self.lambda_functions = lambda_functions
```
CoshUS, line 10: Same comment as `self.graph`.

samcli/commands/check/graph_context.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def generate(self):
+        graph = Graph()
+
+        # Find all entry points
+        for function in self.lambda_functions:
+            if function.get_parents() == []:  # No parent resourecs, so this is an entry point
```
CoshUS, line 17: `if not function.get_parents():` is more pythonic.

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,19 @@
+"""
+Class for graph. All data is stored in the graph directly, or within nodes that are stored in the graph
+"""
+class Graph:
+    def __init__(self):
+        self.entry_points = []
```
CoshUS, line 6: Same comment as `self.graph`. Protected and/or use properties.

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,19 @@
...
+        self.resources_to_analyze = []
+
+    def add_entry_point(self, node):
+        self.entry_points.append(node)
+
+    def get_entry_points(self):
```
CoshUS, line 12: Use `@property` instead of getters.

samcli/commands/check/resources/LambdaFunction.py
```diff
@@ -0,0 +1,66 @@
+"""
+Class object for Lambda Functions. Contains object from template, as well as all data for lambda functions,
+excluding pricing info
+"""
+from samcli.commands.check.resources.TemplateResource import TemplateResource
+
+
+class LambdaFunction(TemplateResource):
+    def __init__(self, resource_object, resource_type):
```
CoshUS, line 9: Do we need getters and setters here since everything is public?
Otherwise, directly referencing the members themselves is a lot cleaner.

samcli/commands/check/resources/TemplateResource.py
```diff
@@ -0,0 +1,16 @@
+"""
+Super class for all resources
+"""
+class TemplateResource:
+    def __init__(self, resource_object, resource_type):
```
CoshUS, line 5: Same as above.

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,73 @@
...
+
+            self.graph.add_resource_to_analyze(current_entry_point)
+
+        click.echo("Running calculations...")
+
+    def ask_bottle_neck_questions(self, resource):
+        if resource is None:
+            return
+
+        if resource.get_resource_type() == "AWS::Lambda::Function":
```
CoshUS: Reference `samcli\commands\_utils\resources.py::AWS_LAMBDA_FUNCTION` instead of hardcoding it here.

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,73 @@
...
+        user_input_tps = ask(
+            "What is the expected per-second arrival rate for [%s]?\n[TPS]" % (lambda_function.get_name())
+        )
+        lambda_function.set_tps(user_input_tps)
+
+    user_input_duration = ask(
+        "What is the expected duration for the Lambda function [%s] in ms?\n[1 - 900,000]"
+        % (lambda_function.get_name()),
+        1,
+        900000,
```
CoshUS, line 55: Let's move this into a constant instead of hardcoding it here.
We can also replace the 900,000 in the prompt with this constant.

samcli/commands/check/lib/command_context.py
```diff
@@ -85,3 +96,23 @@ def read_sam_file(self):
             sam_template = yaml_parse(sam_template.read())
 
         return sam_template
+
+
+def parse_template():
+    all_lambda_functions = []
+
+    # template path
+    path = os.path.realpath("template.yaml")
```
CoshUS, line 105: Why is `template.yaml` hardcoded here instead of getting it from a CLI option?

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,19 @@
+"""
```
CoshUS, line 1: General comment for Graph/LambdaFunction/TemplateResource classes:
Are these better fitted for `lib` folder instead `commands` folder?

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,73 @@
...
+
+class BottleNecks:
+    def __init__(self, graph):
+        self.graph = graph
+
+    def ask_entry_point_question(self):
+        entry_points = self.graph.get_entry_points()
+
+        # All entry points must be calcualted before info can be displayed
+        while entry_points != []:
```
CoshUS, line 16: Use `while not entry_points:` instead of `while entry_points != []:` is more pythonic.

CoshUS (changes requested):
Added some comments. First round of review.

samcli/commands/check/graph_context.py
jordon-ogo: I planned to add something here, but now it is no longer necessary.

samcli/commands/check/resources/LambdaFunction.py
jordon-ogo: I will be making things private, so they will be necessary

samcli/commands/check/lib/command_context.py
jordon-ogo: For now the template file needs to be located in the current directory. Haven't implemented the CLI option yet.

samcli/commands/check/resources/Graph.py
jordon-ogo: I kept them in the command (check/Resources) folder because they are designed for the check command. I'm not sure if they can/will be used for stuff other than `sam check`, given that the parameters seem to be unique to the `check` functionality.

samcli/commands/check/bottle_necks.py
jordon-ogo: Shouldn't this be `while entry_points:`, since I want it to run while there are elements in the list?

samcli/commands/check/resources/LambdaFunction.py
```diff
@@ -0,0 +1,68 @@
...
+        self.duration = -1
+        self.tps = -1
+        self.parents: List = []
+        self.children: List = []
+        self.number_of_requests = -1
+        self.average_duration = -1
+        self.allocated_memory = -1
+        self.allocated_memory_unit = ""
+
+    def set_number_of_requests(self, num: int):
```
hawflau, line 22: As discussed, please update to use `@property`

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,72 @@
...
+Bottle neck questions are asked here. Data is saved in graph, but not calcualted here.
+"""
+import click
+
+from samcli.commands.check.resources.LambdaFunction import LambdaFunction
+from samcli.commands._utils.resources import AWS_LAMBDA_FUNCTION
+
+
+class BottleNecks:
+    def __init__(self, graph):
```
hawflau, line 11: Please add the type annotation for `graph`

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,72 @@
...
+    def ask_entry_point_question(self):
+        entry_points = self._graph.get_entry_points()
+
+        # All entry points must be calcualted before info can be displayed
+        while entry_points:
+            entry_point_question = (
+                "We found the following resources in your application that could be the entry point for a request."
+            )
+            item_number = 1
+            for item in entry_points:
```
hawflau, line 24: you can use `enumerate` to get both `item_number` and `item` instead of declaring `item_number` and incrementing it every loop
Example:
```
for item_number, item in enumerate(entry_points):
    ...
```

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        lambda_function.set_duration(user_input_duration)
+
+    def ask_bottle_neck_questions(self, resource: LambdaFunction):
+        if resource.get_resource_type() == AWS_LAMBDA_FUNCTION:
+            self.lambda_bottle_neck_quesitons(resource)
+
+
+def ask(question: str, min_val=1, max_val=float("inf")) -> int:
```
hawflau, line 62: I think you can also annotate the types for `min_val` and `max_val`

jordon-ogo: I thought about that, but I wasn't sure how to annotate them, given that they have default values

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,72 @@
...
+class BottleNecks:
+    def __init__(self, graph):
+        self._graph = graph
+        self._lambda_max_duration = 900000
+
+    def ask_entry_point_question(self):
+        entry_points = self._graph.get_entry_points()
+
+        # All entry points must be calcualted before info can be displayed
+        while entry_points:
```
hawflau, line 19: Looking at this `while` block, it seems user has to answer all questions, regardless of the order of questions. Two questions:
1. Does the order of questions answer matter the end result?
2. Will there be any way for user to skip answer some questions?

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,72 @@
...
+
+            current_entry_point = entry_points.pop(user_input - 1)
+
+            self.ask_bottle_neck_questions(current_entry_point)
+
+            self._graph.add_resource_to_analyze(current_entry_point)
+
+        click.echo("Running calculations...")
+
+    def lambda_bottle_neck_quesitons(self, lambda_function: LambdaFunction):
```
hawflau, line 40: nit: if the method is only used internally, consider prefixing it with `_` to make it a protected attribute

samcli/commands/check/bottle_necks.py
jordon-ogo: 1. Yes, the order does matter. Right now it technically doesn't matter, but once more resources are added, it will matter.
2. No. Calculations (both bottle neck and pricing) cannot be done unless all questions are answered.

samcli/commands/check/lib/command_context.py
```diff
@@ -68,7 +81,7 @@ def transform_template(self):
 
         return converted_template
 
-    def read_sam_file(self):
+    def read_sam_file(self) -> Any:
```
hawflau, line 84: should the return type be `Dict` instead of `Any`? (assuming `yaml_parse` is returning a dict object)

samcli/commands/check/lib/command_context.py
hawflau: Can you add a `TODO` note for this future implementation?

samcli/commands/check/lib/command_context.py
jordon-ogo: I'll look into it again, but for some reason pylint thinks that it returns an `Any`. Putting anything else there creates a pylint error. For most things that return `Any`, I get this problem.

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Graph:
+    def __init__(self):
+        self._entry_points = []
+        self._resources_to_analyze = []
+
+    def add_entry_point(self, node: LambdaFunction):
+        self._entry_points.append(node)
+
+    def get_entry_points(self) -> Any:
```
hawflau, line 16: Is there a specific type to use instead of `Any`?

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,23 @@
...
+        return self._entry_points
+
+    def get_resources_to_analyze(self) -> Any:
```
hawflau, line 19: same as above comment

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -74,13 +70,29 @@ def is_valid(self):
                 functools.reduce(lambda message, error: message + " " + str(error), e.causes, str(e))
             ) from e
 
+    def _replace_local_codeuri(self):
+        return external_replace_local_codeuri(self.sam_template)
+
+    @staticmethod
+    def is_s3_uri(uri):
```
hawflau, line 77: why defining this static method instead of just using the one that you imported?

jordon-ogo: I have experimented a lot with changing the layout of sam_template_validator. A large amount of functional tests for validate require sam_template_validator to have this exact layout, otherwise they fail. With this format, all types of tests for validate pass. Changing the layout will require many tests to be rewritten.

samcli/commands/check/bottle_necks.py
jordon-ogo: I need `item_number` outside of the `for item in entry_points:` loop, so I can't do that.

samcli/commands/check/resources/Graph.py
jordon-ogo: Same as above. pylint thinks it returns an `Any`, so I need to use that to prevent make-pr from failing. Still looking into this.

samcli/commands/check/bottle_necks.py
CoshUS: Yep, should be `while entry_points:`.

samcli/commands/check/resources/Graph.py
CoshUS: I don't have strong feelings for this. These classes seem to be purely model and don't relate to CLI.
We can keep them here and, if needed in the future, moving them should be relatively easy.

samcli/commands/check/graph_context.py
```diff
@@ -0,0 +1,21 @@
+"""
+This class will generate the graph based on the template data.
+Connections between nodes will occur here.
+"""
+from typing import List
+from .resources.Graph import Graph
+
+
+class GraphContext:
```
CoshUS, line 9: nit: This seems to be better called as GraphFactory or GraphGenerator. Context is already being used for referring to CLI context, such as BuildContext, DeployContext, etc.
Also, can we simply put the generate logic inside of the Graph constructor and eliminate this class entirely?

jordon-ogo: I will put the graph generator logic inside the graph constructor. I agree that is a better place for it.

samcli/commands/check/bottle_necks.py
hawflau: the last updated `item_number` will be available after the `for` loop even if `enumerate` is used.
For example, 
```
for item_number, i in enumerate(range(10)):
    print(item_number)

print(item_number) # 9
```

samcli/commands/check/resources/LambdaFunction.py
```diff
@@ -0,0 +1,84 @@
...
+    def allocated_memory_unit(self, unit: str):
+        self._allocated_memory_unit = unit
+
+    @property
+    def children(self) -> List:
+        return self._children
+
+    @children.setter
+    def children(self, child_node: "LambdaFunction"):
+        self._children.append(child_node)
```
hawflau, line 60: Consider this example:
```
lamdba_function.children = child_a
lambda_function.children = child_b
print(lambda_function.children) # [child_a, child_b]
```

It's a bit not natural to me to use a setter like this. Because `.children` is already returning a list, I would expect I would just use `lambda_function.children.append(child_a)` to add a new child. And the setter is used overwrite `self._children` with another list of children.

jordon-ogo: I changed `def parent` as well to this style

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,75 @@
...
+import click
+
+from samcli.commands.check.resources.Graph import Graph
+from samcli.commands.check.resources.LambdaFunction import LambdaFunction
+from samcli.commands._utils.resources import AWS_LAMBDA_FUNCTION
+
+
+class BottleNecks:
+    def __init__(self, graph: Graph):
+        self._graph = graph
```
CoshUS, line 13: Nice job on adding type hintings!
Can we also add type hintings for class members?
```
class BottleNecks:
    _graph: Graph                     # type hinting for self._graph
    _lambda_max_duration: int         # type hinting for self._lambda_max_duration
    def __init__(self, graph: Graph):
        self._graph = graph
```
Same goes for other classes.

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,75 @@
...
+        self._lambda_max_duration = 900000
+
+    def ask_entry_point_question(self):
```
CoshUS, line 16: Let's add return types for functions as well even if they are None. Functions without annotations will default to `Any` return type.
```
def ask_entry_point_question(self) -> None:
```
https://www.python.org/dev/peps/pep-0484/#the-meaning-of-annotations
> It is recommended but not required that checked functions have annotations for all arguments and the return type. For a checked function, the default annotation for arguments and for the return type is Any.

samcli/commands/check/resources/LambdaFunction.py
```diff
@@ -0,0 +1,84 @@
+"""
+Class object for Lambda Functions. Contains object from template, as well as all data for lambda functions,
+excluding pricing info
+"""
+from typing import List
+from samcli.lib.providers.provider import Function
+from samcli.commands.check.resources.TemplateResource import TemplateResource
+
+
+class LambdaFunction(TemplateResource):
```
CoshUS, line 10: It seems a lot of the properties are directly mapping to class members with both get and set.
 
Going back to the previous comment: https://github.com/aws/aws-sam-cli/pull/3109#discussion_r675202180
If the members are expected to be set and get without checks, we can leave them as public members and switch to properties when needed in the future. https://stackoverflow.com/a/6618184/5284278

samcli/commands/validate/lib/sam_template_validator.py
CoshUS: This is doing it backwards. Code should not be conforming to tests.
We shouldn't be adding random pieces of code just to avoid changes to tests.
If the class architectural changes, we need to reflect that in the tests.

samcli/commands/check/resources/TemplateResource.py
```diff
@@ -0,0 +1,22 @@
+"""
+Super class for all resources
+"""
+from samcli.lib.providers.provider import Function
+
+
+class TemplateResource:
+    def __init__(self, resource_object: Function, resource_type: str):
```
CoshUS, line 8: `resource_object` should not be type of Function.
Adding more child classes of TemplateResource will break this part if they are not Function related.

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+        lambda_function.duration = user_input_duration
+
+    def ask_bottle_neck_questions(self, resource: LambdaFunction):
+        if resource.resource_type == AWS_LAMBDA_FUNCTION:
+            self._lambda_bottle_neck_quesitons(resource)
+
+
+def ask(question: str, min_val: int = 1, max_val: float = float("inf")) -> int:
```
CoshUS, line 65: Let's add some docs for constructors and functions as well.
Use this extension to auto generate snippets: https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

samcli/commands/check/resources/Graph.py
```diff
@@ -0,0 +1,33 @@
...
+    def entry_points(self, node: LambdaFunction):
+        self._entry_points.append(node)
+
+    @property
+    def resources_to_analyze(self) -> List:
+        return self._resources_to_analyze
+
+    @resources_to_analyze.setter
+    def resources_to_analyze(self, resource: LambdaFunction):
+        self._resources_to_analyze.append(resource)
```
CoshUS, line 27: Same issue Wing brought up. Setter shouldn't be doing append.

CoshUS (changes requested):
Second round. Added a few more feedbacks.
Let me know if you have any question.

samcli/commands/validate/lib/sam_template_validator.py
CoshUS: Update: This will be addressed in a separate PR.

samcli/commands/check/resources/TemplateResource.py
jordon-ogo: Right now it can only be of type `Function`. As I add more resources, I will add more options to what type it can be.

CoshUS: Parent classes shouldn't know what child classes need.
For example, if we add a new Layer resource extending TemplateResource, we have to update TemplateResource to:
```
class TemplateResource:
    def __init__(self, resource_object: Union[Function, Layer], resource_type: str):
```
Adding new child classes should not change parent classes.

Instead, we can either make resource_object as Any and let child classes handle checking for specific types, or remove it completely and let child classes create their own resource_object class members.

samcli/commands/check/resources/TemplateResource.py
```diff
@@ -0,0 +1,31 @@
...
+        """
+        Args:
+            resource_object (Function): The resource object form the template file
+            resource_type (str): The resource type
+        """
+        self._resource_object = resource_object
+        self._resource_type = resource_type
+
+    @property
+    def resource_object(self) -> Function:
```
hawflau, line 22: The return type should match the type defined for `self._resource_object` (which is `Any`). Did `mypy` pass?

jordon-ogo: Odd. I ran mypy but never got an error about that. Still, it needs to be changed

samcli/commands/check/bottle_necks.py
```diff
@@ -0,0 +1,106 @@
...
+            entry_point_question = (
+                "We found the following resources in your application that could be the entry point for a request."
+            )
+
+            item_number = 0
+
+            for item_number, item in enumerate(entry_points):
+                item_name = item.resource_name
+                entry_point_question += "\n[%i] %s" % (item_number + 1, item_name)
+                item_number += 1
```
hawflau, line 40: Any reason to increment `item_number` at the end of the loop? `enumerate` does it for you already.

samcli/commands/check/resources/TemplateResource.py
hawflau: oh, probably `self._resource_object` is annotated as `Function` on L9. 
But the input `reousrce_object` in `__init__` is `Any`

samcli/commands/check/resources/TemplateResource.py
```diff
@@ -0,0 +1,30 @@
...
+        self._resource_object = resource_object
+        self._resource_type = resource_type
+
+    @property
+    def resource_object(self) -> Any:
+        return self._resource_object
+
+    @property
+    def resource_name(self) -> Any:
+        return self._resource_object.name
```
CoshUS, line 26: Since `_resource_object` is now `Any`, we don't know whether it has the attribute of `name` or not.
Getting the `name` from `Function` object should now be moved to `LambdaFunction` class.

```
class LambdaFunction(TemplateResource):
    ...
    @property
    def resource_name(self) -> str:
        return cast(Function, self._resource_object).name
```

If we still want to keep the contract that all `TemplateResource` instances should have `resource_name` property, then we can use `@abstractmethod`:

```
class TemplateResource:
     ...
    @property 
    @abstractmethod
    def resource_name(self) -> str:
        raise NotImplementedError()
```

CoshUS (changes requested):
One last comment. Everything looks good.

samcli/commands/check/resources/TemplateResource.py
jordon-ogo: All `TemplateResources` should/will have this property. I'll make it an abstract method

CoshUS (approved):
LGTM!

hawflau (approved):

PR #3111, opened 2021-07-23
Cdk support dev integ test deploy

hawflau:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/iac/interface.py
+++ b/samcli/lib/iac/interface.py
@@ -741,7 +741,7 @@ class Project:
 
     @property
     def default_stack(self) -> Optional[Stack]:
-        if len(self.stacks) > 0:
+        if len(self._stacks) > 0:
             return self._stacks[0]
         return None
 
--- a/samcli/lib/iac/interface.py
+++ b/samcli/lib/iac/interface.py
@@ -740,6 +740,12 @@ class Project:
         self._stacks = stacks
 
     @property
+    def default_stack(self) -> Optional[Stack]:
+        if len(self.stacks) > 0:
+            return self._stacks[0]
+        return None
+
+    @property
     def extra_details(self) -> Optional[Dict[str, Any]]:
         return self._extra_details
 
--- a/tests/integration/deploy/test_deploy_command_cdk.py
+++ b/tests/integration/deploy/test_deploy_command_cdk.py
@@ -1,0 +1,742 @@
+import json
+import os
+import shutil
+import tempfile
+import uuid
+import time
+import copy
+import logging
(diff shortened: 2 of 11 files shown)


wchengru (approved):

tests/integration/deploy/test_deploy_command_cdk.py
```diff
@@ -0,0 +1,742 @@
...
+            if stack_name != SAM_CLI_STACK_NAME:
+                self.cf_client.delete_stack(StackName=stack_name)
+        super().tearDown()
+
+    def _method_to_stack_name(self, method_name):
+        """Method expects method name which can be a full path. Eg: test.integration.test_deploy_command.method_name"""
+        method_name = method_name.split(".")[-1]
+        return f"{method_name.replace('_', '-')}-{CFN_PYTHON_VERSION_SUFFIX}"
+
+    def _replace_cloud_assembly_stack_name(self, manifest_path, original_stack_name, new_stack_name):
```
mildaniel, line 69: Question: What's the purpose of this function/is it used in these tests?

samcli/lib/iac/interface.py
```diff
@@ -739,6 +739,12 @@ def stacks(self) -> List[Stack]:
     def stacks(self, stacks: List[Stack]) -> None:
         self._stacks = stacks
 
+    @property
+    def default_stack(self) -> Optional[Stack]:
+        if len(self.stacks) > 0:
```
mildaniel, line 744: Is there a reason for checking the length of `self.stacks` but returning a stack from `self._stacks`?

tests/integration/deploy/test_deploy_command_cdk.py
hawflau: sorry, it's not needed anymore. Will remove it.

samcli/lib/iac/interface.py
hawflau: It's to ensure `self.stacks` contains at least one stack. Otherwise, it returns `None`.

mildaniel: I just meant why the check is for the public member `self.stacks` but the return is a value of the protected member `self._stacks`. Just for my understanding.

hawflau: ah got it. They are the same but I agree that it should be either both internal or both public. I'll update it

hawflau: updated

mildaniel (approved):

PR #3119, opened 2021-07-27
chore: upgrade pylint to 2.9.0

wchengru:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


.pylintrc
```diff
@@ -92,6 +93,10 @@ disable=
     # W9018, # "%s" differing in parameter type documentation
     # W9019, # "%s" useless ignored parameter documentation
     # W9020, # "%s" useless ignored parameter type documentation
+    C0103, # Class constant name "%s" doesn't conform to UPPER_CASE naming style ('([^\\W\\da-z][^\\Wa-z]*|__.*__)$' pattern) (invalid-name)
```
CoshUS, line 96: Can we get some short justification in the PR description for disabling these?

aahung (approved):
Went through the change with @wchengru, approved.

.pylintrc
wchengru: Thanks! Added.

CoshUS (approved):

samcli/lib/deploy/deployer.py
```diff
@@ -113,6 +113,7 @@ def has_stack(self, stack_name):
             if "Stack with id {0} does not exist".format(stack_name) in str(e):
                 LOG.debug("Stack with id %s does not exist", stack_name)
                 return False
+            return None
```
mndeveci, line 116: Why are we returning `None` for a function which returns boolean? Can we return `True` or `False`?

aahung: We did this so that this PR doesn't change any behavior. I believe the intention of the original author was to raise the exception at line 116 but we didn't confirm that

mndeveci: So if you don't return anything from a function, it will return `None`?

mndeveci (approved):

PR #3121, opened 2021-07-27
sam check calculations

jordon-ogo:
#### Checklist

- [x] Calculations are now done on lambda functions
- [x] Results of calculations are printed based on severity

### TODO

- [x] Add unit tests
- [x] Resolve pylint errors


samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        click.echo("Running calculations...")
+
+        for resource in self._graph.resources_to_analyze:
+            resource_type = resource.resource_type
+            resource_name = resource.resource_name
+
+            if resource_type == AWS_LAMBDA_FUNCTION:
+
+                client = boto3.client("service-quotas")
```
hawflau, line 87: nit: you can create the client once and reuse it over the loop.

samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,115 @@
...
+                burst_concurrency = client.get_aws_default_service_quota(ServiceCode="lambda", QuotaCode="L-548AE339")[
```
hawflau, line 88: please consider handling any potential exception threw by boto3 client calls.

samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,115 @@
...
+    """The check to see if a given resource will cause a bottle neck.
+    Args:
+        tps (int): TPS of lambda function based on parent resource
+        duration (int): Duration of lambda function
+        execution_limit (int): Execution limit based on user account
+
+    Returns:
+        float: percentage of available concurrent executions used
+    """
+    tps_max_limit = (1000 / duration) * execution_limit
```
hawflau, line 114: any precautionary measure to make sure duration is not zero?

jordon-ogo: When the user is asked to input duration, they must enter a value within the range 1-900,000 inclusive. Therefore duration can never be zero.

samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,115 @@
...
+            capacity_used (int): Percentage of capacity used
+            resource_name (str): Recource name that the warning is for
+            concurrent_executions (int): Concurrent executions based on users account
+            duration (int): Duration of lambda function
+            tps (int): TPS of lambda function based on its parent resource
+            burst_concurrency (int): Burst concurrency based on users account
+        """
+        warning = CheckWarning()
+
+        if capacity_used <= 70:
```
hawflau, line 41: sorry for being OCD, but I'm curious to understand why the boundaries are set as:
```
capacity_used <= 70
70 < capacity_used < 90
90 <= capacity_used <= 100
capacity_used > 100
```
rather than:
```
capacity_used < 70
70 <= capacity_used < 90
90 <= capacity_used < 100
capacity_used >= 100
```
or 
```
capacity_used <= 70
70 < capacity_used <= 90
90 < capacity_used <= 100
capacity_used > 100
```

Nit: you can omit the lower boundaries in the `if...elif...else` statements. For example,
```
if capacity_used < 70:
    ...
elif capacity_used < 90:
    ...
elif capacity_used < 100:
    ...
else:
    ...
```

jordon-ogo: 1. Stefano and I discussed it and these are the boundaries he wanted. I can double check with him if he still wants these.
2. There is no `elif capacity_used > 100` in the code if that's what you're referring to. There is a comment stating what the `else` is. I like to see what the else is for without having to look through the entire `if else` statement.

hawflau: 1. I'm good with the numbers of the boundaries (i.e. 70, 90, 100). I think my question is - why are the comparison signs not consistent in all cases? (i.e. either `<` on the right and `<=` on the left, or `<=` on the right and `<` on the left)
2. It's fine to use just `else` to represent the case of `capacity_used > 100`, assuming `capacity_used` is a number. And the comment is good too 👍 and I think you should still keep it. My example for omitting the lower boundary does the exact same thing as yours, but I personally think my example looks cleaner.

jordon-ogo: 1. Stefano defined where each warning should be placed depending on the percentage of concurrency used. Making the signs consistent in either of your examples would not match what Stefano wants. 
2. I understand what you're saying now. I will change it.

CoshUS (changes requested):
First round of review. Focused on styling and OOP.

samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,125 @@
...
+                "For the lambda function [%s], the %ims duration and %iTPS arrival rate is using %i%% of the allowed "
+                + "concurrency on AWS Lambda. A limit increase should be considered:"
+                + "\nhttps://console.aws.amazon.com/servicequotas"
+            ) % (resource_name, duration, tps, round(capacity_used))
+            self._graph.yellow_warnings.append(warning)
+
+        elif capacity_used <= 100:
+            warning.message = (
+                "For the lambda function [%s], the %ims duration and %iTPS arrival rate is using %i%% of the allowed "
+                + "concurrency on AWS Lambda. It is very close to the limits of the lambda function. It is strongly "
```
hawflau, line 62: The `+` sign is not necessary here. Consecutive lines of strings will just concatenate into one when they are in the same `()`. For example,
```
s = (
    "hello, "
    "world!"
)
print(s) # 'hello, world1'
```

samcli/commands/check/bottle_necks.py
```diff
@@ -39,15 +39,16 @@ def ask_entry_point_question(self) -> None:
                 entry_point_question += "\n[%i] %s" % (item_number + 1, item_name)
 
             entry_point_question += "\nWhere should the simulation start?"
-            user_input = ask(entry_point_question, 1, item_number + 1)
+
```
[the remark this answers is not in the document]
jordon-ogo: I assume you're referring to the `\n` on line 41. This is the first question the user is asked in a chain of questions. To help illustrate the importance of the question and that the current "chain" of questions they were asked is now over. The user must now select a new entry point. That's the only reason. Other than printing results, if these extra spaces (new lines) are not expected in cli programs, I can remove it.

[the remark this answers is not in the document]
CoshUS: I'm referring to the code. All the statements are double spaced for this block. 
```
            entry_point_question += "\nWhere should the simulation start?"

            user_input = _ask(entry_point_question, 1, item_number + 1)

            current_entry_point = entry_points.pop(user_input - 1)

            self._ask_bottle_neck_questions(current_entry_point)

            self._graph.resources_to_analyze.append(current_entry_point)

            click.echo("")
```

samcli/commands/check/print_results.py
```diff
@@ -0,0 +1,40 @@
+"""
+Prints the results of bottle neck calculations
+"""
+from typing import List
+import click
+
+from samcli.commands.check.resources.Graph import Graph
+
+
+class Results:
```
CoshUS, line 10: File name also needs to be changed.

samcli/commands/check/calculations.py
```diff
@@ -0,0 +1,126 @@
+"""
```
CoshUS, line 1: File name needs to be changed to match the class name.

samcli/commands/check/bottle_necks.py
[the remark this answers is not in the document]
jordon-ogo: No specific reason. I will reduce the double spaces

hawflau (approved):
LGTM!

samcli/commands/check/resources/Warning.py
```diff
@@ -0,0 +1,12 @@
+"""
+Class object for warnings. Warning messages, as well as the warning type is stored here
+"""
+
+
+class CheckWarning:
+    warning_type: str
+    message: str
+
+    def __init__(self):
```
CoshUS, line 10: Does it make sense to create a CheckWarning object without warning_type or message?
If not, the constructor shouldn't be empty and we should set the class members with parameters taken in from the constructor.

jordon-ogo: `warning_type` was removed, as it is no longer needed. A message must now be passed when `CheckWarning` object is created

CoshUS (approved):
LGTM!

PR #3122, opened 2021-07-27
Sam delete bug fixes

hnnasit:
Bugs fixed and new changes:
- No stack-name present when no-prompts is provided
- No region present when no-prompts is provided
- Retaining everything when a stack has termination protection enabled
- ecr_uploader.delete_artifact tries to delete an image from a repository that is already deleted (In the case of deleting ECR Companion Stack)
- Double checking if s3_bucket and key are strings in packageable_resources
- More integration tests for guided and non-guided delete

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/delete/delete_context.py
+++ b/samcli/commands/delete/delete_context.py
@@ -64,7 +64,7 @@ class DeleteContext:
                 )
             else:
                 raise click.BadOptionUsage(
-                    option_name="delete",
+                    option_name="--stack-name",
                     message="Missing option '--stack-name', provide a stack name that needs to be deleted.",
                 )
 
@@ -107,7 +107,7 @@ class DeleteContext:
                 self.region = region if region else "us-east-1"
             else:
                 raise click.BadOptionUsage(
-                    option_name="delete",
+                    option_name="--region",
                     message="Missing option '--region', region is required to run the non guided delete command.",
                 )
 
--- a/samcli/commands/delete/delete_context.py
+++ b/samcli/commands/delete/delete_context.py
@@ -35,7 +35,7 @@ LOG = logging.getLogger(__name__)
 
 
 class DeleteContext:
-    # TODO: Separate this context into guided and non-guided just like deploy.
+    # TODO: Separate this context into 2 separate contexts guided and non-guided, just like deploy.
     def __init__(self, stack_name: str, region: str, profile: str, config_file: str, config_env: str, no_prompts: bool):
         self.stack_name = stack_name
         self.region = region
--- a/samcli/commands/delete/delete_context.py
+++ b/samcli/commands/delete/delete_context.py
@@ -60,7 +60,7 @@ class DeleteContext:
             LOG.debug("No stack-name input found")
             if not self.no_prompts:
                 self.stack_name = prompt(
-                    click.style("\tEnter stack name you want to delete:", bold=True), type=click.STRING
+                    click.style("\tEnter stack name you want to delete", bold=True), type=click.STRING
                 )
             else:
                 raise click.BadOptionUsage(
--- a/samcli/commands/delete/delete_context.py
+++ b/samcli/commands/delete/delete_context.py
@@ -35,6 +35,7 @@ LOG = logging.getLogger(__name__)
 
 
 class DeleteContext:
+    # TODO: Separate this context into guided and non-guided just like deploy.
     def __init__(self, stack_name: str, region: str, profile: str, config_file: str, config_env: str, no_prompts: bool):
         self.stack_name = stack_name
         self.region = region
@@ -57,9 +58,15 @@ class DeleteContext:
         self.parse_config_file()
         if not self.stack_name:
             LOG.debug("No stack-name input found")
-            self.stack_name = prompt(
-                click.style("\tEnter stack name you want to delete:", bold=True), type=click.STRING
-            )
+            if not self.no_prompts:
+                self.stack_name = prompt(
+                    click.style("\tEnter stack name you want to delete:", bold=True), type=click.STRING
+                )
+            else:
+                raise click.BadOptionUsage(
+                    option_name="delete",
+                    message="Missing option '--stack-name', provide a stack name that needs to be deleted.",
+                )
 
         self.init_clients()
         return self
@@ -94,9 +101,15 @@ class DeleteContext:
         Initialize all the clients being used by sam delete.
         """
         if not self.region:
-            session = boto3.Session()
-            region = session.region_name
-            self.region = region if region else "us-east-1"
+            if not self.no_prompts:
+                session = boto3.Session()
+                region = session.region_name
+                self.region = region if region else "us-east-1"
+            else:
+                raise click.BadOptionUsage(
+                    option_name="delete",
+                    message="Missing option '--region', region is required to run the non guided delete command.",
+                )
 
         if self.profile:
             Context.get_current_context().profile = self.profile
@@ -219,9 +232,6 @@ class DeleteContext:
 
             retain_repos = self.ecr_repos_prompts(ecr_companion_stack_template)
 
-            # Delete the repos created by ECR companion stack if not retained
-            ecr_companion_stack_template.delete(retain_resources=retain_repos)
-
             click.echo(f"\t- Deleting ECR Companion Stack {self.companion_stack_name}")
             try:
                 # If delete_stack fails and its status changes to DELETE_FAILED, retain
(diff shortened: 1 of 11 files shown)


samcli/commands/delete/delete_context.py
```diff
@@ -219,19 +232,23 @@ def delete_ecr_companion_stack(self):
...
+
+                # Delete the repos created by ECR companion stack if not retained
+                ecr_companion_stack_template.delete(retain_resources=retain_repos)
             except CfDeleteFailedStatusError:
                 LOG.debug("delete_stack resulted failed and so re-try with retain_resources")
+                # Delete the non retained repos created by ECR companion stack before deleting the stack
+                ecr_companion_stack_template.delete(retain_resources=retain_repos)
+
                 self.cf_utils.delete_stack(stack_name=self.companion_stack_name, retain_resources=retain_repos)
+                self.cf_utils.wait_for_delete(stack_name=self.companion_stack_name)
```
moelasmar, line 251: sorry I did not get these lines, i have the following questions:
- what is the difference between the call in line 244, and 248 .. for me we are doing the same thing, so what is the value of repeating it.
- why cf_utils.delete_stack does not wait till the stack is really deleted.

samcli/commands/delete/delete_context.py
```diff
@@ -219,19 +232,23 @@ def delete_ecr_companion_stack(self):
...
             click.echo(f"\t- Deleting ECR Companion Stack {self.companion_stack_name}")
             try:
                 # If delete_stack fails and its status changes to DELETE_FAILED, retain
                 # the user input repositories and delete the stack.
                 self.cf_utils.delete_stack(stack_name=self.companion_stack_name)
                 self.cf_utils.wait_for_delete(stack_name=self.companion_stack_name)
                 LOG.debug("Deleted ECR Companion Stack: %s", self.companion_stack_name)
+
+                # Delete the repos created by ECR companion stack if not retained
+                ecr_companion_stack_template.delete(retain_resources=retain_repos)
```
moelasmar, line 244: shall we try to delete the companion stack ecr repositories first before trying to delete the stack

samcli/commands/delete/delete_context.py
```diff
@@ -279,26 +296,30 @@ def delete(self):
...
+
+            # Delete the artifacts and retain resources user selected not to delete
+            template.delete(retain_resources=retain_resources)
         except CfDeleteFailedStatusError:
             LOG.debug("delete_stack resulted failed and so re-try with retain_resources")
+            # Delete the non retained artifacts and resources before deleting stack
+            template.delete(retain_resources=retain_resources)
+
             self.cf_utils.delete_stack(stack_name=self.stack_name, retain_resources=retain_resources)
+            self.cf_utils.wait_for_delete(self.stack_name)
```
moelasmar, line 314: same comments as ecr companion stack deletion

samcli/lib/package/ecr_uploader.py
```diff
@@ -133,8 +133,10 @@ def delete_artifact(self, image_uri: str, resource_id: str, property_name: str):
 
         except botocore.exceptions.ClientError as ex:
             # Handle Client errors such as RepositoryNotFoundException or InvalidParameterException
-            LOG.error("DeleteArtifactFailedError Exception : %s", str(ex))
-            raise DeleteArtifactFailedError(resource_id=resource_id, property_name=property_name, ex=ex) from ex
+            if "RepositoryNotFoundException" not in str(ex):
```
moelasmar, line 136: what should be done in the else, I think just logging the exception is not enough

samcli/lib/package/packageable_resources.py
```diff
@@ -178,7 +178,7 @@ def get_property_value(self, resource_dict):
             return {"Bucket": None, "Key": None}
 
         resource_path = jmespath.search(self.PROPERTY_NAME, resource_dict)
-        if resource_path:
+        if resource_path and isinstance(resource_path, str):
```
moelasmar, line 181: does this change to cove if the S3 information is added as Ref intrinsic function?

samcli/lib/package/packageable_resources.py
```diff
@@ -350,7 +350,9 @@ def get_property_value(self, resource_dict):
         s3_bucket = resource_path.get(self.BUCKET_NAME_PROPERTY, None)
 
         key = resource_path.get(self.OBJECT_KEY_PROPERTY, None)
-        return {"Bucket": s3_bucket, "Key": key}
+        if isinstance(s3_bucket, str) and isinstance(key, str):
```
moelasmar, line 353: the same comment as line 181

moelasmar (changes requested):
You need to add some integration test cases to cover the retain use case (to retain some ECR repo, or to retain some resources artifacts)

samcli/lib/package/packageable_resources.py
hnnasit: Yes that is correct.

samcli/lib/package/ecr_uploader.py
hnnasit: I can add an echo statement to let the user know about this.

samcli/commands/delete/delete_context.py
hnnasit: So the idea was if the stack has termination protection enabled, delete_stack will raise an error and not delete any artifacts. But if it raised CfDeleteFailedStatusError, then delete the artifacts before deleting the stack as there might be repositories which need to be retained.

samcli/commands/delete/delete_context.py
hnnasit: If termination protection is enabled for the ECR companion stack, we don't want to delete the ECR repositories.

samcli/commands/delete/delete_context.py
moelasmar: is `termination protection enabled` something that we can know without just trying to delete, and fail.
what should we do in case if SAM failed to delete the ECR Repos

samcli/lib/package/ecr_uploader.py
moelasmar: my comment is not about writing some info to the customer, my question is shall SAM CLI fail in this case, and if yes, it should fail using some custom user exception, and if not, so will notifying the customer be enough, and what is the message we need to give it to the customer

samcli/lib/package/packageable_resources.py
moelasmar: - shall you add a comment on that, and also can you explain in this case what will be the expected behaviour (shall SAM CLI only delete the stack, and skip deleting the artifact in S3)
- Also add TODO message to fix this
- One more thing the Ref intrinsic can be used in the Image Resources as well, so I think you need to do the same in its class.

samcli/commands/delete/delete_context.py
hnnasit: describe_stacks api can be used to get this information. I can modify has_stack method in cf_utils.py to check if termination protection is enabled. If it is enabled, raise an exception.

samcli/lib/package/packageable_resources.py
hnnasit: That makes sense, I will make those changes.

samcli/commands/delete/delete_context.py
```diff
@@ -57,9 +58,15 @@ def __enter__(self):
...
-            self.stack_name = prompt(
-                click.style("\tEnter stack name you want to delete:", bold=True), type=click.STRING
-            )
+            if not self.no_prompts:
+                self.stack_name = prompt(
+                    click.style("\tEnter stack name you want to delete:", bold=True), type=click.STRING
+                )
+            else:
+                raise click.BadOptionUsage(
+                    option_name="delete",
```
hawflau, line 67: `option_name` should match the option being used. In this case, it should be `"--stack-name"`

samcli/lib/package/ecr_uploader.py
moelasmar: sorry @hnnasit .. it is my bad, I missed the `not` in the condition.

samcli/lib/package/s3_uploader.py
```diff
@@ -189,7 +189,7 @@ def delete_prefix_artifacts(self):
             LOG.error("Bucket not specified")
             raise BucketNotSpecifiedError()
         if self.prefix:
-            response = self.s3.list_objects_v2(Bucket=self.bucket_name, Prefix=self.prefix)
+            response = self.s3.list_objects_v2(Bucket=self.bucket_name, Prefix=self.prefix + "/")
```
hawflau, line 192: Can you please add a note explaining why "/" is needed?

moelasmar (approved):

hawflau (approved):
LGTM!

PR #3124, opened 2021-07-29
fix: integration test case related to recent fix on ruby

wchengru:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/buildcmd/test_build_cmd.py
+++ b/tests/integration/buildcmd/test_build_cmd.py
@@ -1132,8 +1132,8 @@ class TestBuildWithBuildMethod(BuildIntegBase):
 
         # runtime is chosen based off current python version.
         runtime = self._get_python_version()
-        # BuildMethod is set to the ruby2.7, this should cause failure.
-        overrides = {"Runtime": runtime, "CodeUri": "Provided", "Handler": "main.handler", "BuildMethod": "ruby2.7"}
+        # BuildMethod is set to the java8, this should cause failure.
+        overrides = {"Runtime": runtime, "CodeUri": "Provided", "Handler": "main.handler", "BuildMethod": "java8"}
         manifest_path = os.path.join(self.test_data_path, "Provided", "requirements.txt")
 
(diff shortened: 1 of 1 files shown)


c2tarun (approved):

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1132,8 +1132,8 @@ def test_with_wrong_builder_specified_python_runtime(self, use_container):
 
         # runtime is chosen based off current python version.
         runtime = self._get_python_version()
-        # BuildMethod is set to the ruby2.7, this should cause failure.
-        overrides = {"Runtime": runtime, "CodeUri": "Provided", "Handler": "main.handler", "BuildMethod": "ruby2.7"}
+        # BuildMethod is set to the java8, this should cause failure.
+        overrides = {"Runtime": runtime, "CodeUri": "Provided", "Handler": "main.handler", "BuildMethod": "java8"}
```
hawflau, line 1136: Why did "ruby2.7" not work for this test case?

wchengru: Recently we have released a change to unblock Ruby users to build their project from requiring a manifest file. In this test case, the project is provided a python manifest but this will not fail Ruby builder.

mildaniel:
There is already a PR out to fix this issue https://github.com/aws/aws-sam-cli/pull/3100 which takes a slightly different approach to fixing this test.
I think we should consider the purpose of this test, however. If the test case is `test_with_wrong_builder_specified_python_runtime` and we expect it to fail because of a runtime/build method mismatch, the test isn't really doing that if it's only failing because the provided build method is missing a manifest.

mildaniel (approved):

tests/integration/buildcmd/test_build_cmd.py
hawflau: got it, thanks!

PR #3126, opened 2021-07-29
Feature sam check pricing

jordon-ogo:
#### Checklist

- [x] Pricing questions are now asked for Lambda Functions (calculations not done yet)
- [x] Moved pricing data storage to graph instead of lambda function object

#### TODO

- [x] Resolve pylint errors
- [x] Resolve merge conflicts
- [x] Resolve comments


samcli/commands/check/resources/Graph.py
```diff
@@ -11,6 +13,7 @@ class Graph:
     yellow_warnings: List
     red_warnings: List
     red_burst_warnings: List
+    lambda_function_pricing_info: LambdaFunctionPricing
 
     def __init__(self):
         self.entry_points: List = []
```
moelasmar, line 19: can you be more specific while using Typing. as an example, this variable if of type List, can you be more specific and mention list of what

samcli/commands/check/resources/LambdaFunctionPricing.py
```diff
@@ -0,0 +1,17 @@
...
+Object to contain Lambda function pricing info.
+This data gets stored in the graph, not a lambda function object
+"""
+
+
+class LambdaFunctionPricing:
+    number_of_requests: int
+    average_duration: int
+    allocated_memory: float
+    allocated_memory_unit: str
```
moelasmar, line 11: same comment about class variables, and public

samcli/commands/check/resources/LambdaFunctionPricing.py
```diff
@@ -0,0 +1,17 @@
+"""
+Object to contain Lambda function pricing info.
+This data gets stored in the graph, not a lambda function object
+"""
+
+
+class LambdaFunctionPricing:
```
moelasmar, line 7: add some comments about the class usage.

samcli/commands/check/resources/LambdaFunctionPricing.py
```diff
@@ -0,0 +1,17 @@
+"""
+Object to contain Lambda function pricing info.
+This data gets stored in the graph, not a lambda function object
```
moelasmar, line 3: the comment added here should be related to the module itself, not the class or objects

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+from samcli.commands.check.resources.LambdaFunctionPricing import LambdaFunctionPricing
+from samcli.commands._utils.resources import AWS_LAMBDA_FUNCTION
+
+
+class Pricing:
+    _graph: Graph
+    _max_num_requests: int
+    _min_memory_amount: int
+    _max_memory_amount: int
+    _max_duration: int
```
moelasmar, line 20: class variables

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+        user_input_memory, user_input_unit = self._ask_memory(
+            'Enter the amount of memory allocated (128MB - 10GB), followed by a ":", '
+            "followed by a valid unit of measurement [MB|GB]"
+        )
+        lambda_funciton_pricing.allocated_memory = user_input_memory
+        lambda_funciton_pricing.allocated_memory_unit = user_input_unit
+
+        self._graph.lambda_function_pricing_info = lambda_funciton_pricing
+
+    def _correct_memory_input(self, user_input_split: List) -> bool:
```
moelasmar, line 71: i think it is better to name it _validate_memory_input

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+
+from typing import List, Tuple
+import click
+
+from samcli.commands.check.resources.Graph import Graph
+from samcli.commands.check.resources.LambdaFunctionPricing import LambdaFunctionPricing
+from samcli.commands._utils.resources import AWS_LAMBDA_FUNCTION
+
+
+class Pricing:
```
moelasmar, line 15: add class documentation

samcli/commands/check/resources/TemplateResource.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, resource_object: Any, resource_type: str):
             resource_object (Function): The resource object form the template file
             resource_type (str): The resource type
         """
-        self._resource_object = resource_object
-        self._resource_type = resource_type
+        self._resource_object: Any = resource_object
```
moelasmar, line 19: will it be really Any, or dict ?

samcli/commands/check/resources/Graph.py
```diff
@@ -19,6 +22,7 @@ def __init__(self):
         self.yellow_warnings: List = []
         self.red_warnings: List = []
         self.red_burst_warnings: List = []
+        self.lambda_function_pricing_info: LambdaFunctionPricing = None
 
     def generate(self, lambda_functions: List) -> None:
```
moelasmar, line 27: I think this logic should be done in init, unless you want to enable generating the same graph multiple times.

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+        while not valid_user_input:
+            user_input = click.prompt(text=question, type=str)
+            user_input_split = user_input.split(":")
+            if self._correct_memory_input(user_input_split):
+                valid_user_input = True
+
+        return user_input_split[0], user_input_split[1]
+
+
+def _ask(question: str, min_val: int = 1, max_val: float = float("inf")) -> int:
```
moelasmar, line 134: this method is a utility method, I think it should be moved to a common module. I also can find the same method used here https://github.com/aws/aws-sam-cli/blob/3a5322c3cece6b79434b155fc38e61fd8e700d2a/samcli/commands/check/bottle_necks.py#L86

moelasmar (commented):
I have general concern about the design of your classes, your solution is designed with the knowledge that we support lambda function resources for now, and I think in future, a refactoring will be required to extend these classes to support other resources.

Also, some of the module names are not following the python module names pattern .. see this link https://www.python.org/dev/peps/pep-0008/#package-and-module-names

samcli/commands/check/resources/Pricing.py
mndeveci, line 15: This class named as `Pricing` but it looks like it is specific to Lambda, can we rename the class name to reflect that?

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(self, graph: Graph) -> None:
+        """
+        Args:
+            graph (Graph): The graph object. This is where all of the data is stored
+        """
+        self._graph: Graph = graph
+        self._max_num_requests: int = 1000000000000000000000
+        self._min_memory_amount: int = 128
+        self._max_memory_amount: int = 10000
+        self._max_duration: int = 900000
```
mndeveci, line 31: Why we have defaults for these values? Please make them clear that why we have these values as defaults.

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        asked_lambda_questions = False
+        click.echo("Pricing Questions")
+        for resource in self._graph.resources_to_analyze:
+            # Only ask lambda quetions once for all lambda functions
+            if resource.resource_type == AWS_LAMBDA_FUNCTION and not asked_lambda_questions:
+                asked_lambda_questions = True
+                self._ask_lambda_function_questions()
+
+    def _ask_lambda_function_questions(self) -> None:
```
mndeveci, line 46: I was thinking this class is specifically used for Lambda type since it has values of `max_num_requests`, `min_memory_amount`, but now I see `_ask_lambda_function_questions`. 

If this class is generic then it shouldn't contain properties specific to a resource type.
If this class is specific to a resource type, then it doesn't need `_ask_lambda_questions`.

In any case, let's target for something which is easily expandable. We should contain resource specific parameters in the resource specific classes. And when we need information for certain resource type, we can instantiate that class instance and use it to ask the questions.

samcli/commands/check/resources/Pricing.py
```diff
@@ -0,0 +1,155 @@
...
+        """Lambda function pricing questions"""
+
+        lambda_funciton_pricing = LambdaFunctionPricing()
+        user_input_requests = _ask(
+            "What are the total number of requests expected from all lambda functions in a given month?",
+            1,
+            self._max_num_requests,
+        )
+        lambda_funciton_pricing.number_of_requests = user_input_requests
```
mndeveci, line 55: Actually, you kind of implemented what I asked in the previous comment. Since you have a class type called `LambdaFuctionPricing`, all Lambda related questions can be moved there, and you don't need to define it here. For instance, from here you can call;
```
lambda_funciton_pricing = LambdaFunctionPricing()
lambda_function_pricing.ask_questions()
```

and in the `LambdaFunctionPricing`, you can ask the actual questions;
```
def ask_questions():
        user_input_requests = _ask(
            "What are the total number of requests expected from all lambda functions in a given month?",
            1,
            self._max_num_requests,
        )
        self.number_of_requests = user_input_requests

        user_input_duration = _ask(
            "What is the expected average duration of all lambda functions (ms)?", 1, self._max_duration
        )
        self.average_duration = user_input_duration

        user_input_memory, user_input_unit = self._ask_memory(
            'Enter the amount of memory allocated (128MB - 10GB), followed by a ":", '
            "followed by a valid unit of measurement [MB|GB]"
        )
        self.allocated_memory = user_input_memory
        self.allocated_memory_unit = user_input_unit
```

mndeveci (changes requested):

samcli/commands/check/resources/Graph.py
```diff
@@ -11,6 +13,7 @@ class Graph:
     yellow_warnings: List
     red_warnings: List
     red_burst_warnings: List
```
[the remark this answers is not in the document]
jordon-ogo: I initially didn't have them as this. I was informed in a previous pr that if I only use these variables as `getters` and `setters`, then this is how I should orchestrate them instead of using properties.

samcli/commands/check/resources/LambdaFunctionPricing.py
jordon-ogo: I explained why in above comment

samcli/commands/check/resources/Pricing.py
jordon-ogo: Explained above

samcli/commands/check/resources/Graph.py
jordon-ogo: A graph will only be generated once. I will have that logic called in `init`

samcli/commands/check/resources/Pricing.py
jordon-ogo: It looks like that because I only have pricing implemented for lambda functions. Eventually when pricing is implemented for more resources, the logic for them will be handled here.

samcli/commands/check/resources/Pricing.py
jordon-ogo: I will add comments explaining the meaning behind each one

samcli/commands/check/resources/Pricing.py
jordon-ogo: I see what you're saying. This class will eventually be for all resources, so I will extract the lambda pricing logic to its own class.

samcli/commands/check/resources/Pricing.py
jordon-ogo: I agree. I will implement it that way.

samcli/commands/check/resources/TemplateResource.py
jordon-ogo: dict

samcli/commands/check/resources/Graph.py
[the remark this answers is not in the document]
hawflau: These are not class variables as no value is assigned to them, but only for declaring the types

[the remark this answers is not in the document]
jordon-ogo: Since I do not instantiate the variables, they are not defined as class variables. 
https://paste.amazon.com/show/ojordon/1627855166

samcli/commands/check/calculation.py
```diff
@@ -7,14 +7,14 @@
 
 
 from samcli.commands._utils.resources import AWS_LAMBDA_FUNCTION
-from samcli.commands.check.resources.Graph import Graph
-from .resources.Warning import CheckWarning
+from samcli.commands.check.resources.graph import CheckGraph
+from .resources.warning import CheckWarning
 
 
 class Calculation:
```
mndeveci, line 14: Just like Graph, this class have very generic name, please update it with more unique name.

samcli/commands/check/lib/ask_question.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def ask(question: str, min_val: int = 1, max_val: float = float("inf")) -> int:
+    """Prompts the user with a question for some input. A min and max value
+    can be passed and used.
+
+    Args:
+        question (str): The question to ask the user
+        min_val (int, optional): The min value the user can enter. Defaults to 1.
+        max_val (float, optional): The max value the user can enter. Defaults to float("inf").
```
mndeveci, line 16: Why `min_val` is `int` but `max_val` is in `float`?

samcli/commands/check/resources/graph.py
```diff
@@ -0,0 +1,50 @@
...
+    - resources_to_analyze are all resources that need to be directly checked for bottle
+      neck issues (just lambda functions for now).
+    - green/yellow/red/red_burst_warnings are the four different warning types that
+      the user can be presented
+    - lambda_function_pricing_info will contain the user entered data for the lambda
+      pricing questions
+    """
+
+    entry_points: List[Union[LambdaFunction]]
+    resources_to_analyze: List[Union[LambdaFunction]]
```
mndeveci, line 24: AFAIR, `Union` is used if you have 2 different types that might be assigned to the variable. If so why we have a `Union[LambdaFunction]` which suggest to have single type (`LambdaFunction`)?

samcli/commands/check/lib/ask_question.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Returns:
+        [type]: [description]
```
mndeveci, line 19: Please update the method doc

samcli/commands/check/lib/ask_question.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+import click
+
+
+def ask(question: str, min_val: int = 1, max_val: float = float("inf")) -> int:
+    """Prompts the user with a question for some input. A min and max value
+    can be passed and used.
+
+    Args:
```
mndeveci, line 13: Please update documentation to numpy style; https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

samcli/commands/check/calculation.py
jordon-ogo: I will do the same for pricing as well

samcli/commands/check/lib/ask_question.py
jordon-ogo: There wasn't a need for min_val to be a float. I will change it to a float to keep things consistent.

samcli/commands/check/resources/graph.py
jordon-ogo: Once more resources are added to the project, `Union` will be needed there. I just preemptively put it there, since it will be needed very soon.

samcli/commands/check/lib/ask_question.py
```diff
@@ -0,0 +1,36 @@
+"""
+A utility method used to ask the user a question. An int is returned only
+if it is within the specified range.
+"""
+
+import click
+
+
+def ask(question: str, min_val: float = 1, max_val: float = float("inf")) -> int:
```
mndeveci, line 9: nit: Regarding to `int`->`float` discussion, you may think about adding overloaded versions of `ask` which will have different types for the min and max values.

samcli/commands/check/resources/pricing.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        Pricing quetions for various resources get asked here
+        Pricing is only done for Lambda functions now
+        """
+        asked_lambda_questions = False
+        click.echo("Pricing Questions")
+        for resource in self._graph.resources_to_analyze:
+            # Only ask lambda quetions once for all lambda functions
+            if resource.resource_type == AWS_LAMBDA_FUNCTION and not asked_lambda_questions:
+                asked_lambda_questions = True
```
mndeveci, line 36: Why we have `asked_lambda_questions` variable here? Shouldn't we ask these questions for each resource?

samcli/commands/check/resources/template_resource.py
```diff
@@ -0,0 +1,37 @@
...
+        self._resource_type: str = resource_type
+
+    @property
+    def resource_object(self) -> Function:
+        return self._resource_object
+
+    @property
+    @abc.abstractmethod
+    def resource_name(self) -> Any:
+        raise NotImplementedError()
```
mndeveci, line 33: Just curious, why we have `resource_name` as abstract method? I assume this is just a string property, we may think about assigning it in the constructor rather than making it abstract.

samcli/commands/check/resources/graph.py
hawflau: Will this grow to a long list of unions once we add more resources? 
Can there be a parent class or something which collectively represents these resources? not only for type checking, but also for exposing a common interface.

samcli/commands/check/resources/graph.py
```diff
@@ -0,0 +1,53 @@
...
+    green_warnings: List[CheckWarning]
+    yellow_warnings: List[CheckWarning]
+    red_warnings: List[CheckWarning]
+    red_burst_warnings: List[CheckWarning]
+    lambda_function_pricing_info: Optional[LambdaFunctionPricing]
```
mndeveci, line 30: As we discussed yesterday, these resource specific implementations shouldn't be leaked outside of its own class. We can think about moving them into interface definition or abstract classes. Ex;

```
unique_pricing_info: List[UniquePricingInfo]
```

```
class UniquePricingInfo(ABC):
    """Any common methods that we need to call can be defined in this class"""

class LambdaFunctionPricing(UniquePricingInfo):
    """
    This class will extend from UniquePricingInfo, and can implement the common methods 
    which will have Lambda specific business logic here
    """

```

samcli/commands/check/resources/graph.py
jordon-ogo: Yes. `TemplateResource` is the parent class to `LambdaFunction` and all future resources. I will use that for all future type hints involving many resources

samcli/commands/check/resources/pricing.py
jordon-ogo: Not for lambda functions. They are the only resource where we want to ask the pricing questions regarding all lambda functions, not each individual one.

samcli/commands/check/calculation.py
```diff
@@ -7,17 +7,19 @@
...
+    _graph: CheckGraph
 
-    def __init__(self, graph: Graph):
+    def __init__(self, graph: CheckGraph):
         """
-        Args:
-            graph (Graph): The graph object. This is where all of the data is stored
+        Parameters
+        ----------
+            graph: Graph
```
hawflau, line 21: Should this be `CheckGraph` as well?

jordon-ogo: I missed that one. I'll double check all of those docs and make any necessary changes.

samcli/commands/check/lib/command_context.py
```diff
@@ -124,11 +139,13 @@ def _read_sam_file(self) -> Any:
...
 
-def _parse_template() -> Graph:
+def _parse_template() -> CheckGraph:
     """Parses the template to retrieve resources
 
-    Returns:
-        Graph: Returns the generated graph object
+    Returns
+    -------
+        Graph
```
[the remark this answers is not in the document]
jordon-ogo: I forgot to update the docs with my class name changes. I'll update all of them

samcli/commands/check/resources/graph.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, lambda_functions: List):
```
hawflau, line 32: Can we be more specific about the List like `List[LambdaFunction]`?

samcli/commands/check/resources/graph.py
jordon-ogo: Ok. I'll leave it for now.

samcli/commands/check/resources/graph.py
jordon-ogo: I agree. I will create a super class for all pricing objects. Instead of `unique_pricing_info` being a `List`, I think using a `dict` would keep things more organized. `unique_pricing_info = {"LambdaFunction": LambdaFunctionPricing}`.

Until pricing is implemented for more resources, the super class won't be used, as there aren't any common methods since lambda functions are the only resources with pricing implemented. The framework will still be there for other resources once they get pricing implemented for them.

samcli/commands/check/resources/template_resource.py
```diff
@@ -0,0 +1,26 @@
...
+        Parameters
+        ----------
+            resource_object: Function
+                The resource object form the template file
+            resource_type: str
+                The resource type
+            resource_name: str
+                The name of the resource
+        """
+        self.resource_object: Function = resource_object
```
hawflau, line 24: as you've already declared the type above, you don't need to declare the type here again.

samcli/commands/check/lib/command_context.py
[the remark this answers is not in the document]
hawflau: Looks like you updated all others but missed this particular line

samcli/commands/check/resources/graph.py
hawflau: Should this line be `def __init__(self, lambda_function: List[LambdaFunction])`?

hawflau (approved):

samcli/commands/check/resources/lambda_function_pricing.py
```diff
@@ -0,0 +1,144 @@
...
+Object to contain Lambda function pricing info.
+"""
+from typing import List, Tuple
+import click
+
+from samcli.commands.check.lib.ask_question import ask
+from samcli.commands.check.resources.unique_pricing_info import UniquePricingInfo
+
+
+class LambdaFunctionPricing(UniquePricingInfo):
```
mndeveci, line 11: The reason I suggested to have a generic interface (`UniquePricingInfo`) and a special implementation (`LambdaFunctionPricing`) is because you can put lambda specific questions and answers in the implementation where the generic interface can have a generic method which won't expose resource specific stuff to the outside.

Since you created this `UniquePricingInfo`, you can now have an abstract method in that interface;

```
class UniquePricingInfo(ABC):
  
  @abstractmethod
  def ask_questions(self):
    pass
```

And in the `LambdaFunctionPricing` implementation you can implement that method as;

```
class LambdaFunctionPricing(UniquePricingInfo):

  def ask_questions(self):
    # lambda resource specific questions
    self.ask_lambda_function_questions()
    ...
    ...
```

mndeveci (approved):
I left one comment about abstraction that we are trying to achieve. I am approving this to unblock you for next week.

samcli/commands/check/resources/lambda_function_pricing.py
jordon-ogo: I understand. I will implement pricing (and also other features) this way.

Issue #3133, opened 2021-08-03
Cache is missed when building Layer inside a container


Macok:
### Description:
When a Layer is built using `--use-container` option and any `--container-env-var` are passed, the cache is always missed.
This was already fixed by @mndeveci here: https://github.com/aws/aws-sam-cli/pull/2943, but only for Lambda Functions, while the same problem exists for layers.

### Steps to reproduce:

Run below command twice to build any Layer (it will work fine in case of Functions) and see that the cache is not valid.
`sam build --cached --use-container --container-env-var FOO=BAR`

### Observed result:

Cache is missed.

### Expected result:

Cache should be hit.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Linux
2. `sam --version`: SAM CLI, version 1.27.2
3. AWS region: N/A

### Why does this happen?
Just like in case of https://github.com/aws/aws-sam-cli/pull/2943, a dictionary of env variables is being passed to the build function by `DefaultBuildStrategy.build_single_layer_definition`. Additional `LAMBDA_BUILDERS_LOG_LEVEL` variable is added to the dictionary, causing cache miss. We need to make a deepcopy of this dictionary, just like in `DefaultBuildStrategy.build_single_function_definition`.

I'm happy to work on a fix as soon as someone confirms my observations.

sriram-mv:
@Macok sounds like a very good contribution to take on.

Macok:
Awesome, I'll submit a PR later this week.

Macok:
Done, waiting for review: #3139

PR #3139, opened 2021-08-04
Fix: Pass a copy of environment variables when building layers

Macok:

#3133

During the build, additional env variable `LAMBDA_BUILDERS_LOG_LEVEL` is added to the dict.
This additional variable ends up being persisted in build.toml, making cache always invalid.
This was already addressed here #2943 for Lambda Functions, but similar problems affects Lambda Layers.
 
A copy of the dictionary is made, so the original dictionary stays unchanged.

None.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1346,36 +1391,6 @@ def test_cache_build(self, use_container, code_uri, function1_handler, function2
...
-        )
-
-        LOG.info("Re-Running Command (valid cache should exist): %s", cmdlist)
-        command_result_with_cache = run_command(cmdlist, cwd=self.working_dir)
-
-        self.assertTrue(
-            "Valid cache found, copying previously built resources from function build definition"
-            in command_result_with_cache.stderr.decode("utf-8")
-        )
-
```
Macok, line 1378: I removed this test because `TestRepeatedBuildHitsCache` covers all this.

samcli/lib/build/build_strategy.py
```diff
@@ -163,6 +163,10 @@ def build_single_layer_definition(self, layer_definition: LayerBuildDefinition)
             )
 
         single_build_dir = layer.get_build_dir(self._build_dir)
+
+        # pass down a copy of env_vars, to avoid modifying the original env_vars dict
+        container_env_vars = deepcopy(layer_definition.env_vars)
```
jfuss, line 168: I wonder if we need a better way to do this? It seems like we are doing this in a couple places and maybe it is the right thing but seems like the next time we add a resource to build, we will need to do this all over again. @mndeveci what are your thoughts? I am not as familiar with the caching logic as you are.

Not opposed to this, just trying to think through if we can make this better for the future.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1291,6 +1291,51 @@ def _verify_process_code_and_output(self, command_result):
...
+@skipIf(
+    ((IS_WINDOWS and RUNNING_ON_CI) and not CI_OVERRIDE),
+    "Skip build tests on windows when running in CI unless overridden",
+)
+class TestRepeatedBuildHitsCache(BuildIntegBase):
+    # Use template containing both functions and layers
+    template = "layers-functions-template.yaml"
+
+    @parameterized.expand([(True,), (False,)])
+    @skipIf(SKIP_DOCKER_TESTS, SKIP_DOCKER_MESSAGE)
```
jfuss, line 1303: Can we just move this test under the "cache builds" class we already have?

jfuss (changes requested):

samcli/lib/build/build_strategy.py
mndeveci: That is a good point @jfuss. I missed layers in the first PR, and this one will cover that too. And the root cause is coming from pass-by-reference. This `env_vars` dictionary is altered in the way below which changes the `build_definition` signature which causes re-builds for cached build.

But instead of `deepcopy`'ing in different places, we can think about moving `env_vars` property of [`FunctionBuildDefinition`](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/build_graph.py#L406) and [`LayerBuildDefinition`](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/build_graph.py#L350) into [`AbstractBuildDefinition`](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/build_graph.py#L320) and then add a property method which will return the copied instance of dictionary.

```
class AbstractBuildDefinition:
    """
    Abstract class for build definition
    Build definition holds information about each unique build
    """

    def __init__(self, source_md5: str, env_vars: Optional[Dict] = None) -> None:
        self.uuid = str(uuid4())
        self.source_md5 = source_md5
        self._env_vars = env_vars if env_vars else {}

    @property
    def env_vars(self):
        return deepcopy(self._env_vars)
```

@Macok really appreciate your work on this, would you be willing to update and test your implementation with the one I suggested above?

Macok: @mndeveci @jfuss These are all valid points and I like the solution, let me fix this quickly

tests/integration/buildcmd/test_build_cmd.py
Macok: Yes good point :+1:

samcli/lib/build/build_strategy.py
Macok: @mndeveci @jfuss Done :+1:

tests/integration/buildcmd/test_build_cmd.py
Macok: @jfuss You mean to move this test inside the `TestBuildWithCacheBuilds` class?
The problem is that `TestBuildWithCacheBuilds` uses `dedup-functions-template.yaml`, which doesn't have any Layers. So it won't test what we want.

If it's important, I can probably refactor it to use `layers-functions-template.yaml`, and then we can have both tests inside the same class.

mndeveci (changes requested):
Thanks for the changes @Macok!

I verified your changes on my local machine and I can confirm that these changes are working as expected.

One thing though, when I tried to run integration test `pytest tests/integration/buildcmd/test_build_cmd.py  -k TestRepeatedBuildHitsCache` it fails. Can you check what is the issue there?

Macok:
> Thanks for the changes @Macok!
> 
> I verified your changes on my local machine and I can confirm that these changes are working as expected.
> 
> One thing though, when I tried to run integration test `pytest tests/integration/buildcmd/test_build_cmd.py -k TestRepeatedBuildHitsCache` it fails. Can you check what is the issue there?

@mndeveci
Aren't you simply forgetting about `SAM_CLI_DEV` flag?
For me, `SAM_CLI_DEV=1 pytest tests/integration/buildcmd/test_build_cmd.py -k TestRepeatedBuildHitsCache` runs fine on 2 different machines; `make pr` passes as well.

If that's not it, could you post the logs?

Macok:
@mndeveci , @jfuss would you take a look again?

Macok:
@jfuss Only your approval missing.

jfuss (approved):
@Macok Thanks for the contribution and going back to refactor.

PR #3141, opened 2021-08-04
fix: Dependency manager and runtime mismatch

jonife:
Users can pass `dependency-mananager` and `runtime` using options like --dependency-manager  and --runtime to samcli as a way of selecting what dependency-manager and runtime. The dependency-manager passed is not validated against the runtime, so when a customers pass `pip` as a dep for java8 samcli doesn't do any validation. This cause samcli to freeze as it can't spool the appropriate template

add a check to validate the `dependency manager` against `runtime`. To raise an exception informing the Users that the wrong Dependency manager was provide for the  runtime selected.

exception handling for mismatch runtime and dependency manager 

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes


samcli/commands/init/interactive_init_flow.py
```diff
@@ -183,25 +183,33 @@ def _get_runtime_from_image(image):
...
-            click.echo("\nWhich dependency manager would you like to use?")
-            for dm in valid_dep_managers:
-                msg = "\t" + str(choice_num) + " - " + dm
-                click.echo(msg)
-                choice_num = choice_num + 1
-            choice = click.prompt("Dependency manager", type=click.Choice(choices), show_choices=False)
-            dependency_manager = valid_dep_managers[int(choice) - 1]  # zero index
+    valid_dep_managers = RUNTIME_TO_DEPENDENCY_MANAGERS.get(runtime)
+    if dependency_manager:
+        _validate_runtime_dependancy_manager(dependency_manager, runtime)
```
mgrandis, line 188: I think you can simply put the lines 208 -> 210 here instead, this will also avoid repeating the call to `RUNTIME_TO_DEPENDENCY_MANAGERS.get(runtime)`

samcli/commands/init/interactive_init_flow.py
```diff
@@ -183,25 +183,33 @@ def _get_runtime_from_image(image):
...
+            choice_num = choice_num + 1
+        choice = click.prompt("Dependency manager", type=click.Choice(choices), show_choices=False)
+        dependency_manager = valid_dep_managers[int(choice) - 1]  # zero index
     return dependency_manager
 
 
+def _validate_runtime_dependancy_manager(dependency_manager, runtime):
+    valid_dep_managers = RUNTIME_TO_DEPENDENCY_MANAGERS.get(runtime)
+    if dependency_manager not in valid_dep_managers:
+        msg = f"Wrong dependency manager {dependency_manager} selected for Lambda Runtime {runtime}."
```
mgrandis, line 209: Maybe `Unsupported` instead of `Wrong`

tests/unit/commands/init/test_cli.py
```diff
@@ -224,6 +224,27 @@ def test_init_fails_invalid_dep_mgr(self, git_repo_clone_mock):
                 extra_context=None,
             )
 
+    @patch("samcli.lib.utils.git_repo.GitRepo.clone")
+    def test_init_fails_wrong_dep_mgr_for_runtime(self, git_repo_clone_mock):
+        # WHEN the wrong dependency_manager is passed for a runtime
+        # THEN an exception should be raised
+        with self.assertRaises(UserException):
```
mgrandis, line 231: Isn't this a `InvalidInitOptionException`?
We may also want to check the message of the exception as it is a generic exception for init options
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/unit/commands/init/test_cli.py
```diff
@@ -1387,3 +1408,29 @@ def test_init_cli_image_pool_with_base_image_having_one_managed_template_with_pr
                 no_input=None,
                 extra_context=None,
             )
+
+    @patch("samcli.lib.utils.git_repo.GitRepo._ensure_clone_directory_exists")
+    def test_init_cli_with_mismatch_dep_runtime(self, cd_mock):
```
mgrandis, line 1413: There also, check the exception type and message instead of only an `assertTrue`

mgrandis (changes requested):
A few remarks

tests/unit/commands/init/test_cli.py
jonife: `InvalidInitOptionException` is of type `UserException`. I realized that other test used this and as such for consistency I used it, since `InvalidInitOptionException` is of type `UserException`

mgrandis: If another code path raises an exception inheriting `UserException`, your test would still pass so you should expect an `InvalidInitOptionException`.

tests/unit/commands/init/test_cli.py
mgrandis: My bad, on the message, type is not accessible

mndeveci:
@jonife do we still need this PR?

jonife:
> @jonife do we still need this PR?

no I don't think so, I added it to the new init project.

PR #3146, opened 2021-08-06
cdk-support-dev branch merged to develop branch

moelasmar:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/deploy/guided_context.py
```diff
@@ -119,8 +120,8 @@ def guided_prompts(self):
         Start an interactive cli prompt to collection information for deployment
 
         """
-        default_stack_name = self.stack_name
-        default_region = self.region or get_session().get_config_variable("region") or "us-east-1"
+        default_stack_name = self.stack_name or "sam-app"
```
hawflau, line 123: nit: We probably don't need to set the default value "sam-app" as I remember it is set somewhere before this step. But it shouldn't hurt to also have it here for now. My thought is that the default stack name is specific to which IaC platform is being used, so it should be set by the specific IaC plugin.

wchengru (approved):
LGTM.
I noticed a couple of commits are using codeuri and layers. I believe that the codeur in CDK formati has no problem there, but want to give a head up to avoid any potential regression.
* https://github.com/aws/aws-sam-cli/pull/3146/commits/ba81041e969ae696e49ff1ffc46c8c9f64ba83a0
* https://github.com/aws/aws-sam-cli/pull/3146/commits/32c1a5fef8636edca4f37a00054d4be3cc8a4de9

tests/testing_utils.py
```diff
@@ -50,6 +51,12 @@ def run_command_with_input(command_list, stdin_input, cwd=None, env=None, timeou
         raise
 
 
+def run_command_with_inputs(command_list: List[str], inputs: List[str], timeout=TIMEOUT) -> CommandResult:
```
hawflau, line 54: should the arguments here match the full argument list of `run_command_with_input`? (missing `cwd`)

hawflau (approved):
LGTM!

PR #3156, opened 2021-08-10
Import registered command to pyinstaller hook

wchengru:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/cli/command.py
```diff
@@ -62,6 +62,15 @@ def __init__(self, *args, cmd_packages=None, **kwargs):
 
         if not cmd_packages:
             cmd_packages = _SAM_CLI_COMMAND_PACKAGES
+        elif isinstance(cmd_packages, list):
```
CoshUS, line 65: This is in prod code. If we somehow missed importing a new command and the unit tests didn't catch it, users will see this message and have no idea what it's talking about.

CoshUS (changes requested):

samcli/cli/command.py
```diff
@@ -118,3 +127,6 @@ def get_command(self, ctx, cmd_name):
             return None
 
         return mod.cli
+
+    def get_command_packages(self):
```
jfuss, line 131: This looks like it is only used for testing. If so, is there no way to get this information through the tests?

samcli/cli/command.py
jfuss: This else will only execute is `cmd_packages` is set. If that happens (which seems unlikely), then none of the commands would (potentially) work. `cmd_packages` was added to all injection, however, I think this is mainly for test reasons at this time. I don't see a strong need for us to be doing this else, unless I am missing something (which is very much possible).

jfuss (changes requested):

samcli/cli/command.py
wchengru: Removed this one.

samcli/cli/command.py
wchengru: There is no usage of `cmd_packages` now except for some test cases. Here I added this else case to avoid using it without updating package in the future. I will remove this.

jfuss:
@wchengru This looks good to me. One follow up question: Did we test pyinstaller still works as we expect it to? Just being paranoid at this point.

CoshUS (approved):

wchengru:
> @wchengru This looks good to me. One follow up question: Did we test pyinstaller still works as we expect it to? Just being paranoid at this point.

Yes, I have tested it locally and the pyinstaller works as expected.

jfuss (approved):

PR #3170, opened 2021-08-13
Add root_template_path attribute to convert relative to absolute path

xazhao:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -404,6 +404,7 @@ def _get_stacks(self) -> List[Stack]:
                 self._template_file,
                 parameter_overrides=self._parameter_overrides,
                 global_parameter_overrides=self._global_parameter_overrides,
+                root_template_path=os.path.dirname(os.path.abspath(self._template_file)),
```
sriram-mv, line 407: Should we add a comment on why a root template path is required?

sriram-mv (changes requested):

samcli/commands/local/cli_common/invoke_context.py
xazhao: Sure! Since this pr is a little outdated and the branch is deleted. I will recreated a PR for this change.

PR #3179, opened 2021-08-18
Translate Makefile to PS script to run on Windows

ssenchenko:
To make SAM CLI development on Windows more convenient and closer to unix-based system experience

Creating a PowerShell script which performs the same commands in the project as Makefile/make on unix

We should maintain Makefile and Make.ps1 together to keep them in sync.

`update-reproducible-reqs` command from Makefile has not been added to Make.ps1 because it performs linux specific tasks.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


Make.ps1
```diff
@@ -0,0 +1,163 @@
+<#
+.SYNOPSIS
+    Run on Windows the same commands as in ./Makefile without installing any aditional software.
```
sriram-mv, line 3: Overall looks good, we may need to update README or development guide on how to use the new Make.ps1 and at the same time look at other mentions to ensure community knows to keep both the original Makefile and Make.ps1 in sync.

ssenchenko: Good point! I'll update those too.

Make.ps1
```diff
@@ -0,0 +1,163 @@
...
+            Init
+        }
+        $Test {
+            Test
+        }
+        $TestCovReport {
+            pytest --cov samcli --cov-report html --cov-fail-under 95 tests/unit
+        }
+        $IntegTest {
+            Write-Host "Telemetry Status: $env:SAM_CLI_TELEMETRY"
```
CoshUS, line 125: nit:
What's the reason behind using Write-Host instead of Write-Output?
Write-Host will only print on screen and can't be piped into a file.

CoshUS (changes requested):

Make.ps1
ssenchenko: Done

Make.ps1
ssenchenko: Good call-out. Changing to `Write-Output`

ssenchenko: Done. Also changed the formatted error output in the default switch case to displaying help. No `Write-Host` now.

CoshUS (approved):
LGTM!

sriram-mv (approved):

hoffa:
Does the existing Makefile work with a Windows-flavored `make`? (e.g. as installed via `choco install make`)

ssenchenko:
@hoffa No. The point is that there is no need to install any additional tools as windows-flavoured tools are often not 100% as there *nix ancestors.

PR #3215, opened 2021-08-30
sam build - image package type - streaming progress bar

sriram-mv:
* `sam build` image based functions did not stream the progress as they were pulling images, which meant a long wait time when pulling large images, without any progress bar.

* Showcase a progress bar similar to when sam deploys images to ECR, while building images as well.
* Demo of streaming logs on build
sam_build_streaming

* N/A
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,70 @@
+"""
+Log streaming utilities when streaming logs from Docker
+"""
+import os
+
+from samcli.lib.package.stream_cursor_utils import cursor_up, cursor_left, cursor_down, clear_line
+
+
+class LogStreamer:
+    def __init__(self, stream, error_class, error_msg_prefix=""):
```
CoshUS, line 10: Can we add type hinting to this class?

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,70 @@
...
+            status = log.get("status", "")
+            stream = log.get("stream", "")
+            progress = log.get("progress", "")
+            error = log.get("error", "")
+            change_cursor_count = 0
+            if _id:
+                try:
+                    curr_log_line_id = ids[_id]
+                    change_cursor_count = len(ids) - curr_log_line_id
+                    self.stream.write((cursor_up(change_cursor_count) + cursor_left).encode())
```
CoshUS, line 32: With the amount of self.stream.write(string.encode()) maybe we can separate this out into its own method which is easier to read.

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+
+from samcli.lib.package.stream_cursor_utils import cursor_up, cursor_left, cursor_down, clear_line
+
+
+class LogStreamer:
+    def __init__(self, stream, error_class, error_msg_prefix=""):
+        self.stream = stream
+        self.error_class = error_class
+        self.error_msg_prefix = error_msg_prefix
```
CoshUS, line 13: Are these fields intended to be public?

CoshUS (changes requested):
Added a few comments. Rest looks good to me.
Can we also have some unit tests for LogStreamer?

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,96 @@
...
+    CursorUpFormatter,
+    CursorDownFormatter,
+    CursorLeftFormatter,
+    ClearLineFormatter,
+)
+from samcli.lib.utils.stream_writer import StreamWriter
+
+
+class LogStreamer:
+    def __init__(self, stream: StreamWriter, error_class: APIError.__class__, error_msg_prefix=""):
```
CoshUS, line 20: Why do we pass in an exception type here instead of raising a static one, catching it, and throwing whatever is needed in the caller?
Typing might also be wrong, `APIError` is from `docker.errors` but `DockerPushFailedError` is from `UserException`.
For passing class types, `typing.Type[APIError]` should be used.

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,96 @@
...
+            if _id:
+                try:
+                    curr_log_line_id = ids[_id]
+                    change_cursor_count = len(ids) - curr_log_line_id
+                    self._stream.write(
+                        self._cursor_up_formatter.cursor_format(change_cursor_count)
+                        + self._cursor_left_formatter.cursor_format(),
+                        encode=True,
+                    )
+                except KeyError:
```
CoshUS, line 51: I assume this is for catching missing key in the statement of `curr_log_line_id = ids[_id]`.
If so, let's use `if _id in ids:` instead of using exceptions to control flow.

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,96 @@
...
+                    ids[_id] = len(ids)
+            else:
+                ids = dict()
```
CoshUS, line 54: I'm a bit confused on this part. Why do we reset `ids` when `_id` field is missing from a single log?
Some comments here might help.

CoshUS (changes requested):
Added a few more feedbacks.

samcli/lib/docker/log_streamer.py
sriram-mv: Good call.

samcli/lib/docker/log_streamer.py
sriram-mv: I dont think this is required honestly, will do some testing. Looks to be something that has crept in.

samcli/lib/build/app_builder.py
```diff
@@ -366,17 +367,11 @@ def _stream_lambda_image_build_logs(self, build_logs: List[Dict[str, str]], func
...
-                    raise DockerBuildFailed(f"{function_name} failed to build: {error_stream}")
-
-                if log_stream:
-                    self._stream_writer.write(str.encode(log_stream))
-                    self._stream_writer.flush()
+        build_log_streamer = LogStreamer(self._stream_writer)
+        try:
+            build_log_streamer.stream_progress(build_logs)
+        except LogStreamError as ex:
+            raise DockerBuildFailed(f"{function_name} failed to build: {str(ex)}")
```
CoshUS, line 374: nit: `raise DockerBuildFailed(..) from ex` for [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining).

CoshUS (approved):
Thank you for addressing all the comments!
One last small non-blocking nit.

jfuss (commented):
Everything looks fine to me. My only concern is if this breaks consumers. I know we generally view out output as not a contract but I know the AWS Toolkit watches output. Did we check with the AWS Toolkit team to see if this would actually break them at all? Paging @abrooksv and @kiiadi

justinmk3:
> Did we check with the AWS Toolkit team to see if this would actually break them at all? Paging @abrooksv and @kiiadi

in general, more-frequent output is useful to the Toolkits since they use that to check for signs of life.

but, can't SAM CLI detect if it is writing to a terminal to decide whether it should show interactive stuff like this? https://stackoverflow.com/a/1077118/152142

samcli/lib/package/stream_cursor_utils.py
```diff
@@ -16,16 +16,49 @@
...
+class CursorUpFormatter(CursorFormatter):
+    """
+    Class for formatting and outputting moving the cursor up within the stream of bytes.
+    """
+
+    def cursor_format(self, count=0):
+        return ESC + str(count) + "A"
+
+
+class CursorDownFormatter(CursorFormatter):
```
justinmk3, line 40: why is each termcode modeled as a class?

samcli/lib/docker/log_streamer.py
```diff
@@ -0,0 +1,93 @@
...
+        self._stream = stream
+        self._cursor_up_formatter = CursorUpFormatter()
+        self._cursor_down_formatter = CursorDownFormatter()
+        self._cursor_left_formatter = CursorLeftFormatter()
+        self._cursor_clear_formatter = ClearLineFormatter()
+
+    def stream_progress(self, logs: docker.APIClient.logs):
+        """
+        Stream progress from docker push logs and move the cursor based on the log id.
+        :param logs: generator from docker_clent.APIClient.logs
```
justinmk3, line 34: did we consider using https://github.com/tqdm/tqdm or another well-known python library?

samcli/commands/bootstrap/command.py
```diff
@@ -27,6 +27,6 @@ def cli(ctx):
     do_cli(ctx.region, ctx.profile)  # pragma: no cover
 
 
-def do_cli(region, profile):
+def do_cli(region, profile):  # pragma: no cover
```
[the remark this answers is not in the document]
sriram-mv: `sam bootstrap` is not an exposed command, so I was looking at pytest coverage to see which places have no tests, and where it made sense to insert pragma, no cover.

samcli/lib/package/stream_cursor_utils.py
sriram-mv: Worth looking into through a refactor.

samcli/lib/docker/log_streamer.py
sriram-mv: This is to mimic how docker cli showcases the logs, so didn't want to use another library or abstraction.

jfuss (approved):

PR #3029, opened 2021-07-08
chore: Update RIE binary to 1.1

valerena:
#### Why is this change necessary?
Consume the latest version of [RIE](https://github.com/aws/aws-lambda-runtime-interface-emulator/)

#### How does it address the issue?
It bring the latest version of RIE

#### Checklist

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #3216, opened 2021-08-30
chore: Update RIE to 1.1 and fix a bootstrap issue

wchengru:
This PR is an update of https://github.com/aws/aws-sam-cli/pull/3029
It includes this RIE PR: https://github.com/aws/aws-lambda-runtime-interface-emulator/pull/45
Local tests have been passed.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


mndeveci (approved):

wchengru:
> @wchengru Do we have an example of what ended up showing us something was wrong with RIE? My understanding was we (you) caught this manually?
> 
> If so, can we add a test for whatever that was? This was we have more confidence to catch this in the future.

I added a new integration test case to catch it. The integration tests should catch this issue, but all local invoke integration tests are running with Canary. This new test case is to catch it before merging into develop.

wchengru:
> > @wchengru Do we have an example of what ended up showing us something was wrong with RIE? My understanding was we (you) caught this manually?
> > If so, can we add a test for whatever that was? This was we have more confidence to catch this in the future.
> 
> I added a new integration test case to catch it. The integration tests should catch this issue, but all local invoke integration tests are running with Canary. This new test case is to catch it before merging into develop.

According to offline discussion, we can keep the existing tests to catch the RIE binary issue in canary. Rolled back my changes on the tests cases.

jfuss (approved):

Issue #3168, opened 2021-08-12
sam local start-api HttpApi endpoints not returning Set-Cookie headers


hooverdc:
### Description:

When using the SAM `start-api` command, events using the HttpApi ( format 2.0) integration will not return Set-Cookie headers. Events using the format 1.0 will send the Set-Cookie headers.

### Steps to reproduce:

Using the example Python3.8 container image setup for the SAM CLI.

```python
def lambda_handler(event, context):

    return {
        "statusCode": "200",
        "cookies": [
            "mycookie=mycookie; Path=/"
        ],
        "headers": {
            "Content-Type": "text/html; charset=UTF-8",
        },
        "body": "OK"
    }
```
```yaml
 AppFunction:
    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
    Properties:
      PackageType: Image
      Events:
        RootPathV2:
          Type: HttpApi
          Properties:
            Path: /v2/
            Method: ANY
```

1. Run `sam local start-api`
2. Get `http://127.0.0.1:3000/v2/` endpoint

### Observed result:

Did not send Set-Cookie headers.

### Expected result:

Send Set-Cookie headers

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOs
2. `sam --version`: `SAM CLI, version 1.27.2`
3. AWS region: N/A

`Add --debug flag to command you are running`

xazhao:
Hi @hooverdc thanks for posting this. I can reproduce this bug. It seems even if you set `payloadFormatVersion: 2.0`, it still uses payload version 1.0 which resulted `Set-Cookie` only works in headers.

xazhao:
PR merged. It should be fixed in our next release.

PR #3218, opened 2021-08-30
fix: add Set-Cookie header for sam local

xazhao:
https://github.com/aws/aws-sam-cli/issues/3168

Copy cookies from `cookies` field to `Set-Cookie` header

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -526,6 +527,20 @@ def _parse_v2_payload_format_lambda_output(lambda_output: str, binary_types, fla
...
+        if isinstance(json_output.get("cookies"), list):
+            # add all cookies to Set-Cookie header
+            if "Set-Cookie" not in headers:
+                headers["Set-Cookie"] = ""
+            for cookie in json_output.get("cookies"):
+                headers["Set-Cookie"] = headers["Set-Cookie"] + ";" + cookie
+            # remove the first semi colon and trim the header
+            if headers["Set-Cookie"][0] == ";":
+                headers["Set-Cookie"] = headers["Set-Cookie"][1:]
+            headers["Set-Cookie"] = headers["Set-Cookie"].strip()
```
aahung, line 542: nit: There are several branches that we can use "join" to simplify:

```suggestion
            headers["Set-Cookie"] = (
                ";".join([
                    cookie for cookie in [headers.get("Set-Cookie")] + json_output.get("cookies") if cookie
                ])
            ).strip()
```

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -1112,6 +1112,30 @@ def test_custom_content_type_header_is_not_modified(self):
         self.assertIn("Content-Type", headers)
         self.assertEqual(headers["Content-Type"], "text/xml")
 
+    def test_custom_cookies_in_payload_format_version_2(self):
+        lambda_output = (
+            '{"statusCode": 200, "cookies": ["cookie1=test1", "cookie2=test2"], "body": "{\\"message\\":\\"Hello from Lambda\\"}", '
+            '"isBase64Encoded": false}'
+        )
```
aahung, line 1119: can we add a test if "Set-Cookie" pre-exists in the headers?

xazhao: Sure! Added that test

samcli/local/apigw/local_apigw_service.py
```diff
@@ -506,6 +506,7 @@ def _parse_v2_payload_format_lambda_output(lambda_output: str, binary_types, fla
         :return: Tuple(int, dict, str, bool)
         """
         # pylint: disable-msg=too-many-statements
+        # pylint: disable=too-many-branches
```
aahung, line 509: Do we still need this?

aahung (approved):
LGTM

samcli/local/apigw/local_apigw_service.py
xazhao: Nope. Deleted this line

samcli/local/apigw/local_apigw_service.py
```diff
@@ -526,6 +526,15 @@ def _parse_v2_payload_format_lambda_output(lambda_output: str, binary_types, fla
...
         headers = Headers(json_output.get("headers") or {})
 
+        # cookies is a new field in payload format version 2.0 (a list)
+        # https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
+        # we need to move cookies to Set-Cookie headers.
+        if isinstance(json_output.get("cookies"), list):
+            # add all cookies to Set-Cookie header
+            headers["Set-Cookie"] = (
+                ";".join([cookie for cookie in [headers.get("Set-Cookie")] + json_output.get("cookies") if cookie])
+            ).strip()
```
hoffa, line 536: Please add a reference as to why this `Set-Cookie` format is correct.

In [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) the syntax doesn't describe allowing concatenating cookies with `;`. Instead, cookies are on separate `Set-Cookie` values.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie):

> To send multiple cookies, multiple `Set-Cookie` headers should be sent in the same response.

xazhao: Yep you're right. API Gateway only supported single `Set-Cookie` header some time ago and I read a outdated document. So I was trying to make it look like what API Gateway should do. I just found API Gateway can support multiple `Set-Cookie` headers now. Updated the code.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -526,6 +526,15 @@ def _parse_v2_payload_format_lambda_output(lambda_output: str, binary_types, fla
         status_code = json_output.get("statusCode") or 200
         headers = Headers(json_output.get("headers") or {})
 
+        # cookies is a new field in payload format version 2.0 (a list)
+        # https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
+        # we need to move cookies to Set-Cookie headers.
+        if isinstance(json_output.get("cookies"), list):
```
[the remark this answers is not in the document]
xazhao: Yep updated the code.

hoffa (approved):

Issue #2034, opened 2020-06-11
sam deploy --guided doesn't escape quotation marks for parameters written to samconfig.toml


senorkrabs:
### Description

When using `sam deploy --guided`, parameters with quotation marks `"` are written to `parameter_overrides` in samconfig.toml without proper escaping. As a result, the parameter isn't correctly populated the next time `sam deploy` is called.

### Steps to reproduce

Run `sam deploy --guided` and use these parameters:
```
        Stack Name [sftp-lambda]: 
        AWS Region [us-east-1]: 
        Parameter FilesListJSON []: [ { "SrcFile": "/athena/CUR2/CUR2/year=2020/month=6/*.snappy.parquet",  "TargetPath": "/sftp-lambda/test"}, { "SrcFile": "/athena/CUR2/CUR2/year=2020/month=5/*.snappy.parquet",  "TargetPath": "/sftp-lambda/test"}]
        #Shows you resources changes to be deployed and require a 'Y' to initiate deploy
        Confirm changes before deploy [Y/n]: Y
        #SAM needs permission to be able to create roles to connect to the resources in your template
        Allow SAM CLI IAM role creation [Y/n]: Y
        Save arguments to samconfig.toml [Y/n]: Y
```

### Observed result
Written to `samconfig.toml`:
```
parameter_overrides = "FilesListJSON=\"[ { \"SrcFile\": \"/athena/CUR2/CUR2/year=2020/month=6/*.snappy.parquet\",  \"TargetPath\": \"/sftp-lambda/test\"}, { \"SrcFile\": \"/athena/CUR2/CUR2/year=2020/month=5/*.snappy.parquet\",  \"TargetPath\": \"/sftp-lambda/test\"}]\""
```
What happens when `sam deploy` is called again:
```
$ sam deploy

        Deploying with following values
        ===============================
        Stack name                 : sftp-lambda
        Region                     : us-east-1
        Confirm changeset          : True
        Deployment s3 bucket       : aws-sam-cli-managed-default-samclisourcebucket-
        Capabilities               : ["CAPABILITY_IAM"]
        Parameter overrides        : {'FilesListJSON': '[ { '}
```
(Note that Parameter overrides is incorrect)

Please provide command output with `--debug` flag set.

### Expected result

Written to `samconfig.toml`:
```
[...]

PR #3226, opened 2021-09-01
Escape double quotes when saving and unwrap only one layer of quotes when reading parameter values ( #2034)

specious:
The function `_unquote_wrapped_quotes()` in `samcli/cli/types.py` was checking if the string is wrapped in quotes, but then performing a greedy removal on both sides.

The documentation wasn't correct either.

All the tests pass after this change.  Perhaps someone could look over this and see if it makes sense.

---


specious:
Escaping double quotes when preparing parameter values for saving in `GuidedConfig.quote_parameter_values()` properly solves  #2034.

specious:
It is now possible to save parameter overrides containing double quote characters, and they will be double-escaped as `\\\"` in `samconfig.toml`. They will be properly read back as double quotes.

specious:
I apologize for not having the time to adjust the test coverage right now. Could anyone kindly give me some pointers as to what needs to be done?

awood45:
It looks like the failure cause is just not having run `black` formatting. I'll try to pull this down and fix that.

awood45:
I can't push to your fork so I've opened #3533 with formatting fixes.

specious:
Fantastic!

awood45:
And merged! Thanks so much!

PR #3229, opened 2021-09-01
feat: Implementation for Plugin Interface CDK Read Method

mildaniel:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


wchengru (approved):

samcli/lib/iac/cfn/helpers.py
```diff
@@ -0,0 +1,14 @@
+"""
+CFN IaC helper functions
+"""
+from samcli.lib.iac.plugins_interfaces import Stack
+from samcli.lib.providers.sam_base_provider import SamBaseProvider
+from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
+
+
+def get_resolved_cfn_template(template, parameter_overrides, global_parameter_overrides) -> Stack:
```
torresxb1, line 9: nit but could we include the types for the parameters?

mildaniel: Yup, good call out.

samcli/lib/iac/cdk/cdk_iac.py
```diff
@@ -31,3 +115,229 @@ def update_packaged_locations(self, stack: Stack) -> bool:
...
+        assets = _collect_assets(ca_stack)
+        LOG.debug("Found assets: %s", str(assets))
+        asset_parameters = {
+            asset_param
+            for asset in assets.values()
+            for asset_param in asset.extra_details.get("assetParameters", {}).values()
+        }
+        sections = {}
+        options_map = self._context.command_options_map
+        template = get_resolved_cfn_template(
```
hawflau, line 185: I think it's neat to do the intrinsics resolving here. But how do we get the original template back? That's  something we need to revisit when we start to work on the `write_project` implementation.

mildaniel: I believe that topic is part of the planned discussion we have for `write_project`. There are a few options here, but I don't think this should be something difficult.

samcli/lib/iac/cfn/helpers.py
```diff
@@ -0,0 +1,18 @@
...
+CFN IaC helper functions
+"""
+from typing import Dict
+
+from samcli.lib.iac.plugins_interfaces import Stack
+from samcli.lib.providers.sam_base_provider import SamBaseProvider
+from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
+
+
+def get_resolved_cfn_template(
```
hawflau, line 11: Why do we need to define an extra function to just wrap `SamLocalStackProvider.merge_parameter_overrides(...)` and do nothing else?

hawflau, line 11: Why do we need to define an extra function to just wrap `SamLocalStackProvider.merge_parameter_overrides(...)` and do nothing else?

hawflau (changes requested):

samcli/lib/iac/cdk/cdk_iac.py
hawflau: sounds good. Just wanted to keep a note here as a reminder

samcli/lib/iac/cfn/helpers.py
mildaniel: Not really needed. I can remove this.

hawflau (approved):

PR #3232, opened 2021-09-01
fix: SAM Delete CFN Rate Limit

CoshUS:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
+++ b/samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
@@ -46,8 +46,8 @@ class CompanionStackManager:
         self._companion_stack = CompanionStack(stack_name)
         self._builder = CompanionStackBuilder(self._companion_stack)
         self._boto_config = Config(region_name=region if region else None)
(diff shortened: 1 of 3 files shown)


mgrandis (approved):

prenx4x (approved):

SimonCMoore:
It would have been great if you had explained how this fixed the problem. It looks like you increased retries to 3 times and waiting to 10 from 5 (seconds?). Do we know why/what is throttling us might there be a better way not to hit the limit than just retrying? Thanks

CoshUS:
> It would have been great if you had explained how this fixed the problem. It looks like you increased retries to 3 times and waiting to 10 from 5 (seconds?). Do we know why/what is throttling us might there be a better way not to hit the limit than just retrying? Thanks

Our canaries run with 5 concurrent jobs on a single account. The SAM delete calls describe stack from the waiter for every 5 seconds. This will cause the request per second to be around 1 which can cause throttling. The exact number of CFN maximum rps is not known as it's dynamic.
The default waiter timeout is 30 seconds which is what we will be using.

PR #3235, opened 2021-09-02
feat: Sam init interactive flow redesign

jonife:
The changes made in this pr was to take into account 2 things listed below:
1: **Add Sorting the list of dependency manager**: This was aimed at getting a consistent order in the list option for dependency manager, as against an ever changing order in the option for dependency while using only a set. This update was done in `samcli/commands/init/interactive_init_flow.py`

2: **Updated the Init Integration test:** Due to recent UX changes to SAM init, which affected the way the the init integration test previously simulated the user input and as such the integration test needed to be updated to simulate the current flow. This update was done in `tests/integration/init/schemas/test_init_with_schemas_command.py` and `tests/integration/init/schemas/schemas_test_data_setup.py`


mgrandis (approved):

jonife:
it should be 95% now

sriram-mv (approved):

CoshUS (approved):

Issue #2837, opened 2021-04-28
No response from invoke container for Lambda inside docker-compose


andrewx82:
### Description:

Related to https://github.com/aws/aws-sam-cli/issues/2492

When running a docker-compose with an image that is using sam to run a local api, the api can never get a response from the lambda container. Everything works correctly when run locally not in a container.

I have read through all the answers in 2492 and tried the various permutations suggested, nothing has worked. There appeared to be no subsequent documentation follow up for getting this setup to work with docker.

### Steps to reproduce:

DockerFile

```
FROM python:alpine
RUN apk add --no-cache --virtual build-deps build-base=0.5-r2 gcc=10.2.1_pre1-r3 bash=5.1.0-r0 && \
    pip install aws-sam-cli==1.23.0 && \
    apk del build-deps
COPY ./sam_entrypoint.sh /bin/sam_entrypoint.sh
RUN chmod +x /bin/sam_entrypoint.sh

WORKDIR /app
COPY ./node_modules /app/node_modules
COPY ./package.json /app/package.json
COPY ./template.yml /app/template.yml
COPY ./dist /app/dist
EXPOSE 3000 
ENTRYPOINT ["sh","/bin/sam_entrypoint.sh"]
```

sam_entrypoint.sh
```
#!/bin/bash
set -o errexit
/usr/local/bin/sam local start-api --debug \
  --template ./template.yml \
  --host 0.0.0.0 \
  --container-host host.docker.internal \
  --container-host-interface  127.0.0.1 \
  --docker-network host
````

Template.yml
```
AWSTemplateFormatVersion: "2010-09-09"
Transform: "AWS::Serverless-2016-10-31"
Description: "Automated lambda setup for api gateway"
Resources:
  res0:
    Type: "AWS::Serverless::Function"
    Properties:
      CodeUri: "./"
      Handler: "dist/index.handler"
      Runtime: "nodejs12.x"
      Timeout: 10
      Events:
        CatchAll:
          Type: "Api"
          Properties:
            Path: "/"
            Method: ANY
  res1:
    Type: "AWS::Serverless::Function"
    Properties:
      CodeUri: "./"
      Handler: "dist/index.handler"
      Runtime: "nodejs12.x"
      Timeout: 10
      Events:
        CatchAll:
          Type: "Api"
          Properties:
            Path: "/{proxy+}"
            Method: ANY
```

Docker-compose 
```
version: '3.6'
services:
  sam_app:
    build: 
      context: .
[...]

PR #3237, opened 2021-09-02
Add root_template_path to nested stack

xazhao:
https://github.com/aws/aws-sam-cli/issues/2837

The current behavior is if `CodeUri` is a relative path, we convert it to absolute path when we normalize resource path for nested stacks. We will also try to resolve relative code path to absolute code path when we get invoke config using `--docker-volume-basedir` option. There is a bug here. If a customer wants to resolve the path using that option, it won't work since CodeUri is already an absolute path so it won't trying to resolve again.
To fix this issue, we need to make sure if CodeUri is relative path, we keep it relative path so that it can be resolved using `--docker-volume-basedir` properly. `root_template_path` is introduced to make sure relative path is based on root template.

Keep relative path when we normalize resource path.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/_utils/template.py
```diff
@@ -29,20 +29,26 @@ class TemplateFailedParsingException(UserException):
     pass
 
 
-def get_template_data(template_file):
+def get_template_data(template_file, root_template_path: str = ""):
```
aahung, line 32: do we expect `root_template_path` to be `None` or `""`?

samcli/commands/_utils/template.py
```diff
@@ -29,20 +29,26 @@ class TemplateFailedParsingException(UserException):
...
+        Path of the root SAM Template
 
     Returns
     -------
     Template data as a dictionary
     """
 
+    # if template_file is a relative path , we use root_template_path to convert to absolute path
+    if not os.path.isabs(template_file) and root_template_path:
+        template_file = os.path.join(root_template_path, template_file)
```
aahung, line 50: It looks to me that we need to get dirname from `root_template_path` before joining

aahung (commented):
Can we add some tests involving `docker-volume-basedir`?

samcli/commands/_utils/template.py
xazhao: I think either is fine. Just need to keep it consistent in other places. My understanding is if it is string, we prefer using `""`.

samcli/commands/_utils/template.py
xazhao: Actually the name is misleading. I just updated it to `root_template_dir`

samcli/commands/_utils/template.py
xazhao: Updated it to None since it makes more sense.

xazhao:
Added an integration test involving using `--docker-volume-basedir` option.

samcli/commands/_utils/template.py
```diff
@@ -29,20 +30,26 @@ class TemplateFailedParsingException(UserException):
...
+        Optional directory of the root SAM Template
 
     Returns
     -------
     Template data as a dictionary
     """
 
+    # if template_file is a relative path , we use root_template_dir to convert to absolute path
+    if not os.path.isabs(template_file) and root_template_dir:
+        template_file = os.path.join(root_template_dir, template_file)
```
hoffa, line 51: Is there any way to do this without requiring multiple variables to specify a single path?

e.g. in one location make the path absolute, and pass that around, so that callees just read from the path without caring whether it's absolute or relative

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -212,15 +226,16 @@ def get_stacks(
...
+            Optional directory of root SAM Template. By default it is None unless we pass in a value.
         Returns
         -------
         stacks: List[Stack]
             The list of stacks extracted from template_file
         remote_stack_full_paths : List[str]
             The list of full paths of detected remote stacks
         """
-        template_dict = get_template_data(template_file)
+        template_dict = get_template_data(template_file, root_template_dir)
```
aahung, line 238: Do we want to provide `root_template_dir` for all usage of `get_template_data()`? Maybe we can make it not optional

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1082,5 +1082,8 @@ def test_must_pass_custom_region(self, get_stacks_mock):
         invoke_context = InvokeContext("template_file", aws_region="my-custom-region")
         invoke_context._get_stacks()
         get_stacks_mock.assert_called_with(
-            "template_file", parameter_overrides=None, global_parameter_overrides={"AWS::Region": "my-custom-region"}
+            "template_file",
+            parameter_overrides=None,
+            root_template_dir=os.getcwd(),
+            global_parameter_overrides={"AWS::Region": "my-custom-region"},
```
aahung, line 1088: Can you put a comment at line 1087 explaining a little bit why it is `os.getcwd()`?

aahung (approved):
LGTM

samcli/commands/_utils/template.py
xazhao: I think that is possible and useful. It needs a relative larger refactoring since currently we have a few places doing path checking and converting. I'd say it's out of scope for this bug and we can do this in a separate PR.

tests/unit/commands/local/cli_common/test_invoke_context.py
xazhao: Sure!

samcli/lib/providers/sam_stack_provider.py
xazhao: Currently it is used in a few other places including intrinsics resolver. I prefer not changing that part. Ideally as Chris said, we can get rid of all `root_template_dir` and convert all paths to absolute in a single place later.

samcli/commands/_utils/template.py
hoffa: Any way to reduce the number of places where these path modifications are done?

It's already somewhat confusing as for example `normalize_resource_path` already does some path changes, but then further changes are done here.

hoffa: Discussed elsewhere: `normalize_resource_path` no longer converts to absolute path only since a bug was discovered. Agreed to add comment explaining the context.

hoffa (approved):

PR #2935, opened 2021-06-06
fix: Skip build of Docker image if ImageUri is a valid ECR URL (#2934)

alexisfacques:
#### Which issue(s) does this change fix?

#2934 

#### Why is this change necessary?

- Support deployment of `PackageType: Image` lambda functions of which the underlying Docker image is built/deployed externally to SAM.

#### How does it address the issue?

- Checks whether or not the `ImageUri` property of a `AWS::Serverless::Function`, or `Code.ImageUri` property of a `AWS::Lambda::Function` is a valid ECR URL, based on the resource `PackageType`. If so, skips the local image build and prompts the user with a proper warning, similarly to `Zip` packaged lambdas.

```
  Function:
    Type: AWS::Serverless::Function
    Properties:
      PackageType: Image
      ImageUri: 123456789012.dkr.ecr.eu-west-1.amazonaws.com/function:latest
```

```
% sam build
The resource AWS::Serverless::Function 'Function' has specified ECR registry image for ImageUri. It will not be built and SAM CLI does not support invoking it locally.
```

#### What side effects does this change have?

TBD

#### Checklist

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [x] Write unit tests
- [x] `make pr` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #3272, opened 2021-09-14
fix: Revert the behavior when both ImageUri and Metadata exist for im…

### LINKED ISSUE #3239: `sam build` broken after upgrading SAM CLI when using ECR (backward compatibility issue)
### Description:
I have a SAM application with an AWS ECR repository configured as the storage for the app snapshots/releases.

So far I was using just `sam build` and `sam deploy` to deploy my SAM application. It used to work with SAM CLI 1.24.1.

Unfortunately after I upgraded to SAM CLI 1.27.2 it stopped working. Now `sam build` says:

> The resource AWS::Serverless::Function 'HelloFunction' has specified ECR registry image for ImageUri. It will not be built and SAM CLI does not support invoking it locally.

Upgrading SAM CLI to the current latest version 1.30.0 doesn't help either.

**This is a breaking change.** Not sure if it was announced anywhere. Also it seems the problem was introduced as part of #2934.

Honestly I don't know what I should do now and how I should fix it. For now I can stick to the SAM CLI version that works for me but this is only a temporary solution.

### Steps to reproduce:
Create a SAM app with a template including:
```
Resources:
  HelloFunction:
    Properties:
      PackageType: Image
      ImageUri: <account-number>.dkr.ecr.<region>.amazonaws.com/hello-app
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./app
      DockerTag: python3.8-v1
```

Use `sam build` and then guided deploy with SAM CLI 1.24.1. Then upgrade to SAM CLI 1.27.2 - `sam build` will stop working.

### Observed result:
`sam build` doesn't build the image anymore and `sam deploy` doesn't push the image to ECR. As a result `sam deploy` fails.

### Expected result:
`sam build` should build the image, as it used to. `sam deploy` should push the image to ECR and update the app without issues.

1. OS: Ubuntu 20.04
2. `sam --version`: 1.24.1, 1.27.2, 1.30.0
3. AWS region: Ireland (eu-west-1)

aahung:
…age func

Fix #3239 by breaking change #2935
When both ImageUri and Metadata are provided, previously we choose to give "Metadata" the priority and build it. But now we skip it because ImageUri is provided.

Fix the logics so that function is only skipped when ImageUri is provided and related fields cannot be found in Metadata. Previous behavior is partially recovered.

N/A

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/providers/sam_function_provider.py
+++ b/samcli/lib/providers/sam_function_provider.py
@@ -144,8 +144,12 @@ class SamFunctionProvider(SamBaseProvider):
                             SamFunctionProvider._warn_code_extraction(resource_type, name, code_property_key)
                         continue
 
-                    if resource_package_type == IMAGE and SamBaseProvider._is_ecr_uri(
-                        resource_properties.get(image_property_key)
+                    if (
+                        resource_package_type == IMAGE
+                        and SamBaseProvider._is_ecr_uri(resource_properties.get(image_property_key))
+                        and not SamFunctionProvider._metadata_has_necessary_entries_for_image_function_to_be_built(
+                            resource_metadata
+                        )
                     ):
                         # ImageUri can be an ECR uri, which is not supported
                         if not ignore_code_extraction_warnings:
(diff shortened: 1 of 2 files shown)


awood45 (approved):
This looks well-formed, let's make sure another lead agrees that this is the behavior we want going forward.

mndeveci (approved):
LGTM, thanks for fixing this issue 👏🏼

hoffa (approved):

sriram-mv (approved):

PR #3277, opened 2021-09-14
fix: Update Hello World template to match the ones in the templates repository

jonife:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mgrandis (approved):

samcli/lib/init/templates/cookiecutter-aws-sam-hello-dotnet/cookiecutter.json
```diff
@@ -1,4 +1,4 @@
 {
     "project_name": "Name of the project",
-    "runtime": "dotnetcore2.1"
+    "runtime": "dotnetcore3.1"
```
mndeveci, line 3: Should we keep this as `dotnetcore2.1` for backwards compatibility?

samcli/lib/init/templates/cookiecutter-aws-sam-hello-nodejs/cookiecutter.json
```diff
@@ -1,4 +1,4 @@
 {
     "project_name": "Name of the project",
-    "runtime": "nodejs12.x"
+    "runtime": "nodejs14.x"
```
mndeveci, line 3: Same as the other one, should we keep original version inside SAM CLI?

mndeveci (commented):
Thanks for syncing this with external repo!

It looks good in general, I have questions about bumping dotnetcore and nodejs versions

samcli/lib/init/templates/cookiecutter-aws-sam-hello-dotnet/cookiecutter.json
jfuss: `dotnetcore2.1` is no longer supported by Lambda. https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

samcli/lib/init/templates/cookiecutter-aws-sam-hello-nodejs/cookiecutter.json
jonife: So with the current INIT design we default to the latest version of the runtime selected by the customer for hello world templates however customers can still specify the runtime version they would like to use. So defaulting to node14.x as against nodejs12.x was necessary and after working on this project I don't see how this affects backward compatibility.

Issue #3205, opened 2021-08-25
Running sam local start-api has a different NODE_PATH when debugging or not debugging


jondmcelroy:
### Description:
We  were requiring our node js code based on an absolute path. This worked fine when running `sam local start-api`.

For example we have a folder at the root level (same at the template.yml file) called graphql with a js file inside called context.js.
At the handler level we were just calling `require('graphql/context')` and that would work just fine. 

When I run `sam local start-api -d` and attach, it would fail saying: `Error: Cannot find module`, etc...

This lead me down the path of issues like: 
https://github.com/aws/aws-sam-cli/issues/1972 and https://github.com/aws/aws-sam-cli/issues/1246

If I change the require to `require('./context')` it works fine in both debug and regular, but that is strange that there is a difference.

Eventually, I realized that the NODE_PATH was different depending on when you run it with debug or without debug.

NODE_PATH without Debug:
/opt/nodejs/node14/node_modules:/opt/nodejs/node_modules:/var/runtime/node_modules:/var/runtime:/var/task

NODE_PATH with Debug:
/opt/nodejs/node_modules:/opt/nodejs/node14/node_modules:/var/runtime/node_modules

### Steps to reproduce:
This can be checked on any lambda.
At the top of your handler js file: `console.log(process.env.NODE_PATH);`

Run the normal `sam local start-api` and look at the NODE_PATH.
Run the `sam local start-api -d` and look at the NODE_PATH.

They will be different.

IT seems like they shouldn't be different and I wonder if its due to the NODE_PATH being hardcoded in debug mode: https://github.com/aws/aws-sam-cli/pull/1646

### Observed result:
NODE_PATH without Debug:
/opt/nodejs/node14/node_modules:/opt/nodejs/node_modules:/var/runtime/node_modules:/var/runtime:/var/task

NODE_PATH with Debug:
/opt/nodejs/node_modules:/opt/nodejs/node14/node_modules:/var/runtime/node_modules

### Expected result:
They should be the same. The debug should have the `/var/runtime:/var/task` path in the NODE_PATH.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MAC 11.5.2
2. `sam --version`: 1.29.0
3. AWS region: us-west-2

prenx4x:
Hello,

Thank you for bringing this issue to us. We are investigating and working on a fix for it.

PR #3279, opened 2021-09-15
fix: Nodejs debug paths different than non debug

prenx4x:
#3205 

To fix difference in node paths between debug and non debug

Added missing path entries in proper settings object.

None

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_debug_settings.py
+++ b/samcli/local/docker/lambda_debug_settings.py
@@ -108,7 +108,8 @@ class LambdaDebugSettings:
                 + ["--no-lazy", "--expose-gc"]
                 + ["/var/runtime/index.js"],
                 container_env_vars={
-                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node10/node_modules:/var/runtime/node_modules",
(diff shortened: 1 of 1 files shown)


samcli/local/docker/lambda_debug_settings.py
```diff
@@ -108,7 +108,8 @@ def get_debug_settings(debug_port, debug_args_list, _container_env_vars, runtime
                 + ["--no-lazy", "--expose-gc"]
                 + ["/var/runtime/index.js"],
                 container_env_vars={
-                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node10/node_modules:/var/runtime/node_modules",
+                    "NODE_PATH": "/opt/nodejs/node_modules:/opt/nodejs/node10/node_modules:/var/runtime/node_modules:"
+                    "/var/runtime:/var/task",
```
hoffa, line 112: Just curious, is this a pylint rule that requires splitting strings on separate lines?

hoffa (approved):

samcli/local/docker/lambda_debug_settings.py
prenx4x: yes

CoshUS (approved):

PR #3312, opened 2021-09-28
chore: Changed hashing method for build definitions

qingchm:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/sync/command.py
+++ b/samcli/commands/sync/command.py
@@ -53,6 +53,7 @@ Update/sync local artifacts to AWS
 SHORT_HELP = "Sync a project to AWS"
 
 DEFAULT_TEMPLATE_NAME = "template.yaml"
+DEFAULT_CAPABILITIES = ("CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND")
 
 
 @click.command("sync", help=HELP_TEXT, short_help=SHORT_HELP)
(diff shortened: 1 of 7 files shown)


samcli/commands/sync/command.py
```diff
@@ -197,6 +197,9 @@ def do_cli(
     click.echo("\t\tA different default S3 bucket can be set in samconfig.toml")
     click.echo("\t\tOr by specifying --s3-bucket explicitly.")
 
+    click.echo("\n\t\tDefault capabilities applied: CAPABILITY_NAMED_IAM, CAPABILITY_AUTO_EXPAND")
```
mndeveci, line 200: nit: we can move default capabilities into a static var and use it here and at line 99, so that we don't need to change 2 places in the future.

samcli/lib/utils/hash.py
```diff
@@ -41,22 +41,28 @@ def file_checksum(file_name: str, hash_generator: Any = None) -> str:
...
-    md5 checksum of the directory.
+    checksum hash of the directory.
 
     """
     ignore_set = set(ignore_list or [])
-    md5_dir = hashlib.md5()
+    if hashing_method == "sha256":
+        hash_dir = hashlib.sha256()
+    else:
+        hash_dir = hashlib.md5()
```
mndeveci, line 65: Can we use the same way which is used in `file_checksum` method (check `hash_generator` parameter)? Instead of passing the string of the method, we can pass the method itself.

mndeveci (approved):

CoshUS (approved):

PR #3318, opened 2021-09-29
[Auto Dependencies Layer] Combination flag

mingkun2020:
add combine_dependencies to app builder


samcli/lib/build/app_builder.py
```diff
@@ -634,6 +645,8 @@ def _build_function_in_process(
         if lambda_builders_version == "1.8.0":
```
mndeveci, line 645: Can we also change this to `1.9.0` for now?

mingkun2020: What's purpose for this if statement?

mndeveci: We need this until we release new version of lambda builders. Otherwise pylint fail because these new parameters are not available there.

mingkun2020: Updated.

CoshUS (approved):
LGTM, can you take a look at the failing unit tests?

mndeveci (approved):

PR #3325, opened 2021-09-30
Add disable rollback option

torresxb1:
CloudFormation recently launched a new feature that allows users to specify a DisableRollback parameter for the CreateStack, UpdateStack, and ExecuteChangeSet actions. When users pass this parameter and there’s a stack failure, they can preserve the state of successfully provisioned resources and work on the template from the point of failure. Users can also call the UpdateStack and ExecuteChangeSet on stacks that are in the CREATE_FAILED or UPDATE_FAILED states.

More information can be found in this blog post: https://aws.amazon.com/blogs/aws/new-for-aws-cloudformation-quickly-retry-stack-operations-from-the-point-of-failure/

This PR adds support for this new feature in SAM CLI by adding a `--disable-rollback` option when calling `sam deploy`. If users enable this option and experience a failure while deploying, they are now able to do any fixes required and then deploy again, continuing from the point of failure while preserving the successfully created resources from the previous deploy. 

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.1
 Flask~=1.1.2
 #Need to add Schemas latest SDK.
-boto3~=1.14
+boto3~=1.18
 jmespath~=0.10.0
 PyYAML~=5.3
 cookiecutter~=1.7.2
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -9,7 +9,7 @@ pytest-cov==2.10.1
 # here we fix its version and upgrade it manually in the future
 mypy==0.790
 # 1.17.91 has a bug, https://github.com/vemel/mypy_boto3_builder/pull/82
-boto3-stubs[essential]==1.17.90.post1
+boto3-stubs[essential]~=1.18
 
 # Test requirements
 pytest==6.1.1
--- a/samcli/commands/deploy/guided_context.py
+++ b/samcli/commands/deploy/guided_context.py
@@ -59,6 +59,7 @@ class GuidedContext:
         config_section=None,
         config_env=None,
         config_file=None,
+        disable_rollback=None,
     ):
         self.template_file = template_file
         self.stack_name = stack_name
@@ -89,6 +90,7 @@ class GuidedContext:
         self.end_bold = "\033[0m"
         self.color = Colored()
         self.function_provider = None
+        self.disable_rollback = disable_rollback
 
     @property
     def guided_capabilities(self):
@@ -146,6 +148,9 @@ class GuidedContext:
             f"\t{self.start_bold}Allow SAM CLI IAM role creation{self.end_bold}", default=True
         )
 
+        click.secho("\t#Preserves the state of previously provisioned resources when an operation fails")
+        disable_rollback = confirm(f"\t{self.start_bold}Disable rollback{self.end_bold}", default=self.disable_rollback)
+
         if not capabilities_confirm:
             input_capabilities = prompt(
                 f"\t{self.start_bold}Capabilities{self.end_bold}",
@@ -194,6 +199,7 @@ class GuidedContext:
         self.config_env = config_env if config_env else default_config_env
         self.config_file = config_file if config_file else default_config_file
         self.confirm_changeset = confirm_changeset
+        self.disable_rollback = disable_rollback
 
     def prompt_authorization(self, stacks: List[Stack]):
         auth_required_per_resource = auth_per_resource(stacks)
@@ -560,6 +566,7 @@ class GuidedContext:
                 confirm_changeset=self.confirm_changeset,
                 capabilities=self._capabilities,
                 signing_profiles=self.signing_profiles,
+                disable_rollback=self.disable_rollback,
             )
 
     @staticmethod
--- a/samcli/commands/deploy/guided_context.py
+++ b/samcli/commands/deploy/guided_context.py
@@ -148,15 +148,15 @@ class GuidedContext:
             f"\t{self.start_bold}Allow SAM CLI IAM role creation{self.end_bold}", default=True
         )
 
-        click.secho("\t#Preserves the state of previously provisioned resources when an operation fails")
-        disable_rollback = confirm(f"\t{self.start_bold}Disable rollback{self.end_bold}", default=self.disable_rollback)
-
         if not capabilities_confirm:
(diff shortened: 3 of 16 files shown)


requirements/base.txt
```diff
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.1
 Flask~=1.1.2
 #Need to add Schemas latest SDK.
-boto3~=1.14
+boto3~=1.18
```
torresxb1, line 5: Haven't experienced any conflicts

requirements/dev.txt
```diff
@@ -8,8 +8,7 @@ pytest-cov==2.10.1
 # mypy adds new rules in new minor versions, which could cause our PR check to fail
 # here we fix its version and upgrade it manually in the future
 mypy==0.790
-# 1.17.91 has a bug, https://github.com/vemel/mypy_boto3_builder/pull/82
-boto3-stubs[essential]==1.17.90.post1
+boto3-stubs[essential]~=1.18
```
torresxb1, line 11: I also updated this one. Let me know if this should be different from `base.txt`

samcli/commands/deploy/guided_context.py
```diff
@@ -153,6 +156,9 @@ def guided_prompts(self, parameter_override_keys):
                 type=FuncParamType(func=_space_separated_list_func_type),
             )
 
+        click.secho("\t#Preserves the state of previously provisioned resources when an operation fails")
+        disable_rollback = confirm(f"\t{self.start_bold}Disable rollback{self.end_bold}", default=self.disable_rollback)
+
```
torresxb1, line 161: Is the order here okay? After asking for IAM role creation

samcli/commands/deploy/utils.py
```diff
@@ -63,6 +66,7 @@ def print_deploy_args(
     click.echo(f"\tStack name                   : {stack_name}")
     click.echo(f"\tRegion                       : {region}")
     click.echo(f"\tConfirm changeset            : {confirm_changeset}")
+    click.echo(f"\tDisable rollback             : {disable_rollback}")
```
torresxb1, line 69: Another question about order. This time for the table printed after initiating the deployment. Is it okay to print this value after confirm changeset?

samcli/lib/deploy/deployer.py
```diff
@@ -403,9 +407,9 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
 
     @staticmethod
     def _check_stack_complete(status: str) -> bool:
-        return "COMPLETE" in status and "CLEANUP" not in status
+        return "IN_PROGRESS" not in status
```
torresxb1, line 410: This one is easy to miss but it's important. When `disable-rollback` was specified and the deployment failed, the process stayed hanging because `UPDATE_FAILED` or `CREATE_FAILED` don't have `COMPLETE` in them. I'm not sure why it would be considered that it's not complete or still in progress if it has already failed creating/updating, so I just updated this. Would there be any problem doing this (was there a reason why update/create failed were not considered complete)?

tests/integration/deploy/test_deploy_command.py
```diff
@@ -929,6 +929,245 @@ def test_deploy_nested_stacks(self, template_file):
...
+            bytes(
+                f"Error: Failed to create/update the stack: {stack_name}, Waiter StackCreateComplete failed: "
+                f'Waiter encountered a terminal failure state: For expression "Stacks[].StackStatus" '
+                f'we matched expected path: "ROLLBACK_COMPLETE" at least once',
+                encoding="utf-8",
+            ),
+            stderr,
+        )
+
+    @skip("skip until cloudformation bug is fixed")
```
torresxb1, line 969: once cloudformation's bug is fixed, we can just remove this line and hopefully it should just work

torresxb1:
Not exactly sure why, but it seems like bumping the boto3 version is causing the build to fail because of mypy type checks for files I didn't touch. Will look into it.

requirements/base.txt
wchengru: Why do we want to upgrade boto3?

torresxb1: `DisableRollback` parameter hadn't been added to `execute_change_set` in version 1.14:
https://boto3.amazonaws.com/v1/documentation/api/1.14.0/reference/services/cloudformation.html?highlight=execute_change_set#CloudFormation.Client.execute_change_set

mildaniel: Can we confirm what the past issues were that we had with bumping the boto version and make sure we aren't seeing those?

torresxb1: I'm not sure what issues we were experiencing before, but I'll ask Mehment. There's a comment about a bug in  `dev.txt`, but I think that has been resolved.

hawflau (changes requested):

samcli/lib/deploy/deployer.py
```diff
@@ -501,3 +510,14 @@ def get_stack_outputs(self, stack_name, echo=True):
 
         except botocore.exceptions.ClientError as ex:
             raise DeployStackOutPutFailedError(stack_name=stack_name, msg=str(ex)) from ex
+
+    @staticmethod
+    def _gen_deploy_failed_with_rollback_disabled_msg(stack_name):
+        return """\nFailed to deploy stack with automatic rollback disabled\n
+Actions you can take next
+=========================
+[*] Fix issues and try deploying again
```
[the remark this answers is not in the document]
torresxb1: Nope

samcli/lib/deploy/deployer.py
```diff
@@ -403,9 +407,9 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
 
     @staticmethod
     def _check_stack_complete(status: str) -> bool:
```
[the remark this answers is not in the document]
torresxb1: yeah if there's no issue with the changes I made below (left a comment about it), then I think we could either change it to your suggestion or something like `_check_stack_not_in_progress`

samcli/lib/deploy/deployer.py
```diff
@@ -22,6 +22,7 @@
 import time
 from datetime import datetime
 from typing import Dict, List
+import click
```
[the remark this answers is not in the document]
torresxb1: I used it in line 452

Issue #1460, opened 2019-10-12
allow origin isn't sufficient to resolve CORS issue


kimyu92:
### Description
Adding allow origin and allow headers at template.yaml and returning `Access-Control-Allow-Origin` in the lambda response body seems insufficient to resolve cross origin issue after deploying the lambda service.

### Steps to reproduce
sample of `template.yaml`
```yaml
Globals:
  Function:
    Timeout: 60
  Api:
    Cors:
      AllowMethods: "'POST,OPTIONS'"
      AllowHeaders: "'*'"
      AllowOrigin: "'*'"
      AllowCredentials: "'*'"

Resources:
  EndpointFunction:
    Type: AWS::Serverless::Function
      CodeUri: api/v1/
      Handler: endpoint.lambda_handler
      Runtime: ruby2.5
      Events:
        Endpoint:
          Type: Api
          Properties:
            Path: /api/v1/endpoint
            Method: post
```

sample of `endpoint.rb`
```ruby
def lambda_handler(event:, context:)
  {
    statusCode: 201,
    headers: {
      'Access-Control-Allow-Origin' => '*',
      'Content-Type' => 'application/json'
    }
  }
end
```
### Observed result

```
Access to XMLHttpRequest at 'https://<random-sha>.execute-api.<aws-region>amazonaws.com/prod/api/v1/endpoint' from origin 'https://<domain>' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

**Update**:
This turned out to be a url of lambda service is being case sensitive. Aside from that, there is currently no way to set multiple domains for allow origin. Also, the hassle way of defining an option function doesn't seem to work either at least local #1434

### Expected result

able to make request from the browser

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. `sam --version`: 0.22.0

adhamshandle:
@ZHShang how to downgrade in mac?

jfuss:
@kimyu92 

>  in the lambda response body seems insufficient to resolve cross origin issue after deploying the lambda service.

Are you seeing CORS issues with the Lambda Service or locally with SAM CLI? It sounds like this is an issue you are having with the Service.

> Aside from that, there is currently no way to set multiple domains for allow origin.
This is a limitation of API Gateway: https://stackoverflow.com/a/39628785

> Also, the hassle way of defining an option function doesn't seem to work either at least local #1434

[...]

PR #3357, opened 2021-10-12
updates the local API service to include CORS headers

jlapier:
Previous to this commit, `sam local api` would only return CORS headers when an OPTIONS request is made. This causes issues for browsers, which expect at the very least "Access-Control-Allow-Origin" to be present on other HTTP methods. This commit adds the "Access-Control-Allow-Origin" header to a request if it is made with one of the Allowed-Methods.

References:
 * Fetch spec on CORS: https://fetch.spec.whatwg.org/#cors-request
 * Examples from Mozilla: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
 * API Gateway doc on when to supply CORS headers: https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html

#1460 

In order to test a local UI against a local API via sam, proper CORS headers are required or the browser will reject the requests.

Adds the Access-Control-Allow-Origin header if a CORS configuration is present for the API.

The CORS header behavior is not entirely predictable for REST APIs, according to the [AWS API Gateway docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html). This change may mask a need to include the headers at the integration point. However, for [HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-cors.html), the behavior is much more predicable and this change is consistent with that behavior.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


vmarinescu (approved):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -375,8 +375,20 @@ def _request_handler(self, **kwargs):
...
         return self.service_response(body, headers, status_code)
 
+    @staticmethod
+    def _add_cors_headers_to_response(method, response_headers, cors_headers):
+        # If the CORS headers are present and one of the allowed methods is being used,
+        # add the allow-origin header to the response.
+        # On Origin other than the wildcard (*), include Vary: Origin
+        allow_origin = cors_headers.get("Access-Control-Allow-Origin")
+        if method in cors_headers.get("Access-Control-Allow-Methods", []) and allow_origin:
+            response_headers.add_header("Access-Control-Allow-Origin", allow_origin)
```
jfuss, line 388: According to the API Gateway docs, the Lambda function is responsible for adding these integrations: https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html (in the "Enabling CORS support for Lambda or HTTP proxy integrations" section). So SAM CLI should not be doing this.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -375,8 +375,20 @@ def _request_handler(self, **kwargs):
...
+            if allow_origin != "*":
+                response_headers.add_header("Vary", "Origin")
```
jfuss, line 390: Does API Gateway add this?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -768,7 +780,7 @@ def _construct_v_2_0_event_http(
             # Flask does not parse/decode the request data. We should do it ourselves
             request_data = request_data.decode("utf-8")
 
-        query_string_dict, _ = LocalApigwService._query_string_params(flask_request)
+        query_string_dict = LocalApigwService._query_string_params_v_2_0(flask_request)
```
jfuss, line 783: I don't understand how this relates to the CORS headers.

jfuss (changes requested):

jfuss:
Closing given no response. Feel free to address comments and re-open and we will re-review.

Issue #3194, opened 2021-08-23
Multiple queryStringParameters ignored in sam local start-api for HTTP APIs


EpicSquid:
### Description:

I have a simple lambda function which takes multiple query string parameters, however only the last query string parameter is kept. 

### Steps to reproduce:

I have defined a function as follows:
```yaml  
TestFunction:
    Type: AWS::Serverless::Function 
    Properties:
      CodeUri: rest_api/test/
      Events:
        GetTest:
          Type: HttpApi
          Properties:
            Path: /api/test
            Method: get
 ```
 
 Then run the application with `sam local start-api --debug`
 
 Then sent a query to it using postman as a GET request: http://localhost:3000/api/test?test_ids=1&test_ids=2

### Observed result:

The event output for the function is the following
```json
Constructed String representation of Event Version 2.0 to invoke Lambda. Event: 
{
  "version": "2.0",
  "routeKey": "GET /api/test",
  "rawPath": "/api/test",
  "rawQueryString": "test_ids=1&test_ids=2",
  "cookies": [],
  "headers": {
    "Content-Type": "application/json",
    "User-Agent": "PostmanRuntime/7.28.3",
    "Accept": "*/*",
    "Postman-Token": "cb0df40e-920d-4f8d-bbf8-9381f9ee5e45",
    "Host": "localhost:3000",
    "Accept-Encoding": "gzip, deflate, br",
    "Connection": "keep-alive",
    "Content-Length": "21",
    "X-Forwarded-Proto": "http",
    "X-Forwarded-Port": "3000"
  },
  "queryStringParameters": {
    "test_ids": "2"
  },
  "requestContext": {
    "accountId": "123456789012",
    "apiId": "1234567890",
    "http": {
      "method": "GET",
      "path": "/api/test",
      "protocol": "HTTP/1.1",
      "sourceIp": "127.0.0.1",
      "userAgent": "Custom User Agent String"
    },
    "requestId": "46b22e74-8e33-4f90-8f5c-3af887356c8e",
    "routeKey": "GET /api/test",
    "stage": null
  },
  "body": "",
  "pathParameters": {},
  "stageVariables": null,
  "isBase64Encoded": false
}
```

Critically:
```json
"queryStringParameters": {
    "test_ids": "2"
  },
```

### Expected result:

I expected the following parsing for the queryStringParameters, as per the documentation
```json
"queryStringParameters": {
    "test_ids": "1,2"
  },
```

[...]

PR #3365, opened 2021-10-15
fixes multivalue query param requests for HTTP APIs

jlapier:
When a query parameter is specified multiple times in a URL, e.g.
`path?mykey=1&mykey=2&mykey=3`
this should create an event with all appropriate values.

Prior to this commit, when using `sam local start-api` with an HTTP API, only the first value is present in the `queryStringParameters` of the event. In [payload format v2](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) used in HTTP APIs, multiple values should be combined with commas and included in the `queryStringParameters` field. This commit fixes the behavior for query string parameters.

#3194

The local API does not match the behavior of an HTTP API in API Gateway.

It formats multivalue query parameters per the payload format version 2.

Previously truncated query parameter values will be included in the queryStringParameters field of an event.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -825,6 +825,36 @@ def _query_string_params(flask_request):
...
+        -------
+            Empty dict if no query params where in the request otherwise returns a dictionary of key to value
+
+        """
+        query_string_dict = {}
+
+        # Flask returns an ImmutableMultiDict so convert to a dictionary that becomes
+        # a dict(str: list) then iterate over
+        for query_string_key, query_string_list in flask_request.args.lists():
+            query_string_value_length = len(query_string_list)
```
hawflau, line 848: It would be better to just keep the last line of L848-L854
```
query_string_dict[query_string_key] = ",".join(query_string_list)
```

If `query_string_list` is an empty list, `",".join(query_string_list)` will just return an empty string

hawflau: or make the entire function even shorter:
```
...
    query_string_dict = {
        query_string_key: ",".join(query_string_list)
        for query_string_key, query_string_list in flask_request.args.lists()
    }
```

hawflau, line 848: it would be better to just keep the last line of L848-L854, i.e.
```
query_string_dict[query_string_key] = ",".join(query_string_list)
```
`",".join(query_string_list)` returns an empty string if `query_string_list` is an empty list.

Or, we could make the entire function even shorter:
```
query_string_dict = {
    query_string_key: ",".join(query_string_list)
    for query_string_key, query_string_list in flask_request.args.lists()
}
return query_string_dict
```

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -1481,7 +1481,7 @@ def setUp(self):
         self.request_mock.get_data.return_value = b"DATA!!!!"
         self.request_mock.mimetype = "application/json"
         query_param_args_mock = Mock()
-        query_param_args_mock.lists.return_value = {"query": ["params"]}.items()
+        query_param_args_mock.lists.return_value = {"query": ["param1","param2"]}.items()
```
hawflau, line 1484: you can run `make black` to fix the formatting issue here

hawflau (changes requested):
Thanks for the contribution! Just left two comments. Can you also please check the [check failures](https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/41164128) and fix the merge conflict?

hawflau:
Closing due to no response. I re-submitted a new one (#3788) to resolve merge conflicts, apply recommendations and add one more test.

jlapier:
Thanks for following up on this @hawflau! My apologies for being unable to find the time to respond to feedback.

PR #3410, opened 2021-10-26
feat: SAM Init interactive flow redesign

jonife:
Improve the user experience on the sam init interactive experience. Today we "hide" the types of app we support through a layer of questions (custom or pre made examples, Package type for lambda, what runtime, then show apps). While this works, we are moving to the idea of surfacing what examples we have first. 

The new flow is something like: 

custom or pre-made ---> here are the types of examples we have ----> from that list here are the package types ---> from that here are the runtimes. 

Improve the user experience on the sam init interactive experience

Changed the entire cli interactive flow

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes


samcli/commands/init/__init__.py
```diff
@@ -286,9 +286,9 @@ def do_cli(
     image_bool = name and pt_explicit and base_image
     if location or zip_bool or image_bool:
         # need to turn app_template into a location before we generate
-        templates = InitTemplates(no_interactive)
+        templates = InitTemplates()
```
mndeveci, line 289: Why we are not passing `no_interactive` flag anymore?

samcli/commands/init/__init__.py
```diff
@@ -286,9 +286,9 @@ def do_cli(
...
         if package_type == IMAGE and image_bool:
-            base_image, runtime = _get_runtime_from_image(base_image)
+            runtime = _get_runtime_from_image(base_image)
```
mndeveci, line 291: `base_image` is used below, why we are not assigning a value for it?

mndeveci (commented):
I left small comments for now but I want to test this locally and give more detailed feedback later.

samcli/commands/init/__init__.py
jonife: Because prior to this changes `InitTemplates` class did not been use it, it was redundant to pass what we don't use.

samcli/commands/init/__init__.py
jonife: First the _get_runtime_from_image method is confusing as it does 2 things but named something else, inspired good coding practice it should return only the runtime derived from the the image name. This was why I refactored it to return just the runtime derived from the image name and moved other functionalities to the method _get_choice_from_options which would return image options for the customer to select. The option selected (image) is then passed to _get_runtime_from_image which gets the runtime. 

The `base_image` is passed to this method. it seemed redundant passing an argument to method only to return the same argument with manipulating it.

samcli/commands/init/interactive_init_flow.py
```diff
@@ -166,56 +182,136 @@ def _generate_from_app_template(
...
-        choice_num = 1
-        click.echo("\nWhich runtime would you like to use?")
-        for r in INIT_RUNTIMES:
-            msg = "\t" + str(choice_num) + " - " + r
-            click.echo(msg)
-            choice_num = choice_num + 1
-        choice = click.prompt("Runtime", type=click.Choice(choices), show_choices=False)
-        runtime = INIT_RUNTIMES[int(choice) - 1]  # zero index
-    return runtime
+def _generate_default_hello_world_application(use_case, package_type, runtime, dependency_manager, pt_explicit):
```
mndeveci, line 185: Can we add documentation for these new methods?

mndeveci (changes requested):

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +119,80 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+        """
+        self.clone_templates_repo()
+        manifest_path = self.get_manifest_path()
+        with open(str(manifest_path)) as fp:
+            body = fp.read()
+            manifest_body = json.loads(body)
+
+        # This would ensure the Use-Case Hello World Example appears
+        # at the top of list template example displayed to the Customer.
+        preprocessed_manifest = {"Hello World Example": {}}
```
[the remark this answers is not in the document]
hawflau: That should be fine, as Python3 `dict` preserves input order.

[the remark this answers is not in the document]
jonife: @mndeveci and I have addressed this

hawflau (approved):

mndeveci (approved):

samcli/commands/init/init_templates.py
```diff
@@ -171,3 +119,126 @@ def is_dynamic_schemas_template(self, package_type, app_template, runtime, base_
...
+
+    Returns
+    -------
+    bool
+        True if template does not meet filter criteria else False
+    """
+    return bool(
+        (app_template and app_template != template["appTemplate"])
+        or (package_type and package_type != template["packageType"])
+        or (dependency_manager and dependency_manager != template["dependencyManager"])
```
mndeveci, line 243:
```suggestion
        (app_template and app_template != template.get("appTemplate"))
        or (package_type and package_type != template.get("packageType"))
        or (dependency_manager and dependency_manager != template.get("dependencyManager"))
```

mndeveci (approved):

Issue #3417, opened 2021-10-27
Change set is throwing an error while deploying the template.


Deepika-Raj:
### Description:

When I am deploying the template using SAM deploy. It is giving an error : 

**botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in input: "DisableRollback", must be one of: ChangeSetName, StackName, ClientRequestToken**

I am not using either of **--disable-rollback | --no-disable-rollback** while deploying

### Steps to reproduce:

The command I am using : 

sam deploy --template-file packaged.yaml --stack-name <name of stack> --s3-bucket <name of bucket> --capabilities CAPABILITY_IAM --region $AWS_DEFAULT_REGION --parameter-overrides <list of parameters>

### Observed result:

Error with --debug flag:

Changeset created successfully. ***
2021-10-27 09:29:38,186 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': 'fb75737e-a71f-40ff-b874-f9c6d5b53f53', 'installationId': '47f3640c-0c6d-4159-9dac-4f18e6526b4d', 'sessionId': 'a629f491-1932-4811-b3b4-7532d1cdceb3', 'executionEnvironment': 'GitLab', 'ci': True, 'pyversion': '3.8.12', 'samcliVersion': '1.34.1', 'awsProfileProvided': False, 'debugFlagProvided': True, 'region': 'eu-west-1', 'commandName': 'sam deploy', 'duration': 16611, 'exitReason': 'ParamValidationError', 'exitCode': 255}}]}
2021-10-27 09:29:38,810 | HTTPSConnectionPool(host='aws-serverless-tools-telemetry.us-west-2.amazonaws.com', port=443): Read timed out. (read timeout=0.1)
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #3445, opened 2021-11-05
fix: Bumped Minimal boto3 Version to 1.18.32

CoshUS:
https://github.com/aws/aws-sam-cli/issues/3417

Bump boto3 version to be minimal 1.18.32 as DisableRollback is add in this version.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.1
 Flask~=1.1.2
 #Need to add Schemas latest SDK.
-boto3~=1.18
+boto3>=1.18.32,==1.*
 jmespath~=0.10.0
 PyYAML~=5.3
 cookiecutter~=1.7.2


torresxb1 (approved):

requirements/base.txt
```diff
@@ -2,7 +2,7 @@ chevron~=0.12
 click~=7.1
 Flask~=1.1.2
 #Need to add Schemas latest SDK.
-boto3~=1.18
+boto3>=1.18.32,==1.*
```
jfuss, line 5: Any reason we choose `==1.*` over `<2`? I think they are equivalent but was hard to reason about in my head.

CoshUS: No difference. Chose this format as it follows the expanded ~= syntax.
PEP 440: https://www.python.org/dev/peps/pep-0440/#compatible-release

jfuss (approved):

Issue #2990, opened 2021-06-28
sam deploy: Support `-` which indicates stdin


conao3:
### Describe your idea/feature/enhancement

Many CLI commands can interpret `-` as a stdin/out. In `aws-cli`, `s3 cp` can use `-` as a stdin/out.

```
$ BUCKET=<your sample bucket>

$ echo helloworld | aws s3 cp - s3://${BUCKET}/tmp

$ aws s3 cp s3://${BUCKET}/tmp -
helloworld
```

If `sam deploy` use stdin to pass template yaml, user can use any *pre process* step for his template.

```
$ # local transform instead of Cloudformation macro
$ cat template.yml | your-favorite-transformer | sam deploy --template-file -


$ # mustache pre-process with SAM template
$ cat template.yml.mustache | mustache data.json | sam deploy --template-file -


$ # simple `sed` for SAM template (but the user should use *Parameters*.)
$ cat template.yml | sed 's/__company/aws/g' | sam deploy --template-file -
```

### Proposal

Add details of how to add this to the product.

1. Add a special action step if the user specifies `-` for `--template-file` option.
2. The special action is, get template string from stdin instead of reading local file.

Things to consider:
1. What happens when a user really wants to deal with a `-` file?
   - Unix command `cat` address `-` file as a `./-`.
     So you can `cat ./-` if you want to see `-` file.
2. Why can we assume that - implicitly indicates stdin/out? Any authoritative sources?
   - See POSIX specification: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02
     "12.2 Utility Syntax Guidelines", "Guideline 13:"
     > For utilities that use operands to represent files to be opened for either reading or writing, the '-' operand should be used to mean only standard input (or standard output when it is clear from context that an output file is being specified) or a file named -.
### Additional Details
None

mildaniel:
Thanks for the feature request! We have a [roadmap](https://github.com/aws/aws-sam-cli/wiki/SAM-CLI-Roadmap) here, but we also try to take in popular feature requests based on 👍🏽 from the community. We will be adding this to our intake for prioritization.

conao3:
If `sam` supports this feature, users will be able to use like `ytt` as a yaml template engine and input directly into `sam` without  any intermediate files.
https://github.com/vmware-tanzu/carvel-ytt

Issue #3442, opened 2021-11-04
Template file is readed in multiple times


conao3:
### Description:

Template file is loaded multiple times.
This problem is not serious but may have a performance impact.

### Steps to reproduce:

1. apply this diff
```diff
diff --git a/samcli/commands/_utils/template.py b/samcli/commands/_utils/template.py
index 61b1c497..ff52d4c5 100644
--- a/samcli/commands/_utils/template.py
+++ b/samcli/commands/_utils/template.py
@@ -46,6 +46,8 @@ def get_template_data(template_file):
     if not pathlib.Path(template_file).exists():
         raise TemplateNotFoundException("Template file not found at {}".format(template_file))
 
+    print("-"*10 + " Reading template file: {}".format(template_file))
     with open(template_file, "r", encoding="utf-8") as fp:
         try:
             return yaml_parse(fp.read())
```
2. save this file as a `sample.yml`
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      Handler: index.handler
      Runtime: python3.8
      Events:
        ApiEvent:
          Type: Api
          Properties:
            Path: /hello
            Method: GET
      InlineCode: |
        def handler(event, context):
          return {
            'statusCode': 200,
            'body': 'Hello World!'
          }

Outputs:
  HelloWorldAPI:
    Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/hello/"
```
3. deploy that template via sam
```
$ samdev deploy -t sample.yml --stack-name dev-sam-sample --capabilities CAPABILITY_IAM
```
4. you can see multiple log that indicates load template multiple times.
```
$ samdev deploy -t sample.yml --stack-name dev-sam-sample --capabilities CAPABILITY_IAM --debug
---------- Reading template file: /home/conao/dev/forks/aws-sam-cli/sample.yml
---------- Reading template file: /home/conao/dev/forks/aws-sam-cli/sample.yml
---------- Reading template file: /home/conao/dev/forks/aws-sam-cli/sample.yml
---------- Reading template file: /home/conao/dev/forks/aws-sam-cli/sample.yml
2021-11-05 00:23:40,100 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
[...]

PR #3450, opened 2021-11-05
feat: sam deploy: Support - which indicates stdin

conao3:
#2990

`-` support allows users to explicitly type stdin directly into sam.
By supporting this feature, users can transform YAML using a YAML template engine such as ytt, which increases flexibility.

When `-` is specified, it saves stdin to a temporary file and assumes that the file is specified.

When sam finishes executing, it will delete the temporary file it created.

When a user creates a template file that is really `-`, specifying `--template-file -` will result in unintended behavior because it indicates stdin.

As with other commands that support `-`, users can use `./-`, explicitly specifying the current directory `-` file.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -4,7 +4,9 @@ Common CLI options shared by various commands
 
 import os
 import logging
-from functools import partial
+from functools import partial, wraps
+import sys
+import tempfile
 import types
 
 import click
@@ -22,6 +24,7 @@ from samcli.cli.types import (
 from samcli.commands._utils.custom_options.option_nargs import OptionNargs
 from samcli.commands._utils.template import get_template_artifacts_format
 from samcli.lib.utils.packagetype import ZIP, IMAGE
+from samcli.lib.utils import osutils
 
 _TEMPLATE_OPTION_DEFAULT_VALUE = "template.[yaml|yml|json]"
 DEFAULT_STACK_NAME = "sam-app"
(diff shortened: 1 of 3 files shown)


conao3:
Can someone tell me how to add a unit test for this feature, please?

conao3:
My patch writes stdin to a temporary file. This is because stdin can only be read once.
Currently, sam loads the template multiple times randomly. So my implementation writes it to a temporary file so that the template can be read multiple times.

This issue was reported in #3442. Once that issue is resolved, we can read stdin directly without writing to a temporary file.

c2tarun (changes requested):
Change looks reasonable but lacks integration tests. From sam-cli's point of view an integration test basically test if the command works. Can you please add some integration test for this?

samcli/commands/_utils/options.py
```diff
@@ -82,6 +105,18 @@ def get_or_default_template_file_name(ctx, param, provided_value, include_build)
     :return: Actual value to be used in the CLI
     """
 
+    if provided_value == "-":
+        # "-" is a special value that tells us to read the template from stdin
```
mildaniel, line 109: This logic isn't specific to deploy. It can be used with any command, but the cleanup step is only done for deploy. Can we clean up the temp files for all commands?

samcli/commands/deploy/command.py
```diff
@@ -127,6 +128,7 @@
 @track_command
 @check_newer_version
 @print_cmdline_args
+@remove_stdin_tmpfile
```
hawflau, line 131: As @mildaniel suggested, please apply `@remove_stdin_tmpfile` to all commands that accept `--template-file`.

hawflau (changes requested):
Thanks for your contributions!. It looks good overall but please address the issues as @c2tarun and @mildaniel raised. Also please resolve the merge conflicts.

hawflau:
Hi @conao3 would you be able to work on the change requests?

conao3:
I'm busy with my job, so I cannot take time on this PR for now.
And maybe large changes are needed. When I created this PR `sam` was reading its template multiple times, so I had to write it out once to a tmp file to accommodate multiple reads. (also reported #3442)
Perhaps it would be better to modify it first so that it does not have to be read multiple times.

hawflau:
@conao3 no worries. Let me convert this PR to draft and talk to team to see if we can pick up this PR to continue your work. Thanks for your contribution again!

conao3:
Good, thanks!

jfuss:
Closing as inactive.

Happy to review an updated PR from the community, if someone wants to revive the PR.

PR #3451, opened 2021-11-07
feat: support openapi yaml include validation

NickHeap2:
#3041

If you are using OpenAPI yaml specifications it makes sense to use them for the API DefinitionBody like this:
```
      DefinitionBody:
        'Fn::Transform':
          Name: AWS::Include
          Parameters:
            Location: openapi.yaml
```
Whilst this is valid in a template it is currently not supported by sam validate.

This change detects this pattern and replaces the DefinitionBody with the actual contents of the file allowing the validation to succeed.

Potentially someone could be relying on the existing behaviour where it fails for this pattern to prevent that pattern being used.

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -139,6 +141,32 @@ def _replace_local_image(self):
...
+        for _, resource in resources.items():
+            if not resource.get("Type", "") in [AWS_SERVERLESS_API, AWS_SERVERLESS_HTTPAPI]:
+                continue
+
+            transform_dict = resource.get("Properties", {}).get("DefinitionBody", {}).get("Fn::Transform", {})
+            if transform_dict.get("Name", "") != "AWS::Include":
+                continue
+
+            location_prop = transform_dict.get("Parameters", {}).get("Location", "")
+            if not os.path.isfile(location_prop):
```
jfuss, line 165: Can we use `pathlib` instead of `os.path`?

https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file would be the function call.

samcli/commands/validate/lib/sam_template_validator.py
```diff
@@ -180,3 +208,11 @@ def _update_to_s3_uri(property_key, resource_property_dict, s3_uri_value="s3://b
...
 
         resource_property_dict[property_key] = s3_uri_value
+
+    @staticmethod
+    def _replace_with_file_contents(resource_property_dict, location_prop):
+        definition_body = parse_yaml_file(location_prop)
+
+        # replace the definition body from the file
+        resource_property_dict["DefinitionBody"] = definition_body
+        LOG.debug("Imported definition body from %s", location_prop)
```
jfuss, line 218: Nit: Can you move this above line 217. This doesn't have meaningful impact but I find it to be more consistent with the codebase. If something does happen in line 217, we at least would have a log line to help us/customers debug.

tests/unit/commands/validate/lib/openapi/notopenapi.yaml
```diff
@@ -0,0 +1 @@
+this is not yaml
```
jfuss, line 1: This looks unused. Was the suppose to be for a test?

tests/unit/commands/validate/lib/test_sam_template_validator.py
```diff
@@ -306,3 +413,131 @@ def test_replace_local_codeuri_with_no_resources(self):
...
+        validator = SamTemplateValidator(template, managed_policy_mock)
+
+        validator._replace_local_openapi()
+
+        template_resources = validator.sam_template.get("Resources")
+        self.assertIn("DefinitionBody", template_resources.get("ServerlessApi").get("Properties"))
+        self.assertNotIn("openapi", template_resources.get("ServerlessApi").get("Properties").get("DefinitionBody"))
+        self.assertIn("Fn::Transform", template_resources.get("ServerlessApi").get("Properties").get("DefinitionBody"))
+
+    def test_DefinitionBody_gets_replaced_if_json(self):
```
jfuss, line 482: For the tests that read and touch the file system, can we move these to `integ` tests? 

Generally, the unit tests through the repo utilize mocks and (try not to) touch the system at all. Instead, integ tests are preferred for the end to end. This will also allow us to test the functionality from the CLI interface. 

I don't mind having these but would like to have some basic cases covered with integ tests so we can validate the end to end experience and gives us a good guard for functionality we support, in the case of future regression.

jfuss (changes requested):
Thank you for the contribution!

Just a couple small comments. If you could address those, that would be awesome.

samcli/commands/validate/lib/sam_template_validator.py
NickHeap2: Cheers. Have changed to use pathlib.

samcli/commands/validate/lib/sam_template_validator.py
NickHeap2: Have moved the debug message.

tests/unit/commands/validate/lib/openapi/notopenapi.yaml
NickHeap2: Yes it was for a test case that is no longer needed so have deleted.

tests/unit/commands/validate/lib/test_sam_template_validator.py
NickHeap2: I agree they should be mocked (still learning Python mocking!).
Will rework the tests with mocks and add the none mocked versions into integ tests.

jfuss: @NickHeap2 Let me know if you need some help in the mocking area. Happy to lend a hand to help if you get stuck at all.

NickHeap2: @jfuss The part that took me a while was getting the patch path for parse_yaml file correct.
I've finished changing the unit tests to use mocks and added integration test cases now.

NickHeap2:
@jfuss any chance of getting this across the line?

jfuss (approved):
LGTM. Thanks for the contribution @NickHeap2 

I will find another member of the team to review as well (as we required two approvals).

hawflau (approved):

mingkun2020:
Hi @NickHeap2, we have to revert the changes due to the bug reported in this [issue#3676](https://github.com/aws/aws-sam-cli/issues/3676) by [PR#3685](https://github.com/aws/aws-sam-cli/pull/3685). The validation failed when the location is an intrinsic function like `Location: !Join ["", ["s3://", !Ref SomePath, "/openapi.yaml"]]`.

NickHeap2:
OK @mingkun2020 I'll get the change reworked as soon as I get a chance.

PR #3460, opened 2021-11-11
loosen requirements

tomscytale:
use ~= for some entries in requirements.txt in order to make life eaiser for users

If I want to install aws-sam-cli in the same environment as other tools I generally end up with pip complaining about incompatibile dependency versions and occasionally refusing to install or (in recent versions) going into a loop.

uses ~= to allow any [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) as defined in PEP-440

It may change the versions of some dependencies, albeit to compatible versions. 

#### Checklist

I believe none of the below apply here

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,7 +10,7 @@ aws-sam-translator==1.40.0
 #docker minor version updates can include breaking changes. Auto update micro version only.
 docker~=4.2.0
 dateparser~=1.0
-requests>=2.25,<3
+requests~=2.25
 serverlessrepo==0.1.10
 aws_lambda_builders==1.9.0
 tomlkit==0.7.2
@@ -23,4 +23,4 @@ dataclasses==0.8; python_version < '3.7'
 # NOTE: regex is not a direct dependency of SAM CLI, but pin to 2021.9.30 due to 2021.10.8 not working on M1 Mac - https://bitbucket.org/mrabarnett/mrab-regex/issues/399/missing-wheel-for-macosx-and-the-new-m1
 regex==2021.9.30
 # NOTE: tzlocal is not a direct dependency of SAM CLI, but pin to 3.0 as 4.0 break appveyor jobs
-tzlocal<4
+tzlocal~=3.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,17 +10,17 @@ aws-sam-translator==1.40.0
 #docker minor version updates can include breaking changes. Auto update micro version only.
 docker~=4.2.0
 dateparser~=1.0
-requests==2.25.1
+requests>=2.25,<3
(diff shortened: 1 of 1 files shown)


jfuss:
SAM CLI isn't meant to be installed along side other tools. We try to keep tight bounds so what we test is what we release. 

@tomscytale You mention that this is to make "life easier for users", but what is being made easier? We can bump versions as needed but we have had dependencies break the cli on minor versions in the past so I am not inclined to expand requirements.

tomscytale:
> SAM CLI isn't meant to be installed along side other tools. We try to keep tight bounds so what we test is what we release.

Anything beyond the most trivial deployment pipeline is probably going to need more than just SAM CLI - in our case we also use awscli and some in-house python tools.

> but what is being made easier? 

setting up the CI environment, using other python tools along with aws-sam-cli

jfuss:
@tomscytale Our preferred method is using our installers. Any reason you aren't using those? If you are really wanting to use pip (which we don't recommend), the solution would be to install in isolation (through an venv).

tomscytale:
in this case I'm aiming to build a minimal base docker image.

installing using pip means I avoid duplicate installation of dependencies - most of aws-sam-cli's dependencies are already present in my virtual env - I can save probably about 100MB 

It's not a huge deal, but it's enough to make me tweak my requirements file to accommodate aws-sam-cli - it would just make for a smoother experience if it specified a range of dependencies.

I'm sure there are other quite reasonable use-cases where installing via pip makes most sense.

Also, if you're going to publish to pypi then people are very reasonably going to assume that installing via pip is going to work without unexpected, and maybe unnecessary, version conflicts.

And lastly, I don't see it mentioned anywhere that installing with pip is not recommended.

jfuss:
> And lastly, I don't see it mentioned anywhere that installing with pip is not recommended.

We can make that more explicit then but we don't document using pip at all. So yes it is in PyPi, which is due to the early days of the project. The team has since moved away due to the CLI suddenly stops working because of a dependency in the same interpreter or customer forgets which `pip` they used and their system is resolving to an unexpected one (among others).

I understand where you are coming from, however we pin down requirements as much as we can and not something we are planning to change. This avoids a customers installing a newer dependency which breaking the CLI. The way we view this is, a release should be install-able and work. If we don't pin down that can be easily broken (and has). So you are free to continue to use pip but you should isolate the installation.

jfuss:
Again, we are happy to bump versions but we won't open to wide dependency version installations.

jfuss:
Closing as there has been no response. We are happy to update dependencies but they must be pinned to minor versions so we can have some sort of guards/guarantees on the CLI we distribute.

PR #3471, opened 2021-11-17
fix: Correcting workflow for http and rest api code syncs

qingchm:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/sync/flows/rest_api_sync_flow.py
+++ b/samcli/lib/sync/flows/rest_api_sync_flow.py
@@ -114,7 +114,7 @@ class RestApiSyncFlow(GenericApiSyncFlow):
         if api_resource:
             if api_resource.get("Type") == AWS_SERVERLESS_API:
                 # The customer defined stage name
-                stage_name = api_resource.get("Properties").get("StageName")  # type: ignore
+                stage_name = api_resource.get("Properties", {}).get("StageName")
                 if stage_name:
                     stages.add(cast(str, stage_name))
 
@@ -130,8 +130,8 @@ class RestApiSyncFlow(GenericApiSyncFlow):
             # RestApiId is a required field in stage
             stage_dict = get_resource_by_id(self._stacks, stage_resource)
             if stage_dict:
-                rest_api_id = stage_dict.get("Properties").get("RestApiId")  # type: ignore
-                dep_id = stage_dict.get("Properties").get("DeploymentId")  # type: ignore
+                rest_api_id = stage_dict.get("Properties", {}).get("RestApiId")
+                dep_id = stage_dict.get("Properties", {}).get("DeploymentId")
                 # If the stage doesn't have a deployment associated then no need to update
                 if dep_id is None:
                     continue
@@ -139,7 +139,7 @@ class RestApiSyncFlow(GenericApiSyncFlow):
                 for item in get_resource_ids_by_type(self._stacks, AWS_APIGATEWAY_DEPLOYMENT):
                     if item.logical_id == dep_id:
                         if rest_api_id == self._api_identifier:
-                            stages.add(cast(str, stage_dict.get("Properties").get("StageName")))  # type: ignore
+                            stages.add(cast(str, stage_dict.get("Properties", {}).get("StageName")))
 
         return stages
 
--- a/samcli/lib/sync/flows/rest_api_sync_flow.py
+++ b/samcli/lib/sync/flows/rest_api_sync_flow.py
@@ -8,6 +8,7 @@ from samcli.lib.sync.flows.generic_api_sync_flow import GenericApiSyncFlow
 from samcli.lib.providers.provider import ResourceIdentifier, Stack, get_resource_by_id, get_resource_ids_by_type
 from samcli.lib.providers.exceptions import MissingLocalDefinition
 from samcli.lib.utils.colors import Colored
+from samcli.lib.utils.resources import AWS_SERVERLESS_API, AWS_APIGATEWAY_STAGE, AWS_APIGATEWAY_DEPLOYMENT
 
 # BuildContext and DeployContext will only be imported for type checking to improve performance
 # since no instances of contexts will be instantiated in this class
@@ -106,15 +107,16 @@ class RestApiSyncFlow(GenericApiSyncFlow):
         # Get the stage name associated with the previous deployment and update stage
         # Stage needs to be flushed so that new changes will be visible immediately
         api_resource = get_resource_by_id(self._stacks, ResourceIdentifier(self._api_identifier))
-        stage_resources = get_resource_ids_by_type(self._stacks, "AWS::ApiGateway::Stage")
+        stage_resources = get_resource_ids_by_type(self._stacks, AWS_APIGATEWAY_STAGE)
 
         stages = set()
         # If it is a SAM resource, get the StageName property
         if api_resource:
-            if api_resource.get("Type") == "AWS::Serverless::Api":
+            if api_resource.get("Type") == AWS_SERVERLESS_API:
                 # The customer defined stage name
                 stage_name = api_resource.get("Properties").get("StageName")  # type: ignore
-                stages.add(cast(str, stage_name))
+                if stage_name:
+                    stages.add(cast(str, stage_name))
 
                 # The Stage stage
                 if stage_name != "Stage":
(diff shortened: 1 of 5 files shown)


samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -51,12 +55,106 @@ def __init__(
 
     def set_up(self) -> None:
         super().set_up()
-        self._api_client = self._boto_client("apigateway")
+        self._api_client = cast(Session, self._session).client("apigateway")
```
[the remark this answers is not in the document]
qingchm: Thanks! Updated!

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +51,156 @@ def __init__(
...
+        Collect all stages needed to be updated
+
+        Returns
+        ----------
+        Set[str]: The set of stage names to be updated
+        """
+        # Get the stage name associated with the previous deployment and update stage
+        # Stage needs to be flushed so that new changes will be visible immediately
+        api_resource = get_resource_by_id(self._stacks, ResourceIdentifier(self._api_identifier))
+        stage_resources = get_resource_ids_by_type(self._stacks, "AWS::ApiGateway::Stage")
```
mndeveci, line 109: nit: please use static resource type variables for this (plus the ones below)

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +51,156 @@ def __init__(
...
+
+        stages = set()
+        # If it is a SAM resource, get the StageName property
+        if api_resource:
+            if api_resource.get("Type") == "AWS::Serverless::Api":
+                # The customer defined stage name
+                stage_name = api_resource.get("Properties").get("StageName")  # type: ignore
+                stages.add(cast(str, stage_name))
```
mndeveci, line 117: Would there be a case where customer did not defined a `StageName`? If so we should be checking whether the value of that property is not `None`.

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +51,156 @@ def __init__(
...
+                        if item.get("stageName") == "Stage":
+                            stages.add("Stage")
+
+        # For both SAM and ApiGateway resource, check if any refs from stage resources
+        for stage_resource in stage_resources:
+            # RestApiId is a required field in stage
+            stage_dict = get_resource_by_id(self._stacks, stage_resource)
+            if stage_dict:
+                rest_api_id = stage_dict.get("Properties").get("RestApiId")  # type: ignore
+                dep_id = stage_dict.get("Properties").get("DeploymentId")  # type: ignore
```
mndeveci, line 132: Same comment as above, I am not sure if these properties are required. So let's make sure they are not None before working on them.

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +51,156 @@ def __init__(
...
+            try:
+                response_del = cast(
+                    Dict, self._api_client.delete_deployment(restApiId=self._api_physical_id, deploymentId=prev_dep_id)
+                )
+                LOG.debug("%sDelete Deployment Result: %s", self.log_prefix, response_del)
+            except ClientError:
+                LOG.warning(
+                    Colored().yellow(
+                        "Delete deployment for %s failed, it may be due to the it being used by another stage. \
+please check the console if you want to delete it"
```
mndeveci, line 203: May not be true, but should we suggest them to run infra sync?

samcli/lib/sync/flows/rest_api_sync_flow.py
qingchm: Translation would fail if the property StageName is not defined. All serverless api resources are required to have this field. But in the rare case that customer never ran any other sam commands before this and directly call code sync, this might be a problem. So adding a check also makes sense.

samcli/lib/sync/flows/rest_api_sync_flow.py
qingchm: Deployments are not manageable in AWS console so they will have to delete the stage if they want to have that deployment removed, I can update the text to be more descriptive. But infra sync doesn't do the job either. We don't need to guarantee a deployment clean up I would regard it as a good practice instead

samcli/lib/sync/flows/rest_api_sync_flow.py
qingchm: They are also required properties, I don't think we need to do the job that sam validate does here. These two values being None doesn't hurt the sync process so we don't need to have any handling here. The only case that this code crashes would be when a customer have a stage resource that doesn't have the Properties field. Again I think this validation would be too unnecessary. Let me know your thoughts!

samcli/lib/sync/flows/rest_api_sync_flow.py
qingchm: Will update that!

samcli/lib/sync/flows/rest_api_sync_flow.py
mndeveci: Yeah I don't think we need to run validation for `Properties`. But for the sake of safetiness, we can use `stage_dict.get("Properties", {}).get("DeploymentId")` so that even if they have an empty `Properties` this won't end up with any exception

qingchm: Good point! I'll add that

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +52,157 @@ def __init__(
...
+            LOG.debug("%sTrying to update the stage %s through client", self.log_prefix, stage)
+            response_upd = cast(
+                Dict,
+                self._api_client.update_stage(
+                    restApiId=self._api_physical_id,
+                    stageName=stage,
+                    patchOperations=[{"op": "replace", "path": "/deploymentId", "value": deployment_id}],
+                ),
+            )
+            LOG.debug("%sUpdate Stage Result: %s", self.log_prefix, response_upd)
```
CoshUS, line 178: Same comment as above. Do we need to debug log the response dictionary?

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +52,157 @@ def __init__(
...
+        prev_deployment_ids: Set[str]
+            A set of previous deployment IDs to be cleaned up
+        """
+        for prev_dep_id in prev_deployment_ids:
+            LOG.debug("%sTrying to delete the previous deployment %s through client", self.log_prefix, prev_dep_id)
+            try:
+                response_del = cast(
+                    Dict, self._api_client.delete_deployment(restApiId=self._api_physical_id, deploymentId=prev_dep_id)
+                )
+                LOG.debug("%sDelete Deployment Result: %s", self.log_prefix, response_del)
```
CoshUS, line 200: Same as above.
Can we also move `LOG.debug("%sTrying to delete the previous deployment %s through client", self.log_prefix, prev_dep_id)` into the try block so it's easier to read?

CoshUS (changes requested):

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +52,157 @@ def __init__(
...
+    def _update_api(self) -> None:
+        """
+        Update the API content
+        """
+        LOG.debug("%sTrying to update RestAPI through client", self.log_prefix)
+        response_put = cast(
+            Dict,
+            self._api_client.put_rest_api(restApiId=self._api_physical_id, mode="overwrite", body=self._swagger_body),
+        )
+        LOG.debug("%sPut RestApi Result: %s", self.log_prefix, response_put)
```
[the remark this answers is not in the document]
mndeveci: We are using similar debugging when getting SFN logging configuration and that helped us during debugging the other issue. 

I am OK to keep it as it is (since it is debug) as long as it is not really long. 
If the output is really long (as @CoshUS suggestion) we can pick some fields from results and add them to the debug instead of putting everything.

samcli/lib/sync/flows/rest_api_sync_flow.py
```diff
@@ -48,15 +52,157 @@ def __init__(
...
+        stages = set()
+        # If it is a SAM resource, get the StageName property
+        if api_resource:
+            if api_resource.get("Type") == AWS_SERVERLESS_API:
+                # The customer defined stage name
+                stage_name = api_resource.get("Properties", {}).get("StageName")
+                if stage_name:
+                    stages.add(cast(str, stage_name))
+
+                # The Stage stage
```
mingkun2020, line 121: This make me confuse, should it be `The stage named "Stage"`?

CoshUS: Yes. SAM will always create a stage named "Stage".

samcli/lib/sync/flows/rest_api_sync_flow.py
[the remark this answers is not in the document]
qingchm: Yep right now we are logging the full response for the api calls, might want to improve this later but I would say let's prioritize the bug and have that fixed first

CoshUS (approved):
LGTM! Thanks for addressing the comments!

mingkun2020 (approved):
LGTM!

PR #3472, opened 2021-11-17
fix: validate sfn logging configuration

mndeveci:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/observability/cw_logs/cw_log_group_provider.py
+++ b/samcli/lib/observability/cw_logs/cw_log_group_provider.py
@@ -144,7 +144,7 @@ class LogGroupProvider:
             log_group_arn = logging_destination.get("cloudWatchLogsLogGroup", {}).get("logGroupArn")
             LOG.debug("Log group ARN: %s", log_group_arn)
             if log_group_arn:
-                if ":" in log_group_arn:
+                if ":" in log_group_arn and len(log_group_arn.split(":")) > 6:
                     log_group_arn_parts = log_group_arn.split(":")
                     log_group_name = log_group_arn_parts[6]
                     return str(log_group_name)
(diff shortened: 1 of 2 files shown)


samcli/lib/observability/cw_logs/cw_log_group_provider.py
```diff
@@ -144,9 +144,17 @@ def for_step_functions(
             log_group_arn = logging_destination.get("cloudWatchLogsLogGroup", {}).get("logGroupArn")
             LOG.debug("Log group ARN: %s", log_group_arn)
             if log_group_arn:
-                log_group_arn_parts = log_group_arn.split(":")
-                log_group_name = log_group_arn_parts[6]
-                return str(log_group_name)
+                if ":" in log_group_arn:
```
qingchm, line 147: should we validate the length of the split parts as well? in case of invalid arn's

mndeveci: Very good point, will update the PR

qingchm (approved):

samcli/lib/observability/cw_logs/cw_log_group_provider.py
```diff
@@ -144,9 +144,17 @@ def for_step_functions(
...
             if log_group_arn:
-                log_group_arn_parts = log_group_arn.split(":")
-                log_group_name = log_group_arn_parts[6]
-                return str(log_group_name)
+                if ":" in log_group_arn and len(log_group_arn.split(":")) > 6:
+                    log_group_arn_parts = log_group_arn.split(":")
+                    log_group_name = log_group_arn_parts[6]
+                    return str(log_group_name)
+
+                LOG.debug(
```
CoshUS, line 152: Seems like a LOG.warning to me.

CoshUS (changes requested):

samcli/lib/observability/cw_logs/cw_log_group_provider.py
mndeveci: Yeah I am in between `debug` vs `warning` here. I don't want this to be annoying. But since this will be printed once we can change this to `warning` I believe 🤔

CoshUS (approved):

Issue #3443, opened 2021-11-04
sam sync goes into "running incremental build" loop


maplion:
### Description:
After I successfully sync with sam sync, as soon as I do anything in VSCode (even before saving any changes), it will begin going into this "running incremental build" loop and continue to cycle:

![image](https://user-images.githubusercontent.com/8908849/140407806-9f603566-8fab-4354-b2fb-58016d2fcc24.png)

### Steps to reproduce:
Run `sam sync --stack-name <stack name> --watch --profile <profile> --region us-east-1`

### Observed result:
Answer "y" when asked if you want to proceed, then watch Cloudformation initiate deployment, then says "Infra sync completed".  After that, it sits idle, as it should, but as soon as I click into the template.yml, it starts looping.

### Expected result:
It should remain idle until I save something.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10
2. `sam --version`: 1.35.0
3. AWS region: us-east-1

`Add --debug flag to command you are running`

Did this and it is constantly spamming the following (note StageName and LayerArn are Parameters):

2021-11-04 13:41:29,579 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,599 | Found the same SyncFlow in queue. Skip adding.
2021-11-04 13:41:29,599 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,618 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,637 | Found the same SyncFlow in queue. Skip adding.
2021-11-04 13:41:29,637 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,656 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,674 | Found the same SyncFlow in queue. Skip adding.
2021-11-04 13:41:29,674 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,693 | Collected default values for parameters: {'StageName': 'v1', 'LayerArn': 'v1'}
2021-11-04 13:41:29,712 | Found the same SyncFlow in queue. Skip adding.
[...]

PR #3477, opened 2021-11-18
fix: Skip Over .aws-sam Folder

CoshUS:

https://github.com/aws/aws-sam-cli/issues/3443
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/resource_trigger.py
+++ b/samcli/lib/utils/resource_trigger.py
@@ -19,6 +19,9 @@ from samcli.local.lambdafn.exceptions import FunctionNotFound, ResourceNotFound
 from samcli.lib.utils.resources import RESOURCES_WITH_LOCAL_PATHS
 
 
+AWS_SAM_FOLDER_REGEX = "^.*\\.aws-sam.*$"
+
+
 class OnChangeCallback(Protocol):
     """Callback Type"""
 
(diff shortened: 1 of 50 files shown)


samcli/lib/utils/resource_trigger.py
```diff
@@ -58,28 +60,40 @@ def get_single_file_path_handler(file_path_str: str) -> PathHandler:
...
         dir_path = Path(dir_path_str).resolve()
+
+        # watchdog uses PurePath.match() for patterns
+        # It chooses either PurePosixPath or PureWindowsPath depending on case_sensitive
+        # We have to use the correct Path object in order for path.match() to work properly
+        case_sensitive = platform.system().lower() == "windows"
+
         file_handler = PatternMatchingEventHandler(
-            patterns=["*"], ignore_patterns=[], ignore_directories=False, case_sensitive=True
+            patterns=["*"], ignore_patterns=ignore_patterns, ignore_directories=False, case_sensitive=case_sensitive
```
mndeveci, line 96: This was `True` by default but now it is only `True` if that is a windows environment. Linux is also case sensitive, is that going to create issues there?

CoshUS: Fixed. Should be True for Unix and False for Windows.

samcli/lib/utils/resource_trigger.py
```diff
@@ -58,28 +60,40 @@ def get_single_file_path_handler(file_path_str: str) -> PathHandler:
...
         -------
         PathHandler
             The PathHandler for the folder specified
         """
         dir_path = Path(dir_path_str).resolve()
+
+        # watchdog uses PurePath.match() for patterns
+        # It chooses either PurePosixPath or PureWindowsPath depending on case_sensitive
+        # We have to use the correct Path object in order for path.match() to work properly
+        case_sensitive = platform.system().lower() != "windows"
```
mndeveci, line 93: nit: this can be moved into a common method

mndeveci (approved):

mingkun2020 (approved):

PR #3492, opened 2021-11-23
feat: Consume image asset metadata from CDK synth

mildaniel:
#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -56,10 +72,59 @@ def _replace_property(property_key, property_value, resource, logical_id):
...
+        metadata dict
+            Metadata to use for extracting image assets properties
+
+        Returns
+        -------
+        dict
+            metadata properties for image-type lambda function
+
+        """
+        path_from_asset, dockerfile = os.path.split(metadata.get(ASSET_DOCKERFILE_PATH_KEY))
```
jfuss, line 105: Can we use Pathlib instead?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -56,10 +72,59 @@ def _replace_property(property_key, property_value, resource, logical_id):
...
+        asset_path = metadata.get(ASSET_PATH_METADATA_KEY)
+        dockerfile_context = os.path.join(asset_path, path_from_asset)
```
jfuss, line 107: What happens in asset_path is `None`?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -56,10 +72,59 @@ def _replace_property(property_key, property_value, resource, logical_id):
...
+            target_dict[nested_keys[0]] = property_value
         elif property_key or property_value:
             LOG.info(
                 "WARNING: Ignoring Metadata for Resource %s. Metadata contains only aws:asset:path or "
                 "aws:assert:property but not both",
                 logical_id,
             )
+
+    @staticmethod
+    def _extract_image_asset_metadata(metadata):
```
jfuss, line 90: You should have tests that validate windows works as expected. Are Paths provided by the Metadata (aka CDK here) always paths that match the system? Meaning should we expect windows paths on windows or unix paths on windows?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -56,10 +72,59 @@ def _replace_property(property_key, property_value, resource, logical_id):
 
         """
         if property_key and property_value:
-            resource.get(PROPERTIES_KEY, {})[property_key] = property_value
+            nested_keys = property_key.split(".")
```
jfuss, line 75: I assume this is to find/update a nested value, like `Code.ImageUri`? If so, can we use `jmespath` instead? We use that lib in `package` to do similar things: https://github.com/aws/aws-sam-cli/blob/ec9f5d89971b823fa01071c96b7de90ea5e5303a/samcli/lib/package/artifact_exporter.py#L94

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -31,9 +41,15 @@ def normalize(template_dict):
 
         for logical_id, resource in resources.items():
             resource_metadata = resource.get(METADATA_KEY, {})
-            asset_path = resource_metadata.get(ASSET_PATH_METADATA_KEY)
             asset_property = resource_metadata.get(ASSET_PROPERTY_METADATA_KEY)
 
+            if asset_property == IMAGE_ASSET_PROPERTY:
+                asset_metadata = ResourceMetadataNormalizer._extract_image_asset_metadata(resource_metadata)
+                ResourceMetadataNormalizer._update_resource_image_asset_metadata(resource_metadata, asset_metadata)
+                asset_path = logical_id.lower()
```
jfuss, line 49: Why is the logical_id the `asset_path`?

jfuss (changes requested):
Can we add integration tests for this, so we know how this behaves on Windows and Linux systems?

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: This is equivalent to the name of the docker image generated by sam build and sets it to the value of Code.ImageUri.

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: I tried using `set_value_from_jmespath()` but it might not work in this case. We get the template in with `Properties.Code = some_string` and it can't replace a string with a dict. Internally it recurses through the jmespath and sets the next property, so when it gets to the `Properties.Code` string it tries to do `some_string[key] = value` which obviously errors out.

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: In theory it shouldn't ever be None, but if it is then the Lambda function image won't get built, and it'll fail when trying to local invoke.

mildaniel:
> Can we add integration tests for this, so we know how this behaves on Windows and Linux systems?

We have an upcoming task for next sprint to add integration testing for all CDK-based build and local commands so I was deferring it until then. We don't currently have the underlying configurations in the develop branch to add integration tests with CDK.

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: I did some messing around with CDK on Windows and it looks like `aws:asset:path` will return the system-matching paths, whereas `aws:asset:dockerfile-path` is always a unix-style relative path. I added a test for this Windows case.

samcli/lib/samlib/resource_metadata_normalizer.py
jfuss: I don't understand the explanation. Did you try to use it and it error-ed out?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -56,10 +72,60 @@ def _replace_property(property_key, property_value, resource, logical_id):
...
+        metadata dict
+            Metadata to use for extracting image assets properties
+
+        Returns
+        -------
+        dict
+            metadata properties for image-type lambda function
+
+        """
+        asset_path = pathlib.Path(metadata.get(ASSET_PATH_METADATA_KEY, ""))
```
jfuss, line 105: Nit: follow [pep8](https://www.python.org/dev/peps/pep-0008/#imports). Specifically Absolute imports: `from pathlib import Path`

jfuss (changes requested):

jfuss:
> We have an upcoming task for next sprint to add integration testing for all CDK-based build and local commands so I was deferring it until then. We don't currently have the underlying configurations in the develop branch to add integration tests with CDK.

I don't understand why we are doing this. We can write integ tests now and assume the template format. We do this already but the same thing: https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/local/invoke/test_integrations_cli.py#L67

I don't mean to be a strict here but I am not inclined to approve a pr without proper testing. We shouldn't be splitting testing from the implementation. We wouldn't accept this from contributors, why would we accept it for maintainers?

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: Yes, that's correct.

mildaniel:
@jfuss That's a completely fair call-out, I will add some integration testing.

> I don't understand why we are doing this. We can write integ tests now and assume the template format.

When writing this I was under the assumption that CDK integration tests would start from a CDK app -> `cdk synth` -> test CFN output. I see that we decided to go with testing starting from the CFN template which won't need some of the changes that would otherwise be required.

samcli/lib/samlib/resource_metadata_normalizer.py
jfuss: hmm ok. I won't block on this as it's small. I would like to get rid of it to make it easier to maintain but not the end of the world here.

tests/integration/local/invoke/invoke_cdk_templates.py
```diff
@@ -0,0 +1,59 @@
...
+from docker.errors import APIError
+from parameterized import parameterized
+
+from samcli import __version__ as version
+from samcli.local.docker.lambda_image import RAPID_IMAGE_TAG_PREFIX
+from tests.integration.local.invoke.invoke_integ_base import InvokeIntegBase
+from samcli.lib.utils.architecture import X86_64
+
+
+class TestCDKSynthesizedTemplatesFunctions(InvokeIntegBase):
```
jfuss, line 12: Thank you!

tests/integration/local/invoke/invoke_cdk_templates.py
```diff
@@ -0,0 +1,59 @@
...
+    def setUpClass(cls):
+        # Run sam build first to build the image functions
+        super(TestCDKSynthesizedTemplatesFunctions, cls).setUpClass()
+        build_command_list = super().get_build_command_list(cls, template_path=cls.template_path)
+        super().run_command(cls, command_list=build_command_list)
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        # Remove images
+        docker_client = docker.from_env()
```
jfuss, line 31: This feels like we could abstract this into the base class but maybe there is a reason that won't work.

Why do this on `tearDownClass` over `teardown`? I expected to tear down after each function but maybe this is an optimization on tests? Whichever the reason, can you document it for clarity. Just helps us understand the thoughts behind things in the future.

tests/integration/local/invoke/invoke_cdk_templates.py
```diff
@@ -0,0 +1,59 @@
...
+        for function in cls.functions:
+            try:
+                docker_client.api.remove_image(f"{function.lower()}")
+                docker_client.api.remove_image(f"{function.lower()}:{RAPID_IMAGE_TAG_PREFIX}-{version}-{X86_64}")
+            except APIError:
+                pass
+
+    @parameterized.expand(functions)
+    def test_build_and_invoke_image_function(self, function_name):
```
jfuss, line 40: Most of our docker tests are re-runs due to weird networking or docker issues: https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/local/invoke/test_integrations_cli.py#L35 

Should we do the same here?

jfuss (changes requested):
Just a couple small comments. Thanks for adding integ tests to this pr.

tests/integration/local/invoke/invoke_cdk_templates.py
mildaniel: This logic is inherent to this class. Not all classes that extend the base class need to be built and cleaned up. 

As for `tearDownClass` over `teardown`, we only need to run it once when we finish with these Docker images. Re-building and re-removing the images after each individual test will eat up a lot of unnecessary time since they don't change between tests.

tests/integration/local/invoke/invoke_cdk_templates.py
mildaniel: Looks like we also skip build tests on Windows in CI. Not exactly sure why but I'm assuming some Docker weirdness. Will add that in too.

tests/integration/local/invoke/invoke_cdk_templates.py
jfuss: That was my guess but would be good to just document that inline. Might be good for us to apply this logic across our tests. Flagging to @mndeveci and @CoshUS who I think are looking at some improvements in this area. Might not work in all cases but could be general useful for some things at least.

tests/integration/local/invoke/invoke_cdk_templates.py
jfuss: We skip build tests because we run them separately: https://github.com/aws/aws-sam-cli/blob/develop/appveyor-windows-build-dotnet.yml

Might make sense to add these to the correct language windows appveyor job.

jfuss (approved):
Thanks for handling feedback and adding comments to make the code more readable!

wchengru (approved):

PR #3495, opened 2021-11-25
chore: Integration tests for sync command infrastructure sync

qingchm:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/testdata/sync/Python/function/app.py
```diff
@@ -0,0 +1,43 @@
...
+        Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
+    """
+
+    # try:
+    #     ip = requests.get("http://checkip.amazonaws.com/")
+    # except requests.RequestException as e:
+    #     # Send some context about this error to Lambda Logs
+    #     print(e)
+
+    #     raise e
```
mndeveci, line 35: Let's uncomment these lines so that we can test an execution with a dependency. This will be useful with auto-dependency layer.

tests/integration/testdata/sync/Ruby/function/app.rb
```diff
@@ -0,0 +1,39 @@
...
+  #     'statusCode' and 'body' are required
+  #     # api-gateway-simple-proxy-for-lambda-output-format
+  #     Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
+
+  # begin
+  #   response = HTTParty.get('http://checkip.amazonaws.com/')
+  # rescue HTTParty::Error => error
+  #   puts error.inspect
+  #   raise error
+  # end
```
mndeveci, line 30: Similar to python one, let's uncomment these lines so we can test a function with dependency.

tests/integration/testdata/sync/Python/events/event.json
```diff
@@ -0,0 +1,62 @@
+{
```
mndeveci, line 1: Are we using this file in the integration tests?

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,398 @@
...
+        cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "sync")
+
+        # To be used for code sync integ tests
+        # template_path = str(cls.test_data_path.joinpath("template.yaml"))
+        # cls.shared_stack_name = "accelerate-shared-stack"
+        # LOG.info(cls.shared_stack_name)
+
+        # command = "samdev" if os.getenv("SAM_CLI_DEV") else "sam"
+        # sync_command_list = [command, "sync", "-t", template_path, "--dependency-layer", "--stack-name", cls.shared_stack_name]
+        # run_command_with_input(sync_command_list, "y\n".encode())
```
mndeveci, line 75: Do we need these commented-out lines?

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,398 @@
...
+    #         capabilities_list=None,
+    #         role_arn=None,
+    #         notification_arns=None,
+    #         tags="integ=true clarity=yes foo_bar=baz",
+    #         metadata=None,
+    #     )
+
+    #     sync_process_execute = run_command_with_input(sync_command_list, "y\n".encode())
+    #     self.assertEqual(sync_process_execute.process.returncode, 2)
+    #     self.assertIn("", str(sync_process_execute.stderr))
```
mndeveci, line 294: Same as previous comment, let's clean up redundant/commented code if we are not going to use it.

tests/integration/testdata/sync/Python/events/event.json
qingchm: Nope, this was generated with sam init, I will remove this

tests/integration/sync/test_sync_command.py
qingchm: This is useful for the next task so I kept it here

tests/integration/testdata/sync/Python/function/app.py
qingchm: Nice suggestion! Adding the ip call as well as a part of the test

tests/integration/sync/test_sync_command.py
qingchm: I think I will add a image function test eventually that's why I left this here!

tests/integration/sync/test_sync_command.py
CoshUS: Wouldn't code sync be in a different test case file?

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,398 @@
...
+
+from tests.testing_utils import RUNNING_ON_CI, RUNNING_TEST_FOR_MASTER_ON_CI, RUN_BY_CANARY
+from tests.testing_utils import run_command_with_input
+
+# Deploy tests require credentials and CI/CD will only add credentials to the env if the PR is from the same repo.
+# This is to restrict package tests to run outside of CI/CD, when the branch is not master or tests are not run by Canary
+SKIP_SYNC_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
+CFN_SLEEP = 3
+API_SLEEP = 5
+SFN_SLEEP = 10
```
CoshUS, line 40: nit: Can we retry 10 times with a period of 1 second instead of waiting for 10 seconds all the time? 
We can then set longer durations like 20 seconds in case of slow API while having shorter testing time on average.

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,398 @@
...
+    #     sync_process_execute = run_command_with_input(sync_command_list, "y\n".encode())
+    #     self.assertEqual(sync_process_execute.process.returncode, 2)
+    #     self.assertIn("", str(sync_process_execute.stderr))
+
+    def _method_to_stack_name(self, method_name):
+        """Method expects method name which can be a full path. Eg: test.integration.test_deploy_command.method_name"""
+        method_name = method_name.split(".")[-1]
+        return f"{method_name.replace('_', '-')}-{CFN_PYTHON_VERSION_SUFFIX}"
+
+    def _stack_name_to_companion_stack(self, stack_name):
```
CoshUS, line 301: Can we refactor `_stack_name_to_companion_stack` and `_delete_companion_stack` from both deploy and sync into a common file?
Applies to other common methods as well.

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,398 @@
...
+                cfn_client.delete_stack(StackName=stack_name)
+        super().tearDown()
+
+    # To be used for code sync integ tests
+    # @pytest.fixture
+    # def clean_up_shared_stack(self):
+    #     pass
+
+    @parameterized.expand(["template.yaml"])
+    def test_sync_infra(self, template_file):
```
CoshUS, line 109: Do we want to test updating the stack with infra sync as well?

tests/integration/testdata/sync/template.yaml
```diff
@@ -0,0 +1,70 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
CoshUS, line 1: nit: Can we move infra sync testdata into a subfolder of testdata/sync? We probably will need more version of this for code syncs and other tests in the future.
`testdata/sync/base-infra-stack` or something similar.

CoshUS (changes requested):

tests/integration/testdata/sync/template.yaml
```diff
@@ -0,0 +1,70 @@
...
+    Properties:
+      CodeUri: Python/function/
+      Handler: app.lambda_handler
+      Runtime: python3.7
+      Layers:
+        - Ref: HelloWorldLayer
+      Tracing: Active
+
+  HelloWorldApi:
+    Type: AWS::Serverless::Api
```
CoshUS, line 20: Do we want to test HttpApi as well?

qingchm: Yes we definitely want to test HttpApi's, but currently I have not found a way that HttpApi's can be tested successfully with our integ tests. HttpApi's can only be tested by sending actual http requests. The one I commented out was something I attempted to test it but it doesn't return the expected message. If I can get a working test call I will add it.

tests/integration/testdata/sync/template.yaml
qingchm: Yes I'm doing a new ordering of folder structures! Thanks!

tests/integration/sync/test_sync_command.py
qingchm: Yes, right now calling infra twice will cause many unexpected errors. I already have two sets of templates set up for updating tests. Once I figure out how to avoid the errors I will add it (probably before EOD).

tests/integration/sync/test_sync_command.py
qingchm: Good suggestion. Putting it inside package integ base, since they both inherit this class

tests/integration/sync/test_sync_command.py
qingchm: Good point! Added a logic for this

tests/integration/sync/test_sync_command.py
```diff
@@ -354,43 +330,50 @@ def _get_api_message(self, rest_api):
...
+        while count < 20:
+            try:
+                time.sleep(1)
+                api_response = self.api_client.test_invoke_method(
+                    restApiId=rest_api, resourceId=resource_id, httpMethod="GET"
+                )
+                return api_response.get("body")
+            except ClientError as ce:
+                # This test is very unstable, any fixed wait time cannot guarantee a successful invocation
+                if "Invalid Method identifier specified" in ce.response.get("Error", {}).get("Message", ""):
```
mndeveci, line 349: nit: we can LOG the error message here so that it will be visible to person who is investigating possible failure of the test case?

tests/integration/testdata/sync/infra/after/Python/function/app.py
```diff
@@ -0,0 +1,43 @@
...
+        Lambda Context runtime methods and attributes
+
+        Context doc: https://docs.aws.amazon.com/lambda/latest/dg/python-context-object.html
+
+    Returns
+    ------
+    API Gateway Lambda Proxy Output Format: dict
+
+        Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
+    """
```
mndeveci, line 27: nit: can we clean up the comments so code will be easier to read?

tests/integration/testdata/sync/infra/after/Ruby/function/app.rb
```diff
@@ -0,0 +1,39 @@
...
+  #     'statusCode' and 'body' are required
+  #     # api-gateway-simple-proxy-for-lambda-output-format
+  #     Return doc: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html
+
+  # begin
+  #   response = HTTParty.get('http://checkip.amazonaws.com/')
+  # rescue HTTParty::Error => error
+  #   puts error.inspect
+  #   raise error
+  # end
```
mndeveci, line 29: Are we still having issues with this part? I saw that `require 'httparty'` is missing at the top of this file, maybe that might fix the issue here?

mndeveci (commented):
Thanks for addressing the earlier comments! Left couple of small ones.

tests/integration/testdata/sync/infra/after/Ruby/function/app.rb
qingchm: it was purposely removed but since now the ruby build bug is identified this will be added back!

tests/integration/testdata/sync/infra/template-python-after.yaml
```diff
@@ -0,0 +1,48 @@
...
+  Function:
+    Timeout: 3
+
+Resources:
+  HelloWorldFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: after/Python/function/
+      Handler: app.lambda_handler
+      Runtime: python3.7
```
CoshUS, line 14: nit: Are we able to set the runtime to 3.8 or 3.9? If so, 3.7 EOL will be less of a concern.

tests/integration/testdata/sync/infra/before/Python/function/app.py
```diff
@@ -0,0 +1,24 @@
...
+from layer_method import layer_method
+
+import requests
+
+
+def lambda_handler(event, context):
+    """Sample pure Lambda function that returns a message and a location"""
+
+    try:
+        ip = requests.get("http://checkip.amazonaws.com/")
```
CoshUS, line 11: This can be slow and is an external dependency. Is this needed or a simple message will be sufficient?

tests/integration/testdata/sync/infra/after/Ruby/function/app.rb
```diff
@@ -0,0 +1,22 @@
+require 'httparty'
+require 'json'
+require 'layer'
+
+def lambda_handler(event:, context:)
+  # Sample pure Lambda function that returns a message and a location
+
+  begin
+    response = HTTParty.get('http://checkip.amazonaws.com/')
```
CoshUS, line 9: Same comment as the python one.

CoshUS (commented):
Thanks for addressing previous comments! Added a few more regarding the Lambda Functions.

tests/integration/testdata/sync/infra/template-python-after.yaml
qingchm: Yes setting this to 3.8

tests/integration/testdata/sync/infra/before/Python/function/app.py
qingchm: We want to have a simple test of such dependencies, even though a simple message is sufficient in terms of testing the lambda function, increasing the complexity a bit is helpful for the sake of testing function with dependencies.

tests/integration/testdata/sync/infra/after/Ruby/function/app.rb
qingchm: Same as above, considering this to be a useful component of the test

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,363 @@
...
+LOG = logging.getLogger(__name__)
+
+
+@skipIf(SKIP_SYNC_TESTS, "Skip sync tests in CI/CD only")
+class TestSync(BuildIntegBase, SyncIntegBase, PackageIntegBase):
+    @classmethod
+    def setUpClass(cls):
+        # setup signing profile arn & name
+        cls.signing_profile_name = os.environ.get("AWS_SIGNING_PROFILE_NAME")
+        cls.signing_profile_version_arn = os.environ.get("AWS_SIGNING_PROFILE_VERSION_ARN")
```
mndeveci, line 45: Why do we need to set signing profile here?

tests/integration/sync/test_sync_command.py
```diff
@@ -0,0 +1,363 @@
...
+        stack_name = self._method_to_stack_name(self.id())
+        self.stacks.append({"name": stack_name})
+
+        # Run infra sync
+        sync_command_list = self.get_sync_command_list(
+            template_file=template_path,
+            code=False,
+            watch=False,
+            resource_id=None,
+            resource=None,
```
mndeveci, line 90: nit: can we clean up parameters which have `None` value (assuming default values are `None`)?

mndeveci (commented):
Thanks for handling all the feedback, left 2 small comments.

tests/integration/testdata/sync/infra/before/Python/function/app.py
CoshUS: What dependency are we testing here?
I agree that adding complexity and relevant dependencies to tests will increase coverage, but here, we are testing whether Lambda can make requests to and get IP back from `checkip.amazonaws.com` which does not involve SAM CLI.

qingchm: If the imports failed then the lambda would fail since the dependency wasn't included properly

CoshUS: In that case, can we switch to another library like `numpy` which doesn't require rely on a external remote website? This way we can minimize chances of having flaky tests.

mndeveci (approved):

tests/integration/testdata/sync/infra/before/Python/function/app.py
mndeveci: I changed external libraries with `numpy` and `ruby-statistics` so that there won't be an external call during tests

CoshUS (approved):
LGTM! Thanks for driving this and addressing the comments!

Issue #2968, opened 2021-06-17
sam-beta-cdk raises an error without any additional context


jabalsad:
### Description:
`sam-beta-cdk` raises an error when trying to `start-api` or `invoke` locally. I'm hoping I don't have to post the entire CDK stack template here because it is large.

### Steps to reproduce:
1. download and install: `brew install aws-sam-cli-beta-cdk` (osx)
2. run `sam-beta-cdk local start-api`

### Observed result:
```
› sam-beta-cdk local start-api --debug
2021-06-17 16:35:40,390 | Determining project type...
2021-06-17 16:35:40,390 | The project is a CDK project.
2021-06-17 16:35:40,434 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2021-06-17 16:35:40,439 | CDK Toolkit found at ./node_modules/aws-cdk/bin/cdk
2021-06-17 16:35:40,439 | Synthesizing CDK App
2021-06-17 16:35:40,439 | command: ['./node_modules/aws-cdk/bin/cdk', 'synth', '--no-staging', '-o', '.aws-sam/.cdk-out']
2021-06-17 16:35:51,424 | Cloud assembly synthed at .aws-sam/.cdk-out
2021-06-17 16:35:51,424 | _validate_cloud_assembly: .aws-sam/.cdk-out
2021-06-17 16:35:51,431 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': 'c87edbda-20b4-4869-9a03-c866c6fc96f7', 'installationId': '4d5c1751-ca4e-4c25-8c05-f33973bc5430', 'sessionId': '37acf4f4-8fd2-4b40-a68e-67100f4fa95a', 'executionEnvironment': 'CLI', 'ci': False, 'pyversion': '3.8.10', 'samcliVersion': '1.22.0.dev202104291816', 'metricSpecificAttributes': {'projectType': 'CDK'}, 'awsProfileProvided': False, 'debugFlagProvided': True, 'region': '', 'commandName': 'sam-beta-cdk local start-api', 'duration': 10994, 'exitReason': 'KeyError', 'exitCode': 255}}]}
2021-06-17 16:35:51,670 | Telemetry response: 200
Traceback (most recent call last):
  File "/usr/local/bin/sam-beta-cdk", line 8, in <module>
    sys.exit(cli())
[...]

Issue #2849, opened 2021-05-03
KeyError when running sam-beta-cdk build on projen typescript app 


mbuotidem:
### Description:

I have a cdk TypeScript project created using projen. My project contains two stacks, a codepipeline stack and the application stack. In the application stack, I'm creating a lambda resource. When I run `sam-beta-cdk build`, I get the error below:

```
misaac:~/environment/guichet (main) $ sam-beta-cdk build
Synthesizing CDK App
Traceback (most recent call last):
  File "samcli/__main__.py", line 12, in <module>
  File "click/core.py", line 829, in __call__
  File "click/core.py", line 782, in main
  File "click/core.py", line 1259, in invoke
  File "click/core.py", line 1066, in invoke
  File "click/core.py", line 610, in invoke
  File "click/decorators.py", line 73, in new_func
  File "click/core.py", line 610, in invoke
  File "samcli/lib/telemetry/metric.py", line 174, in wrapped
  File "samcli/lib/telemetry/metric.py", line 141, in wrapped
  File "samcli/lib/iac/utils/helpers.py", line 30, in wrapper
  File "samcli/lib/iac/cdk/plugin.py", line 116, in get_project
  File "samcli/lib/iac/cdk/plugin.py", line 207, in _get_project_from_cloud_assembly
  File "samcli/lib/iac/cdk/plugin.py", line 207, in <listcomp>
  File "samcli/lib/iac/cdk/plugin.py", line 222, in _build_stack
  File "samcli/lib/iac/cdk/plugin.py", line 280, in _build_resources_section
KeyError: '/home/ec2-user/environment/guichet/lambda'
[9116] Failed to execute script __main__

```

### Steps to reproduce:

Install projen and sam-cdk-beta

Create a projen TypeScript app - `projen new awscdk-app-ts `

Set ` "@aws-cdk/core:newStyleStackSynthesis": true` in `cdk.json`

Create a codepipeline stack in `main.ts`

Create an application stack that creates a lambda resource using `code: lambda.Code.fromAsset`, eg `code: lambda.Code.fromAsset('lambda')` with the folder lambda placed at the root

Create your lambda and a package.json file in the lambda folder

Initialize your application stack in the codepipeline stack

Run `sam-beta-cdk build`. 

You can also just use the repo here : https://github.com/mbuotidem/guichet/tree/sam-beta-cdk

### Observed result:
```
misaac:~/environment/guichet (sam-beta-cdk) $ sam-beta-cdk build --debug
[...]

Issue #3154, opened 2021-08-10
Native support for Terraform in SAM CLI (now in Beta)


praneetap:
### Terraform Support
Recently we added native support for CDK applications in the SAM CLI - you can see our [launch announcement](https://aws.amazon.com/blogs/compute/better-together-aws-sam-and-aws-cdk/) blog post and download the beta version of the CLI.

We would like to collect the community’s feedback on whether we should add **native support for Terraform** in the SAM CLI. Terraform users would be able to use the SAM CLI’s `build`, `package`, `local invoke`, and `start-api` commands. The `build` and `package` commands make it easy to create deployment packages or OCI images that are ready to be used in AWS Lambda. The `local invoke` and `start-api` commands makes it easy to emulate AWS Lambda and API Gateway in local Docker container for testing purposes.

With this change, the SAM CLI would parse HCL definitions to understand functions and APIs that can be locally tested and built. The SAM CLI would not write HCL, translate to, or force the use of CloudFormation. Deployments should still be performed using the Terraform toolkit.

*What feedback are we looking for?*

1. Is this something the community is interested in? _**Vote with 👍🏽 **._
2. The sam `build`, `package`, and `local invoke` commands rely on the `CodeUri` property of a function resource pointing to a local path where the source code for the Lambda function is stored. The SAM CLI automatically pulls dependencies, compiles if necessary, and prepares a deployment package for AWS Lambda. The package command today produces an output template that contains the S3 URL for the deployment package the SAM CLI prepared. _**How would you like this work with Terraform?**_

antonbabenko:
@Vlaaaaaaad I like the idea of bringing an alternative approach to the `package.py` with the main goal to make the module independent from Python and cross-platform right away.

(When we were making the [terraform-aws-lambda module](https://github.com/terraform-aws-modules/terraform-aws-lambda) module, `sam build` was far from feature-rich. Since there is support for running Lambda functions from a container image, I recommend using that instead of relying on `package.py` which is not perfect. See [container-image example](https://github.com/terraform-aws-modules/terraform-aws-lambda/tree/master/examples/container-image).)

[...]

PR #3496, opened 2021-11-25
fix: KeyError caused by not loading assets

pseudo-su:

I think this fixes the following issues

https://github.com/aws/aws-sam-cli/issues/2968

https://github.com/aws/aws-sam-cli/issues/2849

The current `_extract_assets` method doesn't seem to load the assets correctly, resulting in `KeyError`'s later down the line when trying to extract the stack from the cloud assembly

Reads assets from the manifest successfully when it's in this format

Screen Shot 2021-11-26 at 10 57 34 am

I have no idea

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/iac/cdk/cloud_assembly.py
+++ b/samcli/lib/iac/cdk/cloud_assembly.py
@@ -293,11 +293,13 @@ class CloudAssemblyStack:
         return None
 
     def _extract_assets(self) -> None:
-        for item in self.metadata.get(f"/{self.stack_name}", {}):
-            assert "type" in item
-            if item["type"] == ASSET_TYPE:
-                self._assets_by_id[item["data"]["id"]] = item["data"]
-                self._assets_by_path[item["data"]["path"]] = item["data"]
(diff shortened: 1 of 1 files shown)


jfuss:
@pseudo-su Thanks for the contribution, however this isn't something we can accept at this time. We are working in parts of the CDK beta into `develop` and have taking a little bit of a different direction. We are working through these issues still (https://github.com/aws/aws-sam-cli/pull/3492 works through some of the current gaps we have in the existing implementation that is released).

You can follow https://docs.aws.amazon.com/cdk/latest/guide/sam.html for how this currently works with the stable releases of SAM. 
The short version is:
```
cdk synth --no-staging > template.yaml
sam local invoke MyFunction12345678 --no-event
```
Replace MyFunction12345678 which your function logical Id.

If there are things you think the beta does well that the above experience doesn't, please open an issue. We are always looking for targeted feedback to make the cli better, especially as we look to expand support for different IAC: #3154

Issue #3463, opened 2021-11-12
Difference between the routeKey in event object GET /projects/<id> on local vs GET /projects/{id} on AWS 


samaybhavsar:
### Description:

I created a project using SAM CLI v1.33.0 with Runtime: nodejs14.x. When running on my local machine using sam local start-api the routeKey in event object has angle brackets e.g. GET /projects/<id>  while after deploying on AWS the routeKey in event object has curly brackets e.g.  GET /projects/{id} . I am using the routeKey for access control.

Another similar issue which I noticed is, when running local I get "queryStringParameters": {}, (key with blank object) while after deploying to AWS, the key is not present.

### Steps to reproduce:

- Create a sample project using sam init
- Set up an API endpoint with custom field in URL. e.g. project/{project_id}
- Deploy the same to AWS
- Compare event object for both

### Observed result:

- routeKey has angle brackets instead of curly brackets
- "queryStringParameters": {}, on local but when deployed on AWS, this key is missing

### Expected result:
Should have curly brackets in routeKey

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10
2. `sam --version`: 1.33.0
3. AWS region: ca-central-1

`Add --debug flag to command you are running`

torresxb1:
Thanks for reporting the issue! I was able to reproduce in SAM CLI v1.35 for HttpApi with the following template:
```
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello-world/
      Handler: app.lambdaHandler
      Runtime: nodejs14.x
      Architectures:
        - x86_64
      Events:
        HelloWorld:
          Type: HttpApi 
          Properties:
            ApiId: !Ref MyApi
            Path: /project/{project_id}
            Method: get

  MyApi:
    Type: AWS::Serverless::HttpApi
    Properties:
      StageName: Prod
```

In the handler, I just include the incoming event in the response and I do see `"routeKey":"GET /project/<project_id>"` and `"queryStringParameters":{}` when I run `sam local start-api` and curl the endpoint. When I curl the deployed endpoint, I see `"routeKey":"GET /project/{project_id}"` and `queryStringParameters` is not present. 

We've identified the issue, added it to our backlog, and will prioritize it accordingly. Thanks again for bringing this to our attention!

mndeveci:
[...]

PR #3497, opened 2021-11-26
fix: make routeKey and queryStringParameters for local HttpApi event match deployed one

torresxb1:
#3463

`routeKey` and `queryStringParameters` from HttpApi event when running `sam local start-api` don't match the event from the deployed api. More info is in the linked issue.

Makes them match (make `routeKey` use curly brackets and remove `queryStringParameters` if there's no query params)

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/local/events/test_api_event.py
```diff
@@ -551,7 +551,6 @@ def test_to_dict_with_defaults(self):
             "rawQueryString": None,
             "cookies": None,
             "headers": None,
-            "queryStringParameters": None,
```
awood45, line 554: Can we add a test case with the query string parameters present, since that is also a valid code path?

awood45 (changes requested):

tests/unit/local/events/test_api_event.py
torresxb1: I believe the test case above this one has query string parameters present

awood45: Ok - let me look at the context instead of just the diff :P

awood45 (approved):

mildaniel (approved):

hawflau (approved):

PR #3498, opened 2021-11-26
Add unique id for customer defined cdk functions

wchengru:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/sam_function_provider.py
```diff
@@ -235,12 +265,52 @@ def _convert_sam_function_resource(
             LOG.debug("Found Serverless function with name='%s' and ImageUri='%s'", name, imageuri)
 
         return SamFunctionProvider._build_function_configuration(
-            stack, name, codeuri, resource_properties, layers, inlinecode, imageuri, use_raw_codeuri
+            stack, function_id, name, codeuri, resource_properties, layers, inlinecode, imageuri, use_raw_codeuri
         )
 
+    @staticmethod
+    def _get_function_id(resource_properties: Dict, logical_id: str) -> str:
```
mildaniel, line 272: I believe the metadata generated by CDK is not in the properties, it is on the resource's root level. Can you double check this please?

wchengru: You are right, in the template it is. We pull it into the properties in the function provider: https://github.com/aws/aws-sam-cli/blob/ec9f5d89971b823fa01071c96b7de90ea5e5303a/samcli/lib/providers/sam_function_provider.py#L133

mildaniel: I see, thanks!

mildaniel:
Per our conversation with Jacob, can you add an integration test for invoking with this new function identifier? I started writing some CDK invoke integration tests in [this PR](https://github.com/aws/aws-sam-cli/pull/3492) that you can build off of.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -256,7 +258,12 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
         else:
             # Standard format
             LOG.debug("Environment variables overrides data is standard format")
-            overrides = self.env_vars_values.get(name, None)
+            # Precedence: function_id -> logical_id -> full_path, customer can use any of them
```
jfuss, line 261: Why is this the right order? Shouldn't the previous one (logical_id) be the first to be backwards compatible?

samcli/lib/providers/provider.py
```diff
@@ -29,6 +29,8 @@ class Function(NamedTuple):
     Named Tuple to representing the properties of a Lambda Function
     """
 
+    # Function id, can be Logical ID or any function identifier to define a function in specific IaC
```
jfuss, line 32: What about the function name? `name` can be the Function Name or the logical id (reading the comment below)

samcli/lib/providers/sam_function_provider.py
```diff
@@ -78,23 +78,52 @@ def get(self, name: str) -> Optional[Function]:
         if not name:
             raise ValueError("Function name is required")
 
-        # support lookup by full_path
+        get_f = None
```
jfuss, line 81: It's not clear what `get_f` is. Can you make a more descriptive variable name for this to improve readability?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -78,23 +78,52 @@ def get(self, name: str) -> Optional[Function]:
...
+        if not get_f:
+            # If function is not found by full path, search through all functions
+
+            found_fs = []
+
+            for f in self.get_all():
+                if name in (f.function_id, f.name, f.functionname):
+                    found_fs.append(f)
+
+            # If multiple functions are found, only return one of them
```
jfuss, line 96: Why is returning one of the the right UX? Shouldn't we only invoke if we are sure that is the function they want to invoke?

samcli/lib/providers/sam_function_provider.py
```diff
@@ -78,23 +78,52 @@ def get(self, name: str) -> Optional[Function]:
...
+
+        return get_f
 
     def _deprecate_notification(self, runtime: Optional[str]) -> None:
         if runtime in self._deprecated_runtimes:
             message = (
                 f"WARNING: {runtime} is no longer supported by AWS Lambda, "
                 "please update to a newer supported runtime. SAM CLI "
-                f"will drop support for all deprecated runtimes {self._deprecated_runtimes} on May 1st. "
+                f"will drop support for all deprecated runtimes {self._deprecated_runtimes} on May 1st, 2020. "
```
jfuss, line 126: The linked Github issue is closed. Not in this update, but we should probably remove this as it doesn't look like we support these anymore.

samcli/lib/providers/sam_function_provider.py
```diff
@@ -235,12 +265,52 @@ def _convert_sam_function_resource(
...
+        -------
+        str
+            The unique function id
+        """
+        resource_cdk_path = resource_properties.get("Metadata", {}).get("aws:cdk:path")
+
+        if not isinstance(resource_cdk_path, str) or not resource_cdk_path:
+            return logical_id
+
+        # aws:cdk:path metadata format: {stack_id}/{function_id}/Resource
```
jfuss, line 296: Are we sure this is the only format to `aws:cdk:path`? Maybe link out to docs?

jfuss (changes requested):
Can we add integ tests for the changes to ensure this works as we expect it too?

samcli/lib/providers/sam_function_provider.py
wchengru: Sure. I have added the cdk document about this in the comment.

samcli/lib/providers/sam_function_provider.py
wchengru: Thanks. I removed this line, and only kept the warning, and the Github issue link.

samcli/lib/providers/sam_function_provider.py
wchengru: Here the logic is kept the same as before for backward compatibility. I added a warning there because I noticed that in the nested stack case like this:
`StackA\FunctionA`
`StackA\NestedStackB\FunctionA`
The command `sam local invoke FunctionA` will only invoke one of them, and the order is also not promised. The change here is to (1) Make sure the invoked function is the same for every run. (2) Warn the user and provide the list of matched full paths.
I think if we stop the invoking, that will possible break the behaviour of some existing invocations usages.

samcli/lib/providers/sam_function_provider.py
wchengru: Thanks, I have updated this.

samcli/lib/providers/provider.py
wchengru: Looks like the `name` is function name is always logical ID. When the `FunctionName` is not defined, the function name is the same as logical id. I updated that comment to make it more clear.
https://github.com/aws/aws-sam-cli/blob/ec9f5d89971b823fa01071c96b7de90ea5e5303a/samcli/lib/providers/sam_function_provider.py#L340

https://github.com/aws/aws-sam-cli/blob/ec9f5d89971b823fa01071c96b7de90ea5e5303a/samcli/lib/providers/sam_function_provider.py#L129

samcli/commands/local/lib/local_lambda.py
wchengru: Thanks for pointing it out! I have fixed this order.

wchengru:
> Per our conversation with Jacob, can you add an integration test for invoking with this new function identifier? I started writing some CDK invoke integration tests in [this PR](https://github.com/aws/aws-sam-cli/pull/3492) that you can build off of.



> Per our conversation with Jacob, can you add an integration test for invoking with this new function identifier? I started writing some CDK invoke integration tests in [this PR](https://github.com/aws/aws-sam-cli/pull/3492) that you can build off of.

I will add integration 

> Per our conversation with Jacob, can you add an integration test for invoking with this new function identifier? I started writing some CDK invoke integration tests in [this PR](https://github.com/aws/aws-sam-cli/pull/3492) that you can build off of.

Thanks @mildaniel for providing the integration test templates. I created integration tests based on your templates.

mildaniel (approved):

samcli/lib/providers/sam_function_provider.py
jfuss: Seems reasonable. Thanks for explaining this. Could be worth an in comment code for future reference.

tests/integration/local/invoke/test_invoke_cdk_templates_with_function_id.py
```diff
@@ -0,0 +1,63 @@
...
+            # We don't actually use the build dir so we don't care if it's removed before another process finishes
+            shutil.rmtree(str(Path().joinpath(".aws-sam")))
+        except FileNotFoundError:
+            pass
+
+    @parameterized.expand([
+        ("StandardZipFunctionWithFunctionName", "ThisIsHelloWorldFunction", "LambdaWithFunctionName"),
+    ])
+    @pytest.mark.flaky(reruns=3)
+    def test_invoke_function_with_function_id(self, logical_id, function_name, function_id):
```
jfuss, line 49: Should we have any integs tests around the precedence order or different edge cases we may encounter with templates?

wchengru: I have added integration test cases for invoking functions in the nested stacks.
For the precedence on the env var overrides, the cases are unit tested:
https://github.com/aws/aws-sam-cli/pull/3498/files#diff-09b566488e9a422abe5c417dd6065f26119bd9bca9e9524e379c0534258a539fR223

PR #3500, opened 2021-11-28
feat: prevent deploy throttle with customizable interval for DescribeStack via env

BourgoisMickael:

#2671 

Related PR: #2748 

`sam deploy` often gets throttled because it sends 2 `DescribeStack` per seconds.
As commented in  https://github.com/aws/aws-sam-cli/pull/2748#issuecomment-932178818 this is problematic with a lot of pipeline in parallel that deploy stacks during 10 minutes.
Pipelines don't need real time updates so we can't to customize the interval between `DescribeStack`

Customize the interval between  `DescribeStack` via environment variable `SAM_CLI_DEPLOY_DESCRIBE_STACK_INTERVAL`

None

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


jbernales5 (approved):

samcli/lib/deploy/deployer.py
```diff
@@ -72,13 +73,18 @@
...
+
 
 class Deployer:
     def __init__(self, cloudformation_client, changeset_prefix="samcli-deploy"):
         self._client = cloudformation_client
         self.changeset_prefix = changeset_prefix
-        # 500ms of sleep time between stack checks and describe stack events.
-        self.client_sleep = 0.5
+        try:
+            self.client_sleep = int(os.getenv("SAM_CLI_DEPLOY_DESCRIBE_STACK_INTERVAL", str(DEFAULT_CLIENT_SLEEP)))
```
c2tarun, line 85: This seems like a good change, however we don't use this pattern (of reading from env var) to control command options. We read from env var for very few things like SAM_DEV, REGION, CREDS but not to configure command. Another problem with using env var is that this option is now hidden in the code and is difficult for users to find. If we do it as a command option, it'll be more helpful and easily discoverable.

Thoughts?

c2tarun (dismissed):

samcli/lib/deploy/deployer.py
BourgoisMickael: Is was asked to be an env var (https://github.com/aws/aws-sam-cli/pull/2748#discussion_r604445878) because another PR did that using a command line argument but and got abandoned. Also if this is accepted the documentation could be updated to document this var.

qingchm: I'm also leaning towards a command line option since for this specific value it would be easier for user to acknowledge (help text and documentation wise). This PR overall looks good to me, in terms of design I will engage a conversation within the team to see if we have a consensus on if env vars or options are preferrable! Would love to help push through this PR

jfuss:
@BourgoisMickael Sorry for the mix messaging you are getting from the team. I am trying to clarify with the team now. Personally, I don't think it matters a whole lot and we just need to align on one or the other to start. If we pick one and want the other later, Click has a way to default values from an Env Var so we shouldn't be stuck regardless.

BourgoisMickael:
Ok, let me know when there are news about this. 

I can also update the documentation later to describe this environment variable.

jfuss:
@BourgoisMickael Chatting with the team, we can accept the env var. If we want to move towards a cli option, we still can later on through click. 

The changes you have proposed seem reasonable but I would like if you could move the env var into the command and pass this through to the deployer. This keeps the configuration on the "cli" level while making the deployer more like a lib than reading the env directly. 

The thing I am thinking through is should we make the env var more generic? `SAM_CLI_DEPLOY_DESCRIBE_STACK_INTERVAL` is fine but I maybe it is worth doing something like `SAM_CLI_THROTTLE_RATE` (naming is hard). If we can find a good general name, then we could re-use this configuration in other places. Maybe it's not worth it, but would like to here your thoughts.

BourgoisMickael:
I renamed the variable to `SAM_CLI_POLL_DELAY`. I think it describes best what it does and it could be reused for other poller like for the `sam delete`

samcli/commands/deploy/deploy_context.py
```diff
@@ -142,7 +142,9 @@ def run(self):
                 s3_client, self.s3_bucket, self.s3_prefix, self.kms_key_id, self.force_upload, self.no_progressbar
             )
 
-        self.deployer = Deployer(cloudformation_client)
+        self.deployer = Deployer(
+            cloudformation_client, client_sleep=os.getenv("SAM_CLI_POLL_DELAY", str(DEFAULT_CLIENT_SLEEP))
```
moelasmar, line 146: Could you please move the env variable reading to the deploy command, and pass its value as a parameter to the deploy_context. This will make it easier for us later to refactor this code, and depend on a command options.

moelasmar (changes requested):

moelasmar:
Thanks @BourgoisMickael for your contribution. This PR looks good to me, just a small change requested.

samcli/commands/deploy/deploy_context.py
BourgoisMickael: It's done

samcli/lib/deploy/deployer.py
```diff
@@ -409,6 +416,8 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
                 if self._check_stack_not_in_progress(stack_status):
                     stack_change_in_progress = False
                     break
+                # Reset retry attempts if iteration is a success to use client_sleep again
+                retry_attempts = 0
```
BourgoisMickael, line 420: If found a bug at https://github.com/aws/aws-sam-cli/blob/d1db668924ba74b0f71813da18591de1b80b4142/samcli/lib/deploy/deployer.py#L371-L372

Which causes to ignore the `client_sleep` and always send requests without delay if there was one exception in previous iterations, even if after the retry all the other iterations are successful.

With this change, the function stops only if there are consecutive retries exceptions.

But we could also consider the retry attemps globally no matter if there was successes in between.
In that case I could add a condition on the sleep to skip only if the last iteration was in exception.

Let me know what's best @moelasmar @jfuss

moelasmar: I think it is better to stop only if there consecutive retries exceptions, this increases the probability that the customer can finish the `sam deploy` command successfully.

I have a question why we skip the sleep time after some exception, I think waiting some times help to resolve some exceptions like the throttling exceptions, and this will increase the command success probability. May be the cons of this approach is the command will take a longer time to finish (either success or fail).

jfuss: I think that is ok. It might not be ideal in all cases but we can always adjust this in the future (I would think). Let me ponder this a little more to make sure that all seems reasonable. 

@BourgoisMickael Sorry I missed the update on the PR.

jfuss (approved):
LGTM. Thanks for taking the time to work on this.

moelasmar (approved):
thanks @BourgoisMickael for your contribution

jfuss:
@BourgoisMickael Thanks for pushing this through. There is no further action on you but I need to figure out how/what docs need to get updated. I will merge once I kick off that process so we don't accidentally release without the doc updates.

smurphnturf:
Love this thank you!! :)

BourgoisMickael:
I just found a bug this feature introduces https://github.com/aws/aws-sam-cli/issues/3910

Issue #3494, opened 2021-11-24
Sam sync removes node_modules in lambda layer


cshenrik:
### Description: 
I have a sam application that contains a nodejs14.x Lambda function that references a Lambda layer with a library installed in `node_modules`. The first time I run `sam sync --watch`, all works fine. When I stop the watcher and restart it, it deletes the `node_modules` folder in `.aws-sam/auto-dependency-layer/SharedLayer/nodejs/node_modules` (where `SharedLayer` is my Lambda Layer). Consequently my Lambda function is broken.

My workaround is to modify the `manifest_hash` value of the Lambda layer in `.aws-sam/build.toml`, and make a dummy change in the code of the Lambda layer. This triggers the watcher, so it correctly builds and syncs the layer + dependent functions. 

### Steps to reproduce:
1) Clone this repo: https://github.com/cshenrik/accelerate-bug.
1) Run `sam sync --watch --stack-name <STACK_NAME>`.
1) Wait until the initial sync is done.
1) Exit the watcher by hitting ctrl + c.
1) Run `sam sync --watch --stack-name <STACK_NAME>` again.

Then `.aws-sam/auto-dependency-layer/SharedLayer/nodejs/node_modules` has been deleted, and the Lambda function is broken.  

### Observed result:
This is the output from running the second `sam sync` in step 5 (though with --debug):
```2021-11-24 20:34:31,079 | No Parameters detected in the template
2021-11-24 20:34:31,099 | 3 stacks found in the template
2021-11-24 20:34:31,099 | No Parameters detected in the template
2021-11-24 20:34:31,113 | 3 resources found in the stack 
2021-11-24 20:34:31,113 | No Parameters detected in the template
2021-11-24 20:34:31,126 | No Parameters detected in the template
2021-11-24 20:34:31,140 | --base-dir is not presented, adjusting uri shared/ relative to /home/henrik/code/accelerate-bug/template.yaml
2021-11-24 20:34:31,140 | Found Serverless function with name='SayHelloFunction' and CodeUri='say-hello/'
2021-11-24 20:34:31,140 | --base-dir is not presented, adjusting uri say-hello/ relative to /home/henrik/code/accelerate-bug/template.yaml
2021-11-24 20:34:31,140 | No Parameters detected in the template
2021-11-24 20:34:31,153 | --base-dir is not presented, adjusting uri shared/ relative to /home/henrik/code/accelerate-bug/template.yaml
[...]

PR #3507, opened 2021-12-01
fix: always combine dependencies for layer build

mndeveci:
#3494

Using application wide setting for combining dependencies is wrong when it comes to building layers.

Combining dependencies, 
- For function builds, it will continue to use what is been set inside the `AppBuilder` instance. 
- For layer builds, it will always be `True` since we don't need to separate dependencies.

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


CoshUS (approved):

qingchm (approved):
LGTM!

PR #3518, opened 2021-12-06
Add new metrics projectType

wchengru:
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation


moelasmar (approved):

tests/integration/telemetry/test_experimental_metric.py
```diff
@@ -51,7 +52,11 @@ def test_must_send_experimental_metrics_if_experimental_command(self):
                             "debugFlagProvided": ANY,
                             "region": ANY,
                             "commandName": ANY,
-                            "metricSpecificAttributes": {"experimentalAccelerate": True, "experimentalAll": False},
+                            "metricSpecificAttributes": {
+                                "projectType": "CFN",
```
moelasmar, line 56: I think there is no template used for `sam traces` command, so I think we should not use project type here.

samcli/lib/telemetry/metric.py
```diff
@@ -138,14 +140,21 @@ def wrapped(*args, **kwargs):
...
         try:
             ctx = Context.get_current_context()
+            try:
+                template_dict = ctx.template_dict
+            except AttributeError:
+                LOG.debug("Ignoring CDK project check as template is not provided in context.")
+                template_dict = {}
+            project_type = ProjectTypes.CDK.value if is_cdk_project(template_dict) else ProjectTypes.CFN.value
+            metric_specific_attributes = get_all_experimental_statues() if ctx.experimental else {}
+            metric_specific_attributes["projectType"] = project_type
```
moelasmar, line 150: should the property name `projecttype` ?? .. this value will be part of the JSON value of the column `metricSpecificAttributes` in the metrics data, and I think our queries is sensitive if the value is an uppercase or lowercase.

moelasmar (changes requested):
Could you please check the failing unit test case

samcli/lib/telemetry/metric.py
hawflau: our database is case-insensitive so it doesn't really matter.

samcli/lib/telemetry/metric.py
```diff
@@ -138,14 +140,21 @@ def wrapped(*args, **kwargs):
 
         try:
             ctx = Context.get_current_context()
+            try:
+                template_dict = ctx.template_dict
+            except AttributeError:
+                LOG.debug("Ignoring CDK project check as template is not provided in context.")
+                template_dict = {}
+            project_type = ProjectTypes.CDK.value if is_cdk_project(template_dict) else ProjectTypes.CFN.value
```
hawflau, line 148: What if a user run a command at the location where there is neither CFN nor CDK project?

tests/integration/telemetry/test_experimental_metric.py
wchengru: Updated the metric adding logic, skip adding project type if the template is not provided.

samcli/lib/telemetry/metric.py
wchengru: Similar to Mohamed's comment, if the template is not provided, we don't add the project type metric. For now, we only have these two types of templates, so if template is provided, it can be either CFN, or CDK.

hawflau (approved):

moelasmar (approved):

PR #3522, opened 2021-12-07
test: Add integration tests for sam local commands for CDK projects

mildaniel:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/local/start_api/test_start_api_cdk_template.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pytest
+import requests
+from parameterized import parameterized_class
+
+from tests.integration.local.start_api.start_api_integ_base import StartApiIntegBaseClass
+
+
+@parameterized_class(
+    ("template",),
```
hawflau, line 11: Comparing with `test_start_api.py`, should this be `tempalte_class` instead?

tests/integration/local/start_api/test_start_api_cdk_template.py
```diff
@@ -0,0 +1,83 @@
...
+    [
+        (Path("/testdata/start_api/cdk/template-rest-api.yaml"),),
+        (Path("/testdata/start_api/cdk/template-open-api.yaml"),),
+        (Path("/testdata/start_api/cdk/template-http-api.yaml"),),
+    ],
+)
+class TestStartAPICDKTemplateRestAPI(StartApiIntegBaseClass):
+    template_path = "/testdata/start_api/cdk/template-rest-api.yaml"
```
hawflau, line 19: Suppose it is `template_path` in `@parameterized_class`, do we to override it here?

hawflau (changes requested):

tests/integration/local/start_api/test_start_api_cdk_template.py
mildaniel: Good catch. Just forgot to remove this line when I was cleaning up.

tests/integration/local/start_api/test_start_api_cdk_template.py
mildaniel: You're right, it should be `template_path` instead.

tests/integration/testdata/invoke/cdk/cdk_template.yaml
```diff
@@ -1,73 +1,434 @@
...
+                  - Fn::Split:
+                      - "||"
+                      - Ref: AssetParametersfa3b13f8e96c4bf14d4c6a47337be97af700c05a39b382dd2b756f3f50ffd6b9S3VersionKey5F29F473
+      Role:
+        Fn::GetAtt:
+          - timeoutfunctionServiceRoleA84BD66C
+          - Arn
+      Handler: app.timeout_handler
+      Runtime: python3.8
+      Timeout: 5000
```
wchengru, line 156: The timeout unit is second. It should be 5 here.

tests/integration/local/invoke/invoke_cdk_templates.py
```diff
@@ -15,13 +16,122 @@
...
+        wall_clock_cli_duration = end - start
+
+        process_stdout = stdout.strip()
+
+        # validate the time of the cli (timeout is set to 5s)
+        self.assertGreater(wall_clock_cli_duration, 5)
+        self.assertLess(wall_clock_cli_duration, 20)
+
+        self.assertEqual(return_code, 0)
+        self.assertEqual(
```
wchengru, line 72: Looking at the function, it only returns after 10 seconds, which means the function is not timed out but finished running. Should we add a timeout case, and rename this one?

tests/integration/testdata/invoke/cdk/cdk_template.yaml
```diff
@@ -1,73 +1,434 @@
...
+            Ref: ModeEnvVariable
+      FunctionName: HelloWorldFunction
+      Handler: app.hello_world_handler
+      Runtime: python3.8
+    DependsOn:
+      - helloworldfunctionServiceRole306F1141
+    Metadata:
+      aws:cdk:path: RootStack/helloworld-function/Resource
+      aws:asset:path: ./lambda_code
+      aws:asset:original-path: ./
```
moelasmar, line 55: original-path will be removed in the next CDK release, because of this issue https://github.com/aws/aws-cdk/issues/17706 .. anyway, we do not use it any where right now.

moelasmar (commented):
can you add test cases to cover Lambda function layer from asset - [example](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda.LayerVersion.html#example)

 and bundled Lambda function code - [Example](https://github.com/aws-samples/cdk-build-bundle-deploy-example/blob/main/cdk-bundle-go-lambda-example/lib/api-stack.ts)

tests/integration/local/invoke/invoke_cdk_templates.py
mildaniel: Fixing the timeout value you pointed out should fix this issue.

tests/integration/testdata/invoke/cdk/cdk_template.yaml
mildaniel: Shouldn't make a difference, but I'll remove it to keep things up-to-date.

wchengru (approved):

moelasmar (approved):

hawflau (approved):

PR #3529, opened 2021-12-15
fix: allow sam to deploy templates contain CFN Lambda Function resources with Image package type, and its code is local image uri.

moelasmar:
sam deploy fail to deploy the a template that contains Lambda function with Image Package Type, and refer to a local image uri. example template:
```
Resources:
  HelloWorldFunctionRole:
    Type: AWS::IAM::Role
     ......
  HelloWorldFunction:
    Type: AWS::Lambda::Function
    Properties:
      PackageType: Image
      .... 
      Code:
        ImageUri: func217e03a4:latest
```

This PR also fixes sam build for the same template.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -279,7 +279,7 @@ def update_template(
         for logical_id, resource in template_dict.get("Resources", {}).items():
 
             full_path = get_full_path(stack.stack_path, logical_id)
-            has_build_artifact = full_path in built_artifacts
+            has_build_artifact = full_path in built_artifacts and built_artifacts[full_path]
```
mndeveci, line 282: Should we change `built_artifacts[full_path]` to `built_artifacts.get(full_path)` for not getting any exception if `full_path` is not in the dict?

samcli/lib/build/app_builder.py
```diff
@@ -359,6 +358,14 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
         docker_build_target = metadata.get("DockerBuildTarget", None)
         docker_build_args = metadata.get("DockerBuildArgs", {})
 
+        if not dockerfile or not docker_context:
+            LOG.info(
+                "Skip Building %s function, as it does not contain either Dockerfile or DockerContext "
+                "metadata properties.",
+                function_name,
+            )
+            return ""
```
mndeveci, line 367: nit: should we return `None` or empty string?

torresxb1: should it raise an exception here instead?

samcli/lib/build/app_builder.py
moelasmar: the first half of the condition is to check that full path in the built_artifacts 
`full_path in built_artifacts`

samcli/lib/build/app_builder.py
moelasmar: @mndeveci .. I returned empty string as the method was defined to return str not Optional[str] .. but I believe I can make it None

moelasmar: @torresxb1 .. I think we should not raise an exception, the function can be referring to a remote or local image uri, and so no need to include the  image build related metadata, so sam will skip building it.

tests/integration/testdata/buildcmd/template_cfn_local_prebuilt_image.yaml
```diff
@@ -0,0 +1,16 @@
+Parameteres:
```
torresxb1, line 1:
```suggestion
Parameters:
```

tests/integration/testdata/buildcmd/template_local_prebuilt_image.yaml
```diff
@@ -0,0 +1,20 @@
+AWSTemplateFormatVersion : '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+
+Parameteres:
```
torresxb1, line 4:
```suggestion
Parameters:
```

samcli/lib/build/app_builder.py
mndeveci: I see, we can also simplify it by replacing whole line with `bool(built_artifacts.get(full_path))` but leave it to you

wchengru (approved):

hawflau (approved):

okofish:
Is it intentional that SAM CLI is now requiring both `DockerContext` **and** `Dockerfile` metadata entries to be present? This change broke `sam build` for my template that had only `DockerContext` and no `Dockerfile`.

I'm not clear on the reasoning here but it seems like it would be better to require either property rather than both.

PR #3549, opened 2021-12-23
feat: allow invoking built CDK synthesized templates

moelasmar:
allow invoking built CDK synthesized templates

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -277,8 +278,8 @@ def update_template(
         template_dict = stack.template_dict
 
         for logical_id, resource in template_dict.get("Resources", {}).items():
-
-            full_path = get_full_path(stack.stack_path, logical_id)
+            resource_id = ResourceMetadataNormalizer.get_resource_id(resource, logical_id)
+            full_path = get_full_path(stack.stack_path, resource_id)
```
mndeveci, line 282: nit: let's keep using `logical_id` or `physical_id` for resource ids in the template.

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -44,18 +47,21 @@ def normalize(template_dict):
...
+            if not is_normalized:
+                asset_property = resource_metadata.get(ASSET_PROPERTY_METADATA_KEY)
+                if asset_property == IMAGE_ASSET_PROPERTY:
+                    asset_metadata = ResourceMetadataNormalizer._extract_image_asset_metadata(resource_metadata)
+                    ResourceMetadataNormalizer._update_resource_metadata(resource_metadata, asset_metadata)
+                    # For image-type functions, the asset path is expected to be the name of the Docker image.
+                    # When building, we set the name of the image to be the logical id of the function.
+                    asset_path = logical_id.lower()
+                else:
+                    asset_path = resource_metadata.get(ASSET_PATH_METADATA_KEY)
```
mndeveci, line 60: This for loop will go through all resources in the template and update their `asset_path`. Is this going to work only for function types or for all resource types?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -144,3 +150,53 @@ def _update_resource_metadata(metadata, updated_values):
...
+            LOG.warning(
+                "Cannot detect function id from aws:cdk:path metadata '%s', using default logical id", resource_cdk_path
+            )
+            return logical_id
+
+        cdk_resource_id = cdk_path_partitions[-2]
+
+        # Check if the Resource is nested Stack
+        if resource_properties.get("Type", "") == "AWS::CloudFormation::Stack" and cdk_resource_id.endswith(
+            ".NestedStack"
```
mndeveci, line 198: Please use constant values for some of the string values here (like `AWS::Cloudformation::Stack` or for `.NestedStack`).

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -144,3 +150,53 @@ def _update_resource_metadata(metadata, updated_values):
...
+        resource_cdk_path = resource_metadata.get(RESOURCE_CDK_PATH_METADATA_KEY)
+
+        if not isinstance(resource_cdk_path, str) or not resource_cdk_path:
+            return logical_id
+
+        # aws:cdk:path metadata format of functions: {stack_id}/{function_id}/Resource
+        # Design doc of CDK path: https://github.com/aws/aws-cdk/blob/master/design/construct-tree.md
+        cdk_path_partitions = resource_cdk_path.split("/")
+
+        if len(cdk_path_partitions) < 2:
```
mndeveci, line 188: Logging some debug messages about this values could be useful when trying to find the root cause of a possible error in the future

samcli/lib/build/app_builder.py
moelasmar: resource_id is not always logical Id, this is a new introduced id. In case of CDK (or generally any IaC), it will be the CDK id (or generally the IaC id), in case of CFN, it will be the resource Logical Id

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: this loop is for all resources that has the CDK path metadata, so generally CDK will add the path metadata and set the resource property that will be replaced with the CDK path metadata value. This applied for any CDK L2 constructs that use assets

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: done

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: done

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -144,3 +154,65 @@ def _update_resource_metadata(metadata, updated_values):
...
+
+        if isinstance(customer_defined_id, str) and customer_defined_id:
+            LOG.debug(
+                "Sam customer defined id is more priority than other IDs. Customer defined id for resource %s is %s",
+                logical_id,
+                customer_defined_id,
+            )
+            return customer_defined_id
+
+        resource_cdk_path = resource_metadata.get(RESOURCE_CDK_PATH_METADATA_KEY)
```
mndeveci, line 188: This method and filename is named generic (`resource_metadata_normalizer` and `get_resource_id`), but method itself contains CDK specific stuff. This feels like it should be a generic (let's say `ResourceMetadataNormalizer`) class and we should be assigning different types depending on the project type (`CfnResourceMetadataNormalizer` or `CdkResourceMetadataNormalizer`), what do you think?

samcli/lib/build/app_builder.py
moelasmar: I agree, I renamed it to be `resource_iac_id`

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: I am working on another PR that is related to this part (IaC ids) to use it in other commands. Can I work on this refactoring after I merge this PR, and the one I am working on, as I do not want to miss things, it will be easier to refactor everything together

mndeveci (approved):

mildaniel (approved):

PR #3554, opened 2021-12-29
feat: add CDK synthesized templates support to sam deploy and package commands 

moelasmar:
add CDK synthesized templates support to sam deploy and package commands 

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/commands/package/package_context.py
```diff
@@ -135,8 +136,12 @@ def run(self):
             raise PackageFailedError(template_file=self.template_file, ex=str(ex)) from ex
 
     def _export(self, template_path, use_json):
-        template = Template(template_path, os.getcwd(), self.uploaders, self.code_signer)
+        template = Template(template_path, os.getcwd(), self.uploaders, self.code_signer, normalize_template=True)
         exported_template = template.export()
+        updated_template = copy.deepcopy(exported_template)
+        ResourceMetadataNormalizer.normalize(updated_template, True)
+        if updated_template.get("Parameters", None) and isinstance(updated_template.get("Parameters", None), dict):
```
mndeveci, line 143: nit: no need `None` for `get` operation

samcli/commands/package/package_context.py
```diff
@@ -135,8 +136,12 @@ def run(self):
             raise PackageFailedError(template_file=self.template_file, ex=str(ex)) from ex
 
     def _export(self, template_path, use_json):
-        template = Template(template_path, os.getcwd(), self.uploaders, self.code_signer)
+        template = Template(template_path, os.getcwd(), self.uploaders, self.code_signer, normalize_template=True)
         exported_template = template.export()
+        updated_template = copy.deepcopy(exported_template)
+        ResourceMetadataNormalizer.normalize(updated_template, True)
```
mndeveci, line 142: Where is `updated_template` is used? I saw it is been copied and send it to `ResourceMetadataNormalizer` but we are not using it afterwards.

samcli/lib/package/artifact_exporter.py
```diff
@@ -144,6 +148,8 @@ def __init__(
             self.template_dir = template_dir
             self.code_signer = code_signer
         self.template_dict = yaml_parse(template_str)
+        if normalize_template:
+            ResourceMetadataNormalizer.normalize(self.template_dict)
```
mndeveci, line 152: We are normalizing template dictionary here and above, is one of them redundant?

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -67,6 +70,19 @@ def normalize(template_dict):
...
+            parameters = template_dict.get("Parameters", {})
+
+            for parameter_name, parameter_value in parameters.items():
+                if (
+                    parameter_name.startswith("AssetParameters")
+                    and "Default" not in parameter_value
+                    and parameter_value.get("Type", "") == "String"
+                    and f'"Ref": "{parameter_name}"' not in resources_as_string
+                ):
+                    parameter_value["Default"] = " "
```
mndeveci, line 84: Why we needed to set default to empty string to these parameters? Adding a log line could be helpful when debugging

moelasmar: these are the auto generated CDK Asset parameters, I preferred to set default value instead of removing them to avoid removing a customer parameter that follow the same name pattern, so the customer can use sam deploy or package commands without providing values for these parameters although they are not used.

I will add some comments, and logging to this code

samcli/commands/package/package_context.py
moelasmar: it is used to clone the normalized parameters to the exported template, any way I changed this code to remove the duplicate normalizing step.

samcli/lib/package/artifact_exporter.py
moelasmar: resolved

samcli/commands/package/package_context.py
moelasmar: done

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: added comments and debugging logs

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -70,18 +70,33 @@ def normalize(template_dict, normalize_parameters=False):
...
 
+        # This is a work around to allow the customer to use sam deploy or package commands without the need to provide
+        # values for the CDK auto generated asset parameters. The suggested solution is to let CDK add some metadata to
+        # the autogenerated parameters, so sam can skip them safely.
+
+        # Normalizing CDK auto generated parameters. Search for parameters that meet these conditions:
+        # 1- parameter name starts with AssetParameters
+        # 2- parameter type is string
+        # 3- there is no reference to this parameter any where in the template resources
+        # 4- there is no default value for this parameter
```
mndeveci, line 81: Can we add additional checks for these parameters? I remember they have random chars after `AssetParameters`, can we create a pattern and check for that?

moelasmar: done

samcli/lib/package/image_utils.py
```diff
@@ -44,6 +44,10 @@ def tag_translation(image, docker_image_id=None, gen_tag="latest"):
 
     # NOTE(sriram-mv): Checksum truncation Length is set to 12
     _id = docker_image_id.split(":")[1][:SHA_CHECKSUM_TRUNCATION_LENGTH]
-    name, tag = image.split(":")
+    if ":" in image:
```
mildaniel, line 47: Thanks for fixing this.

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -67,6 +71,39 @@ def normalize(template_dict):
...
+        # We set an empty string as default value for the matching parameters, so the customer can use sam deploy or
+        # package commands without providing values for the auto generated parameters, as these parameters are not used
+        # in SAM (sam set the resources paths directly, and does not depend on template parameters)
+        if normalize_parameters and is_cdk_project(template_dict):
+            resources_as_string = json.dumps(resources)
+            parameters = template_dict.get("Parameters", {})
+
+            default_value = " "
+            cdk_asset_parameter_pattern = re.compile(
+                "AssetParameters[0-9a-f]{64}(?:S3Bucket|S3VersionKey|ArtifactHash)[0-9A-F]{8}"
```
mndeveci, line 93: 👏🏼  One verification; first set of chars are lower case (since it is `[0-9a-f]`) and second set of chars are upper case (since it is `[0-9A-F]`)?
And do we want to keep this at file level so that regex won't be compiled every time we call this method?

moelasmar: Although all generated Asset parameters I saw in CDK examples are following this pattern (the first 64 characters contain only small letters, and last 8 characters contain only capital letters), but I found that in CDK repo they are using [this pattern](https://github.com/aws/aws-cdk/blob/b1ecd3d49d7ebf97a54a80d06779ef0f0b113c16/packages/%40aws-cdk/assert-internal/lib/canonicalize-assets.ts#L19), so I updated the pattern and moved to file level.

mndeveci (approved):

mildaniel (approved):

PR #3580, opened 2022-01-12
chore: Code sync integration tests

qingchm:
#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/sync/test_sync_command.py
```diff
@@ -33,8 +33,9 @@
 SKIP_SYNC_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
 IS_WINDOWS = platform.system().lower() == "windows"
 CFN_SLEEP = 3
+LAMBDA_SLEEP = 3
```
mndeveci, line 36: Can we add some explanation why we need these sleeps? It is important to keep this recorded so that people can understand and adjust these values if it is needed

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,151 @@
...
+            api_resource = self.api_client.get_resources(restApiId=rest_api)
+            for item in api_resource.get("items"):
+                if "GET" in item.get("resourceMethods", {}):
+                    resource_id = item.get("id")
+                    break
+            self.api_client.flush_stage_cache(restApiId=rest_api, stageName="prod")
+        except ClientError as ce:
+            LOG.error(ce)
+        except Exception as e:
+            LOG.error(e)
```
CoshUS, line 103: Same as above.

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,151 @@
...
+        while count < 20:
+            try:
+                time.sleep(1)
+                api_response = self.api_client.test_invoke_method(
+                    restApiId=rest_api, resourceId=resource_id, httpMethod="GET"
+                )
+                return api_response.get("body")
+            except ClientError as ce:
+                if count == 20:
+                    LOG.error(ce)
```
CoshUS, line 116: On 20th failure, can we raise the exception instead of just logging it and returning empty string? The test failure message would be `"" != "expected output"` instead of the exception directly. This will be harder to find the root cause as we need to go back and check the error log.

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,151 @@
...
+                # This test is very unstable, any fixed wait time cannot guarantee a successful invocation
+                if "Invalid Method identifier specified" in ce.response.get("Error", {}).get("Message", ""):
+                    if count == 20:
+                        LOG.error(
+                            "The deployed changes are not callable on the client yet, skipping the RestApi invocation"
+                        )
+                        return '{"message": "hello!!"}'
```
CoshUS, line 123: Why do we return '{"message": "hello!!"}' when deployed changes are not callable?  Shouldn't we fail in this case?

tests/integration/sync/test_sync_code.py
```diff
@@ -0,0 +1,207 @@
...
+TEMPLATE_PATH = ""
+
+
+@skipIf(SKIP_SYNC_TESTS, "Skip sync tests in CI/CD only")
+class TestSyncCode(SyncIntegBase):
+    @pytest.fixture(autouse=True, scope="class")
+    def sync_code_base(self):
+        global TEMP_DIR
+        global STACK_NAME
+        global TEMPLATE_PATH
```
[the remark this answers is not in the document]
CoshUS: All caps is also for constants and not global variables.

CoshUS (changes requested):

tests/integration/sync/test_sync_code.py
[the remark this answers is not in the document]
qingchm: No, those variables will not be shared between instances. Since each test is an object by there own, setting self.var = value in the fixture cannot be applied to all test instances (it's not like setUp() where it's called upon each test object instantiation.) So it needs to be static and global. I can try to make it static class variable like TestSyncCode.temp_dir and see if it works equivalently (I would expect so)

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,151 @@
...
+
+    def _get_lambda_response(self, lambda_function):
+        try:
+            lambda_response = self.lambda_client.invoke(FunctionName=lambda_function, InvocationType="RequestResponse")
+            payload = json.loads(lambda_response.get("Payload").read().decode("utf-8"))
+            return payload.get("body")
+        except ClientError as ce:
+            LOG.error(ce)
+        except Exception as e:
+            LOG.error(e)
```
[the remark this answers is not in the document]
qingchm: Changing all error to directly raise, the only concern that I have is that it would increase the frequency which infra integration tests fail. With the while loop attempt if any one of the attempts gets an client error the whole test case simply fails. For now let's just raise any error and observe the stability of these tests.

tests/integration/sync/sync_integ_base.py
qingchm: Yep changing it to raise as well. But the thing is in real life customers are not guaranteed to have their rest api updated within 20 seconds. So this might become a flaky test in the future where we see failures randomly

tests/integration/sync/sync_integ_base.py
qingchm: Yep changing it to failure now

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,151 @@
...
+                    LOG.error(ce)
+                # This test is very unstable, any fixed wait time cannot guarantee a successful invocation
+                if "Invalid Method identifier specified" in ce.response.get("Error", {}).get("Message", ""):
+                    if count == 20:
+                        LOG.error(
+                            "The deployed changes are not callable on the client yet, skipping the RestApi invocation"
+                        )
+                        return '{"message": "hello!!"}'
+            except Exception as e:
+                if count == 20:
```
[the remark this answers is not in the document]
qingchm: Sure I can put it as a var

tests/integration/sync/test_sync_code.py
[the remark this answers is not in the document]
mndeveci: You can use `request` object and then assign values to the `cls` variable. I've managed to do an example in draft `sam logs` PR: https://github.com/aws/aws-sam-cli/pull/3607/files#diff-c8df8b464cdea941c1cb9dbcdb4b335317ded559bf8835e966e482f7c5694bedR51

samcli/commands/build/build_context.py
```diff
@@ -155,7 +155,7 @@ def __exit__(self, *args):
     def get_resources_to_build(self):
         return self.resources_to_build
 
-    def run(self):
+    def run(self, sync=False):
```
CoshUS, line 158: Can we have a separate PR for this?

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,143 @@
...
+            return physical_ids
+        except Exception:
+            raise
+
+    def _get_lambda_response(self, lambda_function):
+        try:
+            lambda_response = self.lambda_client.invoke(FunctionName=lambda_function, InvocationType="RequestResponse")
+            payload = json.loads(lambda_response.get("Payload").read().decode("utf-8"))
+            return payload.get("body")
+        except Exception:
```
CoshUS, line 87: We can remove these try blocks since we are just raising the exception.

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,143 @@
...
 
     def setUp(self):
+        self.cfn_client = boto3.client("cloudformation")
+        self.ecr_client = boto3.client("ecr")
+        self.lambda_client = boto3.client("lambda")
+        self.api_client = boto3.client("apigateway")
+        self.sfn_client = boto3.client("stepfunctions")
+        self.sns_arn = os.environ.get("AWS_SNS")
+        self.stacks = []
+        time.sleep(CFN_SLEEP)
```
mndeveci, line 41: Sorry if I miss this earlier, but why we are waiting in the `setUp` method where we haven't call/run anything yet?

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,143 @@
...
+        self.s3_prefix = uuid.uuid4().hex
         super().setUp()
 
     def tearDown(self):
+        shutil.rmtree(os.path.join(os.getcwd(), ".aws-sam", "build"), ignore_errors=True)
```
mndeveci, line 46: Do we need to remove `build.toml` file as well, which is at `.aws-sam/build.toml` location?

tests/integration/sync/test_sync_code.py
```diff
@@ -0,0 +1,207 @@
...
+            tags="integ=true clarity=yes foo_bar=baz",
+        )
+        sync_process_execute = run_command_with_input(sync_command_list, "y\n".encode())
+        self.assertEqual(sync_process_execute.process.returncode, 0)
+
+        # CFN Api call here to collect all the stack resources
+        self.stack_resources = self._get_stacks(TestSyncCode.stack_name)
+        # Lambda Api call here, which tests both the python function and the layer
+        lambda_functions = self.stack_resources.get(AWS_LAMBDA_FUNCTION)
+        time.sleep(LAMBDA_SLEEP)
```
mndeveci, line 143: Instead of sleeping and trying once, I would recommend try couple of times and then fail.

samcli/commands/build/build_context.py
qingchm: The integration tests are dependent on this PR's integration test changes but sure I can have two PR's one with just he integration test changes and the other integration test + ux, after integration test merge I can do a merge and have the ux changes merged as a separate PR

tests/integration/sync/test_sync_code.py
qingchm: Makes sense!

tests/integration/sync/sync_integ_base.py
qingchm: Let's remove the whole .aws-sam folder instead, it should be nothing in there that's not removable right?

tests/integration/sync/sync_integ_base.py
qingchm: I couldn't recall the reason why I added this but it was added intentionally. I will test several runs without this.

tests/integration/sync/test_sync_code.py
qingchm: I added a retry mechanism, but some sleep is still needed here, otherwise the lambda response will be the old one, we won't get errors there but the assertion would fail. Or it is a better idea to add an retry mechanism for the content assertion as well?

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,141 @@
...
+            # because of the termination protection, do not delete aws-sam-cli-managed-default stack
+            stack_name = stack["name"]
+            if stack_name != SAM_CLI_STACK_NAME:
+                region = stack.get("region")
+                cfn_client = (
+                    self.cfn_client if not region else boto3.client("cloudformation", config=Config(region_name=region))
+                )
+                ecr_client = self.ecr_client if not region else boto3.client("ecr", config=Config(region_name=region))
+                self._delete_companion_stack(cfn_client, ecr_client, self._stack_name_to_companion_stack(stack_name))
+                cfn_client.delete_stack(StackName=stack_name)
```
mndeveci, line 56: This part is getting duplicated in multiple test cases, can we move it to common place and use it from other places?

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,141 @@
...
         super().tearDown()
 
+    def _get_stacks(self, stack_name):
```
mndeveci, line 59: Seems like function name doesn't match what we are returning. It says `_get_stacks` but it returns `physical_ids`. Can we name the function better?

tests/integration/sync/test_sync_code.py
```diff
@@ -0,0 +1,206 @@
...
+
+# Deploy tests require credentials and CI/CD will only add credentials to the env if the PR is from the same repo.
+# This is to restrict package tests to run outside of CI/CD, when the branch is not master or tests are not run by Canary
+SKIP_SYNC_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
+IS_WINDOWS = platform.system().lower() == "windows"
+# Some wait time for code updates to be reflected on each service
+LAMBDA_SLEEP = 3
+API_SLEEP = 5
+SFN_SLEEP = 5
+CFN_PYTHON_VERSION_SUFFIX = os.environ.get("PYTHON_VERSION", "0.0.0").replace(".", "-")
```
mndeveci, line 34: Can't we use some of these static vars from base class which is also defined there?

tests/integration/sync/sync_integ_base.py
```diff
@@ -1,19 +1,141 @@
...
+                    if nested_resource_type not in physical_ids:
+                        physical_ids[nested_resource_type] = []
+                    physical_ids[nested_resource_type] += nested_physical_ids
+                continue
+            if resource_type not in physical_ids:
+                physical_ids[resource.get("ResourceType")] = []
+            physical_ids[resource_type].append(resource.get("PhysicalResourceId"))
+        return physical_ids
+
+    def _get_lambda_response(self, lambda_function):
```
mndeveci, line 78: Good work on moving these into common places 👏🏼

tests/integration/sync/test_sync_code.py
```diff
@@ -0,0 +1,206 @@
...
+            parameter_overrides="Parameter=Clarity",
+            image_repository=self.ecr_repo_name,
+            s3_prefix=self.s3_prefix,
+            kms_key_id=self.kms_key,
+            tags="integ=true clarity=yes foo_bar=baz",
+        )
+        sync_process_execute = run_command_with_input(sync_command_list, "y\n".encode())
+        self.assertEqual(sync_process_execute.process.returncode, 0)
+
+        time.sleep(API_SLEEP)
```
mndeveci, line 171: I saw that we removed this from function tests, which is nice. But for APIGW and SFN, if there is a possible delay after updating the definition, we can either;
- Check if we can do something in the sync flow to confirm that the changes have been applied and active
- Or add retry here for not having a flaky test case

mndeveci (approved):
Thanks for addressing earlier concerns. Left couple of comments related to some refactoring, if we can address them here or in a follow-up that would be great!

CoshUS (approved):
LGTM! Thanks for addressing the comments!

PR #3582, opened 2022-01-13
fix: force download dependencies if dependency folder doesn't exist

mndeveci:
When running incremental build, `download_dependencies` flag should be `False` if dependecy folder (which should be created earlier) does not exist.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


CoshUS (approved):
LGTM. Is the case of missing dependency dir limited to user manual actions or does it happen in SAM CLI's workflow?

mndeveci:
> LGTM. Is the case of missing dependency dir limited to user manual actions or does it happen in SAM CLI's workflow?

It happened to me when I was testing locally. Anything that causes to first build to fail (internet connection issues or killing process with `CTRL+C`) ends-up all subsequent ones to fail since it assumes dependencies are in that folder but actually that folder doesn't exist.

mingkun2020 (approved):

PR #3601, opened 2022-01-25
feat: sam init changes to improve getting manifest time and clone messaging

jonife:
This pr addresses issues raised by customers with respect to new revamped SAM CLI Init interactive experience:

1. In this PR the wait time to get the manifest file from the cli template was reduced by 80%. Additional robust fault tolerance has been added to ensure that customers always get a vended app template.
2. CLl depends on the manifest file in the cli template repo to create the new, easy and quick interactive experience. This update in the PR ensures that the cli 
      - Attempts to first get just the Manifest file directly from github
      - If this fails, the CLI then resorts to cloning the cli template repo and using the cloned manifest
      - if the cloning fails, a local manifest which acts as reference to all local app templates in SAM CLI is used.
This way ensures that customers always have access to a vended serverless template.
3. Finally, the new SAM CLI has the ability to filter through options therefore reducing the number of steps to a vended serverless application template  via the interactive flow by at-most 70%. This PR adds a text/helper to help make customers aware of those preselected options. 
4. In this pr, the clone message displayed while cloning from github was improved to be more clear.

### DIFF
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -8,6 +8,7 @@ import logging
 import os
 from pathlib import Path
 from typing import Dict, Optional
+import requests
 
 from samcli.cli.global_config import GlobalConfig
 from samcli.commands.exceptions import UserException, AppTemplateUpdateException
@@ -32,6 +33,7 @@ class InitTemplates:
     def __init__(self):
         self._git_repo: GitRepo = GitRepo(url=APP_TEMPLATES_REPO_URL)
         self.manifest_file_name = "manifest.json"
+        self.manifest_url = "https://raw.githubusercontent.com/aws/aws-sam-cli-app-templates/master/manifest.json"
 
     def location_from_app_template(self, package_type, runtime, base_image, dependency_manager, app_template):
         options = self.init_options(package_type, runtime, base_image, dependency_manager)
@@ -171,11 +173,7 @@ class InitTemplates:
         [dict]
             This is preprocessed manifest with the use_case as key
         """
-        self.clone_templates_repo()
-        manifest_path = self.get_manifest_path()
-        with open(str(manifest_path)) as fp:
-            body = fp.read()
-            manifest_body = json.loads(body)
+        manifest_body = self._get_manifest()
 
         # This would ensure the Use-Case Hello World Example appears
         # at the top of list template example displayed to the Customer.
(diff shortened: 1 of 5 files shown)


samcli/commands/init/init_templates.py
```diff
@@ -32,6 +33,7 @@ class InitTemplates:
     def __init__(self):
         self._git_repo: GitRepo = GitRepo(url=APP_TEMPLATES_REPO_URL)
         self.manifest_file_name = "manifest.json"
+        self.manifest_url = "https://raw.githubusercontent.com/aws/aws-sam-cli-app-templates/master/manifest.json"
```
mndeveci, line 36: Let's put it in a static variable just like `APP_TEMPLATES_REPO_URL`.

samcli/commands/init/init_templates.py
```diff
@@ -204,8 +202,23 @@ def get_preprocessed_manifest(
...
+        In an attempt to reduce initial wait time to achieve an interactive
+        flow <= 10sec, This method first attempts to spools just the manifest file and
+        if the manifest can't be spooled, it attempts to clone the cli template git repo or
+        use local cli template
+        """
+        try:
+            response = requests.get(self.manifest_url, timeout=10)
+            body = response.text
+        except (requests.Timeout, requests.ConnectionError):
+            self.clone_templates_repo()
```
mndeveci, line 216: adding a log line could be helpful here to state that getting manifest file directly have failed and we are falling back to cloning the repository.

samcli/commands/init/init_templates.py
```diff
@@ -204,8 +202,25 @@ def get_preprocessed_manifest(
...
+        flow <= 10sec, This method first attempts to spools just the manifest file and
+        if the manifest can't be spooled, it attempts to clone the cli template git repo or
+        use local cli template
+        """
+        try:
+            response = requests.get(MANIFEST_URL, timeout=10)
+            body = response.text
+        except (requests.Timeout, requests.ConnectionError):
+            LOG.debug("Request to get Manifest failed, attempting to clone the repository")
+            LOG.debug("Clone error, attempting to use an old clone from a previous run")
```
mndeveci, line 217: I think one of these log messages are redundant?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -229,7 +230,7 @@ def _generate_default_hello_world_application(
     is_package_type_image = bool(package_type == IMAGE)
     if use_case == "Hello World Example" and not (runtime or base_image or is_package_type_image or dependency_manager):
         if click.confirm("\n Use the most popular runtime and package type? (Nodejs and zip)"):
```
mndeveci, line 232: We forgot to update messaging here, should we use variables here instead of writing them static?

mndeveci (changes requested):

mndeveci (approved):

moelasmar (approved):

PR #3602, opened 2022-01-25
feat: java build changes

mndeveci:
Please don't merge this before merging this PR (https://github.com/aws/aws-lambda-builders/pull/318) and releasing new version of `aws-lambda-builders`.

Related to https://github.com/aws/aws-lambda-builders/pull/318

For using AWS Lambda Layers for Java runtimes and better dependency handling.

- It builds Layers before Functions so that any Layer code can be used during Function build time
- It passes down 2 new parameters (Boolean flag to determine if it is building a Layer and list of experimental flags) to aws-lambda-builders so that changes in aws-lambda-builders can work

Any customer who is using `provided` scope for their `compile` or `runtime` dependency might be affected. Because of that we are making these changes behind a feature flag before making this behavior as default.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -508,3 +509,24 @@ def _is_layer_buildable(layer: LayerVersion):
...
             return False
         return True
+
+    _JAVA_BUILD_WARNING_MESSAGE = (
+        "Test the latest build changes for Java runtime 'SAM_CLI_BETA_MAVEN_SCOPE_AND_LAYER=1 sam build'. "
+        "These changes will replace the existing flow on 1st of April 2022. "
+        "Check __issue_line__ for more information."
+    )
+
+    def _check_java_warning(self):
```
mildaniel, line 519: nit: I know we don't add type hints and docstrings to older files like this, but should we start for consistency going forward?

mildaniel (approved):

samcli/commands/build/build_context.py
mndeveci: Good point, I will update the PR

samcli/commands/build/build_context.py
```diff
@@ -508,3 +509,29 @@ def _is_layer_buildable(layer: LayerVersion):
             LOG.debug("Skip building pre-built layer: %s", layer.full_path)
             return False
         return True
+
+    _JAVA_BUILD_WARNING_MESSAGE = (
+        "Test the latest build changes for Java runtime 'SAM_CLI_BETA_MAVEN_SCOPE_AND_LAYER=1 sam build'. "
+        "These changes will replace the existing flow on 1st of April 2022. "
+        "Check __issue_line__ for more information."
```
CoshUS, line 516: What's the `__issue_line__` referring to?

CoshUS (commented):
Thanks for the PR. Added one question.

samcli/commands/build/build_context.py
mndeveci: We will update this warning message once we have an announcement GH issue ready in SAM CLI. This is just a placeholder for that

CoshUS (approved):

PR #3603, opened 2022-01-26
fix: Add Validation for Incompatible watch and code Options

CoshUS:
N/A

For sam sync, both --watch and --code can be specified at the same time, but this is not a valid workflow.
In such case --code is ignored and a normal --watch is started.

Refactored click_mutex.py to be reusable outside of sam init.

Format of the sam init errors are changed but the content is the same.

#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/click_mutex.py
+++ b/samcli/commands/_utils/click_mutex.py
@@ -80,7 +80,7 @@ class ClickMutex(click.Option):
 
                 msg += self.required_params_hint
                 raise click.UsageError(msg)
-            self.prompt = None
+            self.prompt = ""
 
         return super().handle_parse_result(ctx, opts, args)
 
--- a/samcli/commands/_utils/click_mutex.py
+++ b/samcli/commands/_utils/click_mutex.py
@@ -28,9 +28,9 @@ class ClickMutex(click.Option):
     """
 
     def __init__(self, *args, **kwargs):
-        self.required_param_lists: List = kwargs.pop("required_param_lists", [])
+        self.required_param_lists: List[List[str]] = kwargs.pop("required_param_lists", [])
         self.required_params_hint: str = kwargs.pop("required_params_hint", "")
-        self.incompatible_params: List = kwargs.pop("incompatible_params", [])
+        self.incompatible_params: List[str] = kwargs.pop("incompatible_params", [])
         self.incompatible_params_hint: str = kwargs.pop("incompatible_params_hint", "")
 
         super().__init__(*args, **kwargs)
--- a/samcli/commands/_utils/click_mutex.py
+++ b/samcli/commands/_utils/click_mutex.py
@@ -1,7 +1,7 @@
 """
 Module to check mutually exclusive cli parameters
 """
-from typing import List
+from typing import List, Dict
 
 import click
 
@@ -19,7 +19,7 @@ class Mutex(click.Option):
 
         super().__init__(*args, **kwargs)
 
-    def handle_parse_result(self, ctx, opts, args):
+    def handle_parse_result(self, ctx, opts: Dict, args):
         if self.name not in opts:
             return super().handle_parse_result(ctx, opts, args)
 
--- a/samcli/commands/_utils/click_mutex.py
+++ b/samcli/commands/_utils/click_mutex.py
@@ -1,7 +1,7 @@
 """
 Module to check mutually exclusive cli parameters
 """
-from typing import List, Dict
+from typing import Any, List, Dict, Tuple
 
 import click
 
@@ -19,7 +19,7 @@ class Mutex(click.Option):
 
         super().__init__(*args, **kwargs)
 
-    def handle_parse_result(self, ctx, opts: Dict, args):
+    def handle_parse_result(self, ctx: click.Context, opts: Dict[str, Any], args: List[str]) -> Tuple[Any, List[str]]:
         if self.name not in opts:
             return super().handle_parse_result(ctx, opts, args)
 
--- a/samcli/commands/_utils/click_mutex.py
+++ b/samcli/commands/_utils/click_mutex.py
@@ -6,16 +6,32 @@ from typing import Any, List, Dict, Tuple
 import click
 
 
-class Mutex(click.Option):
+class ClickMutex(click.Option):
     """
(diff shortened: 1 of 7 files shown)


qingchm:
Can we update the title and description to include the mutex option changes? Reading from the changes the watch and code mutex doesn't seem to be the main component of this PR ;) Also please check this test failure test_init_cli_must_pass_with_architecture_and_base_image

CoshUS:
> component

Still WIP. Changing to draft.

samcli/commands/_utils/click_mutex.py
```diff
@@ -0,0 +1,62 @@
...
+import click
+
+
+class Mutex(click.Option):
+    """
+    Preprocessing checks for mutually exclusive or required parameters as supported by click api.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.required_param_lists: List = kwargs.pop("required_param_lists", None)
```
mndeveci, line 15: For a list type, should default value be `None` or an empty list `[]`? We don't need to check if it is `None` when we are looping.

samcli/commands/_utils/click_mutex.py
```diff
@@ -0,0 +1,62 @@
...
+                if self.incompatible_params_hint:
+                    msg += self.incompatible_params_hint
+                raise click.UsageError(msg)
+
+        # Check for required parameters
+        if self.required_param_lists:
+            for required_params in self.required_param_lists:
+                has_all_required_params = False not in [required_param in opts for required_param in required_params]
+                if has_all_required_params:
+                    break
```
mndeveci, line 42: I couldn't understand the `break` statement here. We are looping for each `required_params` and if we find that one we are breaking the loop, wouldn't this miss the next required param?

samcli/commands/_utils/click_mutex.py
```diff
@@ -0,0 +1,62 @@
+"""
+Module to check mutually exclusive cli parameters
+"""
+from typing import Any, List, Dict, Tuple
+
+import click
+
+
+class Mutex(click.Option):
```
mndeveci, line 9: Can we name the class to a more specific name?

mndeveci (commented):
Thanks for working on this. Raised couple of minor concerns, can we also check what is failing in CI jobs?

samcli/commands/_utils/click_mutex.py
CoshUS: self.required_param_lists is a list of combinations.
```
    required_param_lists: List[List[str]]
        List of lists with each supported combination of params
        Ex:
        With option = "a" and required_param_lists = [["b", "c"], ["c", "d"]]
        It is valid to specify --a --b --c or --a --c --d
        but not --a --b --d
```

If one combination is satisfied, then we will `break` and skip over `else` block.

Added documentation for the class and this block.

samcli/commands/_utils/click_mutex.py
CoshUS: Updated to "" and []. However we still need to check `if self.required_param_lists:` to skip over the `for` loop `else` clause.

samcli/commands/_utils/click_mutex.py
CoshUS: Updated name to ClickMutex which also matches the filename.

samcli/commands/_utils/click_mutex.py
mndeveci: I see, so if either `["a", "b", "c"]` or `["a", "b", "d"]` is required, having `["a", "b", "c", "d"]` is still fine because it confirms one or the other requirement. This is not checking it should only be the first case or the second one, right?

samcli/commands/_utils/click_mutex.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, *args, **kwargs):
+        self.required_param_lists: List = kwargs.pop("required_param_lists", [])
+        self.required_params_hint: str = kwargs.pop("required_params_hint", "")
+        self.incompatible_params: List = kwargs.pop("incompatible_params", [])
+        self.incompatible_params_hint: str = kwargs.pop("incompatible_params_hint", "")
+
+        super().__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx: click.Context, opts: Dict[str, Any], args: List[str]) -> Tuple[Any, List[str]]:
+        if self.name not in opts:
```
mndeveci, line 39: Regarding to our discussion below, I feel like having a small doc about this method and its parameters could be useful for the future maintenance.

samcli/commands/_utils/click_mutex.py
CoshUS: Yep, that's correct. This is following the previous behavior of `samcli\commands\local\cli_common\click_mutex.py` so that we don't have regressions.

samcli/commands/_utils/click_mutex.py
CoshUS: Added method docs.

mndeveci (approved):

samcli/commands/_utils/click_mutex.py
```diff
@@ -0,0 +1,89 @@
...
+
+    incompatible_params: List[str]
+        List of incompatible parameters
+
+    incompatible_params_hint: str
+        String to be appended after default incompatible params prompt
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.required_param_lists: List = kwargs.pop("required_param_lists", [])
```
hawflau, line 31: nit: `self.required_param_lists` and `self.incompatible_params` seem to follow different naming patterns.

hawflau, line 31: nit: is it possible to set the type hint to `List[List[str]]`?

hawflau, line 31: nit: is it possible to set the type hint to `List[List[str]]`?

hawflau (approved):
LGTM, just a small question about using typing

samcli/commands/_utils/click_mutex.py
CoshUS: Good catch! Updated.

PR #3613, opened 2022-02-01
feat: Consume esbuild function metadata and pass to lambda builders

mildaniel:
- Check for build method and build options in the template function metadata. If the build method is esbuild, check for additional build properties and pass them to lambda builders.
- Convert the properties dict to snake case to be consistent with sam-cli and lambda builders
- If entry points aren't defined, default the entry point to the lambda handler.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -270,3 +273,18 @@ def get_resource_id(resource_properties, logical_id):
             cdk_resource_id = cdk_resource_id[: -len(CDK_NESTED_STACK_RESOURCE_ID_SUFFIX)]
 
         return cdk_resource_id
+
+    @staticmethod
+    def normalize_build_properties(build_props) -> Dict:
```
hawflau, line 278: question - is Lambda Builder expecting snake case for build properties in general and not only esbuild?

hawflau (approved):
LGTM, approved with one minor question

samcli/lib/samlib/resource_metadata_normalizer.py
mildaniel: BuildProperties is a new field, but everything else we send lambda builders is snake case so I wanted to keep it consistent.

moelasmar (approved):

PR #3617, opened 2022-02-03
feat: Watch Integration Tests

CoshUS:

N/A
Adding integration tests for sam sync --watch.
Updated test data content to 1 and 2 for readability.
Fixed an issue with base dir and definition code triggers where file changes are not triggering.

N/A

#### Checklist

- [x] Write unit tests
- [ ] `make pr` passes
- [x] Write documentation


tests/integration/testdata/sync/infra/template-python-after.yaml
```diff
@@ -11,7 +11,7 @@ Resources:
     Properties:
       CodeUri: after/Python/function/
       Handler: app.lambda_handler
-      Runtime: python3.7
+      Runtime: python3.8
```
qingchm, line 14: The 3.8 changes were reverted in this commit https://github.com/aws/aws-sam-cli/commit/6231a837552ef99fe02d486a38c00be831aad9f1 should check with Mohamed on the reason for sticking with python 3.7

CoshUS: Reverted back to 3.7. Adding a backlog item to update Windows tests to 3.8.

tests/integration/sync/test_sync_watch.py
```diff
@@ -0,0 +1,251 @@
...
+        self.lambda_client = boto3.client("lambda")
+        self.api_client = boto3.client("apigateway")
+        self.sfn_client = boto3.client("stepfunctions")
+        self.stacks = []
+        self.s3_prefix = uuid.uuid4().hex
+        self.test_dir = Path(tempfile.mkdtemp())
+        # Remove temp dir so that shutil.copytree will not throw an error
+        # Needed for python 3.6 and 3.7 as these versions don't have dirs_exist_ok
+        shutil.rmtree(self.test_dir)
+        shutil.copytree(self.test_data_path, self.test_dir)
```
qingchm, line 73: Since we are only using one version in the temp dir, is copying the whole test_data_path content necessary? I think update_file kind of makes the copying action redundant, I would only consider copying one set

qingchm, line 73: Since we are only using one version in the temp dir, is copying the whole test_data_path content necessary? I think update_file kind of makes the copying action redundant, I would only consider copying one set

qingchm: Oh nevermind I think having both copied is useful for infra watch where the template is replaced. Having both would make the template switch smoother. You can ignore this comment

qingchm: Oh nevermind I think having both copied is useful for infra watch where the template is replaced. Having both would make the template switch smoother. You can ignore this comment

qingchm (approved):
Looks good to me!

mingkun2020 (approved):
LGTM!

PR #3618, opened 2022-02-03
feat: sam traces integration test

mingkun2020:
Add sam traces integration test


tests/integration/testdata/traces/python-apigw-sfn/template.yaml
```diff
@@ -0,0 +1,118 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
mndeveci, line 1: I wonder if we can use same projects for both `sam logs` and `sam traces` integration tests

tests/integration/traces/test_traces_command.py
```diff
@@ -0,0 +1,223 @@
...
+
+        cls.stack_resources = cls.cfn_client.describe_stack_resources(
+            StackName=cls.stack_name
+        ).get(
+            "StackResources", []
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.cfn_client.delete_stack(StackName=cls.stack_name)
```
mndeveci, line 51: Will this leave any resource leaking in the account? I remember deploy and sync tests have more complicated cleanup steps, wonder if this would be different or not.

tests/integration/traces/test_traces_command.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        cls.lambda_client = boto3.client("lambda")
+        cls.cfn_client = boto3.client("cloudformation")
+        cls.xray_client = boto3.client("xray")
+
+        cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "traces")
+
+        package = PackageIntegBase()
```
mndeveci, line 30: This might lead to unexpected behavior since you are initiating a test class here. I would recommend changing that method into a static one (since it doesn't use anything from instance) and use the method directly.

tests/integration/traces/test_traces_command.py
```diff
@@ -0,0 +1,223 @@
...
+        cls.stack_name = package._method_to_stack_name("test.integration.test_traces_command")
+
+        LOG.info("Deploying stack %s", cls.stack_name)
+        deploy = DeployIntegBase()
```
mndeveci, line 34: Same goes here, we should not be instantiating test classes inside another one. If you need to use instance methods, please check https://github.com/aws/aws-sam-cli/pull/3607 or https://github.com/aws/aws-sam-cli/pull/3580

tests/integration/traces/test_traces_command.py
mingkun2020: Sync and deploy create stacks in each test case, but in trace we only create stack once in the `setUpClass` then use it for all the rest test cases. So we just need to delete 1 stack in the 'tearDownClass' after all tests are done.

tests/integration/traces/test_traces_command.py
mingkun2020: Create a simply get deploy command list method in traces integration base class to avoid this issue.

tests/integration/traces/test_traces_command.py
mingkun2020: Create a simply get deploy command list method in traces integration base class to avoid this issue.

tests/integration/traces/traces_integ_base.py
```diff
@@ -53,11 +55,38 @@ def get_traces_command_list(
...
\ No newline at end of file
+        return command_list
+
+    @staticmethod
+    def get_basic_deploy_command_list(
+        stack_name=None,
+        template_file=None,
+        capabilities=None,
+        resolve_s3=False,
+    ):
```
mndeveci, line 73: I am in between using this one vs using the one in deploy tests. Even if we are using this one, I would recommend setting default values for some of these params. I believe `capabilities` and `resolve_s3` is always same, right?

tests/integration/traces/traces_integ_base.py
```diff
@@ -53,11 +55,38 @@ def get_traces_command_list(
...
+        if resolve_s3:
+            command_list = command_list + ["--resolve-s3"]
+
+        return command_list
+
+    @staticmethod
+    def _method_to_stack_name(method_name):
+        """Method expects method name which can be a full path. Eg: test.integration.test_deploy_command.method_name"""
+        method_name = method_name.split(".")[-1]
+        return f"{method_name.replace('_', '-')}-{CFN_PYTHON_VERSION_SUFFIX}"
```
mndeveci, line 92: I won't recommend this. If we need to change (or randomize) stack names, we need to change it in multiple places. I would recommend finding more generic approach so that all of these tests (deploy, sync, traces) can use the same method to get the stack name.

mndeveci (changes requested):

tests/integration/traces/traces_integ_base.py
mingkun2020: Agree, let me figure a way to move this method to a utility class.

tests/integration/traces/traces_integ_base.py
mingkun2020: The one in deploy tests is an instance method and if we want to use it I need to inherit deploy base class. Even that I can't use the instance method in the class method `setUpClass()`. I am thinking that should me move all this kind of get_command_list method into a utility class since all the other test class may use this kind of get_command_list method.

tests/integration/traces/traces_integ_base.py
mingkun2020: updated in the latest commit.

tests/integration/traces/traces_integ_base.py
mingkun2020: updated in the latest commit.

tests/integration/traces/test_traces_command.py
mingkun2020: updated in the latest commit.

tests/integration/traces/test_traces_command.py
mingkun2020: updated in the latest commit.

tests/integration/traces/test_traces_command.py
```diff
@@ -0,0 +1,235 @@
...
+        deploy_cmd = DeployIntegBase.get_deploy_command_list(
+            stack_name=cls.stack_name,
+            template_file=cls.test_data_path.joinpath("python-apigw-sfn", "template.yaml"),
+            resolve_s3=True,
+            capabilities="CAPABILITY_IAM",
+        )
+        deploy_result = run_command(deploy_cmd)
+
+        if deploy_result.process.returncode != 0:
+            LOG.error(f"Deployment of the test stack is failed with {deploy_result.stderr}")
```
mndeveci, line 46: Shouldn't we fail the tests here since deployment is failed?

tests/integration/traces/test_traces_command.py
```diff
@@ -0,0 +1,235 @@
...
+class TestTracesCommand(TracesIntegBase):
+    @classmethod
+    def setUpClass(cls):
+        cls.lambda_client = boto3.client("lambda")
+        cls.cfn_client = boto3.client("cloudformation")
+        cls.xray_client = boto3.client("xray")
+
+        cls.test_data_path = Path(__file__).resolve().parents[1].joinpath("testdata", "traces")
+
+        cls.stack_name = method_to_stack_name("test.integration.test_traces_command")
```
mndeveci, line 33: What would be the `stack_name` here? It should start with `test` so that it can be cleaned up afterwards. Can you confirm if the name would be correct. We can also think about changing `method_to_stack_name` to add `test` prefix if it doesn't exist.

tests/integration/traces/test_traces_command.py
mingkun2020: use assertEqual so that it can fail the test if deploy failed.

tests/integration/traces/test_traces_command.py
mingkun2020: Stack name start with test-traces-command-xxxxxx, it should be ok. I also check the log integration test and found it is not start with `test`. We may need to update the log one.

tests/integration/traces/test_traces_command.py
```diff
@@ -22,36 +31,37 @@
...
+        TestTracesCommand.stack_resources = cfn_client.describe_stack_resources(StackName=stack_name).get(
             "StackResources", []
         )
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.cfn_client.delete_stack(StackName=cls.stack_name)
+        yield
+
+        cfn_client.delete_stack(StackName=stack_name)
```
mndeveci, line 64: If any tests fail (including setup) will this line be executed?

mingkun2020: If anything steps before `yield` throw exception before, it won't execute the steps after `yield`. I create one more fixture for deploying stack only and it will be safer to guarantee the teardown.
reference: https://docs.pytest.org/en/6.2.x/fixture.html#safe-teardowns

qingchm (approved):
Did a in person peer review session. LGTM

mndeveci (approved):

PR #3620, opened 2022-02-04
chore: Sync UX updates to remove irrelevant messages

qingchm:
N.A.

Right now after a sync infra call, with the build success comes a series of next actions help text and etc.,  which doesn't apply to the sync context.

Make a new parameter for sync to not print.

Not known of.

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -155,7 +155,7 @@ def __exit__(self, *args):
     def get_resources_to_build(self):
         return self.resources_to_build
 
-    def run(self):
+    def run(self, sync=False):
```
mndeveci, line 158: it is obvious but should we add typing here?

mndeveci (commented):
Is this one contains changes from other PR? If so let's exclude them from here because it makes it harder to review.

CoshUS:
> Is this one contains changes from other PR? If so let's exclude them from here because it makes it harder to review.

We can create a new branch and cherry pick commit 6593e715175131c4b46a5203c046352be3e1be5f into that new branch.

samcli/commands/build/build_context.py
CoshUS: Can we make the name more generic such as `show_message` instead of calling it `sync`?
In general, let's try to isolate different modules so that when going through BuildContext in the future, we don't need to see what sync means.

Another question, what's the reason for putting this parameter in run instead of `__init__`?

CoshUS (changes requested):
As the previous comment mentioned, let's use cherry pick to only include the commit for sync ux change.

qingchm:
> > Is this one contains changes from other PR? If so let's exclude them from here because it makes it harder to review.
> 
> We can create a new branch and cherry pick commit [6593e71](https://github.com/aws/aws-sam-cli/commit/6593e715175131c4b46a5203c046352be3e1be5f) into that new branch.

Right now the changes in integration test would be on test_sync_command.py, but in the code pr, it refactors it into three separate classes. Cherry picking doesn't really work here, after we merge in the code sync pr, the changes would be easier to see

qingchm:
> Is this one contains changes from other PR? If so let's exclude them from here because it makes it harder to review.
https://github.com/aws/aws-sam-cli/commit/6593e715175131c4b46a5203c046352be3e1be5f Please review this part specifically for this PR! The type hint has been added in the latest commit

qingchm:
Since the readability is not good, I'll close this one and open a new one after the sync integration test refactoring has been merged

PR #3622, opened 2022-02-04
chore: Sync UX updates to remove irrelevant messages

qingchm:
N.A.

Right now after a sync infra call, with the build success comes a series of next actions help text and etc.,  which doesn't apply to the sync context.

Make a new parameter for sync to not print.

Not known of.

#### Checklist

- [X] Write unit tests
- [X] `make pr` passes
- [ ] Write documentation


mingkun2020 (commented):
Looks good to me. Do we have irrelevant output messages in package/deploy cmd as well?

samcli/commands/build/build_context.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         aws_region: Optional[str] = None,
         create_auto_dependency_layer: bool = False,
         stack_name: Optional[str] = None,
+        print_message: bool = True,
```
mndeveci, line 73: nit: `print_message` seems too generic, maybe we can use `print_success_message` as the variable name?

mndeveci (approved):

mingkun2020 (approved):

PR #3623, opened 2022-02-04
feat: Add Entry Points for All Handlers in a CodeURI

mildaniel:
SAM-CLI calls Lambda Builders only once for each CodeURI. This works for existing runtimes since everything in the CodeURI gets built. For esbuild, independent handlers within the same CodeURI are built separately, meaning that all of the handlers within a CodeURI need to be added as entry points.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -545,6 +545,7 @@ class ApplicationBuilder:
         container_env_vars: Optional[Dict] = None,
         dependencies_dir: Optional[str] = None,
         download_dependencies: bool = True,
+        functions: Optional[List[Function]] = None,
     ) -> str:
         """
         Given the function information, this method will build the Lambda function. Depending on the configuration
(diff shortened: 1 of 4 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -679,8 +685,11 @@ def _get_build_options(
             build_props = metadata.get(BUILD_PROPERTIES, {})
             # Esbuild takes an array of entry points from which to start bundling
             # as a required argument. This corresponds to the lambda function handler.
-            if handler and not build_props.get("EntryPoints"):
-                entry_points = [handler.split(".")[0]]
+            if functions and not build_props.get("EntryPoints"):
```
moelasmar, line 688: are the build properties metadata properties added to each lambda function that will use ESBuild?

PR #3626, opened 2022-02-06
Make validation errors fatal

tadeboro:
Up until now, running `sam validate` on an invalid template did print a warning to the console, but the last message printed by the sam command was still claiming that the template is valid.

This happens because the template translation process that is used as the validation provider does not consider validation errors to be fatal and ignores them.

This commit makes the validation errors fatal by running validation in a separate step and terminating the validation process if the template is not considered valid by the validator.

We run the `sam validate` command in our CI pipelines to make sure `sam build` will be green once we have things merged into the main branch. But because the validation returns 0 even if there are validation errors in the template, our deployments fail.

This PR adds additional validation step that terminates the validation command with an error if there are validation errors.

Some `sam validate` calls that previously returned status code 0 now return non-zero status. But the only affected calls should be the ones that misreported the state before.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


tadeboro:
Do note that my initial plan was to update the [validation function](https://github.com/tadeboro/serverless-application-model/blob/521abe3d11472225fe4eb6f48a1079c8b0ec55cf/samtranslator/parser/parser.py#L56-L74). Still, I decided against it since, as far as I can understand, the "validation is not always fatal" behavior is there on purpose.

moelasmar:
thanks @tadeboro for the contribution.
Can you give some example where `sam validate` just print warning messages and does not consider the template as invalid. I need to make sure if the same  template will fail or succeed when we deploy it to AWS.

tadeboro:
> thanks @tadeboro for the contribution. Can you give some example where `sam validate` just print warning messages and does not consider the template as invalid. I need to make sure if the same template will fail or succeed when we deploy it to AWS.

The template that got me into trouble contained this snippet:

```yaml
Parameters:
  SubnetIds:
    type: List<AWS::EC2::Subnet::Id>
```

What I get from the `sam validate` command is this:

```
$ sam validate --region us-east-1
2022-02-18 09:04:00 Loading policies from IAM...
2022-02-18 09:04:03 Finished loading policies from IAM.
2022-02-18 09:04:03 Template schema validation reported the following errors: [, P, a, r, a, m, e, t, e, r, s, ., S, u, b, n, e, t, I, d, s, ],  , ', T, y, p, e, ',  , i, s,  , a,  , r, e, q, u, i, r, e, d,  , p, r, o, p, e, r, t, y, ,,  , [, P, a, r, a, m, e, t, e, r, s, ., S, u, b, n, e, t, I, d, s, ],  , A, d, d, i, t, i, o, n, a, l,  , p, r, o, p, e, r, t, i, e, s,  , a, r, e,  , n, o, t,  , a, l, l, o, w, e, d,  , (, ', t, y, p, e, ',  , w, a, s,  , u, n, e, x, p, e, c, t, e, d, )
/home/tadej/devrev/vmotion-jira-extractor/template.yaml is a valid SAM Template
```

And the exit code is 0 in this case. (The issue with bad error message formatting is already fixed in development versions, so ignore that.)

I made the validation error fatal in this PR, this caused a lot of test failures. There are two reasons for those failures:

 1. The test suite contains a lot of invalid templates that now fail validation.
 2. The schema for SAM templates is broken in certain places, which causes `sam validate` to fail on valid templates.

This is why my PR is not finished yet (because I need to shave a lot of yaks before tests will pass ;)). With a little bit of luck, I might be able to finish them over the weekend.

hawflau:
Hey @tadeboro, thanks for your contribution. The warning message `Template schema validation reported the following errors: [,...` is only printed as a warning but would not be thrown as an error from [samtranslator](https://github.com/aws/serverless-application-model/blob/66f6d3b73173bd16d65cc6337c63f94edea24f74/samtranslator/parser/parser.py#L67-L74). It was done this way because the template validation is still not fully developed and we are still exploring different ways to do proper template validation. With that, I'm not sure if we should throw error based on the warning messages. 

One thing we can consider is to accept a flag (e.g. `fail-on-warning`) in the `sam validate` command, and only throw an error if this flag is used.

PR #3630, opened 2022-02-08
feat: Added Persistent Process Utils

CoshUS:
N/A

Adding persistent process testing utilities for using in watch, logs, and traces integration tests.
No prod changes.
N/A
N/A
#### Checklist

- [ ] Write unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -5,13 +5,13 @@ import subprocess
 import tempfile
 
 from threading import Thread
-from typing import Callable, List
+from typing import Callable, List, Optional
 from collections import namedtuple
 from subprocess import Popen, PIPE, TimeoutExpired
 from queue import Queue
 
 import shutil
-import psutil
+import psutil  # type: ignore
 
 IS_WINDOWS = platform.system().lower() == "windows"
 RUNNING_ON_CI = os.environ.get("APPVEYOR", False)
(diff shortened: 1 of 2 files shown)


tests/testing_utils.py
```diff
@@ -61,6 +67,100 @@ def run_command_with_inputs(command_list: List[str], inputs: List[str], timeout=
...
+                if callback(output, outputs):
+                    result_queue.put(True)
+                    return
+        except Exception as ex:
+            result_queue.put(ex)
+
+    reading_thread = Thread(target=_read_output, daemon=True)
+    reading_thread.start()
+    reading_thread.join(timeout=timeout)
+    if reading_thread.isAlive():
```
mndeveci, line 154: Should this method kill the process if it reaches after timeout?

CoshUS: I want to leave that decision up for each individual test case instead. There might be valid use cases where we want to read another line even if the current line timed out.
For the watch integration test, I put the code for killing the process in teardown instead.

mndeveci, line 154:
```suggestion
    if reading_thread.is_alive():
```

mndeveci (approved):
One final suggestion to fix the issues with py39, other than that it looks good 👍🏼

mingkun2020 (approved):

PR #3636, opened 2022-02-10
feat: Update build graph to include handler for esbuild comparison

mildaniel:
Update the build definition comparison to handle esbuild needing a separate build for each handler.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -678,6 +678,8 @@ class ApplicationBuilder:
         """
 
         if metadata and dependency_manager and dependency_manager == "npm-esbuild":
+            # Make a copy of the metadata so that we don't force an updated build definition
+            # when we add the entry points
             build_props = deepcopy(metadata.get(BUILD_PROPERTIES, {}))
             # Esbuild takes an array of entry points from which to start bundling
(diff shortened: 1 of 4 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -676,7 +678,9 @@ def _get_build_options(
         """
 
         if metadata and dependency_manager and dependency_manager == "npm-esbuild":
-            build_props = metadata.get(BUILD_PROPERTIES, {})
+            # Make a copy of the metadata so that we don't force an updated build definition
+            # when we add the entry points
+            build_props = deepcopy(metadata.get(BUILD_PROPERTIES, {}))
```
mndeveci, line 683: Looks like `ResourceMetadataNormalizer` is the method which changes values of `build_props`, can we move `deepcopy` into there for not affecting other users of that method?

mildaniel: Updated to remove deepcopy. The ResourceMetadataNormalizer already returns a fresh copy so I just update that dict.

mndeveci (approved):

CoshUS (approved):

PR #3642, opened 2022-02-11
feat: sam init support for custom runtime

jonife:
This pr help us provide support for custom runtime in SAMCLI

Adding this support will enable us to custom runtimes

None

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -16,8 +16,10 @@ from samcli.lib.utils.git_repo import GitRepo, CloneRepoException, CloneRepoUnst
 from samcli.lib.utils.packagetype import IMAGE
 from samcli.local.common.runtime_template import (
     RUNTIME_DEP_TEMPLATE_MAPPING,
+    get_custom_runtime_base_runtime,
     get_local_lambda_images_location,
     get_local_manifest_path,
+    is_custom_runtime,
 )
 
 LOG = logging.getLogger(__name__)
@@ -179,7 +181,11 @@ class InitTemplates:
         # at the top of list template example displayed to the Customer.
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
-            if filter_value and filter_value != template_runtime:
+            if (
+                is_custom_runtime(filter_value)
+                and filter_value != get_custom_runtime_base_runtime(template_runtime)
+                or (filter_value and not is_custom_runtime(filter_value) and filter_value != template_runtime)
+            ):
                 continue
             template_list = manifest_body[template_runtime]
             for template in template_list:
--- a/samcli/local/common/runtime_template.py
+++ b/samcli/local/common/runtime_template.py
@@ -95,6 +95,7 @@ RUNTIME_TO_DEPENDENCY_MANAGERS = {
     "java8": ["maven", "gradle"],
     "java11": ["maven", "gradle"],
     "java8.al2": ["maven", "gradle"],
+    "rust(provided.al2)": ["cargo"],
 }
 
 SUPPORTED_DEP_MANAGERS: Set[str] = {
--- a/samcli/local/common/runtime_template.py
+++ b/samcli/local/common/runtime_template.py
@@ -2,6 +2,7 @@
 All-in-one metadata about runtimes
 """
 
+import re
 import itertools
 import os
 import pathlib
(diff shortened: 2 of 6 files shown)


samcli/local/common/runtime_template.py
```diff
@@ -138,6 +139,7 @@ def get_local_lambda_images_location(mapping, runtime):
     # ruby runtimes in descending order
     "ruby2.7",
     "ruby2.5",
+    "rust(provided.al2)",
```
mndeveci, line 142: This will require changes whenever a new init template is added. For instance, other than Rust, there is also another PR which is targeting to add new templates for GraalVM, which also uses `provided.al2` for the runtime. Instead of checking runtimes full name, we can check if they are ending with  `(provided.al2)` or `(provided)`, so that we would know that is a custom runtime, and we don't need to check the full string.

mndeveci (changes requested):

samcli/local/common/runtime_template.py
jonife: are we certain that all custom runtimes will end with  the provideded suffix?

mndeveci: If you want to use certain language, then you can use that language's runtime like `node10.x` or `java8`. But if you provide a self executable then you don't need anything extra on lambda image, so you can use `provided` or `provided.al2`.

samcli/local/common/runtime_template.py
```diff
@@ -162,3 +149,172 @@ def get_local_lambda_images_location(mapping, runtime):
...
+    runtime_list : list
+        list of runtime
+
+    Returns
+    -------
+    list
+        list of sorted runtime
+    """
+    runtime_list = verify_runtime_list_provided(runtime_list)
+    _sort_runtimes(runtime_list, 0, len(runtime_list) - 1)
```
hawflau, line 196: Any reason to not use the native `sort()` method but implement a sorting algorithm?

samcli/local/common/runtime_template.py
```diff
@@ -123,6 +106,10 @@ def get_local_lambda_images_location(mapping, runtime):
     # nodejs runtimes in descending order
     "nodejs14.x",
     "nodejs12.x",
+    "nodejs10.x",
```
hawflau, line 109: Do we still need `nodejs10.x`? AFAIK, `nodejs10.x` has been deprecated. It's also removed in the app template repo.

hawflau (changes requested):

samcli/local/common/runtime_template.py
jonife: the complexity of how we want runtimes to be displayed sorting. I am not sure the flexibility sort(). implementing mine allowed me more flexibility.

samcli/commands/init/init_templates.py
```diff
@@ -179,7 +181,11 @@ def get_preprocessed_manifest(
         # at the top of list template example displayed to the Customer.
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
-            if filter_value and filter_value != template_runtime:
+            if (
+                is_custom_runtime(filter_value)
+                and filter_value != get_custom_runtime_base_runtime(template_runtime)
+                or (filter_value and not is_custom_runtime(filter_value) and filter_value != template_runtime)
+            ):
```
mingkun2020, line 188: The condition of this if statement seems too complicated and hard to maintain, could we break it down into multiple if statements and document each cases of these conditions?

samcli/local/common/runtime_template.py
```diff
@@ -162,3 +149,172 @@ def get_local_lambda_images_location(mapping, runtime):
...
+    _type_
+        _description_
+    """
+    if not runtime:
+        return False
+    validation_result = get_custom_runtime_base_runtime(runtime)
+    return runtime in CUSTOM_RUNTIME or bool(validation_result in CUSTOM_RUNTIME)
+
+
+def get_custom_runtime_base_runtime(runtime):
```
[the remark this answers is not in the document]
mingkun2020: Agree on @hawflau 's comment, we need some docstring here and an example about how a custom_base_runtime looks like.

samcli/local/common/runtime_template.py
mingkun2020: A custom comparison can be passed to sorted as parameter called `key`, detail can be found here:(https://docs.python.org/3/howto/sorting.html#key-functions). If it work in this case, I suggest we just need to have a customized key function rather than write the sort function by ourselves.

mingkun2020 (changes requested):

PR #3643, opened 2022-02-12
add local_sqs_service

thoward-godaddy:
Enhances local development when working with SQS event triggers.

Adds basic emulation of SQS event triggers.

If SQS Events are defined for a function and the target resource exists within the stacks the queue is added to an elasticmq configuration. A poller grabs records from the queues and pushes them to the appropriately mapped lambda function using COLD lambda runner.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


thoward-godaddy:
Is it possible to re-initialize this job?

https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/42571393/job/64ogvx02ond7vy2q

praneetap:
hi @thoward-godaddy! Thanks for your contribution 🙂 I am the product owner for SAM and SAM CLI. We have been encouraging people to use cloud based development ([SAM Accelerate](https://aws.amazon.com/blogs/compute/accelerating-serverless-development-with-aws-sam-accelerate/?sc_icampaign=launch_sam-accelerate&sc_ichannel=ha&sc_icontent=awssm-9887_launch&sc_iplace=ribbon&trk=ha_awssm-9887_launch)) in place of local, mainly because you get more test coverage (IAM permissions for ex.) and local emulation is hard to keep on par with the actual AWS service. Have you tried Accelerate yet? I understand that some people still prefer local development, and if that is the case, can we get on a call to chat more about it? I would like to understand your use-case before we can proceed on this PR. [Please DM me](https://twitter.com/praneetaprakash) to set up a time!

praneetap:
Hi @thoward-godaddy , closing this PR as it does not align with our roadmap currently (as discussed over the call). Happy to reopen if something changes in the future.

benjamin9fin:
This would be a really useful feature. Developing locally will always have a faster feedback loop than testing against the cloud and enables test driven development when writing lambdas.

Obviously we get better test coverage when testing against the cloud (possibly with SAM Accelerate) and doing this *in addition* to local testing makes sense.

But it does not replace local testing. Being able to have an isolated dev environment that does not touch the cloud is invaluable.

privateGenish:
Is this issue going to ever resurface again? I feels like it's an integral part of local development

ivan-penchev:
So how can we ensure the team, that this is vital part of DX?

thoward-godaddy:
I'm revisiting this 🙃

hxk1633:
Any update on this PR? I would really like to be able to test SQS events locally.

aalobai1:
👋 Hey another passionate user of AWS SAM, this would be great! I've full sent this with my team of 6 engineers, and being able to locally run SQS would def simplify the dev experience for all of us 🙏

thoward-godaddy:
My team is pivoting to using CDK exclusively. Had positive experience ripping SAM out and letting CDK handle builds/layers as necessary. We already use CDK to handle the majority of our infrastructure so removing SAM would reduce the tooling clutter. Been getting by without local event testing for a while so not losing much by the way of dev capabilities IMO. If I were to take another go on local event testing I'd probably build a little framework around moto proxy:
https://docs.getmoto.org/en/latest/

klexas:
This should be a critical component of the SAM CLI. It's not acceptable to simply not have it in your roadmap. I would like top petition this be reopened. It's unreasonable to be told we can test our lambdas using the AWS SAM CLI, spend all the time creating the templates etc, only to find out that those with SQS in this case, can't be tested local. 

The use case should be obvious. Local, faster testing and development.

PR #3657, opened 2022-02-17
feat: Add Go code generation option to Eventbridge Schemas workflow

kolaamoo25:
No issue number attached for this PR

Allows users to create EventBridge Schema applications using the go1.x runtime

This change maps the go1.x runtime to the Go enum value needed to call the Schemas Download CodeBinding API

N/A

#### Checklist

- [] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [] Write unit tests
- [] Write/update functional tests
- [X] `make pr` passes
- [] `make update-reproducible-reqs` if dependencies were changed
- [] Write documentation


tests/integration/init/schemas/test_init_with_schemas_command.py
```diff
@@ -220,6 +220,37 @@ def test_init_interactive_with_event_bridge_app_aws_schemas_python(self):
             self.assertTrue(expected_output_folder.is_dir())
             self.assertTrue(Path(expected_output_folder, "hello_world_function", "schema").is_dir())
 
+    def test_init_interactive_with_event_bridge_app_aws_schemas_go(self):
```
moelasmar, line 223: please add another use case to test the `event bridge schema app` template, and mark it `@skipif(SKIP_SCHEMA_TESTS)`

moelasmar (changes requested):

tests/integration/init/schemas/test_init_with_schemas_command.py
moelasmar: sorry I got confused, so tag this one with `@skipif(SKIP_SCHEMA_TESTS)`

kolaamoo25: This entire test class is already being skipped by this line: https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/init/schemas/test_init_with_schemas_command.py#L17

samcli/local/common/runtime_template.py
```diff
@@ -164,5 +164,5 @@ def get_local_lambda_images_location(mapping, runtime):
     "python3.6": "Python36",
     "python3.8": "Python36",
     "python3.9": "Python36",
-    "dotnet6": "dotnetcore3.1",
```
kolaamoo25, line 167: Dotnet is currently not a supported language for EventBridge Schemas code bindings

jfuss (approved):

jfuss:
@kolaamoo I know we have app templates pr opened as well but I don't understand the order these need to be released (and merged). Do we need this merged and released first before merging app templates?

kolaamoo25:
> @kolaamoo I know we have app templates pr opened as well but I don't understand the order these need to be released (and merged). Do we need this merged and released first before merging app templates?

If my understanding it correct, the way the sam-cli works is that it clones the sam-cli-app-templates repo for every sam init call. So the sam-cli-app-templates deployment time is pretty much instant after it's approved and merged. There's nothing blocking the merge and deployment on the sam-cli side before sam-cli-app-templates. Let me know if I have something wrong.

samcli/local/common/runtime_template.py
hawflau: minor - If it's not supported, shall we remove this line? Even if it will be supported in the future, I bet the language code would be "dotnetcore3.1". It doesn't hurt to still keep it here as it doesn't impact anything, but just create confusion.

kolaamoo25: I think some other tests rely on this value being here. I tried removing the value and some of the checks failed.

torresxb1 (approved):

moelasmar (approved):

torresxb1:
Thanks for your contribution! @kolaamoo25

Issue #3683, opened 2022-02-23
ADOT layer causes TypeError when using --warm-containers feature with start-api


hooverdc:
For reference: https://aws-otel.github.io/docs/getting-started/lambda/lambda-python

### Description:

When using the AWS managed Lambda layer for ADOT and the sam start-api with `--warm-containers`, the Lambda container will fail to initialize.

### Steps to reproduce:

Require layer `arn:aws:lambda:us-west-2:901920570463:layer:aws-otel-python38-amd64-ver-1-9-1:1` in your SAM function definition. Call `sam local start-api --warm-containers EAGER|LAZY`. Unable to replicate without the `--warm-containers` flag.

### Observed result:

```bash
2022-02-22 17:05:10,378 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2022-02-22 17:05:10,378 | Using config file: samconfig.toml, config environment: default
2022-02-22 17:05:10,378 | Expand command line arguments to:
2022-02-22 17:05:10,378 | --template_file=/Users/hooverd/Projects/account-api-rest/.aws-sam/build/template.yaml --env_vars=/dev/fd/63 --warm_containers=LAZY --host=127.0.0.1 --port=3000 --static_dir=public --layer_cache_basedir=/Users/hooverd/.aws-sam/layers-pkg --container_host=localhost --container_host_interface=127.0.0.1 
2022-02-22 17:05:10,489 | local start-api command is called
2022-02-22 17:05:10,513 | Collected default values for parameters: {}
2022-02-22 17:05:10,541 | There is no customer defined id or cdk path defined for resource AccountApi, so we will use the resource logical id as the resource id
2022-02-22 17:05:10,541 | There is no customer defined id or cdk path defined for resource AuthorizerFunctionLambdaPermission, so we will use the resource logical id as the resource id
2022-02-22 17:05:10,541 | Sam customer defined id is more priority than other IDs. Customer defined id for resource AuthorizerFunction is AuthorizerFunction
2022-02-22 17:05:10,541 | There is no customer defined id or cdk path defined for resource AccountApiSecurityGroup, so we will use the resource logical id as the resource id
2022-02-22 17:05:10,541 | Sam customer defined id is more priority than other IDs. Customer defined id for resource SharedLayer is SharedLayer
[...]

PR #3694, opened 2022-02-26
fix: Warm containers break on non-local layers

mildaniel:

#3683

Running start-api with warm containers and a downloaded layer will throw an exception ```"Lambda functions containers initialization failed because of expected str, bytes or os.PathLike object, not NoneType"```

Check if layer codeuri exists before attempting to watch it

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/file_observer.py
+++ b/samcli/lib/utils/file_observer.py
@@ -112,7 +112,8 @@ class LambdaFunctionObserver:
             """
             code_paths = [function_config.code_abs_path]
             if function_config.layers:
-                code_paths += [layer.codeuri for layer in function_config.layers]
+                # Non-local layers will not have a codeuri property and don't need to be observed
+                code_paths += [layer.codeuri for layer in function_config.layers if layer.codeuri]
             return code_paths
 
         def _get_image_lambda_function_image_names(function_config: FunctionConfig) -> List[str]:
--- a/tests/integration/local/start_api/start_api_integ_base.py
+++ b/tests/integration/local/start_api/start_api_integ_base.py
@@ -12,8 +12,8 @@ from pathlib import Path
 import docker
 from docker.errors import APIError
 
-from tests.testing_utils import kill_process, read_until_string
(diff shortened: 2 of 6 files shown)


moelasmar (approved):

hawflau (approved):

samcli/lib/utils/file_observer.py
```diff
@@ -112,7 +112,8 @@ def _get_zip_lambda_function_paths(function_config: FunctionConfig) -> List[str]
             """
             code_paths = [function_config.code_abs_path]
             if function_config.layers:
-                code_paths += [layer.codeuri for layer in function_config.layers]
+                # Non-local layers will not have a codeuri property and don't need to be observed
```
jfuss, line 115: What is a "non-local" layer? If a Layer is in the template but does not have a CodeUri, shouldn't we fail instead of ignore? Not knowing where the layer code is, means we cannot add it to the container making testing not effective/confusing to customers. Usually we want the container to be as close to when the customer deploys as possible.

jfuss (changes requested):
What about integ tests? It isn't clear from the PR or the Issue in what case this actually happens for.

samcli/lib/utils/file_observer.py
mildaniel: By non-local, I mean a layer referencing a layer arn instead of a layer resource defined within the template. With these types of layers, we download them and they are still usable when running local invoke, start-api, etc. It only breaks when we try to watch the code source with --warn-containers. Since the source isn't on the user's machine there's nothing to watch, but the layer is still usable and testable.

jfuss: I see that makes sense. Should this be an integ tests to cover this case? Would have made understanding the change easier as well.

mildaniel: I'll add an integration test.

jfuss (approved):

PR #3702, opened 2022-03-01
[typing] Unlock boto3-stubs and add all services in use

vemel:
#### Notes

Thank you for using `boto3-stubs` library. I have noticed that `boto3-stubs` version was locked and did not have all necessary services installed. So I hope you find this PR useful.

No issue

- `boto3-stubs` version was locked and is no longer in sync with `boto3` version. The original reason for locking has been resolved.
- `boto3-stubs[essential]` does not include all services that are currently used by the project, so type checking was off for some `boto3.client` calls
- Some `mypy` errors were not shown because of the outdated `botocore-stubs` package

- ~~There is a potential `AssertionError` introduced in `samcli/lib/package/packageable_resources.py`~~ FIXED

#### Checklist

- [ ] Write unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/packageable_resources.py
```diff
@@ -121,6 +121,7 @@ def export(self, resource_id: str, resource_dict: Optional[Dict], parent_dir: st
         temp_dir = None
         if is_local_file(property_value) and not is_zip_file(property_value) and self.FORCE_ZIP:
             temp_dir = copy_to_temp_dir(property_value)
+            assert self.PROPERTY_NAME is not None
```
vemel, line 124: Potential `AssertionError`, but actually this function fails with `self.PROPERTY_NAME is None` earlier.

This assert could be moved higher, because `jmespath.search(None)` fails as well. Probably a proper fix would be to change `PROPERTY_NAME: Optional[str] = None` to `PROPERTY_NAME: str = ""`

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -60,11 +60,13 @@ def manage_stack(
...
-            session = boto3.Session(profile_name=profile, region_name=region if region else None)  # type: ignore
+            session = boto3.Session(profile_name=profile, region_name=region if region else None)
             cloudformation_client = session.client("cloudformation")
         else:
+            # botocore-stubs does not allow Optional region_name, will be fixed soon
             cloudformation_client = boto3.client(
-                "cloudformation", config=Config(region_name=region if region else None)
+                "cloudformation",
+                config=Config(region_name=region if region else None),  # type: ignore
             )
```
vemel, line 70: `Config` works correctly with `region_name=None`. I have added a fix to `botocore-stubs`.

samcli/lib/utils/profile.py
```diff
@@ -1,10 +1,10 @@
...
 """
-from typing import List, cast
+from typing import List
 
 from botocore.session import Session
 
 
 def list_available_profiles() -> List[str]:
-    return cast(List[str], Session().available_profiles)
+    return Session().available_profiles
```
vemel, line 10: `Session().available_profiles` is typed correctly as a `List[str]` now

samcli/lib/package/packageable_resources.py
vemel: I made a change that I described above. `assert` was removed, `PROPERTY_NAME` is now always an `str`.

samcli/lib/utils/managed_cloudformation_stack.py
vemel: New version released, `type: ignore` removed.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -122,16 +123,16 @@ def update_companion_stack(self) -> None:
...
-            waiter = self._cfn_client.get_waiter("stack_update_complete")
+            update_waiter = self._cfn_client.get_waiter("stack_update_complete")
+            update_waiter.wait(StackName=stack_name, WaiterConfig=self._update_stack_waiter_config)
         else:
             self._cfn_client.create_stack(
                 StackName=stack_name, TemplateURL=template_url, Capabilities=["CAPABILITY_AUTO_EXPAND"]
             )
-            waiter = self._cfn_client.get_waiter("stack_create_complete")
+            create_waiter = self._cfn_client.get_waiter("stack_create_complete")
+            create_waiter.wait(StackName=stack_name, WaiterConfig=self._update_stack_waiter_config)
```
vemel, line 133: Even though lines 127 and 133 are identical, it is easier for type checking, because `StackUpdateCompleteWaiter` and `StackCreateCompleteWaiter` do not share the same `wait` method.

samcli/lib/package/packageable_resources.py
```diff
@@ -92,7 +92,7 @@ class ResourceZip(Resource):
     """
 
     RESOURCE_TYPE: Optional[str] = None
-    PROPERTY_NAME: Optional[str] = None
+    PROPERTY_NAME: str = ""
```
vemel, line 95: I cross-checked the code, it is a safe change as all subclasses override this attribute.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -280,7 +280,7 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
             aws_creds=aws_creds,
         )  # EnvironmentVariables is not yet annotated with type hints, disable mypy check for now. type: ignore
 
-    def _get_session_creds(self) -> Credentials:
+    def _get_session_creds(self) -> Optional[Credentials]:
```
vemel, line 283: This function can potentially return `None`. Should be safe for the outside code, only typing was off.

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -122,16 +123,16 @@ def update_companion_stack(self) -> None:
...
 
-        waiter.wait(StackName=stack_name, WaiterConfig=self._update_stack_waiter_config)  # type: ignore
-
-    def _delete_companion_stack(self):
+    def _delete_companion_stack(self) -> None:
```
vemel, line 135: `mypy` does not perform type checking in untyped methods.

requirements/dev.txt
```diff
@@ -8,8 +8,7 @@ pytest-cov==2.10.1
 # mypy adds new rules in new minor versions, which could cause our PR check to fail
 # here we fix its version and upgrade it manually in the future
 mypy==0.790
-# 1.17.91 has a bug, https://github.com/vemel/mypy_boto3_builder/pull/82
-boto3-stubs[essential]==1.17.90.post1
+boto3-stubs[apigateway,cloudformation,ecr,iam,lambda,s3,schemas,secretsmanager,signer,stepfunctions,sts,xray]
```
CoshUS, line 11: Similar to other libraries, can we pin the version to the latest (1.21.21) instead?
This will make CI/builds more stable and reproducible. We can manually update the version if needed in the future without worrying about breaking changes.

CoshUS (changes requested):
Thanks for the PR and really appreciate following up on the previous bug!
Just one small feedback. Rest looks good to me.

requirements/dev.txt
vemel: Just did. However, `boto3` version is not frozen in `requirements/base.txt`, so the latest `boto3` version is installed. Ideally `boto3` and `boto3-stubs` versions should be in sync.

Once `boto3` version reaches `1.22.x` it would make sense to bump `boto3-stubs` as well.

CoshUS (approved):
LGTM! Thanks!

mildaniel (approved):

vemel:
Thank you, guys! Can be safely merged, I suppose.

PR #3706, opened 2022-03-02
fix: container start waits for port

torresxb1:
When doing local invoke, SAM CLI used to consider the container "started" even if the host port was not yet open, so calls to invoke the function could fail (and eventually time out or exhaust the retries if the port took long enough to be open). 

This PR makes it so the container is only considered "started" once the port is open.

#### Checklist

- [x] Write unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/local/docker/container.py
```diff
@@ -194,13 +196,13 @@ def create(self):
 
         return self.id
 
-    def stop(self, time=3):
+    def stop(self, timeout=3):
```
torresxb1, line 199: name conflict after importing `time`

samcli/local/docker/container.py
```diff
@@ -272,6 +274,31 @@ def start(self, input_data=None):
...
+        Waits until the host machine port that Docker binds to is open.
+        """
+        start_time = time.time()
+        sleep = 0.1
+        timeout = 10
+        while True:
+            a_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            location = (self._container_host_interface, self.rapid_port_host)
+            is_port_open = not a_socket.connect_ex(location)
+            a_socket.close()
```
torresxb1, line 291: https://www.kite.com/python/answers/how-to-check-if-a-network-port-is-open-in-python

Should I include this as a comment in code?

samcli/local/docker/container.py
```diff
@@ -272,6 +274,31 @@ def start(self, input_data=None):
...
+        # Wait for port to be open
+        self.wait_for_port()
+
+    def wait_for_port(self):
+        """
+        Waits until the host machine port that Docker binds to is open.
+        """
+        start_time = time.time()
+        sleep = 0.1
+        timeout = 10
```
torresxb1, line 286: for my tests it's only been taking a couple of seconds at most. Is this a good number for the timeout? Should it be configurable by the customer? (seems very specific though)

samcli/local/docker/container.py
```diff
@@ -272,6 +274,31 @@ def start(self, input_data=None):
...
+            location = (self._container_host_interface, self.rapid_port_host)
+            is_port_open = not a_socket.connect_ex(location)
+            a_socket.close()
+
+            if is_port_open:
+                break
+
+            current_time = time.time()
+            if current_time - start_time > timeout:
+                raise RuntimeError("Timed out while starting container")
```
torresxb1, line 298: Is this the right exception to be using?

tests/unit/local/docker/test_container.py
```diff
@@ -481,19 +481,40 @@ def setUp(self):
...
+        socket_mock.connect_ex.return_value = 0
+        patched_socket.return_value = socket_mock
+
         self.container.start()
 
         self.mock_docker_client.containers.get.assert_called_with(self.container.id)
         container_mock.start.assert_called_with()
 
+    @patch("socket.socket")
+    def test_times_out_if_port_not_open(self, patched_socket):
```
torresxb1, line 503: the test also has to wait 10 seconds (or whatever the timeout is set to) for it to time out. Is this okay?

samcli/local/docker/container.py
moelasmar: I think it is Ok

samcli/local/docker/container.py
moelasmar: >For some context, before adding this new function to wait for the port, a local invoke would time out after either exhausting the retries (3) or the function timeout (default is 3 seconds), whichever comes first (meaning it wouldn't wait for the port for more than 3 seconds, or the function timeout, at most).
This new one would be another timeout, before those two start.

samcli/local/docker/container.py
moelasmar: May be use a `UserException` will be better, also you let the customer knows that increasing the timeout and retry may fix the issue (in case if you will make the timeout configurable)

samcli/local/docker/container.py
moelasmar, line 286: can you define these values as Constants values per the module. I believe this will help you customizing them during unit testing

tests/unit/local/docker/test_container.py
moelasmar: if you can not override them in unit testing, so it is Ok

samcli/local/docker/container.py
mildaniel: I think a comment here just explaining what's going on and how this is waiting for a non-error socket connection to the container could be helpful.

samcli/local/docker/container.py
torresxb1: > so do you mean SAM should not wait more than 3 seconds ?
I think it should wait at least like 5 seconds anyway, and making it configurable through an environment variable seems like a good idea 👍

mildaniel: Also having `timeout` as a method parameter might make it easier to configure later on and would make testing a  little easier.

samcli/local/docker/container.py
torresxb1: added a small comment

samcli/local/docker/container.py
torresxb1: decided to make it a constant for now, as it can also be patched in the tests. I think if later on we found that we wanted to wait for a port in other parts of the code, then maybe we could put this in some util in a more generic way (accepting timeout, sleep, port, etc. parameters).

samcli/local/docker/container.py
```diff
@@ -272,6 +277,35 @@ def start(self, input_data=None):
...
+
+            if is_port_open:
+                break
+
+            current_time = time.time()
+            if current_time - start_time > START_CONTAINER_TIMEOUT:
+                raise ContainerNotStartableException(
+                    f"Timed out while starting container. You can increase this timeout by "
+                    f"setting the SAM_CLI_START_CONTAINER_TIMEOUT environment variable. "
+                    f"The current value is {START_CONTAINER_TIMEOUT} (seconds)."
```
torresxb1, line 304: Discussed this with @moelasmar and even though this message is not 100% accurate (we don't time the whole container start process, but rather just waiting for the port), we thought maybe it's more valuable to have a clear error message.

moelasmar (approved):

mildaniel (approved):

PR #3724, opened 2022-03-10
Gating Unsupported runtime

jonife:
Help in gating runtimes SAMCLI is yet to supported to be accessed by customer.

Removes runtime from list of runtimes generated from the manifest file in CLI template repo

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -17,7 +17,7 @@ from samcli.commands.init.interactive_event_bridge_flow import (
 )
 from samcli.commands.exceptions import SchemasApiException, InvalidInitOptionException
 from samcli.lib.schemas.schemas_code_manager import do_download_source_code_binding, do_extract_and_merge_schemas_code
-from samcli.local.common.runtime_template import LAMBDA_IMAGES_RUNTIMES_MAP
+from samcli.local.common.runtime_template import INIT_RUNTIMES, LAMBDA_IMAGES_RUNTIMES_MAP
 from samcli.commands.init.init_generator import do_generate
 from samcli.commands.init.init_templates import InitTemplates, InvalidInitTemplateError
 from samcli.lib.utils.osutils import remove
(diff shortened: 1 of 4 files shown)


samcli/commands/init/interactive_init_flow.py
```diff
@@ -349,6 +350,20 @@ def get_sorted_runtimes(runtime_option_list):
     return sorted_runtime
 
 
+def get_supported_runtime(runtime_list):
```
hawflau, line 353: Please add docstring. Also, can we add a unit test for this function?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -349,6 +350,20 @@ def get_sorted_runtimes(runtime_option_list):
...
+    supported_runtime_list = []
+    error_message = ""
+    for runtime in runtime_list:
+        if runtime not in INIT_RUNTIMES:
+            if not error_message:
```
hawflau, line 358: If there are multiple not-supported runtimes, it seems the debug message will still be shown repeatedly?

hawflau (changes requested):

hawflau (approved):
LGTM!

samcli/commands/init/interactive_init_flow.py
```diff
@@ -300,6 +300,9 @@ def _get_choice_from_options(chosen, options, question, msg):
     click_choices = []
 
     options_list = options if isinstance(options, list) else list(options.keys())
+    options_list = (
+        get_sorted_runtimes(options_list) if msg == "Runtime" and not isinstance(options, list) else options_list
```
torresxb1, line 304: why do we need the condition `not isinstance(options, list)`?

samcli/commands/init/interactive_init_flow.py
```diff
@@ -338,16 +339,45 @@ def get_sorted_runtimes(runtime_option_list):
...
+    error_message = ""
+    for runtime in runtime_list:
+        if runtime not in INIT_RUNTIMES:
+            if not error_message:
+                error_message = "Additional runtimes may be available in the latest SAM CLI version. \
+                    Upgrade your SAM CLI to see the full list."
+                LOG.debug(error_message)
+            continue
+        supported_runtime_list.append(runtime)
+    return supported_runtime_list
```
torresxb1, line 378: Shorter way of doing this
```suggestion
    supported_runtime_list = list(set(runtime) & set(INIT_RUNTIMES))
    if len(supported_runtime_list) != len(runtime):
        LOG.debug(
                    "Additional runtimes may be available in the latest SAM CLI version. Upgrade your SAM CLI to see the full list."
                )
    return supported_runtime_list
```

jonife: This doesn't covers an edge where we INIT_runtime doesn't all the runtime, ie custom runtime support.

torresxb1: but if INIT_runtime doesn't contain all the runtimes, it would throw an error anyway because of the check `if runtime not in INIT_RUNTIMES`, no? (I believe the `supported_runtime_list.append(runtime)` wouldn't be run because of the `continue` before it). 
Also, if that's a valid edge case, could we add a test to cover it?

torresxb1: if you need to return all of the runtimes in `runtime_list` then you could just update my suggestion to return `runtime_list` instead of `supported_runtime_list`. But I'm not sure if I understand correctly

PR #3726, opened 2022-03-12
SAM INIT support for custom runtimes

jonife:

Why is this change necessary?
This pr help us provide support for custom runtime in SAMCLI

Adding this support will enable us to custom runtimes

None

#### Mandatory Checklist

- [] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -182,7 +182,7 @@ class InitTemplates:
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
             if not filter_value_matches_template_runtime(filter_value, template_runtime):
-                LOG.debug("Template runtime {template_runtime} does not match filter value {filter_value}")
+                LOG.debug("Template runtime %s does not match filter value %s", template_runtime, filter_value)
                 continue
             template_list = manifest_body[template_runtime]
             for template in template_list:
@@ -283,8 +283,10 @@ def filter_value_matches_template_runtime(filter_value, template_runtime):
     bool
         True if there is a match else False
     """
+    if not filter_value:
+        return True
     if is_custom_runtime(filter_value) and filter_value != get_provided_runtime_from_custom_runtime(template_runtime):
         return False
-    if filter_value and not is_custom_runtime(filter_value) and filter_value != template_runtime:
+    if not is_custom_runtime(filter_value) and filter_value != template_runtime:
         return False
     return True
--- a/samcli/commands/init/init_templates.py
+++ b/samcli/commands/init/init_templates.py
@@ -16,7 +16,7 @@ from samcli.lib.utils.git_repo import GitRepo, CloneRepoException, CloneRepoUnst
 from samcli.lib.utils.packagetype import IMAGE
 from samcli.local.common.runtime_template import (
     RUNTIME_DEP_TEMPLATE_MAPPING,
-    get_custom_runtime_base_runtime,
+    get_provided_runtime_from_custom_runtime,
     get_local_lambda_images_location,
     get_local_manifest_path,
     is_custom_runtime,
@@ -182,6 +182,7 @@ class InitTemplates:
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
             if not filter_value_matches_template_runtime(filter_value, template_runtime):
+                LOG.debug("Template runtime {template_runtime} does not match filter value {filter_value}")
                 continue
             template_list = manifest_body[template_runtime]
             for template in template_list:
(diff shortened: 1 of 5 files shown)


samcli/commands/init/interactive_init_flow.py
```diff
@@ -192,6 +197,7 @@ def _generate_from_use_case(
     [*] Test Function in the Cloud: sam sync --stack-name {{stack-name}} --watch
     """
     click.secho(next_commands_msg, fg="yellow")
+    runtime = get_custom_runtime_base_runtime(runtime) if is_custom_runtime(runtime) else runtime
```
mildaniel, line 200: Can we define this value once above and re-use it avoid duplication of this line and line 173?

mildaniel (commented):
Thanks @jonife! Since this is more of a complete feature, could we also add an integration test(s) here too?

samcli/local/common/runtime_template.py
```diff
@@ -145,3 +149,44 @@ def get_local_lambda_images_location(mapping, runtime):
...
+    runtime : str
+        runtime to be
+    Returns
+    -------
+    _type_
+        _description_
+    """
+    if not runtime:
+        return False
+    validation_result = get_custom_runtime_base_runtime(runtime)
```
mndeveci, line 170:
```suggestion
    provided_runtime = get_custom_runtime_base_runtime(runtime)
```

samcli/local/common/runtime_template.py
```diff
@@ -145,3 +149,44 @@ def get_local_lambda_images_location(mapping, runtime):
...
+    return runtime in PROVIDED_RUNTIMES or bool(validation_result in PROVIDED_RUNTIMES)
+
+
+def get_custom_runtime_base_runtime(runtime):
```
mndeveci, line 174:
```suggestion
def get_provided_runtime_from_custom_runtime(runtime):
```

samcli/commands/init/init_templates.py
```diff
@@ -262,3 +264,26 @@ def template_does_not_meet_filter_criteria(
...
+        Lambda runtime used to filter through data generated from the manifest
+    template_runtime : str
+        Runtime of the template in view
+
+    Returns
+    -------
+    bool
+        True if there is a match else False
+    """
+    if is_custom_runtime(filter_value) and filter_value != get_custom_runtime_base_runtime(template_runtime):
```
mndeveci, line 285: `is_custom_runtime` is already calling `get_custom_runtime_base_runtime` in it, why do we need to call it here again?

samcli/commands/init/init_templates.py
```diff
@@ -179,7 +181,7 @@ def get_preprocessed_manifest(
         # at the top of list template example displayed to the Customer.
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
-            if filter_value and filter_value != template_runtime:
+            if not filter_value_matches_template_runtime(filter_value, template_runtime):
                 continue
```
mndeveci, line 185: Let's add `debug` logging here so that we can investigate any possible issues in the future.

samcli/commands/init/init_templates.py
```diff
@@ -179,7 +181,8 @@ def get_preprocessed_manifest(
         # at the top of list template example displayed to the Customer.
         preprocessed_manifest = {"Hello World Example": {}}  # type: dict
         for template_runtime in manifest_body:
-            if filter_value and filter_value != template_runtime:
+            if not filter_value_matches_template_runtime(filter_value, template_runtime):
+                LOG.debug("Template runtime {template_runtime} does not match filter value {filter_value}")
```
mndeveci, line 185:
```suggestion
                LOG.debug("Template runtime %s does not match filter value %s", template_runtime, filter_value)
```

samcli/commands/init/init_templates.py
```diff
@@ -262,3 +265,26 @@ def template_does_not_meet_filter_criteria(
...
+        Lambda runtime used to filter through data generated from the manifest
+    template_runtime : str
+        Runtime of the template in view
+
+    Returns
+    -------
+    bool
+        True if there is a match else False
+    """
+    if is_custom_runtime(filter_value) and filter_value != get_provided_runtime_from_custom_runtime(template_runtime):
```
mndeveci, line 286: If `filter_value` is `None` there is no need to run rest of this function. So we can simplify this by adding;
```
if not filter_value:
  return True
```
at the beginning of the function.

samcli/commands/init/interactive_init_flow.py
```diff
@@ -163,9 +168,12 @@ def _generate_from_use_case(
     location = templates.location_from_app_template(package_type, runtime, base_image, dependency_manager, app_template)
 
     final_architecture = get_architectures(architecture)
+    lambda_supported_runtime = (
```
mndeveci, line 171: This `lambda_supported_runtime` is not used below, where we call `is_dynamic_schemas_template` or in the `generate_summary_message`. Is that expected?

mndeveci (commented):
Thanks a lot for addressing earlier feedback! Added some small comments there.

mndeveci (approved):

hawflau (approved):

PR #3729, opened 2022-03-14
Add option to 'deploy' and 'sync' commands to write stack outputs to machine readable (Yaml or JSON) file

jeff-wishnie:
When using `sam deploy` in a monorepo or ci pipeline, stack outputs containing resource identifiers may be needed as inputs to further build and deploy steps. Parsing the stdout of the `sam` command is awkward and brittle (the output may change) and running additional `aws cli` describe-stack command is unneccesary as `sam` already has the necessary data.

Adds an optional argument to `deploy` and `sync` commands so save stack outputs into a local Yaml or JSON file.

Requires test suites to be updated to be aware of additional optional argument.

No other side effects:
- The argument is optional so does not change the CLI interface
- When used, the feature simple writes the same information currently provided to STDOUT to a file as well. No behavior is changed.
- Failure to write the file (e.g. attempting to write to a non-existent directory) results in a warning only.

#### Mandatory Checklist

- [x ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/command.py
+++ b/samcli/commands/deploy/command.py
@@ -2,6 +2,7 @@
 CLI command for "deploy" command
 """
 import logging
+import pathlib
 
 import click
 
@@ -103,6 +104,12 @@ LOG = logging.getLogger(__name__)
     is_flag=True,
     help="Preserves the state of previously provisioned resources when an operation fails.",
 )
+@click.option(
+    "--stack-outputs-file",
+    "-so",
+    type=click.Path(),
+    help="Saves stack outputs as JSON with the file path/name provided",
+)
 @stack_name_option(callback=guided_deploy_stack_name)  # pylint: disable=E1120
 @s3_bucket_option(guided=True)  # pylint: disable=E1120
 @image_repository_option
@@ -157,6 +164,7 @@ def cli(
     config_file,
     config_env,
     disable_rollback,
+    stack_outputs_file,
 ):
     """
     `sam deploy` command entry point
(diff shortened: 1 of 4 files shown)


CoshUS:
Hey, @jeff-wishnie, thanks for the PR!
In terms of the design, I think we should make this generic enough so that other commands can also use this new flag.
Instead of writing to a file directly, we can use `sam deploy --format json` to prints all the information about a deployment in a json object. This is more versatile and can be piped into files or other stdins.
This `--format` parameter can then also be expanded into other command such as build and sync.
I'll bring this up with the team and get back to you on a more concrete path forward.

jeff-wishnie:
Sure, I like the approach of outputing a format and letting people redirect to a file if they want. And also generalizing.

But, as FYI, there are a couple reasons I didn't:
- SAM CLI currently write a lot of notice/info messages to STDOUT (and a few to STDERR), so if you want to use `--format json` to write JSON to STDOUT, you need to decide which of the current messages get supressed, or sent to STDERR instead. 

   Going through all the LOG messages from all commands to make sure they don't pollute STDOUT looked like more trouble than it is worth. 
   
- I didn't see any sort of structured code for handling outputs--like a data structure that commands add messages to that could be filtered or transformed into differ formats. So there was no clean way to add an API for commands to, in a general fashion, provide output data that should be structured (into JSON or YAML or whatever)

 For the purpose of extracting structured info for configuration of components that depend on knowing attributes of various AWS resources, you only need StackOutputs--and the DeployContext does have one place where it prints these out, so that was a clean hook for writing a file, and works for any command that uses a DeployContext--currently Deploy and Sync.

PR #3738, opened 2022-03-15
Revert "fix: Warm containers break on non-local layers"

hawflau:
Reverts aws/aws-sam-cli#3694


mildaniel (approved):

torresxb1 (approved):

matteocontrini:
Why was this fix reverted?

PR #3764, opened 2022-03-21
fix: Resolving symlinks when updating relative paths in template

calabrla:
No issue was created.

Running `sam build` against a template that is in a symlinked directory results in incorrect relative paths in the generated template. (See additional context below)

This change resolves symlinks in the `original_root` and `new_root` before updating relative paths in the generated template at the end of a build.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

#### Additional Context
Given the project structure:
```
.
├── .aws-sam
│   └── build
│       ├── LambdaFunction
│       └── template.yaml
├── build -> /some/other/directory/build
│   └── cdk.out
│       └── Application.template.json
└── lib
    └── app.ts
```
When the project is built via `sam build -t build/cdk.out/Application.template.json`, in `app_builder.py` absolute paths are constructed by resolving all symlinks and then made relative as seen [here](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L332). Before the template is written, the relative file paths are updated to be relative to a `new_root`, however neither the `new_root` nor the `original_root` are resolved to accommodate symlinks. This results in an incorrect code path and `sam local invoke` fails to mount the source.

### DIFF
--- a/samcli/commands/_utils/template.py
+++ b/samcli/commands/_utils/template.py
@@ -247,7 +247,9 @@ def _resolve_relative_to(path, original_root, new_root):
 
     # Value is definitely a relative path. Change it relative to the destination directory
     return os.path.relpath(
-        os.path.normpath(os.path.join(original_root, path)), new_root  # Absolute original path w.r.t ``original_root``
+        # Resolve the paths to take care of symlinks
+        os.path.normpath(os.path.join(pathlib.Path(original_root).resolve(), path)),
+        pathlib.Path(new_root).resolve(),  # Absolute original path w.r.t ``original_root``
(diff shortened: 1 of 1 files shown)


mildaniel:
Hey @calabrla, thanks for the contribution. Can you please add some tests to validate this change and to make sure we don't break it in the future.

calabrla:
Hi @mildaniel I added unit tests for the `_resolve_relative_to` method that was updated and confirmed that the tests fail without my change. Let me know if you need anything else.

mildaniel (approved):

hawflau (approved):

PR #3776, opened 2022-03-24
fix: warm containers break on non-local layers

mndeveci:
Follow-up from; https://github.com/aws/aws-sam-cli/pull/3694

https://github.com/aws/aws-sam-cli/issues/3683

Running start-api with warm containers and a downloaded layer will throw an exception "Lambda functions containers initialization failed because of expected str, bytes or os.PathLike object, not NoneType"

Check if layer codeuri exists before attempting to watch it

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/local/start_api/start_api_integ_base.py
+++ b/tests/integration/local/start_api/start_api_integ_base.py
@@ -92,7 +92,10 @@ class StartApiIntegBaseClass(TestCase):
 
         while True:
             line = cls.start_api_process.stderr.readline()
-            if "(Press CTRL+C to quit)" in str(line):
+            line_as_str = str(line.decode("utf-8")).strip()
+            if line_as_str:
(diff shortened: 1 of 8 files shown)


tests/integration/local/start_api/start_api_integ_base.py
```diff
@@ -88,7 +92,10 @@ def start_api(cls):
 
         while True:
             line = cls.start_api_process.stderr.readline()
-            if "(Press CTRL+C to quit)" in str(line):
+            line_as_str = str(line.decode("utf-8")).strip()
+            if line_as_str:
```
mildaniel, line 96: Can we add this logging to start-lambda also

mildaniel (approved):

hawflau (approved):

PR #3783, opened 2022-03-29
fix: container start waits for port

torresxb1:
When doing local invoke, SAM CLI used to consider the container "started" even if the host port was not yet open, so calls to invoke the function could fail (and eventually time out or exhaust the retries if the port took long enough to be open).

This PR makes it so the container is only considered "started" once the port is open.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -40,6 +40,7 @@ from samcli.lib.utils.colors import Colored
 from samcli.lib.utils import osutils
 from samcli.lib.utils.packagetype import IMAGE, ZIP
 from samcli.lib.utils.stream_writer import StreamWriter
+from samcli.local.docker.container import ContainerStartTimeoutException
 from samcli.local.docker.lambda_build_container import LambdaBuildContainer
 from samcli.local.docker.utils import is_docker_reachable, get_docker_platform
 from samcli.local.docker.manager import ContainerManager
@@ -816,6 +817,8 @@ class ApplicationBuilder:
                     raise UnsupportedBuilderLibraryVersionError(
                         container.image, "{} executable not found in container".format(container.executable_name)
                     ) from ex
+            except ContainerStartTimeoutException as e:
+                raise BuildInsideContainerError(str(e)) from e
 
             # Container's output provides status of whether the build succeeded or failed
             # stdout contains the result of JSON-RPC call
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -29,6 +29,12 @@ class ContainerResponseException(Exception):
     """
 
 
+class ContainerStartTimeoutException(Exception):
+    """
+    Exception raised when timeout was reached while starting a container.
+    """
+
+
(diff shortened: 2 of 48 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -819,6 +820,8 @@ def _build_function_on_container(
                     raise UnsupportedBuilderLibraryVersionError(
                         container.image, "{} executable not found in container".format(container.executable_name)
                     ) from ex
+            except ContainerStartTimeoutException as e:
+                raise BuildInsideContainerError(str(e)) from e
```
torresxb1, line 824: I did this because `BuildInsideContainerError` is caught and re-raised as a UserException here: https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/build/build_context.py#L237-L251 .
`ContainerStartTimeoutException` wouldn't be caught and re-raised as a UserException otherwise and it would show the whole stack trace in a not very user friendly way. 

Is the preferred way to do it like this or to catch `ContainerStartTimeoutException` and re-raise as UserException as in the link above ^?

samcli/local/docker/container.py
```diff
@@ -17,13 +20,21 @@
...
     Exception raised when unable to communicate with RAPID APIs on a running container.
     """
 
 
+class ContainerStartTimeoutException(Exception):
+    """
+    Exception raised when timeout was reached while starting a container.
+    """
+
+
```
torresxb1, line 37: new exception because we don't want to reuse another exception and accidentally throw an exit code of zero when we don't have to (we're only doing it on the invoke to match the previous behaviour)

samcli/local/docker/container.py
```diff
@@ -272,6 +283,35 @@ def start(self, input_data=None):
...
+    def wait_for_port(self):
+        """
+        Waits until the host machine port that Docker binds to is open.
+        """
+        start_time = time.time()
+        sleep = 0.1
+        while True:
+            a_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            location = (self._container_host, self.rapid_port_host)
```
torresxb1, line 297: changed to `self._container_host` to match the host used on the invoke: https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/container.py#L282

hawflau (approved):

samcli/commands/build/build_context.py
```diff
@@ -241,6 +242,7 @@ def run(self):
             UnsupportedBuilderLibraryVersionError,
             ContainerBuildNotSupported,
             InvalidBuildGraphException,
+            ContainerStartTimeoutException,
```
torresxb1, line 245: adding here because it wouldn't be caught and re-raised as a UserException otherwise and it would show the whole stack trace in a not very user friendly way (for building with `--use-container` option since it also goes through the code that waits for the port).

mildaniel (approved):

PR #3794, opened 2022-04-01
fix: increase container start timeout

torresxb1:
Colima seemed to need more than the current default 5s when doing `sam build --use-container`

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -20,7 +20,7 @@ from .utils import to_posix_path, find_free_port, NoFreePortsError
 
 LOG = logging.getLogger(__name__)
 
-START_CONTAINER_TIMEOUT = float(os.environ.get("SAM_CLI_START_CONTAINER_TIMEOUT", 5))
+START_CONTAINER_TIMEOUT = float(os.environ.get("SAM_CLI_START_CONTAINER_TIMEOUT", 10))
 
 
 class ContainerResponseException(Exception):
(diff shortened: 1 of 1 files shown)


hawflau (approved):

samcli/local/docker/container.py
```diff
@@ -20,7 +20,7 @@
 
 LOG = logging.getLogger(__name__)
 
-START_CONTAINER_TIMEOUT = float(os.environ.get("SAM_CLI_START_CONTAINER_TIMEOUT", 5))
+START_CONTAINER_TIMEOUT = float(os.environ.get("SAM_CLI_START_CONTAINER_TIMEOUT", 10))
```
hawflau, line 23: Is there any bad if we just set an even bigger timeout? like 30s?

torresxb1: I don't think so, it's just that in my previous tests it didn't take more than like 3 seconds with Rancher Desktop, so I thought this would be safe, but it seems other alternatives might need more time. We do allow to customize this timeout through the env variable but yeah we should provide a good default. Do you suggest 30 secs? Maybe 20?

torresxb1: Actually I guess our previous default for invoke was like 3 seconds, so we probably shouldn't make it 10x what this was?

hawflau: If 20s is more than double of the average wait time for colima, let's do 20s.

mildaniel (approved):

PR #3815, opened 2022-04-13
bind samcli version to certain commit of app template

mingkun2020:
bind sam cli release with specific commit of app templates

After cloning the repo, checkout a certain commit before copying templates

We need to update the commit each time when aws-sam-cli-app-templates update before releasing

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/init/init_templates.py
```diff
@@ -66,7 +67,12 @@ def clone_templates_repo(self):
         if not self._git_repo.clone_attempted:
             shared_dir: Path = GlobalConfig().config_dir
             try:
-                self._git_repo.clone(clone_dir=shared_dir, clone_name=APP_TEMPLATES_REPO_NAME, replace_existing=True)
+                self._git_repo.clone(
```
CoshUS, line 70: Can we have two step process of clone and checkout instead of combining them into clone?
This is easier to read and we can reuse checkout in the future.

samcli/commands/init/init_templates.py
```diff
@@ -26,6 +26,7 @@
 MANIFEST_URL = "https://raw.githubusercontent.com/aws/aws-sam-cli-app-templates/master/manifest-v2.json"
 APP_TEMPLATES_REPO_URL = "https://github.com/aws/aws-sam-cli-app-templates"
 APP_TEMPLATES_REPO_NAME = "aws-sam-cli-app-templates"
+APP_TEMPLATES_REPO_COMMIT = "773d842c8f721d08c35321defa9087aaabf251f7"
```
CoshUS, line 29: Can we store this into a JSON config file instead?

CoshUS (changes requested):

samcli/commands/init/init_templates.py
```diff
@@ -26,6 +27,7 @@
 MANIFEST_URL = "https://raw.githubusercontent.com/aws/aws-sam-cli-app-templates/master/manifest-v2.json"
```
mndeveci, line 27: We need to update here where we only pull the `manifest.json` file. Otherwise it will lead to error where a template will be listed since this will download from develop HEAD, but when we clone by using this hash, it will be out of date.

mndeveci (changes requested):

samcli/commands/init/init_templates.py
mingkun2020: Good catch! Will update the `mainfest.json` url as well.

config.json
```diff
@@ -0,0 +1,3 @@
+{
```
mndeveci, line 1: Is this file in the root folder? If so are we bundling this folder with our installers? Could this be in somewhere else?

samcli/commands/init/init_templates.py
```diff
@@ -215,7 +229,11 @@ def _get_manifest(self):
         try:
             response = requests.get(MANIFEST_URL, timeout=10)
             body = response.text
-        except (requests.Timeout, requests.ConnectionError):
+            # if the commit is not exist then MANIFEST_URL will be invalid, fall back to use manifest in latest commit
+            if response.status_code == 404:
+                raise ManifestNotFoundException()
```
mndeveci, line 234: Is there a debug message for this case where this exception is handled? Otherwise can we add it here?

samcli/lib/utils/git_repo.py
```diff
@@ -162,3 +175,17 @@ def _persist_local_repo(temp_path: str, dest_dir: Path, dest_name: str, replace_
                 f"Check that you have permissions to create/delete files in {dest_dir} directory "
                 "or file an issue at https://github.com/aws/aws-sam-cli/issues"
             ) from ex
+
+    @staticmethod
+    def _checkout_commit(repo_dir: str, commit: str, git_executable: str):
```
CoshUS, line 180: Does this function need to be protected?

samcli/lib/utils/configuration.py
```diff
@@ -0,0 +1,23 @@
+""" Read info from config.json file"""
```
CoshUS, line 1: We should call this something more specific than configuration. This can be confused with GlobalConfig and toml config. Maybe build-configuration, release-configuration, or something else.

CoshUS (changes requested):

samcli/lib/utils/git_repo.py
mingkun2020: We don't use it in other places, keep it private now

samcli/lib/utils/configuration.py
mingkun2020: rename it to runtime_config.json

CoshUS (approved):

mndeveci (approved):

PR #3817, opened 2022-04-14
feat: sam init support for x-ray tracing

jonife:
This pr provides support for Monitoring in SAM/SAMCLI 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -232,9 +232,8 @@ def non_interactive_validation(func):
     required=False,
 )
 @click.option(
-    "-t",
-    "--tracing",
-    type=click.Choice(["enable", "disable"]),
+    "--tracing/--no-tracing",
+    default=None,
     help="Enable AWS X-Ray tracing for your lambda functions",
     cls=ClickMutex,
 )
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -234,7 +234,6 @@ def non_interactive_validation(func):
 @click.option(
     "-t",
     "--tracing",
-    default="disable",
     type=click.Choice(["enable", "disable"]),
     help="Enable AWS X-Ray tracing for your lambda functions",
     cls=ClickMutex,
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -235,7 +235,6 @@ def non_interactive_validation(func):
     "--tracing/--no-tracing",
     default=None,
     help="Enable AWS X-Ray tracing for your lambda functions",
-    cls=ClickMutex,
 )
 @common_options
 @non_interactive_validation
--- a/samcli/commands/init/__init__.py
+++ b/samcli/commands/init/__init__.py
@@ -231,6 +231,14 @@ def non_interactive_validation(func):
     """ """,
     required=False,
 )
+@click.option(
+    "-t",
+    "--tracing",
+    default="disable",
+    type=click.Choice(["enable", "disable"]),
+    help="Enable AWS X-Ray tracing for your lambda functions",
+    cls=ClickMutex,
+)
 @common_options
 @non_interactive_validation
 @pass_context
@@ -251,6 +259,7 @@ def cli(
     app_template,
     no_input,
     extra_context,
+    tracing,
     config_file,
     config_env,
 ):
@@ -272,6 +281,7 @@ def cli(
         app_template,
         no_input,
         extra_context,
+        tracing,
     )  # pragma: no cover
 
 
@@ -291,6 +301,7 @@ def do_cli(
     app_template,
     no_input,
     extra_context,
+    tracing,
 ):
     """
     Implementation of the ``cli`` method
@@ -339,6 +350,7 @@ def do_cli(
             name,
             no_input,
             extra_context,
+            tracing,
         )
     else:
         if not (pt_explicit or runtime or dependency_manager or base_image or architecture):
@@ -357,6 +369,7 @@ def do_cli(
             name,
             app_template,
             no_input,
+            tracing,
         )
 
 
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -89,7 +89,7 @@ def generate_application(
     no_input,
     location_opt_choice,
     tracing,
-):  # pylint: disable=too-many-arguments
+):  # pylint: disable=too-many-arguments, missing-function-docstring
     if location_opt_choice == "1":
         _generate_from_use_case(
             location,
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -326,14 +326,7 @@ def _get_app_template_properties(
 
 def prompt_user_to_enable_tracing():
     """
-    Parameters
-    ----------
-    tracing : bool
-        Boolean to active AWS Lambda X-ray Tracing
-    Returns
-    -------
-    return
-        True if X-Ray Tracing should activate for functions in the SAM template and vice versa
+    Prompt user to if X-Ray Tracing should activated for functions in the SAM template and vice versa
     """
     tracing = "disable"
     if click.confirm("\nWould you like to enable X-Ray tracing on the function(s) in your application? "):
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -45,6 +45,7 @@ def do_interactive(
     name,
     app_template,
     no_input,
+    tracing,
 ):
     """
     Implementation of the ``cli`` method when --interactive is provided.
@@ -70,6 +71,7 @@ def do_interactive(
         app_template,
         no_input,
         location_opt_choice,
+        tracing,
     )
 
 
@@ -86,6 +88,7 @@ def generate_application(
     app_template,
     no_input,
     location_opt_choice,
+    tracing,
 ):  # pylint: disable=too-many-arguments
     if location_opt_choice == "1":
         _generate_from_use_case(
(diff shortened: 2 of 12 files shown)


moelasmar:
Thanks @jonife for your participation.

moelasmar:
Could you please check the failed test cases. It seems it is failing in something related to the init workflow.

samcli/commands/init/__init__.py
```diff
@@ -231,6 +231,14 @@ def wrapped(*args, **kwargs):
     """ """,
     required=False,
 )
+@click.option(
+    "-t",
+    "--tracing",
+    default="disable",
+    type=click.Choice(["enable", "disable"]),
+    help="Enable AWS X-Ray tracing for your lambda functions",
```
mndeveci, line 239: We can check with UX or documentation team whether we need more information about this flag.

samcli/commands/init/__init__.py
```diff
@@ -231,6 +231,14 @@ def wrapped(*args, **kwargs):
     """ """,
     required=False,
 )
+@click.option(
+    "-t",
+    "--tracing",
+    default="disable",
+    type=click.Choice(["enable", "disable"]),
```
mndeveci, line 238: Aren't we going to set the flag name as `--enable-tracing/--disable-tracing` which would be a boolean flag?

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,188 @@
...
+        Gets data the SAM templates and returns it in a array
+
+        Returns
+        -------
+        list
+            array with updated template data
+        """
+        file = open(self.template_location, "r")
+        template = file.readlines()
+        return template
```
mndeveci, line 188: We should close the file after reading it (for not leaking resource)
```suggestion
        with open(self.template_location, "r") as file:
            return file.readlines()
```

mndeveci (changes requested):

samcli/commands/init/interactive_init_flow.py
```diff
@@ -305,6 +323,25 @@ def _get_app_template_properties(
     return (runtime, base_image, package_type, dependency_manager, template_chosen)
 
 
+def prompt_user_to_enable_tracing():
+    """
```
hawflau, line 327: docstring seems not matching what the function actually accepts and returns. Can you please update it?

samcli/lib/init/__init__.py
```diff
@@ -28,6 +29,7 @@ def generate_project(
     name=None,
     no_input=False,
     extra_context=None,
+    tracing="Disable",
```
hawflau, line 32: My understanding is it's either "enabled" or "disabled". Why not use boolean? 
Also, it looks inconsistent that the "Disable" value is with uppercase "D", but the "enable" value is with lowercase "e".

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,188 @@
...
+"""
+import logging
+from yaml.parser import ParserError
+from samcli.yamlhelper import parse_yaml_file
+
+
+LOG = logging.getLogger(__name__)
+
+
+class TemplateModifier:
```
hawflau, line 12: Since this is a completely new piece of code, I'd encourage you to add type checking

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,188 @@
...
+        Parameters
+        ----------
+        section : str
+            A section in the SAM template
+        position : int
+            position to start searching for the section
+
+        Returns
+        -------
+        tuple
```
hawflau, line 89: It doesn't look like the method is returning any tuple in any branch

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,188 @@
...
+from samcli.yamlhelper import parse_yaml_file
+
+
+LOG = logging.getLogger(__name__)
+
+
+class TemplateModifier:
+    GLOBALS = "Globals:\n"
+    RESOURCES = "Resources:\n"
+    FUNCTION = "  Function:\n"
```
hawflau, line 15: Most of the init templates are using 2 spaces for indentation. But yaml has no strict requirement for indentation as long as it's indented by space(s) but not tab(s). 

If we accidentally accept PR that adds an init template with non-2-spaces indentation, will this break?
If so, can we add some form of protective mechanism? e.g. enforcing 2-spaces indentation in the init templates repo?

hawflau (changes requested):

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,189 @@
...
+        -------
+        bool
+            True if templates passes sanity check else False
+        """
+        try:
+            parse_template = parse_yaml_file(self.template_location)
+            return bool(parse_template)
+        except ParserError:
+            link = "https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html"
+            message = f"Error: Unable to add Tracing to the project. To learn more Tracing visit {link}"
```
valerena, line 162: It should be 
> ... to learn more **about** Tracing

And potentially you can get an overall better message from UX or PM

samcli/commands/init/interactive_init_flow.py
```diff
@@ -86,7 +88,8 @@ def generate_application(
     app_template,
     no_input,
     location_opt_choice,
-):  # pylint: disable=too-many-arguments
+    tracing,
+):  # pylint: disable=too-many-arguments, missing-function-docstring
```
mndeveci, line 92: Can we write function doc string instead of disabling pylint here?

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,191 @@
...
+        Parameters
+        ----------
+        template : list
+            array with updated template data
+        """
+        with open(self.template_location, "w") as file:
+            for line in template:
+                file.write(line)
+
+    def get_template(self) -> list:
```
mndeveci, line 181: If this is array of string, then you can define this in the typing information as `List[str]`

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,191 @@
...
+        list
+            array with updated template data
+        """
+        template = self.template[position:]
+        for index, line in enumerate(template):
+            if not (line.startswith(" ") or line.startswith("#")):
+                return self.template[: position + index] + fields + self.template[position + index :]
+        return self.template
+
+    def field_position(self, position, field) -> Any:
```
mndeveci, line 122: missing typing for `position` and `field`.

samcli/lib/init/cli_template_modifier.py
mndeveci: This class sounds like it could be used for any field to update/change template. But it is only working for `Tracing` at the moment. There is no generic way to extend this to support other fields at the moment. For that reason, I would consider change naming of this class specific to Tracing, and/or extend generic parts into separate class.

samcli/lib/init/cli_template_modifier.py
hawflau: To record the discussions around this topic:
Yes, we discovered several use cases that `Tracing` can't be added successfully. In that case, command will not break but will only output the template without `Tracing`. A warning message will be shown to guide customers to add `Tracing` manually.

We are okay with the current approach for now. But in future if we need extensive support in modifying yaml while preserving comments, we should consider using a yaml library with such support, e.g. raumel.

hawflau: To record the discussions around this topic:
Yes, we discovered several use cases that `Tracing` can't be added successfully. In that case, command will not break but will only output the template without `Tracing`. A warning message will be shown to guide customers to add `Tracing` manually.

We are okay with the current approach for now. But in future if we need extensive support in modifying yaml while preserving comments, we should consider using a yaml library with such support, e.g. raumel.

jonife: `Tracing` will be added successfully except for 1 situation where the customer's template has weird spacing. In which case it reverts back to the users template with a warning message.

hawflau (changes requested):

samcli/commands/init/__init__.py
```diff
@@ -231,6 +231,12 @@ def wrapped(*args, **kwargs):
     """ """,
     required=False,
 )
+@click.option(
+    "--tracing/--no-tracing",
+    default=None,
+    help="Enable AWS X-Ray tracing for your lambda functions",
+    cls=ClickMutex,
```
hawflau, line 238: I don't think we need to use `ClickMutex` class, right? Please consider removing this line

hawflau (changes requested):

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,211 @@
...
+            field and it value
+        """
+        return self.tracing if field == "Tracing" else None
+
+
+class TemplateModifier:
+    def __init__(self, location):
+        self.template_location = location
+        self.template = self._get_template()
+        self.copy_of_original_template = self.template
```
mndeveci, line 43: Should we do a `deepcopy` here for the original template? In case it is been changed in the process, this will still write the changed template instead of the original one.

samcli/lib/init/cli_template_modifier.py
```diff
@@ -0,0 +1,211 @@
+"""
```
mndeveci, line 1: Sorry for bugging this again, but this file and all classes are only working for enabling xray tracing. But they have very generic names, it sounds like these can be used to modify template but it is not actually. Please either extract generic parts into classes and files which will have generic names, otherwise change these names into xray/tracing specific.

mndeveci (changes requested):

mndeveci (approved):

hawflau (approved):

singledigit:
Hey gang, this is a great addition. I wonder of it makes sense to add a Globals entry for Step Functions and API GW to enable tracing as well?

Issue #3208, opened 2021-08-25
When testing locally with sam local start-api... somes properties name got their first letter Capitalize ?!


stevesirois:
### Description:
I'm coding a Lambda function that need to extract some claims from a JWT token. When the code is deploy on AWS, everything is fine but locally I got an error concerning the access to that token. The problem raise from the fact that locally, in the event object, the property read "Authorization" instead of "authorization". So the token is no more accessible with this annotation : event.headers.authorization ( this is now undefined in nodejs)

### Steps to reproduce:
Simply create a Lamdba with sam template using hello world, start it locally then use a tool like Postman and add Bearer Token Authorization to a simple GET REST call. In the code, simply send to the console the event object. 

### Observed result:
When deploy on AWS --> console.info(event)
...
accept: '*/*',
'accept-encoding': 'gzip, deflate',
authorization: 'Bearer eyJraWQiOiJkWk0xe ... Q7zjKbuYFxkb1wvXwlO2',
'cache-control': 'no-cache',
'content-length': '0',
 ....
When running locally ( sam local start-api )  --> console.info(event)
...
START RequestId: d3e77593-5853-419a-86c8-d89f8b3041e0 Version: $LATEST
    Authorization: 'Bearer eyJraWQiOiJkWk0xe ... Q7zjKbuYFxkb1wvXwlO2',
    isBase64Encoded: false
...
### Expected result:
No capital 'A' but all lowercase property name
authorization: 'Bearer eyJraWQiOiJkWk0xe ... Q7zjKbuYFxkb1wvXwlO2'

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Ubuntu 20.04.2 LTS
2. SAM CLI : 1.29.0
3. AWS region: ca-central-1

### Additional comments

I also tried it in debug mode in VS Code and in the launch.json, I added a headers section in the api, all lowercase. But when it run, the output of the object show a capital A for Authorization. I also realized that other property like accept ou origin got the same treatment. I also tried to JSON.parse the object to see if it "normalize" it but it's the same result... AWS all lowercase, local sam get is first letter in uppercase. I also tried a direct invocation with sam local invoke with a event.json file, but this time it keep it in lowercase!? So this behaviour seems only affecting the api invocation.

prenx4x:
Hello,
As per RFC 2616 headers are not case sensitive. This is discussed in depth in previous issue https://github.com/aws/aws-sam-cli/issues/1860.
[...]

Issue #1860, opened 2020-03-10
Headers are received in Camel-Case


jiaming10:
When running `sam local start-api`, I'm getting headers in Camel-Case, (see **X-Api-Key** and **X-User-Id**), however. the deployed lambda function is receiving the headers in lower-case (**x-api-key** and **x-user-id**).

My local code is working as of 2020-03-09 17:00 GMT+8, and I'm trying to understand why the behaviour is different now.

In `sam local`:
```python
{
  'headers': {
    'X-Api-Key': '<redacted>',
    'X-User-Id': '<redacted>',
    'User-Agent': 'PostmanRuntime/7.22.0',
    'Accept': '*/*',
    'Cache-Control': 'no-cache',
    'Host': 'localhost:3000',
    'Accept-Encoding': 'gzip, deflate, br',
    'Connection': 'keep-alive',
    'X-Forwarded-Proto': 'http',
    'X-Forwarded-Port': '3000'
  },
  'multiValueHeaders': {
    'X-Api-Key': ['<redacted>'],
    'X-User-Id': ['<redacted>'],
    'User-Agent': ['PostmanRuntime/7.22.0'],
    'Accept': ['*/*'],
    'Cache-Control': ['no-cache'],
    'Host': ['localhost:3000'],
    'Accept-Encoding': ['gzip, deflate, br'],
    'Connection': ['keep-alive'],
    'X-Forwarded-Proto': ['http'],
    'X-Forwarded-Port': ['3000']
  }
}
```

In AWS Lambda:
```python
{
  'headers': {
    'Accept': '*/*',
    'Accept-Encoding': 'gzip, deflate, br',
    'Cache-Control': 'no-cache',
    'Host': '<redacted>',
    'User-Agent': 'PostmanRuntime/7.22.0',
    'X-Amzn-Trace-Id': 'Root=1-5e6739e1-d6161c849faa6e9a369f680c',
    'x-api-key': '<redacted>',
    'X-Forwarded-For': '175.143.101.237',
    'X-Forwarded-Port': '443',
    'X-Forwarded-Proto': 'https',
    'x-user-id': '<redacted>'
  },
  'multiValueHeaders': {
    'Accept': ['*/*'],
    'Accept-Encoding': ['gzip, deflate, br'],
    'Cache-Control': ['no-cache'],
    'Host': ['<host>'],
    'User-Agent': ['PostmanRuntime/7.22.0'],
[...]

PR #3821, opened 2022-04-17
fix: modify local start-api event headers to lower case

JiJiHoon:

#3208

It fixes event.headers do not match the specification.
[HTTP APIs, Format 2.0](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)

make headers key to lower case.
modify test case expected values.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar:
Thanks @JiJiHoon for your participation.

moelasmar (approved):

mingkun2020 (approved):
LGTM! Thanks @JiJiHoon for your contribution!

jfuss:
This looks like it lowers all the headers keys, which doesn't match Api Gateway. This has been discussed at good lengths here https://github.com/aws/aws-sam-cli/issues/1860, specifically here: https://github.com/aws/aws-sam-cli/issues/1860#issuecomment-888800663. 

Given past discussions, I don't think we should be taking in this contribution.

mndeveci:
I am closing this PR as per previous discussions. This was also mentioned in the issue linked this one, please check [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#page-31) where they specifically mention that http header names are case insensitive;
> HTTP header fields, which include general-header ([section 4.5](https://datatracker.ietf.org/doc/html/rfc2616#section-4.5)),
   request-header ([section 5.3](https://datatracker.ietf.org/doc/html/rfc2616#section-5.3)), response-header ([section 6.2](https://datatracker.ietf.org/doc/html/rfc2616#section-6.2)), and
   entity-header ([section 7.1](https://datatracker.ietf.org/doc/html/rfc2616#section-7.1)) fields, follow the same generic format as
   that given in [Section 3.1 of RFC 822](https://datatracker.ietf.org/doc/html/rfc822#section-3.1) [[9](https://datatracker.ietf.org/doc/html/rfc2616#ref-9)]. Each header field consists
   of a name followed by a colon (":") and the field value. Field names
   are case-insensitive.

Issue #3714, opened 2022-03-07
sam local invoke does not have response every time


Li-ReDBox:
### Description:

Often times, `sam local invoke  HelloWorldFunction` receives output of:
`No response from invoke container for HelloWorldFunction`

### Steps to reproduce:

```
for i in {1..10}
do
   sam local invoke  HelloWorldFunction
   echo; echo
   sleep 5
done
```

### Observed result:

```
for i in {1..10}
> do
>    sam local invoke  HelloWorldFunction
>    echo; echo
>    sleep 5
> done
Invoking app.lambda_handler (python3.9)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-python3.9:rapid-1.40.1-x86_64.

Mounting li-sam-app-py/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
No response from invoke container for HelloWorldFunction


Invoking app.lambda_handler (python3.9)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-python3.9:rapid-1.40.1-x86_64.

Mounting li-sam-app-py/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
No response from invoke container for HelloWorldFunction


Invoking app.lambda_handler (python3.9)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-python3.9:rapid-1.40.1-x86_64.

Mounting li-sam-app-py/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
No response from invoke container for HelloWorldFunction


Invoking app.lambda_handler (python3.9)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-python3.9:rapid-1.40.1-x86_64.

Mounting li-sam-app-py/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
No response from invoke container for HelloWorldFunction


Invoking app.lambda_handler (python3.9)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-python3.9:rapid-1.40.1-x86_64.

Mounting li-sam-app-py/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
START RequestId: 033c0575-1b6d-4828-bc46-ab35e28d39f4 Version: $LATEST
{}
[...]

PR #3831, opened 2022-04-22
fix: wait for socket connection before invoking/executing function in local invoke, start-api, etc.

torresxb1:

https://github.com/aws/aws-sam-cli/issues/3714
https://github.com/aws/aws-sam-cli/issues/3715

When doing a local invoke, we attempt to establish a connection to the socket used to communicate with Docker (when making the post request), but it'll fail if the socket is not listening yet. Notably, this happens with Docker Desktop alternatives like Rancher Desktop and Colima, where it'll make the post request while the socket isn't listening yet, exhausts the retries for this, and come back with the message `No response from invoke container ...`. 

This PR introduces a wait until we're able to establish a connection to the socket before making the post request.

I had to change/move the timer (for function timeout) logic in order to start the timer after waiting for the socket, because the default function timeout is 3 seconds and waiting for the socket can take this amount of time (so it would timeout before even executing/invoking the function).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/local/docker/container.py
```diff
@@ -194,13 +205,13 @@ def create(self):
 
         return self.id
 
-    def stop(self, time=3):
+    def stop(self, timeout=3):
```
torresxb1, line 208: renamed due to naming conflict with import

samcli/local/docker/container.py
```diff
@@ -296,7 +307,16 @@ def wait_for_result(self, full_path, event, stdout, stderr):
...
+        # wait_for_http_response will attempt to establish a connection to the socket
+        # but it'll fail if the socket is not listening yet, so we wait for the socket
+        self.wait_for_socket_connection()
+
+        # start the timer for function timeout right before executing the function, as waiting for the socket
+        # can take some time
+        timer = start_timer() if start_timer else None
         self.wait_for_http_response(full_path, event, stdout)
+        if timer:
+            timer.cancel()
```
torresxb1, line 319: I had to update/move the timer logic (originally everything happened in runtime.py and we started the timer when we called `configure_interrupt`) in order to start the timer after waiting for the socket, because the default function timeout is 3 seconds and waiting for the socket can take this amount of time (so it would timeout before even executing/invoking the function).

samcli/local/docker/container.py
```diff
@@ -314,6 +334,32 @@ def wait_for_logs(self, stdout=None, stderr=None):
...
+
+            if is_port_open:
+                break
+
+            current_time = time.time()
+            if current_time - start_time > START_CONTAINER_TIMEOUT:
+                raise ContainerStartTimeoutException(
+                    f"Timed out while starting container. You can increase this timeout by "
+                    f"setting the SAM_CLI_START_CONTAINER_TIMEOUT environment variable. "
+                    f"The current timeout is {START_CONTAINER_TIMEOUT} (seconds)."
```
torresxb1, line 358: Not sure that this message makes sense anymore, as this wait is not part of the start container flow anymore. This is still related to starting a container, since after the container is started, it can take a bit of time for the socket to start listening. But this is also an independent/standalone function for waiting for a socket connection, so I wonder if I should update this to `Timed out while attempting to establish a connection to the container ...` or something like that (and the corresponding variable/constant/etc changes). In the end, we want a balance between an accurate message but also one that is clear for users. Thoughts?

samcli/local/docker/container.py
```diff
@@ -314,6 +334,32 @@ def wait_for_logs(self, stdout=None, stderr=None):
 
         self._write_container_output(logs_itr, stdout=stdout, stderr=stderr)
 
+    def wait_for_socket_connection(self):
+        """
+        Waits for a successful connection to the socket used to communicate with Docker.
+        """
+        start_time = time.time()
+        sleep = 0.1
+        while True:
```
super132, line 343: Should we wait for `is_port_open` instead?

torresxb1: we have a `break` inside this loop which should accomplish the same thing

super132: My general practice will be waited on a variable instead of `True` as waiting on `True` have a higher chance to lead to infinite loop execution than a variable. This is not critical but I'd generally advocate not to use a constant in a loop if possible.

samcli/local/docker/container.py
super132: This method is being used by `wait_for_result()` only, which is intended to be used by different use cases. The exception `ContainerStartTimeoutException` can be misleading here. Should we rename the exception be `ContainerTimeoutException` or equivalent and using the message that you suggested?

> But this is also an independent/standalone function for waiting for a socket connection

Is it the intent of this method? From the use case in this change, it seems to be a private method of `Container` class to wait for a connection to be ready to represent a container is ready to serve traffic. Is this method designed to be used outside of `Container` class at all? If no then may be adding an underscore prefix is better?

samcli/local/docker/container.py
```diff
@@ -314,6 +334,32 @@ def wait_for_logs(self, stdout=None, stderr=None):
 
         self._write_container_output(logs_itr, stdout=stdout, stderr=stderr)
 
+    def wait_for_socket_connection(self):
```
mndeveci, line 337: Can we make this method protected since it is only used by another method in this class?

samcli/local/docker/container.py
mndeveci: Agreed to change the messaging (`Timed out while attempting to establish a connection to the container ...`) and also the exception name (`ContainerTimeoutException`) to align with that.

mndeveci (changes requested):

samcli/local/docker/container.py
hawflau: should `START_CONTAINER_TIMEOUT` also be changed to another name?

samcli/local/docker/container.py
torresxb1: done 👍

samcli/local/docker/container.py
torresxb1: done 👍

samcli/local/docker/container.py
torresxb1: done 👍

samcli/local/docker/container.py
```diff
@@ -17,13 +20,21 @@
 
 LOG = logging.getLogger(__name__)
 
+CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT", 20))
```
hawflau, line 23: nit: should we handle any error if the provided `SAM_CLI_CONTAINER_CONNECTION_TIMEOUT` is not a number? e.g. if someone sets `export SAM_CLI_CONTAINER_CONNECTION=abc`

hawflau (approved):

samcli/local/docker/container.py
```diff
@@ -334,32 +334,30 @@ def wait_for_logs(self, stdout=None, stderr=None):
...
-    def wait_for_socket_connection(self):
+    def _wait_for_socket_connection(self) -> None:
         """
         Waits for a successful connection to the socket used to communicate with Docker.
         """
-        start_time = time.time()
-        sleep = 0.1
-        while True:
+
+        def can_connect_to_socket() -> bool:
```
mndeveci, line 342: nit: this function could be defined at class level with an underscore in the beginning. The more we add inline functions the harder it is to understand what is going on.

mndeveci (approved):

samcli/local/docker/container.py
torresxb1: done 👍

PR #3838, opened 2022-04-26
chore: add gh action to update app templates hash

mndeveci:
N/A

We recently added a configuration file which keeps the latest commit hash from aws/aws-sam-cli-app-templates repositorty. This GH action is updating that hash value in configuration by running it every hour (or manually).

Test run in fork: https://github.com/mndeveci/aws-sam-cli/actions/runs/2223773270
Result PR in fork: https://github.com/mndeveci/aws-sam-cli/pull/3

### DIFF
--- a/.github/workflows/update_app_templates_hash.yml
+++ b/.github/workflows/update_app_templates_hash.yml
@@ -1,0 +1,43 @@
+name: Update aws/aws-sam-cli with latest commit hash from aws/aws-sam-cli-app-templates
+
+on:
+  schedule:
+    - cron: "0 * * * *"
+  workflow_dispatch: {}
+
+jobs:
+  updateCommitHash:
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout App Templates
+        uses: actions/checkout@v3
+        with:
+          repository: aws/aws-sam-cli-app-templates
+          path: aws-sam-cli-app-templates
+
+      - name: Checkout SAM CLI
+        uses: actions/checkout@v3
+        with:
+          repository: aws/aws-sam-cli
+          path: aws-sam-cli
+
+      - name: Update commit hash in SAM CLI from App Templates
+        run: |
+          cd aws-sam-cli-app-templates
+          export APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
+          cd ../aws-sam-cli
+          cat <<< $(jq --arg commit_hash "$APP_TEMPLATES_COMMIT_HASH" --indent 4 '.app_template_repo_commit =  $commit_hash' samcli/runtime_config.json) > samcli/runtime_config.json
+
+      - name: Raise PR for SAM CLI
+        uses: peter-evans/create-pull-request@v4
+        with:
+          path: aws-sam-cli
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: update_app_templates_hash
+          delete-branch: true
+          title: 'feat: update SAM CLI with latest App Templates commit hash'
+          body:
+            This PR & commit is automatically created by GH action to update the aws/aws-sam-cli with latest hash of the aws/aws-sam-cli-app-templates.
--- a/.github/workflows/update_app_templates_hash.yml
+++ b/.github/workflows/update_app_templates_hash.yml
@@ -2,7 +2,7 @@ name: Update aws/aws-sam-cli with latest commit hash from aws/aws-sam-cli-app-te
 
 on:
   schedule:
-    - cron: "0 * * * *"
+    - cron: "0 * * * *" # run at the top of every hour
   workflow_dispatch: {}
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,43 @@
...
+          path: aws-sam-cli
+
+      - name: Update commit hash in SAM CLI from App Templates
+        run: |
+          cd aws-sam-cli-app-templates
+          export APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
+          cd ../aws-sam-cli
+          cat <<< $(jq --arg commit_hash "$APP_TEMPLATES_COMMIT_HASH" --indent 2 '.app_template_repo_commit =  $commit_hash' samcli/runtime_config.json) > samcli/runtime_config.json
+
+      - name: Raise PR for SAM CLI
```
hawflau, line 34: If the previously PR is not merged, will this close the previous PR? 
The reason for asking is that if we merge the latest PR (with the latest app template commit hash), and leave the older PR(s) open, there could be risk that we would accidentally merge them afterwards.

mndeveci: No this will update the existing PR. And if the branch is out of date (like there are new commits in develop) this will update with them as well.

hawflau (approved):

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,43 @@
...
+  workflow_dispatch: {}
+
+jobs:
+  updateCommitHash:
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout App Templates
```
jfuss, line 15: Do we need to specify a branch?

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,43 @@
+name: Update aws/aws-sam-cli with latest commit hash from aws/aws-sam-cli-app-templates
+
+on:
+  schedule:
+    - cron: "0 * * * *"
```
jfuss, line 5: Can you add a comment on what this cron is? I hate trying to read cron.

jfuss (changes requested):
Requesting changes just to understand the answer of my following question:

How does this handle if the hash is not changing? It seems like this will still create a PR but not sure what the create pr job actually handles this or not.

.github/workflows/update_app_templates_hash.yml
mndeveci: We are cloning default branch in SAM CLI and here as well. So we don't need to maintain a branch configuration here.

.github/workflows/update_app_templates_hash.yml
mndeveci: You are right, this is for every hour, let me add that as a comment

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Update commit hash in SAM CLI from App Templates
+        run: |
+          cd aws-sam-cli-app-templates
+          export APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
+          cd ../aws-sam-cli
+          cat <<< $(jq --arg commit_hash "$APP_TEMPLATES_COMMIT_HASH" --indent 2 '.app_template_repo_commit =  $commit_hash' samcli/runtime_config.json) > samcli/runtime_config.json
+
+      - name: Raise PR for SAM CLI
+        uses: peter-evans/create-pull-request@v4
```
hoffa, line 35: I'd be wary of using third-party actions without reviewing the code and pinning to a hash. If that action introduces buggy/malicious code and they update the tag `v4` to point to that commit, it'll run the code in our repository, which can be an attack vector.

I'd much rather use the official GitHub CLI which is already bundled in GitHub-hosted runners. We can use [`gh pr create`](https://cli.github.com/manual/gh_pr_create).

Something like this (untested):

```yaml
      - run: gh pr create [...]
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

mndeveci:
> How does this handle if the hash is not changing? It seems like this will still create a PR but not sure what the create pr job actually handles this or not.

I verified this behavior in the test repositories. If there is no change and workflow is going to exit without creating a PR (https://github.com/mndeveci/mock-cli/runs/6180064552?check_suite_focus=true).
```
Branch 'update_app_templates_hash' no longer differs from base branch 'develop'
Deleting branch 'update_app_templates_hash'
```

And if next time this workflow runs with a change, it will create a PR (https://github.com/mndeveci/mock-cli/runs/6180080878?check_suite_focus=true).

.github/workflows/update_app_templates_hash.yml
mndeveci: Thanks for the suggestion! I've already have alternative way to create PR through GH CLI. Let me test with that one and update the workflow definition.

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Checkout SAM CLI
+        uses: actions/checkout@v3
+        with:
+          repository: aws/aws-sam-cli
+          path: aws-sam-cli
+
+      - name: Update commit hash in SAM CLI from App Templates
+        run: |
+          cd aws-sam-cli-app-templates
+          export APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
```
hoffa, line 30: nit: `export` is unnecessary AFAICS; the variable is not needed by a subprocess

```suggestion
         APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
```

mndeveci:
Test cases;

- updating existing PR
  - https://github.com/mndeveci/mock-cli/actions/runs/2229214637
- no pr when there is no change
  - https://github.com/mndeveci/mock-cli/actions/runs/2229208188
- create new pr
  - https://github.com/mndeveci/mock-cli/runs/6183049267?check_suite_focus=true
- create new pr when branch is deleted
  - https://github.com/mndeveci/mock-cli/runs/6183019758?check_suite_focus=true

jfuss (approved):
LGTM

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,45 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+          cd aws-sam-cli-app-templates
+          APP_TEMPLATES_COMMIT_HASH=$(git rev-parse HEAD)
+          cd ../aws-sam-cli
+          git checkout -b update_app_templates_hash
+          git reset --hard develop
+          cat <<< $(jq --arg commit_hash "$APP_TEMPLATES_COMMIT_HASH" --indent 4 '.app_template_repo_commit =  $commit_hash' samcli/runtime_config.json) > samcli/runtime_config.json
```
hoffa, line 38: I ran this with `shellcheck`, and it doesn't like reading and writing to same file in the same pipeline: https://github.com/koalaman/shellcheck/wiki/SC2094

Typically what happens is that it'll open two descriptors on the same file, one for writing the other for reading, and that can cause truncation. I know that e.g. `cat foo > foo` causes the file to be truncated, but not sure for this specific case; seems it works for now.

.github/workflows/update_app_templates_hash.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Checkout SAM CLI
+        uses: actions/checkout@v3
+        with:
+          repository: aws/aws-sam-cli
+          path: aws-sam-cli
+
+      - name: Update hash & Raise PR for SAM CLI
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hoffa, line 29: nit (or tip for the future): it can help to reduce attack surface (in practice I don't see any real issue here) by separating the steps, such that _only_ the steps that require `GITHUB_TOKEN` are in their own step, and all others are in another without `GITHUB_TOKEN`

.github/workflows/update_app_templates_hash.yml
hoffa, line 38: Minor fix from https://github.com/koalaman/shellcheck/wiki/SC2046

```suggestion
          cat <<< "$(jq --arg commit_hash "$APP_TEMPLATES_COMMIT_HASH" --indent 4 '.app_template_repo_commit =  $commit_hash' samcli/runtime_config.json)" > samcli/runtime_config.json
```

hoffa (approved):

.github/workflows/update_app_templates_hash.yml
mndeveci: Good suggestion. AFAIK each step runs in separate shell that is why I need to combine everything into single step. Otherwise we need to pass these variables between steps. I will give it another try to limit the scope of this token, if so will update the PR with that.

mndeveci: Applied your suggestion and separated step into 2 to limit the token's visibility

mndeveci:
Updated test cases;
- updating existing PR
  - https://github.com/mndeveci/mock-cli/runs/6184415099?check_suite_focus=true
- no pr when there is no change
  - https://github.com/mndeveci/mock-cli/runs/6184396530?check_suite_focus=true
- create new pr
  - https://github.com/mndeveci/mock-cli/runs/6184406578?check_suite_focus=true
- create new pr when branch is deleted
  - https://github.com/mndeveci/mock-cli/runs/6184427323?check_suite_focus=true

hawflau (approved):

PR #3845, opened 2022-04-28
feat: Added Support for Test Account Management

CoshUS:

N/A

Currently multiple testing jobs share the same account which causes throttling and conflicts.

Added support and scripts for test account management. This will acquire a different account for each CI worker.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-windows.yml
+++ b/appveyor-windows.yml
@@ -79,6 +79,8 @@ install:
   - ps: "$env:CI_ACCESS_ROLE_AWS_SECRET_ACCESS_KEY = $env:AWS_SECRET_ACCESS_KEY"
   - ps: "$env:CI_ACCESS_ROLE_AWS_SESSION_TOKEN = $env:AWS_SESSION_TOKEN"
 
+  - ps: "aws sts get-caller-identity"
+
   - ps: "$env:AWS_ACCESS_KEY_ID = $test_env_var_json.accessKeyID"
   - ps: "$env:AWS_SECRET_ACCESS_KEY = $test_env_var_json.secretAccessKey"
   - ps: "$env:AWS_SESSION_TOKEN = $test_env_var_json.sessionToken"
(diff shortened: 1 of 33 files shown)


appveyor-windows.yml
```diff
@@ -63,6 +63,36 @@ install:
...
+  - ps: "$env:AWS_SECRET_ACCESS_KEY = $test_env_var_json.secretAccessKey"
+  - ps: "$env:AWS_SESSION_TOKEN = $test_env_var_json.sessionToken"
+  - ps: "$env:TASK_TOKEN = $test_env_var_json.taskToken"
+  - ps: "$env:AWS_S3_TESTING = $test_env_var_json.TestBucketName"
+  - ps: "$env:AWS_ECR_TESTING = $test_env_var_json.TestECRURI"
+  - ps: "$env:AWS_KMS_KEY = $test_env_var_json.TestKMSKeyArn"
+  - ps: "$env:AWS_SIGNING_PROFILE_NAME = $test_env_var_json.TestSigningProfileName"
+  - ps: "$env:AWS_SIGNING_PROFILE_VERSION_ARN = $test_env_var_json.TestSigningProfileARN"
+
+  - ps: "aws sts get-caller-identity"
```
mndeveci, line 94: Should we move this part into python file so that we can call it from both windows and linux jobs?

appveyor-windows.yml
```diff
@@ -74,15 +104,25 @@ install:
...
+on_finish:
+  - ps: "aws sts get-caller-identity"
+
+  - ps: "$env:AWS_ACCESS_KEY_ID = $env:CI_ACCESS_ROLE_AWS_ACCESS_KEY_ID"
+  - ps: "$env:AWS_SECRET_ACCESS_KEY = $env:CI_ACCESS_ROLE_AWS_SECRET_ACCESS_KEY"
+  - ps: "$env:AWS_SESSION_TOKEN = $env:CI_ACCESS_ROLE_AWS_SESSION_TOKEN"
+
+  - ps: "aws sts get-caller-identity"
+
+  - ps: 'aws stepfunctions send-task-success --task-token "$env:TASK_TOKEN" --task-output "{}"'
```
mndeveci, line 125: Same goes with this part;
> Should we move this part into python file so that we can call it from both windows and linux jobs?

appveyor-windows.yml
CoshUS: We can't set env vars in a parent process from a child process. This must be done inside of the CI script.

tests/get_testing_resources.py
```diff
@@ -0,0 +1,53 @@
+import json
```
mndeveci, line 1: Let's add some documentation about this file

mingkun2020 (approved):
LGTM

PR #3853, opened 2022-05-05
feat: Nested stack support for sync watch

qingchm:
N.A.

To support nested stacks for sync watch.

By adding multiple template triggers to the watch observer.

Not known

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/watch_manager.py
```diff
@@ -120,10 +120,24 @@ def _add_code_triggers(self) -> None:
...
+                LOG.debug(
+                    "Template %s not watched due to template file validation failed for stack %s.",
+                    template,
+                    stack.name
+                )
+                continue
+
+            if not template_trigger:
+                continue
+            self._observer.schedule_handlers(template_trigger.get_path_handlers())
```
mndeveci, line 140: I would move these 3 lines inside try/except block so that it will be more easy to follow-up. And what is the reason that `template_trigger` might be `None`?

samcli/lib/utils/resource_trigger.py
```diff
@@ -103,22 +103,28 @@ def get_dir_path_handler(dir_path: Path, ignore_regexes: Optional[List[str]] = N
...
+            Stack name of the template
         on_template_change : OnChangeCallback
             Callback when template changes
         """
         super().__init__()
         self._template_file = template_file
+        self._stack_name = stack_name
         self._on_template_change = on_template_change
         self._validator = DefinitionValidator(Path(self._template_file))
+        if not self._validator.validate():
```
mndeveci, line 126: Why we need to call `validate` here since we didn't need it earlier?

qingchm: Because I don't want us to create a code trigger when the template is not valid. Since for nested stacks we have multiple template triggers it is easier to track it this way by providing information on stack + template name before other errors occur. Since watch is still in beta I wouldn't consider this as a breaking change either.

samcli/lib/sync/watch_manager.py
qingchm: The structure is following the add code triggers method above. Yes there is no reason yet that a trigger can be none. It can be removed.

samcli/lib/utils/resource_trigger.py
```diff
@@ -103,22 +103,28 @@ def get_dir_path_handler(dir_path: Path, ignore_regexes: Optional[List[str]] = N
...
         on_template_change : OnChangeCallback
             Callback when template changes
         """
         super().__init__()
         self._template_file = template_file
+        self._stack_name = stack_name
         self._on_template_change = on_template_change
         self._validator = DefinitionValidator(Path(self._template_file))
+        if not self._validator.raw_validate():
+            raise InvalidTemplateFile(self._template_file, self._stack_name)
```
mndeveci, line 127: This will break a use case where user will have an invalid template and start `sam sync --watch` and that template file won't be watched. If user fixes their template and expects sam to sync it, it won't happen. So I think we shouldn't be failing for invalid templates here.

mndeveci (changes requested):

samcli/lib/utils/resource_trigger.py
mndeveci: This will break a use case where user will have an invalid template and start `sam sync --watch` and that template file won't be watched. If user fixes their template and expects sam to sync it, it won't happen. So I think we shouldn't be failing for invalid templates here.

qingchm: @mndeveci Yes I already added warning message on any template validation failure. The executions happen in an order from queuing the infra sync to creating the trigger. I do agree that catching the failure here and not creating the trigger is too late since the failed deployment should have already happened. I will take this raw validation part out of the initialization and print a warning message when a template is invalid

samcli/lib/providers/exceptions.py
```diff
@@ -24,6 +24,35 @@ class RemoteStackLocationNotSupported(Exception):
...
+
+    _template: str
+    _stack_name: str
+
+    def __init__(self, template: str, stack_name: str) -> None:
+        """Exception when template validation fails
+
+        Parameters
+        ----------
+        template_location : str
```
mndeveci, line 38: nit: parameter name is different now

samcli/lib/sync/watch_manager.py
```diff
@@ -113,17 +113,28 @@ def _add_code_triggers(self) -> None:
...
-        template_trigger = TemplateTrigger(self._template, lambda _=None: self.queue_infra_sync())
-        self._observer.schedule_handlers(template_trigger.get_path_handlers())
+        stacks = SamLocalStackProvider.get_stacks(self._template)[0]
+        for stack in stacks:
+            template = stack.location
+            template_trigger = TemplateTrigger(template, stack.name, lambda _=None: self.queue_infra_sync())
+            try:
+                template_trigger.raw_validate()
+            except InvalidTemplateFile:
+                LOG.warning(self._color.yellow("Template validation failed for %s in %s"), template, stack.location)
```
mndeveci, line 135: nit: anything that we can provide a constructive feedback to customer? What they could/should do here?

samcli/lib/utils/definition_validator.py
```diff
@@ -48,13 +48,27 @@ def validate(self) -> bool:
             If detect_change is set, False will also be returned if there is
             no change compared to the previous validation.
         """
+        old_data = self._data
+
+        if not self.raw_validate():
+            return False
+        return old_data != self._data if self._detect_change else True
+
+    def raw_validate(self) -> bool:
```
mndeveci, line 57: I feel like this method should be named as `validate` and the actual `validate` method should be something like `check_change`, what do you think?

mndeveci (commented):
Thanks for addressing the feedback! I think the implementation makes sense, my only feedback is about method naming and some small comments for docs and warning output.

samcli/lib/utils/definition_validator.py
qingchm: Yes makes sense. `validate_template` and `validate_change` would make sense to me.

samcli/lib/sync/watch_manager.py
qingchm: Yep I'll add suggest them to make sure the template path is valid and the template file is in the right format! @mndeveci

mndeveci (approved):

mingkun2020 (approved):
LGTM

Issue #3850, opened 2022-05-04
Enable X-Ray tracing of API Gateway when sam init with Tracing option


Tak1wa:
### Describe your idea/feature/enhancement

Enable X-Ray tracing of API Gateway when sam init with Tracing option

### Proposal

I tried X-Ray Tracing. This is a great feature!  
Release 1.48.0 --Sam Init Support for XRay Tracing

This feature makes Globals --Functions --Tracing Active.  
Globals --Api --TracingEnabled is not set at all, but if it is also set, API Gateway X-Ray tracing will also be enabled.

jfuss:
@Tak1wa Thanks for the issue. This sounds like a good idea and we could even expand it to Step Functions as well. 

The file that would need to be modified is https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py.

We are happy to look at a PR for this addition.

Tak1wa:
@jfuss 

Thank you for reply.

I am considering how to implement it.
Unfortunately, Step Functions doesn't seem to support the global section.
The base code is a mechanism for injecting parameters into the global section. Therefore, I think it is better to carefully consider whether to support Step Functions.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html

mndeveci:
This is been released with [SAM CLI v1.51.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.51.0)

PR #3856, opened 2022-05-06
feat: sam init support for x-ray tracing (API Gateway)

Tak1wa:

#3850 

With #3817, you can now enable X-Ray with sam init. However, it is only a Lambda function and is not enabled for API Gateway. API Gateway can also be enabled using global parameters, so it is useful to include it in this X-Ray trace option.

When the option is enabled, set Api-TracingEnabled in the Globals section.

A field was added to the end of the Globals section when the X-Ray trace of the Lambda function was enabled.
Since Api and Functions may be mixed in random order, it is modified so that the insertion position of the field can be determined according to the nesting level by white space, so the X-Ray tracing enablement process of the Lambda function will be affected. However, the test has passed.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
```diff
@@ -29,27 +32,42 @@ def _add_new_field_to_template(self):
...
-                _section_position = function_section_position
+                else:
+                    self.template = self._add_fields_to_section(function_section_position, [self.TRACING_FUNCTION])
 
             else:
-                new_fields = [self.FUNCTION, self.TRACING]
-                _section_position = global_section_position + 1
+                self.template = self._add_fields_to_section(
+                    global_section_position, [self.FUNCTION, self.TRACING_FUNCTION]
+                )
```
mndeveci, line 47: I wonder if we should move individual sources into their own functions?

mndeveci (commented):
Thanks a lot for your contribution 🎉 Looks good in general, added single comment.

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
Tak1wa: @mndeveci Thank you for your review!

There are several patterns, such as adding the global section itself, adding only the API section or Function section, and adding only the field without affecting the existing fields.

We can consider separating them into individual Functions, but this can be a complex implementation.

mndeveci: I've pushed a change to your PR, please let me know if you are good with it.

mndeveci (approved):

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
```diff
@@ -26,36 +29,58 @@ def _add_new_field_to_template(self):
...
+        ]
+        self.template = (
+                self.template[:resource_section_position]
+                + globals_section_data
+                + self.template[resource_section_position:]
+        )
 
-                new_fields = [self.TRACING]
-                _section_position = function_section_position
+    def _add_tracing_to_api(self, global_section_position):
```
jfuss, line 54: `_add_tracing_to_api` and `_add_tracing_to_function` are almost identical other than the field they looks for and replace. Could we make a single `_add_tracing_to_type(self, global_section_position, type, tracing_config_to_add)` and pass in the different properties from line 32 and 33?

Can you add typing to the function definition too?

jfuss (changes requested):
Looks good overall @Tak1wa. Just a small comment about reducing common functions to make the code a little cleaner than you have it now.

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
jfuss: @mndeveci Reached out and looks like he had the opposite view. I will connect with him tomorrow during work hours and will respond back. Sorry for the back and forth on this, I didn't look at his comments but instead just went to the diff.

mndeveci: Apologies for confusion, I've agreed what you said @jfuss, and addressed them with my latest commit.

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
Tak1wa: @mndeveci 
thank you! I have confirmed the update! It has improved a lot.

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
Tak1wa: @jfuss 
Thank you for your review.
I agree with your advice.

And @mndeveci made a great fix.

jfuss (approved):

Issue #3855, opened 2022-05-05
sam deploy fails with PermissionError when trying to write to  `/.aws-sam`


LaffeyChris:
### Description:
When running `sam deploy` (v1.47.0), I encounter a `PermissionError`, as sam is trying to write to `/.aws-sam`. 

As I am running `sam deploy` on a shared CI/CD instance, I do not have access to write within the root directory. 
Looks like the issue is happening when setting up config/telemetry data. 

I traced this back to the last working version (v1.33.0), and it seems that the OS level errors are no longer caught, as is done here: 
https://github.com/aws/aws-sam-cli/blob/v1.33.0/samcli/cli/global_config.py#L80

### Steps to reproduce:
Run `sam deploy` on an instance that does not have permissions to write to `/.aws-sam` dir.

### Observed result:

```
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/telemetry/metric.py", line 151, in wrapped
    metric = Metric(metric_name)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/telemetry/metric.py", line 318, in __init__
    self._add_common_metric_attributes()
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/telemetry/metric.py", line 341, in _add_common_metric_attributes
    self._data["installationId"] = self._gc.installation_id
[...]

PR #3859, opened 2022-05-06
fix: command should not fail when global config fails to write

mndeveci:
#3855

SAM CLI writes its `GlobalConfig` object when getting or setting its persistent configuration values. If a folder is configured which user doesn't have write access, getting or setting these values might fail, which will also affect the running command.

Move write into a try/except block so that failing while persisting a configuration won't affect running command.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss:
`test_setter_cannot_open_file` test is failing. Overall this looks good to me.

mildaniel (approved):

mingkun2020 (approved):

Issue #3587, opened 2022-01-17
sam sync --watch race condition when syncing Stack with multiple Lambdas within same .NET project


theimowski:
### Description:

When trying to `sam sync --watch` a Stack with multiple .NET Lambdas that use same .NET project, there's a race condition when updating source code - likely related to fact that multiple processes are invoked for same project.

### Steps to reproduce:

1. Clone reproduction repository https://github.com/theimowski/sam-sync-watch-net-multi-lambdas
2. Run `sam sync --watch`

### Observed result:

```
> sam sync --stack-name TheimowskiTestSAMAccelerate --watch

                Managed S3 bucket: aws-sam-cli-managed-default-samclisourcebucket-***

                Default capabilities applied: ('CAPABILITY_NAMED_IAM', 'CAPABILITY_AUTO_EXPAND')
To override with customized capabilities, use --capabilities flag or set it in samconfig.toml

This feature is currently in beta. Visit the docs page to learn more about the AWS Beta terms https://aws.amazon.com/service-terms/.

The SAM CLI will use the AWS Lambda, Amazon API Gateway, and AWS StepFunctions APIs to upload your code without
performing a CloudFormation deployment. This will cause drift in your CloudFormation stack.
**The sync command should only be used against a development stack**.

Confirm that you are synchronizing a development stack and want to turn on beta features.

Enter Y to proceed with the command, or enter N to cancel:
 [y/N]: y

Experimental features are enabled for this session.
Visit the docs page to learn more about the AWS Beta terms https://aws.amazon.com/service-terms/.

Queued infra sync. Wating for in progress code syncs to complete...
Starting infra sync.
Cache is invalid, running build and copying resources to function build definition of f4f2dfc9-eddf-4622-9a5e-6006626996fa
Building codeuri: $path\sam-app\src\HelloWorld runtime: dotnetcore3.1 metadata: {} architecture: x86_64 functions: ['HelloWorldFunction', 'AnotherFunction']
Running DotnetCliPackageBuilder:GlobalToolInstall

Tool 'amazon.lambda.tools' was reinstalled with the latest stable version (version '5.2.0').
Running DotnetCliPackageBuilder:RunPackageAction
Amazon Lambda Tools for .NET Core applications (5.2.0)
Project Home: https://github.com/aws/aws-extensions-for-dotnet-cli, https://github.com/aws/aws-lambda-dotnet

Executing publish command
Deleted previous publish folder
[...]

PR #3862, opened 2022-05-06
fix: Lock functions from building with same codeuri at the same time

mildaniel:

#3587

Resolves a race condition where functions with the same codeuri attempt to modify shared resources simultaneously during a build.

Adds another lock to a function whenever functions share the same codeuri.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mndeveci:
I think we don't need following folders in the repository since they will be created once the function is built;
- `tests/integration/testdata/sync/code/after/dotnet_function/src/HelloWorld/bin/`
- `tests/integration/testdata/sync/code/before/dotnet_function/src/HelloWorld/bin/`

Rest looks good to me 👍🏼

tests/integration/testdata/sync/code/before/template-python.yaml
```diff
@@ -16,6 +16,24 @@ Resources:
         - Ref: HelloWorldLayer
```
mndeveci, line 16: This file name is no longer valid anymore (`template-python.yaml`). Let's move this function into its own template.

tests/integration/testdata/sync/code/before/dotnet_function/src/HelloWorld/obj/HelloWorld.csproj.nuget.dgspec.json
```diff
@@ -0,0 +1,81 @@
+{
```
mndeveci, line 1: Sorry if I missed this my earlier pass, but it looks like we also don't need this `obj` folder under function source. I believe this is generated during build so we don't need to push this in git.

tests/integration/testdata/sync/code/after/dotnet_function/src/HelloWorld/obj/HelloWorld.csproj.nuget.dgspec.json
```diff
@@ -0,0 +1,81 @@
+{
```
mndeveci, line 1: Same for this `obj` folder as well, we probably don't need this.

mndeveci (changes requested):

mndeveci (approved):
LGTM!

qingchm (approved):

PR #3876, opened 2022-05-11
fix: Add in trace event revision to prevent partial trace printing

qingchm:
N.A.

Since for the same trace ID, x-ray client can return multiple trace event records with different revision numbers. We need to always keep track of the latest revision number to prevent missing information.

It adds a trace revision into trace event.

Not known

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mndeveci (approved):
LGTM!

mildaniel (approved):

Issue #3189, opened 2021-08-20
exclude flag in `sam build` which will exclude a particular resource


nagmesh:
### Describe your idea/feature/enhancement
I have templates which have over 10 functions and as application grows the number of functions grows as well.

Provide a clear description. Ex. I wish SAM CLI would [...]
I want SAM CLI to leave specific resources while doing a build. I have lambda functions which I do not want sam cli to build but scan the rest of the resources 
### Proposal

Add details of how to add this to the product.

Add an exclude flag with comma separated values of resources to exclude
When SAM scans in the [iterator](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L110) remove the excluded values before starting the build.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)

ssenchenko:
To be sure, are you asking about having partial build? If your lambdas have interdependencies they need to be build together. If they don't have dependencies you can split them into separate projects and build smaller projects.

nagmesh:
Currently, I can build sam resource specifically using `sam build $LOGICAL_RESOURCE_NAME`
However, I want sam to only exclude one resource in my template and build the rest.

ssenchenko:
We have a [roadmap](https://github.com/aws/aws-sam-cli/wiki/SAM-CLI-Roadmap) here, but we also try to take in popular feature requests based on 👍🏽 from the community. If this issue gets some votes we will be adding this to our intake for prioritization.

mndeveci:
This is been released with [SAM CLI v1.51.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.51.0)

PR #3898, opened 2022-05-25
Add --exclude flag to sam build

Leo10Gama:
#3189 

This change allows users to exclude certain resources from their build.

By using the `--exclude` flag, specified resources are left alone in the build.

An additional parameter was added to the `build` command.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/build_dir/MyFunctionbf3066d1DepLayer/AWS_SAM_CLI_README
+++ b/build_dir/MyFunctionbf3066d1DepLayer/AWS_SAM_CLI_README
@@ -1,0 +1,1 @@
+This layer contains dependencies of function MyFunction and automatically added by AWS SAM CLI command 'sam sync'
--- a/build_dir/MyFunctionbf3066d1DepLayer/AWS_SAM_CLI_README
+++ b/build_dir/MyFunctionbf3066d1DepLayer/AWS_SAM_CLI_README
@@ -1,1 +1,0 @@
-This layer contains dependencies of function MyFunction and automatically added by AWS SAM CLI command 'sam sync'
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -67,7 +67,7 @@ class BuildContext:
         container_env_var: Optional[dict] = None,
         container_env_var_file: Optional[str] = None,
         build_images: Optional[dict] = None,
-        excluded_files: Optional[Tuple[str]] = None,
+        excluded_resources: Optional[Tuple[str, ...]] = None,
         aws_region: Optional[str] = None,
         create_auto_dependency_layer: bool = False,
         stack_name: Optional[str] = None,
@@ -99,7 +99,7 @@ class BuildContext:
         self._container_env_var = container_env_var
         self._container_env_var_file = container_env_var_file
         self._build_images = build_images
-        self._exclude = excluded_files
+        self._exclude = excluded_resources
         self._create_auto_dependency_layer = create_auto_dependency_layer
         self._stack_name = stack_name
         self._print_success_message = print_success_message
@@ -182,7 +182,7 @@ class BuildContext:
                 container_env_var=self._container_env_var,
                 container_env_var_file=self._container_env_var_file,
                 build_images=self._build_images,
-                excluded_files=self._exclude,
+                excluded_resources=self._exclude,
                 combine_dependencies=not self._create_auto_dependency_layer,
             )
         except FunctionNotFound as ex:
@@ -419,7 +419,7 @@ Commands you can use next
             ResourcesToBuildCollector that contains all the buildable resources.
         """
         result = ResourcesToBuildCollector()
-        excludes: List[str] = list(self._exclude) if self._exclude else []
+        excludes: Tuple[str, ...] = self._exclude if self._exclude is not None else ()
         result.add_functions(
             [
                 f
@@ -544,7 +544,7 @@ Commands you can use next
         "You can also enable this beta feature with 'sam build --beta-features'."
     )
 
-    _EXCLUDE_WARNING_MESSAGE = "Resource expected to be built, but marked as excluded.\n" "Building anyways..."
+    _EXCLUDE_WARNING_MESSAGE = "Resource expected to be built, but marked as excluded.\nBuilding anyways..."
 
     def _check_java_warning(self) -> None:
         """
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -420,14 +420,20 @@ Commands you can use next
         """
         result = ResourcesToBuildCollector()
         excludes: List[str] = list(self._exclude) if self._exclude else []
-        result.add_functions([
-            f for f in self.function_provider.get_all()
-            if (f.name not in excludes) and BuildContext._is_function_buildable(f)
-        ])
-        result.add_layers([
-            l for l in self.layer_provider.get_all()
-            if (l.name not in excludes) and BuildContext._is_layer_buildable(l)
-        ])
+        result.add_functions(
+            [
+                f
+                for f in self.function_provider.get_all()
+                if (f.name not in excludes) and BuildContext._is_function_buildable(f)
+            ]
+        )
+        result.add_layers(
+            [
+                l
+                for l in self.layer_provider.get_all()
+                if (l.name not in excludes) and BuildContext._is_layer_buildable(l)
+            ]
+        )
         return result
 
     @property
@@ -538,10 +544,7 @@ Commands you can use next
         "You can also enable this beta feature with 'sam build --beta-features'."
     )
 
-    _EXCLUDE_WARNING_MESSAGE = (
-        "Resource expected to be built, but marked as excluded.\n"
-        "Building anyways..."
-    )
+    _EXCLUDE_WARNING_MESSAGE = "Resource expected to be built, but marked as excluded.\n" "Building anyways..."
 
     def _check_java_warning(self) -> None:
         """
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -5,7 +5,7 @@ import logging
 import os
 import pathlib
 import shutil
-from typing import Dict, Optional, List, cast
+from typing import Dict, Optional, List, Tuple, cast
 
 import click
 
@@ -67,6 +67,7 @@ class BuildContext:
         container_env_var: Optional[dict] = None,
         container_env_var_file: Optional[str] = None,
         build_images: Optional[dict] = None,
+        excluded_files: Optional[Tuple[str]] = None,
         aws_region: Optional[str] = None,
         create_auto_dependency_layer: bool = False,
         stack_name: Optional[str] = None,
@@ -98,6 +99,7 @@ class BuildContext:
         self._container_env_var = container_env_var
         self._container_env_var_file = container_env_var_file
         self._build_images = build_images
+        self._exclude = excluded_files
         self._create_auto_dependency_layer = create_auto_dependency_layer
         self._stack_name = stack_name
         self._print_success_message = print_success_message
@@ -180,6 +182,7 @@ class BuildContext:
                 container_env_var=self._container_env_var,
                 container_env_var_file=self._container_env_var_file,
                 build_images=self._build_images,
+                excluded_files=self._exclude,
                 combine_dependencies=not self._create_auto_dependency_layer,
             )
         except FunctionNotFound as ex:
@@ -188,6 +191,7 @@ class BuildContext:
         try:
             self._check_java_warning()
             self._check_esbuild_warning()
+            self._check_exclude_warning()
             build_result = builder.build()
             artifacts = build_result.artifacts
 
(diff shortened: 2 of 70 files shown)


hawflau (changes requested):

build_dir/MyFunctionbf3066d1DepLayer/AWS_SAM_CLI_README
```diff
@@ -0,0 +1 @@
+This layer contains dependencies of function MyFunction and automatically added by AWS SAM CLI command 'sam sync'
```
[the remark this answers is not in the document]
Leo10Gama: I don't know how this file got added, but I will remove it in the following commit.

samcli/commands/build/build_context.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         container_env_var: Optional[dict] = None,
         container_env_var_file: Optional[str] = None,
         build_images: Optional[dict] = None,
+        excluded_files: Optional[Tuple[str]] = None,
```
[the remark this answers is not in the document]
Leo10Gama: Files is a quite vague term... Perhaps calling them `excluded_resources` would be more appropriate?

samcli/commands/build/build_context.py
```diff
@@ -415,8 +419,15 @@ def collect_all_build_resources(self) -> ResourcesToBuildCollector:
             ResourcesToBuildCollector that contains all the buildable resources.
         """
         result = ResourcesToBuildCollector()
-        result.add_functions([f for f in self.function_provider.get_all() if BuildContext._is_function_buildable(f)])
-        result.add_layers([l for l in self.layer_provider.get_all() if BuildContext._is_layer_buildable(l)])
+        excludes: List[str] = list(self._exclude) if self._exclude else []
```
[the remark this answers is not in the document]
Leo10Gama: Originally I did this since you cannot iterate over `None` and `self._excludes` is an optional value, but using `Tuple` should be just as effective. I'll implement that in the following commit.

samcli/commands/build/build_context.py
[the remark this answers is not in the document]
hawflau: yep that's better

samcli/lib/build/app_builder.py
```diff
@@ -184,6 +188,13 @@ def build(self) -> ApplicationBuildResult:
             Returns the build graph and the path to where each resource was built as a map of resource's LogicalId
             to the path string
         """
+        # Remove excluded resources
```
hawflau, line 191: Why do we need to remove excluded resources here and overwrite `self._resources_to_build`?
Doesn't `self._resources_to_build` already contain only the non-excluded resources when instantiating the `ApplicationBuilder` instance?

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -1149,3 +1150,35 @@ def test_check_esbuild_warning(self, functions, should_print, mocked_click):
...
+            base_dir="base_dir",
+            build_dir="build_dir",
+            cache_dir="cache_dir",
+            cached=False,
+            clean=False,
+            parallel=False,
+            mode="mode",
+            excluded_resources=exclude,
+        )
+        with patch.object(build_context, "get_resources_to_build") as mocked_resources_to_build:
```
hawflau, line 1177: is this patching required?

hawflau (changes requested):

samcli/commands/build/build_context.py
```diff
@@ -560,3 +579,11 @@ def _check_esbuild_warning(self) -> None:
 
         if is_building_esbuild:
             prompt_experimental(ExperimentalFlag.Esbuild, self._ESBUILD_WARNING_MESSAGE)
+
+    def _check_exclude_warning(self) -> None:
+        """
+        Prints warning message if a single resource to build is also being excluded
+        """
+        excludes: List[str] = list(self._exclude) if self._exclude else []
```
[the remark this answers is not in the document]
Leo10Gama: 👍

samcli/lib/build/app_builder.py
Leo10Gama: Initially I thought the parameters were being altered in some way behind the scenes, but further testing has shown me this is not the case, and this chunk isn't needed after all. I will revise this in my next commit.

tests/unit/commands/buildcmd/test_build_context.py
Leo10Gama: Upon further testing, no, it doesn't seem like this patching is required, since `_check_exclude_warning()` can be directly called upon creation of the `BuildContext`. I will fix this in the next commit.

samcli/lib/build/app_builder.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         container_env_var: Optional[Dict] = None,
         container_env_var_file: Optional[str] = None,
         build_images: Optional[Dict] = None,
+        excluded_resources: Optional[Tuple[str, ...]] = None,
```
hawflau, line 113: I suppose we don't need to add the argument?

hawflau (changes requested):

samcli/lib/build/app_builder.py
Leo10Gama: Right, since now the `excluded_resources` isn't included in the app builder

torresxb1 (commented):
left some comments

samcli/commands/build/command.py
```diff
@@ -120,6 +120,13 @@
     "an image URI for all functions, the default SAM CLI build images will be used.",
     cls=ContainerOptions,
 )
+@click.option(
+    "--exclude",
+    "-x",
+    default=None,
```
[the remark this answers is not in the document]
Leo10Gama: While it would ease some of the code, having it default to `None` makes it consistent with other settings like `--build-image` and `--container-env-var`

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1011,6 +1011,44 @@ def _get_python_version(self):
...
+        self._verify_resources_excluded(self.default_build_dir, excluded_resources, function_identifier)
+
+    def _verify_resources_excluded(self, build_dir, excluded_resources, function_identifier):
+        self.assertTrue(build_dir.exists(), "Build directory should be created")
+
+        build_dir_files = os.listdir(str(build_dir))
+
+        if function_identifier is not None and function_identifier not in excluded_resources:
+            for resource in excluded_resources:
+                self.assertNotIn(resource, build_dir_files)
```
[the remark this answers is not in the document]
Leo10Gama: Right, I got the logic for the test backwards 😅 will fix in the next commit.

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -1109,3 +1109,33 @@ def test_check_esbuild_warning(self, functions, should_print, mocked_click):
...
+class TestBuildContext_exclude_warning(TestCase):
+    @parameterized.expand(
+        [
+            ([], [], False),
+            ("Function", ("Function",), True),
+            ("Function", ("NotTheSameFunction",), False),
+        ]
+    )
+    @patch("samcli.commands.build.build_context.LOG")
+    def test_check_exclude_warning(self, resource_id, exclude, should_print, log_mock):
```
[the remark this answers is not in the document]
Leo10Gama: These tests would probably be good to add, yes. I'll add them in the next commit.

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -480,6 +480,96 @@ def test_must_return_many_functions_to_build(
...
+        if resource_identifier is not None and resource_identifier in excluded_resources:
+            # If building 1 and excluding it, build anyway
+            self.assertTrue(context.is_building_specific_resource)
+            self.assertIn(resource_to_exclude, ctx_resources_to_build.functions)
+        else:
+            if excluded_resources:
+                self.assertTrue(all([x not in ctx_resources_to_build.functions for x in excluded_resources]))
+                self.assertTrue(all([x.name in resources_to_build for x in ctx_resources_to_build.functions]))
+            else:
+                self.assertTrue(all([x in ctx_resources_to_build.functions for x in ctx_resources_to_build.functions]))
```
torresxb1, line 571: I think this might always be true? 😅

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -480,6 +480,96 @@ def test_must_return_many_functions_to_build(
...
+        ctx_resources_to_build = context.resources_to_build
+
+        if resource_identifier is not None and resource_identifier in excluded_resources:
+            # If building 1 and excluding it, build anyway
+            self.assertTrue(context.is_building_specific_resource)
+            self.assertIn(resource_to_exclude, ctx_resources_to_build.functions)
+        else:
+            if excluded_resources:
+                self.assertTrue(all([x not in ctx_resources_to_build.functions for x in excluded_resources]))
+                self.assertTrue(all([x.name in resources_to_build for x in ctx_resources_to_build.functions]))
```
torresxb1, line 569: could we add one more check to make sure that the non-excluded resources are in `ctx_resources_to_build.functions`? (right now I think none of the resources were built, even if only one was excluded, these tests wouldn't catch it)

tests/unit/commands/buildcmd/test_build_context.py
Leo10Gama: Right, 569 and 571 will both be true always lol. Will remove in the next commit!

tests/unit/commands/buildcmd/test_build_context.py
Leo10Gama: Was trying to get one going before, but couldn't resolve the logic. Will implement one for the next (hopefully last!) commit.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1011,6 +1011,46 @@ def _get_python_version(self):
...
+        [
+            ((), None),
+            ("FunctionOne", None),
+            ("FunctionThree", None),
+            ("FunctionOne", "FunctionOne"),
+            ("FunctionOne", "FunctionTwo"),
+        ]
+    )
+    @pytest.mark.flaky(reruns=3)
+    def test_build_without_resources(self, excluded_resources, function_identifier):
```
hawflau, line 1031: is `excluded_resources` a string or a list/tuple? The input params are strings but then is treated like a list in `_verify_resources_excluded`?

Also, since `--exclude` can accept multiple values, should there be an input that specifies multiple resources to exclude?

hawflau (changes requested):

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -558,17 +558,19 @@ def test_must_exclude_functions_from_build(
         self.assertEqual(context.stacks, [stack])
         self.assertEqual(context.is_building_specific_resource, bool(resource_identifier))
         ctx_resources_to_build = context.resources_to_build
+        final_resources = [x for x in resources_to_build if x not in excluded_resources] if excluded_resources else []
```
torresxb1, line 561: hmm I think it shouldn't be empty list if there's no excluded resources? Maybe you meant ... if excluded_resources else resources_to_build?

torresxb1: also could we rename to `non_excluded_resources` just to be clear 😅  (or maybe it's just me)

tests/integration/buildcmd/test_build_cmd.py
Leo10Gama: Right, these should be tuples. I will fix this in the next commit.

tests/unit/commands/buildcmd/test_build_context.py
Leo10Gama: Yes that is my mistake in logic, it should be `resources_to_build`. Will clear that up in the next commit!

hawflau (approved):

torresxb1 (approved):
Awesome, thanks for addressing all the feedback! 🚀 🚀

jedwards1211:
where's the documentation for the exclude format?  `sam build --debug --exclude node_modules` still shows `node_modules` being copied

torresxb1:
Hi @jedwards1211 , this is for excluding resources in your template (e.g. Lambda Functions, Layers), but not for things like `node_modules`.

jedwards1211:
@torresxb1 ah gotcha, I guess there's no way to prevent the build from copying over everything from the project dir?

mildaniel:
Hey @jedwards1211, there isn't any way of doing this right now but I understand your concern. These actions are actually done in https://github.com/aws/aws-lambda-builders so I encourage you to open up a issue there if this is a feature that you need.

paul-michalik:
How do you use it, what is the format? No matter what I try, it simply does nothing. 
`--exclude Func1 Func2 Func3` = no effect
`--exclude Func1,Func2,Func3` = no effect
`--exclude Func1` = no effect either

PR #3905, opened 2022-05-26
fix: Add lockchain and status check to prevent sync flows race condition on function update

qingchm:
N.A.

To prevent sync race condition on function flows.

Added a lockchain that both function config update and function code update need to grab.
Added a remote function status check to ensure function update was complete.

Not known.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/sync/flows/function_sync_flow.py
+++ b/samcli/lib/sync/flows/function_sync_flow.py
@@ -1,8 +1,10 @@
 """Base SyncFlow for Lambda Function"""
 from enum import Enum
 import logging
+
 import time
 from typing import Any, Dict, List, TYPE_CHECKING, cast
+from botocore.client import BaseClient
 
 from samcli.lib.providers.sam_function_provider import SamFunctionProvider
 from samcli.lib.sync.flows.alias_version_sync_flow import AliasVersionSyncFlow
@@ -101,20 +103,6 @@ class FunctionSyncFlow(SyncFlow):
 
         return sync_flows
 
-    def _verify_function_status(self) -> bool:
-        """
-        return value:
-        True if last function update was successful
-        False if last function update was still in progress or failed
-        """
-        response = self._lambda_client.get_function(FunctionName=self.get_physical_id(self._function_identifier))
-        if response.get("Configuration", {}).get("LastUpdateStatus", "") == FunctionUpdateStatus.IN_PROGRESS.value:
-            time.sleep(FUNCTION_SLEEP)
-            return self._verify_function_status()
-        if response.get("Configuration", {}).get("LastUpdateStatus", "") == FunctionUpdateStatus.SUCCESS.value:
-            return True
-        return False
-
     def _equality_keys(self):
         return self._function_identifier
 
@@ -125,3 +113,24 @@ class FunctionUpdateStatus(Enum):
     SUCCESS = "Successful"
     FAILED = "Failed"
     IN_PROGRESS = "InProgress"
+
+
+def wait_for_function_update_complete(lambda_client: BaseClient, physical_id: str) -> None:
+    """
+    Parameters
+        ----------
+        lambda_client : boto.core.BaseClient
+            Lambda client that performs get_function API call.
+        physical_id : str
+            Physical identifier of the function resource
+    """
+
+    status = FunctionUpdateStatus.IN_PROGRESS.value
+    while status == FunctionUpdateStatus.IN_PROGRESS.value:
+        response = lambda_client.get_function(FunctionName=physical_id)
+        status = response.get("Configuration", {}).get("LastUpdateStatus", "")
+
+        if status == FunctionUpdateStatus.IN_PROGRESS.value:
+            time.sleep(FUNCTION_SLEEP)
+
+    LOG.debug("Function update status on %s is now %s on cloud.", physical_id, status)
--- a/samcli/lib/sync/flows/function_sync_flow.py
+++ b/samcli/lib/sync/flows/function_sync_flow.py
@@ -118,11 +118,11 @@ class FunctionUpdateStatus(Enum):
 def wait_for_function_update_complete(lambda_client: BaseClient, physical_id: str) -> None:
     """
     Parameters
-        ----------
-        lambda_client : boto.core.BaseClient
-            Lambda client that performs get_function API call.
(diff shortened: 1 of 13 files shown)


samcli/lib/sync/flows/function_sync_flow.py
```diff
@@ -14,6 +16,7 @@
     from samcli.commands.build.build_context import BuildContext
 
 LOG = logging.getLogger(__name__)
+FUNCTION_SLEEP = 1  # used to wait for lambda function last update to be successful
```
mndeveci, line 19: Decrease it to 0.5 seconds maybe?

samcli/lib/sync/flows/layer_sync_flow.py
```diff
@@ -341,3 +350,17 @@ def gather_dependencies(self) -> List["SyncFlow"]:
 
     def _equality_keys(self) -> Any:
         return self._function_identifier, self._layer_arn, self._new_layer_version
+
+    def _verify_function_status(self) -> bool:
```
mndeveci, line 354: I think these functions are the same (the one in function sync flow), can we move them in a common place?

samcli/lib/sync/flows/function_sync_flow.py
```diff
@@ -98,5 +101,27 @@ def gather_dependencies(self) -> List[SyncFlow]:
...
+    def _verify_function_status(self) -> bool:
+        """
+        return value:
+        True if last function update was successful
+        False if last function update was still in progress or failed
+        """
+        response = self._lambda_client.get_function(FunctionName=self.get_physical_id(self._function_identifier))
+        if response.get("Configuration", {}).get("LastUpdateStatus", "") == FunctionUpdateStatus.IN_PROGRESS.value:
+            time.sleep(FUNCTION_SLEEP)
+            return self._verify_function_status()
```
mndeveci, line 113: nit: This is designed as recursive calls, should we change this to a while loop? Python has 1000 recursive call limit by default, I don't think we are going to exceed that, but just checking whether it would be safer or not.
```
>>> import sys
>>> sys.getrecursionlimit()
1000
```

mndeveci (commented):
Good deep dive and implementation! 👏🏼  Added some small comments.

samcli/lib/sync/flows/layer_sync_flow.py
qingchm: This is a non-static method that needs to be class specific. However the only common inheritance between those two classes would be SyncFlow. I don't think we should put in function specific method into the general SyncFlow class. However I can propose a new common class that both FunctionSyncFlow and FunctionLayerReferenceSync inherit. Do you think that would be a better solution?

mndeveci: Or another option, we can make this method static and add parameters for lambda client and function identifier (the only fields that is required from `self`).

I do agree for not putting this into generic sync flow class.

samcli/lib/sync/flows/function_sync_flow.py
qingchm: The reason why while loop is not preferred is that it needs to return a tuple of booleans in order to tell apart Success, InProgress and Failed. Something like
```
while in_progress and not failed:
  in_progress, not failed = verify_function_status()
```
 would work, but I feel that a recursion is cleaner.
 I can add in a max retry number into this recursion that if after 100 retries doesn't if it is still in progress then we just continue. Then the expected exception would be race condition, which is better than max recursion reached.

samcli/lib/sync/flows/layer_sync_flow.py
qingchm: Yep static method with function identifier and lambda client passed in also works! I think this is minimum change here. I'll go with this route.

samcli/lib/sync/flows/function_sync_flow.py
mndeveci: Agreed, let's start safe then 👍🏼

samcli/lib/sync/flows/function_sync_flow.py
```diff
@@ -125,3 +113,24 @@ class FunctionUpdateStatus(Enum):
...
+
+
+def wait_for_function_update_complete(lambda_client: BaseClient, physical_id: str) -> None:
+    """
+    Parameters
+        ----------
+        lambda_client : boto.core.BaseClient
+            Lambda client that performs get_function API call.
+        physical_id : str
+            Physical identifier of the function resource
```
mndeveci, line 125: nit: indentation is wrong here.

samcli/lib/sync/flows/function_sync_flow.py
```diff
@@ -125,3 +113,24 @@ class FunctionUpdateStatus(Enum):
...
+    """
+
+    status = FunctionUpdateStatus.IN_PROGRESS.value
+    while status == FunctionUpdateStatus.IN_PROGRESS.value:
+        response = lambda_client.get_function(FunctionName=physical_id)  # type: ignore
```
mndeveci, line 130: Instead of ignoring the type can we set it as dict?

samcli/lib/sync/flows/layer_sync_flow.py
```diff
@@ -328,8 +327,9 @@ def sync(self) -> None:
 
             self._lambda_client.update_function_configuration(FunctionName=function_physical_id, Layers=layer_arns)
 
-            if self._verify_function_status():
-                LOG.debug(self._color.green("Function configuration update status is now successful on cloud."))
+            lambda_client = self._lambda_client
+            physical_id = self.get_physical_id(self._function_identifier)
+            wait_for_function_update_complete(lambda_client, physical_id)
```
mndeveci, line 332: Why not passing them directly to method instead of extracting to a variable?

mndeveci (approved):

samcli/lib/sync/flows/function_sync_flow.py
qingchm: The type error here is that base client class does not contain get_function method, however boto's create client returns an instance of base client, that's why I'm ignoring the type here!

mildaniel (approved):
LGTM! Just add a few comments as we discussed.

Issue #3246, opened 2021-09-06
sam delete: allow to pass s3_bucket and s3_prefix as arguments


pjungermann:
### Describe your idea/feature/enhancement

At `sam delete` introduced at SAM CLI v1.29.0, you can delete your stack including resources like the ECR image as well as S3 objects.

Currently, you cannot pass the S3 bucket and prefix to the `sam delete` command. It is only obtained from `samconfig.toml` using the "deploy" scope. See the "workaround" at the "Additional Details" section.

If there is no such config, the deletion of S3 objects will be skipped.

Example output
```
2021-09-06 08:23:31,299 | Cannot delete s3 files as no s3_bucket found

Warning: s3_bucket and s3_prefix information could not be obtained from local config file or cloudformation template, delete the s3 files manually if required
```

### Proposal

Add arguments `--s3-bucket` and `--s3-prefix` to the command `sam delete` as they exist e.g. at `sam deploy`.

### Additional Details

Workaround:
```
cat << TOML > samconfig.toml
version=0.1
[default.deploy.parameters]
stack_name = "${STACK_NAME}"
s3_bucket  = "${S3_BUCKET}"
s3_prefix  = "${S3_PREFIX}"
TOML
```

If stack_name is not part of the config or it is not matching the stack name you may have passed as `--stack-name` argument, the config will **not** be used.

aahung:
@pjungermann thanks for the suggestion

qingchm:
I believe that this is addressed in the new 1.53.0 release. Please have a try and download our latest version of SAMCLI. If you still have any questions or issues feel free to re-open the issue!

PR #3915, opened 2022-05-27
feat: Add --s3-bucket and --s3-prefix to sam delete

lucashuy:
#3246 

Adding S3 bucket options will provide the ability to delete S3 buckets and objects if the bucket and prefix cannot be obtained from the local config file or CloudFormation template.

This change adds the `--s3-bucket` and `--s3-prefix` command line options to `sam delete`.

This change adds additional parameters to `sam delete` related components.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/delete/delete_context.py
+++ b/samcli/commands/delete/delete_context.py
@@ -335,12 +335,12 @@ class DeleteContext:
             self.cf_utils.wait_for_delete(self.stack_name)
 
         # Warn the user that s3 information is missing and to use --s3 options
-        if not self.s3_bucket or not self.s3_prefix:
-            LOG.debug("Cannot delete s3 objects as bucket or prefix is missing")
+        if not self.s3_bucket:
+            LOG.debug("Cannot delete s3 objects as bucket is missing")
             click.secho(
                 "\nWarning: Cannot resolve s3 bucket information from command options"
                 " , local config file or cloudformation template. Please use"
-                " --s3-bucket and --s3-prefix command options next time and"
+                " --s3-bucket next time and"
                 " delete s3 files manually if required.",
                 fg="yellow",
             )
--- a/tests/unit/commands/delete/test_delete_context.py
+++ b/tests/unit/commands/delete/test_delete_context.py
@@ -182,7 +182,7 @@ class TestDeleteContext(TestCase):
                 call(
                     "\nWarning: Cannot resolve s3 bucket information from command options"
                     " , local config file or cloudformation template. Please use"
-                    " --s3-bucket and --s3-prefix command options next time and"
+                    " --s3-bucket next time and"
                     " delete s3 files manually if required.",
                     fg="yellow",
                 ),


tests/integration/delete/delete_integ_base.py
```diff
@@ -38,5 +46,9 @@ def get_delete_command_list(
             command_list += ["--profile", str(profile)]
         if no_prompts:
             command_list += ["--no-prompts"]
+        if s3_bucket:
+            command_list += ["--s3-bucket", str(s3_bucket)]
```
mndeveci, line 50: We can set the type information of `s3_bucket` parameter to `Optional[str]` then we don't need to wrap it with `str` call here.

tests/integration/delete/test_delete_command.py
```diff
@@ -45,10 +45,61 @@ def setUpClass(cls):
...
+            s3_bucket=self.bucket_name,
+            s3_prefix=self.s3_prefix,
+            force_upload=True,
+            notification_arns=self.sns_arn,
+            parameter_overrides="Parameter=Clarity",
+            kms_key_id=self.kms_key,
+            no_execute_changeset=False,
+            tags="integ=true clarity=yes foo_bar=baz",
+            confirm_changeset=False,
+            region="us-east-1",
```
mndeveci, line 78: Let's don't use hard coded region information. We had this before and this blocks us running these tests in another region. I would recommend reading region information from environment variables or from boto3 session object. If there are other tests in this file which uses hardcoded region information, updating them would be even better!

samcli/commands/delete/command.py
```diff
@@ -63,12 +64,26 @@
     is_flag=True,
     required=False,
 )
+@click.option(
+    "--s3-bucket",
+    help=("The S3 bucket path you want to delete."),
```
mndeveci, line 69: Let's don't forget to check with docs team before pushing this change.

mndeveci (commented):
LGTM in general! 🥳  Added couple of small comments.

tests/integration/delete/test_delete_command.py
lucashuy: Will also update any other tests that use hardcoded regions.

mndeveci (approved):
LGTM!

samcli/commands/delete/command.py
lucashuy: Checked with docs team, all good to go

samcli/commands/delete/delete_context.py
```diff
@@ -4,8 +4,10 @@
 import logging
 
 import json
-import boto3
 
+from typing import Optional
+
+import boto3
```
qingchm, line 10: Is this an unused import?

samcli/commands/delete/delete_context.py
```diff
@@ -319,12 +334,14 @@ def delete(self):
             self.cf_utils.delete_stack(stack_name=self.stack_name, retain_resources=retain_resources)
             self.cf_utils.wait_for_delete(self.stack_name)
 
-        # If s3_bucket information is not available, warn the user
-        if not self.s3_bucket:
-            LOG.debug("Cannot delete s3 files as no s3_bucket found")
+        # Warn the user that s3 information is missing and to use --s3 options
+        if not self.s3_bucket or not self.s3_prefix:
+            LOG.debug("Cannot delete s3 objects as bucket or prefix is missing")
```
qingchm, line 339: Question mark: is providing s3 bucket without a prefix not a valid use case?

samcli/commands/delete/delete_context.py
lucashuy: The `Optional` class is currently used for the S3 option type as the option could be `None` (ie not specified by the user)

```python
def __init__(
        self,
        stack_name: str,
        region: str,
        profile: str,
        config_file: str,
        config_env: str,
        no_prompts: bool,
        s3_bucket: Optional[str],
        s3_prefix: Optional[str],
):
```

samcli/commands/delete/delete_context.py
lucashuy: Thats a good point! The user could make the choice to not have the prefix be deleted. We could revert the warning message+conditions back or remind the user that `sam delete` did not delete any prefix data.

samcli/commands/delete/delete_context.py
hawflau: even so, the line `import boto3` can be removed, right?

qingchm: @lucashuy Sorry I was referring to import boto3 not Optional!

lucashuy: Oh I see, no worries! boto3 is used in that class to create S3 and ECR clients used to delete data, so its another needed import.

qingchm: Oh I see, this is not a new import you just moved the position, got it!

qingchm (approved):
LGTM! Good job on following up with comments and drafting the PR!

PR #3917, opened 2022-05-27
feat: Added data-safe stack trace collection to telemetry

splouisliu:
N.A.

Previously, only exit code and exit reason are collected when an exception occurs during command executions. In order for us to better understand errors that occur on the user's end, we should also collect stack traces.

Stack trace is now collected when an exception occurs during command executions. The paths contained within are also cleaned to remove user-sensitive information. The final stack trace and exception message are attached to telemetry.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/metric.py
+++ b/samcli/lib/telemetry/metric.py
@@ -133,14 +133,14 @@ def track_command(func):
                 exit_reason = type(ex).__name__
             else:
                 exit_reason = ex.wrapped_from
-            stack_trace = _get_stack_trace()
+            stack_trace = _get_stack_trace(ex)
 
         except Exception as ex:
             exception = ex
             # Standard Unix practice to return exit code 255 on fatal/unhandled exit.
             exit_code = 255
             exit_reason = type(ex).__name__
-            stack_trace = _get_stack_trace()
+            stack_trace = _get_stack_trace(ex)
 
         try:
             ctx = Context.get_current_context()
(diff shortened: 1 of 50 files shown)


hawflau (changes requested):

samcli/lib/telemetry/metric.py
```diff
@@ -170,6 +176,44 @@ def wrapped(*args, **kwargs):
...
+        # If working directory is found within path, remove it
+        dir_index = path.find(working_directory)
+        if dir_index != -1:
+            stack_trace_lines[
+                i
+            ] = f'{prefix}"{path[:dir_index]}{path[dir_index + len(working_directory) + 1 :]}"{suffix}'
+            continue
+
+        # If working directory is not found within path, replace leading path of the last file with: /../ or ..\
+        # i.e. /Users/user1/test.py becomes /../test.py
```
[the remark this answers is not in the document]
splouisliu: The problem is that sometimes the path leads outside the aws-sam-cli folder (i.e. to venv folder). It wouldn't be possible to do this unless we know how the user's outside file directories are structured. AFAIK, Python doesn't have a functionality to return relative path in the stack trace itself.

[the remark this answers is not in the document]
hawflau: I think there's a way to capture more information than keeping only the filename
For example, 
1. Tracing `samcli` - e.g. `/Volumes/workplace/gh/aws-sam-cli/samcli/lib/deploy/deployer.py` -> `/../samcli/lib/deploy/deployer.py`
2. Tracing anything after `site-packages` - e.g. `/usr/local/Cellar/aws-sam-cli/1.49.0/libexec/lib/python3.8/site-packages/botocore/waiter.py` -> `/../botocore/waiter.py`
3. For other cases, it makes sense to just keep the filename as you've implemented.

Handling the cases 1 and 2 should improve the debugging experience a lot.

samcli/lib/telemetry/metric.py
```diff
@@ -170,6 +175,48 @@ def wrapped(*args, **kwargs):
...
+    Returns
+    -------
+    str
+        Stack trace with paths filtered
+    """
+    stack_trace_lines = traceback.format_exc().splitlines()
+
+    for i, line in enumerate(stack_trace_lines):
+        # Make no changes if current line does not contain a path
+        line_split = line.split('"')
```
hawflau, line 191: Will this example break the path detection logic?
```
  File "test.py", line 3, in <module>
    print(a["b"])
```

hawflau (changes requested):
LGTM overall, just want to clarify about one more potential edge case.

tests/unit/lib/telemetry/test_metric.py
```diff
@@ -227,8 +230,11 @@ def real_fn():
             "Measured duration must be in milliseconds and greater than equal to the sleep duration",
         )
 
+    @patch("samcli.lib.telemetry.metric._get_stack_trace")
     @patch("samcli.lib.telemetry.metric.Context")
-    def test_must_record_user_exception(self, ContextMock):
+    def test_must_record_user_exception(self, ContextMock, StackTraceMock):
```
hawflau, line 235: Suggest to match name casing with the method/class being patched (i.e. method -> snake_case, class -> CamelCase)

```suggestion
    def test_must_record_user_exception(self, ContextMock, get_stack_trace_mock):
```

tests/unit/lib/telemetry/test_metric.py
```diff
@@ -244,14 +250,20 @@ def real_fn():
...
+        )
         args, _ = self.telemetry_instance.emit.call_args_list[0]
         metric = args[0]
         assert metric.get_metric_name() == "commandRun"
         self.assertGreaterEqual(metric.get_data().items(), expected_attrs.items())
 
+    @patch("samcli.lib.telemetry.metric._get_stack_trace")
     @patch("samcli.lib.telemetry.metric.Context")
-    def test_must_record_wrapped_user_exception(self, ContextMock):
+    def test_must_record_wrapped_user_exception(self, ContextMock, StackTraceMock):
```
hawflau, line 264:
```suggestion
    def test_must_record_wrapped_user_exception(self, ContextMock, get_stack_trace_mock):
```

hawflau (changes requested):
LGTM overall. Just some minor suggestions on naming. Should be good after fixing them.

moelasmar (approved):

samcli/lib/telemetry/metric.py
```diff
@@ -170,6 +175,58 @@ def wrapped(*args, **kwargs):
...
+    str
+        Stack trace with paths cleaned
+    """
+    tb_exception = traceback.TracebackException.from_exception(exception)
+    _clean_stack_summary_paths(tb_exception.stack)
+    stack_trace = "".join(list(tb_exception.format()))
+    return stack_trace
+
+
+def _clean_stack_summary_paths(stack_summary: traceback.StackSummary):
```
qingchm, line 198: nit: return type -> None

qingchm (approved):

samcli/lib/telemetry/metric.py
```diff
@@ -170,6 +177,62 @@ def wrapped(*args, **kwargs):
...
+
+    Returns
+    -------
+    (str, str)
+        (stack trace, exception message)
+    """
+    tb_exception = traceback.TracebackException.from_exception(exception)
+    _clean_stack_summary_paths(tb_exception.stack)
+    stack_trace = "".join(list(tb_exception.format()))
+    exception_msg = list(tb_exception.format_exception_only())[-1]
```
hawflau, line 199: Are we sure the list always has at least 1 element?

splouisliu: Yup: https://docs.python.org/3/library/traceback.html#traceback.format_exception_only

PR #3919, opened 2022-05-30
Fix deploy with long poll delay events order and stop condition

### LINKED ISSUE #3910: Bug: deploy with large SAM_CLI_POLL_DELAY has a wrong display (nightly)
### Description:
_Not a breaking bug, more like a display and timing bug_

If we use the feature from my last PR #3500, using `SAM_CLI_POLL_DELAY` with a large delay like a minute, the events retrieved during the poll will be displayed in reverse order (because `describe-stack-events` is in reverse chronological order).

Also during the sleep, the deployment can end and another process could start another deploy so the stack will be `IN_PROGRESS` again and when the first `sam deploy` wakes up it misses the end of the first deployment and will keep running and catch the events of the second deployment.

### Steps to reproduce:
Use [`sam-cli-nightly`](https://github.com/aws/aws-sam-cli/releases/tag/sam-cli-nightly)
`SAM_CLI_POLL_DELAY=60 sam deploy ...`

Check on `cloudformation` that the stack completes and while the first CLI is still running (sleeping)
Run `sam deploy ...` again in another tty.

### Observed result:
Wrong order and display 2 deployments
```
-------------------------------------------------------------------------------------------------------------------------
ResourceStatus                 ResourceType                   LogicalResourceId              ResourceStatusReason
-------------------------------------------------------------------------------------------------------------------------
UPDATE_COMPLETE_CLEANUP_IN_P   AWS::CloudFormation::Stack     my-stack                       -
ROGRESS
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     my-stack                       User Initiated
UPDATE_COMPLETE                AWS::CloudFormation::Stack     my-stack                       -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     my-stack                       -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE_CLEANUP_IN_P   AWS::CloudFormation::Stack     my-stack                       -
ROGRESS
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
-------------------------------------------------------------------------------------------------------------------------
```

### Expected result:

```
-------------------------------------------------------------------------------------------------------------------------
ResourceStatus                 ResourceType                   LogicalResourceId              ResourceStatusReason
-------------------------------------------------------------------------------------------------------------------------
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_IN_PROGRESS             AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE_CLEANUP_IN_P   AWS::CloudFormation::Stack     my-stack                       -
ROGRESS
UPDATE_COMPLETE                AWS::CloudFormation::Stack     DynamoApp                      -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     AsyncLambdaApp                 -
UPDATE_COMPLETE                AWS::CloudFormation::Stack     my-stack                       -
-------------------------------------------------------------------------------------------------------------------------
```

`[Unstable] Nightly Release v1.50.0.dev202205260901 - 2022-05-26`

---

### Solution

1: In [`describe_stack_events`](https://github.com/aws/aws-sam-cli/blob/aa75d721b1b463e0a4f2717524cfee80e9fac9d2/samcli/lib/deploy/deployer.py#L364) Instead of immediately print events, they should be buffered during pagination loop and reordered to print.

And in the reordered list find if the stack is not in progress to skip other deployments events (explained in https://github.com/aws/aws-sam-cli/issues/3911).
Maybe we could use `EventId` instead of `Timestamp` to have an exact comparison ?

2: Or maybe we could see with the `cloudformation` team if the [`DescribeStackEvents`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStackEvents.html) could be improved to pass the order and starting `EventId` or `Timestamp` and it does the ordering/filter for us.
(As the second solution might take some time, I'll go on the first one)

---

@jfuss @moelasmar @mndeveci What do you think about this ? I could work on a PR to get this and https://github.com/aws/aws-sam-cli/issues/3911 done quickly

### LINKED ISSUE #3911: Feature request: remove unnecessary describe-stack calls during deploy

Prevent throttling during `sam deploy` by reducing requests.

### Proposal

After my last PR #3500, I saw that during `sam deploy`, the function [`describe_stack_events`](https://github.com/aws/aws-sam-cli/blob/aa75d721b1b463e0a4f2717524cfee80e9fac9d2/samcli/lib/deploy/deployer.py#L364) sends 2 requests for each iteration:

- `DescribeStacks`: to retrieve the status and end the loop if it's not `IN_PROGRESS`
- `DescribeStackEvents`: to display the deployment events

The `DescribeStacks` could be skipped as the status should be accessible via the `DescribeStackEvents`:

```json
{
    "StackId": "arn:aws:cloudformation:region:xxx:stack/my-stack/an-id",
    "EventId": "abc",
    "StackName": "my-stack",
    "LogicalResourceId": "my-stack",
    "PhysicalResourceId": "arn:aws:cloudformation:region:xxx:stack/my-stack/an-id",
    "ResourceType": "AWS::CloudFormation::Stack",
    "Timestamp": "2022-04-27T16:00:33.955000+00:00",
    "ResourceStatus": "UPDATE_COMPLETE"
}
```
So we could check the `ResourceStatus` when `StackId` = `PhysicalResourceId` and `StackName` = `LogicalResourceId`

---

This could be done to fix part of the bug: https://github.com/aws/aws-sam-cli/issues/3910

BourgoisMickael:
Fixes #3910
Fixes #3911

When using the feature from #3500 and deploy with a large `SAM_CLI_POLL_DELAY` events for each poll iterations are in reverse order. And if another process update the stack while sleeping the loop won't stop and catch events from another process.

- Buffer events in a `deque` (push front) to print in chronological order
- Remove `DescribeStacks` request and rely on first `not_in_progress` event on the root stack to exit the loop
- In tests reverse mock events order to match `DescribeStackEvents` api reverse chronological order
- Test print order

- No more requests to `DescribeStack`
- Print refresh delay (`client_sleep`) in table header

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes

### DIFF
--- a/samcli/commands/_utils/table_print.py
+++ b/samcli/commands/_utils/table_print.py
@@ -11,7 +11,9 @@ import click
 MIN_OFFSET = 20
 
 
-def pprint_column_names(format_string, format_kwargs, margin=None, table_header=None, color="yellow"):
+def pprint_column_names(
+    format_string, format_kwargs, margin=None, table_header=None, color="yellow", display_sleep=False
+):
     """
 
     :param format_string: format string to be used that has the strings, minimum width to be replaced
@@ -19,6 +21,7 @@ def pprint_column_names(format_string, format_kwargs, margin=None, table_header=
     :param margin: margin that is to be reduced from column width for columnar text.
     :param table_header: Supplied table header
     :param color: color supplied for table headers and column names.
+    :param display_sleep: flag to format table_header to include deployer's client_sleep
     :return: boilerplate table string
     """
 
@@ -59,7 +62,7 @@ def pprint_column_names(format_string, format_kwargs, margin=None, table_header=
         def wrap(*args, **kwargs):
             # The table is setup with the column names, format_string contains the column names.
             if table_header:
-                click.secho("\n" + table_header)
+                click.secho("\n" + table_header.format(args[0].client_sleep) if display_sleep else table_header)
             click.secho("-" * usable_width, fg=color)
             click.secho(format_string.format(*format_args, **format_kwargs), fg=color)
             click.secho("-" * usable_width, fg=color)
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -53,7 +53,7 @@ DESCRIBE_STACK_EVENTS_DEFAULT_ARGS = OrderedDict(
     }
 )
 
-DESCRIBE_STACK_EVENTS_TABLE_HEADER_NAME = "CloudFormation events from stack operations"
+DESCRIBE_STACK_EVENTS_TABLE_HEADER_NAME = "CloudFormation events from stack operations (refresh every {} seconds)"
 
 DESCRIBE_CHANGESET_FORMAT_STRING = "{Operation:<{0}} {LogicalResourceId:<{1}} {ResourceType:<{2}} {Replacement:<{3}}"
 DESCRIBE_CHANGESET_DEFAULT_ARGS = OrderedDict(
@@ -360,6 +360,7 @@ class Deployer:
         format_string=DESCRIBE_STACK_EVENTS_FORMAT_STRING,
         format_kwargs=DESCRIBE_STACK_EVENTS_DEFAULT_ARGS,
         table_header=DESCRIBE_STACK_EVENTS_TABLE_HEADER_NAME,
+        display_sleep=True,
     )
     def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
         """
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -17,7 +17,7 @@ Cloudformation deploy class which also streams events and changeset information
 
 import sys
 import math
-from collections import OrderedDict
+from collections import OrderedDict, deque
 import logging
 import time
 from datetime import datetime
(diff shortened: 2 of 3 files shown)


jfuss:
@BourgoisMickael Thanks for the quick response and pushing more fixes!

> Remove DescribeStacks request and rely on first not_in_progress event on the root stack to exit the loop

This seems reasonable. I am trying to figure out if there are any edge cases we introduce here. I assume we initially put this in place for "a reason", though not sure what that would be. So I just want to do some digging on this to see if the commit has any indications.

On the ordering:
I am a little confused. From my understanding it seems like the original ordering is consistent before and after your change, unless I am missing something. The reason we did it from top down, was so that we can stream the printing to console and you can read the changes as they come in. I think this change breaks that (from reading the code). Why do you expect the opposite here? What I am really trying to understand is what happened for the ordering to change, and if that was on us or something else changed.

BourgoisMickael:
Because `DesciribeStackEvents` reverse chronological order. So when you poll every 0.5s there will most likely by only one new event for each request and it's printed immediately.

When if a set a delay of 60s, during this time a lot of events will happen, so the `DescribeStackEvents` will return multiple new events with this latests on top, that's wy it needs reordering to print the olddest event first and the latest last

jfuss:
@BourgoisMickael I see. So it all has to do with the polling time. Sorry I did not connect those dots initially. 


I hate diffs, makes changes harder to read sometimes. I see what is happening now. There is a while "stack is in progress" do thing going on. Which allows for the streaming. Or at least keeps the same thing we had before in how we "streamed".

jfuss (approved):
Overall changes look good to me. We appreciate the testing you did on nightly to catch this and provide a patch to address it!

I will ping some team members tomorrow when everyone is in the office to get a second reviewer.

moelasmar (approved):

PR #3922, opened 2022-05-30
feat: remove the created intermediate containers while building the lambda functions of Image Type

moelasmar:
The default behaviour of [docker image create API](https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.build) is to keep the intermediate containers created. So, when customer execute `sam build` for an Image type Lambda Function, the customer will find some created containers exist after `sam build` command finish. This PR is to clean these intermediate containers, and let the docker API delete these containers after the image built process finish.

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -414,7 +414,6 @@ class ApplicationBuilder:
             "decode": True,
             "platform": get_docker_platform(architecture),
             "rm": True,
-            "forcerm": True,
         }
         if docker_build_target:
             build_args["target"] = cast(str, docker_build_target)
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -413,6 +413,8 @@ class ApplicationBuilder:
             "buildargs": docker_build_args,
             "decode": True,
             "platform": get_docker_platform(architecture),
+            "rm": True,
+            "forcerm": True,
         }
         if docker_build_target:
             build_args["target"] = cast(str, docker_build_target)
--- a/tests/integration/buildcmd/test_build_cmd.py
+++ b/tests/integration/buildcmd/test_build_cmd.py
@@ -46,11 +46,7 @@ LOG = logging.getLogger(__name__)
 SKIP_SAR_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
 
 
-@skipIf(
-    # Hits public ECR pull limitation, move it to canary tests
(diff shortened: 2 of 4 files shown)


jfuss:
Is this the right behavior we want? I ask because from https://docs.docker.com/engine/reference/commandline/images/: "Docker images have intermediate layers that increase reusability, decrease disk usage, and speed up docker build by allowing each step to be cached. These intermediate layers are not shown by default." This suggests that (at least currently) some customers may actually want this as the behavior because it saves the layers on the machine to be re-used. And I assume this leads to faster builds, if you build twice.

moelasmar:
my understanding is this change will only remove the intermediate containers, but will keep the images. This is the same behaviour we implement in `sam build --use-container`, we cleaned the created containers after build is done.

By the way, this also the same behaviour of `docker build` command.

samcli/lib/build/app_builder.py
```diff
@@ -413,6 +413,8 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
             "buildargs": docker_build_args,
             "decode": True,
             "platform": get_docker_platform(architecture),
+            "rm": True,
+            "forcerm": True,
```
torresxb1, line 417: I see that here we have `rm` True but not `forcerm`: https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/lambda_image.py#L266
Should we keep them consistent?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -46,11 +46,6 @@
 SKIP_SAR_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
 
 
-@skipIf(
-    # Hits public ECR pull limitation, move it to canary tests
-    ((not RUN_BY_CANARY) or (IS_WINDOWS and RUNNING_ON_CI) and not CI_OVERRIDE),
-    "Skip build tests on windows when running in CI unless overridden",
-)
```
torresxb1, line 53: we don't want to skip anymore?

torresxb1 (commented):
couple of questions

tests/integration/buildcmd/test_build_cmd.py
moelasmar: no, it is my bad .. I removed them to run testing in my local machine, and forget to return them back :(

samcli/lib/build/app_builder.py
moelasmar: actually I was hesitated to add `forcerm`, I will remove it.

jfuss (approved):

torresxb1 (approved):

PR #3929, opened 2022-06-02
Add project metadata tracking with telemetry

Leo10Gama:
N/A

This change allows the observation of sequences of commands run within a project, allowing for the joining of currently-disjoint data sets, providing a clearer picture on how customers use the SAM CLI platform.

The git remote origin URL, project name, and initial commit can all be used to follow a project through the querying of `metricSpecificAttributes`. This metadata can be classified as critical data, and as such, it is encrypted to preserve digital confidentiality.

The `metricSpecificAttributes` key collected by any command with the `@track_command` decorator will now always be present, and the three keys `gitOrigin`, `initialCommit`, and `projectName` will always be present (though the first two may have `None` values).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


hawflau (changes requested):
LGTM! Just one suggestion to refactor.

Also, we might want to review the use uuid5 as a hashing mechanism and get opinions from security. But for now, it's okay. If anything, we can change later.

samcli/lib/telemetry/project_metadata.py
```diff
@@ -0,0 +1,103 @@
...
+from os import getcwd
+from os.path import basename
+import re
+import subprocess
+from uuid import uuid5, NAMESPACE_URL
+
+from samcli.cli.global_config import GlobalConfig
+
+
+def get_git_origin():
```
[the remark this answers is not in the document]
Leo10Gama: 👍

samcli/lib/telemetry/project_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+        current directory that the command is running in.
+        If telemetry is opted out of by the user, returns None
+    """
+    if not bool(GlobalConfig().telemetry_enabled):
+        return None
+
+    project_name = ""
+    try:
+        runcmd = subprocess.run(
+            ["git", "config", "--get", "remote.origin.url"], capture_output=True, shell=True, check=True, text=True
```
mildaniel, line 59: Does this still work if the remote's name isn't "origin"?

Leo10Gama: Unfortunately it would not, and it would return `None` if there is no "origin". However, it would be safe to assume the default remote name is "origin", since that's often the convention used in Git projects.

mildaniel (approved):

Issue #3093, opened 2021-07-21
Deduplicated build is still not fast


yskkin:
### Describe your idea/feature/enhancement
I have ~10 lambdas written in JS.
These lambdas share CodeUri, Runtime and Metadata and thus dedeplicated build is enabled.

`sam build`ing my project is still slow.

This is last output of `sam build --debug`:

```
2021-07-21 12:17:46,950 | Copying artifacts from /path/to/repo/.aws-sam/build/func1 to /path/to/repo/.aws-sam/build/func2
2021-07-21 12:17:56,472 | Copying artifacts from /path/to/repo/.aws-sam/build/func1 to /path/to/repo/.aws-sam/build/func3
```

Copying built lambda took 10 seconds due to large node_modules dependency.
Copying 10 lambdas took 90 seconds.

### Proposal

Symlinking instead?

aahung:
Wonder is it feasible to use symlink in this case. Need to investigate whether `local invoke` and other SAM CLI features can handle symlinks or not.

mildaniel:
The fix has been released in 1.54.0! Resolving. You can now use this functionality using the `experimentalBuildPerformance` feature flag!

yskkin:
Saw significant build performance improvement with `sam build --beta-features`. Thanks ❤️

having-fun-serverless:
@mildaniel why aren't you promoting this beta feature to be released?

Issue #2688, opened 2021-03-05
Cache Use Hard/Soft Link for Instead of Copying All the Files


OGoodness:
### Description
Currently with SAM, you are able to specify a cache, which is nice. But sometimes, using that cache is actually slower than just running a regular `sam build`. SAM copies all the files from the cache, and then writes them in the build directory. This is a lot of I/O operations, and can take forever on old and newer computers, especially if your project runs in node. 

### Proposal
Instead of copying all the files directly, how about you create a symbolic or hard link to each files. This way, there will be less overhead when performing builds, no need to copy the entire file, just create a reference to the associated file. 

Maybe this isn't the default way of doing it, but provide an option to do so. Maybe a parameter you can pass into the `--cached` parameter? 

Things to consider:
1. Speed up builds significantly
2. Saves disk space

### Additional Details
This might introduce issues where, a user decides to manually edit a hard link, and that will effect the original cache. But I think this feature would be a great optional use.

rdt712:
I agree with this. I'm not sure what the solution is, but I've also had a bad experience using a cached build in Node. It takes longer than a normal build and I've never gotten it to work like I would expect at a normal cache to work.

CoshUS:
Thanks for the feedback. For compatibility reasons, symlink might not be the solution as Windows has poor support for it. We will look into other avenues to improve the cached performance.

mildaniel:
Support for this has been released in 1.54.0! You can now use this functionality using the `experimentalBuildPerformance` feature flag!

timmattison:
@mildaniel is there an example of how to use these feature flags? I can find it in the code, and it looks like it can be specified as an environment variable but it would be good to be sure I'm doing it right so I can make a comparison.

PR #3935, opened 2022-06-04
feat: dedup & cached build performance improvements

mndeveci:
#3093
#2688

Previously, SAM CLI improved build performance by introducing de-duplicated & cached builds. These improvements are using copying previously built resources to skip some of the build operations. But runtimes like `nodejs` might contain lots of files inside `node_modules` folder which makes copy operation slow.

In order to address this issue, this PR has following improvements;
- For de-duplicated builds (where multiple functions uses same `CodeUri`) we will be building into a shared folder and point all of the functions to that folder rather than copying first built function into rest of them.
- For cached builds, we will be symlinking previously built artifacts to the target folder, instead of copying them.
- For Auto Dependency Layer (SAM Accelerate feature, where dependencies of a function is moved to a separate layer), we will be symlinking the dependencies to the target layer folder.

Certain operating systems might not support symlinking correctly. Because of that, we are checking if any errors are thrown with symlink operation, if so we will fall-back to copy. 
This feature will also be released behind a feature flag to address some issues before marking it as GA.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/build/build_graph.py
```diff
@@ -625,7 +626,10 @@ def get_build_dir(self, artifact_root_dir: str) -> str:
         Return the directory path relative to root build directory
         """
         self._validate_functions()
-        return self.functions[0].get_build_dir(artifact_root_dir)
+        build_dir = self.functions[0].get_build_dir(artifact_root_dir)
+        if is_experimental_enabled(ExperimentalFlag.BuildImprovements22) and len(self.functions) > 1:
+            build_dir = f"{build_dir}-Shared"
```
sriram-mv, line 631: How is this postfix "Shared" determined?

mildaniel: When determining the artifacts directory that we will have our resultant build in, before we would copy all the contents in the same code URI (technically the same build definition) over and over. Now if there are multiple functions with the same build definition we will just put them into one single shared artifacts directory.

mildaniel:
> I wonder if we can recommend this option/flag based on the runtime...if the project is using node, we can probably add a hint to use this in the CLI.

These changes aren't necessarily runtime-specific except for the accelerate related changes which are only applicable to Python and esbuild.

samcli/lib/build/build_graph.py
sriram-mv: I see, putting up that comment above the line and also making the suffix a constant string thats defined and importable would be good.

sriram-mv (approved):

samcli/lib/build/build_graph.py
```diff
@@ -625,7 +627,12 @@ def get_build_dir(self, artifact_root_dir: str) -> str:
         Return the directory path relative to root build directory
         """
         self._validate_functions()
-        return self.functions[0].get_build_dir(artifact_root_dir)
+        build_dir = self.functions[0].get_build_dir(artifact_root_dir)
+        if is_experimental_enabled(ExperimentalFlag.BuildPerformance) and len(self.functions) > 1:
```
mildaniel, line 631: Add message to customers about this change and add a link to a doc about all the changes that this includes.

qingchm (approved):
Discussion for review session has been captured in the comments above. The main focus would be to release this feature in the next release and then construct a corresponding documentation link explaining on the experimental features. With that customers can better understand the improvements.

PR #3948, opened 2022-06-08
feat: Add in support for nested stack's code sync with Api's and StepFunctions

qingchm:
N.A.

Since currently we are implementing code sync support for nested stack, this will make Api and StepFunction resources within child stacks also syncable through sam sync --code.

It resolves the definition files from path relative to child template, to path relative to root template. For base_dir use cases, this will still be base_dir based.

Not known.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/sync/flows/generic_api_sync_flow.py
+++ b/samcli/lib/sync/flows/generic_api_sync_flow.py
@@ -75,13 +75,13 @@ class GenericApiSyncFlow(SyncFlow):
             return None
         properties = api_resource.get("Properties", {})
         definition_file = properties.get("DefinitionUri")
+        definition_path = None
         if definition_file:
-            if self._build_context.use_base_dir:
-                definition_file = Path(self._build_context.base_dir).joinpath(definition_file)
-            else:
+            definition_path = Path(self._build_context.base_dir).joinpath(definition_file)
+            if not self._build_context.use_base_dir:
                 child_stack = Stack.get_stack_by_full_path(ResourceIdentifier(api_identifier).stack_path, self._stacks)
-                definition_file = Path(child_stack.location).parent.joinpath(definition_file)
-        return cast(Optional[Path], definition_file)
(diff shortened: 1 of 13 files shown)


samcli/lib/sync/flows/generic_api_sync_flow.py
```diff
@@ -65,19 +65,24 @@ def gather_resources(self) -> None:
...
-        if self._build_context.base_dir and definition_file:
-            definition_file = str(Path(self._build_context.base_dir).joinpath(definition_file))
-        return cast(Optional[str], definition_file)
+        definition_path = None
+        if definition_file:
+            definition_path = Path(self._build_context.base_dir).joinpath(definition_file)
+            if not self._build_context.use_base_dir:
+                child_stack = Stack.get_stack_by_full_path(ResourceIdentifier(api_identifier).stack_path, self._stacks)
+                if child_stack:
+                    definition_path = Path(child_stack.location).parent.joinpath(definition_file)
```
mndeveci, line 84: This seems duplicated lines with SFN flow. Can we move this one into a common function so that we can re-use it in sfn sync flow as well?
Can we also add some debug logs if we need to investigate an issue in the future?

mndeveci (commented):
Looks good to me in general, left a small comment there

samcli/lib/sync/sync_flow.py
```diff
@@ -315,3 +316,18 @@ def execute(self) -> List["SyncFlow"]:
             dependencies = self.gather_dependencies()
         LOG.debug("%sFinished", self.log_prefix)
         return dependencies
+
+
+# A helper method used by non-function sync flows to resolve definition file path
+# that are relative to the child stack to absolute path for nested stacks
```
mndeveci, line 322: nit: let's put this as method documentation rather than a comment here.

mndeveci (approved):

mildaniel (approved):

Issue #3946, opened 2022-06-01
Bug: Windows x64 quick start web template error


CoreyCole:
### Description:

Follwing the documentation [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install-windows.html) I ran `sam init` and it threw the error:
```
Error: Can't find application template quick-start-web - check valid values in interactive init.
``` 

### Steps to reproduce:

```
sam init

        SAM CLI now collects telemetry to better understand customer needs.

        You can OPT OUT and disable telemetry collection by setting the
        environment variable SAM_CLI_TELEMETRY=0 in your shell.
        Thanks for your help!

        Learn More: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-telemetry.html


You can preselect a particular runtime or package type when using the `sam init` experience.
Call `sam init --help` to learn more.

Which template source would you like to use?
        1 - AWS Quick Start Templates
        2 - Custom Template Location
Choice: 1

Choose an AWS Quick Start application template
        1 - Hello World Example
        2 - Multi-step workflow
        3 - Serverless API
        4 - Scheduled task
        5 - Standalone function
        6 - Data processing
        7 - Infrastructure event management
        8 - Machine Learning
Template: 3

Which runtime would you like to use?
        1 - dotnet6
        2 - dotnetcore3.1
        3 - nodejs16.x
        4 - nodejs14.x
        5 - nodejs12.x
        6 - python3.9
        7 - python3.8
Runtime: 4

Based on your selections, the only Package type available is Zip.
We will proceed to selecting the Package type as Zip.

Based on your selections, the only dependency manager available is npm.
We will proceed copying the template using npm.

Would you like to enable X-Ray tracing on the function(s) in your application?  [y/N]: y
X-Ray will incur an additional cost. View https://aws.amazon.com/xray/pricing/ for more details

Project name [sam-app]:

Cloning from https://github.com/aws/aws-sam-cli-app-templates (process may take a moment)
2022-05-31 17:59:58,957 |
Cloning from https://github.com/aws/aws-sam-cli-app-templates (process may take a moment)
2022-05-31 18:00:02,025 | Clone error, attempting to use an old clone from a previous run
2022-05-31 18:00:02,025 | Template is not provided in context, skip adding project type metric
[...]

PR #3950, opened 2022-06-08
Fix issue in cloning app templates repo in windows

hawflau:
#3946 

App templates repo contains some very long paths in which Windows doesn't support them as it has a limitation of 260 chars. 

This change modifies the `git clone` command by appending `--config core.longpaths=true` if the system is Windows. The config is only applied for the cloned repo (i.e. the app templates repo) and won't affect user's system settings. 

No

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/git_repo.py
+++ b/samcli/lib/utils/git_repo.py
@@ -129,8 +129,11 @@ class GitRepo:
                 temp_path = os.path.normpath(os.path.join(tempdir, clone_name))
                 git_executable: str = GitRepo._git_executable()
                 LOG.info("\nCloning from %s (process may take a moment)", self.url)
+                command = [git_executable, "clone", self.url, clone_name]
+                if platform.system().lower() == "windows":
+                    command += ["--config", "core.longpaths=true"]
                 check_output(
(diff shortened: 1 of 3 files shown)


samcli/lib/utils/git_repo.py
```diff
@@ -129,8 +129,11 @@ def clone(self, clone_dir: Path, clone_name: str, replace_existing: bool = False
                 temp_path = os.path.normpath(os.path.join(tempdir, clone_name))
                 git_executable: str = GitRepo._git_executable()
                 LOG.info("\nCloning from %s (process may take a moment)", self.url)
+                command = [git_executable, "clone", self.url, clone_name]
+                if platform.system().lower() == "windows":
+                    command += ["--config", "core.longpaths=true"]
```
mndeveci, line 134: Users might need to enable long path settings in windows registry as well. Should we add a debug message to remind that too?

hawflau: yes, we should. Will update it.

moelasmar (approved):

Issue #3945, opened 2022-06-07
Bug: "DepLayer doesn't have any versions in remote"  -  layer versions / SyncFlow issue when using sam sync


Leigh-M:
After upgrading to SAM 1.51.0  -  I am seeing a stack error across all my stacks to do with missing dependency layer versions: 

After initially running it syncs initially.. then you make some changes which you expect to be uploaded to AWS but fails with:
    "Syncing Layer externalJobsLoadingAPIHandler44912949DepLayer..." 
    "DepLayer doesn't have any versions in remote"  

### Steps to reproduce:
`sam sync --watch --stack-name external-jobs-loader --region eu-west-1 --parameter-overrides EnvType=dev --debug`

### Observed result:
```
2022-06-07 16:31:00,657 | SyncFlow [Lambda Function externalJobsLoadingAPIHandler]: Finished
2022-06-07 16:31:00,710 | Finished syncing Lambda Function externalJobsLoadingAPIHandler.
2022-06-07 16:31:00,814 | SyncFlow [Layer externalJobsLoadingAPIHandler44912949DepLayer]: Setting Up
2022-06-07 16:31:00,820 | Syncing Layer externalJobsLoadingAPIHandler44912949DepLayer...
2022-06-07 16:31:01,121 | Code sync encountered an error.
Traceback (most recent call last):
  File "samcli/lib/sync/sync_flow_executor.py", line 335, in _sync_flow_execute_wrapper
  File "samcli/lib/sync/sync_flow.py", line 307, in execute
  File "samcli/lib/sync/flows/auto_dependency_layer_sync_flow.py", line 72, in set_up
samcli.lib.sync.exceptions.NoLayerVersionsFoundError: external-jobs-lodcdd4f9b-externalJobsLoadingAPI44912949-DepLayer doesn't have any versions in remote.
```

I did also see a new yellow warning:  "CodeTrigger not created as CodeUri or DefinitionUri is missing for externalJobsLoaderApi" - so there appears to be requirement for DefinitionUri  applied to an API provisioned in my template.yaml? 

But this dependency/ S3 URI should be controlled by SAM I understand.. ?

1. OS:   Ubuntu Linux 22.04
2. If using SAM CLI, `sam --version`:  1.51.0
3. AWS region:  eu-west-1

mndeveci:
Thanks for reporting this issue @Leigh-M 

Can you provide us an example template that we can re-produce the issue?

[...]

PR #3952, opened 2022-06-09
fix: revert not creating ADL for empty dependencies

mndeveci:
https://github.com/aws/aws-sam-cli/issues/3945

We recently introduced a fix; sam sync will not create dependency layers for functions which doesn't any dependency defined in them. This feature works during initial infra sync but it fails subsequent sync flow executions.

Revert that latest change for now and revert back to original behavior.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/sync/test_sync_adl.py
```diff
@@ -0,0 +1,168 @@
...
+        self.update_file(
+            self.test_dir.joinpath("code", "after", "python_function_no_deps", "requirements.txt"),
+            self.test_dir.joinpath("code", "before", "python_function_no_deps", "requirements.txt"),
+        )
+        read_until_string(
+            self.watch_process,
+            "\x1b[32mFinished syncing Function Layer Reference Sync HelloWorldFunction.\x1b[0m\n",
+            timeout=60
+        )
+        time.sleep(LAMBDA_SLEEP)
```
mndeveci, line 164: Add retry rather than fixed sleep here

tests/integration/sync/test_sync_watch.py
```diff
@@ -29,7 +29,7 @@
     RUN_BY_CANARY,
     kill_process,
     read_until_string,
-    start_persistent_process,
+    start_persistent_process, run_command,
```
mndeveci, line 32: redundant import

tests/integration/sync/test_sync_watch.py
```diff
@@ -91,6 +91,10 @@ def tearDown(self):
                 cfn_client.delete_stack(StackName=stack_name)
         super().tearDown()
 
+    def should_run_initial_infra_validation(self):
```
mndeveci, line 94: Change this method to validate initial infra and override in other classes if needed to

qingchm (approved):

PR #3955, opened 2022-06-09
Implement event data structure and relevant unit tests

Leo10Gama:
N/A

This change adds Events to the codebase, which will be used alongside the Telemetry class to keep record of which specific events occur during the runtime of a given command.

With the new data structure, Events and their available options can be created and used to represent the settings used during command runs.

Since the Event class and its test method are currently disjoint from the rest of the codebase, there are no immediate side effects to this addition.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -4,7 +4,7 @@ Represents Events and their values.
 
 from abc import ABC
 from enum import Enum
-from typing import Any, List
+from typing import List
 
 
 class EventName(Enum):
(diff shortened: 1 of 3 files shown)


samcli/lib/telemetry/event.py
```diff
@@ -0,0 +1,64 @@
...
+    """Class to represent Events that occur in SAM CLI."""
+
+    event_name: EventName
+    event_value: str  # Validated by EventType.get_accepted_values to never be an arbitrary string
+
+    def __init__(self, event_name: str, event_value: str):
+        Event._verify_event_name(event_name)
+        self.event_name = EventName(event_name)
+        if event_value not in EventType.get_accepted_values(self.event_name):
+            raise KeyError(f"Event '{self.event_name.value}' does not accept value '{event_value}'.")
```
hawflau, line 54: Can you create a new exception class for this instead of using native `KeyError`? That would make debugging easier. Also, consider where this Event instance will be created, and thus where the potential places that we need to handle this exception.

samcli/lib/telemetry/event.py
```diff
@@ -0,0 +1,64 @@
...
+        self.event_value = event_value
+
+    def __eq__(self, other):
+        return self.event_name == other.event_name and self.event_value == other.event_value
+
+    @staticmethod
+    def _verify_event_name(event_name: Any) -> None:
+        """Raise a NameError if the passed parameter is not an EventName Enum."""
+        if event_name not in [event.value for event in EventName]:
+            raise NameError(f"Event '{event_name}' does not exist.")
```
hawflau, line 64: same as above, try to create a new exception class instead of using native ones.

hawflau (changes requested):

samcli/lib/telemetry/event.py
Leo10Gama: Very true! Will implement

samcli/lib/telemetry/event.py
Leo10Gama: 👍

samcli/lib/telemetry/event.py
```diff
@@ -0,0 +1,64 @@
...
+        self.event_name = EventName(event_name)
+        if event_value not in EventType.get_accepted_values(self.event_name):
+            raise KeyError(f"Event '{self.event_name.value}' does not accept value '{event_value}'.")
+        self.event_value = event_value
+
+    def __eq__(self, other):
+        return self.event_name == other.event_name and self.event_value == other.event_value
+
+    @staticmethod
+    def _verify_event_name(event_name: Any) -> None:
```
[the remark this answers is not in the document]
Leo10Gama: It can be, yes! Will fix this in the following commit

samcli/lib/telemetry/event.py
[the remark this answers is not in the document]
Leo10Gama: It would likely be consistent to have its validation in a separate method, yes. Merging with the other verify will keep the "verification" logic in one central function, so I will implement that in the next commit

samcli/lib/telemetry/event.py
```diff
@@ -0,0 +1,73 @@
...
+
+class EventName(Enum):
+    """Enum for the names of available events to track."""
+
+    USED_FEATURE = "UsedFeature"
+    DEPLOY = "Deploy"
+    BUILD_RUNTIME = "BuildRuntime"
+
+
+class EventType(ABC):
```
hawflau, line 18: As discussed previously, we probably don't need this class to be an ABC?

Leo10Gama: Right, almost forgot about that 😅 will remove and fix for the next commit

hawflau (approved):

torresxb1 (approved):

PR #3966, opened 2022-06-13
chore: replace InvokeContextException with granular ones

mndeveci:
`InvokeContextException` for `sam local` command is used for multiple different error cases. With this PR we are replacing each individual case with specific exception type.

To get more granular information about `InvokeContextException`'s.

By replacing `InvokeContextException` with specific ones.

N/A, some notes;
- `InvokeContextException` or parent class `UserException` have not caught anywhere in the codebase.
- The newly created exceptions are also extending from `UserException` which should cause SAM CLI to fail gracefully without logging the entire trace stack.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/user_exceptions.py
+++ b/samcli/commands/local/cli_common/user_exceptions.py
@@ -5,6 +5,12 @@ Class containing error conditions that are exposed to the user.
 from samcli.commands.exceptions import UserException
 
 
+class InvokeContextException(UserException):
+    """
+    Something went wrong invoking the function.
+    """
+
+
 class InvalidSamTemplateException(UserException):
     """
     The template provided was invalid and not able to transform into a Standard CloudFormation Template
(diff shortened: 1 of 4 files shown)


samcli/commands/local/cli_common/user_exceptions.py
```diff
@@ -5,12 +5,6 @@
 from samcli.commands.exceptions import UserException
 
 
-class InvokeContextException(UserException):
```
jfuss, line 8: Why replace instead of extending further? Wouldn't that make less changes overall and we can still `catch` `InvokeContextException` as we did before

mndeveci: That was my original idea, but then I found these;
- `InvokeContextException` is not caught in any place, same goes with parent classes `UserException` and `ClickException`. 
- `InvokeContextException` is not used in any place that would make it convenient to leave there. After replacing the newer exceptions, no usage is left so I removed the entire class.

jfuss: But the hierarchy of exceptions can be no? For example, the main CLI would catch `InvokeContextException` and do something specific with it like communicating further help. If we flatten everything, we loose that. `InvokeContextException` is a user error Exception.

qingchm (approved):

PR #3969, opened 2022-06-14
feat: Adding --s3-bucket and --s3-prefix to sam delete

lucashuy:
#3246 

Adding S3 bucket options will provide the ability to delete S3 buckets and objects if the bucket and prefix cannot be obtained from the local config file or CloudFormation template.

This change adds the `--s3-bucket` and `--s3-prefix` command line options to `sam delete`.

This change adds additional parameters to sam delete related components.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


tests/integration/delete/delete_integ_base.py
```diff
@@ -22,10 +23,19 @@ def base_command(self):
         return command
 
     def get_delete_command_list(
```
lucashuy, line 25: This PR is identical to #3915 except here I've reverted the changes to the get delete command method to what it was initially to resolve integration tests failing in certain conditions on Windows machines.

torresxb1 (approved):

PR #3970, opened 2022-06-14
Implement tracker methods

Leo10Gama:
N/A

This implementation of tracking events will allow for any arbitrary event to be tracked and stored in a class variable that can be accessed across multiple files. It is necessary to allow for simplified collection of events that may occur during a SAM CLI session.

An `EventTracker` class has been added to encapsulate the logic of tracking events. It holds a private variable to store events, which are accumulated by scattering `EventTracker.track_event()` method calls throughout other code paths. When called, it appends those Events (if they are valid Events) to the private variable, which can later be accessed to send through Telemetry before the end of the session. The list of tracked events can also be cleared at the end of the session to prevent inaccuracies.

While this new class uses the existing `Event` class in its methods, it exists disjoint to existing code, thus there are no notable side effects to this change.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -34,7 +34,7 @@ class EventType:
 
     @staticmethod
     def get_accepted_values(event_name: EventName) -> List[str]:
-        """Get all acceptable values for a given Event."""
+        """Get all acceptable values for a given Event name."""
         if event_name not in EventType._events:
             return []
         return EventType._events[event_name]
@@ -54,6 +54,9 @@ class Event:
     def __eq__(self, other):
         return self.event_name == other.event_name and self.event_value == other.event_value
 
+    def __repr__(self):
+        return f"({self.event_name.value}, {self.event_value})"
+
     @staticmethod
     def _verify_event(event_name: str, event_value: str) -> None:
         """Raise an EventCreationError if either the event name or value is not valid."""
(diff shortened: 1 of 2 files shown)


samcli/lib/telemetry/event.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, event_name: str, event_value: str):
     def __eq__(self, other):
         return self.event_name == other.event_name and self.event_value == other.event_value
 
+    def __repr__(self):
```
mingkun2020, line 57: nit: `__repr__` method usually would provide some developer info, like here return "EventType(event_name, event_value)" maybe better. Not sure how we use this method in our cli.

mingkun2020 (approved):

samcli/lib/telemetry/event.py
Leo10Gama: I mostly implemented this to make debugging a little easier, since otherwise it prints something along the lines of `<Event object at 0x12345678>`. Labelling it more clearly would likely give further clarification on the object itself; I'll implement it in a following commit

hawflau: That's a good call out.
Another note - when we embed the events into the metric, we will need to convert the `Event` instances into a json-compatible format

Leo10Gama: Implementing a method to quickly convert the class to something JSON-compatible shouldn't be too tricky. I can implement that as well to append to this PR

hawflau (approved):

PR #3973, opened 2022-06-14
chore: remove beta experimental flag

mingkun2020:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/logs/command.py
```diff
@@ -163,9 +150,6 @@ def do_cli(
     from samcli.lib.utils.boto_utils import get_boto_client_provider_with_config, get_boto_resource_provider_with_config
 
     if not names or len(names) > 1:
```
mndeveci, line 152: I think this if statement needs some adjustments right? It was either asking for beta feature opt in or printing an informational message.

samcli/commands/sync/command.py
```diff
@@ -77,20 +70,8 @@
...
-This feature is currently in beta. Visit the docs page to learn more about the AWS Beta terms https://aws.amazon.com/service-terms/.
-
-The SAM CLI will use the AWS Lambda, Amazon API Gateway, and AWS StepFunctions APIs to upload your code without 
-performing a CloudFormation deployment. This will cause drift in your CloudFormation stack. 
-**The sync command should only be used against a development stack**.
-
-Confirm that you are synchronizing a development stack and want to turn on beta features.
-
-Enter Y to proceed with the command, or enter N to cancel:
-"""
```
mndeveci, line 90: We still want this confirmation in the beginning of `sam sync`. So we only need to remove first line here, rest should stay there and we should get a confirmation from customer every time they run `sam sync`.

mndeveci (changes requested):

samcli/commands/sync/command.py
mingkun2020: We already have one name `SYNC_CONFIRMATION_TEXT` above, just remove this beta one.

mndeveci: Got it, I thought we had one text and we display it all the time. Thanks!

samcli/commands/sync/command.py
```diff
@@ -244,15 +224,9 @@ def do_cli(
 
     confirmation_text = SYNC_CONFIRMATION_TEXT
```
mndeveci, line 225: nit: then we can remove this line here by directly using `SYNC_CONFIRMATION_TEXT` below

samcli/commands/logs/command.py
mingkun2020: This if statement check if the name is not provided we will print the hint. In the previous beta version we check if the beta flag is off then we don't print the hint then exist but now we don't need the beta check anymore.

mndeveci: But there was an `else` statement there which was checking;
- If no names or multiple names provided, ask for beta opt-in
- Else print informational message that `sam logs` can be used without any name flag.

So now it seems like it is printing the information message if no names or multiple names are provided. It should be opposite.

mingkun2020: I see. Thanks for the explanation. Updated the logic.

Issue #2971, opened 2021-06-21
sam deploy: support OnFailure


pjungermann:
### Describe your idea/feature/enhancement

At aws-cli, you can use the optional `--on-failure` argument at e.g., the `aws cloudformation create-stack` command. The default behavior (when not set/`None`) is `ROLLBACK`.

In the case of a rollback, stack will end up in state ROLLBACK_COMPLETE which requires to delete the stack first in order to attempt another deployment.

Adding the option, will allow users to customize the behavior and e.g., choose `DELETE` to directly delete the stack to not require a (manual) delete.

Additionally, aws-cli supports `[--disable-rollback | --no-disable-rollback]. 
You can specify either `OnFailure` or `DisableRollback`, but not both. 

### Proposal

Add the optional argument `--on-failure DO_NOTHING|ROLLBACK|DELETE` to the `sam deploy` command which then results in adding this config at `Deployer.create_changeset` (`kwargs`).

jfuss:
@pjungermann Thanks for the feature request. Seems like a good addition to `sam deploy` command. 

Updating labels.

mildaniel:
Support for this has been released in 1.54.0! Resolving.

PR #3974, opened 2022-06-14
feat: Adding --on-failure to sam deploy

lucashuy:
#2971

This change primarily provides the ability to delete a stack upon a failed deployment, reducing the amount of manual work required to remove failed stacks.

This change adds `--on-failure` to `sam deploy`. The possible options are:
 - `DELETE`: This will attempt to rollback the stack to a successful state, otherwise it will delete the stack
 - `ROLLBACK`: This is the default behaviour, it will only try to rollback the stack
 - `DO_NOTHING`: Upon failure, `sam deploy` will not attempt to rollback

This changes some internal methods by adding support for `on_failure` arguments. Additionally, `--on-delete` and `--disable-rollback/--no-disable-rollback` are now mutually exclusive.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/deploy/deployer.py
```diff
@@ -481,15 +499,20 @@ def wait_for_execute(self, stack_name: str, stack_operation: str, disable_rollba
...
-        outputs = self.get_stack_outputs(stack_name=stack_name, echo=False)
-        if outputs:
-            self._display_stack_outputs(outputs)
+        try:
+            outputs = self.get_stack_outputs(stack_name=stack_name, echo=False)
+            if outputs:
+                self._display_stack_outputs(outputs)
+        except DeployStackOutPutFailedError as ex:
+            # Show exception if we aren't deleting stacks
+            if not stack_operation == "DELETE":
```
lucashuy, line 514: Change this to check the FailureMode enum instead of checking for "DELETE" since that is not being used.

samcli/lib/deploy/deployer.py
```diff
@@ -424,6 +431,9 @@ def describe_stack_events(self, stack_name, time_stamp_marker, **kwargs):
                 # Reset retry attempts if iteration is a success to use client_sleep again
                 retry_attempts = 0
             except botocore.exceptions.ClientError as ex:
+                if "Stack with id {0} does not exist".format(stack_name) in str(ex) or on_failure == FailureMode.DELETE:
```
qingchm, line 434: let's change the or to and as discussed in code review session, as we don't want to accidentally catch the "Stack with id {0} does not exist" exception when the on failure is not specified
```suggestion
                if "Stack with id {0} does not exist".format(stack_name) in str(ex) and on_failure == FailureMode.DELETE:
```

samcli/commands/deploy/deploy_context.py
```diff
@@ -264,14 +267,25 @@ def deploy(
                     if not click.confirm(f"{self.MSG_CONFIRM_CHANGESET}", default=False):
                         return
 
-                self.deployer.execute_changeset(result["Id"], stack_name, disable_rollback)
-                self.deployer.wait_for_execute(stack_name, changeset_type, disable_rollback)
+                # Stop the rollback in the case of DO_NOTHING or if this is a new stack and DELETE is specified
+                do_disable_rollback = (
```
qingchm, line 271: Can we have a comment here explaining that DO_NOTHING is exactly the same thing as disable rollback? Quoting from CFN documentation:

> To preserve successfully provisioned resources during a stack create operation
Specify the disable-rollback option or on-failure DO_NOTHING enumeration during a [create-stack](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html) operation.

samcli/lib/deploy/deployer.py
```diff
@@ -642,6 +675,104 @@ def get_stack_outputs(self, stack_name, echo=True):
...
+                self._rollback_wait(stack_name)
+
+                current_state = self._get_stack_status(stack_name)
+
+            failed_states = ["CREATE_FAILED", "UPDATE_FAILED", "ROLLBACK_COMPLETE", "ROLLBACK_FAILED"]
+
+            if current_state in failed_states:
+                LOG.info("Stack %s failed to create/update correctly, deleting stack", stack_name)
+
+                self._client.delete_stack(**kwargs)
```
lucashuy, line 708: Add a case to check for DELETE_FAILED, where we were not able to delete some resources (eg S3 bucket) and let them know that they need to clean it up.

samcli/commands/deploy/command.py
```diff
@@ -107,6 +108,23 @@
...
+@click.option(
+    "--on-failure",
+    default="ROLLBACK",
+    type=click.Choice(["ROLLBACK", "DELETE", "DO_NOTHING"]),
+    required=False,
+    help="""
+    Provide an action to determine what will happen when a stack fails to create. Three actions are available:\n
+    - ROLLBACK: This will rollback a stack to a previous known good state.\n
+    - DELETE: The stack will rollback to a previous state if one exists, otherwise the stack will be deleted.\n
+    - DO_NOTHING: The stack will not rollback or delete.\n
```
qingchm, line 123: Can we mention the incompatibility with disable-rollback flag in the help text? Also can we have a comment here explaining that DO_NOTHING is exactly the same thing as disable rollback? Quoting from CFN documentation:

> To preserve successfully provisioned resources during a stack create operation
Specify the disable-rollback option or on-failure DO_NOTHING enumeration during a [create-stack](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html) operation.

samcli/lib/deploy/deployer.py
```diff
@@ -481,15 +499,20 @@ def wait_for_execute(self, stack_name: str, stack_operation: str, disable_rollba
             waiter.wait(StackName=stack_name, WaiterConfig=waiter_config)
         except botocore.exceptions.WaiterError as ex:
             LOG.debug("Execute stack waiter exception", exc_info=ex)
-            if disable_rollback:
+            if disable_rollback and on_failure is not FailureMode.DELETE:
```
lucashuy, line 502: Make sure to clarify that the only valid case here is when the failure mode is DO_NOTHING or disable_rollback is set to True

mingkun2020 (approved):

mildaniel (approved):

samcli/lib/deploy/deployer.py
```diff
@@ -511,7 +515,7 @@ def wait_for_execute(
                 self._display_stack_outputs(outputs)
         except DeployStackOutPutFailedError as ex:
             # Show exception if we aren't deleting stacks
-            if not stack_operation == "DELETE":
+            if on_failure != FailureMode.DELETE:
```
qingchm, line 518: Can you briefly explain why we don't need to "or" this condition in for stack_operation == "DELETE" anymore?

lucashuy: Yup, so previously I had added a waiter for stack_delete_complete, this meant that I had to add DELETE as a possible stack_operation. Now its been removed and the waiter is now located in the rollback_delete_stack() method, so we check the on_failure parameter instead.

qingchm (approved):
LGTM!

PR #3976, opened 2022-06-17
feat: add terraform hook package skeleton

torresxb1:
Adds basic skeleton terraform hook package with Config file

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [x] Write documentation


samcli/hook_packages/terraform/prepare/main.bat
```diff
@@ -0,0 +1,14 @@
+:: ensure the script works regardless of where it's invoked from
+@echo off&setlocal
+for %%i in ("%~dp0.") do set "parent_path=%%~fi"
+cd %parent_path%
+
+:: set up virtual environment
+python -m venv ./venv
```
moelasmar, line 7: - do we need to check if the virtual environment is already there?
- do we need to check the python is installed and return errors if it is not installed?
- we need to make sure that these commands are not logging to the output stream, we need to forward it to some logs file. The idea is the wrapper will read the prepare hook output from the command output stream.

samcli/hook_packages/terraform/Config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "hook_package_id": "terraform",
+    "description": "Terraform IaC hooks package",
+    "hook_use_case": "IaC",
+    "iac_framework": "Terraform",
+    "version": "1.0",
+    "hook_specification": "1.0.0",
+    "functionalities": {
+         "prepare": {
+              "entry_script": "./prepare/main"
```
hawflau, line 10: probably question to @moelasmar, do we need `parameters` here?

torresxb1: I believe we're not supporting parameters for now

samcli/hook_packages/terraform/prepare/main.bat
hawflau: I also have some questions:
- If we use `./venv` here, are we assuming the script runs in the current project root? 
- What if the customer already has a virtual env of the same name created for a different purpose? 
- if we are re-using an existing virtual env and there's a version conflict in one of the dependencies, how shall we handle it?

samcli/hook_packages/terraform/prepare/main.bat
```diff
@@ -0,0 +1,29 @@
...
+    :: check that python 3 is installed
+    py -3 --version >NUL 2>&1
+    if errorlevel 1 goto errorNoPython
+
+    :: set up virtual environment
+    py -3 -m venv ./venv
+    CALL venv/Scripts/activate
+
+    :: install dependencies
+    pip install -r requirements.txt
```
hawflau, line 17: this seems to assume the cwd of running this script is `/path/to/samcli/hook_packages/terraform/prepare/`. So back to the question about the location of the virtual env, are we creating the virtual env in the same cwd?

samcli/hook_packages/terraform/prepare/main.bat
torresxb1: - nope, the script runs in the `prepare` directory always (check out the `ensure the script works regardless of where it's invoked from` line)
- not a problem due to above ^

samcli/hook_packages/terraform/prepare/main.bat
torresxb1: addressed above

samcli/hook_packages/terraform/prepare/main.bat
hawflau: do we want to create a virtual env in the `prepare` directory? Since it will live under where the SAM CLI is installed. I supposed this should be touched only by Homebrew or any other package manager like pip.

For example, homebrew installed my SAM CLI at `/usr/local/Cellar/aws-sam-cli/1.52.0/libexec/lib/python3.8/site-packages`. So if we create the venv in the `prepare` directory, it will be `/usr/local/Cellar/aws-sam-cli/1.52.0/libexec/lib/python3.8/site-packages/samcli/hook_packages/terraform/prepare/venv`. I'm not sure if we should touch anything under `/usr/local/Cellar/aws-sam-cli/1.52.0/libexec/lib/python3.8/site-packages`

samcli/hook_packages/terraform/Config.json
```diff
@@ -0,0 +1,16 @@
...
+    "hook_package_id": "terraform",
+    "description": "Terraform IaC hooks package",
+    "hook_use_case": "IaC",
+    "iac_framework": "Terraform",
+    "version": "1.0",
+    "hook_specification": "1.0.0",
+    "functionalities": {
+         "prepare": {
+              "entry_method": {
+                    "module": "samcli.hook_packages.terraform.prepare.hook",
```
torresxb1, line 11: do we still want to store the hook somwhere like this, or in another location?

samcli/hook_packages/terraform/Config.json
```diff
@@ -0,0 +1,16 @@
...
+                    "method": "hook"
```
torresxb1, line 12: wasn't very creative here lol, open to suggestions

moelasmar: method name can be `prepare`, `deploy` .. etc

moelasmar (changes requested):
I was expecting to define a method on the python file you created that will accept a dict, and return a dict. In this method, we will initiate the TF_Hook class and call the prepare method

hawflau (changes requested):
probably better to have the `prepare` method defined in hook.py (or `samcli/hook_packages/terraform/main.py` as @moelasmar suggested). It's okay to just define the method which does nothing but only returns the expected output.

hawflau (approved):

moelasmar (approved):

PR #3977, opened 2022-06-17
feat: Adding --s3-bucket to sam sync

lucashuy:
#3537

This change adds the ability to specify a custom S3 bucket when using `sam sync`, overriding the default managed bucket.

This change adds `--s3-bucket` to `sam sync`.

This change will modify the base `do_cli` method to include `s3_bucket`.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


qingchm (changes requested):
Since there is an s3_bucket_option already created in samcli.commands._utils.options and just using @s3_bucket_option(guided=False) here can help with what we want to input, is it possible to refactor that option to not use the callback for sync like @s3_bucket_option(guided=False, sync=True) or something that disables the callback for sync? It feels like this option will be a duplicate to me.

lucashuy:
Yup, it should be possible to refactor the existing util option to disable a callback if the sync argument is provided without breaking anything else. Looking into it

samcli/commands/_utils/options.py
```diff
@@ -474,19 +474,22 @@ def stack_name_option(f, required=False, callback=None):
...
-    )
+def s3_bucket_click_option(guided, callback):
+    kwargs = {
+        "required": False,
+        "help": "The name of the S3 bucket where this command uploads the artifacts "
+        "that are referenced in your template.",
+    }
+
+    if callback:
+        kwargs["callback"] = None if guided else partial(artifact_callback, artifact=ZIP)
```
mndeveci, line 485: Can we move & combine all if checks (`callback` & `guided`) into one place?

mndeveci (changes requested):

samcli/commands/_utils/options.py
```diff
@@ -475,13 +475,16 @@ def stack_name_option(f, required=False, callback=None):
...
-        help="The name of the S3 bucket where this command uploads the artifacts that are referenced in your template.",
-    )
+    kwargs = {
+        "required": False,
+        "help": "The name of the S3 bucket where this command uploads the artifacts "
+        "that are referenced in your template.",
+    }
+
+    if guided:
+        kwargs["callback"] = partial(artifact_callback, artifact=ZIP)
```
mndeveci, line 485: This if statement was other way around (`callback` was `None` if `guided` is provided). Can we check if this is the correct behaviour?

mndeveci (changes requested):

samcli/commands/_utils/options.py
lucashuy: Oh yep, nice catch. I accidentally forgot about the entire other half of that statement.

tests/integration/sync/sync_integ_base.py
```diff
@@ -262,6 +263,8 @@ def get_sync_command_list(
             command_list += ["--image-repository", str(image_repository)]
         if image_repositories:
             command_list += ["--image-repositories", str(image_repositories)]
+        if s3_bucket:
+            command_list += ["--s3-bucket", str(s3_bucket)]
```
qingchm, line 267: Since we added support for s3 bucket for integration test, we can also write a simple one with s3_bucket option using the package integration test's testing bucket.

qingchm (approved):
LGTM! Thanks for addressing the comments

qingchm (approved):

mndeveci (approved):

PR #3978, opened 2022-06-17
chore: fix canary issues

mndeveci:
- Update regex to match the new build outputs
- Remove specific layer names from nested stack tests, which leads to random failures (same layer name should not be used in nested stacks)
- Add missing class fields for ADL tests
- Add missing method parameter for `sam logs` tests

### DIFF
--- a/tests/integration/testdata/sync/code/before/template-python-no-dependencies.yaml
+++ b/tests/integration/testdata/sync/code/before/template-python-no-dependencies.yaml
@@ -11,4 +11,4 @@ Resources:
     Properties:
       CodeUri: python_function_no_deps/
       Handler: app.lambda_handler
-      Runtime: python3.9
+      Runtime: python3.7
--- a/tests/integration/testdata/sync/infra/template-python-after.yaml
+++ b/tests/integration/testdata/sync/infra/template-python-after.yaml
@@ -31,7 +31,6 @@ Resources:
   HelloWorldLayer:
     Type: AWS::Serverless::LayerVersion
     Properties:
-      LayerName: HelloWorldLayer
       Description: Hello World Layer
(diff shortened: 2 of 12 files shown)


mildaniel (approved):

tests/integration/testdata/sync/infra/template-python-after.yaml
```diff
@@ -31,7 +31,6 @@ Resources:
   HelloWorldLayer:
     Type: AWS::Serverless::LayerVersion
     Properties:
-      LayerName: HelloWorldLayer
```
mildaniel, line 34: Why are we removing the LayerName properties?

tests/integration/testdata/sync/code/before/template-python-no-dependencies.yaml
```diff
@@ -11,4 +11,4 @@ Resources:
     Properties:
       CodeUri: python_function_no_deps/
       Handler: app.lambda_handler
-      Runtime: python3.9
```
mildaniel, line 14: Why does 3.9 not work?

mndeveci: Windows canaries doesn't have 3.9 configured with it, so we need to use 3.7 until we update configuration there.

tests/integration/testdata/sync/infra/template-python-after.yaml
mndeveci: If you use same `LayerName` in different templates in nested stack, they will have same ARN therefore their sync flow will start to clash each other. If we remove its name, it will have a randomly generated ARN for every layer.

torresxb1 (approved):

PR #3979, opened 2022-06-17
Send Event metrics via Telemetry

Leo10Gama:
N/A

With this change, previously implemented Event tracker methods will be able to have their information sent via Telemetry. This provides us with a clear and concise way to actually look at the Event data.

The previously added `EventTracker` class is imported into the Metric class, allowing access to the Events that are added to it by other methods. Since the decorator function is run after the primary SAM commands are run, it guarantees that any Event information is collected after all commands are executed.

With this new metric in place, the `Metric` class' `metricSpecificAttributes` field will have one additional value, "events". This will be a list, containing JSON objects with the following format:
```
{
  "event_name":  "SomeName",
  "event_value": "SomeValue"
}
```

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/telemetry/metric.py
```diff
@@ -158,6 +159,9 @@ def wrapped(*args, **kwargs):
             metric_specific_attributes["gitOrigin"] = get_git_remote_origin_url()
             metric_specific_attributes["projectName"] = get_project_name()
             metric_specific_attributes["initialCommit"] = get_initial_commit_hash()
+            # Event metrics
+            metric_specific_attributes["events"] = [e.to_json() for e in EventTracker.get_tracked_events()]
```
hawflau, line 163: Can we add some tests to validate:
1/ if some events are tracked, `metric_specific_attributes["events"]` contains the expected events; and
2/ if no event is tracked, `metric_specific_attributes["events"]` is an empty array

samcli/lib/telemetry/metric.py
```diff
@@ -158,6 +159,9 @@ def wrapped(*args, **kwargs):
...
+            EventTracker.clear_trackers()
```
hawflau, line 164: would be good to make sure `clear_trackers` is called in the `track_command` test

hawflau (changes requested):

samcli/lib/telemetry/metric.py
Leo10Gama: Definitely! I wasn't sure whether tests such as those would be necessary since the attribute existence is covered by the existing metric test, but I'll implement them to be more specific and detailed to the events

samcli/lib/telemetry/metric.py
Leo10Gama: I'm not sure I understand. Since `clear_trackers` is called in `track_command`, any existing tests would call it by default, as all tests that call `track_command` will also call `clear_trackers`. Could you let me know which types of tests should be calling `clear_trackers`?

hawflau: My understanding is - `clear_trackers` must be called in `track_command`. 
If someone accidentally removes this line, we won't be able to catch it.

Leo10Gama: With the additional `Event` unit tests I added in the latest commit (1b5efd5), if line 164 were to be accidentally deleted and things not function properly, one or more of the three test cases will fail, as previous Events will be carried over into the new tests (I tested on my machine with that line commented out, and that is the result I get). Another solution could be to leave a comment on that line, or update the `clear_trackers` docstring, as to provide more thorough reasoning for why the line is necessary where it is?

tests/unit/lib/telemetry/test_metric.py
```diff
@@ -339,6 +340,46 @@ def real_fn(a, b=None):
...
+                ],
+            ),
+            (
+                [
+                    Mock(event_name=Mock(value="T1"), event_value="1"),
+                    Mock(event_name=Mock(value="T2"), event_value="2"),
+                    Mock(event_name=Mock(value="T3"), event_value="3"),
+                ],
+            ),
+            ([],),
```
qingchm, line 360: Is this parameter list specifically designed this way? If so what's the rationale behind this order?

Leo10Gama: Yes, I specifically had the "empty list" test repeat to ensure that the result is the same after additional tests. My rationale is that this ensures the `EventTracker._events` field is truly reset back to an empty list, though looking back, I can see it may be cleaner to keep one "empty list" test somewhere in between or at the end of the tests containing multiple events, as the list of events would be empty by default when the tests begin. It may also be worth implementing a check that the length of `EventTracker._events` is what is expected, just to be thorough. I will implement these changes as soon as I have a chance

samcli/lib/telemetry/metric.py
hawflau: > one or more of the three test cases will fail, as previous Events will be carried over into the new tests
We should avoid relying on this behavior. Each test should be standalone. States from previous test should not be carried over to next test.

Consider, if we create a test, and patch `EventTracker.clear_trackers` in that test. Then we'd be able to assert if `EventTracker.clear_trackers` is called. See [here](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called) for more info.

Leo10Gama: I see, that would definitely help in ensuring the functionality remains the same. In addition, that would also remove the need for test ordering to be in a specific fashion. I will implement these changes as soon as I get the chance

hawflau (approved):
LGTM! Thanks for making the changes!

qingchm (approved):

PR #3980, opened 2022-06-17
Refactored stack outputs command to the producer mapper consumer design pattern

andrew-zhan139:
#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/list/mapper.py
```diff
@@ -0,0 +1,13 @@
...
+import abc
+from typing import Generic, TypeVar
+
+T = TypeVar("T")
+
+
+class Mapper(Generic[T]):
+    @abc.abstractmethod
+    def map(self, data: T):
+        pass
```
mildaniel, line 13: Instead of `pass` for these abstract methods, let's add docstrings.

samcli/lib/list/mapper_consumer_factory_interface.py
```diff
@@ -0,0 +1,10 @@
+"""
+Interface for MapperConsumerFactory
+"""
+import abc
+
+
+class MapperConsumerFactoryInterface:
```
mildaniel, line 7: I believe that abstract classes in Python should inherit from `ABC`.

samcli/lib/list/mapper.py
```diff
@@ -0,0 +1,13 @@
+"""
+Interface for Mappers
+"""
+import abc
+from typing import Generic, TypeVar
+
+T = TypeVar("T")
```
mildaniel, line 7: Is there a way to also define a Generic as the output type for the mapper. Maybe something like 

```
InputType = TypeVar("InputType")
OutputType = TypeVar("OutputType")

class Mapper(Generic[InputType, OutputType]):
    @abstractmethod
    def map(self, input: InputType) -> OutputType:
        pass
```
I'm not sure about ^, it needs to be confirmed.

mildaniel (changes requested):
Thanks for doing the refactoring! Left some small comments.

samcli/lib/list/json_consumer.py
```diff
@@ -0,0 +1,10 @@
+"""
+The json consumer for 'sam list'
+"""
+import click
+from samcli.lib.list.consumer import Consumer
+
+
+class JsonConsumer(Consumer):
+    def consume(self, data: str) -> None:
+        click.echo(data)
```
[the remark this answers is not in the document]
andrew-zhan139: Moved it.

samcli/lib/list/mapper_consumer_factory.py
```diff
@@ -0,0 +1,16 @@
...
+from samcli.lib.list.mapper_consumer_factory_interface import MapperConsumerFactoryInterface
+from samcli.lib.list.data_to_json_mapper import DataToJsonMapper
+from samcli.lib.list.json_consumer import JsonConsumer
+from samcli.lib.list.mapper_consumer_container import MapperConsumerContainer
+
+
+class MapperConsumerFactory(MapperConsumerFactoryInterface):
+    def create(self, producer, output):
+        # Will add conditions here to return different sorts of containers later on
+        new_data_to_json_mapper = DataToJsonMapper()
```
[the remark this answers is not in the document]
andrew-zhan139: removed

samcli/lib/list/mapper.py
andrew-zhan139: Added the docstring

samcli/lib/list/mapper_consumer_factory_interface.py
andrew-zhan139: Added the inheritance

samcli/commands/list/exceptions.py
```diff
@@ -6,7 +6,7 @@
 from samcli.commands.exceptions import UserException
 
 
-class StackOutputsError(UserException):
+class SamListError(UserException):
```
mndeveci, line 9: Small doc could be useful for this exception. I assume this is a base one that we can use in `sam list` command?

samcli/commands/list/json_consumer.py
```diff
@@ -0,0 +1,10 @@
+"""
+The json consumer for 'sam list'
+"""
+import click
+from samcli.lib.list.list_interfaces import ListInfoPullerConsumer
+
+
+class JsonConsumer(ListInfoPullerConsumer):
```
mndeveci, line 8: It says JSON consumer but it is accepting `str`, should this be re-named to `StringConsumer`?

samcli/lib/list/mapper_consumer_container.py
```diff
@@ -0,0 +1,9 @@
+"""
+Container for a mapper and a consumer
+"""
+
+
+class MapperConsumerContainer:
```
mndeveci, line 6: What is this been used for?

samcli/lib/list/mapper_consumer_factory.py
```diff
@@ -0,0 +1,16 @@
...
+from samcli.commands.list.json_consumer import JsonConsumer
+from samcli.lib.list.mapper_consumer_container import MapperConsumerContainer
+
+
+class MapperConsumerFactory(MapperConsumerFactoryInterface):
+    def create(self, producer, output):
+        # Will add conditions here to return different sorts of containers later on
+        data_to_json_mapper = DataToJsonMapper()
+        json_consumer = JsonConsumer()
+        container = MapperConsumerContainer(mapper=data_to_json_mapper, consumer=json_consumer)
```
mndeveci, line 15: nit: no need to use named parameters, you can just write this `container = MapperConsumerContainer(data_to_json_mapper, json_consumer)`

samcli/lib/list/stack_outputs/stack_outputs.py
```diff
@@ -0,0 +1,16 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class StackOutputs:
+    OutputKey: str
+    OutputValue: str
+    Description: str
+
+    def __init__(self, OutputKey, OutputValue, Description):
```
mndeveci, line 13: As far as I remember, you don't need constructor, if you are using `@dataclass`

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,68 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class StackOutputsProducer(Producer):
+    def __init__(self, stack_name, output, region, profile, cloudformation_client):
+        self.stack_name = stack_name
+        self.output = output
+        self.region = region
+        self.profile = profile
```
mndeveci, line 22: `self.profile` seems redundant, any plans to use it in this class?

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,68 @@
...
+            raise SamListError(msg=str(e)) from e
+        except BotoCoreError as e:
+            LOG.error("Botocore Exception : %s", str(e))
+            raise SamListError(msg=str(e)) from e
+
+    def produce(self):
+        self.factory = MapperConsumerFactory()
+        new_container = self.factory.create(producer="stackoutputsproducer", output=self.output)
+        self.mapper = new_container.mapper
+        self.consumer = new_container.consumer
```
mndeveci, line 59: Since `output` value won't change after execution is started, you can also create mapper & consumer and pass them through constructor. So that this class would be more dummy where it will assume there will be a mapper and consumer that it can use.

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,68 @@
...
+        response = self.get_stack_info()
+        for stack_output in response["Stacks"][0]["Outputs"]:
```
mndeveci, line 61: I would recommend using `response.get("Stacks", [])` so that you will have default values in case that key doesn't exist in the dictionary.
And for accessing first element of it, I would recommend checking it first before accessing it since it might raise out of index error.

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,68 @@
...
+                raise StackDoesNotExistInRegionError(stack_name=self.stack_name, region=self.region) from e
+            LOG.error("ClientError Exception : %s", str(e))
+            raise SamListError(msg=str(e)) from e
+        except BotoCoreError as e:
+            LOG.error("Botocore Exception : %s", str(e))
+            raise SamListError(msg=str(e)) from e
+
+    def produce(self):
+        self.factory = MapperConsumerFactory()
+        new_container = self.factory.create(producer="stackoutputsproducer", output=self.output)
```
mndeveci, line 57: For value of `"stackoutputproducer"`, I would recommend writing an enum for it, so that when you are using it somewhere else, you won't make a type mistake while writing the string value.

mndeveci (changes requested):

samcli/commands/list/json_consumer.py
andrew-zhan139: It's called json consumer because it outputs the data in json format. Maybe I could write a docstring to clarify that?

samcli/lib/list/mapper_consumer_container.py
andrew-zhan139: It's a container for each producer's mapper and consumer to be returned together from the factory

samcli/commands/list/json_consumer.py
mndeveci: I see but from implementation point of view, it prints whatever string that is been given to it. So I would rather name it as string consumer.

samcli/lib/list/mapper_consumer_container.py
mndeveci: I see, you can use `@dataclass` for this one to simplify the implementation

samcli/commands/list/json_consumer.py
andrew-zhan139: Sure, but can I leave some sort of indication of the json output in the name? Since there will be a consumer that outputs a table format

mndeveci: Sure, I would suggest to put some documentation in it.

samcli/commands/list/exceptions.py
andrew-zhan139: Added doc

samcli/commands/list/json_consumer.py
andrew-zhan139: Changed it and added doc

samcli/lib/list/mapper_consumer_factory.py
andrew-zhan139: removed named parameters

samcli/lib/list/stack_outputs/stack_outputs_producer.py
andrew-zhan139: Switched

samcli/lib/list/stack_outputs/stack_outputs_producer.py
andrew-zhan139: Switched to use enum

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            response = self.cloudformation_client.describe_stacks(StackName=self.stack_name)
+            if not response.get("Stacks", []):
+                raise StackDoesNotExistInRegionError(stack_name=self.stack_name, region=self.region)
+            if len(response.get("Stacks", [])) > 0 and "Outputs" not in response.get("Stacks", [])[0]:
+                raise NoOutputsForStackError(stack_name=self.stack_name, region=self.region)
+            return response["Stacks"][0]["Outputs"]
+
+        except ClientError as e:
+            if "Stack with id {0} does not exist".format(self.stack_name) in str(e):
```
mndeveci, line 43: You can use this method to get the error code and compare against it: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/boto_utils.py#L140

It would be better since exception message might change but it is highly unlikely that the error code will change.

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,61 @@
...
+                LOG.debug("Stack with id %s does not exist", self.stack_name)
+                raise StackDoesNotExistInRegionError(stack_name=self.stack_name, region=self.region) from e
+            LOG.error("ClientError Exception : %s", str(e))
+            raise SamListError(msg=str(e)) from e
```
mndeveci, line 47: Please raise unique exception for each different error case. In that way we can track any increase for an error.

samcli/lib/list/stack_outputs/stack_outputs_producer.py
```diff
@@ -0,0 +1,61 @@
...
+class StackOutputsProducer(Producer):
+    def __init__(self, stack_name, output, region, cloudformation_client, mapper, consumer):
+        self.stack_name = stack_name
+        self.output = output
+        self.region = region
+        self.cloudformation_client = cloudformation_client
+        self.mapper = mapper
+        self.consumer = consumer
+
+    def get_stack_info(self):
```
mndeveci, line 25: return type is missing

samcli/lib/list/mapper_consumer_factory.py
```diff
@@ -0,0 +1,16 @@
...
+The factory for returning the appropriate mapper and consumer
+"""
+from samcli.lib.list.list_interfaces import MapperConsumerFactoryInterface
+from samcli.lib.list.data_to_json_mapper import DataToJsonMapper
+from samcli.commands.list.json_consumer import StringConsumerJsonOutput
+from samcli.lib.list.mapper_consumer_container import MapperConsumerContainer
+
+
+class MapperConsumerFactory(MapperConsumerFactoryInterface):
+    def create(self, producer, output):
```
mndeveci, line 11: return type is missing

mndeveci (changes requested):

samcli/commands/list/exceptions.py
```diff
@@ -6,7 +6,11 @@
...
+    """
+    Base exception for the 'sam list' command
+    """
+
```
mndeveci, line 13: nit: I think we don't need this `__init__` method here? Can you remove and verify if we need it or not?

mndeveci (commented):
Thanks for addressing all the feedback, left one final comment there, rest looks good to me!

samcli/commands/list/exceptions.py
andrew-zhan139: I removed it and got errors because there's I can't pass in the msg anymore

mndeveci (approved):

mildaniel (approved):

PR #3989, opened 2022-06-22
fix: add sync state and context to check previous run details

mndeveci:
N/A

If customer runs `sam sync` with a following `sam sync --no-dependency-layer` for a Java function without making a code change, the second run will fail since cached build will kick in and use the package which doesn't have dependencies in it (since previous run was using ADL which omits the dependencies). This issue only happens for Java runtimes, since we exclude dependencies after build is completed.

By introducing "Sync State" which will keep track of used options in previous session so that `sam sync` can check if there were any changes and run cleaning operation before starting the sync.

If customer runs `sam sync` (with ADL) and then switches to `sam sync --no-dependency-layer` (without ADL), their cached/incremental builds will be cleaned up so a full build needs to be completed. Sub-sequent runs after this will have cached/incremental builds.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/sync_context.py
```diff
@@ -0,0 +1,105 @@
...
+        # if adl parameter is changed between sam sync runs, cleanup build, cache and dependencies folders
+        if self._previous_state and self._previous_state.dependency_layer != self._current_state.dependency_layer:
+            self._cleanup_build_folders()
+
+        return self
+
+    def __exit__(self, *args):
+        self._write()
+
+    def _write(self) -> None:
```
qingchm, line 81: Can we protect this with a lock since open + write isn't atomic?

qingchm (approved):
To capture a few discussion points mentioned in the code review session: 
1. The wrapping of the withs for context creation in sync seems to have too many layers, it would be nice to have it refactored to be cleaner. However the shared temp dirs and template file location between contexts, as well as the scheduled delays needs to be preserved, so we need to be careful about it. It would be nice to have a separate PR to address it, but for current release scope we will play safe and leave it as it is right now.
2. As the current behaviour of the sync state stores the dependency layer's status, we explored the possibility of auto-saving this into samconfig.toml. That would help in the scenario where the user deployed the initial application with no-dependency-layer but forget to add this flag in later executions, current solutions will not fill in the previous option value,  but instead safely switch to ADL under this scenario. A suggested solution is that after each execution we prompt the user if they want to store the setting into toml. But after discussion with UX, this improvement will be addressed by environment and is not in this PR's scope.
3. Locks will protect one process's thread safety, since for each process only one sync state instance will be instantiated, we will not need of locking mechanism on toml write even though it's not atomic. As for write errors, we will not be catching it since this is consistent with how we do it for build.toml, currently we don't see any issues/dangers.

tests/integration/sync/test_sync_infra.py
```diff
@@ -315,3 +315,50 @@ def test_cdk_templates(self, template_file, template_after, function_id, depende
                 lambda_response = json.loads(self._get_lambda_response(lambda_function))
                 self.assertIn("extra_message", lambda_response)
                 self.assertEqual(lambda_response.get("message"), "9")
+
+
+@skipIf(SKIP_SYNC_TESTS, "Skip sync tests in CI/CD only")
+@parameterized_class([{"dependency_layer": True}, {"dependency_layer": False}])
+class TestSyncInfraWithJava(SyncIntegBase):
+    dependency_layer = True
```
mildaniel, line 323: Should we remove this initialization? It is already being set by the parameterized class.

mndeveci: Good catch, I kept it for testing without parameters, let me remove it 👍🏼

mildaniel (approved):

PR #3990, opened 2022-06-22
bug: make sure all pullers in child combine puller also get stopped

mingkun2020:
ctrl+c to sam log command with --include-traces option currently doesn't kill the traces' combine puller, with this fix now it should works correctly

overwrite the combine puller's `stop_tailing()` method to make sure it call `stop_tailing() `in all its children pullers.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/observability/observability_info_puller.py
+++ b/samcli/lib/observability/observability_info_puller.py
@@ -223,5 +223,7 @@ class ObservabilityCombinedPuller(ObservabilityPuller):
         async_context.run_async()
 
     def stop_tailing(self):
+        # if ObservabilityCombinedPuller A is a child puller in other ObservabilityCombinedPuller B, make sure A's child
+        # pullers stop as well when B stops.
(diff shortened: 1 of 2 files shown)


samcli/lib/observability/observability_info_puller.py
```diff
@@ -188,7 +191,7 @@ def tail(self, start_time: Optional[datetime] = None, filter_pattern: Optional[s
         except KeyboardInterrupt:
             LOG.info(" CTRL+C received, cancelling...")
             for puller in self._pullers:
-                puller.cancelled = True
+                puller.stop_tailing()
```
mndeveci, line 194: We don’t need this for loop anymore, you can call `self.stop_tailing()` which will do exactly same thing.

mndeveci (changes requested):

mndeveci (approved):

PR #3992, opened 2022-06-22
feat: Add --resource option input validation for sync --code

qingchm:
N.A.

When running sync --code, users can use --resource flag to run sync on specific resource types, for example --resource AWS::Serverless::Function means code sync on all serverless function resources. However if an invalid resource type is provided, the sync is skipped without letting the customer to be aware. This adds resource type validation and fail the command in that scenario.

By having a class that stores all supported resource types for sync code.

Not known

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/sync/test_sync_code.py
+++ b/tests/integration/sync/test_sync_code.py
@@ -119,7 +119,6 @@ class TestSyncCode(TestSyncCodeBase):
             s3_prefix=self.s3_prefix,
             kms_key_id=self.kms_key,
             tags="integ=true clarity=yes foo_bar=baz",
-            debug=True,
         )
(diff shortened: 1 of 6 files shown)


tests/integration/sync/test_sync_code.py
```diff
@@ -119,6 +119,7 @@ def test_sync_code_function(self):
             s3_prefix=self.s3_prefix,
             kms_key_id=self.kms_key,
             tags="integ=true clarity=yes foo_bar=baz",
+            debug=True,
```
qingchm, line 122: Will remove this flag for simplicity

mndeveci (approved):

PR #3993, opened 2022-06-22
Support OIDC

sidhujus:
Preliminary work to add support for using OIDC for identities instead of IAM Users in SAM Pipelines 

The changes in this don't currently get used but will be required. It adds a new Resource called OidcProvider, as well as additional methods to check whether an OidcProvider has already been created in IAM, and also a method to generate the thumbprint that is required when creating OidcProviders

It adds a dependency on pyOpenSSL. I haven't been able to check with Python 3.7 because it doesn't seem to work on my laptop so I can't run the update-reproducible-reqs command

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/pipeline/bootstrap/stage.py
+++ b/samcli/lib/pipeline/bootstrap/stage.py
@@ -3,16 +3,21 @@ import json
 import os
 import pathlib
 import re
+import socket
+import hashlib
 from itertools import chain
 from typing import Dict, List, Optional, Tuple
 
 import boto3
 import click
+import requests
+
+from OpenSSL import SSL, crypto  # type: ignore
 
 from samcli.lib.config.samconfig import SamConfig
 from samcli.lib.utils.colors import Colored
 from samcli.lib.utils.managed_cloudformation_stack import manage_stack, StackOutput
-from samcli.lib.pipeline.bootstrap.resource import Resource, IAMUser, ECRImageRepository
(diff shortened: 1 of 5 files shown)


samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -120,6 +142,46 @@ def _get_non_user_provided_resources_msg(self) -> str:
         )
         return "\n".join([f"\t- {comment}" for comment in resource_comments])
 
+    def _should_create_new_provider(self) -> bool:
```
hoffa, line 145: Where is this used? Applies to other methods as well.

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -120,6 +142,46 @@ def _get_non_user_provided_resources_msg(self) -> str:
...
+            return False
+        iam_client = boto3.client("iam")
+        providers = iam_client.list_open_id_connect_providers()
+        url_to_compare = self.oidc_provider.provider_url.replace("https://", "")
+        for provider_resource in providers["OpenIDConnectProviderList"]:
+            if url_to_compare in provider_resource["Arn"]:
+                return False
+        return True
+
+    def _generate_thumbprint(self) -> Optional[str]:
```
hoffa, line 156: This could be static; it seems generating the thumbprint for some domain is a well-defined process that makes sense on its own. Would be helpful to also e.g. some authoritative reference that defines what a thumbprint is technically.

PR #3995, opened 2022-06-22
Add OIDC support

sidhujus:
Adds support for Pipeline bootstrap command to create an OIDC identity provider instead of an IAM User to be used in CI/CD systems

Currently using pipelines bootstrap only has an option for IAM Users that requires people to copy over secret access keys. Using the built in OIDC support that CICD systems have will allow pipeline execution role to be assumed via external identity providers rather than an IAM user

It adds a dependency on pyOpenSSL. I haven't been able to check with Python 3.7 because it doesn't seem to work on my laptop so I can't run the update-reproducible-reqs command. It also requires us to send an HTTP GET request and then open another connection, which adds some time to the flow of the command

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -100,14 +136,64 @@ def __init__(
...
+            comment="IAM OIDC Identity Provider",
+            arn="",
+        )
+
+    def _should_create_new_provider(self) -> bool:
+        if not self.oidc_provider.provider_url:
+            return False
+        iam_client = boto3.client("iam")
+        providers = iam_client.list_open_id_connect_providers()
+        url_to_compare = self.oidc_provider.provider_url.replace("https://", "")
```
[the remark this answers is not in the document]
sidhujus: Yea since the ARN has the entire URL except for the protocol, and then by seeing if that part is in the ARN of one of the existing providers we can figure out if a new provider needs to be created or not without having to ask.

arn:aws:iam::<ACCOUNT>:oidc-provider/api.bitbucket.org/2.0/workspaces/<USERNAME>/pipelines-config/identity/oidc

is the format of the ARN that using the public bitbucket instance for OIDC ends up as.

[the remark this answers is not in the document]
hoffa: Oh gotcha, makes sense. A comment with e.g. example ARN would be helpful; it's not immediately obvious why a URL is compared with an ARN.

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -100,14 +136,64 @@ def __init__(
...
+        for provider_resource in providers["OpenIDConnectProviderList"]:
+            if url_to_compare in provider_resource["Arn"]:
+                return False
```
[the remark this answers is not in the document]
sidhujus: The only time the thumbprint gets used is when a new OIDC provider needs to be made in IAM, so if the provider already exists theres no more reason to re-generate the thumbprint

samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -100,14 +136,64 @@ def __init__(
...
+        r = requests.get(oidc_config_url, timeout=5)
+        jwks_uri = r.json()["jwks_uri"]
+        url_start = jwks_uri.find("//") + 2
+        url_end = jwks_uri[url_start:].find("/") + url_start
+        url_for_certificate = jwks_uri[url_start:url_end]
+
+        # Create connection to retrieve certificate
+        address = (url_for_certificate, 443)
+        ctx = SSL.Context(SSL.TLS_METHOD)
+        # s = socket.create_connection(address, timeout=5)
```
xazhao, line 170: Can remove this line :)

samcli/commands/pipeline/bootstrap/pipeline_oidc_provider.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, oidc_parameters: dict) -> None:
+        self.oidc_parameters = oidc_parameters
+
+    def verify_common_parameters(self) -> str:
+        error_string = ""
+        for parameter_name in self.COMMON_OIDC_PARAMETERS_NAMES:
+            if not self.oidc_parameters[parameter_name]:
+                error_string += f"Missing required parameter '{parameter_name}'\n"
+        return error_string
```
xazhao, line 23: `verify_common_parameters` and `verify_subject_claim_parameters` seems share the same logic. Can we create a method to be shared between these classes so we can avoid some duplicates?

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -144,43 +168,138 @@ def _prompt_image_repository(self) -> None:
...
+            "", type=click.Choice((["1", "2"])), show_choices=False, show_default=False, default="1", prompt_suffix=""
+        )
+        self.permissions_provider = OPEN_ID_CONNECT if user_provider == "2" else "iam"
+
+    def _prompt_oidc_provider(self) -> None:
+        click.echo("Select an OIDC provider:")
+        for (key, provider) in self.SUPPORTED_OIDC_PROVIDERS.items():
+            click.echo("\t{key} - {provider}".format(key=key, provider=self.OIDC_PROVIDER_NAME_MAPPINGS[provider]))
+        oidc_provider = click.prompt(
+            "",
```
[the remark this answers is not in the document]
sidhujus: Yea it prompts on an empty line. Ill change it to "Choice (OPTIONS):" since right now it hides the options on the prompt line

samcli/commands/pipeline/bootstrap/pipeline_oidc_provider.py
```diff
@@ -0,0 +1,61 @@
...
+class PipelineOidcProvider:
+
+    PROVIDER_URL_PARAMETER = "oidc-provider-url"
+    CLIENT_ID_PARAMETER = "oidc-client-id"
+
+    def __init__(self, oidc_parameters: dict, oidc_parameter_names: List[str]) -> None:
+        self.oidc_parameters = oidc_parameters
+        self.oidc_parameter_names = [self.PROVIDER_URL_PARAMETER, self.CLIENT_ID_PARAMETER] + oidc_parameter_names
+
+    def verify_parameters(self) -> None:
```
hoffa, line 20: If this is needed for the object to considered correct, should this be in the constructor instead?

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -241,5 +357,20 @@ def _load_saved_pipeline_user_arn() -> Optional[str]:
...
+        return {}
+    config: Dict[str, str] = samconfig.get_all(cmd_names=_get_bootstrap_command_names(), section="parameters")
+    oidc_parameters: Dict[str, Optional[str]] = {}
+    oidc_parameters[OIDC_PROVIDER] = config.get(OIDC_PROVIDER)
+    oidc_parameters[OIDC_PROVIDER_URL] = config.get(OIDC_PROVIDER_URL)
+    oidc_parameters[OIDC_CLIENT_ID] = config.get(OIDC_CLIENT_ID)
+    oidc_parameters[GITHUB_ORG] = config.get(GITHUB_ORG)
+    oidc_parameters[GITHUB_REPO] = config.get(GITHUB_REPO)
+    oidc_parameters[DEPLOYMENT_BRANCH] = config.get(DEPLOYMENT_BRANCH)
+    return oidc_parameters
```
hoffa, line 372: nit: can reduce verbosity by `return`ing the dict directly

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -144,43 +168,134 @@ def _prompt_image_repository(self) -> None:
...
+
+    def _prompt_github_org(self) -> None:
+        self.github_org = click.prompt(
+            "Enter the GitHub Organization that the code repository belongs to."
+            " If there is no organization enter your username instead",
+            type=click.STRING,
+        )
+
+    def _prompt_github_repo(self) -> None:
+        self.github_repo = click.prompt("Enter GitHub Repository name", type=click.STRING)
```
hoffa, line 220:
```suggestion
        self.github_repo = click.prompt("Enter GitHub repository name", type=click.STRING)
```

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -144,43 +168,134 @@ def _prompt_image_repository(self) -> None:
...
+            if not self.github_org:
+                self._prompt_github_org()
+            if not self.github_repo:
+                self._prompt_github_repo()
+            if not self.deployment_branch:
+                self._prompt_github_branch()
+
+    def _prompt_github_org(self) -> None:
+        self.github_org = click.prompt(
+            "Enter the GitHub Organization that the code repository belongs to."
```
hoffa, line 214:
```suggestion
            "Enter the GitHub organization that the code repository belongs to."
```

hoffa (approved):

PR #3998, opened 2022-06-23
Hooks wrapper

hawflau:
#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/hook/hook_config.py
+++ b/samcli/lib/hook/hook_config.py
@@ -1,27 +1,22 @@
 """Hook Package Config"""
 import json
 from pathlib import Path
-from typing import Dict, Iterable, List, NamedTuple, Optional
+from typing import Dict, NamedTuple, cast
 
-import jsonschema
+import jsonschema  # type: ignore
 from .exceptions import InvalidHookPackageConfigException
 
 
-class HookFunctionalityParam(NamedTuple):
-    long_name: str
-    short_name: str
-    description: str
-    mandatory: bool
-    type: str
-
-
 class HookFunctionality(NamedTuple):
-    entry_script: str
-    parameters: Optional[List[HookFunctionalityParam]]
+    entry_method: Dict[str, str]
 
     @property
-    def mandatory_parameters(self) -> List[HookFunctionalityParam]:
-        return [params for params in self.parameters if params.mandatory]
+    def module(self) -> str:
+        return self.entry_method["module"]
+
+    @property
+    def method(self) -> str:
+        return self.entry_method["method"]
 
 
 class HookPackageConfig:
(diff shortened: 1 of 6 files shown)


samcli/lib/hook/hook_wrapper.py
```diff
@@ -0,0 +1,176 @@
...
+        Parameters
+        ----------
+        hook_package_id: str
+            Hook package ID
+        """
+        # locate hook package from internal first
+        for child in self._INTERNAL_PACKAGES_ROOT.iterdir():
+            try:
+                hook_package_config = HookPackageConfig(child)
+                if hook_package_config.package_id == hook_package_id:
```
moelasmar, line 110: let's do not use the Config.json package Id as the `hook-pacakge-id` parameter value, because in future if we will move to 3rd party hooks, and we implement the `sam hooks install` command, the customer may be able to customize the hook package id to solve the issue if there are multiple hooks that were developed and their owners use the same package id.
So, for now for the internal use, my assumption was to use the directory name as the hook_package_id.

samcli/lib/hook/hook_wrapper.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        self._config = None
+        self._load_hook_package(hook_package_id)
+
+    def prepare(
+        self,
+        output_dir_path: str,
+        iac_project_path: Optional[str] = None,
+        debug: bool = False,
+        logs_path: Optional[str] = None,
```
moelasmar, line 53: no need for this parameter now, as hooks will be called as python libs. Even in the future, we agreed to use the error stream for logs, and output stream for output communication.

moelasmar (changes requested):
This PR looks great, just left minor comments

samcli/lib/hook/hook_config.py
```diff
@@ -0,0 +1,89 @@
...
+    def version(self) -> str:
+        return cast(str, self._config["version"])
+
+    @property
+    def specification(self) -> str:
+        return cast(str, self._config["hook_specification"])
+
+    @property
+    def description(self) -> str:
+        return cast(str, self._config["description"])
```
torresxb1, line 85: description is not required, so `self._config["description"]` could be `None` and it seems like calling cast on this just results in None / NoneType. Maybe we should change the return type to Optional[str]?

moelasmar (approved):

moelasmar (approved):

tests/unit/lib/hook/test_hook_config.py
```diff
@@ -0,0 +1,105 @@
...
+from copy import deepcopy
+from unittest import TestCase
+from unittest.mock import patch, MagicMock, mock_open
+
+from samcli.lib.hook.hook_config import HookFunctionality, HookPackageConfig
+from samcli.lib.hook.exceptions import InvalidHookPackageConfigException
+
+TEST_HOOK_PACKAGE_CONFIG = {
+    "hook_package_id": "my_test_hook_package_id",
+    "hook_use_case": "IaC",
```
torresxb1, line 12: this example made me think, should we require iac_framework if the use case is IaC?
I think we can use dependentRequired: https://json-schema.org/understanding-json-schema/reference/conditionals.html#id4

torresxb1: oh wait I think dependentRequired is only for when the depending property is present but not that it equals some specific value

torresxb1 (approved):

PR #4003, opened 2022-06-24
Add UsedFeature event and trackers

Leo10Gama:
N/A

This change adds trackers for the `UsedFeature` event, which allow for the tracking of usage of features such as Accelerate, CDK, and ESBuild.

Trackers are placed in the respective methods for each of these events, and their information is sent to the `EventTracker`, which will hold the information that will be sent by the subsequent `@track_command` decorator.

Since all trackers are placed within functions under the `@track_command` decorator, their Event information will be sent immediately upon completion of the function. This will result in no additional side effects, besides creating a slightly larger payload for the `metricSpecificAttributes` value of the `Metric` class.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -270,6 +270,8 @@ class BuildContext:
 
             for f in self.get_resources_to_build().functions:
                 EventTracker.track_event("BuildFunctionRuntime", f.runtime)
+                if f.metadata and f.metadata.get("BuildMethod", "") == "esbuild":
+                    EventTracker.track_event("UsedFeature", "ESBuild")
 
             click.secho("\nBuild Succeeded", fg="green")
 
--- a/samcli/commands/sync/command.py
+++ b/samcli/commands/sync/command.py
@@ -27,6 +27,7 @@ from samcli.commands._utils.options import (
 )
 from samcli.cli.cli_config_file import configuration_option, TomlProvider
 from samcli.commands._utils.click_mutex import ClickMutex
+from samcli.lib.telemetry.event import EventTracker
 from samcli.lib.utils.colors import Colored
 from samcli.lib.utils.version_checker import check_newer_version
 from samcli.lib.bootstrap.bootstrap import manage_stack
@@ -260,6 +261,7 @@ def do_cli(
 
     set_experimental(ExperimentalFlag.Accelerate)
     update_experimental_context()
+    EventTracker.track_event("UsedFeature", "Accelerate")
 
     with BuildContext(
         resource_identifier=None,
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -3,10 +3,13 @@ Represents Events and their values.
 """
 
 from enum import Enum
+import logging
 from typing import List
 
 from samcli.local.common.runtime_template import INIT_RUNTIMES
 
+
+LOG = logging.getLogger(__name__)
 
 class EventName(Enum):
     """Enum for the names of available events to track."""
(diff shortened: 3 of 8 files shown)


samcli/commands/build/build_context.py
```diff
@@ -270,6 +270,8 @@ def run(self):
 
             for f in self.get_resources_to_build().functions:
                 EventTracker.track_event("BuildFunctionRuntime", f.runtime)
+                if f.metadata and f.metadata.get("BuildMethod", "") == "esbuild":
+                    EventTracker.track_event("UsedFeature", "ESBuild")
```
hawflau, line 274: a question - Did we anticipate that `EventTracker.track_event` would become the point of failure? i.e. Will it cause the command to fail?

Leo10Gama: I don't think we ever did... One solution I can think of might be to put the method in a try-catch? That way instead of throwing an error and breaking the program flow, it could print to console or `LOG`

samcli/lib/telemetry/event.py
```diff
@@ -110,7 +113,10 @@ def track_event(event_name: str, event_value: str):
                 EventTracker.track_event("UsedFeature", "FeatureY")
                 return some_value
         """
-        EventTracker._events.append(Event(event_name, event_value))
+        try:
+            EventTracker._events.append(Event(event_name, event_value))
+        except EventCreationError as e:
+            LOG.error("Error occurred while trying to track an event: %s", e)
```
hawflau, line 119: minor - `LOG.debug` is enough, as we are not throwing an exception.

hawflau (changes requested):

samcli/commands/sync/command.py
```diff
@@ -260,6 +261,7 @@ def do_cli(
 
     set_experimental(ExperimentalFlag.Accelerate)
     update_experimental_context()
+    EventTracker.track_event("UsedFeature", "Accelerate")
```
mndeveci, line 264: I would recommend using enums for `UsedFeature` and `Accelerate` values, so that we don't need to repeat same string in other places. And it will also eliminate any possible typo in them.

mndeveci (changes requested):

samcli/commands/sync/command.py
Leo10Gama: While we do have the enum `EventName`, the method currently only accepts a string as its first parameter. The `Accelerate` value is verified via the `EventType` class, which is cross-referenced upon creation of an Event, as I found that was the cleanest way to keep track of each event's potential values, as well as the simplest way to add on new events in the future. A refactor for `track_event` to take an `EventName` instead is possible, and would require an extra import of the enum, but it may be best to push changes like those in a separate PR since it would affect all (7) currently existing `track_event` calls, as well as its unit test methods.

samcli/lib/telemetry/event.py
```diff
@@ -110,7 +113,10 @@ def track_event(event_name: str, event_value: str):
                 EventTracker.track_event("UsedFeature", "FeatureY")
                 return some_value
         """
-        EventTracker._events.append(Event(event_name, event_value))
+        try:
+            EventTracker._events.append(Event(event_name, event_value))
+        except EventCreationError as e:
```
hawflau, line 118: just one last question - is `EventCreationError` the only exception that can occur? Any possibility that `EventTracker._events.append` would fail?

Leo10Gama: There's no situation I can see where `append` would fail, since it's a standard list operation. The only place I can see that may go wrong is something being thrown when the Event itself is created, which is why I capture that error specifically.

hawflau: sounds good. makes sense to me.

hawflau (approved):

samcli/commands/sync/command.py
mndeveci: Sure, since this is merging to feature branch we can make those changes later on then.

mndeveci (approved):

Issue #4004, opened 2022-06-26
Bug: infinite deploy loop on small stack edge case


BourgoisMickael:
_Note: I'll make a PR, it's a small change_

### Description:
Introduced by my last PR #3919 
`sam deploy` is stuck in `describe_stack_events` loop without catching events because the stack already finished deploying. 
Happens when cloudformation throttle because there are many deploys in parralel.

### Steps to reproduce:
Use a small stack like:
```yaml
SomeTopic:
  Type: AWS::SNS::Topic
  Properties:
    Subscription:
    - Protocol: email
      Endpoint:  emailh@example.com
```
This will take like ~5 seconds to deploy. And if deployed alone, there is no problem.
But if you have 50+ stacks being deployed at the same time (in CI pipelines) with env `AWS_MAX_ATTEMPTS: 30` to retry a lot, then there will be a lot of calls to `DescribeStackEvents` and cloudformation will throttle.
So just after [`execute_changeset`](https://github.com/aws/aws-sam-cli/blob/6110cb948f3291ed1ffe064c3b7b8da1c8b9ceee/samcli/commands/deploy/deploy_context.py#L267), the function [`get_last_event_time`](https://github.com/aws/aws-sam-cli/blob/6110cb948f3291ed1ffe064c3b7b8da1c8b9ceee/samcli/lib/deploy/deployer.py#L346) might take a few seconds and return the wrong time.

### Example timestamp output of [`wait_for_execute`](https://github.com/aws/aws-sam-cli/blob/6110cb948f3291ed1ffe064c3b7b8da1c8b9ceee/samcli/lib/deploy/deployer.py#L447):
**No parallel deploy - no throttle - 0s time diff**

```
Thu, 21 Jun 2022 14:24:32 GMT 2022-06-21 14:24:32 - Waiting for stack create/update to complete
Thu, 21 Jun 2022 14:24:32 GMT CloudFormation events from stack operations (refresh every 0.5 seconds) 
```

**Many parallel deploy - throttle - 11s diff**
```
Mon, 20 Jun 2022 14:09:28 GMT 2022-06-20 14:09:28 - Waiting for stack create/update to complete
Mon, 20 Jun 2022 14:09:39 GMT CloudFormation events from stack operations (refresh every 0.5 seconds)
 ```
[...]

PR #4005, opened 2022-06-26
Fix infinite deploy loop on small stack edge case

BourgoisMickael:
#4004 

In case you have lots of parallel deploys that might throttle cloudformation (or network condition) and a small stack/changeset that could complete deployment before the CLI reach the `describe_stack_events` loop. (so a delay in the `get_last_event_time`)

Rely on `execute_change_set` / `create_stack` / `update_stack` request date instead of last event date to start the describe loop. 
There should be no possible delay between the stack operation and the first `time_stamp_marker` for `descibe_stack_events` .
So no way to skip events anymore.

`get_last_event_time` is not used aymore
https://github.com/aws/aws-sam-cli/blob/6110cb948f3291ed1ffe064c3b7b8da1c8b9ceee/samcli/lib/deploy/deployer.py#L346
(should I delete the function ?)

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes


hawflau (approved):
LGTM!

samcli/lib/deploy/deployer.py
```diff
@@ -444,7 +445,9 @@ def _is_root_stack_event(event: Dict) -> bool:
     def _check_stack_not_in_progress(status: str) -> bool:
         return "IN_PROGRESS" not in status
 
-    def wait_for_execute(self, stack_name: str, stack_operation: str, disable_rollback: bool) -> None:
+    def wait_for_execute(
+        self, stack_name: str, stack_operation: str, disable_rollback: bool, execution_time: float
```
hawflau, line 449: Should we default `execution_time` to `time.time()? This make the method a bit more intuitive to use - If I don't need to check against a specific timestamp, I just don't provide this argument.

```suggestion
        self, stack_name: str, stack_operation: str, disable_rollback: bool, execution_time: float = time.time()
```

samcli/lib/deploy/deployer.py
```diff
@@ -457,14 +460,16 @@ def wait_for_execute(self, stack_name: str, stack_operation: str, disable_rollba
             The type of the stack operation, 'CREATE' or 'UPDATE'
         disable_rollback : bool
             Preserves the state of previously provisioned resources when an operation fails
+        execution_time : float
+            Changeset execution time (current time)
```
hawflau, line 464: Can we update the description to explain clearer about what this argument is for?

hawflau (changes requested):

samcli/lib/deploy/deployer.py
BourgoisMickael, line 464:
```suggestion
            Time of the last stack change execution request (like `execute_change_set`, `update_stack`, `create_stack`)
            Prevents missing events if the event streaming is delayed from the change request by any action in between
```

BourgoisMickael: I've improved the documentation

hawflau (approved):
Thanks for your contribution! LGTM

mildaniel (approved):

PR #4020, opened 2022-06-30
Local transform and resource collection

andrew-zhan139:
#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/translate/translate_utils.py
```diff
@@ -0,0 +1,31 @@
+"""
+Library for Validating Sam Templates
+"""
+import os
+import logging
+import click
```
torresxb1, line 6: I believe we should not use click outside of the `commands` directory: https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#code-conventions

samcli/lib/translate/translate_utils.py
```diff
@@ -0,0 +1,31 @@
+"""
+Library for Validating Sam Templates
```
torresxb1, line 2: probably doesn't belong here

samcli/lib/translate/translate_utils.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+import os
+import logging
+import click
+
+
+LOG = logging.getLogger(__name__)
+
+
+def _read_sam_file(template):
```
torresxb1, line 12: if it's meant to be used from outside this file, we should remove the leading underscore

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
+"""
+in progress
```
torresxb1, line 2: 😬

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
...
+from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
+from samcli.lib.translate.translate_utils import _read_sam_file
+from samcli.lib.translate.exceptions import InvalidSamDocumentException
+from samcli.commands.local.cli_common.user_exceptions import InvalidSamTemplateException
+from samcli.commands.exceptions import UserException
+
+LOG = logging.getLogger(__name__)
+
+
+class ResourceMappingProducer(Producer):
```
torresxb1, line 27: can we add type hints to the functions here? 🙏

samcli/lib/translate/sam_template_validator.py
```diff
@@ -45,6 +45,37 @@ def __init__(self, sam_template, managed_policy_loader, profile=None, region=Non
...
+
+        try:
+            template = sam_translator.translate(sam_template=self.sam_template, parameter_values={})
+            LOG.debug("Translated template is:\n%s", yaml_dump(template))
+            return yaml_dump(template)
+        except InvalidDocumentException as e:
+            raise InvalidSamDocumentException(
+                functools.reduce(lambda message, error: message + " " + str(error), e.causes, str(e))
+            ) from e
+
```
torresxb1, line 78: this is almost all copy-paste from is_valid, should probably do some refactoring here 😅

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
...
+        self.iam_client = iam_client
+        self.mapper = mapper
+        self.consumer = consumer
+
+    def get_translated_dict(self, template_file_dict) -> Optional[Any]:
+        try:
+            validator = SamTemplateValidator(
+                template_file_dict, ManagedPolicyLoader(self.iam_client), profile=self.profile, region=self.region
+            )
+            translated_dict = yaml.load(validator.get_translated_template(), Loader=yaml.FullLoader)
```
torresxb1, line 46: I wonder if we should really be making use of the validator here 🤔 . It seems what we really want is to just get the translated template. So maybe we should extract that logic into somewhere else where both this code and the validator code make use of it

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
+"""
+in progress
+"""
+from typing import Optional, Any
+import dataclasses
+import logging
+import yaml
+import click
```
torresxb1, line 8: same comment as the other one on not using click outside of `commands`

samcli/lib/providers/sam_stack_provider.py
```diff
@@ -218,6 +219,8 @@ def get_stacks(
             that might want to get substituted within the template and its child templates
         metadata: Optional[Dict]
             Optional dictionary of nested stack resource metadata values.
+        template_dict_format: Optional[Dict]
+            Optional dictionary representing the sam template file to be used instead of the template file
```
torresxb1, line 223: I was a bit confused by this before I looked into the implementation. I think just `template_dict` might be a better name, and saying `representing the sam template` rather than `representing the sam template file`. Also should leave some message that one of this one (and only one) of either this one or template file is required.

torresxb1: Also, since now the template file is not required (as they could provide the dict) we should probable reflect that in the function's parameters and make it an Optional type. Would also need to update the comment to reflect this (and probably also include again here a line that one and only one of either this or template_dict is required). Also, in that case, should handle below in the code what happens if the user doesn't provide template file or dict (should throw an exception)

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
...
+        stacks, _ = SamLocalStackProvider.get_stacks(template_file="", template_dict_format=translated_dict)
+        if not stacks or not stacks[0].resources:
+            raise SamListLocalResourcesNotFoundError(msg="No local resources found.")
+        resources_dict = {}
+        for local_resource in stacks[0].resources:
+            # Set the PhysicalID to "-" if there is no corresponding PhysicalID
+            resources_dict[local_resource] = "-"
+
+        for logical_id, physical_id in resources_dict.items():
+            resource_data = ResourcesDef(LogicalResourceId=logical_id, PhysicalResourceId=physical_id)
```
torresxb1, line 78: not sure if the dataclass is needed here, since it seems it's immediately turned into a dict in the next line 🤔

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
...
+
+        translated_dict = self.get_translated_dict(template_file_dict=sam_template)
+
+        stacks, _ = SamLocalStackProvider.get_stacks(template_file="", template_dict_format=translated_dict)
+        if not stacks or not stacks[0].resources:
+            raise SamListLocalResourcesNotFoundError(msg="No local resources found.")
+        resources_dict = {}
+        for local_resource in stacks[0].resources:
+            # Set the PhysicalID to "-" if there is no corresponding PhysicalID
+            resources_dict[local_resource] = "-"
```
torresxb1, line 75: is this part unfinished? Will we look for the actual physical ID later on?

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,80 @@
...
+
+        for logical_id, physical_id in resources_dict.items():
```
torresxb1, line 77: it seems like you could just do the logic that's under this loop in the loop above, no?

torresxb1 (changes requested):

samcli/commands/list/cli_common/list_common_context.py
```diff
@@ -0,0 +1,33 @@
...
+            region = session.region_name
+            if region:
+                self.region = region
+            else:
+                raise RegionError(
+                    message="No region was specified/found. "
+                    "Please provide a region via the --region parameter or by the AWS_REGION environment variable."
+                )
+
+        client_provider = get_boto_client_provider_with_config(region=self.region, profile=self.profile)
```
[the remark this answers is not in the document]
andrew-zhan139: ListContext is inherited by ResourcesContext that may have profile

samcli/lib/translate/translate_utils.py
andrew-zhan139: Moved it to the commands directory

samcli/lib/translate/translate_utils.py
andrew-zhan139: Fixed

samcli/lib/translate/translate_utils.py
andrew-zhan139: Removed the underscore

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: Fixed

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: Resolved

samcli/lib/translate/sam_template_validator.py
andrew-zhan139: Removed it and combined it with is_valid. is_valid was renamed to get_translated_template_if_valid

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: Fixed

samcli/lib/providers/sam_stack_provider.py
andrew-zhan139: Made the fixes and added the exception raise for if neither are passed

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: I think the dataclass makes it more organized and conforming to the design pattern

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: Added it to the above loop

samcli/commands/list/resources/resources_context.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def run(self) -> None:
+        """
+        Get the resources for a stack
+        """
+        factory = MapperConsumerFactory()
+        container = factory.create(producer=ProducersEnum.RESOURCES_PRODUCER, output=self.output)
+        resource_producer = ResourceMappingProducer(
+            stack_name=self.stack_name,
+            output=self.output,
```
mildaniel, line 49: Do we need to be passing the output to the producer if we're already passing the mapper and consumer?

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,87 @@
...
+        self.output = output
+        self.region = region
+        self.profile = profile
+        self.template_file = template_file
+        self.cloudformation_client = cloudformation_client
+        self.iam_client = iam_client
+        self.mapper = mapper
+        self.consumer = consumer
+
+    def get_translated_dict(self, template_file_dict: dict) -> Optional[Any]:
```
mildaniel, line 49: Let's include docstrings for all new methods please. Also for the return type, let's use `Dict[Any, Any]` as that's a little more accurate.

samcli/commands/translate/translate_utils.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+import os
+import logging
+import click
+
+
+LOG = logging.getLogger(__name__)
+
+
+def read_sam_file(template):
```
mildaniel, line 12: I found a similar function in `samcli/commands/_utils/template.py` maybe we can just re-use that one.

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            validator = SamTemplateValidator(
+                template_file_dict, ManagedPolicyLoader(self.iam_client), profile=self.profile, region=self.region
```
mildaniel, line 52: Let's leave a note to check if we can mock IAM here and get around doing a translate without it.

samcli/lib/providers/exceptions.py
```diff
@@ -84,3 +84,12 @@ def resource_identifier(self) -> "ResourceIdentifier":
     @property
     def property_name(self) -> str:
         return self._property_name
+
+
+class MissingTemplateFile(UserException):
```
mildaniel, line 89: Similar to the above comment, we already have a `TemplateNotFoundException` we can re-use.

mildaniel (changes requested):

samcli/commands/list/resources/resources_context.py
andrew-zhan139: Removed it

samcli/commands/translate/translate_utils.py
andrew-zhan139: switched to using that one

samcli/lib/list/resources/resource_mapping_producer.py
andrew-zhan139: Added the note

samcli/commands/translate/translate_utils.py
mildaniel: Do we still need this file?

andrew-zhan139: File removed

mildaniel (approved):

torresxb1 (approved):
Thanks for addressing all the feedback! 
Andrew and I talked about about unit testing, and we concluded that we'll merge in this PR and make some changes to unit testing in another PR:
- making sure each function has unit tests
- making sure the unit tests for each function test the functionality of *that* specific  function (and not the internal helper functions called, since those should have their own unit tests)

PR #4024, opened 2022-06-30
FEAT: Test Runner - Test Runner CloudFormation Template Generation

bentvelj:
This PR adds code responsible for generating the CloudFormation Template for the Test Runner.

### What is the Test Runner Feature?
The Test Runner (exact naming TBD) will be a tool allowing customers to run integration tests using AWS Fargate.
Through the SAM CLI, the customer creates and deploys a Test Runner CloudFormation stack with the following resources:

- S3Bucket: A bucket to store test code, test requirements (`requirements.txt`), and results after the test is run.
- [TaskDefinition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html): Defines the task which runs the customer tests. It uses a general base image that downloads customer tests and requirements from the S3 bucket, runs the tests, and uploads results.
- ContainerIAMRole: An IAM role that is assumed by the ECS task. Contains permissions to access the Test Runner S3 bucket (download tests, upload results). Also contains any _default permissions_ generated. To make writing permissions easier for the customer, a set of commonly used resources are supported with commonly used actions. An IAM policy statement is generated for each customer resource, and the common action permissions are added if the resource is one of the supported. For example, a Lambda Function will be given a policy statement with the permission `lambda:InvokeFunction`. See all currently supported default permissions [here](https://github.com/aws/aws-sam-cli/blob/f849e7a2f1b66d0b061fb157f93786ef64eb7e92/samcli/lib/test_runner/test_runner_template_generator.py#L67-L88). Any generated default permissions will be **commented out**. This is to establish a barrier of consent, to prevent customers from accidentally granting permissions that they did not mean to.
- LogGroup: For logging container activity
- [ECSCluster](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/clusters.html): To host the test running task.

##### Why Fargate, and not Lambda?
Lambda has a 15 minute time limit per execution, whereas Fargate is free of such time limits.

##### How are customer resources discovered?
Customers can specify tags for the resources they want IAM statements generated for. Tags are used in order to support both CloudFormation and Terraform customers.

N/A

In order to deploy a Test Runner Stack, customers need the Test Runner CloudFormation template.

This change contains the code to generate the Test Runner CloudFormation Template.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+    Returns a dictionary mapping a Regular Expression representing the arn of a specific resource to the default permissions generated for that resource.
+
+    Returns
+    -------
+    default_permissions_map : dict
+        A dictionary mapping a Regular Expression representing the arn of a resource to a list of IAM Action permissions generated for that resource.
+    """
+
+    partition_ex = "(aws|aws-cn|aws-us-gov)"
+    region_ex = "\w+-\w+-\d"
```
hoffa, line 47: This won't work for GovCloud regions.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns
+    -------
+    List[str]
+        The list of IAM Action permissions to generate for a given resource.
+    """
+
+    permissions_map = __get_permissions_map()
+
+    for arnExp in permissions_map.keys():
```
hoffa, line 94: You can iterate over `items()` instead, so you don't need the extra access when `return`ing.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+import re
+import logging
+from jinja2 import Template
+from cfn_flip import to_yaml
+from typing import *
+
+LOG = logging.getLogger(__name__)
+
+
+def __extract_api_id_from_arn(api_arn: str) -> str:
```
hoffa, line 12: nit: a single underscore prefix is the [more Pythonic](https://peps.python.org/pep-0008/#descriptive-naming-styles) notation for "private" functions

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+    -------
+    dict
+        A response object containing ResourceTagMappingList, which contains a list of resource ARNs and tags associated with each.
+
+        NOTE: https://docs.aws.amazon.com/cli/latest/reference/resourcegroupstaggingapi/get-resources.html#output
+
+
+    """
+    try:
+        return boto3.client("resourcegroupstaggingapi").get_resources(tag_filters=tag_filters)
```
hoffa, line 191: If only `ResourceTagMappingList` is accessed anyway, might consider returning that directly.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+        }
+        arn = resource["ResourceARN"]
+
+        new_statement["Action"].extend(
+            # Default permissions are commented out
+            ["# " + action for action in __get_permissions(arn)]
+        )
+
+        # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-control-access-using-iam-policies-to-invoke-api.html
+        if "# execute-api:Invoke" in new_statement["Action"]:
```
hoffa, line 236: Getting the `Action` and `Resource` for an ARN are coupled, but here in an indirect way that relies on the output of `__get_permissions`, instead of the ARN.

Could we instead have a function that returns the full IAM statement for some ARN? Should help make this loop cleaner as well.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+    There will be no other instances of `- # ` other than the ones created during the template generation process.
+
+    The purpose is to establish a barrier of consent, to prevent customers from accidentally granting permissions that they did not mean to.
+
+    e.g.
+
+        `- '# lambda:InvokeFunction'` => `# - lambda:InvokeFunction`
+    """
+    # Move comment hashtag to be in front of list items so they're properly commented out
+    return raw_yaml_string.replace("'", "").replace("- # ", "# - ")
```
hoffa, line 263: nit: I wonder if it would be cleaner to have the base template as YAML with a template variable for the IAM policy, then inject it with `"\n".join("# - " + action for action in actions)`. At a glance it would get rid of the need for `cfn-flip` and this hack

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Replacing with `(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d`

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Mhmm, can easily refactor back if we need other data from the response object

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Refactored so that an ARN maps to the `AWS::Service::Resource` as well as the IAM action list. Now the whole IAM statement is generated in a single function, and it is more clear which Resource Type the ARN regex maps to. APIGW resources can also be dealt with directly in the statement generating function.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Update: Found a way to do this. We can keep everything in YAML, avoid the `cfn-flip` dependency, and the hack.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    -------
+    List[str]
+        The list of IAM statements in the form of YAML strings corresponding to the resources specified by the given tags.
+    """
+
+    resource_tag_mapping_list = _query_tagging_api(tag_filters)
+
+    if resource_tag_mapping_list is None:
+        LOG.error("Failed to receive get_resources response, cannot generate any IAM statements.")
+        return ["# Failed to query tagging api, can't generate IAM permissions for your resources."]
```
mndeveci, line 212: This seems like an unexpected return value from this function, we should rather return `None` or raise an exception so that caller can handle this error case.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    -------
+    dict
+        A ResourceTagMappingList, which contains a list of resource ARNs and tags associated with each.
+
+        NOTE: https://docs.aws.amazon.com/cli/latest/reference/resourcegroupstaggingapi/get-resources.html#output
+
+
+    """
+    try:
+        return boto3.client("resourcegroupstaggingapi").get_resources(TagFilters=tag_filters)["ResourceTagMappingList"]
```
mndeveci, line 185: I would suggest using `get_boto_client_provider_with_config` which will return a client with `region` and `profile`  settings as well as updating user agent with SAM CLI.

That function will return a function reference which can be used later on, so that once you have `region` and `profile` settings, you can initalize the function reference, and you can use it later to initialize clients. 
```
# initalize the client provider
boto_client_provider = get_boto_client_provider_with_config(region=region, profile=profile)

# use it later to initialize the client
resource_group_tagging_api_client = boto_client_provider("resourcegroupstaggingapi")
```

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+        A dictionary mapping a Regular Expression representing the arn of a resource to an object containing the AWS Resource Type and the list of IAM Action permissions generated for that resource.
+    """
+
+    partition_ex = r"(aws|aws-cn|aws-us-gov)"
+    region_ex = r"(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d"
+    account_ex = r"\d{12}"
+
+    default_permissions_map = {
+        rf"^arn:{partition_ex}:lambda:{region_ex}:{account_ex}:function:[\w-]+(:\d+)?$": {
+            "Resource": "AWS::Lambda::Function",
```
mndeveci, line 50: Will this give permission to execute all lambda functions in the same account? If so is there a way to limit this permission to the single lambda resource?

mndeveci (commented):
Thanks for the PR! I left some comments (most of them trying to understand the implementation).

samcli/lib/test_runner/base_template.j2
```diff
@@ -0,0 +1,68 @@
...
+                  PolicyDocument:
+                      Version: 2012-10-17
+                      Statement:
+                          - Sid: S3BucketAccess
+                            Effect: Allow
+                            Action:
+                                - s3:PutObject
+                                - s3:GetObject
+                            Resource: !Sub
+                                - arn:aws:s3:::${bucket}/*
```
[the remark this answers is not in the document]
bentvelj: Hm, I don't see any instances of that in [bucket policy examples](https://docs.aws.amazon.com/AmazonS3/latest/userguide/example-bucket-policies.html#example-bucket-policies-use-case-8). Also the [S3 ARN Format](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html) explicitly says the ARN of a bucket reduces to `arn:aws:s3:::bucket_name/key_name`. Doesn't look like it's provided, but we could add a `Principal` field such as `"Principal":{"AWS":"arn:aws:iam::${AWS::AccountId}:root"}`

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    str
+        The api id from the api arn.
+    """
+
+    return api_arn[api_arn.rindex("/") + 1 :]
+
+
+def _get_permissions_map() -> dict:
+    """
+    Returns a dictionary mapping a Regular Expression representing the arn of a specific resource to the AWS Resource Type and the default permissions generated for that resource.
```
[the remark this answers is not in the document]
bentvelj: Formatter doesn't seem to have an issue, `make black` and `make pr` pass. Probably ignores docstrings. I'll put it on two lines anyway for readability

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    dict
+        A ResourceTagMappingList, which contains a list of resource ARNs and tags associated with each.
+
+        NOTE: https://docs.aws.amazon.com/cli/latest/reference/resourcegroupstaggingapi/get-resources.html#output
+
+
+    """
+    try:
+        return boto3.client("resourcegroupstaggingapi").get_resources(TagFilters=tag_filters)["ResourceTagMappingList"]
+    except Exception as query_error:
```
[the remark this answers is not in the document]
bentvelj: Yep, added more specific exception catching

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+        "  Resource: {{arn}}\n"
+    )
+    mapping = _get_action_mapping(resource_arn)
+
+    if mapping is None:
+        return jinja2.Template(new_statement_template).render(action_list=[], arn=resource_arn).rstrip()
+
+    if mapping["Resource"] in ("AWS::ApiGateway::Api", "AWS::ApiGateway::RestApi"):
+        apiId = _extract_api_id_from_arn(resource_arn)
+        execute_api_arn = f"!Sub arn:${{AWS::Partition}}:execute-api:${{AWS::Region}}:${{AWS::AccountId}}:{apiId}/<STAGE>/GET/<RESOURCE_PATH>"
```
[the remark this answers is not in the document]
bentvelj: They're intended to be filled in by the customer, since there's no way for us to know which resource paths they want to grant. As for the stage, we don't know which stages go with which APIs

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: What are the advantages of this? Why does the client need `region` and `profile` settings? 

I'm also not sure what `updating user agent with SAM CLI` means

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    api_arn : str
+        The arn of the HTTP or REST api from which the api id is extracted.
+
+    Returns
+    -------
+    str
+        The api id from the api arn.
+    """
+
+    return api_arn[api_arn.rindex("/") + 1 :]
```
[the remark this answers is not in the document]
bentvelj: I don't see what input would cause this to fail, since this function is only executed if a string matches 
`^arn:{partition_ex}:apigateway:{region_ex}::\/restapis\/\w+$` 
or 
`^arn:{partition_ex}:apigateway:{region_ex}::\/apis\/\w+$`
meaning it will only extract from valid [HTTP or REST api arns](https://docs.aws.amazon.com/apigateway/latest/developerguide/arn-format-reference.html)

To validate, I'm pretty sure the api id is 10 alphanumeric characters (from this [example](https://docs.aws.amazon.com/apigateway/latest/developerguide/arn-format-reference.html)), but I cannot find it formally documented. I can make the regex more specific to reflect the length of the api id.

bentvelj:
Checks failed because of `The remote server returned an error: (429) Too Many Requests.`

If someone can restart checks, that'd be appreciated :)

tests/unit/lib/test_runner/test_template_generator.py
```diff
@@ -0,0 +1,347 @@
...
+            "            ClusterName: cloud-test-fargate-cluster\n",
+            "\n",
+            "    S3Bucket:\n",
+            "        Type: AWS::S3::Bucket\n",
+            "        Properties:\n",
+            "            BucketName: cloud-test-bucket-unique-name",
+        ]
+
+    @patch("samcli.lib.test_runner.test_runner_template_generator._query_tagging_api")
+    def test_lambda_function(self, query_tagging_api_patch):
```
acristin, line 95: Most of the test_<resource type> functions are nearly identical; it is probably worth factoring out the common core such that each test function sets the resource arn and expected statements and the common code constructs the expected and actual results, then compares them.

samcli/lib/test_runner/base_template.j2
```diff
@@ -0,0 +1,69 @@
...
+                      LogDriver: awslogs
+                      Options:
+                          awslogs-region: !Ref AWS::Region
+                          awslogs-group: !Ref LogGroup
+                          awslogs-stream-prefix: ecs
+
+    LogGroup:
+        Type: AWS::Logs::LogGroup
+        Properties:
+            LogGroupName: cloud-test-loggroup
```
acristin, line 59: In the unlikely event a customer already has such a log group, what should we do?

Can this value be parameterized or derived from other inputs?

samcli/lib/test_runner/base_template.j2
```diff
@@ -0,0 +1,69 @@
...
+
+    ECSCluster:
+        Type: AWS::ECS::Cluster
+        Properties:
+            ClusterName: cloud-test-fargate-cluster
```
acristin, line 64: In the unlikely event a customer already has such a cluster, what should we do?

Can this value be parameterized or derived from other inputs?

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,279 @@
...
+    dict:
+        A dictionary containing two keys: `'Resource'` and `'Action'`. `'Resource'` contains the AWS Resource for which the IAM actions list contained in `'Action'` maps to.
+
+        For example `{"Resource": "AWS::S3::Bucket", "Action": ["s3:PutObject", "s3:GetObject"]}`
+
+    """
+
+    permissions_map = _get_permissions_map()
+
+    for arn_exp, mapping in permissions_map.items():
```
acristin, line 114: nit: arn_exp -> arn_regex?

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,279 @@
...
+        if re.search(arn_exp, resource_arn) is not None:
+            LOG.info("Matched ARN `%s` to IAM actions %s", resource_arn, mapping["Action"])
+            return mapping
+
+    # If the supplied ARN is for a resource without any default actions supported
```
acristin, line 119: Since we log in the case where we find a match, is it worth logging when we fail to?

samcli/lib/test_runner/test_runner_template_generator.py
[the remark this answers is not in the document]
acristin: Can we use wildcards for stage, etc? Will we grant permissions we'd rather let the customer explicitly allow?

acristin (commented):
Thanks for the PR. As discussed offline, let's add a design doc with the PR so folks understand what we're building towards.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Edit: Andrew mentioned that it marks requests as coming from the SAM CLI instead of just boto3. I think this is good to have for consistency

samcli/lib/test_runner/base_template.j2
bentvelj: I feel like the easiest thing would be to just delete the LogGroup resource and replace `awslogs-group` with your own in the `TaskDefinition`, no?

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: I had actually done this, not sure why I removed it :/

samcli/lib/test_runner/test_runner_template_generator.py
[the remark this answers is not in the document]
bentvelj: Alex says it should be reasonable. I think it makes sense, to allow any `GET` request on a given API.

I'll change it :)

tests/unit/lib/test_runner/test_template_generator.py
bentvelj: Will do!

samcli/lib/test_runner/test_runner_template_generator.py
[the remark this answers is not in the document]
mndeveci: I see, my concern was around not getting an arn but something else (I am not sure how this parameter is populated).

samcli/lib/test_runner/test_runner_template_generator.py
mndeveci: So `"AWS::Lambda::Function"` is just a placeholder. If that is the case, can we use a different value here since using type as placeholder can lead to misunderstanding as if we are using the type value here?

samcli/lib/test_runner/test_runner_template_generator.py
mndeveci: If we will tie this to a command, then users might run their command with specific region and profile settings. We usually use `default` profile which is used by default as the name suggest. But if user invoke their command with another profile name using default might fail since there might not be a default profile or their default profile might not have sufficient permission to execute this.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,296 @@
...
+        The arn of the resource for which the IAM statement is generated.
+
+    Returns
+    -------
+    str:
+        An IAM Statement in the form of a YAML string.
+    """
+
+    # NOTE: The spacing within these strings is needed to keep the resulting YAML indentation valid
+    new_statement_template = (
```
mndeveci, line 143: Regarding to the comment here, since this template is programatically generated, we might not need to use jinja here. You could keep starting template as yaml, read it as dictionary into the memory, then make changes to its fields (or add new fields), and then serialize back as yaml. In that way you don't need to worry about indentation.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: The reason the types are there are at all is because APIGW IAM Statements have their `Resource` field as [something other than the API arn](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-control-access-using-iam-policies-to-invoke-api.html), so they require special logic.

When we match an arn to a regular expression, I thought it would be more clear to denote what resource type the ARN is supposed to match to. That way, we can more clearly detect when an ARN is a APIGW RestApi arn, or something else, by checking the `Resource` field of the matched object.

I get why this is confusing, because the information is really only used in the APIGW cases. I think a better refactor would be to just have a list of constants with clear names as the regular expressions.

Eg. `LAMBDA_FUNCTION_ARN_REGEX` = `....` and `APIGW_RESTAPI_REGEX`. Then, we when checking for a match, we can say "if the expression matched was `APIGW_RESTAPI_REGEX`, do this logic.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: This is similar to an approach I had tried, but the main issue is that we need generated IAM actions to be commented out in YAML. There isn't a way to comment out specific lines programmatically when converting from dictionary to yaml. This meant I had to use some funky string manipulation hacks to place `#`s in front of the lines I wanted.

Although having necessary whitespace isn't ideal, I don't think it is likely to change. The base template can have more resources added, but that will not change the indentation for the existing ones. 

The advantage of using jinja is that we can just render it, inject our string, and done. There's no flipping between yaml to json and back to yaml and then do some string manipulation to comment out generated IAM actions

samcli/lib/test_runner/test_runner_template_generator.py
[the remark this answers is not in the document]
bentvelj: This function is only called on strings that match the expression representing an APIGW RestApi or Api

acristin (approved):
Looks good to me.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -0,0 +1,271 @@
...
+    Raises
+    ------
+    botocore.ClientError
+        If the `get_resources` call fails.
+
+        See # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services
+    """
+    return boto_client_provider("resourcegroupstaggingapi").get_resources(TagFilters=tag_filters)[
+        "ResourceTagMappingList"
+    ]
```
mndeveci, line 157: I would recommend using `get` and probably with default values so that it will be easier to handle some of the edge cases.

If you use `["KeyName"]` directly, it will raise an error if the key is not there. If you use `get`, it will return `None` if the key is not there. You can also use `get('KeyName', DefaultValue)` as well for dict or list types.

```suggestion
    return boto_client_provider("resourcegroupstaggingapi").get_resources(TagFilters=tag_filters).get("ResourceTagMappingList", [])
```

mndeveci (approved):
Left one small comment there, approving it for now to unblock you.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: We probably would like to throw an exception if the `ResourceTagMappingList` is not present in the response, because that means the query delivered an unexpected result.

Still use `.get("ResourceTagMappingList")` and if the response is `None`, throw an exception

PR #4027, opened 2022-06-30
Refactor Event data structure

Leo10Gama:
N/A

With these added attributes, we can not only tell when Events occur, but also match Events to one command run via their `thread_id`, which is set to be the arbitrary ID of whatever thread is being used by the system.

The new attributes allow for queries to be run to match groups of Events by thread, which may be useful for SAM commands that employ multithreading.

Since the [thread_id](https://docs.python.org/3/library/threading.html#threading.get_ident) method picks an arbitrary integer to represent the thread ID, it takes no customer-specific data, and is safe for raw use as such. Otherwise, this would have the impact of adding two new fields to the `Event` data structure: `thread_id: int` and `timestamp: datetime`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -2,6 +2,7 @@
 Represents Events and their values.
 """
 
+from datetime import datetime
 from enum import Enum
 import logging
 import threading
@@ -52,6 +53,8 @@ class Event:
 
     event_name: EventName
     event_value: str  # Validated by EventType.get_accepted_values to never be an arbitrary string
+    thread_id = threading.get_ident()  # The thread ID; used to group Events from the same command run
+    timestamp = datetime.utcnow()
 
     def __init__(self, event_name: str, event_value: str):
         Event._verify_event(event_name, event_value)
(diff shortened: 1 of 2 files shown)


samcli/lib/telemetry/event.py
```diff
@@ -62,10 +65,20 @@ def __eq__(self, other):
...
+            f"timestamp={self.timestamp})"
+        )
 
     def to_json(self):
-        return {"event_name": self.event_name.value, "event_value": self.event_value}
+        return {
+            "event_name": self.event_name.value,
+            "event_value": self.event_value,
+            "thread_id": self.thread_id,
+            "timestamp": str(self.timestamp),
```
hawflau, line 80: Will different system generate different output? (e.g. different timezone offset?)

hawflau (changes requested):

samcli/lib/telemetry/event.py
Leo10Gama: No, since it uses [datetime.utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow), which returns the system time in UTC. As such, there will be no offset for timezones.

samcli/lib/telemetry/event.py
```diff
@@ -62,10 +65,20 @@ def __eq__(self, other):
...
+            f"timestamp={self.timestamp})"
+        )
 
     def to_json(self):
-        return {"event_name": self.event_name.value, "event_value": self.event_value}
+        return {
+            "event_name": self.event_name.value,
+            "event_value": self.event_value,
+            "thread_id": self.thread_id,
+            "timestamp": str(self.timestamp)[:-3],
```
mildaniel, line 80: I'm assuming trimming the last few characters is to make the timestamp format SQL compatible but can we drop a comment somewhere explaining this?

mildaniel (approved):

hawflau (approved):

PR #4037, opened 2022-07-06
Support GitLab OIDC for pipeline bootstrap

sidhujus:
Supports GitLab for OIDC provider during bootstrap flow

Currently only the GitHub Actions OIDC provider is implemented and because configuring the assume role policy differs slightly for each OIDC provider they need to be supported individually

adds click prompts/options to get the necessary information from the user before configuring the assume role policy for the pipeline execution role correctly

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -292,6 +320,13 @@ def do_cli(
                 "deployment-branch": deployment_branch,
             }
             pipeline_oidc_provider = GitHubOidcProvider(github_oidc_params, common_oidc_params, GITHUB_ACTIONS)
+        elif oidc_provider == GITLAB:
+            gitlab_oidc_params: dict = {
+                "gitlab-project": gitlab_project,
+                "gitlab-group": gitlab_group,
+                "deployment-branch": deployment_branch,
+            }
```
xazhao, line 328: Thinking if we could re-use the logic here. The `deployment_branch` is the same. `github-org` seems match to `gitlab-group` and `github_repo` seems match to `gitlab-project`. Maybe we could use general parameters name for all?

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -251,7 +273,15 @@ def _get_user_inputs(self) -> List[Tuple[str, Callable[[], None]]]:
...
-                        (f"Deployment branch:  {self.deployment_branch}", self._prompt_github_branch),
+                        (f"Deployment branch:  {self.deployment_branch}", self._prompt_deployment_branch),
+                    ]
+                )
+            elif self.oidc_provider == GITLAB:
+                inputs.extend(
+                    [
+                        (f"GitLab group: {self.gitlab_group}", self._prompt_gitlab_group),
+                        (f"GitLab project: {self.gitlab_project}", self._prompt_gitlab_project),
+                        (f"Deployment branch:  {self.deployment_branch}", self._prompt_deployment_branch),
```
hoffa, line 284:
```suggestion
                        (f"Deployment branch: {self.deployment_branch}", self._prompt_deployment_branch),
```

samcli/lib/pipeline/bootstrap/stage_resources.yaml
```diff
@@ -159,7 +159,7 @@ Resources:
                       },
                       "Action": "sts:AssumeRoleWithWebIdentity",
                       "Condition": {
-                        "ForAllValues:StringEquals": {
+                        "ForAllValues:StringLike": {
```
hoffa, line 162: Why the change?

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -292,6 +320,13 @@ def do_cli(
                 "deployment-branch": deployment_branch,
             }
             pipeline_oidc_provider = GitHubOidcProvider(github_oidc_params, common_oidc_params, GITHUB_ACTIONS)
+        elif oidc_provider == GITLAB:
+            gitlab_oidc_params: dict = {
```
hoffa, line 324: I wish there was stronger typing here, it's quite easy to mistype or spend time figuring out the correct names. Can't probably use `TypedDict` since it's 3.8+, but should be able to use the constants from `GitLabOidcProvider` and friends directly.

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -19,16 +19,17 @@
...
 class GuidedContext:
 
-    SUPPORTED_OIDC_PROVIDERS = {"1": GITHUB_ACTIONS}
-    OIDC_PROVIDER_NAME_MAPPINGS = {GITHUB_ACTIONS: "GitHub Actions"}
-    DEFAULT_OIDC_URLS = {GITHUB_ACTIONS: "https://token.actions.githubusercontent.com"}
-    DEFAULT_CLIENT_IDS = {GITHUB_ACTIONS: "sts.amazonaws.com"}
+    SUPPORTED_OIDC_PROVIDERS = {"1": GITHUB_ACTIONS, "2": GITLAB}
+    OIDC_PROVIDER_NAME_MAPPINGS = {GITHUB_ACTIONS: "GitHub Actions", GITLAB: "GitLab"}
+    DEFAULT_OIDC_URLS = {GITHUB_ACTIONS: "https://token.actions.githubusercontent.com", GITLAB: "https://gitlab.com"}
+    DEFAULT_CLIENT_IDS = {GITHUB_ACTIONS: "sts.amazonaws.com", GITLAB: "https://gitlab.com"}
```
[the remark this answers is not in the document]
sidhujus: GitHub Actions has an official action to assume the IAM roles and the client id that they have configured with that is "sts.amazonaws.com" GitLab instead just uses the instance URL for their client id.
https://docs.gitlab.com/ee/ci/cloud_services/aws/#add-the-identity-provider
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-the-identity-provider-to-aws

samcli/lib/pipeline/bootstrap/stage_resources.yaml
sidhujus: Its not really needed for GitHub or GitLab because its pretty easy to build the entire sub claim in the OIDC token, but for Bitbucket the claim changes weather or not a step is assigned to a deployment environment (our templates dont do this) and it also contains a UUID for the pipeline step which I couldn't find a way to get so using StringLike becomes necessary for that.

https://support.atlassian.com/bitbucket-cloud/docs/deploy-on-aws-using-bitbucket-pipelines-openid-connect/#Using-claims-in-ID-tokens-to-limit-access-to-the-IAM-role-in-AWS

samcli/commands/pipeline/bootstrap/cli.py
sidhujus: GitHub and GitLab are close enough to each other in information that we need that it would probably be fine for those two, but with Bitbucket we will need only a single new parameter, instead of the three that GitHub and GitLab need, the repository UUID which is a bitbucket specific value so I don't think it would be possible to generalize the parameters once we add Bitbucket and if in the future we decide to add any other providers. I think keeping them all separate helps reinforce that to anybody whos reading the code for the first time and is unfamiliar with the concepts

hoffa (approved):

PR #4039, opened 2022-07-08
Add send_events method to EventTracker

Leo10Gama:
N/A

Previously, Event information could only be sent through Metric's `track_command` decorator. This change looks to decouple the functionality, allowing Events to be sent to the backend of Telemetry independently of the decorator.

With a new `send_events()` method, Events will be sent to the Telemetry backend whenever requested. This method is called in place of the existing format used in the `track_command` decorator. In addition, new functionality is given to the `track_event` method, and when there exceeds a certain number of tracked events, their information is sent automatically via Telemetry, and the existing Events are cleared from the list.

Now, `metricSpecificAttributes` for the `commandRun` metrics will no longer contain `events` as a field, as they will be sent under a different metric name. Additionally, the unit test for testing that the list of Events (in `..metric.py`) have been removed, as the functionality of `send_events` is checked separately, and is now decoupled from the decorator, thus its testing in this situation is no longer necessary.
In addition, the `Event.timestamp: datetime.datetime` attribute has been refactored, and is now `Event.time_stamp: str`, so as to not use the same keyword as the SQL backend, as well as allow for the correct cropping of the timestamp to allow for SQL casting.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -166,7 +166,9 @@ class EventTracker:
 
         try:
             metric = Metric("events")
-            metric.add_data("metricSpecificAttributes", [e.to_json() for e in EventTracker.get_tracked_events()])
+            msa = {}
+            msa['events'] = [e.to_json() for e in EventTracker.get_tracked_events()]
+            metric.add_data("metricSpecificAttributes", msa)
             telemetry.emit(metric)
             EventTracker.clear_trackers()
         except RuntimeError:
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -167,7 +167,7 @@ class EventTracker:
         try:
             metric = Metric("events")
             msa = {}
-            msa['events'] = [e.to_json() for e in EventTracker.get_tracked_events()]
+            msa["events"] = [e.to_json() for e in EventTracker.get_tracked_events()]
             metric.add_data("metricSpecificAttributes", msa)
             telemetry.emit(metric)
             EventTracker.clear_trackers()
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -71,7 +71,7 @@ class Event:
             f"Event(event_name={self.event_name.value}, "
             f"event_value={self.event_value}, "
             f"thread_id={self.thread_id}, "
-            f"timestamp={self.time_stamp})"
+            f"time_stamp={self.time_stamp})"
         )
 
     def to_json(self):
@@ -79,7 +79,7 @@ class Event:
             "event_name": self.event_name.value,
             "event_value": self.event_value,
             "thread_id": self.thread_id,
-            "timestamp": self.time_stamp,
+            "time_stamp": self.time_stamp,
         }
 
     @staticmethod
@@ -157,10 +157,10 @@ class EventTracker:
     @staticmethod
     def send_events():
         """Sends the current list of events via Telemetry."""
+        from samcli.lib.telemetry.metric import Metric  # pylint: disable=cyclic-import
+
         if not EventTracker.get_tracked_events():  # Don't do anything if there are no events to send
             return
-
-        from samcli.lib.telemetry.metric import Metric  # creates circular import otherwise
 
         telemetry = Telemetry()
 
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -165,15 +165,12 @@ class EventTracker:
 
             telemetry = Telemetry()
 
-            try:
-                metric = Metric("events")
-                msa = {}
-                msa["events"] = [e.to_json() for e in EventTracker._events]
-                metric.add_data("metricSpecificAttributes", msa)
-                telemetry.emit(metric)
-                EventTracker._events = []  # Manual clear_trackers() since we're within the lock
-            except RuntimeError:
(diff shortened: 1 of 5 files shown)


tests/integration/telemetry/test_experimental_metric.py
```diff
@@ -152,8 +150,11 @@ def test_must_send_cdk_project_type_metrics(self):
             process.communicate()
 
             all_requests = server.get_all_requests()
-            self.assertEqual(1, len(all_requests), "Command run metric must be sent")
-            request = all_requests[0]
+            self.assertGreaterEqual(len(all_requests), 1, "Command run metric must be sent")
+            request = None
+            for req in all_requests:
+                print(req["data"])
+                request = req if "commandRun" in req["data"]["metrics"][0].keys() else all_requests[-1]
```
hawflau, line 157: Are we only looking for the `req` in `all_requests` with the `commandRun` metric? I don't understand why we need `else all_requests[-1]`. Looks like `request` is assigned as `all_requests[-1]` regardless

hawflau (changes requested):

tests/integration/telemetry/test_experimental_metric.py
Leo10Gama: Yes we only want the `commandRun` metric for this test. Initially I wanted to check which of the metrics had `commandRun` in the name, so I tried iterating through all of them. However on second look, I can see that the `commandRun` metric will be guaranteed as the last metric thanks to the ordering of `EventTracker.send_events(); telemetry.emit(metric)`. Will fix this in the next commit. Good catch!

samcli/lib/telemetry/event.py
```diff
@@ -145,6 +154,26 @@ def clear_trackers():
...
+
+        try:
+            metric = Metric("events")
+            msa = {}
+            msa["events"] = [e.to_json() for e in EventTracker.get_tracked_events()]
+            metric.add_data("metricSpecificAttributes", msa)
+            telemetry.emit(metric)
+            EventTracker.clear_trackers()
+        except RuntimeError:
+            LOG.debug("Unable to find Click Context for getting session_id.")
```
torresxb1, line 175: discussed with Leo in a huddle and we decided we should probably do all this while holding a lock in order to avoid race conditions (e.g. sending duplicate  events, etc.), as this function is called without holding a lock

samcli/lib/telemetry/event.py
```diff
@@ -145,6 +154,27 @@ def clear_trackers():
...
+
+            try:
+                metric = Metric("events")
+                msa = {}
+                msa["events"] = [e.to_json() for e in EventTracker._events]
+                metric.add_data("metricSpecificAttributes", msa)
+                telemetry.emit(metric)
+                EventTracker._events = []  # Manual clear_trackers() since we're within the lock
+            except RuntimeError:
+                LOG.debug("Unable to find Click Context for getting session_id.")
```
torresxb1, line 176: It seems we copy pasted this from some metric methods that use click context. Since we're not using it here, I believe we actually don't need the try and catch

tests/unit/lib/telemetry/test_event.py
```diff
@@ -93,3 +93,83 @@ def test_track_event(self, event_mock, lock_mock):
         self.assertEqual(len(EventTracker._events), 0)
         lock_mock.__enter__.assert_called()  # Lock should have been accessed
         lock_mock.__exit__.assert_called()
+
+    @patch("samcli.lib.telemetry.event.Telemetry")
+    def test_events_get_sent(self, telemetry_mock):
+        # Create fake emit to capture tracked events
+        dummy_telemetry = Mock()
+        tracked_events = []
```
torresxb1, line 101: nit: could we rename to `emitted_metrics`? (if I understand correctly)

tests/unit/lib/telemetry/test_event.py
```diff
@@ -93,3 +93,83 @@ def test_track_event(self, event_mock, lock_mock):
...
+                    {
+                        "event_name": "Test",
+                        "event_value": "SomeValue",
+                        "thread_id": ANY,
+                        "time_stamp": ANY,
+                    }
+                ]
+            },
+        }
+        print(metric_data)
```
torresxb1, line 144: probably left from testing 😅 ?

torresxb1 (changes requested):

samcli/lib/telemetry/event.py
Leo10Gama: Right, just double checked and none of the methods in the try-catch throw anything, so it should be safe to remove!

tests/unit/lib/telemetry/test_event.py
Leo10Gama: `emitted_events` may be more accurate, but yes that would probably be more specific and correct

tests/unit/lib/telemetry/test_event.py
Leo10Gama: Yup, not necessary to be there lol!

tests/unit/lib/telemetry/test_event.py
torresxb1: but doesn't `emit` get called with a `Metric`? 🤔 
Maybe `emitted_event_metrics`? lol

torresxb1:
seems like some unit tests are failing btw

Issue #3899, opened 2022-05-25
Bug: sam cli does not throw error and exit when --config-file does not exist


edify42:
### Description:
Hello maintainers 👋 

Currently when deploying via `sam deploy --config-file <file>`, if `<file>` does not exist sam ignores that without throwing an error.

While it's unlikely this has caused issues for anyone (as the stack-name being missing throws an error), it does make the debugging experience strange for the first time you see the error and not realise that you might have passed in the wrong location to the config file.

### Steps to reproduce:
Run the below anywhere:

```bash
sam deploy --config-file file-that-does-not-exist.toml
```

### Observed result:
```bash
Usage: sam deploy [OPTIONS]
Try 'sam deploy -h' for help.

Error: Missing option '--stack-name', 'sam deploy --guided' can be used to provide and save needed parameters for future deploys.
```

### Expected result:
I'd expect the error be thrown when trying to read / validate the toml file

```bash
Usage: sam deploy [OPTIONS]
Try 'sam deploy -h' for help.

Error: File <file> could not be found

OR

Error: File <file> could not be read as permissions on the file/directory prevent it being read by the current user
```

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macos / linux
2. `sam --version`: SAM CLI, version 1.46.0
3. AWS region: ap-southeast-2

jlgoolsbee:
I'm unsure if I should open a new bug or just comment on this one; I left a comment on #4044 but I'm also unsure if that's "visible" enough:

I believe this change has broken deploys that use a relative path to resolve a custom location for `samconfig.toml`; passing an absolute path works – so there is a workaround – but it isn't optimal.

Now busted (worked previously):

```
$ sam deploy --template-file "$app_name/template.yaml" --config-file ../samconfig.toml --config-env $env
Error: Config file ../samconfig.toml does not exist or could not be read
```

... but works when full path is provided:

```
$ sam deploy --template-file "$app_name/template.yaml" --config-file /full/path/to/samconfig.toml --config-env $env
File with same data already exists at [...], skipping upload

	Deploying with following values
	===============================
... etc.
```

mildaniel:
A fix has been released in version 1.56.0 on 09/01. Please re-open if it persists.

PR #4044, opened 2022-07-10
Error out and exit when --config-file does not exist

cibinmathew:

Throw error and exit when --config-file does not exist
Fixes: https://github.com/aws/aws-sam-cli/issues/3899

Currently it throws an irrelevant error making it hard to debug.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -129,12 +129,11 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
     param.default = None
     config_env_name = ctx.params.get("config_env") or DEFAULT_ENV
 
-    config_file = ctx.params.get("config_file")
+    config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
     if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file).is_file():
         error_msg = f"Config file {config_file} does not exist or could not be read"
         LOG.debug(error_msg)
         raise ConfigException(error_msg)
-    config_file = config_file or DEFAULT_CONFIG_FILE_NAME
 
     config_dir = getattr(ctx, "samconfig_dir", None) or os.getcwd()
     # If --config-file is an absolute path, use it, if not, start from config_dir
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -128,7 +128,14 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
     cmd_name = cmd_name or ctx.info_name
     param.default = None
     config_env_name = ctx.params.get("config_env") or DEFAULT_ENV
-    config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
+
+    config_file = ctx.params.get("config_file")
+    if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file).is_file():
+        error_msg = f"Config file {config_file} does not exist or could not be read"
+        LOG.debug(error_msg)
+        raise ConfigException(error_msg)
+    config_file = config_file or DEFAULT_CONFIG_FILE_NAME
+
     config_dir = getattr(ctx, "samconfig_dir", None) or os.getcwd()
     # If --config-file is an absolute path, use it, if not, start from config_dir
     config_file_name = config_file if os.path.isabs(config_file) else os.path.join(config_dir, config_file)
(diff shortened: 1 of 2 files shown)


mndeveci (commented):
Thanks for your contribution. I've verified following use cases and it seems this is better experience from customer's perspective.

- I've tried with an application with no default `samconfig.toml` settings, I got new error instead of `Missing option` error.
- I've tried with a default settings in `samconfig.toml`, and I got new error instead of `Missing option` one.

Can you also add some tests to verify this behavior which will also make sure that we won't remove this by mistake?

cibinmathew:
> Thanks for your contribution. I've verified following use cases and it seems this is better experience from customer's perspective.
> 
> * I've tried with an application with no default `samconfig.toml` settings, I got new error instead of `Missing option` error.
> * I've tried with a default settings in `samconfig.toml`, and I got new error instead of `Missing option` one.
> 
> Can you also add some tests to verify this behavior which will also make sure that we won't remove this by mistake?

Can you pls provide the full commands used for both scenarios. Also pls add the settings used in second scenario.


Currently with this pr, only custom config file is handled. An observation from the code flow:
If this check is enabled for default config file, `sam deploy --help` will error out and cannot be run from any dir (based on the current code, these checks happen earlier than code for rendering help text). I will still dig deep into the code to see what needs to be changed.

mndeveci:
> 



> > Thanks for your contribution. I've verified following use cases and it seems this is better experience from customer's perspective.
> > 
> > * I've tried with an application with no default `samconfig.toml` settings, I got new error instead of `Missing option` error.
> > * I've tried with a default settings in `samconfig.toml`, and I got new error instead of `Missing option` one.
> > 
> > Can you also add some tests to verify this behavior which will also make sure that we won't remove this by mistake?
> 
> Can you pls provide the full commands used for both scenarios. Also pls add the settings used in second scenario.
> 
> Currently with this pr, only custom config file is handled. An observation from the code flow: If this check is enabled for default config file, `sam deploy --help` will error out and cannot be run from any dir (based on the current code, these checks happen earlier than code for rendering help text). I will still dig deep into the code to see what needs to be changed.

I've tried with following command;
```
sam deploy --config-file this-file-does-not-exist.toml
```

with different environments.
- For the first one, I tried to run it in a folder where no `samconfig.toml` exists. Before your change, it was giving an error that `--stack-name` is required. But with your change, it is now giving an error which states that the settings file is missing.
- For the second one, I tried to run it in a folder where `samconfig.toml` file exists. And results were like the first one.

The behavior looks like correct to me, but I will do another round of check later.

samcli/cli/cli_config_file.py
```diff
@@ -128,7 +128,14 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
     cmd_name = cmd_name or ctx.info_name
     param.default = None
     config_env_name = ctx.params.get("config_env") or DEFAULT_ENV
-    config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
+
+    config_file = ctx.params.get("config_file")
```
qingchm, line 132:
```suggestion
    config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
```

samcli/cli/cli_config_file.py
```diff
@@ -128,7 +128,14 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
...
+    if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file).is_file():
+        error_msg = f"Config file {config_file} does not exist or could not be read"
+        LOG.debug(error_msg)
+        raise ConfigException(error_msg)
+    config_file = config_file or DEFAULT_CONFIG_FILE_NAME
```
qingchm, line 137:
```suggestion
```

jfuss (approved):

hnnasit (approved):
LGTM!

jlgoolsbee:
I believe this change has broken deploys that use a relative path to resolve a custom location for `samconfig.toml`; passing an absolute path works – so there is a workaround – but it isn't optimal.

Now busted (worked previously):

```
$ sam deploy --template-file "$app_name/template.yaml" --config-file ../samconfig.toml --config-env $env
Error: Config file ../samconfig.toml does not exist or could not be read
```

... but works when full path is provided:

```
$ sam deploy --template-file "$app_name/template.yaml" --config-file /full/path/to/samconfig.toml --config-env $env
File with same data already exists at [...], skipping upload

	Deploying with following values
	===============================
... etc.
```

PR #4045, opened 2022-07-11
feat: Added Transform Command to sam-cli

Crvanetten15:
### Which issue does this change fix and why is this change necessary?
Currently when working with SAM templates and using AWS SAM CLI, getting a transformed version for the file is only able to be done using the command `sam validate --template {name_of_template} --debug`. This output also will contain unwanted information for a user who only wants a transformed version of the template to be returned or saved to a file. The validate command also requires AWS credentials to be present when running a template transformation. Preferably users would like to be able to receive a transformed template locally or without AWS credentials present. 

The creation of this `transform` will remove the unwanted excess information that the command `sam validate --template {name_of_template} --debug` returns to the command line, and **only** return the template after transformation. This command will also allow a user to locally transform the template, which is done by removing the need for AWS credentials which was previously required.

No side effects have been shown as `transform` utilizes previously integrated libraries and a similar functionality to that of `validate`. All files and implementations are newly added, while utilizing previously imported libraries that are present elsewhere in the repository.

### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


Crvanetten15:
@jfuss Will do. Currently adding a Github Issue, look forward from hearing from you and your team.

PR #4052, opened 2022-07-13
fix: Don't use deps dir for non-cached builds

mildaniel:
After accelerate GA, we stopped gating the creating of a dependencies dir with a feature flag which means that now a dependencies dir is used for all builds where applicable. This means that the downstream lambda builders workflow performs an accelerate-specific workflow even when it shouldn't. 

Don't build with dependencies dir if cached is False.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -186,7 +186,7 @@ class ApplicationBuilder:
         """
         build_graph = self._get_build_graph(self._container_env_var, self._container_env_var_file)
         build_strategy: BuildStrategy = DefaultBuildStrategy(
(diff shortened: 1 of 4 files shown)


qingchm (approved):
LGTM, wondering if the appbuilder parameter should just be called cached or using_dependencies_dir

mndeveci (approved):

mildaniel:
> LGTM, wondering if the appbuilder parameter should just be called cached or using_dependencies_dir

Happy to switch it to cached. Might be better if we decide to use it for other things in the future.

PR #4061, opened 2022-07-15
Upgrade encryption algorithm from SHA-1 to SHA-256

Leo10Gama:
N/A

As discussed, the current SHA-1 hashing of project metadata info has proven not cryptographically secure enough, and thus should be upgraded to SHA-256.

Any and all prior uses of the `uuid.uuid5()` method, which uses SHA-1, have been replaced by `hashlib.sha256()` and its relevant encryption methods.

Rather than sending a string, the encryption methods will return bytes. This shouldn't have significant impact on existing structures and frameworks in the codebase.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/project_metadata.py
+++ b/samcli/lib/telemetry/project_metadata.py
@@ -11,14 +11,14 @@ from typing import List, Optional
 from samcli.cli.global_config import GlobalConfig
 
 
-def get_git_remote_origin_url() -> Optional[bytes]:
+def get_git_remote_origin_url() -> Optional[str]:
     """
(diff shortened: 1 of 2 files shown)


hawflau (approved):
LGTM!

mildaniel (approved):

sriram-mv:
Even though this is merged, I wanted to highlight that we are changing the "hash" algorithm, even though the title says changing the encryption algorithm. Hashing is one way, whereas encryption is 2 way (as in decryption is possible).

Leo10Gama:
Thanks for the correction!

PR #4062, opened 2022-07-15
feat: Enable source maps if either Sourcemap: true or --enable-source-maps is provided

lucashuy:
Related to lambda builders change: https://github.com/aws/aws-lambda-builders/pull/375

Previously, both `Sourcemap: true` and `NODE_OPTIONS: --enable-source-maps` needed to be provided to enable source map usage. This is redundant since explicitly defining one would imply source map usage.

This change allows users to specify either `NODE_OPTIONS: --enable-source-maps` under environment variables or `Sourcemap: true` under esbuild build properties and have source maps enabled and added to their functions.

`NODE_OPTIONS` can be defined at either the Globals level or the Function level.

This change will modify the final template file in `sam build` by adding either `Sourcemap` or `NODE_OPTIONS` at the function level.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -313,7 +313,8 @@ class BuildContext:
 
     def _enable_source_maps(self):
         """
-        Appends the --enable-source-maps NODE_OPTIONS if Sourcemap is set to true and vice versa
+        Appends ``NODE_OPTIONS: --enable-source-maps``, if Sourcemap is set to true
+        and sets Sourcemap to true if ``NODE_OPTIONS: --enable-source-maps`` is provided.
         """
         using_source_maps = False
         incorrect_node_option = False
@@ -331,7 +332,7 @@ class BuildContext:
                 source_map = build_properties.get("Sourcemap", None)
 
                 if source_map and not node_option_set:
-                    LOG.debug("Sourcemap set without --enable-source-maps, adding")
+                    LOG.debug("Sourcemap set without --enable-source-maps, adding Sourcemap to Metadata BuildProperties")
 
                     resource.setdefault("Properties", {})
                     resource["Properties"].setdefault("Environment", {})
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -1,6 +1,7 @@
 """
 Context object used by build command
 """
+from copy import deepcopy
 import logging
 import os
 import pathlib
@@ -265,6 +266,9 @@ class BuildContext:
                         stack, self._stack_name, self.build_dir, modified_template, build_result
                     )
                     modified_template = nested_stack_manager.generate_auto_dependency_layer_stack()
+
+                modified_template = self._enable_source_maps(modified_template)
+
                 move_template(stack.location, output_template_path, modified_template)
 
             click.secho("\nBuild Succeeded", fg="green")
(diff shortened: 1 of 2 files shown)


samcli/commands/build/build_context.py
```diff
@@ -307,6 +311,99 @@ def run(self):
             wrapped_from = deep_wrap if deep_wrap else ex.__class__.__name__
             raise UserException(str(ex), wrapped_from=wrapped_from) from ex
 
+    def _enable_source_maps(self):
+        """
+        Appends the --enable-source-maps NODE_OPTIONS if Sourcemap is set to true and vice versa
```
sriram-mv, line 316: Let's be clear on the docstring itself, to showcase both scenarios.

samcli/commands/build/build_context.py
```diff
@@ -307,6 +311,99 @@ def run(self):
...
+                    resource.setdefault("Metadata", {})
+                    resource["Metadata"].setdefault("BuildProperties", {})
+                    resource["Metadata"]["BuildProperties"]["Sourcemap"] = True
+
+                    using_source_maps = True
+
+        if using_source_maps:
+            self._warn_using_source_maps()
+
+        if incorrect_node_option:
```
sriram-mv, line 364: nit: change incorrect to invalid?, because its just missing a flag right?

samcli/commands/build/build_context.py
```diff
@@ -307,6 +311,99 @@ def run(self):
...
+                    else:
+                        resource["Properties"]["Environment"]["Variables"]["NODE_OPTIONS"] = " ".join(
+                            [existing_options, "--enable-source-maps"]
+                        )
+
+                    using_source_maps = True
+
+                # check if --enable-source-map is provided and append Sourcemap: true if it is not set
+                if source_map is None and node_option_set:
+                    LOG.debug("--enable-source-maps set without Sourcemap, adding")
```
sriram-mv, line 353: .. adding SourceMap to Metadata BuildProperties.

samcli/commands/build/build_context.py
lucashuy: Its actually when the user provides an invalid NODE_OPTIONS, the customer is able to provide an array of strings in the template file when we are expecting a space delimited string. Still, "invalid" could be used in place of "incorrect" too.

samcli/commands/build/build_context.py
```diff
@@ -307,6 +311,99 @@ def run(self):
...
+                if metadata.get("BuildMethod", "") != "esbuild":
+                    continue
+
+                node_option_set = self._is_node_option_set(resource)
+
+                # check if Sourcemap is provided and append --enable-source-map if not set
+                build_properties = metadata.get("BuildProperties", {})
+                source_map = build_properties.get("Sourcemap", None)
+
+                if source_map and not node_option_set:
```
qingchm, line 333: Based on the logic here it doesn't seem to have special handling to the scenario where node_option_set is true and the build_properties.get("Sourcemap") is an explicit value False, which I feel we might want to handle differently from it being None? Open to any reason for this self conflict input should not be a concern

lucashuy: Yup, you're correct that it does not handle the case where NODE_OPTIONS is set and Sourcemap is false. In this case, no source map is generated, so the environment variable to enable source maps will not do anything even though it is set. The only potential concern the performance hit with source maps, I'm not quite sure if the performance is degraded with a source map file, or if its just solely by using the enable source map environment variable.

We can modify this anyways to not include that option to reduce unnecessary parameters passed. Will update this if there are no concerns with this.

sriram-mv: Unsure of the performance concerns here, but to @qingchm 's question is there a change in behavior from before this PR and after? If there is a conflict, we can raise that as LOG warning that there is a conflict.

lucashuy: (not @qingchm answering) Before this PR, setting Sourcemaps to false would effectively disable source map usage since no map file is generated, even if the env variable was provided, this was expected behaviour. The behaviour remains the same after the PR. The `Sourcemap: false` option is basically just a way to opt out of source maps if it was defined at the Global level.

sriram-mv (approved):

lucashuy:
Currently blocked awaiting further discussion/planning.

samcli/commands/build/build_context.py
```diff
@@ -307,6 +311,105 @@ def run(self):
...
+            "\nNODE_OPTIONS is not a string! As a result, the NODE_OPTIONS environment variable will "
+            "not be set correctly, please make sure it is a string. "
+            "Visit https://nodejs.org/api/cli.html#node_optionsoptions for more details.\n",
+            fg="yellow",
+        )
+
+    @staticmethod
+    def _warn_using_source_maps():
+        click.secho(
+            "\nYou are using source maps, note that this comes with a performance hit!"
```
mildaniel, line 407: If they try to remove just the env var it will still be set next time if the source maps are still set in the metadata, right? Should we clarify that it needs to be removed from both spots?

lucashuy: Nice catch, I'll update it to reflect this change

mildaniel (approved):

Issue #4034, opened 2022-07-06
Bug: Cannot execute CDK application locally ("sam local invoke" cannot work)


sinokuma:
### Description:

I tried to execute a Lambda function defined by CDK locally, but an error occurred.

### Steps to reproduce:

$ cdk --version
2.29.0 (build 47d7ec4)

Execute "sam local invoke" against the CloudFormation template generated by "cdk synth".
The sample code is here.

```TypeScript
import * as cdk from "aws-cdk-lib";
import { Stack, StackProps, Duration } from "aws-cdk-lib";
import { Construct } from "constructs";
import * as cloudfront from "aws-cdk-lib/aws-cloudfront";
import * as origins from "aws-cdk-lib/aws-cloudfront-origins";
import * as s3 from "aws-cdk-lib/aws-s3";
import * as acm from "aws-cdk-lib/aws-certificatemanager";
import * as lambda from "aws-cdk-lib/aws-lambda";
import * as path from "path";

export class TestStack extends Stack {
  constructor(scope: Construct, id: string, props?: StackProps) {
    super(scope, id, props);

    const originBucketArn = "hogehoge";
    const acmArn = "fugafuga";
    const domainName = "example.com";

    const viewerRequestFunction = new cloudfront.experimental.EdgeFunction(
      this,
      "ViewerRequestFunction",
      {
        runtime: lambda.Runtime.NODEJS_14_X,
        handler: "app.lambdaHandler",
        code: lambda.Code.fromAsset(
          path.join(__dirname, "./src/viewer-request-function")
        ),
        memorySize: 128,
        timeout: Duration.seconds(5),
      }
    );

    const originResponseFunction = new cloudfront.experimental.EdgeFunction(
      this,
      "OriginResponseFunction",
      {
        runtime: lambda.Runtime.NODEJS_14_X,
        handler: "app.lambdaHandler",
        code: lambda.Code.fromAsset(
          path.join(__dirname, "./src/origin-response-function")
        ),
        memorySize: 512,
        timeout: Duration.seconds(10),
      }
    );

    const customCachePolicy = new cloudfront.CachePolicy(
      this,
      "TestCachePolicy",
      {
        cachePolicyName: "TestCachingOptimized",
        comment: "Custom policy optimized for xxxxx",
        defaultTtl: Duration.days(30),
        minTtl: Duration.days(30),
        maxTtl: Duration.days(30),
        queryStringBehavior: cloudfront.CacheQueryStringBehavior.all(),
[...]

PR #4064, opened 2022-07-16
fix: duplicate CDK resource ID

WinterYukky:

#4034 
https://github.com/aws/aws-cdk/issues/21134

Currently, SAM CLI extract SamResourceId from last second string of `aws:cdk:path` value.
But, CDK possibly generate `aws:cdk:path` like next.

- Stack/aaa/Fn/Resource -> SamResouceId: Fn
- Stack/bbb/Fn/Resource -> SamResouceId: Fn
- Stack/ccc/Fn/Resource -> SamResouceId: Fn

In such cases, the SAM CLI considers these resources to have the same SamResouceId (i.e., Fn) and treats only the last lambda function found as an executable lambda function. In other words, it will generate lambda functions that are not found.

See https://github.com/aws/aws-sam-cli/pull/4064#issuecomment-1187724150

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss:
Left a comment in https://github.com/aws/aws-sam-cli/issues/4034 but putting the important details here too for visibility with the team. 

We need to understand if this is the right approach. Combining the two fields seems reasonable but I need to do a deeper check in what we support now and why before I can understand the right approach (allowing the logical id and the cdk path, using "_" as a delimiter, etc).

Again thanks to @WinterYukky for catching this and @sinokuma for the initial report.

WinterYukky:
Thanks @jfuss for checking the PR.
I was wondering why you use part of the cdk path (i.e Fn) for the ID value, but your explanation helped me understand.
This PR works by using underscores as delimiters, but may not be a good solution for users. It is better to use slashes as delimiters than to use underscores, at least.

Currently, SAM CLI find functions by `function_id` `name` `functionname`.
https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/providers/sam_function_provider.py#L110-L112

Consider the following example.
```json
{
  "Resources": {
    "aaaFn67B877B1": {
      "Type": "AWS::Lambda::Function",
      "Properties": {},
      "Metadata": {
        "aws:cdk:path": "Stack/aaa/Fn/Resource",
        "aws:asset:path": "/workspaces/cdk-sam-local/lib/handler",
        "aws:asset:is-bundled": false,
        "aws:asset:property": "Code"
      }
    },
    "bbbFn9F2D7876": {
      "Type": "AWS::Lambda::Function",
      "Properties": {},
      "Metadata": {
        "aws:cdk:path": "Stack/bbb/Fn/Resource",
        "aws:asset:path": "/workspaces/cdk-sam-local/lib/handler",
        "aws:asset:is-bundled": false,
        "aws:asset:property": "Code"
      }
    },
    "cccFn256E40E6": {
      "Type": "AWS::Lambda::Function",
      "Properties": {},
      "Metadata": {
        "aws:cdk:path": "Stack/ccc/Fn/Resource",
        "aws:asset:path": "/workspaces/cdk-sam-local/lib/handler",
        "aws:asset:is-bundled": false,
        "aws:asset:property": "Code"
      }
    },
    "CDKMetadata": {
      "Type": "AWS::CDK::Metadata",
      "Properties": {
        "Analytics": "v2:deflate64:H4sIAAAAAAAA/zWNXQqDMBCEz+J73FqFvtdCD2APIGvcyvqTgJu0DyF3b6IUFr7ZYZipoamhKvArpR6XcuUBwsuhXlSy+rDiNowI4emNdmyNerzNX0fFuEHo7ErZzoxKmh5FyAncM9IPofV6IdeikDpljp8qRnXk0uTEZjp6SKzfdeoydiSY5fK53iBdVczCXO7eON4IupM/SkZ1q8EAAAA="
      },
      "Metadata": {
        "aws:cdk:path": "Stack/CDKMetadata/Default"
      }
    }
  }
}
```

In this case, the SAM CLI resolves the functions into the following objects.

| LogicalId | function_id | name | functionname |
|-|-|-|-|
| aaaFn67B877B1 | Fn | aaaFn67B877B1 | aaaFn67B877B1 |
| bbbFn9F2D7876 | Fn | bbbFn9F2D7876 | bbbFn9F2D7876 |
| cccFn256E40E6 | Fn | cccFn256E40E6 | cccFn256E40E6 |

How about changing it as follows.

| LogicalId | function_id | name | functionname |
|-|-|-|-|
| aaaFn67B877B1 | aaaFn67B877B1 | Fn | aaaFn67B877B1 |
| bbbFn9F2D7876 | bbbFn9F2D7876 | Fn | bbbFn9F2D7876 |
| cccFn256E40E6 | cccFn256E40E6 | Fn | cccFn256E40E6 |

When change to the proposed solution, then I think SAM CLI allows use these lambda functions. When call by aaaFn67B877B1 (i.e LogicalId) then invoke the found function and when call by Fn then print warning log "Multiple functions found with keyword {name}! Function ... ".

jfuss:
@WinterYukky I chatted with the team. Instead of concatenating parts of the `cdk:path`, we want to ensure that the LogicalId is callable. Concatenating, still requires the customer to understand how the `cdk:path` works and map it to their code. Instead, we feel that supporting the LogicalId should always work. We know this does not solve the how we handle the `cdk:path` but want to start from the LogicalId as a first step.

Would you be willing to update this PR to support this?

WinterYukky:
@jfuss Of course! I'll try it.

WinterYukky (commented):
@jfuss 
From the following results, the behavior appears to be good. Could you check the PR?

```bash
vscode ➜ /workspaces/aws-sam-cli (fix/duplicate-CDK-ID ✗) $ samdev local invoke Fn -t cdk.out/ExampleStack.template.json 

        SAM CLI now collects telemetry to better understand customer needs.

        You can OPT OUT and disable telemetry collection by setting the
        environment variable SAM_CLI_TELEMETRY=0 in your shell.
        Thanks for your help!

        Learn More: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-telemetry.html

Multiple functions found with keyword Fn! Function Example1FnCBC0C9FD will be invoked! If it's not the function you are going to invoke, please choose one of them from below:
Example1FnCBC0C9FD
Example2FnFBA58E3D
Invoking app.handler (nodejs16.x)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-nodejs16.x:rapid-1.53.0-x86_64.

Mounting /workspaces/aws-sam-cli as /var/task:ro,delegated inside runtime container
START RequestId: a5b9b93a-d047-4901-924c-4405e7fb338a Version: $LATEST
{"statusCode":200,"body":"{\"message\":\"Hello from Lambda!\"}"}END RequestId: a5b9b93a-d047-4901-924c-4405e7fb338a
REPORT RequestId: a5b9b93a-d047-4901-924c-4405e7fb338a  Init Duration: 0.20 ms  Duration: 219.29 ms   Billed Duration: 220 ms  Memory Size: 128 MB     Max Memory Used: 128 MB
```

<details><summary>Used template</summary><div>

```json
{
    "Resources": {
     "Example1FnServiceRole6C38A161": {
      "Type": "AWS::IAM::Role",
      "Properties": {
       "AssumeRolePolicyDocument": {
        "Statement": [
         {
          "Action": "sts:AssumeRole",
          "Effect": "Allow",
          "Principal": {
           "Service": "lambda.amazonaws.com"
          }
         }
        ],
        "Version": "2012-10-17"
       },
       "ManagedPolicyArns": [
        {
         "Fn::Join": [
          "",
          [
           "arn:",
           {
            "Ref": "AWS::Partition"
           },
           ":iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
          ]
         ]
        }
       ]
      },
      "Metadata": {
       "aws:cdk:path": "ExampleStack/Example1/Fn/ServiceRole/Resource"
      }
     },
     "Example1FnCBC0C9FD": {
      "Type": "AWS::Lambda::Function",
      "Properties": {
       "Code": {
        "S3Bucket": {
         "Fn::Sub": "cdk-hnb659fds-assets-${AWS::AccountId}-us-east-1"
        },
        "S3Key": "448f44cb8523ba7e460e321e56bc18df73372207e9166e1089b2ef72cbe168c8.zip"
       },
       "Role": {
        "Fn::GetAtt": [
         "Example1FnServiceRole6C38A161",
         "Arn"
        ]
       },
       "Handler": "app.handler",
       "Runtime": "nodejs16.x"
      },
      "DependsOn": [
       "Example1FnServiceRole6C38A161"
      ],
      "Metadata": {
       "aws:cdk:path": "ExampleStack/Example1/Fn/Resource",
       "aws:asset:path": "/workspaces/aws-sam-cli",
       "aws:asset:is-bundled": false,
       "aws:asset:property": "Code"
      }
     },
     "Example2FnServiceRole839DD1B5": {
      "Type": "AWS::IAM::Role",
      "Properties": {
       "AssumeRolePolicyDocument": {
        "Statement": [
         {
          "Action": "sts:AssumeRole",
          "Effect": "Allow",
          "Principal": {
           "Service": "lambda.amazonaws.com"
          }
         }
        ],
        "Version": "2012-10-17"
       },
       "ManagedPolicyArns": [
        {
         "Fn::Join": [
          "",
          [
           "arn:",
           {
            "Ref": "AWS::Partition"
           },
           ":iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
          ]
         ]
        }
       ]
      },
      "Metadata": {
       "aws:cdk:path": "ExampleStack/Example2/Fn/ServiceRole/Resource"
      }
     },
     "Example2FnFBA58E3D": {
      "Type": "AWS::Lambda::Function",
      "Properties": {
       "Code": {
        "S3Bucket": {
         "Fn::Sub": "cdk-hnb659fds-assets-${AWS::AccountId}-us-east-1"
        },
        "S3Key": "448f44cb8523ba7e460e321e56bc18df73372207e9166e1089b2ef72cbe168c8.zip"
       },
       "Role": {
        "Fn::GetAtt": [
         "Example2FnServiceRole839DD1B5",
         "Arn"
        ]
       },
       "Handler": "app.handler",
       "Runtime": "nodejs16.x"
      },
      "DependsOn": [
       "Example2FnServiceRole839DD1B5"
      ],
      "Metadata": {
       "aws:cdk:path": "ExampleStack/Example2/Fn/Resource",
       "aws:asset:path": "/workspaces/aws-sam-cli",
       "aws:asset:is-bundled": false,
       "aws:asset:property": "Code"
      }
     },
     "CDKMetadata": {
      "Type": "AWS::CDK::Metadata",
      "Properties": {
       "Analytics": "v2:deflate64:H4sIAAAAAAAA/zWM0QqDMAxFv8X3mjmFvc/BPkA/QGLtJLamYNrtofjvq45B4CTnkltDU0NV4EdKPdnS0QipD6itympIDtdxQkjPyDqQZ/V48X/fFeEKqfPOHPrgrqQZUMQEgfuBfEMbtTWhRcnxKXP/TDyfT0Z83HRO2E8GFrm8rzfIUxWLEJVb5ECrge7HL6MfX3uuAAAA"
      },
      "Metadata": {
       "aws:cdk:path": "ExampleStack/CDKMetadata/Default"
      }
     }
    },
    "Parameters": {
     "BootstrapVersion": {
      "Type": "AWS::SSM::Parameter::Value<String>",
      "Default": "/cdk-bootstrap/hnb659fds/version",
      "Description": "Version of the CDK Bootstrap resources in this environment, automatically retrieved from SSM Parameter Store. [cdk:skip]"
     }
    },
    "Rules": {
     "CheckBootstrapVersion": {
      "Assertions": [
       {
        "Assert": {
         "Fn::Not": [
          {
           "Fn::Contains": [
            [
             "1",
             "2",
             "3",
             "4",
             "5"
            ],
            {
             "Ref": "BootstrapVersion"
            }
           ]
          }
         ]
        },
        "AssertDescription": "CDK bootstrap stack version 6 required. Please run 'cdk bootstrap' with a recent version of the CDK CLI."
       }
      ]
     }
    }
   }
```

</div></details>

jfuss:
@WinterYukky I should be able to get you a review this week. Sorry for the delay.

samcli/lib/providers/provider.py
```diff
@@ -858,5 +893,6 @@ def _get_build_dir(resource: Union[Function, LayerVersion], build_root: str) ->
     """
     Return the build directory to place build artifact
     """
+    resource_id = resource.function_id if isinstance(resource, Function) else resource.layer_id
```
jfuss, line 896: I would prefer to find ways to leave the default to be `resource.name`. This is a much more maintainable solution. For example, if we add a new builder this will try to access `resource.layer_id` instead of "just working". If we need to change this, I am ok but let's change this to check for each resource and access the attribute we need. Otherwise default to `resource.name`.

samcli/lib/providers/sam_layer_provider.py
```diff
@@ -138,11 +149,21 @@ def _convert_lambda_layer_resource(
...
-            codeuri,
-            compatible_runtimes,
-            metadata,
+            arn=layer_logical_id,
+            codeuri=codeuri,
+            compatible_runtimes=compatible_runtimes,
+            metadata=metadata,
             compatible_architectures=compatible_architectures,
             stack_path=stack.stack_path,
+            layer_id=layer_id,
```
jfuss, line 167: I am not exactly understanding what layer_id is meant to represent. It looks like it is just the `layer_logical_id` but we have that within the `arn` property. So it's not clear why we need another property for the same thing.

samcli/lib/providers/sam_layer_provider.py
```diff
@@ -138,11 +149,21 @@ def _convert_lambda_layer_resource(
             LOG.debug("--base-dir is not presented, adjusting uri %s relative to %s", codeuri, stack.location)
             codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
 
+        # If the layer from CDK, name will be CDK resource id.
+        # Normally, LayerVersion will infer the name and layer id from the arn,
+        # but if the logical id is different from the name,
+        # specify the name and layer id in LayerVersion to use that name.
+        name = ResourceMetadataNormalizer.get_resource_name(resource_properties, layer_logical_id)
+        name = name if name != layer_logical_id else None
```
jfuss, line 157: I think I will need more time to digest this. But it's not super clear why this is here, how this changes part of the code base for getting the name or logical id for this resource.

jfuss (changes requested):
I left a couple comments but think I need to do a fresh pass when my brain is fresh. Just having a hard time following how `name` and `layer_id` fit into everything. Ideally, we could follow the pattern we have for functions but maybe there is a reason they need to be different that I don't quite get yet. 

I at least wanted to get some feedback out to you now.

WinterYukky:
> I left a couple comments but think I need to do a fresh pass when my brain is fresh. Just having a hard time following how `name` and `layer_id` fit into everything. Ideally, we could follow the pattern we have for functions but maybe there is a reason they need to be different that I don't quite get yet.
> 
> I at least wanted to get some feedback out to you now.

Thanks for reviewing @jfuss!
I recheck the my change and will reply.

samcli/lib/providers/provider.py
```diff
@@ -710,6 +734,16 @@ def get_full_path(stack_path: str, resource_id: str) -> str:
...
 
 
+def get_full_name(stack_path: str, resource_name: str) -> str:
+    """
+    Return the posix path-like name
+    while will used for name a resource from resources in a multi-stack situation
+    """
+    if not stack_path:
+        return resource_name
+    return posixpath.join(stack_path, resource_name)
```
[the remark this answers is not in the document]
WinterYukky: Yes, it is similar to the `get_full_path` implementation.

samcli/lib/providers/provider.py
WinterYukky: You are correct.
This time, we noticed that the value of resource.name coming from the CDK output template has been changed, and the name of the generated build directory has changed. Therefore, we have included this workaround for backward compatibility. If you are not affected by changing the name of the build directory, you can use the resource.name as it is. If there is no guarantee of this, then resource.name is the default.

samcli/lib/providers/sam_layer_provider.py
WinterYukky: Like `Function`, `Layer` also used potentially duplicate values (e.g. Fn) extracted from `aws:cdk:path` for `arn`. Thus, the data structure was as follows, which could cause the same problem as with `Function`.

| LogicalId | arn | layer_id(SamResourceId) | name |
|-|-|-|-|
| aaaLayer123 | Layer | Layer | Layer |
| bbbLayer456 | Layer | Layer | Layer |
| cccLayer789 | Layer | Layer | Layer |

To deal with the same problem as `Function`, it was necessary to change the data structure to the following one.

| LogicalId | arn | layer_id(SamResourceId) | name |
|-|-|-|-|
| aaaLayer123 | aaaLayer123 | aaaLayer123 | Layer |
| bbbLayer456 | bbbLayer456 | bbbLayer456 | Layer |
| cccLayer789 | cccLayer789 | cccLayer789 | Layer |

However, the `layer_id` and `name` resolution logic was structured in the following order of reference.

```mermaid
flowchart LR
    layer_id -- refer if not --> name -- refer if not --> arn
```

Therefore, changing `arn` to a `logical_id` would change `name`, and specifying `name` would next change `layer_id`, which references `name`, so there was a need to specify each completely.

I think this change is necessary to have backward compatibility because it supported potentially duplicate values (e.g. Fn) extracted from `aws:cdk:path` as search conditions for `Layer`.

samcli/lib/providers/sam_layer_provider.py
WinterYukky: The code was indeed a bit difficult to read the intent. This is a change that applies only to CDK resources, and requires that `layer_id` and `name` be specified when initializing the `LayerVersion` as described at https://github.com/aws/aws-sam-cli/pull/4064#discussion_r939480604 This is a change that applies only to CDK resources.
I have refactored it so that it is clear that it applies only to CDK resources.

jfuss: Maybe layer is different but it might be easier to understand and read if we don't do conditional things per IAC. This can make expanding and supporting Terraform or other IAC much more difficult in the future.

samcli/lib/providers/sam_layer_provider.py
jfuss: So if I am understanding what you are saying here is that they way we get the information from the `aws:cdk:path` and apply it into our object is the root of the issue here because we override properties in the generated template from CDK.

Is that a good summary?

jfuss (changes requested):
Still processing the changes but overall I think this looks good. 

Would you be able to add integration tests for this? And cover the cases like 1) invoking with logical id if two have similar cdk paths 2) invoke the using the "cdk id" (for backwards compatibility) 3) If there are any others you can think through.

Integration tests will really help me in understanding this is working as we intended.

samcli/lib/providers/sam_layer_provider.py
WinterYukky: Yes, the cdk id must be set to name and the logical id to layer_id, and this is the escape hatch that makes it happen.

jfuss (approved):
Seems good to me. Thanks for handling all the feedback.

We do require two approvals before merging. So I will find another person to review. Hopefully, we got most of the things handled through my review.

Thanks again for the contribution!

samcli/lib/providers/provider.py
```diff
@@ -88,14 +88,24 @@ class Function(NamedTuple):
...
 
+    @property
+    def full_name(self) -> str:
+        """
+        Return the path-like name of this Function. If it is in root stack, full_name = name.
+        Example:
+            "HelloWorldFunction"
+            "ChildStackA/GrandChildStackB/AFunctionInNestedStack"
+        """
+        return get_full_name(self.stack_path, self.name)
```
moelasmar, line 107: I think this function will mix between the CDK ID, and logical ID. It should return the same value as full_path in CFN use cases, but in CDK use case it will return the <stack CDK ID>/<Function Logical ID>. I am not sure how this mix will help the customer to identify a lambda function.

samcli/lib/providers/provider.py
```diff
@@ -360,14 +374,24 @@ def compatible_runtimes(self) -> Optional[List[str]]:
...
 
+    @property
+    def full_name(self) -> str:
+        """
+        Return the path-like name of this Layer. If it is in root stack, full_name = name.
+        Example:
+            "HelloWorldLayer"
+            "ChildStackA/GrandChildStackB/ALayerInNestedStack"
+        """
+        return get_full_name(self.stack_path, self.name)
```
moelasmar, line 393: same as `full_name` in Function

samcli/lib/providers/sam_function_provider.py
```diff
@@ -108,7 +109,7 @@ def get(self, name: str) -> Optional[Function]:
             found_fs = []
 
             for f in self.get_all():
-                if name in (f.function_id, f.name, f.functionname):
+                if name in (f.function_id, f.name, f.functionname, f.full_name):
```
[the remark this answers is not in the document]
moelasmar: why do not you use the `f.full_path` as well? The customer should be able to use `<stack cdk id>/<function cdk id>` as an identifier for a function.

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -242,29 +242,53 @@ def get_resource_id(resource_properties, logical_id):
             )
             return customer_defined_id
 
+        return logical_id
+
+    @staticmethod
+    def get_resource_name(resource_properties, logical_id):
```
moelasmar, line 248: if we add this change, the resource name is something related to CDK only. This way will break the Other IaC platforms support. 

The idea is we either define a lambda function using the logical Id or the custom defined Id (IaC Id) using `SamResourceId` metadata. Only in case of CDK, as the `cdk synth` command does not add the `SamSourceId` metadata, we use the `aws:cdk:path` to be the custom defined id (IaC Id).

moelasmar:
Thanks @WinterYukky for your effort. I have some concern about the solution you suggested to fix this problem. My concern is it could not be extend to other IaC frameworks. The `full_name` property you added to the lambda functions, and layers seems is only applicable for CDK use case, and this breaks our current implementation to support other IaC frameworks like Terraform.

As per [Jacob's comment](https://github.com/aws/aws-sam-cli/issues/4034#issuecomment-1187347966) in the opened issue. The solution we are looking for is to enable the customer to use logical id for search, and not to fix the CDK id extraction.So, I would suggest to define two ids for Functions, Layers, and Stacks (Logical Id, and Custom Id), and to have `full_logical_id` which will be `<parent stack full_logical_id>/<resource logical id>` and `full_custom_id` which will be `<parent stack full_custom_id>/<resource custom id>`. So, the customer can use either the a path of logical ids or a path of custom ids to define a resource, as I do not think that it is a valid use case for the customer to mix between the 2 id types to define a resource.

I also suggest to keep `ResourceMetadataNormalizer.get_resource_id` as it is, and remove `ResourceMetadataNormalizer.get_resource_name`. As, the custom logic to determine the resource custom id (CDK id) is only applicable for CDK use case, and this id is always equivalent to the custom id. The best solution here is to add a change to CDK so during synthesize step, it adds the `SamResourceId` metadata, and set its value to the resource CDK Id.

WinterYukky:
Thanks for reviewing @moelasmar!
Your propose that using new properties `custom_id` and `full_custom_id` looks like nice. I also understand your thoughts on other IaC frameworks. 

However, I don't agree to keep `ResourceMetadataNormalizer.get_resource_id` as it is. Because CDK ID (which SAM expects) can be duplicates. The solution can't fix this bug.

So I thought I should use the `IaCPluginInterface` to implement per-IaC behavior, but I don't know if that is appropriate. Can you help me?

samcli/lib/providers/provider.py
WinterYukky: You are correct.
I will fix to using only CDK ID.

moelasmar:
Sorry @WinterYukky for my slow reply. I want to leave the `ResourceMetadataNormalizer.get_resource_id` for now as in my opinion we should fix it from the CDK side, we need to push a PR in the CDK repo to add a new Metadata property to get the resource CDK id. Any fix we are going to do in SAM CLI is a workaround in my opinion, that is why I am thinking to leave it as is right now, till we can do the CDK change.

For this PR, we can change its purpose to enable the customers to use the Logical Ids to invoke a lambda function in case if there is any issue with using the CDK ID, so the customers can unblock themselves.

mndeveci:
Thanks for your contribution @WinterYukky ! I will be converting this PR into draft until you will have some time to address comments from second part of @moelasmar's suggestion.

WinterYukky:
@mndeveci 
I'm sorry. I'm busy now but working this issue slowly. I'm never abandonment.
When I'm ready let you know.

jfuss:
Closing as this appears stale. If you are still interesting in contribution this, feel free to reopen or create a new PR.

PR #4065, opened 2022-07-17
FEAT: Test Runner - Kick Off Testsuite

bentvelj:
Adds functionality to kick off a testsuite by running a Fargate task with a set of environment variables, using a [runTask](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html) query.

In order for customers to run their testsuites on Fargate, there needs to be functionality to make a runTask query to start a test-running task.

This change contains the code to make a test-running runTask query

N/A

### NOTE: 

The general image used in the Test Runner Task Definition is not yet public, so testing this is properly is not yet possible. Once remaining features are added, and the general image is public, thorough end to end testing can be done. Given that we are merging with a feature branch, I think waiting for further implementation and end to end tests is reasonable.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/test_runner/invoke_testsuite.py
+++ b/samcli/lib/test_runner/invoke_testsuite.py
@@ -1,103 +1,110 @@
+"""
+Kicks off a testsuite in a Fargate container
+"""
 from typing import List
 from samcli.commands.exceptions import ReservedEnvironmentVariableException
 
 
-def invoke_testsuite(
-    boto_ecs_client,
-    boto_s3_client,
-    bucket: str,
-    path_in_bucket: str,
-    ecs_cluster: str,
-    container_name: str,
-    task_definition_arn: str,
-    other_env_vars: dict,
-    test_command_options: str,
-    subnets: List[str],
-) -> None:
+class SuiteRunner:
+    def __init__(self, boto_ecs_client, boto_s3_client):
+        self.boto_ecs_client = boto_ecs_client
+        self.boto_s3_client = boto_s3_client
 
-    """
-    Kicks off a testsuite by making a runTask query.
+    def invoke_testsuite(
+        self,
+        bucket: str,
+        path_in_bucket: str,
+        ecs_cluster: str,
+        container_name: str,
+        task_definition_arn: str,
+        other_env_vars: dict,
+        test_command_options: str,
+        subnets: List[str],
+    ) -> None:
 
-    NOTE: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html
+        """
+        Kicks off a testsuite by making a runTask query.
 
-    Parameters
-    ----------
-    boto_client_provider : BotoProviderType
-        Provides a boto3 client in order to make a runTask query.
+        NOTE: https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html
 
-        Also used for waiting for results to appear in the s3 bucket.
+        Parameters
+        ----------
+        boto_client_provider : BotoProviderType
+            Provides a boto3 client in order to make a runTask query.
 
-    bucket : str
-        The name of the bucket used to store results.
+            Also used for waiting for results to appear in the s3 bucket.
 
-    path_in_bucket : str
-        The path within the bucket where results are stored.
+        bucket : str
+            The name of the bucket used to store results.
 
-    ecs_cluster : str
-        The name of the ECS Cluster to run the task on.
+        path_in_bucket : str
+            The path within the bucket where results are stored.
 
-    container_name : str
-        The name of the container in which the testsuite runs.
+        ecs_cluster : str
+            The name of the ECS Cluster to run the task on.
 
-        This is required to specify environment variables in a containerOverride.
+        container_name : str
+            The name of the container in which the testsuite runs.
 
-    task_definition_arn : str
-        The ARN of the task definition to run.
+            This is required to specify environment variables in a containerOverride.
 
-    other_env_vars : dict
-        Other environment variables to be sent to the container.
+        task_definition_arn : str
+            The ARN of the task definition to run.
 
-    test_command_options : str
-        Options to be passed to the test command.
+        other_env_vars : dict
+            Other environment variables to be sent to the container.
 
-        For example, '--maxfail=2
+        test_command_options : str
+            Options to be passed to the test command.
 
-    do_await : Optional[bool]
-        If enabled, wait for the task to finish and for results to appear in the bucket.
+            For example, '--maxfail=2
 
-    Raises
-    ------
-    botocore.ClientError
-        If run_task fails
+        do_await : Optional[bool]
+            If enabled, wait for the task to finish and for results to appear in the bucket.
 
-    botocore.WaiterError
-        If the results.tar.gz fails to appear in the bucket
-    """
+        Raises
+        ------
+        botocore.ClientError
+            If run_task fails
 
-    container_env_vars = {
-        "TEST_RUNNER_BUCKET": bucket,
-        "TEST_COMMAND_OPTIONS": test_command_options,
-        "TEST_RUN_ID": path_in_bucket,
-    }
+        botocore.WaiterError
+            If the results.tar.gz fails to appear in the bucket
+        """
 
-    reserved_vars = []
-    for key in other_env_vars.keys():
-        if key in container_env_vars.keys():
-            reserved_vars.append(key)
-        if not str.isidentifier(key):
-            raise ValueError(f"{key} is not a valid environment variable name.")
+        container_env_vars = {
+            "TEST_RUNNER_BUCKET": bucket,
+            "TEST_COMMAND_OPTIONS": test_command_options,
+            "TEST_RUN_ID": path_in_bucket,
+        }
 
-    if len(reserved_vars) > 0:
-        raise ReservedEnvironmentVariableException(
-            f"The following are reserved environment variable, ensure they are not present in your environment variables file: {reserved_vars}"
+        reserved_vars = []
+        for key in other_env_vars.keys():
+            if key in container_env_vars.keys():
+                reserved_vars.append(key)
+            if not str.isidentifier(key):
+                raise ValueError(f"{key} is not a valid environment variable name.")
+
+        if len(reserved_vars) > 0:
+            raise ReservedEnvironmentVariableException(
+                f"The following are reserved environment variable, ensure they are not present in your environment variables file: {reserved_vars}"
+            )
+
+        container_env_vars.update(other_env_vars)
+
+        self.boto_ecs_client.run_task(
+            cluster=ecs_cluster,
+            launchType="FARGATE",
+            networkConfiguration={"awsvpcConfiguration": {"subnets": subnets, "assignPublicIp": "ENABLED"}},
+            overrides={
+                "containerOverrides": [
+                    {
+                        "name": container_name,
+                        "environment": [{"name": key, "value": val} for key, val in container_env_vars.items()],
+                    }
+                ]
+            },
+            taskDefinition=task_definition_arn,
         )
 
-    container_env_vars.update(other_env_vars)
-
-    boto_ecs_client.run_task(
-        cluster=ecs_cluster,
-        launchType="FARGATE",
-        networkConfiguration={"awsvpcConfiguration": {"subnets": subnets, "assignPublicIp": "ENABLED"}},
-        overrides={
-            "containerOverrides": [
-                {
-                    "name": container_name,
-                    "environment": [{"name": key, "value": val} for key, val in container_env_vars.items()],
-                }
-            ]
-        },
-        taskDefinition=task_definition_arn,
-    )
-
-    results_upload_waiter = boto_s3_client.get_waiter("object_exists")
-    results_upload_waiter.wait(Bucket=bucket, Key=path_in_bucket + "/results.tar.gz")
+        results_upload_waiter = self.boto_s3_client.get_waiter("object_exists")
+        results_upload_waiter.wait(Bucket=bucket, Key=path_in_bucket + "/results.tar.gz")
--- a/samcli/lib/test_runner/invoke_testsuite.py
+++ b/samcli/lib/test_runner/invoke_testsuite.py
@@ -1,4 +1,4 @@
-from typing import *
+from typing import List
 import click
 from samcli.lib.utils.boto_utils import BotoProviderType
 from samcli.commands.exceptions import ReservedEnvironmentVariableException
--- a/samcli/lib/test_runner/invoke_testsuite.py
+++ b/samcli/lib/test_runner/invoke_testsuite.py
@@ -1,33 +1,7 @@
-from typing import List
 import click
-from samcli.lib.utils.boto_utils import BotoProviderType
+from typing import List
 from samcli.commands.exceptions import ReservedEnvironmentVariableException
-
-
-def get_subnets(boto_client_provider: BotoProviderType) -> List[str]:
-    """
-    Queries describe-subnets to get subnets associated with the customer account.
-
-    NOTE: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSubnets.html
-
-    Parameters
-    ----------
-    boto_client_provider : BotoProviderType
-        Provides a boto3 client in order to query the describe-subnets API
-
-    Returns
-    -------
-    List[str]
-        A list of subnet ids associated with a customer account.
-
-    Raises
-    ------
-    botocore.ClientError
-        If the describe_subnets call fails
-    """
-    ec2_client = boto_client_provider("ec2")
-    subnets = ec2_client.describe_subnets().get("Subnets")
-    return [subnet["SubnetId"] for subnet in subnets]
+from samcli.lib.utils.boto_utils import BotoProviderType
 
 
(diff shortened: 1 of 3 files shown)


samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,140 @@
...
+    botocore.ClientError
+        If the describe_subnets call fails
+    """
+    ec2_client = boto_client_provider("ec2")
+    subnets = ec2_client.describe_subnets().get("Subnets")
+    return [subnet["SubnetId"] for subnet in subnets]
+
+
+def invoke_testsuite(
+    boto_client_provider: BotoProviderType,
```
hoffa, line 34: How does the caller specify the region?

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+    click.secho("Successfully kicked off testsuite!\n", fg="green")
+
+    # If await is specified, wait for the task to finish and the results.tar.gz to appear in the bucket
+    if do_await:
+        click.secho("Awaiting testsuite completion...\n", fg="yellow")
+
+        task_arn = response.get("tasks")[0].get("taskArn")
+        task_waiter = ecs_client.get_waiter("tasks_running")
```
hoffa, line 133: These are only used in L140, any reason they're separate?

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    # If the customer specifies their own subnet, use it, otherwise query describe-subnets.
+
+    subnets = other_env_vars.get("subnets") or get_subnets(boto_client_provider)
+
+    container_env_vars = [
+        {"name": "BUCKET", "value": bucket},
+        {"name": "TEST_RUN_ID", "value": path_in_bucket},
+        {"name": "OPTIONS", "value": test_command_options},
```
hoffa, line 98: `BUCKET` and `OPTIONS` are maybe too generic as env vars, not unreasonable someone to already be using something similar. Even prefixing everything with `TEST_` (or whatever) could be a better way to reduce risk.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,140 @@
...
+    ]
+
+    for key, value in other_env_vars.items():
+        if key in ("BUCKET", "TEST_RUN_ID", "OPTIONS"):
```
hoffa, line 102: To avoid the risk of typos (hence risk of forgetting something), we can define `container_env_vars` as a regular dict, and before passing to the ECS API, convert to whatever weird notation it requires:

```python
[{"name": k, "value": v} for k, v in container_env_vars.items()]
```

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: Caller wouldn't specify the region directly, but would have the region and profile through context, and then construct a `boto_client_provider` to pass. Why would we need region passed directly?

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: sounds good, I think even `TEST_RUNNER_BUCKET`

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: Nope, I'll move the line down.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,138 @@
+from typing import *
```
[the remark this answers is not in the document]
bentvelj: Will do

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,138 @@
...
+    -------
+    List[str]
+        A list of subnet ids associated with a customer account.
+
+    Raises
+    ------
+    botocore.ClientError
+        If the describe_subnets call fails
+    """
+    ec2_client = boto_client_provider("ec2")
```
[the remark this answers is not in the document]
bentvelj: There is a `get_boto_client_provider_from_session_with_config`, I've used the `get_boto_client_provider_with_config` as per [Mehmet's recommendation](https://github.com/aws/aws-sam-cli/pull/4024#discussion_r914697665). Is that what you mean by "respect sessions"?

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,138 @@
...
+
+    click.secho("Successfully kicked off testsuite!\n", fg="green")
+
+    # If await is specified, wait for the task to finish and the results.tar.gz to appear in the bucket
+    if do_await:
+        click.secho("Awaiting testsuite completion...\n", fg="yellow")
+
+        task_waiter = ecs_client.get_waiter("tasks_running")
+        task_arn = response.get("tasks")[0].get("taskArn")
+        task_waiter.wait(cluster=ecs_cluster, tasks=[task_arn])
```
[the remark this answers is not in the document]
bentvelj: I do not see anything in the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Waiter.TasksRunning) about events. It seems that it just waits for a successful state.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Raises
+    ------
+    botocore.ClientError
+        If get_subnets, run_task, or waiters fail
+    """
+
+    # If the customer specifies their own subnet, use it, otherwise query describe-subnets.
+
+    subnets = other_env_vars.get("subnets") or get_subnets(boto_client_provider)
```
hoffa, line 93: Is `subnets` an environment variable? If yes, it should be consistent with the rest.

hoffa (approved):

samcli/lib/test_runner/invoke_testsuite.py
acristin: Why is the customer specifying subnets via an environment variable?

bentvelj: Yeah it is confusing. Originally the customer would specify a bunch of options (e.g. bucket & ecs cluster overrides) in this file, but those are now options for the `run` command and the input file is solely for environment varibales. Will change subnet overrides to an option, same as the others, not sure how I missed this.

jfuss (changes requested):

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
+import click
```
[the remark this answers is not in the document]
bentvelj: Ah fair enough, we can move that message to the caller.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    "name": container_name,
+                    "environment": [{"name": key, "value": val} for key, val in container_env_vars.items()],
+                }
+            ]
+        },
+        taskDefinition=task_definition_arn,
+    )
+
+    click.secho("Successfully kicked off testsuite!\n", fg="green")
```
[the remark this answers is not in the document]
bentvelj: Will do, noted for command PRs.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # If await is specified, wait for the task to finish and the results.tar.gz to appear in the bucket
+    if do_await:
```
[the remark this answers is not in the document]
bentvelj: I think we're gonna scrap the await option altogether, maybe include a fast-succeed option as a stretch feature. But I think above 80% of our customers will want to wait for tests to run and see results by default

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
...
+
+    for key in other_env_vars.keys():
+        if key in container_env_vars.keys():
+            raise ReservedEnvironmentVariableException(
+                f"{key} is a reserved environment variable, ensure that it is not present in your environment variables file."
+            )
+
+    container_env_vars.update(other_env_vars)
+
+    ecs_client = boto_client_provider("ecs")
```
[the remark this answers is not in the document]
bentvelj: Sure, I'll change the param to an ecs client

tests/unit/lib/test_runner/test_invoke_testsuite.py
```diff
@@ -0,0 +1,57 @@
...
+            "ecs_cluster": "test-cluster-name",
+            "container_name": "test-container-name",
+            "task_definition_arn": "test-task-def-arn",
+            "other_env_vars": {"TEST_RUNNER_BUCKET": "oh-no"},
+            "test_command_options": "--max-fail=1",
+            "subnets": ["subnet-xxxxxxxxxxxxxxxxx"],
+            "do_await": False,
+        }
+
+        with self.assertRaises(ReservedEnvironmentVariableException):
```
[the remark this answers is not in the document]
bentvelj: Agreed, will do

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
+import click
+from typing import List
+from samcli.commands.exceptions import ReservedEnvironmentVariableException
+from samcli.lib.utils.boto_utils import BotoProviderType
+
+
+def invoke_testsuite(
```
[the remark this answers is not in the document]
bentvelj: Is there a reason it should be in a class, if it's just a single function?

tests/unit/lib/test_runner/test_invoke_testsuite.py
```diff
@@ -0,0 +1,57 @@
+from botocore.exceptions import ClientError
+from unittest import TestCase
+from samcli.lib.test_runner.invoke_testsuite import invoke_testsuite
+from samcli.commands.exceptions import ReservedEnvironmentVariableException
+from samcli.lib.utils.boto_utils import get_boto_client_provider_with_config
+
+
+class Test_InvokeTestsuite(TestCase):
+    def test_specify_reserved_variable(self):
```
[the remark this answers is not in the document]
bentvelj: Adding a `isidentifier` check in the customer specifies invalid variable names

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        reserved_vars = []
+        for key in other_env_vars.keys():
+            if key in container_env_vars.keys():
+                reserved_vars.append(key)
+            if not str.isidentifier(key):
+                raise ValueError(f"{key} is not a valid environment variable name.")
+
+        if len(reserved_vars) > 0:
```
acristin, line 87: Nit: I think this is counter to PEP 8 guidance; should look more like `if reserved_vars`. This is something that might be best addressed as a followup pass to catch all code that is inconsistent with PEP 8, and then we can make a call on which ones to change.

tests/unit/lib/test_runner/test_invoke_testsuite.py
```diff
@@ -0,0 +1,75 @@
...
+            "ecs_cluster": "test-cluster-name",
+            "container_name": "test-container-name",
+            "task_definition_arn": "test-task-def-arn",
+            "other_env_vars": {},
+            "test_command_options": "--max-fail=1",
+            "subnets": ["subnet-xxxxxxxxxxxxxxxxx"],
+        }
+
+    def test_invalid_var_names(self):
+
```
acristin, line 23: Nit: Remove extra blank lines

acristin (approved):

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    container_env_vars = {
+        "TEST_RUNNER_BUCKET": bucket,
+        "TEST_COMMAND_OPTIONS": test_command_options,
+        "TEST_RUN_ID": path_in_bucket,
+    }
+
+    for key in other_env_vars.keys():
+        if key in container_env_vars.keys():
```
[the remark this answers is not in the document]
jfuss: > I'm adding an `isidentifier` check to each anyway

Not sure I understand what this is. Could you summarize this for me? Mainly for my understanding on the trade offs you are making.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,116 @@
+"""
+Kicks off a testsuite in a Fargate container
```
jfuss, line 2: Are there going to be other types of `SuiteRunner`s? You may want to consider making an interface and extend it to be Fargate. This gives you the ability to make different `SuiteRunner`s and interchange them with a factory or something similar.

An example of this is our "provider" for getting different types of resources parsed out of templates: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/providers/provider.py#L472.

I wouldn't require this now but something you may want to consider having in place, which could make this much more expandable in the future.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+from typing import List
+import logging
+from samcli.commands.exceptions import ReservedEnvironmentVariableException
+from samcli.lib.utils.colors import Colored
+
+LOG = logging.getLogger(__name__)
+
+
+class SuiteRunner:
```
jfuss, line 12: If this `SuiteRunner` is specific to Fargate, I would indicate that in the class name here.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,116 @@
...
+                    {
+                        "name": container_name,
+                        "environment": [{"name": key, "value": val} for key, val in container_env_vars.items()],
+                    }
+                ]
+            },
+            taskDefinition=task_definition_arn,
+        )
+
+        LOG.info(Colored().yellow("=> Successfully kicked off testsuite, waiting for completion...\n"))
```
jfuss, line 113: Nit: You can move `Colored()` into the `__init__` of the class over doing it here. This way the class (as a whole) has access to it and you don't have to create it multiple times. You can also make it optional parameter to allow easier mocking and allow a caller to change it (if required). 

I wouldn't block the PR on this but trying to give you some additional things you may want to consider.

samcli/lib/test_runner/invoke_testsuite.py
```diff
@@ -0,0 +1,116 @@
...
+            "TEST_COMMAND_OPTIONS": test_command_options,
+            "TEST_RUN_ID": path_in_bucket,
+        }
+
+        reserved_vars = []
+        for key in other_env_vars.keys():
+            if key in container_env_vars.keys():
+                reserved_vars.append(key)
+            if not str.isidentifier(key):
+                raise ValueError(f"{key} is not a valid environment variable name.")
```
jfuss, line 89: I don't recommend raising `ValueError` or any of the default python Error or Exceptions. Main reason is they become hard to catch and know the reason, as `ValueError` could be something thrown by the interpreter or your code. Which forces you to inspect the message in order to understand deeper. Instead, create an Error specific to this case. You can extend ValueError to give the caller the ability to catch all.

tests/unit/lib/test_runner/test_invoke_testsuite.py
```diff
@@ -0,0 +1,66 @@
...
+            "test_command_options": "--max-fail=1",
+            "subnets": ["subnet-xxxxxxxxxxxxxxxxx"],
+        }
+
+    def test_invalid_var_names(self):
+        self.params["other_env_vars"] = {r"0variable_name": "value", r"\othervariable_name": "othervalue"}
+        with self.assertRaises(ValueError):
+            self.runner.invoke_testsuite(**self.params)
+
+    def test_specify_variable_bucket(self):
```
jfuss, line 27: Nit: you can parameterize this test with the next two. Up to you, but calling it out incase you were unaware of test parameterization.

jfuss (changes requested):

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: Great idea, will do

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: Ah, good point, I just did it how I saw it done elsewhere, but it makes sense with lots of LOG messages

samcli/lib/test_runner/invoke_testsuite.py
bentvelj: Will do, I'll make note of this for other aspects of code.

tests/unit/lib/test_runner/test_invoke_testsuite.py
bentvelj: I wasn't, but I am now! This is much better, very easy to expand.

PR #4069, opened 2022-07-20
feat: Adding Image PackageType lambda function mapping

moelasmar:
This PR to add the Image PackageType Lambda Function related properties to the generated metadata file. We mapped the 
`image_uri`, and `image_config` properties from terraform application to the generated metadata.


tests/unit/hook_packages/terraform/test_prepare_hook.py
```diff
@@ -73,6 +90,29 @@ def setUp(self) -> None:
             "Code": "file.zip",
         }
 
+        self.tf_iamge_package_type_function_properties: dict = {
```
torresxb1, line 93: small typo on `iamge`

torresxb1 (approved):
approving with minor comments

hawflau (approved):

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -298,14 +298,49 @@ def _build_lambda_function_code_property(tf_properties: dict) -> Any:
...
+
+    Returns
+    -------
+    dict
+        The built ImageConfig property of a CloudFormation AWS Lambda Function resource
+    """
+    image_config = tf_properties.get("image_config")
+    if not image_config:
+        return None
+    image_config = image_config[0]
```
sriram-mv, line 327: do we need to assert that image_config type here, before attempting to get the first element.

sriram-mv: Primarily because we are relying on terraform not changing their contract on how they operate with their files, we should sanity check types of whatever fields we are accessing.

moelasmar: as per the terraform plan output, it should be a list. Terraform translate Optional attributes into List, and either it is empty, or contain one item.

moelasmar: as per our discussion, I will add this check to guard our code in case if terraform changed there contract, and to fail elegantly

sriram-mv (approved):

PR #4071, opened 2022-07-20
fix: Fixed describe stack events reporting wrong events

lucashuy:
This change fixes `describe_stack_events()` reporting old events due to using epoch time in seconds instead of milliseconds.

The execution time parameter has been removed from `describe_stack_events()`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


sriram-mv:
What kind of tests can be written to ensure wrong events are not reported?

lucashuy:
> What kind of tests can be written to ensure wrong events are not reported?

One test can write is an integration test. The test would first deploy a new stack and result in the CREATE_COMPLETE status. Then we can redeploy the stack and the stack should end in UPDATE_COMPLETE with no CREATE_COMPLETE events. This can be done by just changing the runtime in the template (eg. Python 3.7 -> 3.8).

Will update with a test.

samcli/lib/deploy/deployer.py
```diff
@@ -446,7 +446,7 @@ def _check_stack_not_in_progress(status: str) -> bool:
         return "IN_PROGRESS" not in status
 
     def wait_for_execute(
-        self, stack_name: str, stack_operation: str, disable_rollback: bool, execution_time: float = time.time()
+        self, stack_name: str, stack_operation: str, disable_rollback: bool, execution_time: float = time.time() * 1000
```
sriram-mv, line 449: Having anything mutable in the arguments makes me uneasy. In this case though it looks like it should be okay, since its not a list or a dictionary.

tests/integration/deploy/test_deploy_command.py
```diff
@@ -1285,3 +1285,56 @@ def test_deploy_logs_warning_with_cdk_project(self, template_file):
...
+            no_execute_changeset=False,
+            tags="integ=true clarity=yes foo_bar=baz",
+            confirm_changeset=False,
+        )
+
+        deploy_process_execute = run_command(deploy_command_list)
+        self.assertEqual(deploy_process_execute.process.returncode, 0)
+
+        # Deploy template that modifies existing resource, this should only UPDATE
+        template_path = self.test_data_path.joinpath("aws-serverless-function-cdk.yaml")
```
sriram-mv, line 1317: why a cdk template? I'm assuming this could be any different template that's not the same template as before, lets make it so its explicit. something like "aws-serverless-function-changed.yaml"

Otherwise the PR looks good to me.

sriram-mv (approved):
Approving with comments.

tests/integration/deploy/test_deploy_command.py
lucashuy: Yup, it just has to be a different template file to trigger an update. The CDK template was used here since it happens to use the same resource name.

samcli/lib/deploy/deployer.py
qingchm: For execution time parameter where else is this used? If this is the desired value can we just completely deprecate the execution time parameter if we never want values to be passed in?

lucashuy: Will take a look at all instances of where this gets called and remove if the references just pass the current time in. If I recall correctly, there was only one instance where it takes in a value other than current time, but even then it is just the time that an execute command finished.

I'll look into it and see if this argument is required.

lucashuy: After looking into it, the only time where execution time was provided was when it was called after `execute_changeset()`, however it was just passing the same thing (`time.time()`). This parameter is now removed.

mildaniel (approved):

PR #4072, opened 2022-07-20
Send events via thread when at capacity

Leo10Gama:
N/A

Previously, when at capacity, Events would be sent inline. This could create slowdowns during execution if the Telemetry happens to be slow on a given day.

By sending these Events in a separate thread, we ensure that they are sent without disrupting the flow of execution of other commands and functions.

There are no notable side effects to this change, besides the use of an additional thread during execution.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


hawflau (approved):
LGTM!

mildaniel (approved):

Issue #4059, opened 2022-07-15
SAM CDK -> sam build -> DockerImageCode.from_image_asset: Unable to set parent folder as the DockerContext 


dunika:
### Description:

I am having trouble with the following setup.

I have the following folder structure:
```
/root
  .Pipfile
  /shared_dependencies
  /project
    /runtime
      DockerFile
```

I need to set `/root` as the `DockerContext` and  `/root/project/runtime/DockerFile` as the path to the `Dockerfile`

If I create a `DockerImageCode` instance and use it to create a `DockerImageFunction` like so:
```
code = DockerImageCode.from_image_asset(
  directory="/filesystem_path/root",
  file="/project/runtime/DockerFile",
)

DockerImageFunction(
  code=code,
)
```

`cdk synth >| template.yaml` will output a template and an `AWS:Serverless:Function` resource with the following metadata:

```
Metadata:
  aws:asset:path: /filesystem_path/root
  aws:asset:dockerfile-path: /project/runtime/DockerFile
```

My `DockerFile` contains the following:

```
COPY Pipfile .
```

To facilitate local development, I need to run `sam build`. When I do so, I get the following error:

```
COPY failed: file not found in build context or excluded by .dockerignore:
 stat Pipfile: file does not exist
```

Upon inspecting `.aws-sam/build.toml`, it would seem that the `DockerContext` is being set to `/filesystem_path/root/project/runtime`, the folder in which the `DockerFile` resides.

I would have assumed that the `DockerContext` would correspond to the `directory` parameter that gets passed into `DockerImageCode.from_image_asset`? (Which corresponds to `aws:asset:path` in the template)

As far as I can tell, there is no way of setting the `DockerContext` to any folder that is not the folder in which the `DockerFile` resides.

[This pull request ](https://github.com/aws/aws-sam-cli/pull/3492)added the mapping logic for `cdk image asset metadata -> sam build params`. 

Looking at the [changes in this file](https://github.com/aws/aws-sam-cli/pull/3492/files#diff-58fdce738a6b28772eb8a9a742194ffe8023e756bd9d0d4594e72cec55d12a25R107), it would seem that the `DockerContext` will always be equal to the folder in which the DockerFile resides.

Is there a way of setting the `DockerContext` that I am not aware of? Or is it currently impossible?

I came across [this issue](https://github.com/aws/aws-cdk/issues/3342) which is now closed but does not offer a satisfactory resolution to this problem.

### Additional environment details:

[...]

PR #2425, opened 2020-12-01
feat(images): SAM CLI support for Lambda Images

sriram-mv:
- [ ] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #4074, opened 2022-07-21
Use the same DockerContext and Dockerfile paths as the CDK is using

i18n-tribe:

#4059

When defining an `aws_lambda.DockerImageFunction` construct using the CDK, the docker image is configured with code like this:

```python
code = aws_lambda.DockerImageCode.from_image_asset(
    directory="/Users/john/PycharmProjects/project-root",
    file="./application/runtime/Dockerfile",
)

docker_lambda = aws_lambda.DockerImageFunction(code=code)
```

When the CDK project is built, the CDK uses `directory` as the `DockerContext` and `file` as the `Dockerfile`.

See [the documentation](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/DockerImageCode.html#aws_cdk.aws_lambda.DockerImageCode.from_image_asset):

> directory (str) – the directory from which the asset must be created.
> file (Optional[str]) – Path to the Dockerfile (relative to the directory). Default: ‘Dockerfile’

This means `/Users/john/PycharmProjects/project-root` is the `DockerContext`

and the `Dockerfile` is a path to a docker file from within the `DockerContext` - i.e. `./application/runtime/Dockerfile`

This capability of the CDK is useful because...

Say you have a folder structure like:

```
project-root
└───application
│   └───runtime
│       │   Dockerfile
│       │   app.py
│       │   requirements.txt
└───another_application
│   └───runtime
│       │   Dockerfile
│       │   app.py
│       │   requirements.txt
└───shared_code
    │   shared.py

```

There are 2 apps in the repo - in folders `application` and `another_application`

Each app contains a lambda.

The 2 lambdas share some common code, which is in a folder called `shared_code`

In each lambda's `Dockerfile`, we need to copy the `shared_code` module into the container.

Therefore, when we are building each lambda's container, we need to set the `DockerContext` to be the root folder of the repo `/project-root`, because this means `/project-root/shared_code` will be accessible when we build the container.

The `DockerContext` cannot be the folder that contains `Dockerfile`. I.e. it cannot be `project-root/application/runtime` - because the `/project-root/shared_code` folder would not be readable while the container is being built.  

This example works fine in the CDK.

However - when we wish to test locally using the `sam-cli` - it does not work.

The reason it does not work with the `sam-cli` is - the `sam-cli` code is not using the same `DockerContext`  and  `Dockerfile` paths as the CDK. The `sam-cli` code is currently ensuring that `DockerContext` is _always_ the folder that contains `Dockerfile`, irrespective of how `DockerContext` has been configured in the `aws_lambda.DockerImageFunction` CDK construct.

So in this example - `DockerContext` is `/Users/john/PycharmProjects/project-root/application/runtime`

Side-by-side comparison:

|                   | CDK                                                | SAM                                                            |
|-------------------|----------------------------------------------------|----------------------------------------------------------------|
| `DockerContext`   | `/Users/john/PycharmProjects/project-root`         | `/Users/john/PycharmProjects/project-root/application/runtime` |
| `Dockerfile`      | `./application/runtime/Dockerfile`                 | `Dockerfile`                                                   |

Side Note:

This example would be the same with a single app in the repo.

This example is the same in any case where there are 2 lambdas that share some code, but don't share a Dockerfile.

The SAM CLI will now use the same `DockerContext` and `Dockerfile` paths as the CDK is using.

No effect at all on production code. 

It could in theory effect local testing in some cases.

However – we cannot think of a scenario where local testing is currently working, but will stop working because of this change. 

If there is any code that is currently working that will stop working because of this change – the currently working code must be relying on in some kind of hack to work around the issue this PR is fixing. 

We wrote our own hack - it looks like this:

```python
   code = create_docker_image_code_from_asset(
        dockerfile_path=dockerfile_path,
        directory=docker_context,
        cmd=cmd,
    )

    docker_lambda = aws_lambda.DockerImageFunction(code=code)

    # Required for local development
    if docker_lambda.node.default_child:
        docker_lambda.node.default_child.add_metadata("DockerContext", docker_context)
        docker_lambda.node.default_child.add_metadata("Dockerfile", dockerfile_path)
        docker_lambda.node.default_child.add_metadata("SamNormalized", True)
```

So we are manually setting `DockerContext` and `Dockerfile`  to be the same as the CDK. 

And by setting `SamNormalized=True` metadata, we are able to bypass the current `sam-cli` code that is setting `DockerContext` and `Dockerfile` metadata to be different to the CDK - which would overwrite what we had manually set.

`sam build` works fine with our hack. 

Our code (that includes the hack) will still work after the change in this PR is merged. So if a team has done a similar hack to us, their ability to test locally will not break when this PR is merged.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -186,11 +186,9 @@ def _extract_image_asset_metadata(metadata):
         """
         asset_path = Path(metadata.get(ASSET_PATH_METADATA_KEY, ""))
         dockerfile_path = Path(metadata.get(ASSET_DOCKERFILE_PATH_KEY), "")
-        dockerfile, path_from_asset = dockerfile_path.stem, dockerfile_path.parent
-        dockerfile_context = str(Path(asset_path.joinpath(path_from_asset)))
         return {
-            SAM_METADATA_DOCKERFILE_KEY: dockerfile,
-            SAM_METADATA_DOCKER_CONTEXT_KEY: dockerfile_context,
+            SAM_METADATA_DOCKERFILE_KEY: str(dockerfile_path),
```
qingchm, line 190: A bit concerned about this change since this will break the current code behaviour, might need to investigate on the current usage volume to decide if we can change this part.

i18n-tribe: > might need to investigate on the current usage volume to decide if we can change this part.

Sounds like a good move 👍 

> this change... will break the current code behaviour

As noted, I think it's good to investigate this.

From our own investigation...

I would not describe this as `breaking` any current behavior.

It does `change` the current behavior. 

But if we say the change `breaks` something... that sounds like: something will stop working.

So - would this `change` cause any currently working code to `break`? 

We do not think it will. 

To illustrate...

Say your CDK code looks like this:

```python
code = aws_lambda.DockerImageCode.from_image_asset(
    directory="/Users/john/PycharmProjects/project-root/application/runtime/",
    file="Dockerfile",
)

docker_lambda = aws_lambda.DockerImageFunction(code=code)
```

This works with the current code.

And this will still work after this PR is merged.

Currently - the SAM CLI only works with CDK code where `file` is just the `Dockerfile` and `directory` is the full path to the `Dockerfile`.

In contrast, CDK code like this...

```python
code = aws_lambda.DockerImageCode.from_image_asset(
    directory="/Users/john/PycharmProjects/project-root",
    file="./application/runtime/Dockerfile",
)

docker_lambda = aws_lambda.DockerImageFunction(code=code)
```

...does not currently work with the SAM CLI. But it will work when this PR is merged.

An important question would be – 

Why was the code written like this to begin with?

In the [PR where this code was introduced](https://github.com/aws/aws-sam-cli/pull/3492) – this does not seem to have been discussed in the comments. 

What this code behaviour specified in your internal tools e.g. Jira? If so, what were the reasons given?

The goal of this feature is to be compatible with the CDK.

But currently, the code is not supporting an important capability of the CDK.

Both `cdk deploy` and `sam build` work fine with metadata like:

```python
DockerContext ="/Users/john/PycharmProjects/project-root",
Dockerfile ="./application/runtime/Dockerfile",
```

So we ware not sure why code has been written to ensure sam always works with metadata like:

```python
DockerContext ="/Users/john/PycharmProjects/project-root/application/runtime",
Dockerfile ="Dockerfile",
```

jfuss: > So it is not clear to us why the code in the PR was been written to ensure sam cli always works with metadata like

This is how SAM CLI handles it. So it was written to match SAM CLI expectations.

jfuss:
@i18n-tribe I am still trying to grasp what is being changed and how that impacts the CLI. At the very least, we need integration tests added to ensure this behaves as expected. I would have to check to see if we have integ tests that cover the current integrations, but if we don't have those, we need them added. This ensures we are not breaking current customers and allows us to better regression test going further.

mndeveci:
Thanks for your contribution @i18n-tribe. Until you will find sometime and get back to this PR, I am marking it as draft.

samcli/lib/samlib/resource_metadata_normalizer.py
i18n-tribe: @jfuss apologies for the slow reply, I was on vacation. 

So I had a look at the SAM CLI code to see how `DockerContext` and `Dockerfile` metadata are used by the SAM CLI.

I found 4 other usages. 

Just one usage is relevant to this PR: 

In [this code block](https://github.com/i18n-tribe/aws-sam-cli/blob/bdb8b961bdcb237105a2776b09cde9944a1f7e5d/samcli/lib/build/app_builder.py#L384-L423) in `samcli/lib/build/app_builder.py` 

We read the metadata properties:

```python
dockerfile = cast(str, metadata.get("Dockerfile"))
docker_context = cast(str, metadata.get("DockerContext"))
```

Then we make `docker_context` into an absolute path:

```python
docker_context_dir = pathlib.Path(self._base_dir, docker_context).resolve()
```

Then we call `docker build` using the python docker client, passing the properties as arguments:

```python
build_args = {
    "path": str(docker_context_dir),
    "dockerfile": dockerfile,
}
build_logs = self._docker_client.api.build(**build_args)
```

So in summary - 

- `Dockerfile` is passed without modification to `_docker_client.api.build` as the `dockerfile` argument
- `DockerContext` is converted to an absolute path, and passed to  `_docker_client.api.build` as the `path` argument

Here is the documentation for ['_docker_client.api.build'](https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.build)

And those arguments are:

```
path (str) – Path to the directory containing the Dockerfile
dockerfile (str) – path within the build context to the Dockerfile
```

In my opinion, the documentation for `path` is a little bit ambiguous.

Does this mean that `path` is:

1. the immediate parent directory containing the Dockerfile? 
2. a directory that ultimately contains the dockerfile, which might be nested in child folder(s)?

I would completely understand if something thinks it means `1.` - this is what I thought at first.

But if we look at the documentation for `dockerfile` - it says `path within the build context to the Dockerfile`

Because `dockerfile` can be a `path within the build context` - it can be a path (potentially via subfolders) to the `Dockerfile`.

So in fact, `2.` is the correct interpretation.

This interpretation is reinforced by the `docker cli` documentation for [docker build](https://docs.docker.com/engine/reference/commandline/build/)

Usage is:

```
docker build [OPTIONS] PATH 
```

Where `PATH` will be `DockerContext` in the SAM CLI.

And the option `--file` will be `Dockerfile` in the SAM CLI.

In the `docker cli` documentation, they refer to something called the `build context`.

The `build context` is very similar to `PATH` / `DockerContext`.

The `build context` is a copy of the folder specified in `PATH` / `DockerContext`, excluding files specified in `.dockerignore`.

So `--file` / `Dockerfile` is a path to a docker file within the `build context` (usually a relative path is used).

There is no requirement for the docker file to be in the root of the `build context`.

I.E. `--file` / `Dockerfile` can be a path via subfolders to the docker file.

Here's [the documentation for `--file`](https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f).

The documentation has examples where `--file` is a path via subfolders to the docker file like:

```
docker build --file dockerfiles/Dockerfile.debug .
docker build --file dockerfiles/Dockerfile.prod .
```

In the example above, `PATH` is `.` - the current directory.


And there's a quote in [this section of the docs](https://docs.docker.com/engine/reference/commandline/build/#text-files).

```
By default the docker build command will look for a Dockerfile at the root of the build context. 
The -f, --file, option lets you specify the path to an alternative file to use instead. 
This is useful in cases where the same set of files are used for multiple builds. 
```

The use case `where the same set of files are used for multiple builds` is exactly what this PR is trying to solve.

This is the use case described in the [PR description](https://github.com/aws/aws-sam-cli/pull/4074#issue-1313551509).

This use-case is supported by the `AWS CDK`.

This use-case is also - in fact - already supported by the `SAM CLI`.

The [`AWS CDK` docs](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/DockerImageCode.html#aws_cdk.aws_lambda.DockerImageCode.from_image_asset) support this use-case explicitly:

```
directory (str) – the directory from which the asset must be created.
file (Optional[str]) – Path to the Dockerfile (relative to the directory). Default: ‘Dockerfile’
```

The [`SAM CLI` docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-build.html#build-container-image) do not support this use-case explicitly:

```python
Dockerfile - The name of the Dockerfile associated with the Lambda function.
DockerContext - The location of the Dockerfile.
```

The `SAM CLI` developer guide sound like interpretation `1.` of the documentation for ['_docker_client.api.build'](https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.build).

But in fact, the `SAM CLI` code works perfectly well if `Dockerfile` is a path via subfolders to the docker file.

And I've personally used this 'undocumented feature' in multiple `SAM` projects I have worked on.

In other words -

This PR is removing code that added a limitation to the `SAM CLI`, when the limitation did not previously exist in either `SAM` or `CDK`, but `SAM CLI` docs said this limitation exists.

So I've created a [PR to update the `SAM CLI` developer guide](https://github.com/awsdocs/aws-sam-developer-guide/pull/159), so people will be aware of this capability of the `SAM CLI`.

I've also created a PR to [update the documentation for the Docker Python Client](https://github.com/docker/docker-py/pull/3043) to be more clear.

---

A further note -

Above there is a link do documentation for the `docker cli`. 

The python docker SDK doesn't use the `docker cli`, it uses the `docker api`.

So I checked the docker API documentation for [build](https://docs.docker.com/engine/api/v1.40/#operation/ImageBuild)

It takes an argument:

```
dockerfile - Path within the build context to the Dockerfile.
```

It also takes a tar archive which contains the `build context` - documented as:

```
Build an image from a tar archive with a Dockerfile in it.

The Dockerfile specifies how the image is built from the tar archive.
It is typically in the archive's root, but can be at a different path or have a different name by specifying the dockerfile parameter.
```

I.e. `dockerfile` 'is typically in the archive's root' - but it can be a path to a dockerfile in a subfolder.

And I checked the source code of the docker python sdk, to see how it's interacting with the docker API.

Here's the [code where it creates the `build context tar archive`](https://github.com/docker/docker-py/blob/b7daa52feb8b897fb10fbe82c6e49a273746352a/docker/api/build.py#L150-L163).

This is a copy of the folder specified in `path` / `DockerContext`, excluding files specified in `.dockerignore`.

And [on this line](https://github.com/docker/docker-py/blob/b7daa52feb8b897fb10fbe82c6e49a273746352a/docker/api/build.py#L174) it passes `Dockerfile` to the docker API as `dockerfile`.
 
---

For reference, here are the 3 other usages of `DockerContext` metadata, which are not relevant to this PR:

- In samcli/commands/build/build_context.py - [If it's a docker lambda, confirm these metadata properties are defined](https://github.com/i18n-tribe/aws-sam-cli/blob/c2b947d6087cd88f2174e6a3b3079865dc06394e/samcli/commands/build/build_context.py#L562-L573).
- In samcli/lib/utils/resource_trigger.py - [trigger a code refresh if this metadata is changed the the developer](https://github.com/i18n-tribe/aws-sam-cli/blob/1cda797850907cf28c7b4f80180cba3627360fd4/samcli/lib/utils/resource_trigger.py#L260-L264)
- In samcli/lib/providers/sam_function_provider.py - [converting `DockerContext` to be an absolute path if necessary](https://github.com/i18n-tribe/aws-sam-cli/blob/4b4e6e44ea7dd7e4d9f73001214dc34a5aa59701/samcli/lib/providers/sam_function_provider.py#L413-L422)

i18n-tribe:
> @i18n-tribe I am still trying to grasp what is being changed and how that impacts the CLI. 

@jfuss perhaps it could be productive if we arrange a quick video call to discuss?

I can completely understand you not yet finding this clear. 

It's quite a subtle issue, and it takes quite a lot of words to explain in writing. 

But I have a feeling if we discuss it briefly in realtime, we will quickly be able to align on a common understanding.

If you have Calendly or a similar app I could use to book a meeting at a convenient time for you, I'd be very happy to do so. 

We could potentially also include @sriram-mv who originally wrote the [code to support docker lambdas using SAM](https://github.com/aws/aws-sam-cli/pull/2425).

In the meantime, I have written a [systematic analysis](https://github.com/aws/aws-sam-cli/pull/4074#discussion_r967206796) of how the `SAM CLI` uses `Dockerfile` and `DockerContext`.

From this analysis, and from my personal experience working on multiple SAM projects, I believe the `SAM CLI` code does in fact support the use-case where `Dockerfile` is inside a nested subfolder within `DockerContext`.

Thus - the code that is removed in this PR seems to be enforcing a supposed limitation of the `SAM CLI`, which sounded like a limitation in the SAM developer guide, but was not a real limitation in the code.

i18n-tribe:
> Thanks for your contribution @i18n-tribe. Until you will find sometime and get back to this PR, I am marking it as draft.

@mndeveci  sorry for slow reply to @jfuss 's comments, I was on vacation.

I'd be more than happy to help move this forward when you have some time - by discussing on a brief call, and if you like by helping to write the integration tests. 

I think a call would be really helpful as a next step, because we do not yet have a shared understanding of what this PR is doing.

Once we have a shared understanding of the PR, it will be easier to prioritise this PR accurately - at an appropriate level of priority.

samcli/lib/samlib/resource_metadata_normalizer.py
moelasmar: Thanks @i18n-tribe for your contribution and your very detailed explanation. I think you have a point regarding this change, I think we miss understood the `DockerContext` and `DockerFile` properties and we assumed that the `DockerFile` should be a file in the `DockerContext` directory path. I did some investigations to make sure that this change will not break any customer. I think it will work as expected, and fix this bug.

my only request is to add some integration testing to cover this case

moelasmar (changes requested):
Looks good to me. Could you please add some integration testing to cover this case.

samcli/lib/samlib/resource_metadata_normalizer.py
FredTopTal: You're most welcome, I'll write some integration tests 👍

i18n-tribe:
> Looks good to me. Could you please add some integration testing to cover this case.

Thanks @moelasmar , I have added a [new commit](https://github.com/aws/aws-sam-cli/pull/4074/commits/ec8d936d251f114977cc4a103acc18a857c30313) with integration testing for the case where `Dockerfile` is a path to a docker file via subfolders, which is useful in the case where some images use some shared code.

All tests are passing in CI. I'm not sure if this new test class is running in CI, but I have manually run it locally with the following command, and it's passing 👍 

```
SAM_CLI_DEV=1 APPVEYOR_CI_OVERRIDE=1 BY_CANARY=1 pytest tests/integration/buildcmd/test_build_cmd.py::TestBuildCommand_PythonFunctions_ImagesWithSharedCode
```

moelasmar:
Thanks @i18n-tribe for your update. Can I ask you to update the integration test case to be a CDK application and add it [here](https://github.com/aws/aws-sam-cli/tree/develop/tests/iac_integration/cdk) or at least use a synthesized template, as the sample test case you added not really testing the  metadata normalizer change you introduce.

i18n-tribe:
@moelasmar there is one caveat - at the time of writing, I have not yet been able to fully run these new integration tests. It looks like it's going to take several hours, so I may have to try again first thing tomorrow morning.

I thought I would push the commit now anyhow, in case it'll be run by CI. 

CI checks are all passing.

But if the CDK integration tests are not run by CI, then I will confirm when I have chance to run the tests all locally.

i18n-tribe:
OK - all the integration tests are passing on my machine 👍 

image

Ran with command `SAM_CLI_DEV=1 pytest tests/iac_integration/cdk/test_sam_cdk_integration.py`

moelasmar:
Thanks @i18n-tribe for your contribution

moelasmar (approved):

hawflau (approved):
LGTM and thanks for your contribution!

hawflau:
Hey @i18n-tribe, can you please checking the failing tests? - https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/45273404

i18n-tribe:
> Hey @i18n-tribe, can you please checking the failing tests? - https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/45273404

Fixed in [this commit](https://github.com/aws/aws-sam-cli/pull/4074/commits/b36eaa061dcf411d1b43515340cc8743d4c88f51) @hawflau 👍

PR #4076, opened 2022-07-22
Add `sam sync` Events

Leo10Gama:
N/A

New Events related to the `sam sync` command will provide deeper insights into how customers use these features and commands.

The two new Events created are `SyncUsed` (which tracks when a `sam sync` has started and ended, which will provide us with a total duration of the sync thanks to `Event.time_stamp`) and `SyncFlow` (which tells us which sync flow was used, similar to the `BuildRuntime` Event).

The only notable side effects will be slightly more Events in the `EventTracker` when `sam sync` is used.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


qingchm:
Should we leave this as a draft for now since it doesn't look to have tests added yet?

Leo10Gama:
> Should we leave this as a draft for now since it doesn't look to have tests added yet?

I don't think tests are necessary for a change like this, since it's just populating the codebase with trackers. My way of thinking is that, if the `Event` and `EventTracker` methods work as intended (verified by their unit tests), then this change shouldn't require specific tests. Further, we don't have any sorts of other unit or integration tests for the other `Event` types under the same line of reason, so I wouldn't see adding one here as necessary.

samcli/commands/sync/command.py
```diff
@@ -403,10 +410,39 @@ def execute_code_sync(
         else set(get_all_resource_ids(stacks))
     )
 
+    sync_function_event_mapper = {
```
hawflau, line 413: Is there some way that we can avoid using maps? Reason is we will need to maintain these maps in the future whenever there's a new sync flow

samcli/lib/telemetry/event.py
```diff
@@ -39,6 +41,23 @@ class EventType:
             "CreateChangeSetSuccess",
         ],
         EventName.BUILD_FUNCTION_RUNTIME: INIT_RUNTIMES,
+        EventName.SYNC_USED: [
+            "Start",
+            "End",
+        ],
+        EventName.SYNC_FLOW: [
```
hawflau, line 48: Why don't we just use the class name of the classes defined under "samcli/lib/sync/flows/..."? 
We can probably update samcli/lib/sync/flows/__init__.py to first collect all SyncFlow classes and sub-classes, and then just do something like `from samcli.lib.sync.flows import *` to import all these classes/sub-classes. If we do this change, we might also want to update other places where these SyncFlow classes/sub-classes are imported.

Reason for asking for this change is mainly to reduce our manual effort needed to maintain this list in the future

samcli/commands/sync/command.py
```diff
@@ -403,10 +410,39 @@ def execute_code_sync(
...
+        "RestApi": "RestApi",
+        "StepFunctions": "StepFunctions",
+    }
+
     for resource_id in sync_flow_resource_ids:
         sync_flow = factory.create_sync_flow(resource_id)
         if sync_flow:
             executor.add_sync_flow(sync_flow)
+            # Track the type of sync flow
+            name = sync_flow.log_name
```
hawflau, line 431: If we track the class name of sync flows as event value, would that make this block code simpler?

hawflau (changes requested):

samcli/commands/sync/command.py
Leo10Gama: I'm not sure whether there's a way to avoid it completely, but there may be a way to refactor the maps to be simpler. Initially I used multiple `if .. elif` statements, but pylint didn't seem too fond of it lol

samcli/lib/telemetry/event.py
Leo10Gama: I considered looking into this, but there wasn't a consistent list I could find to collect all the SyncFlows, especially since some files included additional resources (i.e. abstract classes, module functions). If we don't anticipate many additional SyncFlows to be added, I don't suspect it should be much more of a hassle to add them into the tracker.

samcli/commands/sync/command.py
Leo10Gama: I tried that initially, but I can definitely take a deeper look into doing that! I think I might have a new idea for how that can be simplified

samcli/commands/sync/command.py
```diff
@@ -144,6 +144,7 @@
 @capabilities_option(default=DEFAULT_CAPABILITIES)  # pylint: disable=E1120
 @experimental
 @pass_context
+@track_long_event("SyncUsed", "Start", "SyncUsed", "End")
```
hawflau, line 147: are we able to measure the duration of each sync flow if we use `track_long_event` here? My understanding is this would track the entire `sam sync` command run.

samcli/lib/telemetry/event.py
hawflau: Another thought - is it possible to do the event registration and event tracking in the `__init__` method of a SyncFlow subclass?

For example, within the `__init__` method of `AliasVersionSyncFlow`:
1. append  `AliasVersionSyncFlow.__name__` (or `type(self).__name__`) to the list `EventName.SYNC_FLOW` if it is not in it already
2. `EventTracker.track_event("SyncFlow", type(self).__name__)`

We might just create a method to do both in the `SyncFlow` class, then every subclass of `SyncFlow` would have it called automatically.

Just trying to think of ways that we don't need to do anything manually.

hawflau (changes requested):

samcli/commands/sync/command.py
Leo10Gama: This would create two Events, `("SyncUsed", "Start")` and `("SyncUsed", "End")`. Since both Events are given timestamps at the time of their creation, we would be able to interpolate the duration of the command by taking the difference of the two times.

samcli/lib/telemetry/event.py
Leo10Gama: It's possible, but my worry is that creating a method to add new values dynamically could potentially lead to misuse (i.e. one wrong line accidentally storing a possible value as some arbitrary string, or Events not being stored when they should be). The hard-coded route does create more of a hassle for the implementation, but unless the `SyncFlow` classes themselves (or other modules that implement them) are refactored in such a way that we have a concrete list of all SyncFlows, I think it's still the safest solution.

samcli/commands/sync/command.py
hawflau: We also get the same measurement from `track_command` if I didn't misunderstand anything.
If we do it this way, can we get individual duration for each sync flow?

samcli/lib/telemetry/event.py
hawflau: okay it makes sense to me about your concern in the event registration part

samcli/commands/sync/command.py
Leo10Gama: We could, though with this current implementation, we wouldn't be able to tell which pairs of events would match with which since there's the possibility that they share a thread with the main function (unless I'm mistaken). I could add new Events `SyncFlowStart` and `SyncFlowEnd` which would have the same values as `SyncFlow` to measure duration and allow us to see the types of flows and their duration, which may allow us to collect that information.

hawflau (approved):

samcli/lib/telemetry/event.py
```diff
@@ -21,11 +21,27 @@ class EventName(Enum):
...
     BUILD_FUNCTION_RUNTIME = "BuildFunctionRuntime"
+    SYNC_USED = "SyncUsed"
+    SYNC_FLOW_START = "SyncFlowStart"
+    SYNC_FLOW_END = "SyncFlowEnd"
 
 
 class EventType:
     """Class for Events and the types of values they may have."""
 
+    _SYNC_FLOWS = [
```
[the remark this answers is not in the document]
Leo10Gama: Yup it's definitely something I'll keep in mind in the near future 👍

samcli/lib/telemetry/event.py
```diff
@@ -21,11 +21,27 @@ class EventName(Enum):
...
+        "AliasVersionSyncFlow",
```
mildaniel, line 33: If we're including `FunctionSyncFlow` here as the parent of Zip and Image functions, should we also include `GenericApiSyncFlow`? I'm not certain on how these definitions are used downstream in the metrics.

mildaniel (commented):
LGTM! Just one question.

samcli/lib/telemetry/event.py
Leo10Gama: My logic was that, since `GenericApiSyncFlow` is described as an interface for `HttpApiSyncFlow` and `RestApiSyncFlow`, it wouldn’t exist unless something were to go wrong. Though, that was also the reason I included `FunctionSyncFlow`. Adding in the extra value is definitely possible, I’ll get on that as soon as I’ve got a minute!

PR #4079, opened 2022-07-25
FEAT: Test Runner - Refactor of Template Generation

bentvelj:
Improvements upon this [PR](https://github.com/aws/aws-sam-cli/pull/4024).
+ Refactor Template Generation into a class, move the query to resource tagging api to the caller.
+ Prevent IAM statements from being generated for IAM resources.
+ More tests

Encapsulating the template generation process into a class makes the process much simpler, and extendable.
We do not want to generate IAM statements for IAM resources, as IAM permissions are particularly dangerous, and are not applicable for integration testing.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,214 @@
...
+            `None` if the `resource_arn` corresponds to an IAM or STS resource.
+
+        Raises
+        ------
+        jinja2.exceptions.TemplateError:
+            If the IAM Statement template render fails.
+        """
+
+        partition_regex = r"(aws|aws-cn|aws-us-gov)"
+        region_regex = r"(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d"
```
hoffa, line 79: Any reason not to use `[\w-]+` for both? Having to maintain a list of current regions seems painful.

bentvelj: Yeah, I see. I wanted to keep the regex strict for clarity sake, but we don't want to take up the burden of maintaining the regions.

hoffa (approved):
:shipit:

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,215 @@
...
+            LOG.debug("ARN `%s` is for an IAM or STS resource, will not generate permissions.", resource_arn)
+            return None
+
+        LAMBDA_FUNCTION_REGEX = rf"^arn:{partition_regex}:lambda:{region_regex}:{account_regex}:function:[\w-]+(:\d+)?$"
+        APIGW_API_REGEX = rf"^arn:{partition_regex}:apigateway:{region_regex}::\/apis\/\w+$"
+        APIGW_RESTAPI_REGEX = rf"^arn:{partition_regex}:apigateway:{region_regex}::\/restapis\/\w+$"
+        SQS_QUEUE_REGEX = rf"^arn:{partition_regex}:sqs:{region_regex}:{account_regex}:[\w-]+$"
+        S3_BUCKET_REGEX = rf"^arn:{partition_regex}:s3:::[\w-]+$"
+        DYNAMODB_TABLE_REGEX = rf"^arn:{partition_regex}:dynamodb:{region_regex}:{account_regex}:table\/[\w-]+$"
+        STEPFUNCTION_REGEX = rf"^arn:{partition_regex}:states:{region_regex}:{account_regex}:stateMachine:[\w-]+$"
```
jfuss, line 96: This is slightly concerning. Why do we need to maintain regex's here? This is also brittle, as (if) new partitions are build, this requires us to have to maintain that map.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,215 @@
...
+                f"Failed to open jinja template, File Not Found."
+            ) from jinja_template_not_found_ex
+
+        if self.resource_arn_list:
+
+            statements_list = []
+            for arn in self.resource_arn_list:
+                try:
+                    statement = self._create_iam_statement_string(arn)
+                except jinja2.exceptions.TemplateError as template_error:
```
jfuss, line 189: I know this is a refactor but I assume we added `jinja` to our dependencies?

Any reason we need to use jinja here instead of just string manipulation? We have tended to do string manipulation in the past and output a template that way (reduces the need for additional dependencies).

jfuss (changes requested):

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: [Jinja is a dependency of Flask](https://flask.palletsprojects.com/en/2.1.x/installation/#dependencies), so it was already in the SAMCLI, I didn't add it.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Yes, I agree it is definitely not ideal.

I can send the design document for this component offline, but the idea is to make writing IAM statements a bit easier by having a group of commonly used resources for which the template will include a set of commonly used actions. For example, `Lambda` would have `InvokeFunction`.

Which resources to generate IAM statements for is determined via tags (to support CFN & Terraform), so by querying tagging API, we get a list of resource ARNs. Determining which ARN belongs to which resource requires some logic when some arns are subsets of others, e.g. [APIGW ARNs](https://docs.aws.amazon.com/apigateway/latest/developerguide/arn-format-reference.html). It was decided that regex was most pragmatic to link resource ARNs to their type.

As for the partitions, their expressions are general, and should not require changing in the future. I left them in the `partition_regex` variable for clarity.

There's also no real "obligation" to maintain this list of supported resources, it's really just to include the (estimated) most commonly used resources for integration testing.

samcli/lib/test_runner/test_runner_template_generator.py
jfuss: I know that but you need to be adding it to our dependencies. For example, since we directly use Click, Click is a dependency directly. This allows us to pin to specific versions we need from the CLI. Let's say Flask removes Jinja, now you just impacted SAM CLI. Direct dependencies (one you use/import) MUST be defined in our requirements file.

Still curious why `jinja` is needed as a dependency (directly).

samcli/lib/test_runner/test_runner_template_generator.py
jfuss: I will ping you off this. I don't fully understand what is going on here. 

> There's also no real "obligation" to maintain this list of supported resources,

I disagree. Once you release something, you (or the team) signs up to maintain it. So if arns change (for whatever reason), we have to change it. So it is something we have to maintain.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Right, I see.

I don't think Jinja is needed, I just thought it would be more convenient than direct string manipulation, but it makes sense that it is not worth adding another dependency

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: That was poorly worded, I meant that we don't necessarily need to _add_ resources, the tool will not break if we don't. But yes, we definitely are signing up to maintain any changes to the arns, and even to add new resources to the pool. I didn't intend to say it did not require maintenance, it does.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,215 @@
...
+        for arn_regex, action_list in default_permissions_map.items():
+            if re.search(arn_regex, resource_arn) is not None:
+                LOG.debug("Matched ARN `%s` to IAM actions %s", resource_arn, action_list)
+
+                # APIGW API IAM Statements:
+                # https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-control-access-using-iam-policies-to-invoke-api.html
+                if arn_regex in (APIGW_API_REGEX, APIGW_RESTAPI_REGEX):
+                    apiId = self._extract_api_id_from_arn(resource_arn)
+                    execute_api_arn = (
+                        f"!Sub arn:${{AWS::Partition}}:execute-api:${{AWS::Region}}:${{AWS::AccountId}}:{apiId}/*/GET/*"
```
acristin, line 132: I can't recall; why do we assume GET instead of defaulting to `arn:${{AWS::Partition}}:execute-api:${{AWS::Region}}:${{AWS::AccountId}}:{apiId}/*`?

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,215 @@
...
+            action_list=[], arn=resource_arn
+        )
 
-    image_uri : str
-        The URI of the Image to be used by the Test Runner Fargate task definition.
+    def _get_default_bucket_name(self) -> str:
+        """
+        Returns a guaranteed unique default S3 Bucket name
+        """
+        return "test-runner-bucket" + str(uuid.uuid4())
```
acristin, line 152: Minor nit; consider a `-` separator between 'bucket' and the uuid.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Update: Obviously regex is not ideal. An ideal solution would be a proper ARN parser available in boto3 (Go and Java SDKs have ARN libraries, but not boto), or a tag API that returns resource types as well as arns for resources matching a given tag.

Neither of those will be available any time soon, so I'm going to abstract away as much of the regex to something that can easily be changed later. Once more important aspects of the project are finished and merged, I can go back and replace the mechanism responsible for matching ARNs to IAM statements with a simple ARN parser.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Changed to `/*`, this way any action is authorized from any stage for any resource for a **single api**.

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Will do, should be there.

samcli/lib/test_runner/test_runner_template_generator.py
```diff
@@ -1,276 +1,283 @@
...
-                )
-
-    # If the supplied ARN is for a resource without any default actions supported
-    LOG.info("Found no match for ARN `%s` to any IAM actions.", resource_arn)
-    return jinja2.Template(source=new_statement_template, keep_trailing_newline=True).render(
-        action_list=[], arn=resource_arn
-    )
-
-
-def _query_tagging_api(tag_filters: dict, boto_client_provider: BotoProviderType) -> Union[dict, None]:
```
acristin, line 123: Should I expect to see these again in a future PR?

acristin (approved):

samcli/lib/test_runner/test_runner_template_generator.py
bentvelj: Yep, but it's not relevant to this one

jfuss (approved):

PR #4094, opened 2022-08-02
FEAT: Test Runner - Init CLI + Environment Variable File Generation

bentvelj:
Code for `sam test-runner init` command entrypoint and environment variable file generation

Note that there does not yet exist a public ECR repository to store Test Runner images, so the image URI is specified as an option for now. In the future, a public image will be selected corresponding to the specified runtime.

A PR will be raised to include the proper image URIs as soon as the public ECR repository is created.

In order to make the process of referencing resources in integration tests, the `sam test-runner init` command will generate a YAML file containing mappings between a resource name (e.g. Lambda function name) and it's corresponding ARN. This file can then be passed to `sam test-runner run` to expose the mappings as environment variables in the Fargate container.

The environment variable names are the resource names, prepended with an underscore, set to all capitals, and with all hyphens replaced with underscores.

e.g. `_MY_LAMBDA_FUNCTION : arn:aws:lambda:us-east-1:123456789012:function:my-lambda-function`

The customer can then reference the `_MY_LAMBDA_FUNCTION` environment variable in their test code.

The customer can also change the names of the environment variables, or add new ones, which is convenient for getting information to the Test Runner.

Includes code to generate environment variable file and code for the `sam test-runner init` CLI entry.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/test_runner/generate_env_vars.py
```diff
@@ -0,0 +1,94 @@
...
+
+            NOTE: https://docs.aws.amazon.com/quicksight/latest/APIReference/qs-arn-format.html
+        """
+
+        # Extract resource-id from ARN
+        # Cases:
+        # 1) arn:partition:service:region:account-id:resource-id
+        # 2) arn:partition:service:region:account-id:resource-type/resource-id
+        # 3) arn:partition:service:region:account-id:resource-type:resource-id
+        components = resource_arn.split(":")
```
hoffa, line 38: Do we have any existing methods to parse ARNs? If not, this is a prime candidate for a separate function.

samcli/lib/test_runner/generate_env_vars.py
```diff
@@ -0,0 +1,94 @@
+import yaml
+from typing import List
+from samcli.commands.exceptions import FailedArnParseException
+
+
+class FargateRunnerArnMapGenerator:
```
hoffa, line 6: Is a class needed?

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,199 @@
...
+from typing import Optional, Any
+from samcli.lib.test_runner.utils import get_image_uri
+from samcli.lib.test_runner.test_runner_template_generator import FargateRunnerCFNTemplateGenerator
+from samcli.commands.exceptions import NoResourcesMatchGivenTagException
+from samcli.cli.main import pass_context
+from samcli.lib.utils.boto_utils import get_boto_client_provider_with_config
+from samcli.lib.test_runner.utils import write_file
+from samcli.lib.test_runner.utils import query_tagging_api
+from samcli.lib.test_runner.generate_env_vars import FargateRunnerArnMapGenerator
+from samcli.lib.utils.colors import Colored
```
acristin, line 17: Should these be sorted?

Can we combine the imports from samcli.lib.test_runner.utils into one line? E.g., `from samcli.lib.test_runner.utils import get_image_uri, query_tagging_api, write_file`?

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,199 @@
...
+        LOG.info(
+            COLOR.yellow(
+                "These environment variables hold the ARNs of your testable resources, and passing this file to `sam test_runner run` will make them available in the container running your tests.\n\n"
+                "Feel free to change them, or add new variables, but keep in mind that the test runner reserves the keys TEST_RUNNER_BUCKET, TEST_COMMAND_OPTIONS, and TEST_RUN_ID, and will throw an Exception if they are specified.\n\n"
+                "If you add new resources or change existing ones, you can use `sam test-runner get-arns` to generate an updated environment variable specification file.\n"
+            )
+        )
+    else:
+        # If no tags were provided we cannot pass resource ARNs to the template generator to generate IAM statements
+        # We also cannot generate a resource ARN map
```
acristin, line 181: Should we still generate an empty mapping file so customers can add their own env variables?

samcli/lib/test_runner/generate_env_vars.py
```diff
@@ -0,0 +1,94 @@
...
+    def _get_env_var_name(self, resource_arn: str) -> str:
```
acristin, line 7: Do we want any checks to ensure we don't collide with reserved env variables? I'm thinking of aws related ones that the fargate task needs to have to support auth, our own variables for S3 bucket, requirements, etc. 

It may be unlikely, but I am worried customers might break in a confusing way if they have an unfortunately named resource.

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,199 @@
...
+        template_generator = FargateRunnerCFNTemplateGenerator(resource_arn_list)
+        arn_map_generator = FargateRunnerArnMapGenerator()
+
+        resource_arn_map_yaml_string = arn_map_generator.generate_env_vars_yaml_string(resource_arn_list)
+
+        write_file(env_file, resource_arn_map_yaml_string)
+
+        LOG.info(
+            COLOR.green(
+                f"✓ Sucessfully generated an environment variable specification file `test_runner_arn_map.yaml`.\n"
```
hoffa, line 169: What is it used for?

samcli/lib/test_runner/generate_env_vars.py
bentvelj: Unfortunately we do not, so yes I'll abstract out the parsing code.

samcli/commands/test_runner/init/cli.py
bentvelj: I put the explanation in a separate LOG statement below, just to use a different color

https://github.com/aws/aws-sam-cli/blob/0f34f6218a4a6fba36c82e7a885e32a55c01f92d/samcli/commands/test_runner/init/cli.py#L172-L178

samcli/commands/test_runner/init/cli.py
bentvelj: Maybe not an empty file, but definitely a message reminder saying "if you'd like to pass environment variables, create a YAML file etc."

I feel automatically creating empty YAML files might be more annoying for those who don't want to include environment variables than creating one will be for those who do

samcli/lib/test_runner/generate_env_vars.py
bentvelj: This is done in `run` process: https://github.com/aws/aws-sam-cli/blob/d81eec8ddbb9c1ede36cb14947f51c0c76c13314/samcli/lib/test_runner/invoke_testsuite.py#L79-L90

It is also impossible to have any collisions by default, since all generated environment variable names begin with an underscore, and none of the reserved names do. The `init` output warns of the reserved names so that the customer is aware of which names they cannot add or change existing ones to.

samcli/commands/test_runner/init/cli.py
bentvelj: Sure, I'll organize the imports

samcli/lib/test_runner/generate_env_vars.py
bentvelj: Not necessarily, I was just following OOP practice for consistency and to make it easier to extend in the future.

samcli/commands/test_runner/init/cli.py
hoffa: I still don't understand what it's used for. What uses the ARNs?

bentvelj: Oh sorry, the ARNs can be used to access resources within customer tests. E.g. If a customer wanted to invoke a lambda in their integration test, they can do so by specifying the ARN. Having the arns exist as environment variables within the Fargate container allows the customer to write tests around these variables instead of copy pasting ARNs.

For example, test code can look something like this:

```
ENGLISH_LAMBDA_ARN = os.getenv('ENGLISH_LAMBDA')
def test_english_lambda():
    lambdaClient = boto3.client('lambda')
    response = lambdaClient.invoke(FunctionName=ENGLISH_LAMBDA_ARN)
    ...
```

acristin (approved):

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,204 @@
...
+    if tag_value and tag_key:
+        tag_filters = [{"Key": tag_key, "Values": [tag_value]}]
+    else:
+        tag_filters = None
+
+    # TODO: When ready, public ECR repositories will be created holding images capable of running tests for each available runtime.
+    #       Until ready for public images, image URIs are supplied by option
+    # image_uri = get_image_uri(runtime)
+
+    if tag_filters is not None:
```
jfuss, line 149: Generally, I avoid using `if: else:` in this way. Over time things becomes hard to maintain because you have to think through all the `if`'s to understand what is being caught in the `else`. There are times for it though, so not trying to suggest never to do this. 

In this case, here is what I would do:

Instead of checking for `is not None`, I would check the opposite first. So doing a `if tag_filers is None:` and then exist early from the command (looks like that case done nothing). Then you can assume, `tag_filters` is not None, allowing this block of code to become unnested and making it easier to read and maintain.

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,204 @@
...
+    # TODO: Merge these two options into one to follow AWS CLI (if we can get access to their parser)
+    if tag_key and not tag_value:
+        LOG.info(COLOR.red("Specified tag key without a tag value."))
+        return
+
+    if tag_value and not tag_key:
+        LOG.info(COLOR.red("Specified tag value without a tag key."))
+        return
+
+    if tag_value and tag_key:
```
jfuss, line 140: Similar to the below comment on line 149. I would do `tag_filters = None` and then assign it with this if statement.

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,204 @@
...
+DEFAULT_TEST_RUNNER_CFN_TEMPLATE_FILE_NAME = "test_runner_template.yaml"
+SUPPORTED_RUNTIMES = ["python3.8"]
+
+LOG = logging.getLogger(__name__)
+COLOR = Colored()
+
+
+@click.command("init", help=HELP_TEXT, short_help=SHORT_HELP)
+@click.option(
+    "--tag-key",
```
jfuss, line 43: Are you only allowing one tag to be provided? Is there never a case to provide multiple? 

How would a customer provide multiple keys if they want?

Seems like https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/resourcegroupstaggingapi.html#ResourceGroupsTaggingAPI.Client.get_resources supports multiple keys, so wouldn't this implementation limit our ability to do that since key and values are two different options provided to the command?

jfuss (changes requested):

samcli/commands/test_runner/init/cli.py
bentvelj: Right, this should have been noted in the PR. Having the tags specified as two separate options is a temporary solution, and a point of improvement. Ideally, we'd want to mimic the way AWS CLI enters tags, this "shorthand" form, e.g. `Key=string,Values=string,string`. However, [their parser](https://github.com/aws/aws-cli/blob/2f09fcc0e28784affb472d9aa0d3dd2c3ab513de/awscli/shorthand.py) is not something I can easily copy. 

I don't want to implement my own simple version of it for fear of edge cases that produce behavior inconsistent with AWS CLI, as that would be really confusing for customers.

I couldn't find one grepping around, but does SAM CLI have similar shorthand parser?

samcli/commands/test_runner/init/cli.py
bentvelj: For sure, shouldn't have missed that

samcli/commands/test_runner/init/cli.py
jfuss: I would check what we do for deploy. This can be something to follow and would make the cli consistent on how tags are done

samcli/commands/test_runner/init/cli.py
jfuss: Ya I missed that because it was hard to read and indention (maybe it's the diff). At the very list, create private methods for the different cases. That will help with the readability aspect.

bentvelj: Sounds good

samcli/commands/test_runner/init/cli.py
bentvelj: Ah, I found a `CfnTags` class which is awesome. It's not the same format as AWS CLI, but that's fine, as long as this tag option is consistent with other tag options in SAM CLI

bentvelj:
Refactored out the `utils.py` into the init CLI, since they were more init CLI utilities, so we don't need the generic utils file.

jfuss (changes requested):

samcli/commands/test_runner/init/cli.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+import logging
+from typing import Any, List, Optional, Union
+
+import click
+
+from samcli.cli.main import pass_context
+from samcli.cli.types import CfnTags
+from samcli.commands._utils.custom_options.option_nargs import OptionNargs
+from samcli.commands.exceptions import NoResourcesMatchGivenTagException
```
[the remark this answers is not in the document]
bentvelj: Will do!

jfuss (approved):

PR #4096, opened 2022-08-03
Add workflow events

Leo10Gama:
N/A

Events previously exist to capture the runtime of functions as they are built. However, up until now, we do not have a way to capture the information of workflow configs.

With a new `WorkflowUsed` Event, we will be able to capture and analyze data on which workflows customers are using most. This can reveal information such as the most common languages of workflows, most common dependency managers, and how many different workflow configs are used per build.

A new file, `workflows.py`, has been created to prevent a cyclic import, and allow for dynamic computing of Event values. Thus, for any new workflow to be added, it would need to be created there, and then be imported by `workflow_config.py`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/workflow_config.py
+++ b/samcli/lib/build/workflow_config.py
@@ -209,8 +209,6 @@ def get_workflow_config(
         config = cast(WorkFlowSelector, selector).get_config(code_dir, project_dir)
 
         EventTracker.track_event("WorkflowUsed", f"{config.language}-{config.dependency_manager}")
-        if config.dependency_manager == "npm-esbuild":
-            EventTracker.track_event("UsedFeature", "ESBuild")
 
         return config
     except ValueError as ex:
--- a/samcli/lib/build/workflow_config.py
+++ b/samcli/lib/build/workflow_config.py
@@ -4,97 +4,24 @@ Contains Builder Workflow Configs for different Runtimes
 
 import os
 import logging
-from collections import namedtuple
 from typing import Dict, List, Optional, Tuple, Union, cast
 
+from samcli.lib.build.workflows import (
+    CONFIG,
+    PYTHON_PIP_CONFIG,
+    NODEJS_NPM_CONFIG,
+    RUBY_BUNDLER_CONFIG,
+    JAVA_GRADLE_CONFIG,
+    JAVA_KOTLIN_GRADLE_CONFIG,
+    JAVA_MAVEN_CONFIG,
+    DOTNET_CLIPACKAGE_CONFIG,
+    GO_MOD_CONFIG,
+    PROVIDED_MAKE_CONFIG,
+    NODEJS_NPM_ESBUILD_CONFIG,
+)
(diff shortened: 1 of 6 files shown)


samcli/lib/build/workflow_config.py
```diff
@@ -278,6 +207,11 @@ def get_workflow_config(
 
         # Identify workflow configuration from the workflow selector.
         config = cast(WorkFlowSelector, selector).get_config(code_dir, project_dir)
+
+        EventTracker.track_event("WorkflowUsed", f"{config.language}-{config.dependency_manager}")
+        if config.dependency_manager == "npm-esbuild":
```
mildaniel, line 212: Please confirm with Wing if we need this. Otherwise let's get rid of this `UsedFeature` metric since we can get this same data from the `WorkflowUsed` metric.

samcli/lib/telemetry/event.py
```diff
@@ -19,11 +20,11 @@ class EventName(Enum):
     """Enum for the names of available events to track."""
 
     USED_FEATURE = "UsedFeature"
-    DEPLOY = "Deploy"
```
mildaniel, line 22: Why is `Deploy` and the corresponding logic removed here?

mildaniel (commented):
Looks good overall,left some minor comments. Can we also add some tests here to validate this functionality?

samcli/lib/build/workflow_config.py
Leo10Gama: 👍

samcli/lib/telemetry/event.py
Leo10Gama: I noticed that, while the `Deploy` Event was present in the design docs, it was never actually used or implemented. Also, I found it redundant to have an Event providing the status of the `sam deploy` command, when that information would already be present in the `commandRun` metrics with its exit codes and traceback.

Leo10Gama:
> Can we also add some tests here to validate this functionality?

I'm not sure whether tests would be necessary for a change like this. I intentionally designed the `EventTracker` to be lightweight, such that Events could be tracked quickly with only a line or two of code, and able to be removed just as quickly. My line of logic is, if in the future we didn't need this information and wanted to remove the tracker, we'd also have to dig through tests and remove the relevant ones to allow `make pr` to pass. I think it should be enough that the trackers are verified locally through a dummy project or two to ensure their logic is correct, though I could implement tests to verify this if we really wanted to be sure.

mildaniel:
I would generally disagree that the ability to make lightweight changes quickly is a reason against testing. If someone accidentally removes this logic in the future, they might think it's okay because all of the tests still pass. Not everyone has and will have the same context. For that reason, I think having tests as a guardrail against regression is already more valuable than the couple extra minutes someone will need to spend to remove a few tests later on.

hawflau (changes requested):

hawflau (approved):

PR #4099, opened 2022-08-06
add color to required options, change help text width and help text on long options.

sriram-mv:
No issue, but an UX improvement

For overall better UX, showcase which options are required and make options less verbose than they need to be.

BEFORE:
Screen Shot 2022-08-05 at 9 01 18 PM

AFTER:

Screen Shot 2022-08-05 at 9 00 56 PM

Less confusing to view the help output.

N/A

#### Mandatory Checklist

* No unit tests, need to vet the concept.

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -292,6 +292,9 @@ docs/_build/
 # pyenv
 .python-version
 
+# pyright
+pyrightconfig.json
+
 # celery beat schedule file
 celerybeat-schedule.*
 
--- a/samcli/commands/_utils/custom_options/option_help.py
+++ b/samcli/commands/_utils/custom_options/option_help.py
@@ -1,5 +1,5 @@
 """
-Custom click option for changing color of the help text
+Custom click option for customizing help text.
 """
 
 import click
@@ -15,3 +15,14 @@ class RequiredOptionColorTextOption(click.Option):
     def get_help_record(self, ctx):
         option, help_text = super(RequiredOptionColorTextOption, self).get_help_record(ctx=ctx)
         return (self.colored.yellow(option), self.colored.yellow(help_text)) if self.required else (option, help_text)
+
+
+class ReplaceHelpSummaryOption(click.Option):
+    def __init__(self, *args, **kwargs):
+        self.replace_help_option = kwargs.get("replace_help_option", "")
+        kwargs.pop("replace_help_option")
+        super(ReplaceHelpSummaryOption, self).__init__(*args, **kwargs)
+
+    def get_help_record(self, ctx):
+        option, help_text = super(ReplaceHelpSummaryOption, self).get_help_record(ctx=ctx)
+        return self.replace_help_option, help_text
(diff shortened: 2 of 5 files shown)


samcli/commands/_utils/custom_options/option_help.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class RequiredOptionColorTextOption(click.Option):
+    def __init__(self, *args, **kwargs):
+        super(RequiredOptionColorTextOption, self).__init__(*args, **kwargs)
+        self.colored = Colored()
+
+    def get_help_record(self, ctx):
+        option, help_text = super(RequiredOptionColorTextOption, self).get_help_record(ctx=ctx)
+        return (self.colored.yellow(option), self.colored.yellow(help_text)) if self.required else (option, help_text)
```
hoffa, line 17: I think the idea is great (especially consider the number of parameters), although a little on the fence whether yellow is ideal. It's typically used to convey warnings and mutations, but at the same time it's a common accent color in SAM CLI. 🤔

sriram-mv: my other thinking is we could make it bold or italic?

.gitignore
```diff
@@ -292,6 +292,9 @@ docs/_build/
 # pyenv
 .python-version
 
+# pyright
+pyrightconfig.json
```
jfuss, line 296: Curious: what is pyrightconfig?

samcli/commands/_utils/custom_options/option_help.py
```diff
@@ -0,0 +1,28 @@
+"""
+Custom click option for customizing help text.
+"""
+
+import click
+
+from samcli.lib.utils.colors import Colored
+
+
+class RequiredOptionColorTextOption(click.Option):
```
jfuss, line 10: Is there a way to can do this on `click` directly? Which would allow us to not have to attach it to individual options?

jfuss (commented):
Have you thought about ordering at all?

How does this impact color blind individuals?

.gitignore
sriram-mv: its my config file for showcasing the virtual env which talks to the LSP server through my neovim.

samcli/commands/_utils/custom_options/option_help.py
sriram-mv: Not with click directly, Yeah I did think about ordering! Terraform cli has a nice way of doing this, showcasing the core options first.

In terms of colors, other ideas are to just to think about bold text instead.

samcli/commands/_utils/custom_options/option_help.py
hoffa: Italic support is sketchy in some terminals/fonts. Making only the option bold but regular text color might be nice as well. To be clear I like colors especially if we can define color guidelines to avoid inconsistent "over-saturation".

PR #4103, opened 2022-08-08
Add option to CfnTag parameter to accept multiple values per key

bentvelj:
With `multiple_values_per_key=True`:
`Key1=Value1 Key1=Value2` -> `{Key1: [Value1, Value2]}`

Without `multiple_values_per_key` specified (before this change):
`Key1=Value1 Key1=Value2` -> `{Key1: Value2}`

N/A

Currently, there is no way for a customer to specify "tag filters" consisting of keys with multiple values. This is useful for SAM CLI commands that make API calls such as [get-resources](https://docs.aws.amazon.com/cli/latest/reference/resourcegroupstaggingapi/get-resources.html#get-resources). If a customer would like to specify two resources with the same tag key but different values, this change makes that possible.

An optional flag is added to the `CfnTags` parameter class. If enabled, the resulting dictionary will map each key to a list of values instead of the latest entered one. If not specified, `CfnTags` will function exactly as it did previously, only accepting one value per key.

None, all uses of CfnTags will function the same as before this change.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -239,7 +239,7 @@ class CfnTags(click.ParamType):
 
         return result
 
-    def _add_value(self, result, key, new_value):
+    def _add_value(self, result : dict, key : str, new_value : str):
         """
         Add a given value to a given key in the result map.
         """
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -239,7 +239,7 @@ class CfnTags(click.ParamType):
 
         return result
 
-    def _add_value(self, result : dict, key : str, new_value : str):
+    def _add_value(self, result: dict, key: str, new_value: str):
         """
         Add a given value to a given key in the result map.
         """
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -178,7 +178,15 @@ class CfnTags(click.ParamType):
     """
     Custom Click options type to accept values for tag parameters.
     tag parameters can be of the type KeyName1=string KeyName2=string
+
+    If multiple_values_per_key is set to True, the returned dictionary will map
+    each key to a list of corresponding values.
+
+    E.g. Input: KeyName1=Value1 KeyName1=Value2 Output: {KeyName1 : [Value1, Value2]}
     """
+
+    def __init__(self, multiple_values_per_key=False):
+        self.multiple_values_per_key = multiple_values_per_key
 
     _EXAMPLE = "KeyName1=string KeyName2=string"
     # Tags have additional constraints and they allow "+ - = . _ : / @" apart from alpha-numerics.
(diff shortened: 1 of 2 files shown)


jfuss (approved):
LGTM

samcli/cli/types.py
```diff
@@ -232,6 +239,17 @@ def convert(self, value, param, ctx):
 
         return result
 
+    def _add_value(self, result, key, new_value):
```
hoffa, line 242: Missing type hints.

samcli/cli/types.py
```diff
@@ -232,6 +239,17 @@ def convert(self, value, param, ctx):
...
+        """
+        Add a given value to a given key in the result map.
+        """
+        if self.multiple_values_per_key:
```
hoffa, line 246: Can this be deduced from the type?

samcli/cli/types.py
bentvelj: Yeah, the whole file is missing type hints, but I've added them for my additions

tests/unit/cli/test_types.py
```diff
@@ -241,6 +241,64 @@ def test_successful_parsing(self, input, expected):
...
+        self.param_type = CfnTags(multiple_values_per_key=True)
+
+    @parameterized.expand(
+        [
+            # Just a string
+            ("some string"),
+            # Wrong notation
+            # ("a==b"),
+            # Wrong multi-key notation
+            # ("a==b,c==d"),
```
hoffa, line 259: Why commented?

samcli/cli/types.py
bentvelj: The values are always strings, `multiple_values_per_key` just determines whether or not to accumulate values in the same key or to replace them.

With `multiple_values_per_key=True`:
Key1=Value1 Key1=Value2 -> {Key1: [Value1, Value2]}

Without `multiple_values_per_key` specified (before this change):
Key1=Value1 Key1=Value2 -> {Key1: Value2}

hoffa (approved):

tests/unit/cli/test_types.py
bentvelj: The comments were in the original `CfnTags` tests, I'm not entirely sure why

samcli/cli/types.py
```diff
@@ -232,6 +239,17 @@ def convert(self, value, param, ctx):
 
         return result
 
+    def _add_value(self, result: dict, key: str, new_value: str):
+        """
+        Add a given value to a given key in the result map.
+        """
+        if self.multiple_values_per_key:
+            if not result.get(key):
```
sriram-mv, line 247: nit: something like defaultdict(list) could be used here, not a blocker.

sriram-mv (approved):

samcli/cli/types.py
bentvelj: The type of `result` would have to change based on whether `multiple_values_per_key` is set though, because if result is a `defaultdict(list)`, we can't do the normal `result[key] = value` if `multiple_values_per_key` is not set.

It's possible, but I think it's a bit more complexity than it's worth.

PR #4104, opened 2022-08-09
Fix send_events lock logic

Leo10Gama:
N/A

While reviewing my code, I noticed that the structure of the `send_events` method was poorly formatted for multi-threading, as the method holds the `event` lock for its entire duration, including when connecting to Telemetry to send the events. This can cause long waits, which we attempted to get rid of by sending events in a separate thread in the first place.

I rearranged the method, accessing the lock only for what is solely necessary. This means keeping the `telemetry.emit` method out of the lock, allowing it to properly execute in parallel if other Events are added to the tracker while sending.

This should cause notable increase in the speed of Events being tracked, especially on days that the Telemetry itself may be slower than usual.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -179,18 +179,19 @@ class EventTracker:
         """Sends the current list of events via Telemetry."""
         from samcli.lib.telemetry.metric import Metric  # pylint: disable=cyclic-import
 
+        msa = {}
+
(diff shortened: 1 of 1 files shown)


hawflau (approved):

torresxb1 (approved):
nice improvement 👌

PR #4111, opened 2022-08-10
Make minor tweaks before launch

Leo10Gama:
N/A

There were a few things I wasn't happy with reflecting on some of my previous code, and they were too small to fit into PRs on their own. Thus, they've been delegated to *this* PR, which has many tiny adjustments and tweaks, outlined below in order of significance.

- `EventTracker.send_events` now executes on a thread! Previously, this logic was only implemented when the function was called at capacity. Now, it's the default!
- `EventTracker` now stores `session_id` locally. This fixes a bug I was only recently made aware of, where Events sent in a thread did not contain the session ID, thus preventing us from linking the data. With the new implementation, the session ID is appended to the Metric before it is sent out.
- The decorators on `sam sync` are rearranged (namely, `track_command` and `track_long_event` have swapped places). This will prevent one `send_events` from being called, then the "end event" being added, and then one more `send_events` being called.
  - This reasoning has also been added to the docstring for the `track_long_event` decorator.
- The attribute `EventType._events` has been renamed to `EventType._event_values`. This not only prevents ambiguity, but also prevents confusion with `EventTracker._events`.
- Added a docstring to `EventTracker.get_tracked_events` method.
- Added a comment beside `EventType._event_values` to provide explanation as to what the value represents.
- Added comments in the `track_long_event` decorator to provide clarity on the process of the method.

Aside from some readability improvements within my code, the only notable side effect is the `EventTracker.send_events` executing on a separate thread always.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/event.py
+++ b/samcli/lib/telemetry/event.py
@@ -8,6 +8,7 @@ import logging
 import threading
 from typing import List
 
+from samcli.cli.context import Context
 from samcli.lib.build.workflows import ALL_CONFIGS
 from samcli.lib.telemetry.telemetry import Telemetry
 from samcli.local.common.runtime_template import INIT_RUNTIMES
@@ -121,6 +122,7 @@ class EventTracker:
 
     _events: List[Event] = []
     _event_lock = threading.Lock()
+    _session_id = None
 
     MAX_EVENTS: int = 50  # Maximum number of events to store before sending
 
@@ -156,6 +158,12 @@ class EventTracker:
             should_send: bool = False
             with EventTracker._event_lock:
                 EventTracker._events.append(Event(event_name, event_value))
+                # Get the session ID (needed for multithreading sending)
+                if not EventTracker._session_id:
+                    try:
+                        EventTracker._session_id = Context.get_current_context().session_id
+                    except RuntimeError:
+                        pass
                 if len(EventTracker._events) >= EventTracker.MAX_EVENTS:
                     should_send = True
             if should_send:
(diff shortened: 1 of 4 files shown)


samcli/lib/telemetry/event.py
```diff
@@ -155,16 +158,24 @@ def track_event(event_name: str, event_value: str):
             should_send: bool = False
             with EventTracker._event_lock:
                 EventTracker._events.append(Event(event_name, event_value))
+                # Get the session ID (needed for multithreading sending)
+                if not EventTracker._session_id:
+                    try:
+                        ctx = Context.get_current_context()
+                        if ctx:
+                            EventTracker._session_id = ctx.session_id
+                    except RuntimeError:
```
hawflau, line 167: nit: adding a debug log would help

hawflau (changes requested):

samcli/lib/telemetry/event.py
```diff
@@ -155,16 +158,25 @@ def track_event(event_name: str, event_value: str):
             should_send: bool = False
             with EventTracker._event_lock:
                 EventTracker._events.append(Event(event_name, event_value))
+                # Get the session ID (needed for multithreading sending)
+                if not EventTracker._session_id:
+                    try:
+                        ctx = Context.get_current_context()
+                        if ctx:
+                            EventTracker._session_id = ctx.session_id
+                            LOG.debug("EventTracker: Obtained session ID '%s'", EventTracker._session_id)
```
hawflau, line 167: Thanks for updating. But I believe it's more useful to log in the failure path (i.e. in the except block) than in the happy path.

hawflau (changes requested):

hawflau (approved):
Thanks for updating the PR! LGTM!

torresxb1 (approved):
💯 thanks for the improvements!

PR #4126, opened 2022-08-17
FEAT: Test Runner - Allow init --tags option to accept multiple values per key

bentvelj:
N/A

The init command makes a [get-resources](https://docs.aws.amazon.com/cli/latest/reference/resourcegroupstaggingapi/get-resources.html) call with a tag filter, to get the ARNs of customer application resources with the specified tag.

It is possible that customers have resources tagged with the same key, but different values and would like to both to be discovered by the Test Runner (for IAM policy & Environment Variable generation).

Related: https://github.com/aws/aws-sam-cli/pull/4103

Makes use of the `multiple_values_per_key` option for the `CfnTags` type.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/tests/unit/cli/test_types.py
+++ b/tests/unit/cli/test_types.py
@@ -241,6 +241,64 @@ class TestCfnTags(TestCase):
         self.assertEqual(result, expected, msg="Failed with Input = " + str(input))
 
 
+class TestCfnTagsMultipleValues(TestCase):
+    """
+    Tests for the CfnTags parameter allowing multiple values per key.
+    """
+
+    def setUp(self):
+        self.param_type = CfnTags(multiple_values_per_key=True)
+
+    @parameterized.expand(
+        [
+            # Just a string
+            ("some string"),
+            # Wrong notation
+            # ("a==b"),
+            # Wrong multi-key notation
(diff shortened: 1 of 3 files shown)


tests/unit/cli/test_types.py
```diff
@@ -241,6 +241,64 @@ def test_successful_parsing(self, input, expected):
...
+        self.param_type = CfnTags(multiple_values_per_key=True)
+
+    @parameterized.expand(
+        [
+            # Just a string
+            ("some string"),
+            # Wrong notation
+            # ("a==b"),
+            # Wrong multi-key notation
+            # ("a==b,c==d"),
```
acristin, line 259: Should line 257 and 259 be commented?

acristin (approved):

tests/unit/cli/test_types.py
bentvelj: Those are commented in many other tests in that file, I've only included them for consistency. Maybe they're going to be re-enabled in the future? If we remove these I think we should remove them all.

jfuss (approved):

PR #4130, opened 2022-08-18
FEAT: Test Runner - Add --allow-iam option to init command

bentvelj:
N/A

This option improves UX by providing a shortcut to customers who are confident in the generated IAM permissions, and do not want to go through and "enable" them manually by un-commenting them.

A TODO: Having readily available public documentation describing which resources will have which IAM actions generated for them. This is vital knowledge for the customer to be aware of which actions they are granting by enabling `--allow-iam`.

Includes code to generate IAM actions that are not commented out in the Test Runner CloudFormation Template if the `--allow-iam` flag is present for the `init` command.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/tests/unit/commands/test_runner/test_init_command.py
+++ b/tests/unit/commands/test_runner/test_init_command.py
@@ -8,6 +8,8 @@ import botocore.exceptions
 from samcli.commands.exceptions import NoResourcesMatchGivenTagException
 from samcli.commands.test_runner.init.cli import do_cli, query_tagging_api
 
+from samcli.yamlhelper import parse_yaml_file
+
 
 class TestCli(TestCase):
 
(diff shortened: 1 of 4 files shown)


tests/unit/commands/test_runner/test_init_command.py
```diff
@@ -123,10 +128,38 @@ def test_valid_tagging_api_response(self, boto_client_provider_patch, query_tagg
...
+                template_name=self.TEST_TEMPLATE_NAME,
+                env_file=self.TEST_ENV_FILE_NAME,
+                image_uri="test_image_uri",
+                allow_iam=True,
+                runtime="python3.8",
+            )
+            self.assertTrue(os.path.exists(self.TEST_TEMPLATE_NAME))
+            self.assertTrue(os.path.exists(self.TEST_ENV_FILE_NAME))
+
+            self.assertFalse("#" in Path.read_text(Path(self.TEST_TEMPLATE_NAME)))
```
acristin, line 162: This seems brittle if we later add comments to the yaml file; would it be better to assert that the actions do/don't show up if you parse the yaml?

acristin (approved):
I'm not sure I like the `--allow-iam` naming, but the change seems good otherwise.

tests/unit/commands/test_runner/test_init_command.py
bentvelj: Good point, I think so

bentvelj:
> I'm not sure I like the `--allow-iam` naming, but the change seems good otherwise.

Yeah, I'll do some brainstorming to come up with a better name. Maybe `--allow-actions` or `--enable-actions` or `--enable-iam-actions` (kinda long)?

We can discuss with UX and come up with solid names for this option (every other one).

acristin (approved):

jfuss (changes requested):
Seems reasonable. I know we chatted offline about this. I am ok accepting now with the understanding we will look deeper into this in the future and see if we can find ways to make this easier on the customer and have a deeper review on IAM resources created.

jfuss (approved):

Issue #1045, opened 2019-03-08
Specify globals in --env-vars file


terlar:
### Describe your idea/feature/enhancement

I wish SAM CLI JSON file provided by `--env-vars` had support for globals. In our `template.yaml` we specify certain environment variables via `Globals: Function: Environment: Variables`. It would be nice if such shared variables could be added in one place inside the JSON file you can provide to `--env-vars`. Currently as far as I understand it you can only specify env vars per function in this file.

### Proposal

Add something like `Globals`, `Function` or similar to the JSON file. What do you think? Any other workarounds? Now I frequently get caught in missing specifying this thing when adding a new function as I rely on the globals for that in the regular template.

mnapoli:
@terlar I tested this and it worked:

```json
{
    "Parameters": {
        "BAR": "helloworld"
    }
}
```

That seems to define the `BAR` environment variable for all functions.

jfuss:
@terlar Thanks for the issue. You can achieve this though the option @mnapoli has shown. We currently support two formats but you cannot intermix them. So if two functions have the same Env Var but you wanted different values, you would need to specify with the LogicalIds instead of Parameters. Marking this as a feature request as I can see some use-cases where it could be beneficial to have some kind of mixed format.

[Here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/local/lib/local_lambda.py#L175) is the spot in the code where we parse the env vars file contents.

pwrmiller:
Can this be issue be modified to be a documentation change? Specifically, [this section](https://github.com/awslabs/aws-sam-cli/blob/develop/docs/advanced_usage.md) of documentation needs to document the `Parameters` option.

ozbillwang:
Seems `--env-vars` is only available for `sam local invoke`, not for `sam deploy`. 

Could we have the same for `sam deploy` command? So I can easily manage the parameter overrides with a readible Json file.

qingchm:
@ozbillwang To understand the requirement here, the --env-vars option is provided because invoke happens inside a container. Same for build inside containers we support --container-env-var-file option. If you want to use env vars for deploy what's blocking you from directly setting the env vars and reading it? Deploy does not create an isolated execution environment AFAIK

qingchm:
[...]

PR #4132, opened 2022-08-18
feat: Allow mix of global and function specific env vars for sam local

hnnasit:

#1045 

Users cannot provide both global and function specific environment variables
in the env-vars json file for sam local.

This change allows users to use a mix of cloudformation and standard format for
environment variables

None found

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -238,14 +238,15 @@ class LocalLambdaRunner:
         """
 
         function_id = function.function_id
-        name = function.name
+        logical_id = function.name
+        function_name = function.functionname
         full_path = function.full_path
 
         variables = None
         if isinstance(function.environment, dict) and "Variables" in function.environment:
             variables = function.environment["Variables"]
         else:
-            LOG.debug("No environment variables found for function '%s'", name)
+            LOG.debug("No environment variables found for function '%s'", logical_id)
 
         # This could either be in standard format, or a CloudFormation parameter file format, or mix of both.
         #
@@ -268,10 +269,11 @@ class LocalLambdaRunner:
             parameter_result = self.env_vars_values.get("Parameters", {})
             overrides.update(parameter_result)
 
-        # Precedence: logical_id -> function_id -> full_path, customer can use any of them
+        # Precedence: logical_id -> function_id -> function name -> full_path, customer can use any of them
         fn_file_env_vars = (
-            self.env_vars_values.get(name, None)
+            self.env_vars_values.get(logical_id, None)
             or self.env_vars_values.get(function_id, None)
+            or self.env_vars_values.get(function_name, None)
             or self.env_vars_values.get(full_path, None)
         )
         if fn_file_env_vars:
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -251,6 +251,7 @@ class LocalLambdaRunner:
         #
         # Standard format is {FunctionName: {key:value}, FunctionName: {key:value}}
         # CloudFormation parameter file is {"Parameters": {key:value}}
+        # Mixed format is {FunctionName: {key:value}, "Parameters": {key:value}}
 
         for env_var_value in self.env_vars_values.values():
             if not isinstance(env_var_value, dict):
--- a/samcli/commands/local/lib/local_lambda.py
+++ b/samcli/commands/local/lib/local_lambda.py
@@ -216,6 +216,10 @@ class LocalLambdaRunner:
     def _make_env_vars(self, function: Function) -> EnvironmentVariables:
         """Returns the environment variables configuration for this function
 
+        Priority order for environment variables (high to low):
+        1. Function specific env vars from json file
+        2. Global env vars from json file
+
         Parameters
         ----------
         function : samcli.commands.local.lib.provider.Function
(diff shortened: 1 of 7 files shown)


samcli/commands/local/lib/local_lambda.py
```diff
@@ -243,33 +247,36 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
         else:
             LOG.debug("No environment variables found for function '%s'", name)
 
-        # This could either be in standard format, or a CloudFormation parameter file format.
+        # This could either be in standard format, or a CloudFormation parameter file format, or mix of both.
         #
         # Standard format is {FunctionName: {key:value}, FunctionName: {key:value}}
         # CloudFormation parameter file is {"Parameters": {key:value}}
```
qingchm, line 253: Can we should an example of the mixed format here?

qingchm (approved):
LGTM, please update the commenting and it's good to go

samcli/commands/local/lib/local_lambda.py
```diff
@@ -243,33 +247,37 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
...
         if "Parameters" in self.env_vars_values:
-            LOG.debug("Environment variables overrides data is in CloudFormation parameter file format")
+            LOG.debug("Environment variables data found in the CloudFormation parameter file format")
             # CloudFormation parameter file format
-            overrides = self.env_vars_values["Parameters"]
-        else:
+            parameter_result = self.env_vars_values.get("Parameters", {})
+            overrides.update(parameter_result)
+
+        # Precedence: logical_id -> function_id -> full_path, customer can use any of them
```
moelasmar, line 271: shall we enable the customer to use the Lambda function name as well?

samcli/commands/local/lib/local_lambda.py
```diff
@@ -243,33 +247,37 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
...
             if not isinstance(env_var_value, dict):
-                reason = """
-                            Environment variables must be in either CloudFormation parameter file
-                            format or in {FunctionName: {key:value}} JSON pairs
-                            """
+                reason = "Environment variables {} in incorrect format".format(env_var_value)
                 LOG.debug(reason)
                 raise OverridesNotWellDefinedError(reason)
 
+        overrides = {}
```
moelasmar, line 262: lets keep overrides to be None if the customer does not provide any env vars.

moelasmar (changes requested):
looks good to me, just add small comment

samcli/commands/local/lib/local_lambda.py
```diff
@@ -243,33 +247,37 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
         else:
             LOG.debug("No environment variables found for function '%s'", name)
 
-        # This could either be in standard format, or a CloudFormation parameter file format.
+        # This could either be in standard format, or a CloudFormation parameter file format, or mix of both.
         #
         # Standard format is {FunctionName: {key:value}, FunctionName: {key:value}}
         # CloudFormation parameter file is {"Parameters": {key:value}}
+        # Mixed format is {FunctionName: {key:value}, "Parameters": {key:value}}
```
hawflau, line 254: What happens if a function name is `Parameters`?

hawflau (changes requested):

samcli/commands/local/lib/local_lambda.py
moelasmar: I had the same question, but it seems it is how we currently handle global parameters. If we changed it, we may break our customers

hnnasit: Yes I agree with @moelasmar. I checked and sam build handles the global env variables in a similar way [here](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L934-L948).  If a function name was Parameters then with the current code it would not allow reading specific env variables for other lambda functions. With the new changes if function specific env vars are given, they would have precedence over the "Parameters" env vars which is similar to sam build.

samcli/commands/local/lib/local_lambda.py
hnnasit: I thought it was redundant to set overrides to None because when overrides is passed to EnvironmentVariables by default it is set to empty dict and lint throws and error when I try to reassign overrides to None if it is empty. Is there a particular reason it should be None?

moelasmar: thanks @hnnasit for the clarification. I think it is Ok to keep it empty dict as we already set to empty dict during the env vars processing, so no issue.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -234,42 +238,48 @@ def _make_env_vars(self, function: Function) -> EnvironmentVariables:
...
-            overrides = self.env_vars_values["Parameters"]
-        else:
+            parameter_result = self.env_vars_values.get("Parameters", {})
+            overrides.update(parameter_result)
+
+        # Precedence: logical_id -> function_id -> function name -> full_path, customer can use any of them
+        fn_file_env_vars = (
+            self.env_vars_values.get(logical_id, None)
+            or self.env_vars_values.get(function_id, None)
+            or self.env_vars_values.get(function_name, None)
```
moelasmar, line 276: we need to update our docs to let the customers know that they can use function name in the env vars file.

moelasmar (approved):
it looks good to me, thanks Haresh for your contribution.

samcli/commands/local/lib/local_lambda.py
hnnasit: I will add this to the doc update ticket. Thanks @moelasmar

hawflau (approved):

PR #4133, opened 2022-08-18
feat: Disable ADL for esbuild

mildaniel:
Speeds up sam sync for esbuild by disabling auto-dependency layer.

Also moves some existing code to be more consistent with where we handle template post-processing today.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -10,6 +10,7 @@ from typing import Dict, Optional, List, Tuple, cast
 import click
 
 from samcli.commands._utils.experimental import ExperimentalFlag, prompt_experimental
+from samcli.lib.build.bundler import EsbuildBundlerManager
 from samcli.lib.providers.sam_api_provider import SamApiProvider
 from samcli.lib.utils.packagetype import IMAGE
 
@@ -34,6 +35,7 @@ from samcli.lib.build.app_builder import (
     BuildError,
     UnsupportedBuilderLibraryVersionError,
     ContainerBuildNotSupported,
+    ApplicationBuildResult,
 )
 from samcli.commands._utils.options import DEFAULT_BUILD_DIR
 from samcli.lib.build.workflow_config import UnsupportedRuntimeException
@@ -222,9 +224,6 @@ class BuildContext:
         if is_sam_template:
             SamApiProvider.check_implicit_api_resource_ids(self.stacks)
 
-        # modify the stack resources to support source maps
-        self._enable_source_maps()
-
         try:
             builder = ApplicationBuilder(
                 self.get_resources_to_build(),
@@ -262,12 +261,7 @@ class BuildContext:
                 )
                 output_template_path = stack.get_output_template_path(self.build_dir)
 
-                if self._create_auto_dependency_layer:
-                    LOG.debug("Auto creating dependency layer for each function resource into a nested stack")
-                    nested_stack_manager = NestedStackManager(
-                        stack, self._stack_name, self.build_dir, modified_template, build_result
-                    )
-                    modified_template = nested_stack_manager.generate_auto_dependency_layer_stack()
+                modified_template = self._handle_template_post_processing(stack, modified_template, build_result)
 
                 move_template(stack.location, output_template_path, modified_template)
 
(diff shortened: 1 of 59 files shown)


samcli/lib/build/bundler.py
```diff
@@ -0,0 +1,143 @@
...
+
+ESBUILD_PROPERTY = "esbuild"
+
+
+class EsbuildBundlerManager:
+    def __init__(self, stack: Stack, template: Optional[Dict] = None):
+        self._stack = stack
+        if template is None:
+            template = dict()
+        self._previous_template = template
```
lucashuy, line 21: nit/question: this vs `self._previous_template = template or dict()`?

mildaniel: Updated.

samcli/lib/build/bundler.py
```diff
@@ -0,0 +1,141 @@
...
+        using_source_maps = False
+        invalid_node_option = False
+
+        template = deepcopy(self._previous_template)
+        resources = template.get("Resources", {})
+
+        for name, resource in resources.items():
+            metadata = resource.get("Metadata", {})
+
+            node_option_set = self._is_node_option_set(resource)
```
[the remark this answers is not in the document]
mildaniel: Good call-out.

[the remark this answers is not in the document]
mildaniel: The other issue with this particular use-case is that it needs to be pre-processed and not post. Otherwise the flag won't be set to true in time. I'll have to adjust this too.

lucashuy (approved):

samcli/commands/sync/command.py
```diff
@@ -415,3 +419,20 @@ def execute_watch(
...
+    :param template_file: template file string
+    :return: True if ADL should be enabled, False otherwise
+    """
+    stacks, _ = SamLocalStackProvider.get_stacks(template_file)
+    for stack in stacks:
+        esbuild = EsbuildBundlerManager(stack)
+        if esbuild.esbuild_configured():
+            # Disable ADL if esbuild is configured. esbuild already makes the package size
+            # small enough to ensure that ADL isn't needed to improve performance
+            click.secho("esbuild is configured, disabling auto dependency layer.", fg="yellow")
```
torresxb1, line 436: discussed through huddle: potentially display some explanation to users on why ADL is disabled for esbuild?

samcli/commands/build/build_context.py
```diff
@@ -263,12 +265,7 @@ def run(self):
                 )
                 output_template_path = stack.get_output_template_path(self.build_dir)
 
-                if self._create_auto_dependency_layer:
-                    LOG.debug("Auto creating dependency layer for each function resource into a nested stack")
-                    nested_stack_manager = NestedStackManager(
-                        stack, self._stack_name, self.build_dir, modified_template, build_result
-                    )
-                    modified_template = nested_stack_manager.generate_auto_dependency_layer_stack()
+                modified_template = self._handle_build_post_processing(stack, modified_template, build_result)
```
torresxb1, line 268: discussed in huddle: some of the lines above could also go into the post_processing function

tests/integration/sync/test_sync_adl.py
```diff
@@ -166,3 +168,42 @@ def _verify_lambda_response(_lambda_response):
...
+        self.assertIn("Sync infra completed.", str(sync_process_execute.stderr))
+
+        self.stack_resources = self._get_stacks(stack_name)
+        lambda_functions = self.stack_resources.get(AWS_LAMBDA_FUNCTION)
+        for lambda_function in lambda_functions:
+            lambda_response = json.loads(self._get_lambda_response(lambda_function))
+            self.assertEqual(lambda_response.get("message"), "hello world")
+
+        layers = self.stack_resources.get(AWS_LAMBDA_LAYERVERSION)
+        self.assertIsNone(layers)
```
torresxb1, line 209: should there be another test where there are other layers and we check that those are still there, but no dependency layer?

torresxb1 (approved):

tests/integration/sync/test_sync_adl.py
mildaniel: Good idea. I'll include this in a separate PR since this one is getting a little large.

PR #4136, opened 2022-08-19
Add hook_package_id option to sam build

hawflau:

NA

To add hook_package_id option to `sam build` command.
Note that we want the prepare to always run in `sam build`, thus setting `force_prepare` to True

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


moelasmar (approved):

samcli/commands/build/command.py
```diff
@@ -222,6 +229,7 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     container_env_var_file: Optional[str],
     build_image: Optional[Tuple[str]],
     exclude: Optional[Tuple[str, ...]],
+    hook_package_id: Optional[str],
```
torresxb1, line 232: why is the type hint different here than in the `cli` function? (shouldn't it be optional in `cli`?)

torresxb1 (approved):
approving with question

samcli/commands/build/command.py
hawflau: good catch, it should be optional there too. will update

PR #4139, opened 2022-08-23
Add Table of Generated IAM Actions

bentvelj:
N/A

Shows the customer which IAM actions are being generated for which of their testable resources.

This can serve as a substitute for commenting out permissions. Instead of having the customer go through and enable the permissions they want to grant, permissions can be enabled by default, and the customer can see exactly what is being granted.

Prints a table when the `sam test-runner init` command is executed
E.g.
image

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


jfuss (approved):
LGTM

samcli/commands/test_runner/init/cli.py
```diff
@@ -233,3 +250,56 @@ def query_tagging_api(tags: dict, boto_client_provider: BotoProviderType) -> Uni
     )
 
     return [resource.get("ResourceARN") for resource in resource_tag_mapping_list]
+
+
+def _extract_action(iam_action_string: str) -> str:
+    indx = iam_action_string.index(":")
```
acristin, line 283: Is there a reason for `indx` as opposed to `index`?

bentvelj: nope, will fix :)

acristin (approved):

PR #4153, opened 2022-08-25
feat: do metadata resource type property validations

moelasmar:
Add the sam metadata resource type property validations.

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [ ] Write/update integration tests
- [ ] Write/update functional tests if needed
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #4154, opened 2022-08-25
feat: get the cfn resource that match the sam metadata resource name property

moelasmar:
Use the sam metadata resource_name property to retrieve the CFN resource to be enriched. We first try to use this value as the resource only postfix address, so we should concatenate it to the module address to get the full address. If we could not find a resource that match the previous address, so may be the resource name value is a full address, so we will try to use it without any concatenation. Finally if we did not find any resources, we will raise an error.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -7,7 +7,7 @@ import json
 import os
 from pathlib import Path
 from subprocess import run, CalledProcessError
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 import hashlib
 import logging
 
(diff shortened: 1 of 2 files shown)


torresxb1:
[For next reviewer]
This PR builds on top of another: https://github.com/aws/aws-sam-cli/pull/4153
Before the other PR is merged, see the actual new changes from this PR here: https://github.com/moelasmar/aws-sam-cli/compare/develop-terraform-hooks-do-source-type-validation...moelasmar:aws-sam-cli:develop-terraform-hooks-get-the-cfn-resource-that-match-the-resource-name-property

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -206,9 +243,174 @@ def _translate_to_cfn(tf_json: dict) -> dict:
...
+    if not resource_name:
+        raise InvalidSamMetadataProperties(
+            f"sam cli expects the sam metadata resource {sam_metadata_resource_address} to contain a resource name "
+            f"that will be enriched using this metadata resource"
+        )
+
+    # the provided resource name can be the name without the module address, or can be a complete resource address name
+    # check first if the provided name is without the module address
+    LOG.info(
+        "Check if the input source name %s is a postfix to the current module address %s",
```
torresxb1, line 382: I think let's change `source` to `resource`, to be consistent with the other info messages here.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -206,9 +243,174 @@ def _translate_to_cfn(tf_json: dict) -> dict:
...
+        else resource_name
+    )
+    LOG.debug("check if the resource address %s has a relevant cfn resource or not", full_resource_address)
+    logical_id = _build_cfn_logical_id(full_resource_address)
+    cfn_resource = cfn_resources.get(logical_id)
+    if cfn_resource:
+        LOG.info("The CFN resource that match the input resource name %s is %s", resource_name, logical_id)
+        return cfn_resource, logical_id
+
+    LOG.info("Check if the input source name %s is a complete address", resource_name)
```
torresxb1, line 398: I think let's change `source` to `resource`, to be consistent with the other info messages here.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -206,9 +243,174 @@ def _translate_to_cfn(tf_json: dict) -> dict:
...
+    # check if the provided name is a complete resource address
+    if sam_metadata_resource.current_module_address:
+        full_resource_address = resource_name
+        LOG.debug("check if the resource address %s has a relevant cfn resource or not", full_resource_address)
+        logical_id = _build_cfn_logical_id(full_resource_address)
+        cfn_resource = cfn_resources.get(logical_id)
+        if cfn_resource:
+            LOG.info("The CFN resource that match the input resource name %s is %s", resource_name, logical_id)
+            return cfn_resource, logical_id
```
torresxb1, line 407: I'm a little bit confused, I feel like this is the same as above logic? (shouldn't the above be able to handle this case?)

moelasmar: it is not similar, the first case we try to build a full address assuming that the resource name is just the address prefix, and we should add the module address to it. This one is assuming that the resource name is a full address by it self.

The only case that both scenarios will be the same is in case of root modules, that is why I add the condition 
 ```
  if sam_metadata_resource.current_module_address:
```

to make sure that I am not doing the same logic twice.

hawflau (approved):

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: discussed outside this thread: when defining the metadata, customer's won't be able to put the full address (the metadata is defined in the module, but the full address will include an instance of the module). So, the second check shouldn't be reached.

moelasmar: done

torresxb1 (approved):

PR #4155, opened 2022-08-25
feat: extract the lambda function source code or docker context paths

moelasmar:
SAM Metadata will contain specific properties for the lambda function source code or docker context. The value of this property can be either string, array, or encoded JSON object. Prepare hook will  try to decode the value of this property, and then check the output type. Based on each type, the hook will try to get the required path, and do some validations if metadata property is not as expected.

In case if this property value is an array, the hook will use only the first item as the required path, as SAM CLI currently only supports functions that refer to the source code directory, and does not support referring to an array of files.

in case if this property value is an encoded JSON object, the hook will use another property value to get the required path from this object.

The hook will also validate that the retrieved path is a string, and an exist path.
 
#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -5,6 +5,7 @@ Terraform prepare hook implementation
 from dataclasses import dataclass
 import json
 import os
+from json.decoder import JSONDecodeError
 from pathlib import Path
 from subprocess import run, CalledProcessError
 from typing import Any, Callable, Dict, List, Optional, Tuple
@@ -353,7 +354,7 @@ def _get_lambda_function_source_code_path(
             LOG.debug("Try to decode the %s value in case if it is a encoded JSON string.", src_code_attribute_name)
             source_code = json.loads(source_code)
             LOG.debug("The decoded value of the %s value is %s", src_code_attribute_name, source_code)
-        except Exception:
+        except JSONDecodeError:
             LOG.debug("Source code value could not be parsed as a JSON object. Handle it as normal string value")
 
     if isinstance(source_code, dict):
@@ -383,7 +384,10 @@ def _get_lambda_function_source_code_path(
             )
     elif isinstance(source_code, list):
         # SAM CLI does not process multiple paths, so we will handle only the first value in this list
-        LOG.debug("Process the extracted %s as list, and get the first value", src_code_attribute_name)
+        LOG.debug(
+            "Process the extracted %s as list, and get the first value as SAM CLI does not support multiple paths",
+            src_code_attribute_name,
+        )
         if len(source_code) < 1:
             raise InvalidSamMetadataPropertiesException(
                 f"The sam metadata resource {sam_metadata_resource_address} should contain the lambda function "
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -301,6 +301,124 @@ def _validate_referenced_resource_matches_sam_metadata_type(
         )
 
 
+def _get_lambda_function_source_code_path(
+    sam_metadata_attributes: dict,
+    sam_metadata_resource_address: str,
+    project_root_dir: str,
+    src_code_property_name: str,
+    property_path_property_name: str,
+    src_code_attribute_name: str,
+) -> str:
+    """
+    Validate that sam metadata resource contains the valid metadata properties to get a lambda function source code.
+    Parameters
+    ----------
+    sam_metadata_attributes: dict
+        The sam metadata properties
+    sam_metadata_resource_address: str
+        The sam metadata resource address
+    project_root_dir: str
+        the terraform project root directory path
+    src_code_property_name: str
+        the sam metadata property name that contains the lambda function source code or docker context path
+    property_path_property_name: str
+        the sam metadata property name that contains the property to get the source code value if it was provided
+        as json string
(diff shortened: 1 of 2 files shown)


torresxb1:
[For next reviewer]
This PR builds on top of another: https://github.com/aws/aws-sam-cli/pull/4154
Before the other PR is merged, see the actual new changes from this PR here: https://github.com/moelasmar/aws-sam-cli/compare/develop-terraform-hooks-get-the-cfn-resource-that-match-the-resource-name-property...moelasmar:aws-sam-cli:develop-terraform-hooks-extract-the-lambda-resource-source-directory

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -301,6 +301,124 @@ def _validate_referenced_resource_matches_sam_metadata_type(
...
+                src_code_attribute_name,
+                source_code,
+            )
+            raise InvalidSamMetadataPropertiesException(
+                f"The sam metadata resource {sam_metadata_resource_address} should contain a valid lambda function "
+                f"{src_code_attribute_name} property in property {property_path_property_name} as the "
+                f"{src_code_property_name} value is an object"
+            )
+    elif isinstance(source_code, list):
+        # SAM CLI does not process multiple paths, so we will handle only the first value in this list
```
torresxb1, line 385: should we say this in an info log to inform users?

tests/unit/hook_packages/terraform/test_prepare_hook.py
```diff
@@ -1020,6 +1022,254 @@ def test_validate_referenced_resource_matches_sam_metadata_type_invalid_types(
...
+            ('"/src/code/path"', None, "/src/code/path", True),
+            ('"src/code/path"', None, "src/code/path", False),
+            ('{"path":"/src/code/path"}', "path", "/src/code/path", True),
+            ('{"path":"src/code/path"}', "path", "src/code/path", False),
+            ({"path": "/src/code/path"}, "path", "/src/code/path", True),
+            ({"path": "src/code/path"}, "path", "src/code/path", False),
+            ('["/src/code/path"]', "None", "/src/code/path", True),
+            ('["src/code/path"]', "None", "src/code/path", False),
+            (["/src/code/path"], "None", "/src/code/path", True),
+            (["src/code/path"], "None", "src/code/path", False),
```
torresxb1, line 1038: can we add a case with multiple items in the list?

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -301,6 +301,124 @@ def _validate_referenced_resource_matches_sam_metadata_type(
...
+        The sam metadata properties
+    sam_metadata_resource_address: str
+        The sam metadata resource address
+    project_root_dir: str
+        the terraform project root directory path
+    src_code_property_name: str
+        the sam metadata property name that contains the lambda function source code or docker context path
+    property_path_property_name: str
+        the sam metadata property name that contains the property to get the source code value if it was provided
+        as json string
```
torresxb1, line 326: why are these two customizable?

moelasmar: sorry .. did not get your comment.

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: hmmm .. not sure, I was thinking of documenting this behaviour, but we can add logs as well if the list contains more than 1 item.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -301,6 +301,124 @@ def _validate_referenced_resource_matches_sam_metadata_type(
...
+        raise InvalidSamMetadataPropertiesException(
+            f"The sam metadata resource {sam_metadata_resource_address} should contain the lambda function "
+            f"{src_code_attribute_name} in property {src_code_property_name}"
+        )
+    if isinstance(source_code, str):
+        try:
+            LOG.debug("Try to decode the %s value in case if it is a encoded JSON string.", src_code_attribute_name)
+            source_code = json.loads(source_code)
+            LOG.debug("The decoded value of the %s value is %s", src_code_attribute_name, source_code)
+        except Exception:
```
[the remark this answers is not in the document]
moelasmar: sure

tests/unit/hook_packages/terraform/test_prepare_hook.py
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare.py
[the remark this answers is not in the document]
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: Done

hawflau (approved):

torresxb1 (approved):

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: discussed outside this, but I believe I was confused into why it wouldn't always be the same property name for these. It is because this function can be used for image and zip types, which will have different property names. We could potentially add some type hints for the valid values but yeah not sure how much benefit/value that brings

PR #4157, opened 2022-08-25
feat: enrich image lambda functions

moelasmar:
enrich the Image lambda functions with the data included in the sam metadata resource.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -15,7 +15,7 @@ import logging
 from samcli.lib.hook.exceptions import PrepareHookException, InvalidSamMetadataPropertiesException
 from samcli.lib.utils import osutils
 from samcli.lib.utils.hash import str_checksum
-from samcli.lib.utils.packagetype import ZIP
+from samcli.lib.utils.packagetype import ZIP, IMAGE
 from samcli.lib.utils.resources import (
     AWS_LAMBDA_FUNCTION as CFN_AWS_LAMBDA_FUNCTION,
 )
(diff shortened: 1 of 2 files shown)


samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -323,10 +441,102 @@ def _enrich_mapped_resources(
...
+        if cfn_docker_build_args_string:
+            try:
+                LOG.debug("Parse the docker build args %s", cfn_docker_build_args_string)
+                cfn_docker_build_args = json.loads(cfn_docker_build_args_string)
+                if not isinstance(cfn_docker_build_args, dict):
+                    raise InvalidSamMetadataPropertiesException(
+                        f"The sam metadata resource {sam_metadata_resource_address} should contain a valid json "
+                        f"encoded string for the lambda function docker build arguments."
+                    )
+            except Exception as exc:
```
hawflau, line 519: can we have a specific exception type to catch?

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -323,10 +441,102 @@ def _enrich_mapped_resources(
...
+                raise InvalidSamMetadataPropertiesException(
+                    f"The sam metadata resource {sam_metadata_resource_address} should contain a valid json encoded "
+                    f"string for the lambda function docker build arguments."
+                ) from exc
+
+        cfn_docker_tag = sam_metadata_attributes.get("docker_tag")
+
+        if cfn_resource_properties.get("Code"):
+            cfn_resource_properties.pop("Code")
```
torresxb1, line 528: why do we do this?

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: We do not need the Code attribute where we already mapped the image uri. This will be added during the build step.

torresxb1: hmmm but do we need to remove it or are you just removing it because it won't be used? Could we add a comment in the code with an explanation maybe, in case somebody else wonders this?

torresxb1 (approved):
approving with a comment

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: We need to remove the `Code.IamgeUri` property for the Lambda function as SAM CLI will use theImage Build metadata properties to build a local image uri for this function to be used in local testing. I do not want SAM CLI to use the provided image uri, and so the customer will not be able to change its code, and then test it locally.

moelasmar: let me add some comment to explain that in the code

PR #4159, opened 2022-08-26
UX Polishes

bentvelj:
Help message rewords, option renames, removed colours except for tables and test failures.

Also fixed typo in `test_no_fail_on_no_update`, where the error message was in the stack name instead of the message.

N/A

Improved customer UX

Changed help message wording and options to be more intuitive, reduced use of colour for more consistent experience.

None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/test_runner/init/cli.py
+++ b/samcli/commands/test_runner/init/cli.py
@@ -14,19 +14,22 @@ from samcli.commands._utils.table_print import newline_per_item, pprint_column_n
 from samcli.lib.utils.boto_utils import BotoProviderType
 from samcli.lib.utils.colors import Colored
 
-SHORT_HELP = "Generates a Test Runner CloudFormation Template in YAML format, as well as a resource-ARN map."
+SHORT_HELP = """
+Generates a Test Runner CloudFormation Template file in YAML format. It also generates a resource-ARN map YAML file, where the keys are your resource
+names in the format of an environment variable name, and the values are the corresponding resource ARN."""
+
 HELP_TEXT = """
 This command generates a CloudFormation Template file that can be deployed to instantiate a Test Runner Stack.
 
-It also generates a resource-ARN map, where the keys are your resource names in the format of an environment variable name, and the values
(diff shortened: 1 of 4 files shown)


samcli/commands/test_runner/init/cli.py
```diff
@@ -14,19 +14,22 @@
...
 are the corresponding resource ARN. 
 
 For example, 
 
 _MY_LAMBDA_FUNCTION : arn:aws:lambda:us-east-1:123456789012:function:my-lambda-function
 
-This resource-ARN map can be passed to the `sam test-runner run` command which will expose these mappings as environment variables to the Test Runner Fargate instance,
-and thus can be used in test code.
+This file can be passed to `sam test-runner run` (using the `--container-vars` option), to expose the mappings as environment variables to the Fargate container.
+These environment variables can be picked up and used by your test code.
```
acristin, line 32: > can be picked up and

Is this wording precise enough? It might be better to phrase like "These environment variables are available to your test suite" or "These environment variables are exported to your test suite."

acristin (approved):
Overall fine. Minor question around wording of environment variables & test suite, but non-blocking.

samcli/commands/test_runner/init/cli.py
bentvelj: Yeah, I agree it's a bit too colloquial. I like the "These environment variables are available to your test suite", but we should probably stay consistent with the `exposed` verb, so maybe "These environment variables are exposed to your testsuite"?

jfuss: @bentvelj Either of them sound fine honestly. I would pick the one you think it best. No need to overthink it, it's something we can always adjust later too. :)

bentvelj: Yep! Easy change, fixed to "These environment variables are exposed to your testsuite" :)

jfuss (approved):

acristin (approved):

PR #4160, opened 2022-08-26
fix: Fix Sourcemap Generation

mildaniel:
Sourcemaps now aren't being generated by the build if only set in the NODE_OPTIONS. This is a bug that occurs because the function provider isn't refreshed when the stack in the build context is updated. 

This change forces the function provider to update its contents after the stack has been modified.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -10,6 +10,7 @@ from typing import Dict, Optional, List, Tuple, cast
 import click
 
 from samcli.commands._utils.experimental import ExperimentalFlag, prompt_experimental
+from samcli.lib.build.bundler import EsbuildBundlerManager
 from samcli.lib.providers.sam_api_provider import SamApiProvider
 from samcli.lib.utils.packagetype import IMAGE
 
@@ -34,6 +35,7 @@ from samcli.lib.build.app_builder import (
     BuildError,
     UnsupportedBuilderLibraryVersionError,
     ContainerBuildNotSupported,
+    ApplicationBuildResult,
 )
 from samcli.commands._utils.options import DEFAULT_BUILD_DIR
 from samcli.lib.build.workflow_config import UnsupportedRuntimeException
@@ -222,9 +224,6 @@ class BuildContext:
         if is_sam_template:
             SamApiProvider.check_implicit_api_resource_ids(self.stacks)
 
-        # modify the stack resources to support source maps
-        self._enable_source_maps()
-
         try:
             builder = ApplicationBuilder(
                 self.get_resources_to_build(),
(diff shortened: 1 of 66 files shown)


samcli/lib/providers/sam_function_provider.py
```diff
@@ -79,6 +79,26 @@ def stacks(self) -> List[Stack]:
         """
         return self._stacks
 
+    def update(
```
lucashuy, line 82: Would it be possible to call this `update()` function in `__init__()` to remove the repeat code? Otherwise this LGTM

mildaniel: Not sure if I'm following. The provider doesn't get re-instantiated after the first time we instantiate it in the `BuildContext`, so how would it update?

lucashuy: Its more so that the new update function and __init__ both extract the functions from the stack, so I was wondering if we could just call `self.update()` inside of the constructor instead of having `self.functions = SamFunctionProvider._extract_functions(...)` in two places.

mildaniel: The only thing with this is that instance variables should be instantiated in the constructor otherwise our linter will complain.

lucashuy (approved):

samcli/commands/build/build_context.py
```diff
@@ -318,32 +304,47 @@ def _handle_build_pre_processing(self) -> List[Stack]:
...
         """
         stacks = []
-        for stack in self.stacks:
-            stacks.append(EsbuildBundlerManager(stack=stack).set_sourcemap_metadata_from_env())
-        return stacks
+        if any(EsbuildBundlerManager(stack).esbuild_configured() for stack in self.stacks):
+            # esbuild is configured in one of the stacks, will check and update stack metadata accordingly
+            for stack in self.stacks:
+                stacks.append(EsbuildBundlerManager(stack).set_sourcemap_metadata_from_env())
+            self.function_provider.update(stacks, self._use_raw_codeuri, locate_layer_nested=self._locate_layer_nested)
```
mndeveci, line 311: Will `any` return `True` with first instance, or it will loop through the entire array? I was asking because we are looping over `stacks` couple of times here and was wondering if we can improve it by looping only once.

samcli/commands/build/build_context.py
```diff
@@ -318,32 +304,47 @@ def _handle_build_pre_processing(self) -> List[Stack]:
...
+        return stacks if stacks else self._stacks
```
mndeveci, line 312: Should we use `self.stacks` here as well (instead of `self._stacks`)?

mndeveci (commented):
LGTM in general, just asking some questions to clarify some of the behavior there.

samcli/commands/build/build_context.py
mildaniel: `any` will exit early on the first `True` that it sees so it won't re-run the logic more than once.

mndeveci (approved):

PR #4165, opened 2022-08-29
fix: allow relative paths for `--config-file`

sriram-mv:
* PR https://github.com/aws/aws-sam-cli/pull/4044 has caused a
  regression.

* https://github.com/aws/aws-sam-cli/pull/4044 

* Re allow usage of relative paths for `--config-file`

* Only check for the path being a file after resolving it to be absolute.

* N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -292,6 +292,9 @@ docs/_build/
 # pyenv
 .python-version
 
+# pyright
+pyrightconfig.json
+
 # celery beat schedule file
 celerybeat-schedule.*
 
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -130,20 +130,20 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
     config_env_name = ctx.params.get("config_env") or DEFAULT_ENV
 
     config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
-    if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file).is_file():
-        error_msg = f"Config file {config_file} does not exist or could not be read"
+    config_dir = getattr(ctx, "samconfig_dir", None) or os.getcwd()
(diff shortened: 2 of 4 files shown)


samcli/cli/cli_config_file.py
```diff
@@ -130,20 +130,20 @@ def configuration_callback(cmd_name, option_name, saved_callback, provider, ctx,
     config_env_name = ctx.params.get("config_env") or DEFAULT_ENV
 
     config_file = ctx.params.get("config_file") or DEFAULT_CONFIG_FILE_NAME
-    if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file).is_file():
-        error_msg = f"Config file {config_file} does not exist or could not be read"
+    config_dir = getattr(ctx, "samconfig_dir", None) or os.getcwd()
+    # If --config-file is an absolute path, use it, if not, start from config_dir
+    config_file_path = config_file if os.path.isabs(config_file) else os.path.join(config_dir, config_file)
+    if config_file and config_file != DEFAULT_CONFIG_FILE_NAME and not Path(config_file_path).absolute().is_file():
```
mndeveci, line 136: We didn't have `absolute` call before, why we need it now?

mndeveci (commented):
LGTM in general just couple of questions to understand the behavior here.

.gitignore
```diff
@@ -292,6 +292,9 @@ docs/_build/
 # pyenv
 .python-version
 
+# pyright
+pyrightconfig.json
```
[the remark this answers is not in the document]
sriram-mv: This is my LSP when I'm editing via vim.

samcli/cli/cli_config_file.py
sriram-mv: if the config file supplied has a relative path, we need to resolve it, which looks like was a regression from the previous PR, which is why I added a test particularly for this.

mndeveci (approved):

jfuss (approved):
LGTM

PR #4182, opened 2022-09-01
fix: make terraform project root arg required, and move script into the terraform hook package dir

torresxb1:
The current logic would make it so that if the terraform project root arg was not provided, extracted relative paths would be made relative to the current working directory (wherever the script was executed, so from the prepare hook's dir). 

Also, since hook packages should be independent/isolated, I moved the script into the terraform hook package. 
**NOTE:** not sure, however, where to move the integration test

The arg was made required.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
+++ b/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
@@ -217,27 +217,24 @@ if __name__ == "__main__":
     argparser.add_argument(
         "--terraform-project-root",
         type=str,
-        required=False,
+        required=True,
         default=None,
-        help="Extracted expression (path) will be relative to the terraform project root if provided.",
+        help="Extracted expression (path), if relative, will be relative to the provided terraform project root.",
     )
 
     arguments = argparser.parse_args()
     directory_path = os.path.abspath(arguments.directory)
 
-    terraform_project_root = (
-        os.path.abspath(arguments.terraform_project_root) if arguments.terraform_project_root else None
-    )
+    terraform_project_root = os.path.abspath(arguments.terraform_project_root)
     extracted_attribute_path = None
     data_object = None
 
     if not os.path.exists(directory_path) or not os.path.isdir(directory_path):
         LOG.error("Expected --directory to be a valid directory!")
         cli_exit()
-    if terraform_project_root:
-        if not os.path.exists(terraform_project_root) or not os.path.isdir(terraform_project_root):
-            LOG.error("Expected --terraform-project-path to be a valid directory!")
-            cli_exit()
+    if not os.path.exists(terraform_project_root) or not os.path.isdir(terraform_project_root):
+        LOG.error("Expected --terraform-project-path to be a valid directory!")
+        cli_exit()
 
     # Load and Parse
     try:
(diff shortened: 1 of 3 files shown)


moelasmar (approved):

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
```diff
@@ -217,27 +217,24 @@ def cli_exit():
     argparser.add_argument(
         "--terraform-project-root",
         type=str,
-        required=False,
+        required=True,
```
sriram-mv, line 220: If this script is moved, the corresponding pyinstaller hook also needs to be changed to ensure that this file is included.

sriram-mv (changes requested):

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
torresxb1: It seems like it already includes the samcli, so I believe I only need to remove these files now? https://github.com/sriram-mv/aws-sam-cli/blob/85cfa00ecb49e12ae04685b74f20617656e0a917/installer/pyinstaller/hook-samcli.py#L8-L10

sriram-mv: https://github.com/sriram-mv/aws-sam-cli/blob/85cfa00ecb49e12ae04685b74f20617656e0a917/installer/pyinstaller/hook-samcli.py#L10 

`include_datas` is the piece that matters there, it collects all the py files first and then only includes the files that are passed to `include_datas`, and that takes in a path.

torresxb1: got it, thank you! 
done

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
```diff
@@ -249,6 +251,11 @@ def cli_exit():
         LOG.error(ex.message, exc_info=True)
         cli_exit()
 
+    if not isinstance(extracted_attribute_path, string_types):
+        LOG.error("Expected the extracted attribute to be a string")
+        cli_exit()
+    extracted_attribute_path = str(extracted_attribute_path)
```
torresxb1, line 257: mypy was complaining about this
```
samcli/hook_packages/terraform/copy_terraform_built_artifacts.py:261: error: Argument 2 to "join" has incompatible type "Union[Any, None, str]"; expected "Union[str, _PathLike[str]]"
samcli/hook_packages/terraform/copy_terraform_built_artifacts.py:262: error: Argument 1 to "isabs" has incompatible type "Union[Any, None, str]"; expected "Union[str, bytes, _PathLike[str], _PathLike[bytes]]"
samcli/hook_packages/terraform/copy_terraform_built_artifacts.py:263: error: Argument 1 to "abspath" has incompatible type "Union[Any, None, str]"; expected "_PathLike[Any]"
```

torresxb1: @sriram-mv this check is what I meant with my comment here https://github.com/aws/aws-sam-cli/pull/4117#discussion_r950594500
😅

tests/integration/scripts/test_copy_terraform_built_artifacts.py
```diff
@@ -11,7 +11,7 @@
 
 class TestCopyTerraformBuiltArtifacts(TestCase):
     def setUp(self) -> None:
-        self.script_dir = pathlib.Path(__file__).parents[3].joinpath("samcli").joinpath("scripts")
+        self.script_dir = pathlib.Path(__file__).parents[3].joinpath("samcli/hook_packages/terraform")
```
sriram-mv, line 14: I don't know if this will work on windows too because of the different path separators on windows. I'm going to approve this, but check on the gates to ensure.

sriram-mv (approved):

tests/integration/scripts/test_copy_terraform_built_artifacts.py
torresxb1: I realized I used joinpath wrong, and just had to provide the different parts of the path separated by commas. Fixed now

PR #4185, opened 2022-09-01
fix: check for `hook-package-id` only being allowed if a beta feature on `sam build`

sriram-mv:
There is no sam build logic for terraform applications yet. This PR only ensures that we check to see if beta features are enabled when executing a command with build command with `--hook-package-id`

N/A

 Only allow for `sam build` to continue if a hook package id is specified if executed within a beta context.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/buildcmd/build_integ_base.py
+++ b/tests/integration/buildcmd/build_integ_base.py
@@ -78,6 +78,7 @@ class BuildIntegBase(TestCase):
         build_image=None,
         exclude=None,
         region=None,
+        hook_package_id=None,
         beta_features=False,
     ):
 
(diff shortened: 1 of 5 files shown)


hawflau (approved):

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -86,7 +87,7 @@ def get_command_list(
         if function_identifier:
             command_list += [function_identifier]
 
-        command_list += ["-t", self.template_path]
```
moelasmar, line 89: can we override the template_path to be None in case of Terraform testing, and to depend on it in building the command list.

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -78,6 +78,7 @@ def get_command_list(
         build_image=None,
         exclude=None,
         region=None,
+        hook_package_id=None,
         beta_features=False,
```
moelasmar, line 82: can we allow accepting None, so we can test both `--beta-features` and `--no-beta-features` use cases.

moelasmar (commented):
it looks good, just left small comments

moelasmar (approved):

PR #4187, opened 2022-09-02
feat: implement _enrich_resources_and_generate_makefile

torresxb1:
Adds to `_enrich_resources_and_generate_makefile` the logic for generating a makefile with a rule for each lambda resource to be built. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
torresxb1, line 1: Should we maybe think about pulling the translation stuff into its own module?

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -110,7 +112,7 @@ def prepare(params: dict) -> dict:
 
         # store in supplied output dir
         if not os.path.exists(output_dir_path):
-            os.mkdir(output_dir_path)
+            os.makedirs(output_dir_path, exist_ok=True)
```
torresxb1, line 115: makedirs would create any intermediate directories too

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -276,9 +278,66 @@ def _generate_custom_makefile(
...
+    Returns
+    -------
+    str
+        The name of the python command installed
+    """
+    # TODO
+    return ""
+
+
+def _write_makefile_rule_for_lambda_resource(
```
torresxb1, line 318: will implement in followup PR(s)

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -276,9 +278,66 @@ def _generate_custom_makefile(
...
+    shutil.copy("../copy_terraform_built_artifacts", output_directory_path)
+
+    # check python is installed and get python command
+    python_command_name = _get_python_command_name()
+
+    # create makefile
+    makefilePath = os.path.join(output_directory_path, "Makefile")
+    with open(makefilePath, "a+") as makefile:
+        for sam_metadata_resource in sam_metadata_resources:
+            _write_makefile_rule_for_lambda_resource(
```
moelasmar, line 300: how will you determine the Lambda resource that you are going to generate a building rule for it?

torresxb1: planning to use similar logic of the enrich function, using `_get_relevant_cfn_resource`

moelasmar: My issue is repeating this logic, so in case if we have to add more complex logic to determine the resource, we will repeat this logic, and may cause performance issues. 

Can we move the makefile rules generation to the enrichment method, or change the enrichment method to add the found resource to each sam_metadata object

torresxb1: it doesn't seem very complex or performance heavy right now, though 🤔 . Can we just leave as is, and if at some point in the future we introduce changes that will impact performance then we just refactor at that point?

torresxb1: or we can also do it as a followup PR

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -276,9 +278,66 @@ def _generate_custom_makefile(
...
+    terraform_application_dir: str
+        the terraform project root directory
+    """
+
+    # create output directory if it doesn't exist
+    if not os.path.exists(output_directory_path):
+        os.makedirs(output_directory_path, exist_ok=True)
+
+    # copy copy_terraform_built_artifacts.py script into output directory
+    shutil.copy("../copy_terraform_built_artifacts", output_directory_path)
```
sriram-mv, line 291: not sure on how this would be handled in windows, always have to be careful with paths to ensure they are cross platfrom.

sriram-mv (changes requested):

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: good call out to be careful just in case, I think my latest commit should address it

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: I prefer to do that now better than later. The effort to do the change later will be more expensive than now. I agree that the performance now is not heavy, but if we changed how we determine the cfn resource later it may be heavy specially that we already defined this information before, and this fact will never change.

tests/unit/hook_packages/terraform/test_prepare_hook.py
```diff
@@ -1101,7 +1097,7 @@ def test_get_lambda_function_source_code_path_valid_metadata_resource(
             "source_code_property",
             "source code",
         )
-        self.assertEquals(path, expected_path)
```
torresxb1, line 1104: deprecated

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: done

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -245,44 +247,16 @@ def _translate_to_cfn(tf_json: dict, output_directory_path: str, terraform_appli
...
-    cfn_resources: dict
-        CloudFormation resources
-    output_directory_path: str
-        the output directory path to write the generated metadata and makefile
-    terraform_application_dir: str
-        the terraform project root directory
-    """
-    # TODO
-
-
```
torresxb1, line 285: combined into/with the enrichment stuff, as suggested by Mohamed

samcli/hook_packages/terraform/hooks/prepare.py
sriram-mv: Looks like the gates passed, so my question is moot. 

But I will throw this out there anyway, sometimes params like exist_ok or any new params that get added are from a certain python version. So It might work locally, but will fail at the gates, just something to have in mind for the future.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -621,20 +596,47 @@ def _enrich_mapped_resources(
...
+            sam_metadata_resource_address = sam_metadata_resource.resource.get("address")
+            enrichment_function = resources_types_enrichment_functions.get(resource_type)
+            if enrichment_function is None:
+                raise InvalidSamMetadataPropertiesException(
+                    f"The resource type {resource_type} found in the sam metadata resource "
+                    f"{sam_metadata_resource_address} is not a correct resource type. The resource type should be one "
+                    f"of these values {resources_types_enrichment_functions.keys()}"
+                )
+            cfn_resource, logical_id = _get_relevant_cfn_resource(sam_metadata_resource, cfn_resources)
+            enrichment_function(
```
sriram-mv, line 628: I'm not following this bit, is `resources_types_enrichment_functions` a dictionary of functions, where is it defined?

torresxb1: the enrichment logic I didn't add, that was already there. I just put my changes and the enrichment stuff in the same function.
But yes, resources_types_enrichment_functions is a dictionary of functions, defined in this same function (a few lines above)

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -621,20 +596,47 @@ def _enrich_mapped_resources(
...
-        resource_type = sam_metadata_resource.resource.get("values", {}).get("triggers", {}).get("resource_type")
-        sam_metadata_resource_address = sam_metadata_resource.resource.get("address")
-        enrichment_function = resources_types_enrichment_functions.get(resource_type)
-        if enrichment_function is None:
-            raise InvalidSamMetadataPropertiesException(
-                f"The resource type {resource_type} found in the sam metadata resource {sam_metadata_resource_address} "
-                f"is not a correct resource type. The resource type should be one of these values "
-                f"{resources_types_enrichment_functions.keys()}"
+    # create output directory if it doesn't exist
+    if not os.path.exists(output_directory_path):
```
moelasmar, line 600: can we push the make file creation to be after the enrichment logic. As in case if we find any issue during the enrichment, and the prepare hook failed, we should not create the Makefile, or the metadata file.

moelasmar (commented):
I left one comment

samcli/hook_packages/terraform/hooks/prepare.py
moelasmar: thanks

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: done

moelasmar (approved):
thanks Rup :)

sriram-mv (approved):

Issue #4172, opened 2022-08-31
Feature request: AWS::LanguageExtensions support


jdel:
### Describe your idea/feature/enhancement

I wish SAM CLI would handle the new AWS::LanguageExtensions transform as specified in the documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-languageextension-transform.html

Example template:
```yml
AWSTemplateFormatVersion: 2010-09-09
Transform:
  - AWS::LanguageExtensions
  - AWS::Serverless-2016-10-31

Parameters:
  Environment:
    Type: String
    Default: dev
    AllowedValues:
      - dev
      - prod

Conditions:
  IsProd: !Equals [!Ref Environment, prod]

Resources:
  Bucket:
    Type: AWS::S3::Bucket
    DeletionPolicy: !If [IsProd, Retain, Delete]
    UpdateReplacePolicy: !If [IsProd, Retain, Delete]
```

`AWS CLI`, version 2.7.7 succeeds with:
```
❯ aws cloudformation deploy --stack-name test-language-extension --template-file sam/test-template.yml

Waiting for changeset to be created..
Waiting for stack create/update to complete
Successfully created/updated stack - test-language-extension
```

`SAM CLI`, version 1.55.0 fail with:
```
sam deploy --stack-name test-language-extension --template-file sam/test-template.yml
Traceback (most recent call last):
[... redacted stack trace ...]
  File "/usr/local/Cellar/aws-sam-cli/1.55.0/libexec/lib/python3.8/site-packages/samcli/lib/samlib/wrapper.py", line 70, in run_plugins
    raise InvalidSamDocumentException(
samcli.commands.validate.lib.exceptions.InvalidSamDocumentException: [InvalidTemplateException('Every DeletionPolicy member must be a string.')] Every DeletionPolicy member must be a string.
```

### Proposal

Implement the same changes as `aws-cli` and `cfn-lint`

For reference, the releases/announcements from both tools 
  - https://github.com/aws-cloudformation/cfn-lint/compare/v0.62.0..v0.63.0
  - https://github.com/aws/aws-cli/issues/3825#issuecomment-1231608258

Things to consider:
The order of the transforms is important and using multiple transforms is currently broken in `cfn-lint` https://github.com/aws-cloudformation/cfn-lint/issues/2346

### Additional Details

No additional details

jfuss:
[...]

PR #4188, opened 2022-09-02
fix: Patch sam-cli to support language extensions

mildaniel:

#4172

SAM-CLI shouldn't error out when validating certain intrinsic properties if `Transform: AWS::LanguageExtensions` is in the template.

Temporarily patch the validation function to bypass certain validations.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/samlib/wrapper.py
+++ b/samcli/lib/samlib/wrapper.py
@@ -66,6 +66,7 @@ class SamTranslatorWrapper:
             additional_plugins, parameters=self.parameter_values if self.parameter_values else {}
         )
 
+        # Temporarily disabling validation for DeletePolicy and UpdateReplacePolicy when language extensions are set
         self._patch_language_extensions()
 
         try:
@@ -81,7 +82,11 @@ class SamTranslatorWrapper:
     def template(self):
         return copy.deepcopy(self._sam_template)
 
-    def _patch_language_extensions(self):
+    def _patch_language_extensions(self) -> None:
+        """
+        Monkey patch SamResource.valid function to exclude checking DeletePolicy
+        and UpdateReplacePolicy when language extensions are set
+        """
         template_copy = self.template
         if self._check_using_langauge_extension(template_copy):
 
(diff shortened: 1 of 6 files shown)


mndeveci (approved):
LGTM!

samcli/lib/samlib/wrapper.py
```diff
@@ -77,6 +81,30 @@ def run_plugins(self, convert_local_uris=True):
     def template(self):
         return copy.deepcopy(self._sam_template)
 
+    def _patch_language_extensions(self):
```
mndeveci, line 84: Can we mention here that we are temporarily disabling validation for `DeletePolicy` and `UpdateReplacePolicy` for language extensions?

mndeveci (approved):

lucashuy (approved):

samcli/lib/samlib/wrapper.py
```diff
@@ -77,6 +82,39 @@ def run_plugins(self, convert_local_uris=True):
...
+    @staticmethod
+    def _check_using_langauge_extension(template: Dict) -> bool:
+        """
+        Check if language extensions are set in the template's Transform
+        :param template: template to check
+        :return: True if language extensions are set in the template, False otherwise
+        """
+        transform = template.get("Transform")
+        if transform:
+            if isinstance(transform, str) and transform == "AWS::LanguageExtensions":
```
jlhood, line 112: It's ok to leave this in, but is this a "can't happen" case? Doesn't SAM CLI verify that the template contains the Serverless transform?

samcli/lib/samlib/wrapper.py
```diff
@@ -77,6 +82,39 @@ def run_plugins(self, convert_local_uris=True):
...
+                return True
+            if isinstance(transform, list) and "AWS::LanguageExtensions" in transform:
```
jlhood, line 114: Can we just check for transforms with the prefix "AWS::LanguageExtensions" rather than checking for this exact string? Will sync with you offline...

samcli/lib/samlib/wrapper.py
mildaniel: This was a case I was a little unsure about. Since SAM-CLI supports raw CFN resources, there's no requirement for a template to have the serverless transform for it to work with SAM. Going off this logic, this seemed like a valid case.

samcli/lib/samlib/wrapper.py
mildaniel: I'll raise an updated PR to check for the prefix instead of the absolute string

samcli/lib/samlib/wrapper.py
jlhood: Gotcha, makes sense to keep it then. Thanks!

samcli/lib/samlib/wrapper.py
jlhood: Thank you!

PR #4189, opened 2022-09-02
Fix parse remote origin url

hawflau:
N/A

`_parse_remote_origin_url` didn't handle certain url types that git supports, resulting in error and breaking a command run.

- refactor `_parse_remote_origin_url` to return either str if the given url is one of the supported style (http or git ssh). Otherwise, return None
- update methods that use `_parse_remote_origin_url`
- revert "get_project_name" to use only the basename of cwd as the fallback return, as it is intended to be used as a combination key instead of a unique key. We don't have strong uniqueness requirement for it.
- Skip capturing project metadata for `sam init` command runs, because it is not applicable (as project is not initiated yet).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/telemetry/project_metadata.py
+++ b/samcli/lib/telemetry/project_metadata.py
@@ -36,9 +36,7 @@ def get_git_remote_origin_url() -> Optional[str]:
         # Ignoring, None git_url will be handled later
         pass
 
-    if not git_url:
-        return None
-    return _hash_value(git_url)
+    return _hash_value(git_url) if git_url else None
 
 
 def get_project_name() -> Optional[str]:
--- a/samcli/lib/telemetry/project_metadata.py
+++ b/samcli/lib/telemetry/project_metadata.py
@@ -3,15 +3,12 @@ Creates and hashes metadata regarding SAM CLI projects.
 """
 
 import hashlib
-from os import getcwd
 import re
 import subprocess
-import logging
+from os import getcwd
 from typing import Optional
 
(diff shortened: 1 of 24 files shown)


samcli/lib/telemetry/project_metadata.py
```diff
@@ -88,24 +94,46 @@ def get_initial_commit_hash() -> Optional[str]:
...
+    if not found:
+        return None
+    if found.endswith("/"):
+        found = found[:-1]
+    if found.endswith(".git"):
+        found = found[:-4]
+    found = re.sub(r":\d{1,5}", "", found)  # remove port number if any
+    found = found.replace(":", "/")  # e.g. github.com:aws/aws-sam-cli -> github.com/aws/aws-sam-cli
+
+    return found
```
hoffa, line 132: This can be replaced with:

```python
import re
from urllib.parse import urlparse

def parse(url):
    u = urlparse(url)
    s = (u.hostname or "") + u.path
    s = re.sub("/$", "", s)
    s = re.sub(".git$", "", s)
    s = re.sub("^git@", "", s)
    s = s.replace(":", "/")
    return s
```

And it'll pass all the test cases in `test_retrieve_git_origin`.

hawflau: Thanks for the suggestion! Will update. 👍

hoffa (approved):

PR #4192, opened 2022-09-02
feat: raise error in case of build in container with hooks package and no build image.

moelasmar:
- Raise error in case of building a terraform application using container and there is no build image provided.
- Added some invalid build options integration test cases using simple terraform project 
- moved the beta features integration test cases to the Terraform integration test cases.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -2565,29 +2565,3 @@ def test_sar_application_with_location_resolved_from_map(self, use_container, re
             # will fail the build as there is no mapping
             self.assertEqual(process_execute.process.returncode, 1)
             self.assertIn("Property \\'ApplicationId\\' cannot be resolved.", str(process_execute.stderr))
-
```
moelasmar, line 2568: moved it to the terraform build testing modules

samcli/commands/_utils/custom_options/hook_package_id_option.py
```diff
@@ -50,6 +50,16 @@ def handle_parse_result(self, ctx, opts, args):
                     raise click.BadParameter(
                         f"Parameters {opt_name}, and {','.join(self._invalid_coexist_options)} can not be used together"
                     )
+
+            # validate build-image is provided in case of build using container
+            # add this validation here to avoid running hook prepare and there is issue
+            if command_name == "build" and opts.get("use_container") and not opts.get("build_image"):
+                msg = (
+                    "Missing required parameter, need the --build-image parameter in order to use --use-container flag "
```
sriram-mv, line 58: Missing required parameter "--build-image", should that suffice?

sriram-mv (changes requested):

samcli/commands/_utils/custom_options/hook_package_id_option.py
moelasmar: hmmm ... will it be clear for the customer .. I like explaining things :D

sriram-mv: It should be, since there is context on the user who just ran the command with a certain number of options. I like the explanation aspect, but if I wonder if we should start codifying it so that its consistent across commands.

sriram-mv (approved):

appveyor-ubuntu.yml
```diff
@@ -54,7 +54,7 @@ install:
   # AppVeyor's apt-get cache might be outdated, and the package could potentially be 404.
   - sh: "sudo apt-get update"
 
-  - sh: "gvm use go1.13"
+  - sh: "gvm use go1.15"
```
moelasmar, line 57: appveyor jobs were failing as go1.13 is not available

hawflau (approved):

PR #4210, opened 2022-09-08
feat: Implement _get_python_command_name

torresxb1:
Verifies that python is installed and returns the name of the python command, to use for running python scripts.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -8,6 +8,7 @@ import json
 import os
 from json.decoder import JSONDecodeError
 from pathlib import Path
+import re
 from subprocess import run, CalledProcessError
 from typing import Any, Callable, Dict, List, Optional, Tuple
 import hashlib
@@ -718,15 +719,20 @@ def _get_python_command_name() -> str:
     str
         The name of the python command installed
     """
-    command_names_to_try = ["python", "python3"]
+    command_names_to_try = ["python3", "py3", "python", "py"]
     for command_name in command_names_to_try:
         try:
-            run([command_name, "--version"], check=True, capture_output=True)
+            run_result = run([command_name, "--version"], check=True, capture_output=True, text=True)
         except CalledProcessError:
             pass
         else:
+            # check for python 3, 3.7 or above
+            # regex: search for 'Python', whitespace, '3.', digits 7-9 or 2+ digits, any digit or '.' 0+ times
+            match = re.search(r"Python\s*3.([7-9]|\d{2,})[\d.]*", run_result.stdout)
+            if not match:
+                continue
             return command_name
-    raise PrepareHookException("Python not found. Please ensure that python is installed.")
+    raise PrepareHookException("Python not found. Please ensure that python 3.7 or above is installed.")
 
 
 def _generate_makefile_rule_for_lambda_resource(
--- a/samcli/hook_packages/terraform/hooks/prepare.py
+++ b/samcli/hook_packages/terraform/hooks/prepare.py
@@ -718,8 +718,15 @@ def _get_python_command_name() -> str:
     str
         The name of the python command installed
     """
-    # TODO
-    return ""
+    command_names_to_try = ["python", "python3"]
+    for command_name in command_names_to_try:
+        try:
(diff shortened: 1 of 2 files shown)


moelasmar (approved):

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -688,6 +709,53 @@ def _get_relevant_cfn_resource(
...
+def _get_python_command_name() -> str:
+    """
+    Verify that python is installed and return the name of the python command
+
+    Returns
+    -------
+    str
+        The name of the python command installed
+    """
+    command_names_to_try = ["python", "python3"]
```
hawflau, line 721: Will these work in Windows?
Also, is it possible for `python` be pointed to Python2? Is the script being run Python version agnostic?

torresxb1: Good points.

At least one of these options should work on Windows. I believe the default one on windows might be `python`, but I can add `py` and `py3` to `command_names_to_try` just in case

The script which is the intended use case for this function should work with python 2 or 3, but @sriram-mv  can confirm. Here's the PR: https://github.com/aws/aws-sam-cli/pull/4117
Regardless, this is an independent function that could be used for other scripts, so I can at least add some comments here to point that out.

mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare.py
sriram-mv: The script should work across all python versions.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -718,15 +719,20 @@ def _get_python_command_name() -> str:
...
             pass
         else:
+            # check for python 3, 3.7 or above
+            # regex: search for 'Python', whitespace, '3.', digits 7-9 or 2+ digits, any digit or '.' 0+ times
+            match = re.search(r"Python\s*3.([7-9]|\d{2,})[\d.]*", run_result.stdout)
+            if not match:
+                continue
             return command_name
-    raise PrepareHookException("Python not found. Please ensure that python is installed.")
+    raise PrepareHookException("Python not found. Please ensure that python 3.7 or above is installed.")
```
sriram-mv, line 735: We need to have doc updates to ensure this is a pre-req to use.

samcli/hook_packages/terraform/hooks/prepare.py
```diff
@@ -718,15 +719,20 @@ def _get_python_command_name() -> str:
...
     for command_name in command_names_to_try:
         try:
-            run([command_name, "--version"], check=True, capture_output=True)
+            run_result = run([command_name, "--version"], check=True, capture_output=True, text=True)
         except CalledProcessError:
             pass
         else:
+            # check for python 3, 3.7 or above
+            # regex: search for 'Python', whitespace, '3.', digits 7-9 or 2+ digits, any digit or '.' 0+ times
+            match = re.search(r"Python\s*3.([7-9]|\d{2,})[\d.]*", run_result.stdout)
```
sriram-mv, line 731: nit: this maybe ok in perf front since we have a limit on the command names to try, but in general its better to separate the match by creating a re.compile(expr) ahead of time and then re-using it.

you can benchmark things like this by using `timeit`

python -m timeit -s "<stmt>"

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: It's the same version supported by the sam cli (setup.py)

samcli/hook_packages/terraform/hooks/prepare.py
torresxb1: It seems like due to caching (and like you mentioned, our small number of commands to try) this should be a negligible difference: https://stackoverflow.com/a/47269110

torresxb1: made the change to pre-compile anyway ✅

sriram-mv (approved):

Issue #4082, opened 2022-07-26
Bug: Wrong handler mapping after esbuild


santiperone:
### Description:
  When running esbuild on a Lambda function, the output doesn't mantain the same folder structure as the enry points, and this remains unchange in the built template, resulting in SAM not finding the lambda Handler:

### Steps to reproduce:
Project structure:
```
- package.json
- template.yaml
-- src
--- functions
---- FunctionName
----- app.js
--- utils
---- some-util.js
```

Template.yaml:
```
FunctionName:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: esbuild
    Properties:
      Handler: src/functions/FunctionName/app.Handler
      Events:
        HttpEvent:
          ....
```

### Observed result:
code is budled to FunctionName folder in .aws-build resulting in:
```
- .aws-build
-- build
--- template.yaml
--- FunctionName
---- app.js
```
However, the built template results in:

```
FunctionName:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: esbuild
    Properties:
      CodeUri: FunctionName
      Handler: src/functions/FunctionName/app.Handler
      Events:
        HttpEvent:
          ....
```

Resulting in an error since FunctionName/src/functions/FunctionName/app.Handler

### Expected result:
Handler in built template should match esbuild output. I.E: 

```
FunctionName:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: esbuild
    Properties:
      CodeUri: FunctionName
      Handler: app.Handler
      Events:
        HttpEvent:
          ....
```

jfuss:
@santiperone You need to provide a `CodeUri` in your template. If you do not, SAM CLI assumes `CodeUri` is the `cwd`. This behavior cannot be changed since it would cause a breaking change. 

Looks like you want the `CodeUri` to be `src/functions/FunctionName` and Handler to be `app.Handler`, from your example above.

santiperone:
@jfuss I am aware of this. Actually no, I need to bundle packages and files from the entire cwd, (i.e. node_modules and the 'some_util' file from the example), and if I set my codeUri as you say this files would not be found by esbuild.

santiperone:
This approach would allow to have shared, and a single node_modules folder (which makes testing a set of lambdas within a project easier) whilst keeping only the needed code for each function upon build, which I guess is the point of using esbuild.

moelasmar:
[...]

PR #4216, opened 2022-09-09
fix: Resolve handler paths to be relative to output dir

mildaniel:

#4082

When bundling other runtimes, we copy the resulting file tree to the artifact directory. In the case of esbuild, we bundle straight into the artifact directory, meaning that the initial file structure is lost and the artifact directory just contains the bundle artifacts. 

The change post-processes the template to check for this discrepancy, and adjusts the handler property to be relative to the artifact directory.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/build/bundler.py
```diff
@@ -11,12 +12,15 @@
 LOG = logging.getLogger(__name__)
 
 ESBUILD_PROPERTY = "esbuild"
+HANDLER_WITHOUT_FUNCTION_INDEX = 0
```
sriram-mv, line 15: I'm squeamish about magic numbers, do we know if there any enough safeguards for this?

samcli/lib/build/bundler.py
```diff
@@ -32,6 +36,12 @@ def esbuild_configured(self) -> bool:
                 return True
         return False
 
+    def handle_template_post_processing(self) -> Dict:
+        template = deepcopy(self._previous_template)
```
sriram-mv, line 40: Idea: If there are multiple deepcopies , there might be something to just copy once and set it in thread local storage.

samcli/lib/build/bundler.py
```diff
@@ -134,6 +143,53 @@ def set_sourcemap_env_from_metadata(self) -> Dict:
...
+    @staticmethod
+    def _get_path_and_filename_from_handler(handler: str) -> str:
+        """
+        Takes a string representation of the handler defined in the
+        template, returns the file name and location of the handler.
+
+        :param handler: string representation of handler property
+        :return: string path to built handler file
+        """
+        return handler.split(".")[HANDLER_WITHOUT_FUNCTION_INDEX] + ".js"
```
sriram-mv, line 170: do we need any error handling here?

samcli/lib/build/bundler.py
```diff
@@ -134,6 +143,53 @@ def set_sourcemap_env_from_metadata(self) -> Dict:
...
+
+        :param template: deepcopy of template dict
+        :return: Updated template with resolved handler property
+        """
+        for name, resource in self._stack.resources.items():
+            if self._esbuild_in_metadata(resource.get("Metadata", {})):
+                long_path_handler = resource.get("Properties", {}).get("Handler", "")
+                if not long_path_handler or not self._check_invalid_lambda_handler(long_path_handler, name):
+                    continue
+                resolved_handler = long_path_handler.split("/")[HANDLER_FILE_INDEX]
```
sriram-mv, line 188: does it matter on windows?

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -370,11 +370,16 @@ def _test_with_various_properties(self, overrides):
...
-        filename = handler.split(".")[0]
+        filename = self._extract_filename_from_handler(handler)
         resource_artifact_dir = build_dir.joinpath(function_logical_id)
         self._verify_sourcemap_created(filename, resource_artifact_dir)
         self._verify_function_minified(filename, resource_artifact_dir)
 
+    @staticmethod
+    def _extract_filename_from_handler(handler):
+        # Takes a handler in the form /a/b/c/file.function and returns file
+        return (handler.split(".")[0]).split("/")[-1]
```
sriram-mv, line 381: can we use file objects attribute to load instead of splits and indexing?

samcli/lib/build/bundler.py
mildaniel: Good question.. this depends on the CFN spec and what's expected in the handler property. I'm not sure if CFN expects Windows file separators but can look into it more tomorrow.

samcli/lib/build/bundler.py
mildaniel: There should only be two deepcopies, one for the pre-processed template and one for the post-processed template which should be different templates.

samcli/lib/build/bundler.py
mildaniel: File separators for in the template should be the same regardless of OS. For example, our templates don't make a distinction https://github.com/aws/aws-sam-cli-app-templates/blob/master/nodejs18.x/web/%7B%7Bcookiecutter.project_name%7D%7D/template.yaml.

samcli/lib/build/bundler.py
mildaniel: I can't find any exact documentation, but I do believe the patter is always `<path_to_file>.<handler_function_name>` for JavaScript Lambda functions. There's a brief mention of it in the documentation here: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html

mildaniel: I will add a comment to the constant.

mildaniel: Removed all magic numbers if favour of file objects.

lucashuy (approved):

samcli/lib/build/bundler.py
```diff
@@ -134,6 +142,61 @@ def set_sourcemap_env_from_metadata(self) -> Dict:
...
+        Takes a string representation of the handler defined in the
+        template, returns the file name and location of the handler.
+
+        :param handler: string representation of handler property
+        :return: string path to built handler file
+        """
+        try:
+            path = str(Path(handler).parent / Path(handler).stem) + ".js"
+        except (AttributeError, TypeError):
+            raise InvalidHandlerPathError("Unable to parse the handler")
```
mndeveci, line 176: If this exception won't be returned back to user (I saw `_should_update_handler` catches this exception and returns `False`), I would instead suggest to return `None` here. 

It will be easier for method user to see method returns `Optional[str]` and take an action (with exceptions in python, it is unclear if a method will raise an exception or not). It will be also easier (and more performant) since we don't need to have 2 times of try/except block.

mndeveci (changes requested):

samcli/lib/build/bundler.py
mildaniel: Good call-out, updated.

mndeveci (approved):

PR #4249, opened 2022-09-20
feat: define resource linking data representation structures

torresxb1:
Define a few structures to be used for parsing the Terraform plan file and linking resources

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -2,3 +2,41 @@
...
+        # TODO
+        pass
+
+
+@dataclass
+class Resource:
+    address: str
+    type: str
+    # the module this resource is defined in
+    module: "Module"
```
torresxb1, line 37: unsure whether `parent_module` or `module` is clearer

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -2,3 +2,41 @@
...
+class Value:
+    constant_value: Any
+    references: List[str]
+
+
+@dataclass
+class Module:
+    # full path to the module, including parent modules
+    full_address: str
+    parent_module: Optional["Module"]
```
torresxb1, line 20: have to either put it in quotes like this or do `from __future__ import annotations`, in order to self reference like this.
Preference on any way?

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -2,3 +2,41 @@
...
+
+
+@dataclass
+class Value:
+    constant_value: Any
+    references: List[str]
+
+
+@dataclass
+class Module:
```
moelasmar, line 17: There was a comment on the classes names to make it clear that it is related to terraform. I am not really good in naming, so I left this comment to let you think about the names :)

moelasmar (changes requested):

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: I would say let's aim for consistency through the CLI. We've typically been doing it this way so far so let's just keep going with it unless anyone else feels strongly about it.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: I think you're right that just `module` might be clearer here since it's the resource's own module within which it's defined and not the parent.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -2,3 +2,41 @@
...
+    outputs: Dict[str, Value]
+
+    # current module's + all child modules' resources
+    def get_all_resources(self) -> List["Resource"]:
+        # TODO
+        pass
+
+
+@dataclass
+class Resource:
```
mildaniel, line 33: nit: We can move this class declaration above `Module` since it is used inside of `Module`. Then we can reference `Resources` in the type hints without the `""`.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
torresxb1: I thought about naming, but I feel like this is clear enough within this module (if anybody imports it, they could just rename the import). Happy to hear others' thoughts though

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
torresxb1: `Resource` also references a `Module`, so it's kinda the same either way 😅

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -2,3 +2,41 @@
...
+    address: str
+    type: str
+    # the module this resource is defined in
+    module: "Module"
+    attributes: Dict[str, Value]
+
+    @property
+    def full_address(self) -> str:
+        return f"{self.module.full_address}.{self.address}"
```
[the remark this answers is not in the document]
torresxb1: I don't think a module should have an empty string address though. Even the root module will have an address. Unless you mean as an edge/error case. I could add some error handling here but I wonder if that's better suited for the function that builds modules (it shouldn't even set the address of a module if it's an incorrect address)

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
lucashuy: I'm not completely opposed to prepending the classes with `TF` (eg `TFResource` removes potential ambiguity about whether its a Terraform or CFN resource), but I'm also okay with leaving it as is for the reason you mentioned

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: I must've missed that, my bad. We can probably remove the quotes around Module defined here then.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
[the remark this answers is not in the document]
moelasmar: but root module address will be an empty string. Why do you think it should have address?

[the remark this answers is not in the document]
torresxb1: you're right, I got mixed up. Will make the address optional, so that it's not empty string.

moelasmar (approved):

mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -3,8 +3,62 @@
...
+        return all_resources
+
+
+@dataclass
+class TFResource:
+    address: str
+    type: str
+    # the module this resource is defined in
+    module: TFModule
+    attributes: Dict[str, Expression]
```
torresxb1, line 55: left as `attributes` for now, but now the type includes `Expression`.
If we wanted to stick to the terminology of the plan file / configuration, I think we'd have to switch this to `expressions` and also switch module `variables` to `expressions`. But unsure which one is clearer to be honest.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -3,8 +3,62 @@
...
+    value: List[str]
+
+
+Expression = Union[ConstantValue, References]
+
+
+@dataclass
+class ResolvedReference:
+    value: str
+    module_address: str
```
torresxb1, line 27: updated these to consolidate the representation of expressions and references

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -3,8 +3,62 @@
...
+    child_modules: Dict[str, "TFModule"]
+    outputs: Dict[str, Expression]
+
+    # current module's + all child modules' resources
+    def get_all_resources(self) -> List["TFResource"]:
+        all_resources = self.resources.copy()
+        for _, module in self.child_modules.items():
+            all_resources += module.get_all_resources()
+
+        return all_resources
```
torresxb1, line 46: implemented

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -95,3 +100,34 @@ def test_get_configation_address(self, input_addr, expected_addr):
...
+
+    def test_resource_full_address(self):
+        module = TFModule("module.full_address", None, {}, {}, {}, {})
+        resource = TFResource("resource_address", "type", module, {})
+        self.assertEqual(resource.full_address, "module.full_address.resource_address")
+
+    def test_resource_full_address_root_module(self):
+        module = TFModule(None, None, {}, {}, {}, {})
+        resource = TFResource("resource_address", "type", module, {})
+        self.assertEqual(resource.full_address, "resource_address")
```
torresxb1, line 133: added tests

mildaniel (approved):

moelasmar (approved):

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -3,8 +3,62 @@
...
+    value: str
+    module_address: str
+
+
+@dataclass
+class TFModule:
+    # full path to the module, including parent modules
+    full_address: Optional[str]
+    parent_module: Optional["TFModule"]
+    variables: Dict[str, Expression]
```
moelasmar, line 35: I like the change to use Expression instead of the old TFValue class. It looks cleaner. Just let Daniel and Lucas know about this change as it may affect their implementation

lucashuy (approved):

PR #4259, opened 2022-09-23
feat: Added module output resolving

lucashuy:
This change resolves any references in the module's `output`.

Adds a function that recursively searches and resolves the module tree.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +99,62 @@ def _get_configuration_address(address: str) -> str:
...
+                results += _resolve_module_variable(module, stripped_reference)
+            elif reference.startswith("module."):
+                # aaa.bbb.ccc => bbb
+                module_name = reference[7 : reference.rfind(".")]
+                # aaa.bbb.ccc => ccc
+                output_name = reference[reference.rfind(".") + 1 :]
+
+                stripped_reference = _get_configuration_address(module_name)
+
+                results += _resolve_module_output(module.child_modules[stripped_reference], output_name)
```
moelasmar, line 152: can we use get instead of `[]`, and also can we raise an exception if the module is not in the child modules or if child modules is none, it means either the plan configuration has some issue or we have some logic issue.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +99,62 @@ def _get_configuration_address(address: str) -> str:
...
     Returns
-    =======
+    -------
     str
         The address clean of indices
     """
     return re.sub(r"\[[^\[\]]*\]", "", address)
+
+
+def _resolve_module_output(module: TFModule, output_name: str) -> List[Union[ConstantValue, ResolvedReference]]:
```
moelasmar, line 114: can you add some debugging logs to help in issues investigation later

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,41 @@ def test_resource_full_address_root_module(self):
...
+    @patch("samcli.hook_packages.terraform.hooks.prepare.resource_linking._get_configuration_address")
+    def test_resolve_module_output_with_var(self, config_mock, resolve_var_mock):
+        module = TFModule("", None, {"mycoolref": "mycoolvar"}, [], {}, {"mycooloutput": References(["var.mycoolref"])})
+
+        config_mock.return_value = "mycoolref"
+
+        _resolve_module_output(module, "mycooloutput")
+
+        config_mock.assert_called_with("mycoolref")
+        resolve_var_mock.assert_called_with(module, "mycoolref")
```
moelasmar, line 146: can we assert that we return what we received from `_resolve_module_variable`

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,41 @@ def test_resource_full_address_root_module(self):
...
+
+    @patch("samcli.hook_packages.terraform.hooks.prepare.resource_linking._get_configuration_address")
+    def test_resolve_module_output_with_module(self, config_mock):
+        module = TFModule("", None, {}, [], {}, {"mycooloutput": References(["module.mycoolmod"])})
```
moelasmar, line 150: `References(["module.mycoolmod"]` is missing the `mycoolmod` module output name

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +99,62 @@ def _get_configuration_address(address: str) -> str:
...
+        results.append(output)
+    elif isinstance(output, References):
+        cleaned_references = _clean_references_list(output_value)
+
+        for reference in cleaned_references:
+            if reference.startswith("var."):
+                stripped_reference = _get_configuration_address(reference[4:])
+                results += _resolve_module_variable(module, stripped_reference)
+            elif reference.startswith("module."):
+                # aaa.bbb.ccc => bbb
```
moelasmar, line 145: update this comment to `module.bbb.ccc` to be easiest for reader to map why you use the su bstring [7:..]

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +99,62 @@ def _get_configuration_address(address: str) -> str:
...
+        The value to resolve
+
+    Returns
+    -------
+    List[Union[ConstantValue, ResolvedReference]]
+        A list of resolved values
+    """
+    results: List[Union[ConstantValue, ResolvedReference]] = []
+
+    output = module.outputs[output_name]
```
moelasmar, line 132: use get instead of `[]`, and please raise an exception if it is not there

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,41 @@ def test_resource_full_address_root_module(self):
...
+        module2 = TFModule("module.mycoolmod", module, {}, [], {}, {"mycoolmod": ConstantValue("mycoolconst")})
+        module.child_modules.update({"mycoolmod": module2})
+
+        config_mock.return_value = "mycoolmod"
+
+        results = _resolve_module_output(module, "mycooloutput")
+
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].value, "mycoolconst")
```
moelasmar, line 159: I am not sure how this test case passed :)

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,41 @@ def test_resource_full_address_root_module(self):
...
+            (TFModule("", None, {}, [], {}, {"mycooloutput": ConstantValue("mycoolconst")}),),
+            (TFModule("", None, {}, [], {}, {"mycooloutput": References(["mycoolconst"])}),),
+        ]
+    )
+    @patch("samcli.hook_packages.terraform.hooks.prepare.resource_linking._get_configuration_address")
+    def test_resolve_module_output_already_resolved(self, module, config_mock):
+        results = _resolve_module_output(module, "mycooloutput")
+
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].value, "mycoolconst")
```
moelasmar, line 172: can we validate that in case of `References(["mycoolconst"])`, the output is of type `ResolvedReference` and has the correct address.

moelasmar (changes requested):
left couple of comments

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
lucashuy: So for this test case, both the name of the module and output were the same name (I only set the reference as `module.mycoolmod` and not `module.mycoolmod.output`). Since I didn't have an output, the function managed to set both `module_name` and `output_name` to the same thing, but since the output name happened to be the same as the module name in the test, everything worked out fine and it managed to pass.

I can probably add a check in the function to make sure that we have a reference in the format `module.name.output` and throw an exception if it isn't there.

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
lucashuy: At the moment, this method isn't implemented (won't actually return anything). Additionally, should we be testing the output of a different method in this test, since this test is for outputs?

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +104,111 @@ def _get_configuration_address(address: str) -> str:
...
+        for reference in cleaned_references:
+            if reference.startswith("var."):
+                LOG.debug(
+                    "Resolving variable reference {%s} for module {%s} for output {%s}",
+                    reference,
+                    module.full_address,
+                    output_name,
+                )
+
+                stripped_reference = _get_configuration_address(reference[4:])
```
sriram-mv, line 169: anytime absolute numbers are used one has to be careful, would len("var.") be better?

Also if we are to use multiple if else (in this case its 3 (if, elif, else)), it might be better to use a dispatcher pattern. eg: https://www.oreilly.com/library/view/python-cookbook/0596001673/ch01s07.html

lucashuy: Yup, changing the magic number to `len("var.")` will be nicer. Might also consider `reference[reference.find("."):]` to be inline with the other if case that uses `rfind()`

Will look into the dispatcher pattern.

lucashuy: So if we were to go with a dispatching design, we would still need some way to handle the `else` case. Since Python doesn't have a switch-like statement until 3.10, we would need to do some checking if `var` or `module` was in the dictionary and throw all the rest of the cases into a default function/branch. From what I can make out, we would still need some `if` or `try/except` to catch a `KeyError` for "default" cases.

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
moelasmar: My comment was not clear, sorry for that. What I meant is to let this mock function return some value, and let's your unit test validate that what `_resolve_module_output` return is what `_resolve_module_variable` mock function return. In your testing you validate that we call `_resolve_module_variable` with the correct value, but you do not validate how `_resolve_module_output` will use `_resolve_module_variable` output.

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
moelasmar: I think we should validate this case. `module.name` is not a correct format, and we should not be able to know what is the output name.

sriram-mv (commented):
Overall this looks good, I will defer to @moelasmar 's comments being addressed before approving.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -99,13 +104,111 @@ def _get_configuration_address(address: str) -> str:
...
+                results += _resolve_module_output(child_module, output_name)
+            else:
+                LOG.debug(
+                    "Resolved reference {%s} for module {%s} for output {%s}",
+                    reference,
+                    module.full_address,
+                    output_name,
+                )
+
+                results.append(ResolvedReference(reference, module.full_address or ""))
```
torresxb1, line 208: Haven't reviewed the whole PR but remembered that we discussed this part in the office. Let's make the module address optional instead, and set this as None here (instead of empty string) 🙏

moelasmar (approved):
it looks good to me, thanks Lucas :)

sriram-mv (approved):

PR #4260, opened 2022-09-23
feat: Resolve Module Variables

mildaniel:
Iterate over a module's variables and resolve them to either their final constant value or a reference that we don't yet fully resolve (e.g. locals).

Adds recursive function for module variable resolution.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
@@ -2,7 +2,6 @@
 Use Terraform plan to link resources together
 e.g. linking layers to functions
 """
-
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Union
 import re
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
         The address clean of indices
     """
     return re.sub(r"\[[^\[\]]*\]", "", address)
+
+
+def _resolve_module_output(module, output_name):
+    pass
+
+
+def _resolve_module_variable(module: TFModule, variable_name: str) -> List[Union[ConstantValue, ResolvedReference]]:
+    # return a list of the values that resolve the passed variable
+    # name in the input module.
+    results: List[Union[ConstantValue, ResolvedReference]] = []
+
+    var_value = module.variables[variable_name]
+
+    # check the possible constant value for this variable
+    if isinstance(var_value, ConstantValue) and var_value is not None:
+        results.append(ConstantValue(var_value.value))
+
+    # check the possible references value for this variable
+    if isinstance(var_value, References) and var_value is not None:
+        cleaned_references = _clean_references_list(var_value.value)
+        for reference in cleaned_references:
+            # refer to a variable passed to this module from its parent module
+            if reference.startswith("var."):
+                config_var_name = _get_configuration_address(reference[4:])
+                if module.parent_module:
+                    results += _resolve_module_variable(module.parent_module, config_var_name)
+            # refer to another module output. This module will be defined in the same level as this module
+            elif reference.startswith("module."):
+                module_name = reference[reference.find(".") + 1 : reference.rfind(".")]
+                config_module_name = _get_configuration_address(module_name)
+                output_name = reference[reference.rfind(".") + 1 :]
+                if module.parent_module and module.parent_module.child_modules:
+                    results += _resolve_module_output(
+                        module.parent_module.child_modules[config_module_name], output_name
+                    )
+            # this means either a resource, data source, or local.variables.
+            elif module.full_address is not None:
+                results.append(ResolvedReference(reference, module.full_address))
+
+    return results
--- a/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
@@ -150,7 +150,7 @@ def _resolve_module_variable(module: TFModule, variable_name: str) -> List[Union
             LOG.debug("Resolving reference: %s", reference)
             # refer to a variable passed to this module from its parent module
             if reference.startswith("var."):
-                config_var_name = _get_configuration_address(reference[4:])
+                config_var_name = _get_configuration_address(reference[len("var.") :])
                 if module.parent_module:
                     results += _resolve_module_variable(module.parent_module, config_var_name)
             # refer to another module output. This module will be defined in the same level as this module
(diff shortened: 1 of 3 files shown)


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
...
+                if module.parent_module:
+                    results += _resolve_module_variable(module.parent_module, config_var_name)
+            # refer to another module output. This module will be defined in the same level as this module
+            elif reference.startswith("module."):
+                module_name = reference[reference.find(".") + 1 : reference.rfind(".")]
+                config_module_name = _get_configuration_address(module_name)
+                output_name = reference[reference.rfind(".") + 1 :]
+                if module.parent_module and module.parent_module.child_modules:
+                    results += _resolve_module_output(
+                        module.parent_module.child_modules[config_module_name], output_name
```
moelasmar, line 144: can we use get instead of `[]`

moelasmar: also, can we add an exception case if the module name is not in the child modules.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
...
+                    )
+            # this means either a resource, data source, or local.variables.
+            elif module.full_address is not None:
```
moelasmar, line 147: can we add an else case to raise an exception, as it means that there is something wrong in the plan or in our logic.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
         The address clean of indices
     """
     return re.sub(r"\[[^\[\]]*\]", "", address)
+
+
+def _resolve_module_output(module, output_name):
+    pass
+
+
+def _resolve_module_variable(module: TFModule, variable_name: str) -> List[Union[ConstantValue, ResolvedReference]]:
```
moelasmar, line 117: can we add some debugging logs to help us in issues investigation later

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,144 @@ def test_resource_full_address_root_module(self):
...
+            resources=[],
+            child_modules={},
+            outputs={},
+            full_address="",
+            parent_module=None,
+        )
+        results = _resolve_module_variable(module, "layer")
+        expected_result = "layer.arn"
+        for result in results:
+            self.assertEqual(result.value, expected_result)
```
moelasmar, line 149: I am wondering why do you do it in loop!! .. we should expect only one item in this list. I think we should add an assert to validate that the list length is only 1 and it contains the value we expect.

tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
```diff
@@ -131,3 +132,144 @@ def test_resource_full_address_root_module(self):
...
+            resources=[],
+            child_modules={},
+            outputs={},
+            full_address="",
+            parent_module=parent_module,
+        )
+        results = _resolve_module_variable(child_module, "layer")
+        expected_result = "layer.arn"
+        for result in results:
+            self.assertEqual(result.value, expected_result)
```
moelasmar, line 178: can we replace it with validating the list as we expect .. this loop can pass if there are multiple items returned in the loop but all of them are the same value.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
...
+                module_name = reference[reference.find(".") + 1 : reference.rfind(".")]
+                config_module_name = _get_configuration_address(module_name)
+                output_name = reference[reference.rfind(".") + 1 :]
+                if module.parent_module and module.parent_module.child_modules:
+                    results += _resolve_module_output(
+                        module.parent_module.child_modules[config_module_name], output_name
+                    )
+            # this means either a resource, data source, or local.variables.
+            elif module.full_address is not None:
+                results.append(ResolvedReference(reference, module.full_address))
```
moelasmar, line 148: I think we should return the parent module address instead of the current module address. If  the value of the variable is not a constant value, so it means it got passed from the parent module.

moelasmar (changes requested):
I left couple of comments

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +108,43 @@ def _get_configuration_address(address: str) -> str:
...
+def _resolve_module_output(module, output_name):
+    pass
+
+
+def _resolve_module_variable(module: TFModule, variable_name: str) -> List[Union[ConstantValue, ResolvedReference]]:
+    # return a list of the values that resolve the passed variable
+    # name in the input module.
+    results: List[Union[ConstantValue, ResolvedReference]] = []
+
+    var_value = module.variables[variable_name]
```
moelasmar, line 122: can we used `get` instead of `[]`, and please raise an exception if it is not there

moelasmar (approved):
It looks good to me. Thanks Daniel :)

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -109,3 +117,60 @@ def _get_configuration_address(address: str) -> str:
...
+
+    # check the possible references value for this variable
+    if isinstance(var_value, References) and var_value is not None:
+        LOG.debug("Found references (%s) in module (%s)", var_value.value, module.module_name)
+        cleaned_references = _clean_references_list(var_value.value)
+        for reference in cleaned_references:
+            LOG.debug("Resolving reference: %s", reference)
+            # refer to a variable passed to this module from its parent module
+            if reference.startswith("var."):
+                config_var_name = _get_configuration_address(reference[4:])
```
sriram-mv, line 153: `4:` should we look at len("var.") instead?

sriram-mv (approved):
approved with a comment.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: I like this for readability. Updated.

PR #4264, opened 2022-09-28
Add integ test for sam build terraform image based Lambda Function lo…

hawflau:
- Adding integration tests for `sam build` (Terraform Hook) with Image-based Lambda Functions

- Add integration tests
- Fix bug

No

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/testdata/buildcmd/terraform/image_based_lambda_functions_local_backend/main.tf
```diff
@@ -0,0 +1,90 @@
+provider "aws" {
+    region = "us-west-1"
```
sriram-mv, line 2: we create buckets in us-east-1 in our own integration test class code, do these 2 being different not matter?

tests/integration/buildcmd/test_build_terraform_applications.py
```diff
@@ -69,6 +68,91 @@ def run_command(self, command_list, env=None, input=None):
...
+    @classmethod
+    def tearDownClass(cls):
+        """Clean up and delete the bucket if it is not pre-created"""
+        cls.s3_bucket.objects.all().delete()
+        time.sleep(S3_SLEEP)
+        if not cls.pre_created_bucket:
+            cls.s3_bucket.delete()
+
+    def setUp(self):
+        self.backend_key = str(Path("terraform-backend") / str(uuid.uuid4()))
```
sriram-mv, line 129: does using '/' automatically join? if so should we use pathlib join operation instead as this may cause issues on windows.

sriram-mv (commented):
Mostly looks good, just some questions.

tests/integration/buildcmd/test_build_terraform_applications.py
hawflau: This should be fine. `Path("some/path")` will produce an instance of either `WindowsPath("some/path")` or `PosixPath("some/path"), depending on the OS. The `/` operator is just similar to pathlib join. 

On Windows:
```
>>> str(Path("some/path") / "child")
'some\\path\\child'
>>> Path("some/path") / "child"
WindowsPath('some/path/child')
>>>
```

tests/integration/testdata/buildcmd/terraform/image_based_lambda_functions_local_backend/main.tf
hawflau: yeah, it does not matter. The buckets being created are for storing the backend state file generated by `terraform`, which is not related where the resources in the `tf` files are deployed.

tests/integration/buildcmd/test_build_terraform_applications.py
```diff
@@ -249,16 +308,51 @@ def test_build_and_invoke_lambda_functions(self, function_identifier):
...
-class TestBuildTerraformApplicationsWithImageBasedLambdaFunctionAndLocalBackend(BuildTerraformApplicationIntegBase):
-    pass
+@skipIf(
+    not CI_OVERRIDE,
+    "Skip Terraform test cases unless running in CI",
+)
+class TestBuildTerraformApplicationsWithImageBasedLambdaFunctionAndS3Backend(
+    BuildTerraformApplicationS3BackendIntegBase
+):
+    terraform_application = Path("terraform/image_based_lambda_functions_local_backend")
```
sriram-mv, line 329: `TestBuildTerraformApplicationsWithImageBasedLambdaFunctionAndS3Backend` and terraform application seems to contradict, because the terraform application still talks about a `local_backend` , am I missing something?

sriram-mv (commented):
Generally looks good.

tests/integration/buildcmd/test_build_terraform_applications.py
hawflau: my bad. I corrected the path, and tested it in my laptop.

sriram-mv (approved):

lucashuy (approved):

PR #4265, opened 2022-09-28
Update makefile rule generate to create temp backend override

hawflau:
n/a

- The makefile rule for the terraform hook unavoidably modifies the backend state file. This could be undesirable to some customers, particularly those who use remote backend.

- update Makefile rule for build to create a temp backend file and override to use it. As the Make build workflow runs in a scratch directory, the temp backend state will be deleted after the build completes.

No

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -918,14 +919,38 @@ def _generate_makefile_rule_for_lambda_resource(
...
+        f"{apply_command_recipe}"
+        f"{show_command_recipe}"
+        f"{rm_override_file_recipe}"
+        f"{rm_backend_file_recipe}"
+    )
+
+
+def _generate_and_format_init_local_backend_recipe(override_filename: str, backend_filename: str) -> str:
+    command_list = [
+        f'echo "terraform {{" > {override_filename}',
```
moelasmar, line 946: do you think this way will work in windows ?

moelasmar (commented):
left a question, we need to test the generated make file in windows

samcli/hook_packages/terraform/hooks/prepare/hook.py
hawflau: Right, this might not work. Maybe it's better to create another python script to fulfill this purpose. Let's discuss about what the best approach is. And then I will update it.

hawflau: After discussion, this PR only focuses on making it work for linux and mac. We will have another task to make it work for Windows

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -46,6 +47,7 @@
 PropertyBuilderMapping = Dict[str, PropertyBuilder]
 
 TERRAFORM_BUILD_SCRIPT = "copy_terraform_built_artifacts.py"
+TF_BACKEND_OVERRIDE_FILENAME = "z_samcli_backend_override.tf"
```
moelasmar, line 50: as we discussed, let's name it `z_samcli_backend_override` and when we copy it in the makefile we can set the correct name

moelasmar (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -813,6 +818,22 @@ def _generate_makefile(
...
+    """
+    Generates an override tf file to use a temporary backend
+
+    Parameters
+    ----------
+    output_directory_path: str
+        the output directory path to write the generated makefile
+    """
+    statefile_filename = f"{uuid.uuid4()}.tfstate"
+    override_content = "terraform {\n" '  backend "local" {\n' f'    path = "./{statefile_filename}"\n' "  }\n" "}\n"
```
mildaniel, line 831: Do the file contents matter here and will it be read/executed on WIndows? If so should we use `os.linesep`?

mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
hawflau: We will have another task to address the issues on Windows. Potentially we will generate make rules differently based on OS.

PR #4270, opened 2022-09-30
Feat/sam pipeline OIDC

xazhao:
Adding OIDC support to SAM Pipeline.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


requirements/base.txt
```diff
@@ -15,6 +15,7 @@ serverlessrepo==0.1.10
 aws_lambda_builders==1.19.0
 tomlkit==0.7.2
 watchdog==2.1.2
+pyopenssl==22.0.0
```
mndeveci, line 18: Please run this step in PR description to update reproducable linux file;
> make update-reproducible-reqs

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -5,16 +5,42 @@
 from textwrap import dedent
 from typing import Any, Dict, List, Optional
 
+import logging
```
mndeveci, line 8: Please move non-required imports into command execution block so that running `--help` won't need to import any of these. See an example here; https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/sync/command.py#L244

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -141,14 +251,63 @@ def do_cli(
...
+    gitlab_config = GitLabOidcConfig(
+        gitlab_group=gitlab_group, gitlab_project=gitlab_project, deployment_branch=deployment_branch
+    )
+    github_config = GitHubOidcConfig(
+        github_org=github_org, github_repo=github_repo, deployment_branch=deployment_branch
+    )
+    bitbucket_config = BitbucketOidcConfig(bitbucket_repo_uuid=bitbucket_repo_uuid)
+    if config_parameters:
+        saved_provider = config_parameters.get(PERMISSIONS_PROVIDER)
+        if saved_provider == "OpenID Connect (OIDC)":
```
mndeveci, line 291: I would recommend moving `"OpenID Connect (OIDC)"` into a static variable

samcli/lib/cookiecutter/template.py
```diff
@@ -121,6 +121,7 @@ def run_interactive_flows(self, context: Optional[Dict] = None) -> Dict:
         """
         try:
             context = context if context else {}
+            context["shared_values"] = "default"
```
mndeveci, line 124: Can you explain why this is needed? And can you also confirm that it won't break `sam init` flow?

samcli/lib/utils/managed_cloudformation_stack.py
```diff
@@ -33,6 +33,62 @@ def get(self, key) -> Optional[str]:
             return None
 
 
+def update_stack(
```
mndeveci, line 36: Are these changes going to affect SAM CLI's managed stack, or they are specific to Pipeline command's managed stack?

mndeveci (changes requested):

requirements/base.txt
xazhao: Sure I will update the file

samcli/commands/pipeline/bootstrap/cli.py
xazhao: Looking at the import. I don't think it's needed. Will remove it. Thanks for catching it.

samcli/lib/utils/managed_cloudformation_stack.py
xazhao: Yes this is only used in Pipeline command's managed stack.

samcli/commands/pipeline/bootstrap/cli.py
mndeveci: Not at the moment. We used to have something in development guide or in contributing but couldn't find it now.

mndeveci: OK it was in design document; 
> Each command, along with any subcommands, is implemented using Click annotations. They are not directly wired with the core of the CLI. Instead, commands are dynamically loaded into the CLI at run time by importing the Python package implementing the command. For example, assuming two commands are implemented at Python packages foo.cli.cmd1 and foo.cli.cmd2, then the CLI framework will dynamically import these two packages and connect them to parent Click instance. The CLI framework expects the command's Click object to be exposed through an attribute called cli.

https://github.com/aws/aws-sam-cli/blob/develop/DESIGN.md

samcli/lib/cookiecutter/template.py
xazhao: No it won't break the `sam init` flow. This value is used in `sam pipeline init` templates to utilize previous answer to skip some questions.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -9,12 +9,37 @@
...
+    DEPLOYMENT_BRANCH,
+    GITHUB_ORG,
+    GITHUB_REPO,
+    GITLAB_GROUP,
+    GITLAB_PROJECT,
+    OIDC_CLIENT_ID,
+    OIDC_PROVIDER,
+    OIDC_PROVIDER_URL,
+    PERMISSIONS_PROVIDER,
+    OIDC_SUPPORTED_PROVIDER,
```
mndeveci, line 36: This variable `OIDC_SUPPORTED_PROVIDER` is not been used in command parameters. Just to clarify my earlier comment, please move any imports that is not used until line 269 into that line.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -241,5 +451,13 @@ def _load_saved_pipeline_user_arn() -> Optional[str]:
     return config.get("pipeline_user")
 
 
+def _load_config_values() -> Dict[str, str]:
+    samconfig: SamConfig = SamConfig(config_dir=PIPELINE_CONFIG_DIR, filename=PIPELINE_CONFIG_FILENAME)
+    if not samconfig.exists():
+        return {}
+    config: Dict[str, str] = samconfig.get_all(cmd_names=_get_bootstrap_command_names(), section="parameters")
```
mndeveci, line 458: Why we are loading parameters like this [instead of this](https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/sync/command.py#L97)?

samcli/commands/pipeline/bootstrap/cli.py
xazhao: Got it. Thanks for catching it. Updated the imports

samcli/commands/pipeline/bootstrap/cli.py
xazhao: We do have this at line 59. This loading parameters part was used for some conditional OIDC config update. I tried using `@configuration_option` only but it doesn't work as we expect.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -9,13 +9,20 @@
...
 from samcli.lib.config.samconfig import SamConfig
-from samcli.lib.pipeline.bootstrap.stage import Stage
+
+from samcli.commands.pipeline.bootstrap.oidc_config import (
+    BitbucketOidcConfig,
+    GitHubOidcConfig,
+    OidcConfig,
+    GitLabOidcConfig,
+)
+
```
mndeveci, line 22: Apologies for been picky, but these imports can be moved down as well.

mndeveci (dismissed):

samcli/commands/pipeline/bootstrap/cli.py
xazhao: These imports are also used in another function as input, so I think they need to be at the top. Correct me if I'm wrong.

samcli/commands/pipeline/bootstrap/cli.py
```diff
@@ -87,6 +96,66 @@
...
+)
+@click.option(
+    "--oidc-provider",
+    help="The name of the CI/CD system that will be used for OIDC permissions "
+    "we currently only support GitLab, GitHub, and Bitbucket",
+    type=click.Choice([GITHUB_ACTIONS, GITLAB, BITBUCKET]),
+    required=False,
+)
+@click.option(
+    "--gitlab-group",
```
sriram-mv, line 139: Some of these options don't play well with each other I'm assuming. There needs to be appropriate error messaging if they are specified together.

samcli/commands/pipeline/bootstrap/cli.py
sriram-mv: we are duplicating this? I'm still not sure why its stated as not working.

samcli/lib/pipeline/bootstrap/stage_resources.yaml
```diff
@@ -94,31 +120,55 @@ Resources:
...
+                    aws:PrincipalTag/Role: aws-sam-pipeline-codebuild-service-role
+          - Fn::Sub: 
+              - |
+                {
+                  "Version": "2012-10-17",
+                  "Statement": [
+                    {
+                      "Effect": "Allow",
+                      "Principal": {
+                        "Federated": "arn:aws:iam::${AWS::AccountId}:oidc-provider/${Url}"
```
sriram-mv, line 158: use AWS::Partition instead of :aws: , other places need to be cross checked as well.

sriram-mv (changes requested):
This is a large PR and will require another pass.

samcli/lib/pipeline/bootstrap/stage_resources.yaml
xazhao: Good catch! Will update it.

samcli/commands/pipeline/bootstrap/cli.py
xazhao: Currently it takes all options then decides which option to use based on what `oidc-provider` is. I see your point is helpful. I'm thinking we can leave it to another PR for improvements which includes some refactoring.

hawflau: I learned from @xazhao that these new options, if not applicable, will just be ignored. I think it makes sense to handle them for now, as at least these different groups (GH Action, Gitlab and Bitbucket) won't conflict each other. But IMO there's a room to improve in terms of UX, to give customers clearer messages about what options they really need to supply.

samcli/commands/pipeline/bootstrap/guided_context.py
```diff
@@ -144,43 +183,181 @@ def _prompt_image_repository(self) -> None:
         else:
             self.create_image_repository = False
 
+    def _prompt_permissions_provider(self) -> None:
+        click.echo("Select a user permissions provider:")
+        click.echo("\t1 - IAM (default)")
+        click.echo("\t2 - OpenID Connect (OIDC)")
+        user_provider = click.prompt("Choice", type=click.Choice((["1", "2"])), show_default=False, default="1")
+        self.permissions_provider = OPEN_ID_CONNECT if user_provider == "2" else "iam"
```
hawflau, line 191: nit: why not also use a constant for "iam"?

xazhao: Good catch. Updating it to a constant.

samcli/commands/pipeline/bootstrap/cli.py
hawflau: looks like they are used in other functions only for type indication. We can wrap them by quotes instead, e.g.
```
def my_function(arg: "MyClass"):
    pass
```

hawflau (changes requested):

samcli/commands/pipeline/bootstrap/cli.py
xazhao: Good suggestion! Make it work. Thanks!

sriram-mv (approved):
Conditionally approving this based on following assumptions.

* Improved error messaging when conflicting options are chosen.
* Improved handling on changing the entities referenced within the configuration file through `parameters`
* Showcase in the help text which options are related and go together, there are a lot of options for this command at present.
* Lazy imports within cli() instead of top level imports, this causes the help text loading to slow down.

xazhao:
> Conditionally approving this based on following follow-on improvements.
> 
>     * Improved error messaging when conflicting options are chosen.
> 
>     * Improved handling on changing the entities referenced within the configuration file through `parameters`
> 
>     * Showcase in the help text which options are related and go together, there are a lot of options for this command at present.
> 
>     * Lazy imports within cli() instead of top level imports, this causes the help text loading to slow down.

Thanks! I will create SIMs to keep track of each improvement.

hawflau (approved):

Issue #4012, opened 2022-06-28
Bug: Incorrect required parameter combinations for `no-interactive` Flag


kmfischer3:
### Description:

There is a discrepancy between the required parameter combinations for the `no-interactive` flag and the SAM CLI output from using that flag.

According to the CLI output, one of the following parameter combinations must be used with the `no-interactive` flag:
- `--name` and `--location`
- `--name` `--package-type` and `--base-image`
- `--name` `--runtime` `--dependency-manager` and `--app-template`

However, when I use the first option, I still get an error:

```
$ sam init --no-interactive --name hello --location sam-init-template
Usage: sam init [OPTIONS]

Error: Missing parameter --dependency-manager
```

Strangely, even if I _do_ provide the requested flag, it's still not happy because then my flags don't match one of the accepted combinations:

```
$ sam init --no-interactive --name hello --location sam-init-template --dependency-manager npm
Error: You must not provide both the --location and --dependency-manager parameters.
...
```

### Steps to reproduce:

1. Start with a template (I'm using a cookiecutter template similar to [this](https://github.com/aws-samples/cookiecutter-aws-sam-python)) named `sam-init-template` in your current working directory
2. run `sam init --no-interactive --name hello --location sam-init-template`

### Observed result:

`Error: Missing parameter --dependency-manager`

### Expected result:

The CLI should accept the required parameter combination `--name` and `--location` without needing `--dependency-manager`.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOS Monterey 12.4
4. `sam --version`: 1.51.0
5. AWS region: N/A

`Add --debug flag to command you are running` 
N/A

kmfischer3:
Hey @hawflau sorry for not being clear - I _want_ the `no-interactive` flag because I want the disable interactive prompting for the init parameters.

I'm trying to automatically test the `sam init` process for our custom template, so I don't want any prompts.

Thanks :)

hawflau:
Right, it makes sense.

The limitation about `--no-interactive` and `--dependency-manager` was introduced in the #2463. I'm not sure if it was a miss or something. But I think we need to revisit and fix it.

qingchm:
Hey @kmfischer3 does the solution suggested above help?

torresxb1:
Closing as a workaround was provided. Team will track improving the clarity of the messaging.

PR #4288, opened 2022-10-07
fix: Fixed requiring dep manager when using location flag

lucashuy:
#4012 

This changes updates the `--no-interactive` check to consider `--location`. Post change, users can specify `--no-interactive --location` again without having to include `--dependency-manager`.

Add an additional check in the if condition to check for `--location`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/init/command.py
```diff
@@ -118,9 +118,12 @@ def wrapped(*args, **kwargs):
             package_type = ctx.params.get("package_type")
             base_image = ctx.params.get("base_image")
             dependency_manager = ctx.params.get("dependency_manager")
+            location = ctx.params.get("location")
+
+            java_base_image = base_image and "java" in base_image
 
             # dependency manager is only required for ZIP package type and for java based IMAGE package types
```
moelasmar, line 125: can you update the comment

moelasmar (commented):
can you add an integration test case to cover this update?

samcli/commands/init/command.py
```diff
@@ -118,9 +118,13 @@ def wrapped(*args, **kwargs):
             package_type = ctx.params.get("package_type")
             base_image = ctx.params.get("base_image")
             dependency_manager = ctx.params.get("dependency_manager")
+            location = ctx.params.get("location")
 
-            # dependency manager is only required for ZIP package type and for java based IMAGE package types
-            if package_type == ZIP or (base_image and "java" in base_image):
+            java_base_image = base_image and "java" in base_image
```
sriram-mv, line 123: Why are we bleeding this information over? especially checking for a string in a base_image?

sriram-mv (changes requested):

samcli/commands/init/command.py
lucashuy: Sorry not sure if I 100% follow, but the base image option is optional and can has a couple of java options (`amazon/java11-base`, `amazon/java8.al2-base`, `amazon/java8-base`) hence the `in` check.

moelasmar (approved):
It looks good to me. Thanks Lucas

samcli/commands/init/command.py
sriram-mv: Could you add a fixme/todo with respect to our code organization here, we are looking up an arbitrary string in the base image to determine if its a java image.

sriram-mv (approved):
Approving with conditional feedback.

PR #4293, opened 2022-10-10
Update options.py  help message

ccannell67:
Remove reference to `--s3-guided parameter` as it is not a valid parameter for `sam deploy`


sriram-mv:
Nice catch! thanks for the PR. I believe it should be do not use ```--guided``` with this option.

mndeveci (approved):

PR #4295, opened 2022-10-11
feat: Add logic for linking functions to layers

mildaniel:
We need the TF prepare hook to call our layer linking logic.

Add the connection logic from the hook to iterate through the resources in the TF modules and call the layer linking logic for each resource.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar (approved):
it looks good to me. Thanks Daniel :)

sriram-mv (approved):

Issue #4289, opened 2022-10-08
`pywintypes-stubs` is now available through `types-pywin32`


Avasam:
Just a heads-up that `types-pywin32` has been added to typeshed, so you don't need to ignore `pywintypes` for mypy and pylint anymore.

sriram-mv:
Thanks for the issue! would you be open for creating a PR on this? :)

Avasam:
> Thanks for the issue! would you be open for creating a PR on this? :)

I had to fight with the CI a bit, but it's done: #4297

jfuss:
Typically, we only close issues once released. Since this is for typing/developing and the PR is merged, going to close the issue now. 

Thanks for the issue and contribution @Avasam.

PR #4297, opened 2022-10-12
Add type-stubs and remove redundant ignores

Avasam:

#4289

More accurate type-checking

`types-pywin32` has been added to typeshed, so you don't need to ignore `pywintypes` for mypy and pylint anymore.
`types-PyYAML`, `types-chevron`, `types-psutil` were also already on Typeshed

No runtime effect.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes (has a file not found error locally)
- [x] `make update-reproducible-reqs` if dependencies were changed (I can't run the script and I'm on windows. Feel free to do so for me and update this PR)
- [x] Write documentation


samcli/local/docker/utils.py
```diff
@@ -85,7 +85,7 @@ def is_docker_reachable(docker_client):
     if platform.system() == "Windows":
         import pywintypes  # pylint: disable=import-error
```
lucashuy, line 86: question: Is this suppression still required? Noticed that it was removed in `test_manager.py` but remains here.

lucashuy (commented):
Looks good to me so far! Left a question about one of the pylint suppression

samcli/local/docker/utils.py
lucashuy: Ah, saw the commit where you had added it back, so I'm guessing this is still required 😃

Avasam: Indeed, Pylint is not great at working with stubs yet. While this wasn't an issue locally, there could be a bunch of different factors why it's different on the CI. And opted to just add the suppression back, rather than waste time trying to figure this specific oddity or start updating too many dependencies.

Avasam:
I've realized yesterday that there's other missing stubs, not just pywin32. May as well do it at once.

samcli/commands/local/lib/local_lambda.py
```diff
@@ -152,8 +152,7 @@ def invoke(
             # trying to connect to the socket for Docker it would throw ContainerResponseException but now it's this.
             LOG.info(str(e))
         except OSError as os_error:
-            # pylint: disable=no-member
-            if hasattr(os_error, "winerror") and os_error.winerror == 1314:  # type: ignore
+            if getattr(os_error, "winerror", None) == 1314:
```
mndeveci, line 155: Isn't there still a chance that this could be thrown in a non windows machine? Then would this change be compatible with Linux or Mac?

Avasam: No change of logic here, just simplified the check. If winerror isn't an attribute, getattr returns None, which is not 1314.

mndeveci (approved):

requirements/base.txt
```diff
@@ -20,6 +20,12 @@ pyopenssl==22.0.0
 
 # Needed for supporting Protocol in Python 3.7, Protocol class became public with python3.8
 typing_extensions==3.10.0.0
+# Typeshed type stubs
+types-pywin32==304.0.0.1
```
sriram-mv, line 24: will this cause issues when installing on a linux/mac system, the classifiers on the project indicate it to be just for windows.

hawflau: @Avasam Can you please address this feedback? Thanks!

Avasam: Thanks for the ping, I forgot about this.

The stub package only contains type stubs (`.pyi files`). There's no functional code.
At worst they'll keep the typing errors in their IDE with `pywintypes` if they *don't* install it.

requirements/dev.txt
```diff
@@ -21,4 +21,4 @@ pytest-json-report==1.5.0
 
 # formatter
 black==22.6.0
-psutil==5.9.0
\ No newline at end of file
+psutil==5.9.0
```
jfuss, line 24: What changed in this file?

requirements/base.txt
```diff
@@ -21,6 +21,12 @@ pyopenssl==22.0.0
...
+types-PyYAML==6.0.12
```
jfuss, line 26: Why are we adding these to `base.txt` instead of `dev.txt`? We shouldn't need these outside of development right?

jfuss (changes requested):

requirements/dev.txt
Avasam: Likely EOF

jfuss (approved):

PR #4300, opened 2022-10-12
Get TrimGoPath from build properties

### LINKED ISSUE #3258: sam build includes absolute paths in Golang binaries
### Description:

When you run ```sam build``` and ```sam package``` in two instances of the same projects with different absolute paths and upload to the same S3 bucket, the compiled binaries will be uploaded both times, even though the binaries are identical.
According to #2110 this has been fixed, but I can still reproduce the issue on Windows.

### Steps to reproduce:

1. Clone [this repository](https://github.com/suroh1994/sampackagesample) into a directory on your local Windows machine.
1. ```cd``` into the directory.
1. Run ```sam build```.
1. Run ```sam package --s3-bucket YOUR_BUCKETNAME```
1. Clone the same repository into a second directory on your local windows machine.
1. Copy the second directory into a third directory using ```xcopy seconddirectory thirddirectory /s /e```.
1. ```cd``` into the second directory.
1. Run ```sam build```.
1. Run ```sam package --s3-bucket YOUR_BUCKETNAME```
1. ```cd``` into the third directory.
1. Run ```sam build```.
1. Run ```sam package --s3-bucket YOUR_BUCKETNAME```

### Observed result:

The calculated hashes change, when the parent directory changes. This results in the compiled binaries being re-uploaded for every directory, although no files have been modified.

### Expected result:

Hashes should not depend on the parent directory. Compiled binaries should only be uploaded, when they have been modified.

1. OS: Windows 10 Pro Version 20H2 (Build 19042.1165)
2. `sam --version`: 1.31.0
3. AWS region: eu-central-1

### Further notes:
I can imagine this is also the cause for the issue #2872, where running ```sam package``` in an ephemeral container results in the files being uploaded every time.

scrofungulus:
This PR closes #3258. Changes to the [aws-lambda-builders](https://github.com/aws/aws-lambda-builders) repository have been made to add a `trim_go_path` option with this [pull request](https://github.com/aws/aws-lambda-builders/pull/389).

This change prevents CloudFormation from triggering new Lambda function deployments for the Go runtime when the project has no code changes. 

By building Go binaries with `-trimpath` instead of absolute file system paths, the recorded file names will begin either a module path@version (when using modules), or a plain import path. Without absolute paths enabled the md5 hexdigest of binaries built from the same codebase are always the same and in turn will only replace Lambda functions that have had code changes. 

This should have no side effects because the new option will default to `False`

With this new build property Lambda functions with the Go runtime can now be defined like the following to prevent the above issue: 
```yaml
Resources:
  HelloWorldFunction:
    Properties:
      CodeUri: functions/hello-world/
      Handler: main
      Runtime: go1.x
      Architectures:
        - x86_64
      Events:
        HelloWorld:
          Type: Api
          Properties:
            Path: /hello
            Method: get
    Metadata:
      BuildProperties:
        TrimGoPath: True
``` 

#### Mandatory Checklist

- [x ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


sriram-mv (approved):

moelasmar (commented):
Is it possible to add an integration test case for building a Go lambda function with the new `TrimGoPath` build property?

collen-maxex:
@moelasmar Good question. @sriram-mv please correct me if I'm wrong...but I would think the integration tests written in the [aws-lambda-builder](https://github.com/aws/aws-lambda-builders) repo would suffice. Here are those tests if you would like to [look](https://github.com/aws/aws-lambda-builders/blob/develop/tests/integration/workflows/go_modules/test_go.py#L133-L169) and my [comment](https://github.com/aws/aws-lambda-builders/pull/389#discussion_r990501206) on my thought process. I think since these repositories already have a separation of concerns we should be confident that as long as we're passing the proper build options that we are fine. 

I am happy to add the integration test if necessary.

sriram-mv:
Another thing to consider here is that even if an integration test is added, it will not work yet without a release on `aws-lambda-builders` and then updating the requirements.txt

moelasmar (approved):

scrofungulus:
> Another thing to consider here is that even if an integration test is added, it will not work yet without a release on aws-lambda-builders and then updating the requirements.txt

@sriram-mv based on this comment I'm not sure if you're asking me to take care of some additional steps. Will you be taking over from here by cutting a `aws-lambda-builders` release?

sriram-mv:
@scrofungulus. Yep, we will be cutting a release on aws lambda builders.

PR #4309, opened 2022-10-17
test: Add integration test cases for local testing terraform projects with lambda layers support

moelasmar:
Adding integration test cases for local testing terraform projects that contain both lambda functions and lambda layers.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -207,6 +207,12 @@ def _build_module_resources_from_configuration(module_configuration: Dict, modul
 
         expressions = config_resource.get("expressions", {})
         for expression_name, expression_value in expressions.items():
+            # we do not process the attributes of type dictionary
+            # Todo add dictionary type attributes post beta
+            if isinstance(expression_value, list):
+                LOG.debug("Skip processing the attribute %s as its value is a map.", expression_name)
+                continue
+
```
moelasmar, line 215: Found a use case that is not supported in this code. If the attribute value is a block (dictionary) .. it got translated as a list of dictionaries, and each key in the attribute value dictionary will be an item in this list. 

I skipped processing this attribute for now as it is not affecting this beta version, as no of the attributes that we care about for beta 1 stories has a block value

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -538,9 +544,8 @@ def _resolve_resource_attribute(
 
     attribute_value = resource.attributes.get(attribute_name)
     if attribute_value is None:
```
moelasmar, line 546: if the value of the attribute is None, it means that the customer does not set value for it in the terraform application, and so we should not raise an error from this function. The function caller if it expect a value for this attribute, so it should raise the error

sriram-mv: What does `The function caller if it expect a value for this attribute, so it should raise the error` mean? we raise the error higher up in the chain?

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -638,14 +643,18 @@ def _link_lambda_function_to_layer(
...
     # The agreed limitation to support only 1 lambda layer reference.
     if len(layers) > 1:
         LOG.debug(
             "SAM CLI does not support mapping the lambda function %s to more than one layer.",
             function_tf_resource.full_address,
         )
         raise OneLambdaLayerLinkingLimitationException(layers, function_tf_resource.full_address)
-    _update_mapped_lambda_function_with_resolved_layers(cfn_functions, layers, tf_layers)
+    if not layers:
+        LOG.debug("There is no layers defined for lambda function %s", function_tf_resource.full_address)
```
sriram-mv, line 655: nit: is/are

sriram-mv (approved):
Mostly looks good.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: Yes, in this function, we will not raise an error for this scenario in this function (this is a utility function). So, for example if we used this utility function in a scenario that we expect the resource contains a a specific attribute, and we received it as None, so we should raise this error from the caller function and not the utility function

mildaniel (approved):

tests/integration/local/start_lambda/test_start_lambda_terraform_applications.py
```diff
@@ -95,6 +110,124 @@ def test_invoke_function(self, function_name):
...
+class TestLocalStartLambdaTerraformApplicationWithLayersWithoutBuild(StartLambdaTerraformApplicationIntegBase):
+    terraform_application = "/testdata/invoke/terraform/simple_application_with_layers_no_building_logic"
+    pre_create_lambda_layers = ["simple_layer1", "simple_layer2", "simple_layer3"]
+    template_path = None
+    hook_package_id = "terraform"
+    beta_features = True
+
+    @classmethod
+    def setUpClass(cls):
+        cls.region_name = os.environ.get("AWS_DEFAULT_REGION", "us-east-1")
```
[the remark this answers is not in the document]
moelasmar: We already set this property in the appveyor jobs environment variables by our lambda functions before invoke it, and this is how we did it in the deploy and package commands integration testing. But, in general it should not make any difference

PR #4318, opened 2022-10-18
Create IAM user & Access/Secret keys for all Envs

russellballestrini:
We always want to create IAM user & Access/Secret keys for each isolated stage/environment especially when the environments are located in different AWS Accounts.

https://github.com/aws/aws-sam-cli-pipeline-init-templates/issues/55

When a SAM pipeline uses multiple AWS Accounts we need to create a separate IAM user and access/secret keys for each. 

Always creates the AWS Resources needed for all stages of the SAM Pipeline

Always creates the AWS Resources needed for all stages of the SAM Pipeline

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/pipeline/bootstrap/cli.py
+++ b/samcli/commands/pipeline/bootstrap/cli.py
@@ -265,9 +265,6 @@ def do_cli(
         Stage,
     )
 
-    if not pipeline_user_arn and not permissions_provider == OPEN_ID_CONNECT:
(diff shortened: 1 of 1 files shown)


xazhao:
Replied in https://github.com/aws/aws-sam-cli-pipeline-init-templates/issues/55. This change will ask customers to create multiple IAM Users even they're within the same account. It may not be the way we want to implement now.

PR #4321, opened 2022-10-18
fix: add `MaxAttempts` to `sam delete`

sriram-mv:
* Add better error messaging when a stack fails to delete.

https://github.com/aws/aws-sam-cli/issues/3754

* Try a finite number of times and exit if unsuccessful.

* The waiter config now has a finite number of attempts baked in and when it fails it showcases the last stack status.

* N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/delete/cfn_utils.py
+++ b/samcli/lib/delete/cfn_utils.py
@@ -116,13 +116,16 @@ class CfnUtils:
 
         # Wait for Delete to Finish
         waiter = self._client.get_waiter("stack_delete_complete")
-        # Poll every 5 seconds.
-        waiter_config = {"Delay": 30}
+        # Poll every 5 seconds and set max attempts to be 3.
+        waiter_config = {"Delay": 5, "MaxAttempts": 3}
         try:
             waiter.wait(StackName=stack_name, WaiterConfig=waiter_config)
         except WaiterError as ex:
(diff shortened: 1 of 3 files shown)


samcli/lib/delete/cfn_utils.py
```diff
@@ -116,13 +116,16 @@ def wait_for_delete(self, stack_name):
 
         # Wait for Delete to Finish
         waiter = self._client.get_waiter("stack_delete_complete")
-        # Poll every 5 seconds.
-        waiter_config = {"Delay": 30}
+        # Poll every 5 seconds and set max attempts to be 3.
+        waiter_config = {"Delay": 5, "MaxAttempts": 3}
```
lucashuy, line 120: Quick question: with max tries at 3 and delay at 5 seconds, is it safe to assume that delete operations will take at most 15 seconds?

lucashuy (commented):
Looks good to me in general, left a question

samcli/lib/delete/cfn_utils.py
sriram-mv: This is just for polling, it could take longer, but we poll for status every 5 seconds. 30 seconds seemed long to me.

lucashuy (approved):

hnnasit (approved):
LGTM!

PR #4323, opened 2022-10-19
feat: Update client-side telemetry

mildaniel:
To collect hook-specific data.

Update the telemetry that is sent by sam-cli to the telemetry back end to include hook-specific data.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar (approved):
It looks good to me. Thanks Daniel

samcli/lib/telemetry/metric.py
```diff
@@ -21,7 +24,10 @@
 from samcli.lib.telemetry.event import EventTracker
 from samcli.lib.telemetry.project_metadata import get_git_remote_origin_url, get_project_name, get_initial_commit_hash
 from samcli.commands._utils.experimental import get_all_experimental_statues
+from samcli.lib.hook.hook_config import HookPackageConfig
+from samcli.lib.hook.hook_wrapper import INTERNAL_PACKAGES_ROOT
 from .telemetry import Telemetry
+from ..hook.exceptions import InvalidHookPackageConfigException
```
hawflau, line 30: nit: should we change all relative imports to absolute imports?

samcli/lib/telemetry/metric.py
```diff
@@ -185,6 +202,22 @@ def wrapped(*args, **kwargs):
...
 
 
+def _get_project_details(hook_package_id: str, template_dict: Dict) -> ProjectDetails:
+    if not hook_package_id:
+        project_type = ProjectTypes.CDK.value if is_cdk_project(template_dict) else ProjectTypes.CFN.value
+        return ProjectDetails(project_type=project_type, hook_package_id=None, hook_package_version=None)
+    hook_location = Path(INTERNAL_PACKAGES_ROOT, hook_package_id)
+    try:
+        hook_package_config = HookPackageConfig(package_dir=hook_location)
+    except InvalidHookPackageConfigException:
```
hawflau, line 212: The way we now invoke the hook (via a click option class), if it fails, seems like it won't  trigger the `track_command` method. 

So this seems like an unlikely path that will get triggered.

But we should think about how to capture that kind of failures and make sure we report them via Telemetry.

mildaniel: We discussed this point offline. We generally don't send telemetry for bad click options which is a miss on our end. We will create a ticket internally to track this and follow up.

samcli/lib/hook/hook_config.py
```diff
@@ -87,3 +87,7 @@ def description(self) -> Optional[str]:
     @property
     def functionalities(self) -> Dict[str, HookFunctionality]:
         return cast(Dict[str, HookFunctionality], self._config["functionalities"])
+
+    @property
+    def iac_framework(self) -> str:
+        return cast(str, self._config.get("iac_framework") if self._config.get("iac_framework") else "")
```
hawflau, line 93: Would this be the same?
```suggestion
        return cast(str, self._config.get("iac_framework", ""))
```

hawflau (approved):
approved with a minor comment

samcli/lib/telemetry/metric.py
```diff
@@ -1,29 +1,35 @@
...
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Dict
 
 import click
 
 from samcli import __version__ as samcli_version
 from samcli.cli.context import Context
 from samcli.cli.global_config import GlobalConfig
-from samcli.lib.warnings.sam_cli_warning import TemplateWarningsChecker
+from samcli.commands._utils.experimental import get_all_experimental_statues
```
sriram-mv, line 21: Nothing to do with this particular PR.

but note regardless: this file gets imported a LOT across all commands, so we need to be mindful of the imports we put in here and where possible rely on lazy-imports.

sriram-mv (approved):

PR #4325, opened 2022-10-19
feat: link the lambda resources to the correct s3_object in case if the s3 bucket data are not available in the planned values

moelasmar:
The Lambda resources that are linked to a S3 bucket that will be created in the same terraform project, will have a Null value for the s3 bucket details. This will cause an issue in SAM CLI while preparing the TF project, and could not link this lambda resource to the correct code artifacts.

This PR will depend on the TF configuration section in the `terraform state` incase if the bucket details in the lambda resource is Null to get the reference values, and then use these values to find the matched S3 Object that has the code artifacts details.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


hawflau (commented):
some minor comments first. Will continue dive deeper into the integration test cases

hawflau (approved):

tests/integration/local/start_lambda/test_start_lambda_terraform_applications.py
```diff
@@ -228,6 +246,64 @@ def test_invoke_function(self, function_name, expected_output):
...
+        try:
+            shutil.rmtree(str(Path(self.working_dir).joinpath(".aws-sam-iacs")))
+            shutil.rmtree(str(Path(self.working_dir).joinpath(".terraform")))
+            os.remove(str(Path(self.working_dir).joinpath(".terraform.lock.hcl")))
+        except FileNotFoundError:
+            pass
+
+    @pytest.mark.flaky(reruns=3)
+    def test_invoke_function(self):
+        self.working_dir = self.integration_dir + self.terraform_application
```
lucashuy, line 280: Should we consider moving the directory and/or docker setup to `setUp()` or a helper function? There is only one test for this class and I'm not sure if we even need other tests for this case, but just something to keep in mind

lucashuy (approved):
Looks good to me, left a general comment

Issue #3691, opened 2022-02-24
Support for Click >= 8.0


jdbodyfelt:
### Description:

In a number of my travis builds, I'm getting errors relating to conflicts in versions of `click` used between `aws-sam-cli 1.39.0` and `black 22.1.0` (newest formatter) - the former requiring =~7.1 and the later with >=8.0.0. It would be great if that can be addressed on 1.39.1. 
A quick fix for me is rolling our black back to 20.8 or so, but we'd love to have most stable on that and aws-sam-cli. Thanks!

mingkun2020:
Thanks for the feature request! We have a [roadmap](https://github.com/aws/aws-sam-cli/wiki/SAM-CLI-Roadmap) here, but we also try to take in popular feature requests based on 👍🏽 from the community. We will be adding this to our intake for prioritization.

thyagram-aws:
Hi guys, we ran into issue with the 'click' python package as a dependency of aws-sam-cli (v1.45.0) while working on our solution: https://github.com/aws-solutions / [aws-virtual-waiting-room](https://github.com/aws-solutions/aws-virtual-waiting-room)

Could you look into upgrading the dependency version to >=8.0.0. Python safety package has version 7.1.2 flagged. 
https://pyupio.github.io/safety-db/ 

Running safety (v. 1.10.3) on a sample requirements.txt produces the following output. 
```
$ cat requirements.txt
click==7.1.2
$ safety check -r requirements.txt
+==============================================================================+
|                                                                              |
|                               /$$$$$$            /$$                         |
|                              /$$__  $$          | $$                         |
|           /$$$$$$$  /$$$$$$ | $$  \__//$$$$$$  /$$$$$$   /$$   /$$           |
|          /$$_____/ |____  $$| $$$$   /$$__  $$|_  $$_/  | $$  | $$           |
|         |  $$$$$$   /$$$$$$$| $$_/  | $$$$$$$$  | $$    | $$  | $$           |
|          \____  $$ /$$__  $$| $$    | $$_____/  | $$ /$$| $$  | $$           |
|          /$$$$$$$/|  $$$$$$$| $$    |  $$$$$$$  |  $$$$/|  $$$$$$$           |
|         |_______/  \_______/|__/     \_______/   \___/   \____  $$           |
|                                                          /$$  | $$           |
|                                                         |  $$$$$$/           |
|  by pyup.io                                              \______/            |
|                                                                              |
+==============================================================================+
[...]

PR #4326, opened 2022-10-19
chore: Upgrade Click and Flask

jfuss:
#3691

Keep up to date with dependencies.

Upgrades dependencies.

Shouldn't have any that I know of. I did patch things that needed to be updated but relying on tests to ensure behavior is backwards compatible.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/common/runtime_template.py
```diff
@@ -83,11 +83,13 @@ def get_local_lambda_images_location(mapping, runtime):
     return os.path.join(_lambda_images_templates, runtime, dir_name + "-lambda-image")
 
 
-SUPPORTED_DEP_MANAGERS: Set[str] = {
-    c["dependency_manager"]  # type: ignore
-    for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))
-    if c["dependency_manager"]
-}
+SUPPORTED_DEP_MANAGERS: List[str] = list(
```
jfuss, line 86: This and `LAMBDA_IMAGES_RUNTIMES` was changed to list due to a mypy error: `samcli/commands/init/command.py:195: error: Argument 1 to "Choice" has incompatible type "Set[Any]"; expected "Sequence[str]"`.

Instead of just a list, I could do `list(set()`, to ensure only one value is present.

sriram-mv (approved):

samcli/local/common/runtime_template.py
```diff
@@ -83,11 +83,15 @@ def get_local_lambda_images_location(mapping, runtime):
...
+    set(
+        {
+            c["dependency_manager"]  # type: ignore
+            for c in list(itertools.chain(*(RUNTIME_DEP_TEMPLATE_MAPPING.values())))
+            if c["dependency_manager"]
```
qingchm, line 91: should we use methods like c.get() which seems safer in this context?

qingchm (approved):
LGTM, just commented one NIT opinion

samcli/local/common/runtime_template.py
jfuss: Not critical as this would be caught at build time if `c["dependency_manager"]` throws a `KeyError` but generally a good practice. Not sure why we wrote it this way to begin with but updated to your suggestion.

PR #4344, opened 2022-10-27
fix: building terraform projects in container

moelasmar:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -513,6 +513,11 @@ class ApplicationBuilder:
             manifest_path = self._manifest_path_override or os.path.join(manifest_context_path, config.manifest_name)
 
             # By default prefer to build in-process for speed
+            scratch_dir_path = (
+                LambdaBuildContainer.get_container_dirs(code_dir, manifest_path)["scratch_dir"]
+                if self._container_manager
+                else scratch_dir
+            )
             build_runtime = specified_workflow
             options = ApplicationBuilder._get_build_options(
                 layer_name,
@@ -521,7 +526,7 @@ class ApplicationBuilder:
                 None,
                 metadata=layer_metadata,
                 source_code_path=code_dir,
-                scratch_dir=scratch_dir,
+                scratch_dir=scratch_dir_path,
             )
             if self._container_manager:
                 # None key represents the global build image for all functions/layers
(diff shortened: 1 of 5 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -513,6 +513,11 @@ def _build_layer(
             manifest_path = self._manifest_path_override or os.path.join(manifest_context_path, config.manifest_name)
 
             # By default prefer to build in-process for speed
+            scratch_dir_path = (
```
sriram-mv, line 516: Why is this a change that needs to be made for terraform specific projects?

moelasmar: this is not required only for terraform, but the scratch_dir is different in container, and if there is any runtime that depend on this value it will not work on containers.

tests/integration/buildcmd/test_build_terraform_applications.py
```diff
@@ -25,11 +26,30 @@
...
+                .joinpath("integration", "testdata", "buildcmd", "terraform", "build_image_docker")
+            )
+            # Directly build an image that will be used across all local invokes in this class.
+            for log in cls.client.api.build(
+                path=cls.terraform_sam_build_image_context_path,
+                dockerfile="Dockerfile",
+                tag=cls.docker_tag,
+                decode=True,
+            ):
+                print(log)
```
hawflau, line 52: Should we use `LOG.info` instead?

samcli/lib/build/app_builder.py
sriram-mv: the question I ask is then should this be a change we need to make in `develop` branch as well?. Integration tests should have been failing if that is the case, so I'm thinking thats probably not the case.

moelasmar: I think Terraform is the first thing that depend on the scratch directory on the logic.

tests/integration/buildcmd/test_build_terraform_applications.py
moelasmar: done

hawflau (approved):

sriram-mv (approved):

PR #4349, opened 2022-10-27
docs: ignore pycache when listing available hook package ids in help text

torresxb1:
Help message looks like this:
```
  --hook-package-id TEXT          The id of the hook package to be used to
                                  extend the SAM CLI commands functionality.
                                  As an example, you can use `terraform` to
                                  extend SAM CLI commands functionality to
                                  support terraform applications. Available
                                  Hook Packages Ids ['terraform',
                                  '__pycache__']
```
At the end you can see pycache listed as one of the available hook package ids. We should ensure it doesn't appear there.

Ignores `pycache`.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/hook/hook_wrapper.py
+++ b/samcli/lib/hook/hook_wrapper.py
@@ -191,7 +191,7 @@ def get_available_hook_packages_ids() -> List[str]:
     LOG.debug("Return available internal hook packages")
     hook_packages_ids = []
     for child in INTERNAL_PACKAGES_ROOT.iterdir():
-        if child.is_dir() and child.name != "__pycache__":
+        if child.is_dir() and child.name[0].isalpha():
             hook_packages_ids.append(child.name)
 
     return hook_packages_ids
--- a/samcli/lib/hook/hook_wrapper.py
+++ b/samcli/lib/hook/hook_wrapper.py
@@ -191,7 +191,7 @@ def get_available_hook_packages_ids() -> List[str]:
     LOG.debug("Return available internal hook packages")
     hook_packages_ids = []
     for child in INTERNAL_PACKAGES_ROOT.iterdir():
-        if child.is_dir():
+        if child.is_dir() and child.name != "__pycache__":
             hook_packages_ids.append(child.name)
 
     return hook_packages_ids
--- a/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
+++ b/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
@@ -51,8 +51,6 @@ import subprocess
 import zipfile
 import logging
 
-from six import string_types
-
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


torresxb1:
new help text: 
Screen Shot 2022-10-27 at 3 43 55 PM

samcli/lib/hook/hook_wrapper.py
```diff
@@ -191,7 +191,7 @@ def get_available_hook_packages_ids() -> List[str]:
     LOG.debug("Return available internal hook packages")
     hook_packages_ids = []
     for child in INTERNAL_PACKAGES_ROOT.iterdir():
-        if child.is_dir():
+        if child.is_dir() and child.name != "__pycache__":
```
hawflau, line 194: is __pycache__ the only one we want to exclude?

torresxb1: right now, yes. If we wanted to make this less likely to break, we could just hardcode the package ids here? But we'd have to make sure we keep them up to date.

lucashuy: We could also change this to check in a list instead if we wanted to be set up for any future excluded folders, but presumably anything inside of that hook_packages folder is a hook_package right (aside from the cache folder)?

moelasmar: let's ignore any name that does not start with [a-z] .. these names will be the hooks packages names we choose, so we can assume that it will always start with small alphabet letters

hawflau: yeah, a list to exclude would be good

torresxb1: done

samcli/lib/hook/hook_wrapper.py
```diff
@@ -191,7 +191,7 @@ def get_available_hook_packages_ids() -> List[str]:
     LOG.debug("Return available internal hook packages")
     hook_packages_ids = []
     for child in INTERNAL_PACKAGES_ROOT.iterdir():
-        if child.is_dir():
+        if child.is_dir() and child.name[0].isalpha():
```
hawflau, line 194: uppercase letter would still pass this check?

maybe something like this?
```suggestion
string.ascii_lowercase

if child.is_dir() and child.name[0] in string.ascii_lowercase:
```

and just to certain, child.name[0] should never throw any error right? (as in is it possible to have an empty string as dir name?

hawflau (approved):

samcli/lib/hook/hook_wrapper.py
moelasmar: I think it is OK .. it does not really matter

moelasmar (approved):
it looks good .. thanks Rup :)

PR #4351, opened 2022-10-27
fix: Updated build success message for hook packages

lucashuy:
Changes the build success messages to not mention commands that won't work with current hook packages.

Checks for the hook package and changes which commands are mentioned in the message.

eg when using a hook package:
```
Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Invoke Function: sam local invoke
[*] Start a local API: sam local start-api --hook-package-id terraform
[*] Emulate local Lambda functions: sam local start-lambda --hook-package-id terraform
```

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -371,7 +371,6 @@ class BuildContext:
         invoke_suggestion = "Invoke Function: sam local invoke"
         sync_suggestion = "Test Function in the Cloud: sam sync --stack-name {{stack-name}} --watch"
         deploy_suggestion = "Deploy: sam deploy --guided"
-        start_api_suggestion = "Start a local API: sam local start-api"
         start_lambda_suggestion = "Emulate local Lambda functions: sam local start-lambda"
 
         if not is_default_build_dir:
@@ -383,10 +382,9 @@ class BuildContext:
         # check if we have used a hook package before building
         if self._hook_package_id:
             hook_package_flag = f" --hook-package-id {self._hook_package_id}"
-            start_api_suggestion += hook_package_flag
             start_lambda_suggestion += hook_package_flag
 
-            commands = [invoke_suggestion, start_api_suggestion, start_lambda_suggestion]
+            commands = [invoke_suggestion, start_lambda_suggestion]
 
         msg = f"""\nBuilt Artifacts  : {artifacts_dir}
 Built Template   : {output_template_path}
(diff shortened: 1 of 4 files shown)


samcli/commands/build/build_context.py
```diff
@@ -344,29 +348,55 @@ def _handle_build_post_processing(self, builder: ApplicationBuilder, build_resul
...
-            deploy_cmd += " --template-file {}".format(output_template_path)
+            invoke_suggestion += " -t {}".format(output_template_path)
+            deploy_suggestion += " --template-file {}".format(output_template_path)
+
+        commands = [validate_suggestion, invoke_suggestion, sync_suggestion, deploy_suggestion]
+
+        # check if we have used a hook package before building
+        if self._hook_package_id:
+            hook_package_flag = f" --hook-package-id {self._hook_package_id}"
+            start_api_suggestion += hook_package_flag
```
moelasmar, line 386: can we skip the local api testing as well. It is not supported in this beta release

moelasmar (changes requested):
Thanks Lucas, I juts left one small comment.

moelasmar (approved):
thanks Lucas

samcli/commands/build/build_context.py
```diff
@@ -344,29 +348,53 @@ def _handle_build_post_processing(self, builder: ApplicationBuilder, build_resul
...
-            deploy_cmd += " --template-file {}".format(output_template_path)
+            invoke_suggestion += " -t {}".format(output_template_path)
+            deploy_suggestion += " --template-file {}".format(output_template_path)
+
+        commands = [validate_suggestion, invoke_suggestion, sync_suggestion, deploy_suggestion]
+
+        # check if we have used a hook package before building
+        if self._hook_package_id:
+            hook_package_flag = f" --hook-package-id {self._hook_package_id}"
+            start_lambda_suggestion += hook_package_flag
```
hawflau, line 385: nit: it works without the hook package flag, but should we allow add it for the local invoke command?

moelasmar: it will work, but let's here at least printed with the `--hook-package-id ` ... as in case if the project does not need to be built, the customers have to provide it.

lucashuy: Nice catch, will add hook package flag to invoke

hawflau (approved):

PR #4353, opened 2022-10-28
feat: link sam metadata resource to lambda resource if there is no resource_name attribute

moelasmar:
This PR is to process the metadata resources that does not contain value for the resource name property. For these resources, SAM CLI will depend on the built_artifact property and try to find a lambda resource that use the same value as code property.

This PR also contains integration test cases that cover the Serverless TF modules usage.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -290,43 +308,76 @@ def _translate_to_cfn(tf_json: dict, output_directory_path: str, terraform_appli
...
 
             if resource_type == TF_AWS_LAMBDA_FUNCTION:
-                resolved_config_address = _get_configuration_address(resource_address)
+                resolved_config_address = _get_configuration_address(resource_full_address)
                 matched_lambdas = lambda_funcs_conf_cfn_resources.get(resolved_config_address, [])
                 matched_lambdas.append(translated_resource)
                 lambda_funcs_conf_cfn_resources[resolved_config_address] = matched_lambdas
                 lambda_config_funcs_conf_cfn_resources[resolved_config_address] = config_resource
 
+                resource_type = translated_properties.get("PackageType", "Zip")
```
sriram-mv, line 359: lets use the ZIP and IMAGE constants

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -346,6 +401,73 @@ def _translate_to_cfn(tf_json: dict, output_directory_path: str, terraform_appli
...
+    Parameters
+    ----------
+    sam_metadata_resource: SamMetadataResource
+        The mapped metadata resource
+    sam_metadata_resource_planned_values: Dict
+        The metadata resource in planned values section
+    sam_metadata_resources: List[SamMetadataResource]
+        The list of metadata resources
+    """
+    if sam_metadata_resource_planned_values.get("values", {}).get("triggers", {}).get("resource_name"):
```
sriram-mv, line 465: we may want to move to constants here as well, especially if we are going to be re-using.

sriram-mv (commented):
I'll go through another pass at this since its a large PR.

lucashuy (commented):
Left a minor comment change, will review the test cases later and provide additional comments if there are any

samcli/hook_packages/terraform/hooks/prepare/hook.py
moelasmar: done

samcli/hook_packages/terraform/hooks/prepare/hook.py
moelasmar: done

lucashuy (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -40,6 +44,18 @@
...
+
+SAM_METADATA_DOCKER_FILE_ATTRIBUTE = "docker_file"
+
+SAM_METADATA_RESOURCE_TYPE_ATTRIBUTE = "resource_type"
+
+SAM_METADATA_ADDRESS_ATTRIBUTE = "address"
+
+SAM_METADATA_RESOURCE_NAME_ATTRIBUTE = "resource_name"
+
 REMOTE_DUMMY_VALUE = "<<REMOTE DUMMY VALUE - RAISE ERROR IF IT IS STILL THERE>>"
```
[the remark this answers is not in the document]
moelasmar: I used this dummy value to set as a value for the lambda resources Code.S3 properties in the case if the S3 properties in Terraform are null, which means that it is referring to an S3 Bucket resource that will be created in the same terraform project. Also this means that we should find an s3_object that upload the lambda resource artifacts to this S3 Bucket. 

So, we set this value while we map the terraform Lambda resource into CFN format, and after we process the s3_objects terraform resources, we check for all lambda resources we mapped, and if there is any of them that still use the dummy value, this means that there is an issue in this TF project and we will raise an error

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -290,43 +324,76 @@ def _translate_to_cfn(tf_json: dict, output_directory_path: str, terraform_appli
...
                 matched_lambdas = lambda_funcs_conf_cfn_resources.get(resolved_config_address, [])
                 matched_lambdas.append(translated_resource)
                 lambda_funcs_conf_cfn_resources[resolved_config_address] = matched_lambdas
                 lambda_config_funcs_conf_cfn_resources[resolved_config_address] = config_resource
 
+                resource_type = translated_properties.get("PackageType", ZIP)
+                resource_type_constants = {"Zip": ("zip", "filename"), "Image": ("image", "image_uri")}
+                planned_value_function_code_path = (
+                    translated_properties.get("Code")
+                    if resource_type == "Zip"
```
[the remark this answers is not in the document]
moelasmar: resolved, and checked for similar usages

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -1523,14 +1719,23 @@ def _map_s3_sources_to_functions(s3_hash_to_source: Dict[str, str], cfn_resource
...
                 if source:
-                    tf_address = resource.get("Metadata", {}).get("SamResourceId")
-                    LOG.debug(
-                        "Found S3 object resource with matching bucket and key for function %s."
-                        " Setting function's Code property to the matching S3 object's source: %s",
-                        tf_address,
-                        source,
-                    )
-                    resource["Properties"][code_property] = source
+                    if source[0]:
```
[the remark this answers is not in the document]
moelasmar: no, this is a tuple of 2 values, that we add to the `s3_hash_to_source` dictionary. we map the hash value of the S3 bucket properties defined in the s3_object to a tuple of the planned value, and the configuration value of the code artifact path attached to this s3_object.

PR #4360, opened 2022-10-31
Extract full file name from Dockerfile path

### LINKED ISSUE #4355: Bug: sam build omits Dockerfile suffix from build.toml when aws:asset:property == "Code.ImageUri"
### Description:

Hi all! I was unable to find an open or closed issue describing this exact problem.

The problem is very specific: when your SAM template specifies a Dockerfile containing a suffix in its name (e.g., `.production`) AND the `aws:asset:property` is set to `"Code.ImageUri"`, the Dockerfile suffix is not copied over to the generated `build.toml`, and thus SAM is unable to locate the Dockerfile. This occurs for both the `aws:asset:dockerfile-path` and `Dockerfile` properties in the template JSON.

This is important because this is the pathway by which CDK generates SAM templates which can be used for local testing and other deployment internals.

The problem boils down to the metadata normalization when `aws:asset:property` is `Code.ImageUri`. This leads to `_extract_image_asset_metadata` being called in which only the Dockerfile path's `stem` is extracted ([code permalink](https://github.com/aws/aws-sam-cli/blob/d634991615476a3980496694f90ac6c23800ab73/samcli/lib/samlib/resource_metadata_normalizer.py#L189)).

It seems to me that `dockerfile_path.name` should be used here instead of `dockerfile_path.stem` ([Python docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name)). Is there some other rationale for using `.stem` here?

### Steps to reproduce:

1. Create a CDK project (e.g., [CDK hello world app](https://docs.aws.amazon.com/cdk/v2/guide/hello_world.html))
2. Add a simple Dockerfile **named `Dockerfile.production`**
3. Define a CDK Docker Lambda function. E.g.:

```typescript
new lambda.DockerImageFunction(
      this,
      "TestFunction",
      {
        code: lambda.DockerImageCode.fromImageAsset(
          path.join(__dirname, "..", "path", "to", "dockerfile"),
          {
            file: "Dockerfile.production",
          },
        ),
      },
    );
```

5. Change the name of your `Dockerfile` to `Dockerfile.prod`. Also update this reference in the template JSON/YAML.
6. `npx cdk synth --no-staging`
7. `sam build -t cdk.out/HelloCdk.template.json`
8. `sam local invoke`

### Observed result:

The build will "succeed" but you should not see a Docker build triggered. Open `.aws-sam/build.toml` and examine the function build definition metadata. You should see

```
"aws:asset:dockerfile-path" = "Dockerfile.production"
```

and

```
Dockerfile = "Dockerfile"
```

Upon attempting to invoke, you will see that SAM is unable to locate the Docker image to invoke the function (because it has not been built).

### Expected result:

The outputted `build.toml` should include:

```
Dockerfile = "Dockerfile.production"
```

`sam build` and `sam local invoke` should trigger a build of the Docker image and invoke the function by running that image, respectively.

1. macOS 13.0
3. `sam --version`: SAM CLI, version 1.61.0
9. AWS region: local development

christiangaetano:
Fixes #4355. Use `Path`/`PurePath`'s `name` property to extract a full Dockerfile file name. The current implementation uses `.stem` which omits the final suffix, is present.

Fixes #4355.

The `sam build` command does not properly locate Dockerfiles with extension(s) in their file names without this change.

By using the `.name` property on `Path`/`PurePath` as opposed to `.stem`, the entire file name is guaranteed to be included.

See https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name and https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem.

None. File names with no extension (e.g., `Dockerfile`) will be parsed in the same way after this change. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/samlib/resource_metadata_normalizer.py
+++ b/samcli/lib/samlib/resource_metadata_normalizer.py
@@ -186,7 +186,7 @@ class ResourceMetadataNormalizer:
         """
         asset_path = Path(metadata.get(ASSET_PATH_METADATA_KEY, ""))
         dockerfile_path = Path(metadata.get(ASSET_DOCKERFILE_PATH_KEY), "")
-        dockerfile, path_from_asset = dockerfile_path.stem, dockerfile_path.parent
+        dockerfile, path_from_asset = dockerfile_path.name, dockerfile_path.parent
         dockerfile_context = str(Path(asset_path.joinpath(path_from_asset)))
         return {
             SAM_METADATA_DOCKERFILE_KEY: dockerfile,
--- a/tests/integration/buildcmd/test_build_cmd.py
+++ b/tests/integration/buildcmd/test_build_cmd.py
@@ -86,6 +86,34 @@ class TestBuildCommand_PythonFunctions_Images(BuildIntegBase):
             self.built_template, self.FUNCTION_LOGICAL_ID_IMAGE, self._make_parameter_override_arg(overrides), expected
         )
 
+    @parameterized.expand([("3.6", False), ("3.7", False), ("3.8", False), ("3.9", False)])
+    @pytest.mark.flaky(reruns=3)
+    def test_with_dockerfile_extension(self, runtime, use_container):
+        _tag = f"{random.randint(1,100)}"
+        overrides = {
+            "Runtime": runtime,
+            "Handler": "main.handler",
+            "DockerFile": "Dockerfile.production",
+            "Tag": _tag,
+        }
+        cmdlist = self.get_command_list(use_container=use_container, parameter_overrides=overrides)
+
+        LOG.info("Running Command: ")
+        LOG.info(cmdlist)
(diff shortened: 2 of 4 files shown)


mndeveci (commented):
Thanks for quickly raising this PR! Can we add some unit/integration tests for this use case?

christiangaetano:
I added one unit test and one integration test. There are several integration tests that verify that images can be built and invoked locally, but I think it is sufficient to duplicate only the primary test for building/invoking. If you think I should duplicate the other `sam build` integration tests as well, let me know.

Thank you!

hawflau (approved):
LGTM. yep, I agree that the integration test cases added would be enough. Thanks for your contribution!

sriram-mv (approved):

Issue #4361, opened 2022-10-31
Bug: SAM-CLI 1.61.0 is breaking our sam delete with "Max attempts exceeded" error


hung-tang:
### Description:

The `sam delete` command has broken our pipeline with the release of 1.61.0.  I suspect it's due to this change [here](https://github.com/aws/aws-sam-cli/pull/4321).  After I rolled back to 1.60.0, our pipeline is working again.

We have Lambdas with VPC connectivity configured.  As documented [here](https://docs.aws.amazon.com/lambda/latest/dg/foundation-networking.html), this can take up to 20 minutes for Lambdas to clean up HENIs.  The latest 1.61.0 release looks like it will timeout prematurely which will return an error to our pipeline job.  Prior to 1.61.0, the `sam delete` command will just wait until stack is deleted (i.e. the Lambdas is able to delete the HENIs)

Please consider rolling back [4321](https://github.com/aws/aws-sam-cli/pull/4321).

### Steps to reproduce:

Make sure you have SAM CLI 1.61.0 installed
1. Create and deploy lambda with VPC connectivity using `sam deploy`
2. Then run `sam delete`

### Observed result:

After a minute, you will see the "Max attempts exceeded" error message:

```
2022-10-31 15:56:13,029 | Using config file: samconfig.toml, config environment: default
2022-10-31 15:56:13,029 | Expand command line arguments to:
2022-10-31 15:56:13,029 | --stack_name=htang-dev2 --no_prompts 
2022-10-31 15:56:13,120 | Config file location: /home/htang/samconfig.toml
2022-10-31 15:56:13,120 | Config file '/home/htang/samconfig.toml' does not exist
2022-10-31 15:56:13,539 | Input stack is deployed, continue deleting
2022-10-31 15:56:13,765 | Stack with id htang-dev2-d1ad6991-CompanionStack does not exist
2022-10-31 15:56:14,102 | Could not find the S3 file with the key 93c23669f53871dd23571c90926e562e
2022-10-31 15:56:14,103 | 	- Could not find and delete the S3 object with the key 93c23669f53871dd23571c90926e562e
2022-10-31 15:56:14,175 | Could not find the S3 file with the key 54250197b1767e869139074ec10db986
2022-10-31 15:56:14,175 | 	- Could not find and delete the S3 object with the key 54250197b1767e869139074ec10db986
[...]

PR #4362, opened 2022-11-01
fix: regresssion on `sam delete`

sriram-mv:
- Remove `MaxAttempts` and let it be set to default of 120 per https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#waiters

https://github.com/aws/aws-sam-cli/issues/4361

- Removes the total number of attempts before change of state to be the default. This stems from a wrong understanding of what `MaxAttempts` meant during the previous implementation of the fix.

- Unsets `MaxAttempts` and restores default behavior.

- Fixes regression https://github.com/aws/aws-sam-cli/issues/4361 and should not have any other side effects.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/delete/cfn_utils.py
+++ b/samcli/lib/delete/cfn_utils.py
@@ -116,8 +116,9 @@ class CfnUtils:
 
         # Wait for Delete to Finish
         waiter = self._client.get_waiter("stack_delete_complete")
-        # Poll every 30 seconds and set max attempts to be 3.
-        waiter_config = {"Delay": 30, "MaxAttempts": 3}
+        # Do not set `MaxAttempts` in waiter_config
+        # Regression: https://github.com/aws/aws-sam-cli/issues/4361
+        waiter_config = {"Delay": 30}
         try:
             waiter.wait(StackName=stack_name, WaiterConfig=waiter_config)
         except WaiterError as ex:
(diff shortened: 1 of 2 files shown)


samcli/lib/delete/cfn_utils.py
```diff
@@ -116,8 +116,10 @@ def wait_for_delete(self, stack_name):
 
         # Wait for Delete to Finish
         waiter = self._client.get_waiter("stack_delete_complete")
-        # Poll every 30 seconds and set max attempts to be 3.
-        waiter_config = {"Delay": 30, "MaxAttempts": 3}
+        # Poll every 30 seconds which is the default, but just explicitly called out.
+        # Do not set `MaxAttempts` in waiter_config - regression: https://github.com/aws/aws-sam-cli/issues/4361
+        # Waiter configuration docs at: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#waiters
+        waiter_config = {"Delay": 30}
```
super132, line 122: If we are using the default, why do we need to put in a definition here? If you were being explicit about `Delay`, should we also be explicit about the MaxAttempt (`120` by default) ?

sriram-mv: It was a point of education for ourselves, however I have instead removed those lines and just added a link to the regression issue.

mndeveci (approved):

hnnasit (approved):

samcli/lib/delete/cfn_utils.py
hnnasit: Looking at the commit history, it was initially set to 5 seconds and changed to 30 later in this [PR](https://github.com/aws/aws-sam-cli/pull/3232) .

PR #4368, opened 2022-11-01
Fix escape double quotes in generating Make rule

hawflau:

N/A

Double quotes in resource names are not properly escaped when used in generating make rules. This caused failures in terraform if the resource name contains double quotes, e.g. `null_resource.sam_metadata_aws_lambda_layer_version_layers["layer3"]`

Escape double quotes when generating make rules

No

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -1159,7 +1159,10 @@ def _build_jpath_string(sam_metadata_resource: SamMetadataResource, resource_add
     parent_modules = _get_parent_modules(module_address)
     for module in parent_modules:
         full_module_path += child_modules_template.format(module_address=module)
-    return jpath_string_template.format(child_modules=full_module_path, resource_address=resource_address)
+    jpath_string = jpath_string_template.format(child_modules=full_module_path, resource_address=resource_address)
+    # return jpath_string.replace("\"", "\\\"")
```
sriram-mv, line 1163: we could get rid of the comments.

sriram-mv (approved):

moelasmar (commented):
can we add an integration test case for the for_each usage .. we can add a new lambda function in our invocations tf sample project

lucashuy (approved):

moelasmar (approved):

PR #4369, opened 2022-11-01
feat: app insights monitoring integration for sam init

vsukameti:

New feature to add CLI option to enable application insights.

This allows customers to enable automated monitoring for their applications

No side effects

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/init/template_modifiers/cli_template_modifier.py
```diff
@@ -125,27 +126,10 @@ def _sanity_check(self) -> bool:
...
 
-        Returns
-        -------
-        list
-            array with updated template data
-        """
-        with open(self.template_location, "r") as file:
-            return file.readlines()
+    @abstractmethod
+    def _get_template(self) -> Any:
```
hawflau, line 133: We can remove the type hint here as it's an abstract method and doesn't really implement anything

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
```diff
@@ -22,7 +23,19 @@ class XRayTracingTemplateModifier(TemplateModifier):
         "https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst\n"
     )
 
-    def _add_new_field_to_template(self):
+    def _get_template(self) -> Any:
```
hawflau, line 26: should the type hint be `List[str]` instead?

hawflau (changes requested):

samcli/lib/telemetry/event.py
```diff
@@ -48,10 +48,7 @@ class EventType:
     _WORKFLOWS = [f"{config.language}-{config.dependency_manager}" for config in ALL_CONFIGS]
 
     _event_values = {  # Contains allowable values for Events
-        EventName.USED_FEATURE: [
-            "Accelerate",
-            "CDK",
-        ],
+        EventName.USED_FEATURE: ["Accelerate", "CDK", "ApplicationInsights"],
```
hawflau, line 51:
```suggestion
        EventName.USED_FEATURE: ["Accelerate", "CDK", "InitWithApplicationInsights"],
```

samcli/lib/init/__init__.py
```diff
@@ -121,9 +128,19 @@ def generate_project(
...
         template_modifier = XRayTracingTemplateModifier(template_file_path)
         template_modifier.modify_template()
+
+
+def _enable_application_insights(application_insights: bool, output_dir: str, name: str) -> None:
+    if application_insights:
+        template_file_path = f"{output_dir}/{name}/template.yaml"
+        template_modifier = ApplicationInsightsTemplateModifier(template_file_path)
+        template_modifier.modify_template()
+        EventTracker.track_event("UsedFeature", "ApplicationInsights")
```
hawflau, line 146:
```suggestion
        EventTracker.track_event("UsedFeature", "InitWithApplicationInsights")
```

hawflau (changes requested):

hawflau:
@vsukameti Please also run `make update-reproducible-reqs` to update "requirements/reproducible-linux.txt".

@mndeveci Do we need to update the attribution doc as we add ruamel yaml as dependency?

samcli/lib/init/template_modifiers/cli_template_modifier.py
vsukameti: The type is different of XRayTemplate and AppInsightsTemplate. Thats why I used Any

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
vsukameti: return type of yaml.load() is not List(str)

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,82 @@
...
+    AUTO_CONFIG_ENABLED_KEY = "AutoConfigurationEnabled"
+    DEPENDS_ON_KEY = "DependsOn"
+    RESOURCE_GROUP_TYPE = "AWS::ResourceGroups::Group"
+    CFN_STACK_TYPE = "CLOUDFORMATION_STACK_1_0"
+    APPLICATION_INSIGHTS_TYPE = "AWS::ApplicationInsights::Application"
+    RESOURCE_GROUP_REF = "ApplicationResourceGroup"
+    APPLICATION_INSIGHTS_REF = "ApplicationInsightsMonitoring"
+    AUTO_CONFIG_VALUE = "true"
+
+    def _get_template(self) -> Any:
```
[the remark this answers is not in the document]
vsukameti: The type is different of XRayTemplate and AppInsightsTemplate. Thats why I used Any

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,92 @@
...
+import logging
+from typing import Any
+import ruamel.yaml
+from ruamel.yaml import YAML
+from ruamel.yaml.comments import CommentedMap
+from samcli.lib.init.template_modifiers.cli_template_modifier import TemplateModifier
+
+LOG = logging.getLogger(__name__)
+yaml = YAML()
+# set ignore aliases to true
```
mndeveci, line 13: Can we add more details why we are setting this config?

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,92 @@
...
+                self.RESOURCE_GROUP_NAME_KEY: self.RESOURCE_GROUP_NAME,
+                self.AUTO_CONFIG_ENABLED_KEY: self.AUTO_CONFIG_VALUE,
+            },
+            self.DEPENDS_ON_KEY: self.RESOURCE_GROUP_REF,
+        }
+
+        self.template[self.RESOURCES_KEY][self.RESOURCE_GROUP_REF] = CommentedMap(resourceGroup)
+        self.template[self.RESOURCES_KEY][self.APPLICATION_INSIGHTS_REF] = CommentedMap(appInsightsApplication)
+        print("test")
+        print(self.template)
```
mndeveci, line 69: Let's remove testing prints with actual log statements

samcli/lib/telemetry/event.py
mndeveci: Should we move these events names into constants so that they can be used in different places? I saw we are duplicating the string here and somewhere above.

mndeveci (changes requested):

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: Sure, will do. 
FYI, this is to avoid automatic usage of YAML aliases. In this case, yaml aliases is applied to self.RESOURCE_GROUP_NAME which is used twice. 
I tried to reference resourceGroupName from Resource group logical ID. However, I keep getting this error "Requested attribute Name must be a readonly property in schema for AWS::ResourceGroups::Group" 
Looks like we cannot use reference for Name attribute

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: Sure, will do

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,82 @@
+"""
+Class used to parse and update template when application-insights is enabled
+"""
+import logging
+from typing import Any
+from ruamel.yaml import YAML
```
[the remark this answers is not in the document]
vsukameti: Sure, will do

samcli/lib/telemetry/event.py
vsukameti: I think making UsedFeature an Enum class makes more sense

requirements/reproducible-linux.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.7
+# This file is autogenerated by pip-compile with python 3.9
```
hawflau, line 2: we should use Python3.7 to update this file

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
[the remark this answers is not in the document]
hawflau: Can we use different types for `XRayTemplateModifier` and `ApplicationInsightsTemplateModifier`? I suppose they can have their own type hint, while the abstract method in the parent class doesn't require any type hint. Using `Any` doesn't provide any value for adding type hint here

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,94 @@
...
+                self.AUTO_CONFIG_ENABLED_KEY: self.AUTO_CONFIG_VALUE,
+            },
+            self.DEPENDS_ON_KEY: self.RESOURCE_GROUP_REF,
+        }
+
+        self.template[self.RESOURCES_KEY][self.RESOURCE_GROUP_REF] = \
+            ApplicationInsightsTemplateModifier.CommentedMap(resourceGroup)
+        self.template[self.RESOURCES_KEY][self.APPLICATION_INSIGHTS_REF] = \
+            ApplicationInsightsTemplateModifier.CommentedMap(appInsightsApplication)
+        LOG.info(self.template)
```
hawflau, line 74: Why do we need to log the template?

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
hawflau: This is under `XRayTracingTemplateModifier`, so it's not relevant to what the `_get_template` is returning under the `ApplicationInsightsTemplateModifier` class.

samcli/lib/telemetry/event.py
```diff
@@ -27,7 +27,11 @@ class EventName(Enum):
     SYNC_FLOW_END = "SyncFlowEnd"
     BUILD_WORKFLOW_USED = "BuildWorkflowUsed"
 
-
+class UsedFeature(Enum):
+    """Enum for the names of event values of UsedFeature"""
+    ACCELERATE = "Accelerate"
+    CDK = "CDK"
+    INIT_WITH_APPLICATION_INSIGHTS = "InitWithApplicationInsights"
```
hawflau, line 34: did linting fail for not having two empty lines between this line and the next?

hawflau (changes requested):

requirements/reproducible-linux.txt
```diff
@@ -21,24 +21,6 @@ aws-sam-translator==1.53.0 \
     --hash=sha256:84d780ad82f1a176e2f5d4c397749d1e71214cc97ee7cccd50f823fd7c7e7cdf \
     --hash=sha256:85252646cf123642d08442137b60445e69e30bfd2f8b663b1202b20ab3782b10
     # via aws-sam-cli (setup.py)
-backports-zoneinfo==0.2.1 \
```
mndeveci, line 24: As Wing mentioned above, these file should be generated with python3.7, otherwise we will miss some dependencies due to version difference.

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,94 @@
...
+Class used to parse and update template when application-insights is enabled
+"""
+import logging
+from typing import Any
+from samcli.lib.init.template_modifiers.cli_template_modifier import TemplateModifier
+
+LOG = logging.getLogger(__name__)
+
+
+class ApplicationInsightsTemplateModifier(TemplateModifier):
```
mndeveci, line 11: Can we add some description (comment) what is this class supposed to do.

tests/integration/init/test_init_command.py
```diff
@@ -371,6 +371,64 @@ def test_init_command_passes_with_disabled_tracing(self):
...
+                ]
+            )
+            try:
+                process.communicate(timeout=TIMEOUT)
+            except TimeoutExpired:
+                process.kill()
+                raise
+
+            self.assertEqual(process.returncode, 0)
+            self.assertTrue(Path(temp, "sam-app").is_dir())
```
mndeveci, line 401: Can we check if resources are added into template? And other way around with the test below?

tests/unit/lib/init/test_cli_template_modifier.py
```diff
@@ -328,3 +351,183 @@ def test_must_pass_sanity_check(self, parse_yaml_file_mock):
         parse_yaml_file_mock.return_value = {"add: add_value"}
         result = template_modifier._sanity_check()
         self.assertTrue(result)
+
+    @patch(
+        "samcli.lib.init.template_modifiers.application_insights_template_modifier.ApplicationInsightsTemplateModifier._get_template"
+    )
+    def test_must_add_application_insights_monitoring(self, get_template_patch):
```
mndeveci, line 358: This class (or file) already has too many lines. Since these test will cover a new implementation, can we move them into their own file?

mndeveci (changes requested):

tests/integration/init/test_init_command.py
vsukameti: we test this part in test_cli_template_modifier.py file

hawflau:
Please run `make black` to address the format issue caught [here](https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli/builds/45272814)

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: Sure, will do

requirements/reproducible-linux.txt
vsukameti: Sure, will change it

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: Just for debugging. No specific reason

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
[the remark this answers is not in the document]
vsukameti: The return type of yaml.load is itself Any

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
vsukameti: Sure, will change this

tests/unit/lib/init/test_cli_template_modifier.py
vsukameti: Sure, will change that

samcli/lib/init/__init__.py
```diff
@@ -121,9 +129,19 @@ def generate_project(
...
+
 
 def _apply_tracing(tracing: bool, output_dir: str, name: str) -> None:
     if tracing:
         template_file_path = f"{output_dir}/{name}/template.yaml"
         template_modifier = XRayTracingTemplateModifier(template_file_path)
         template_modifier.modify_template()
+
+
+def _enable_application_insights(application_insights: bool, output_dir: str, name: str) -> None:
```
hoffa, line 142: So if I understand correctly, this enables Application Insights by changing the input template. This means that unless someone is using SAM CLI (e.g. deploying directly with `aws cloudformation deploy`), it won't include the extra resources and might cause them to get deleted.

Should this be done in the [SAM transform](https://github.com/aws/serverless-application-model/) instead?

hoffa: Ah on second glance I might've misunderstood. I assume this only runs once during `sam init` and is stored in the final template?

If that's the case, this could potentially also live as part of the [`init` templates](https://github.com/aws/aws-sam-cli-app-templates). 🤔

mndeveci:
> @mndeveci Do we need to update the attribution doc as we add ruamel yaml as dependency?
I've missed this one earlier, yeah we need to update attribution doc with this new addition (and any possible dependency that comes with this library).

samcli/lib/init/__init__.py
hawflau: > I assume this only runs once during sam init and is stored in the final template?

Instead of the final template, the snippet (App Insights resources) is added into the working template. 

This change is similar to the [X-Ray Tracing change](https://github.com/aws/aws-sam-cli/pull/3817). We considered options like the current choice (adding an option in the init flow to add or not add the resources) and adding the snippet in init templates. We decided to go with the current approach because it's not flexible to allow enable/no-enable in the init templates repo. Also, if we made the change in init templates instead, we would have to modify each template to have one version with the added resources and one version without.

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
hawflau: Can we remove it? using `LOG.info` will print the template to customers' console when they run the command.

hawflau (changes requested):

requirements/reproducible-linux.txt
```diff
@@ -191,9 +191,9 @@ idna==2.10 \
     --hash=sha256:b307872f855b18632ce0c21c5e45be78c0ea7ae4c15c828c20788b26921eb3f6 \
     --hash=sha256:b97d804b1e9b523befed77c48dacec60e6dcb0b5391d57af6a65a312a90648c0
     # via requests
-importlib-metadata==4.11.3 \
-    --hash=sha256:1208431ca90a8cca1a6b8af391bb53c1a2db74e5d1cef6ddced95d4b2062edc6 \
-    --hash=sha256:ea4c597ebf37142f827b8f39299579e31685c31d3a438b59f469406afd0f2539
+importlib-metadata==5.0.0 \
```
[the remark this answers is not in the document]
mndeveci: And can we also re-run `make update-reproducible-reqs` to update this file with up to date dependencies?

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
mndeveci: @vsukameti were you able to push your changes to address this comment?

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,95 @@
...
+
+    TYPE_KEY = "Type"
+    RESOURCES_KEY = "Resources"
+    PROPERTIES_KEY = "Properties"
+    NAME_KEY = "Name"
+    RESOURCE_QUERY_KEY = "ResourceQuery"
+    RESOURCE_GROUP_NAME_KEY = "ResourceGroupName"
+    AUTO_CONFIG_ENABLED_KEY = "AutoConfigurationEnabled"
+    DEPENDS_ON_KEY = "DependsOn"
+    RESOURCE_GROUP_TYPE = "AWS::ResourceGroups::Group"
```
mndeveci, line 24: We usually keep resource definition types in this file, can we move this one and `AWS::ApplicationInsights::Application` there as well?

https://github.com/aws/aws-sam-cli/blob/d0f99f2f0160e53f04f08949ac3f779eb6f61f79/samcli/lib/utils/resources.py

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,95 @@
...
+    CFN_STACK_TYPE = "CLOUDFORMATION_STACK_1_0"
+    APPLICATION_INSIGHTS_TYPE = "AWS::ApplicationInsights::Application"
+    RESOURCE_GROUP_REF = "ApplicationResourceGroup"
+    APPLICATION_INSIGHTS_REF = "ApplicationInsightsMonitoring"
+    AUTO_CONFIG_VALUE = "true"
+    RESOURCE_GROUP_NAME = {"Fn::Join": ["", ["ApplicationInsights-SAM-", {"Ref": "AWS::StackName"}]]}
```
mndeveci, line 30: Will this create any issues if resource group name has size limitations, if how can we eliminate it?

tests/unit/lib/init/test_cli_template_modifier.py
mndeveci: @vsukameti same as previous, were you able to push your changes to address this comment?

mndeveci (changes requested):

requirements/reproducible-linux.txt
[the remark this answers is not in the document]
vsukameti: Re-ran `make update-reproducible-reqs`. No new dependencies were added to the file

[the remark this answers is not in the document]
vsukameti: @hawflau not sure why that was upgraded. I committed the diff after running `make update-reproducible-reqs`

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: Yes, I added a one-line description

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: A stack name cannot be longer than 128 chars. Size limit of ResourceGroup name is 300 chars. So, that shouldn't be a problem

tests/unit/lib/init/test_cli_template_modifier.py
vsukameti: Yes, moved tests to a new file, "test_application_insights_template_modifier"

samcli/commands/init/command.py
```diff
@@ -240,6 +240,11 @@ def wrapped(*args, **kwargs):
     default=None,
     help="Enable AWS X-Ray tracing for your lambda functions",
 )
+@click.option(
+    "--application-insights/--no-application-insights",
+    default=None,
+    help="Enable CloudWatch Application Insights monitoring for your application",
```
sriram-mv, line 246: "Enable AWS CloudWatch Application Insights". has the help text been bar raised with UX? Its a crucial component on how customers understand and interpret.

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
```diff
@@ -0,0 +1,97 @@
...
+            return self.yaml.load(file)
+
+    def _update_template_fields(self):
+        """
+        Add new resources to SAM template
+        """
+        self._add_app_insights_monitoring_section()
+
+    def _add_app_insights_monitoring_section(self):
+        resourceGroup = {
```
sriram-mv, line 55: should this be versioned? how will this change over time if it does? will there be new cfn properties?

sriram-mv (changes requested):

mndeveci (approved):

mndeveci:
As I synced offline, can you run `make black` so that it can fix the formatting issue which is blocking CI now?
```
would reformat samcli/lib/init/template_modifiers/application_insights_template_modifier.py
Oh no! 💥 💔 💥
1 file would be reformatted, 747 files would be left unchanged.
```

samcli/commands/init/command.py
vsukameti: Yes, it was discussed and approved by UX

samcli/lib/init/template_modifiers/application_insights_template_modifier.py
vsukameti: I don't think it needs to be versioned

sriram-mv: As discussed offline, lets continue the conversation on keeping this section upto date. Since we are injecting new resources from client side tooling.

sriram-mv (approved):

hawflau (approved):

qingchm:
@vsukameti thanks for the contribution! Can you please pull for the latest changes and resolve the conflicts listed above? I myself usually run git pull upstream develop --no-rebase (with this remote named as upstream). We updated the third party license file and the requirements file so please take a look and address the conflict! Thanks!

PR #4377, opened 2022-11-02
feat: Added --skip-prepare-infra to skip the prepare hook stage

lucashuy:
This change adds the `--skip-prepare-infra` flag to `sam build`, `invoke` and `start-lambda`. This flag will skip the infrastructure setup stage if there are metadata files.

Adds a new click option to skip the prepare hook if we already have a metadata file.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/unit/commands/_utils/custom_options/test_hook_package_id_option.py
```diff
@@ -13,6 +13,12 @@ def setUp(self):
         self.name = "hook-package-id"
         self.opt = "--hook-package-id"
         self.terraform = "terraform"
+        self.invalid_coexist_options = ["t", "template", "template-file", "parameters-override"]
+        self.metadata_path = "path/metadata.json"
+        self.cwd_path = "path/current"
+
+        self.iac_hook_wrapper_instance_mock = MagicMock()
+        self.iac_hook_wrapper_instance_mock.prepare.return_value = self.metadata_path
```
lucashuy, line 21: This file has a bunch of changed lines. The bulk of the changes is to move a wrapper mock and some setup to the `setUp` class to reduce repeat code.

samcli/commands/_utils/custom_options/hook_package_id_option.py
```diff
@@ -74,15 +74,26 @@ def handle_parse_result(self, ctx, opts, args):
                 LOG.debug("beta-feature flag is disabled and prepare hook is not run")
                 return super().handle_parse_result(ctx, opts, args)
 
+            iac_project_path = os.getcwd()
+            output_dir_path = os.path.join(iac_project_path, ".aws-sam-iacs", "iacs_metadata")
+
+            # check if user provided --skip-prepare-iac
+            skip_prepare_hook = opts.get("skip_prepare_iac")
+            metadata_file_path = os.path.join(output_dir_path, "template.json")
```
lucashuy, line 82: question: Is it safe to assume that this template file will always be generated in this location (`.aws-sam-iacs/iacs_metadata`?

tests/unit/commands/local/invoke/test_cli.py
```diff
@@ -46,22 +46,13 @@ def setUp(self):
...
-        ctx_mock.profile = self.profile
-
-        # Mock the __enter__ method to return a object inside a context manager
-        context_mock = Mock()
-        InvokeContextMock.return_value.__enter__.return_value = context_mock
+        self.ctx_mock = Mock()
+        self.ctx_mock.region = self.region_name
+        self.ctx_mock.profile = self.profile
 
+    def call_cli(self):
```
lucashuy, line 53: I originally modified this file by including a parameter for `skip_prepare_iac`. While I was in this file, I moved the CLI invoke call to a helper function to reduce repeat code. I then realised that I didn't need to add `skip_prepare_iac`, but I had already made the help function improvements. That is why this file is modified.

samcli/commands/_utils/custom_options/hook_package_id_option.py
moelasmar: This will work for terraform case which is OK, but I think we should push it to the Prepare Hook, and the hook itself should do this check. The idea later we may have multiple templates to be generated by the hook (like CDK for example) ... also, the template file name can be different based on how the hook will name it (also CDK)

lucashuy: Sounds good, in that case I'll move this check into the prepare hook under the terraform folder.

samcli/commands/_utils/options.py
```diff
@@ -666,6 +683,24 @@ def hook_package_id_click_option(force_prepare=True, invalid_coexist_options=Non
...
+def skip_prepare_iac_click_option():
+    """
+    Click option to skip the hook preparation stage
+    """
+    return click.option(
+        "--skip-prepare-iac",
+        is_flag=True,
+        required=False,
+        callback=skip_prepare_iac_callback,
+        help="Skips the preparation stage for the hook package if the metadata file has already been generated. "
```
sriram-mv, line 695: Reading this, makes me question what this magic metadata file is. Do we explain this anywhere in our help texts. if not, we should provide context here.

samcli/commands/_utils/custom_options/hook_package_id_option.py
```diff
@@ -74,15 +74,26 @@ def handle_parse_result(self, ctx, opts, args):
...
             # call prepare hook
             built_template_path = DEFAULT_BUILT_TEMPLATE_PATH
             if not self._force_prepare and os.path.exists(built_template_path):
                 LOG.info("Skip Running Prepare hook. The current application is already prepared.")
+            elif skip_prepare_hook:
+                if os.path.exists(metadata_file_path):
+                    LOG.info("Skipping prepare hook, --skip-prepare-iac was provided and metadata file already exists.")
+                    opts["template_file"] = metadata_file_path
+                else:
+                    raise click.UsageError("No metadata file found, rerun without the --skip-prepare-iac flag.")
```
sriram-mv, line 93: This needs to be more expressive, what is a metadata file, how do customers know how to generate it.

On the error message itself, should we instead have the command to be run?

eg: Try: ` sam build --skip-prepare-iac`

samcli/commands/_utils/options.py
moelasmar: I think we need to rephrase in in a way to let the customers know that they can set this flag if they did not change their infrastructure since last time they use sam commands. The customers actually need to regenerate the metadata file after the change the infrastructure  code

samcli/commands/_utils/options.py
```diff
@@ -200,6 +200,23 @@ def resolve_s3_callback(ctx, param, provided_value, artifact, exc_set, exc_not_s
...
+    Parameters
+    ----------
+    ctx: click.core.Context
+        Click context
+    param: click.Option
+        Parameter properties
+    provided_value: bool
+        True if option was provided
+    """
+    if provided_value and not ctx.params.get("hook_package_id", None):
```
moelasmar, line 216: also check in `ctx.default_map` .. so we also cover the case if the customer added this option to samconfig toml file

samcli/commands/build/command.py
```diff
@@ -84,6 +85,7 @@
     force_prepare=True,
     invalid_coexist_options=["t", "template-file", "template", "parameter-overrides"],
 )
+@skip_prepare_iac_option
```
moelasmar, line 88: move it to be after the `@configuration_option`

samcli/commands/local/invoke/cli.py
```diff
@@ -39,6 +39,7 @@
 @hook_package_id_click_option(
     force_prepare=False, invalid_coexist_options=["t", "template-file", "template", "parameter-overrides"]
 )
+@skip_prepare_iac_option
```
moelasmar, line 42: move it to be after the `@configuration_option`

samcli/commands/local/start_lambda/cli.py
```diff
@@ -63,6 +63,7 @@
 @hook_package_id_click_option(
     force_prepare=False, invalid_coexist_options=["t", "template-file", "template", "parameter-overrides"]
 )
+@skip_prepare_iac_option
```
moelasmar, line 66: move it to be after the `@configuration_option`

samcli/commands/_utils/options.py
lucashuy: New help message is: `This option should be used to skip the preparation stage if there have not been any infrastructure changes. The --hook-name option should also be specified when using this option.`. This is pending UX comments, but this is what is currently proposed.

samcli/commands/_utils/custom_options/hook_package_id_option.py
lucashuy: New behaviour is as follows:
- if option is provided and metadata file exists: console log a message saying we skipped prepare since we found that file at `/path/here`
- if option if not provided, or if option is provided and metadata file does not exist: continue with execution without printing any message

Sample output when using `--skip-prepare-infra`:
```
Running Prepare Hook to prepare the current application
Executing prepare hook of hook package "terraform"
Skipping preparation stage, the metadata file already exists at /Users/lucashuy/repos/Aws-sam-cli-terraform-sample-projects/working_examples/needs_build_nodejs/.aws-sam-iacs/iacs_metadata/template.json
Prepare hook completed and metadata file generated at: /Users/lucashuy/repos/Aws-sam-cli-terraform-sample-projects/working_examples/needs_build_nodejs/.aws-sam-iacs/iacs_metadata/template.json
Building layer 'aws_lambda_layer_version.hello_world_layer'
```

samcli/commands/_utils/options.py
```diff
@@ -693,6 +713,25 @@ def decorator(f):
...
+def skip_prepare_infra_click_option():
+    """
+    Click option to skip the hook preparation stage
+    """
+    return click.option(
+        "--skip-prepare-infra",
+        is_flag=True,
+        required=False,
+        callback=skip_prepare_infra_callback,
+        help="This option should be used to skip the preparation stage if there "
```
sriram-mv, line 725: Subject to change.

samcli/commands/_utils/options.py
```diff
@@ -200,6 +200,26 @@ def resolve_s3_callback(ctx, param, provided_value, artifact, exc_set, exc_not_s
...
+    param: click.Option
+        Parameter properties
+    provided_value: bool
+        True if option was provided
+    """
+    is_option_provided = provided_value or ctx.default_map.get("skip_prepare_infra")
+    is_hook_provided = ctx.params.get("hook_package_id") or ctx.default_map.get("hook_package_id")
+
+    if is_option_provided and not is_hook_provided:
+        raise click.BadOptionUsage(option_name=param.name, ctx=ctx, message="Missing option --hook-package-id")
```
sriram-mv, line 220: `--hook-package-id` to `--hook-name`

sriram-mv (approved):

moelasmar (approved):

PR #4378, opened 2022-11-03
feat: add hook metadata to generated template and use for telemetry

torresxb1:
Currently sam local invoke works for a terraform project although we do not provide --hook-package-id. This is not an issue in itself, but the way the telemetry is set up, it would be considered a CFN project (in telemetry data). 

This adds the hook package id to a metadata section of the generated template. This is then used by telemetry if no package id is provided from the command option. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/hook/utils.py
```diff
@@ -0,0 +1,28 @@
+"""
+Hook-related utils
+"""
+
+from typing import Dict, Optional, Any
+
+HOOK_METADATA_KEY = "AWS::SAM::Hook"
+
+
+def get_hook_metadata(template: Dict) -> Optional[Any]:
```
torresxb1, line 10: the type hint of the return was supposed to be `Optional[Dict]` but mypy complained because technically it's `Any` since we're just getting the value of a dict at some key.

samcli/lib/telemetry/metric.py
```diff
@@ -199,8 +200,11 @@ def wrapped(*args, **kwargs):
 
 def _get_project_details(hook_package_id: str, template_dict: Dict) -> ProjectDetails:
     if not hook_package_id:
-        project_type = ProjectTypes.CDK.value if is_cdk_project(template_dict) else ProjectTypes.CFN.value
-        return ProjectDetails(project_type=project_type, hook_package_id=None, hook_package_version=None)
+        hook_metadata = get_hook_metadata(template_dict)
+        if not hook_metadata:
+            project_type = ProjectTypes.CDK.value if is_cdk_project(template_dict) else ProjectTypes.CFN.value
+            return ProjectDetails(project_type=project_type, hook_package_id=None, hook_package_version=None)
+        hook_package_id = str(hook_metadata.get("HookPackageId"))
```
torresxb1, line 207: only casting because mypy complained that `hook_package_id` was `str` but `hook_metadata.get("HookPackageId")` is `Any`. 

What should really happen if HookPackageId is None or not a string? Asking because it seems like even when we have an exception, we still return some project details (see a couple lines below, on the except).

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -141,6 +141,11 @@ def prepare(params: dict) -> dict:
         if cfn_dict.get("Resources"):
             _update_resources_paths(cfn_dict.get("Resources"), terraform_application_dir)  # type: ignore
 
+        # Add hook metadata
+        cfn_dict["Metadata"]["AWS::SAM::Hook"] = {
```
moelasmar, line 145: could you please define constant variables for these values

samcli/lib/hook/utils.py
moelasmar: I think if you defined the `template` parameter as `Dict[str, Dict]` will solve it

moelasmar (approved):
approve with minor comment. Thanks Rup

samcli/lib/telemetry/metric.py
sriram-mv: https://github.com/aws/aws-sam-cli/pull/4378/files#diff-848f38e3508120c82700f37e74bbd8688f7cbfb99dec7b44725771222f076fa6R212 

why are project_type and hook_package_id the same thing? I find this function signature weird.

samcli/hook_packages/terraform/hooks/prepare/hook.py
sriram-mv: AWS::SAM::Hook makes it appear as if its a CFN deployable resource, is this intentional?

moelasmar: we just want to come with a name to be similar to the one we used for SAR Application metatdata https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-template-publishing-applications-metadata-properties.html .. but we can change it for sure

samcli/lib/telemetry/metric.py
moelasmar: I am not sure if I correctly understood your concern, but `hook_package_id` and `project_type` can have different values. For example, once we start allowing others to add their own hooks, someone can add another Terraform hook, that can be named anything but the `project_type` will be always Terraform ...

torresxb1: @sriram-mv  if you mean a few lines below when we catch InvalidHookPackageConfigException, I think it's because at that point it failed to read the config so we don't know the project type. So I'm guessing the next best thing is to use the package id? I don't even know if that's the behaviour we'd want but that's what I assumed it did right now.

samcli/hook_packages/terraform/hooks/prepare/hook.py
torresxb1: >could you please define constant variables for these values

good call, done

samcli/lib/hook/utils.py
torresxb1: that worked, thank you!

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -77,6 +77,11 @@
 TERRAFORM_BUILD_SCRIPT = "copy_terraform_built_artifacts.py"
 TF_BACKEND_OVERRIDE_FILENAME = "z_samcli_backend_override"
 
+HOOK_METADATA_KEY = "AWS::SAM::Hook"
+TERRAFORM_HOOK_METADATA = {
+    "HookPackageId": "terraform",
```
hawflau, line 82: Should we update this as we are changing from `--hook-package-id` to `--hook-name`? (https://github.com/aws/aws-sam-cli/pull/4382/files)

```suggestion
    "HookName": "terraform",
```

samcli/lib/telemetry/metric.py
mildaniel: > So I'm guessing the next best thing is to use the package id

That's right. In theory it shouldn't ever enter that state, but if it does, then we can at least emit some metric instead of nothing.

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -77,6 +77,11 @@
 TERRAFORM_BUILD_SCRIPT = "copy_terraform_built_artifacts.py"
 TF_BACKEND_OVERRIDE_FILENAME = "z_samcli_backend_override"
 
+HOOK_METADATA_KEY = "AWS::SAM::Hook"
```
mildaniel, line 80: Do we need this constant defined here and in the utils fle?

samcli/hook_packages/terraform/hooks/prepare/hook.py
moelasmar: we did not change the collected telemetry .. to do that we need to update our telemetry service .. I thought it is Ok to leave it as it is as it is something internal and does not affect our customers

samcli/hook_packages/terraform/hooks/prepare/hook.py
torresxb1: Well, my thinking is that if/when this was actually a hook package, it wouldn't have access to stuff in the sam cli repo/code. So I thought to code it as if it was a hook package. Not a big deal for me if we want to refer to the one in the utils or vice versa, however.

lucashuy (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
torresxb1: we discussed this outside here, but yes we can do this change independently of the telemetry one (that would be a separate change, if we decided to do that, but we can also keep that internally as Mohamed mentions). 

I just pushed the change to rename

PR #4379, opened 2022-11-03
feat: Run subprocess with loading pattern

mildaniel:
When running long external subprocesses, sam-cli can appear to be "stuck" when in reality it's just waiting for the process.

Runs the long subprocesses asynchronously and prints a timed pattern to the screen while waiting.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -112,20 +113,30 @@ def prepare(params: dict) -> dict:
...
+                # Terraform project expects input, and customer does not provide it.
+                command_args={
+                    "args": ["terraform", "plan", "-out", temp_file.name, "-input=false"],
+                    "cwd": terraform_application_dir,
+                    "stdout": PIPE,
+                }
             )
+            if isinstance(process_output, bytes):
+                process_output = process_output.decode("utf-8")
+            LOG.debug(process_output)
```
sriram-mv, line 138: won't this LOG all the debug at once? can we stream it instead?

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -112,20 +113,30 @@ def prepare(params: dict) -> dict:
...
-
         with osutils.tempfile_platform_independent() as temp_file:
-            run(
-                # input false to avoid SAM CLI to stuck in case if the Terraform project expects input, and customer
-                # does not provide it.
-                ["terraform", "plan", "-out", temp_file.name, "-input=false"],
-                check=True,
-                capture_output=True,
-                cwd=terraform_application_dir,
+            process_output = invoke_subprocess_with_loading_pattern(
```
sriram-mv, line 127: are we loading the `.` to stderr by default? There are CI runners that will error out that moment they see something printed to stderr, we have to be conscious of that.  

What is our current pattern for the dots, which stream do they land on ?

sriram-mv (changes requested):

samcli/hook_packages/terraform/hooks/prepare/hook.py
mildaniel: I was looking at other places in the codebase where we do something similar and we typically default to `stderr`. For example, we write loading dots when building Docker images https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/lambda_image.py#L236.  We can default it to `stdout` instead here if it makes more sense.

sriram-mv: I see, maybe its less invasive that I thought it to be, still something to be conscious about.

moelasmar: we use the stdout to write the `local invoke` output. I think to make it easier for people to read the lambda output programatically

samcli/hook_packages/terraform/hooks/prepare/hook.py
moelasmar: I think in this case (debug is enabled)  no idea of having the dots .

samcli/hook_packages/terraform/hooks/prepare/hook.py
sriram-mv: Fair, good callout.

samcli/hook_packages/terraform/hooks/prepare/hook.py
mildaniel: So you're saying we should stream one or the other? Is there a way of knowing what log level they have configured?

samcli/lib/utils/subprocess_utils.py
```diff
@@ -0,0 +1,115 @@
...
+        A function generating a pattern to the given stream
+    stream_writer: Optional[StreamWriter]
+        The stream to which to write the pattern
+
+    Returns
+    -------
+    str
+        A string containing the process output
+    """
+    stream_writer = stream_writer or StreamWriter(sys.stdout)
```
moelasmar, line 66: should it be `sys.stderr`, to leave the stdout for the lambda function output?

samcli/lib/utils/subprocess_utils.py
```diff
@@ -0,0 +1,115 @@
...
+            elif process.stdout:
+                # Logging level is DEBUG, streaming logs instead
+                for line in process.stdout:
+                    decoded_line = _check_and_process_bytes(line)
+                    LOG.debug(decoded_line)
+                    process_output += decoded_line
+
+            return_code = process.wait()
+
+            stream_writer.write("\n")
```
moelasmar, line 84: let's use `os.linesep` instead of `\n`

samcli/lib/utils/subprocess_utils.py
mildaniel: Yup, good catch, didn't mean to commit that.

moelasmar (approved):

sriram-mv (approved):

Issue #4375, opened 2022-11-02
Packaging: aws-sam-cli files conflict with the python "installer" package


javmorin:
I maintain a [package](https://aur.archlinux.org/packages/aws-sam-cli) for `aws-sam-cli` in Arch Linux. One of the users of the package notified me that it conflicted with the `python-installer` package (See refs [#1](https://pypi.org/project/installer/), [#2](https://archlinux.org/packages/community/any/python-installer/)) shipped by ArchLinux. I have confirmed this is the case, that aws-sam-cli does "ship" several files in an installer directory which when installed via pypi or setup.py could conflict with this other package.

I did some review of the source code, but I cannot see any direct calls to the `installer` files within `aws-sam-cli`. I am wondering if the `installer` files are necessary post-install, or if I can remove them as part of the packaging process.

Would appreciate clarification from the devs on this matter.

hawflau:
Hey @javmorin, thanks for raising the issue.

>  I am wondering if the installer files are necessary post-install, or if I can remove them as part of the packaging process.
They are not. You can remove them as part of the packaging process

However, I'm curious why it only happened recently. It looks like the `installer` library has existed for a while.

javmorin:
> They are not. You can remove them as part of the packaging process
Good to know, thanks! Barring other feedback, I will look to do this.

> However, I'm curious why it only happened recently. It looks like the `installer` library has existed for a while.
I think it was only noticed recently as I don't think many people install the `python-installer` package (that's a guess, I have no visibility to install #'s). Those files have been present in the repo for some time, but on my personal machine I didn't have the conflicting page installed (either directly or indirectly). 

I do think in the case of a system install via PyPi, this could lead to silent corruption as files are placed outside of the samcli directory, possibly overwriting existing files. I wonder if it's possible to alter the setup/install mechanism of `aws-sam-cli` to not install these files?

hawflau:
I checked with some team members. It seems like we can exclude "installers" in setup.py on this line:
https://github.com/aws/aws-sam-cli/blob/develop/setup.py#L54

But in general, we advise installing SAM CLI in an isolated environment (e.g. in a virtual environment), pretty much like how it's installed through our installer binary. This reduces the chance of having conflicting dependencies. Is there any chance you can explore in updating your package to install AWS SAM CLI in an isolated env?

javmorin:
[...]

PR #4389, opened 2022-11-08
exclude installer dir in build

hawflau:

#4375 

SAM CLI installation copies the unnecessary "installer" directory, which might conflict with the Python "installer" library

Exclude "installer"

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,5 +5,6 @@ include requirements/pre-dev.txt
 include requirements/dev.txt
 recursive-include samcli *
 prune tests
+prune installer
 global-exclude *.py[cod]
 global-exclude *.DS_Store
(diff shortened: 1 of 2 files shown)


jfuss (approved):
I wonder if we should write a simple tests that ensures the `.whl` doesn't include directories we don't want.

My thinking here is if we write a tests that generates the `.whl` and validates the output only includes `samcli/*` and specific top level files like License, setup, etc. Then when we add more things to the repo, we can ensure we aren't accidentally including new files. I am not blocking on this but something I think we should consider adding (or event adding it to a backlog/Github Issues).

sriram-mv (approved):

Issue #1921, opened 2020-04-14
sam local start-api doesn't update on changes


fakekamrankhan:
### Description

I am using the sample **hello-world** application on nodejs12.x using `sam init` but when I invoke `sam build` and `sam local start-api` any subsequent changes I make to the lambda function don't update on the port.

### Steps to reproduce

```
sam init
Choice: 1 (AWS Quick Start Templates)
Runtime: 1 (nodejs12.x)
cd sam-app
sam build && sam local start-api
```

Inside the `hello-world/app.js` response object I change the message property and save.

### Observed result

I refresh 127.0.0.1:3000/hello on the browser and nothing changes.

### Expected result

I should see the message updated as per how `sam local start-api` is meant to work.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

`sam --version`
SAM CLI, version 0.47.0

`docker --version`
Docker version 19.03.8, build afacb8b

`aws --version`
aws-cli/1.11.80 Python/2.7.10 Darwin/18.7.0 botocore/1.5.43

jfuss:
@fakekamrankhan Did you rebuild after changes? You need to run `sam build` again for code changes to be reflected.

fakekamrankhan:
Hmmm, I was under the impression that I could work on an existing unction without needing to rebuild:

> You can now browse to the above endpoints to invoke your functions. You do not need to restart/reload SAM CLI while working on your functions, changes will be reflected instantly/automatically. You only need to restart SAM CLI if you update your AWS SAM template

fakekamrankhan:
Apologies, I have dug a little further on this github repository and realize this isn't possible for now. Kind of a bummer because it is quite annoying as a development experience to have to constantly build after every change.

I believe this is the correct issue for tracking: https://github.com/awslabs/aws-sam-cli/issues/921

For now, anyone coming up here for the same confusion, this is what I'm doing to improve my experience in the interim:

```
npm i -g nodemon
nodemon --exec sam build
```

I run this in the root folder of my sam application in one terminal and run `sam local start-api` in another. This builds on every save so I don't have to do it manually.

Closing this issue.

rpstreef:
When looking at this issue, what is reported in the CLI tool ``You only need to restart SAM CLI if you update your AWS SAM template`` and what it actually does, does not correspond.

Is there an intention to fix this bug? Or was this never intended to work like this in the first place?

jfuss:
[...]

Issue #901, opened 2018-12-30
start-api automatic reload does not appear to be doing anything


ryancole:
Hi folks,

I'm running the following command: `sam local start-api`

I'm running the above command inside of a directory that I created using the `sam --init` command. So, it's a directory with a `template.yaml` file in it, and an actual node.js lamba function project folder.

The console output from `sam` says that I do not need to restart the CLI for changes to take affect, but modifying my code does not appear to cause any sort of automatic reload.

Do I have to do anything specific to allow this to auto reload? I'm running Windows 10.

jfuss:
You will need to build every time. The behavior is you do not need to restart the command and can update the folders with new code and have it reflected when hitting the api. With the introduction of `sam build`, you will need to build anytime your source or dependencies change. We are looking to make this easier with `sam build --watch` but that will not change the behavior of the `sam local start-api` command.

Closing as this is working as intended.

felixhageloh:
when I run `sam local start-api` I get this output 

>  You can now browse to the above endpoints to invoke your functions. You do not need to restart/reload SAM CLI while working on your functions changes will be reflected instantly/automatically. You only need to restart SAM CLI if you update your AWS SAM template

Which suggests that automatic reloading is the intended behavior

jfuss:
@felixhageloh Yes but that 'auto reloading` is not `auto building`. The reloading in this is describing that you can update the contents of the `CodeUri` and see the changes without restarting the api. With the introduction of `sam build`, this notion changed slightly.

If you are using build, what this is saying is you can build the function without needing to restart. This is because build generates a template that is used behind the scenes that points to a build location. The confusion comes from, you as a customer only interact with the non-built function/template while the invoke can (if using build) interact with the generated built template. Maybe a rewording is in order here to make the expectation better but it is working as intended. 

I create https://github.com/awslabs/aws-sam-cli/issues/921 to track the work for 'auto building` through `sam build --watch`.

Note: That output was from the pre-`sam build` stage.

felixhageloh:
@jfuss thanks for the clarification. I spent some time reading the docs in the meantime and they also all clearly state that you need to run build again.

eamarce:
Hey guys, 

We created a small npm package based in nodemon (samwatch):

https://www.npmjs.com/package/samwatch


[...]

PR #4423, opened 2022-11-16
chore: Clarify note about code building on `sam local start-api`

yuyokk:
Hi team,

Running code locally with `sam local start-api` and see this message 

> You can now browse to the above endpoints to invoke your functions. You do not need to restart/reload SAM CLI while working on your functions, changes will be reflected instantly/automatically. You only need to restart SAM CLI if you update your AWS SAM template

I assumed (similar to what was described here https://github.com/aws/aws-sam-cli/issues/901 and https://github.com/aws/aws-sam-cli/issues/1921) that my typescript code will be built as well but it is not the case.

Thought of clarifying this in the text of that note.

Please let me know if this makes sense.

Thank you

> By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

### DIFF
--- a/samcli/commands/local/lib/local_api_service.py
+++ b/samcli/commands/local/lib/local_api_service.py
@@ -72,8 +72,8 @@ class LocalApiService:
         LOG.info(
             "You can now browse to the above endpoints to invoke your functions. "
             "You do not need to restart/reload SAM CLI while working on your functions, "
-            "changes will be reflected instantly/automatically. You only need to restart "
-            "SAM CLI if you update your AWS SAM template"
+            "changes will be reflected instantly/automatically (note: code still needs to be built). "
+            "You only need to restart SAM CLI if you update your AWS SAM template"
         )
 
         service.run()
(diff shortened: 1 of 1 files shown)


samcli/commands/local/lib/local_api_service.py
```diff
@@ -72,7 +72,8 @@ def start(self):
         LOG.info(
             "You can now browse to the above endpoints to invoke your functions. "
             "You do not need to restart/reload SAM CLI while working on your functions, "
-            "changes will be reflected instantly/automatically. You only need to restart "
+            "changes will be reflected instantly/automatically (note: you will still need to "
```
mildaniel, line 75: I think there's actually a little more to it than that. This message was correct for certain runtimes before `sam build`. For example, if you're using Node.js or Python without running `sam build` then this message is true. As we evolved it seems like some of our messaging hasn't.

Maybe something along the lines of?
```suggestion
            changes will be reflected instantly/automatically. If you used sam build before running local commands, you will need to re-run sam build for the changes to be picked up. You only need to restart
```

mildaniel (commented):
Thanks for bringing this up! It's a great user experience improvement to have more accurate messaging! Left a small comment to discuss some of the wording.

samcli/commands/local/lib/local_api_service.py
yuyokk: @mildaniel this makes sense!

mildaniel (approved):

mndeveci (approved):

mildaniel:
Thank you @yuyokk for your contribution!

PR #4441, opened 2022-11-25
Added Next Command Suggestions

Harleenijjar:
we are targeting to add a section for sam local start-lambda and sam local start-api commands. When users run these commands, a message will be printed as shown below to suggest a few commands that they can use next.

```
Commands you can use next
=========================

[*] Validate SAM template: sam validate
[*] Deploy: sam deploy --guided

=========================
```

Informative messages suggesting users some commands that can be used next improves user experience.

Added 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -5,6 +5,7 @@ Common CLI options shared by various commands
 import os
 import logging
 from functools import partial
+from typing import List
 
 import click
 from click.types import FuncParamType
@@ -790,7 +791,7 @@ def _space_separated_list_func_type(value):
 _space_separated_list_func_type.__name__ = "LIST"
 
 
-def generate_next_command_recommendation(command_tuples: list):
+def generate_next_command_recommendation(command_tuples: List[tuple]):
     """
     Generates a message containing some suggested commands to run next.
 
@@ -806,5 +807,5 @@ Commands you can use next:
     """
     command_list_txt = ""
     for command, description in command_tuples:
-        command_list_txt += "\n[*] " + f"{os.linesep}[*] ".join([f"{command} : {description}"])
+        command_list_txt += f"{os.linesep}[*] {command}: {description}"
     return template.format(command_list_txt)
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -801,11 +801,9 @@ def generate_next_command_recommendation(command_tuples: List[tuple]):
     template = """
 Commands you can use next:
 ================================
-{}
-
-================================
-    """
+{}"""
     command_list_txt = ""
     for description, command in command_tuples:
-        command_list_txt += f"{os.linesep}[*] {description}: {command}"
+        command_list_txt += f"[*] {description}: {command}\n"
+    command_list_txt.rstrip()
     return template.format(command_list_txt)
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -806,6 +806,6 @@ Commands you can use next:
 ================================
     """
     command_list_txt = ""
-    for command, description in command_tuples:
-        command_list_txt += f"{os.linesep}[*] {command}: {description}"
+    for description, command in command_tuples:
+        command_list_txt += f"{os.linesep}[*] {description}: {command}"
     return template.format(command_list_txt)
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -788,3 +788,23 @@ def _space_separated_list_func_type(value):
 
 
 _space_separated_list_func_type.__name__ = "LIST"
+
+
+def generate_next_command_recommendation(command_tuples: list):
+    """
+    Generates a message containing some suggested commands to run next.
+
+    :type command_tuples: list
+    :param command_tuples: list of tuples containing the command with their respective description
+    """
+    template = """
+Commands you can use next:
+================================
+{}
+
+================================
+    """
+    command_list_txt = ""
+    for command, description in command_tuples:
+        command_list_txt += "\n[*] " + f"{os.linesep}[*] ".join([f"{command} : {description}"])
+    return template.format(command_list_txt)
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -795,7 +795,7 @@ def generate_next_command_recommendation(command_tuples: List[tuple]):
     """
     Generates a message containing some suggested commands to run next.
 
-    :type command_tuples: list
+    :type command_tuples: list[tuple]
     :param command_tuples: list of tuples containing the command with their respective description
     """
     template = """
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -801,6 +801,7 @@ def generate_next_command_recommendation(command_tuples: List[Tuple[str, str]]) 
     template = """
 Commands you can use next
 =========================
-{}"""
+{}
+"""
     command_list_txt = "\n".join(f"[*] {description}: {command}" for description, command in command_tuples)
     return template.format(command_list_txt)
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -5,7 +5,7 @@ Common CLI options shared by various commands
 import os
 import logging
 from functools import partial
-from typing import List
+from typing import List, Tuple
 
 import click
 from click.types import FuncParamType
@@ -791,7 +791,7 @@ def _space_separated_list_func_type(value):
 _space_separated_list_func_type.__name__ = "LIST"
 
 
-def generate_next_command_recommendation(command_tuples: List[tuple]):
+def generate_next_command_recommendation(command_tuples: List[Tuple[str, str]]) -> str:
     """
     Generates a message containing some suggested commands to run next.
 
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -799,8 +799,8 @@ def generate_next_command_recommendation(command_tuples: List[tuple]):
     :param command_tuples: list of tuples containing the command with their respective description
     """
     template = """
-Commands you can use next:
-================================
+Commands you can use next
+=========================
 {}"""
     command_list_txt = ""
     for description, command in command_tuples:
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -802,8 +802,5 @@ def generate_next_command_recommendation(command_tuples: List[tuple]):
 Commands you can use next
 =========================
 {}"""
-    command_list_txt = ""
-    for description, command in command_tuples:
-        command_list_txt += f"[*] {description}: {command}\n"
-    command_list_txt.rstrip()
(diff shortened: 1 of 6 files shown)


samcli/commands/local/start_api/cli.py
```diff
@@ -32,6 +32,13 @@
  compiled languages or projects requiring complex packing support, we recommended you run your own building solution
 and point SAM to the directory or file containing build artifacts.
 """
+COMMAND_SUGGESTIONS = """
+Commands you can use next
+=========================
+[*]Validate SAM template: sam validate
```
hoffa, line 38: Missing space (also the other help text).

Is this supposed to be verbatim the message that shows up on `sam init`? If yes, should this be a shared variable instead? So next time we add some command/parameter we want customers to use we don't need to remember updating in multiple places.

Harleenijjar: its similiar to what sam init but the words are different.

hoffa: How are these different from the one in `sam init`?

(Comment about missing spaces after `[*]` still applies)

Harleenijjar: for sam init there is another line: [*] Create pipeline: cd hello-world-for-command-suggestions && sam pipeline init --bootstrap rather than [*]Deploy: sam deploy --guided

samcli/commands/local/start_api/cli.py
```diff
@@ -33,6 +33,14 @@
 and point SAM to the directory or file containing build artifacts.
 """
 
+COMMAND_SUGGESTIONS = """
```
hawflau, line 36: Can we make this a template that can be shared across command? Then, for each applicable command, we just provide a list of potential next commands. 
This will ensure consistency for this message across commands.

samcli/commands/_utils/options.py
```diff
@@ -788,3 +788,23 @@ def _space_separated_list_func_type(value):
...
+    template = """
+Commands you can use next:
+================================
+{}
+
+================================
+    """
+    command_list_txt = ""
+    for command, description in command_tuples:
+        command_list_txt += "\n[*] " + f"{os.linesep}[*] ".join([f"{command} : {description}"])
```
hawflau, line 809: As you are already iterating `command_tuples`, I don't think you need `str.join()`
Also, is `\n[*] " + "` not needed?

```suggestion
        command_list_txt += f"{os.linesep}[*] {command} : {description}"
```

samcli/commands/_utils/options.py
```diff
@@ -788,3 +788,23 @@ def _space_separated_list_func_type(value):
 
 
 _space_separated_list_func_type.__name__ = "LIST"
+
+
+def generate_next_command_recommendation(command_tuples: list):
```
hawflau, line 793: - Since we now have this new function, can we have some unit tests for it?
- we are expecting `command_tuples` to be a list of tuples, can we be more precise about the type hint?
- Can we also update other commands which give "next command suggestions" to use this function?

hawflau (changes requested):

samcli/commands/_utils/options.py
Harleenijjar: Done

tests/unit/commands/_utils/test_options.py
```diff
@@ -527,3 +528,20 @@ def test_skip_without_hook_package(self):
...
+    def test_generate_next_command_recommendation(self):
+        listOfTuples = [("command1", "description1"), ("command2", "description2"), ("command3", "description3")]
+        output = generate_next_command_recommendation(listOfTuples)
+        expectedOutput = """
+Commands you can use next:
+================================
+
+[*] command1: description1
+[*] command2: description2
+[*] command3: description3
```
hoffa, line 543: The example in the summary has it the other way around.

Harleenijjar: will fix!

samcli/commands/_utils/options.py
Harleenijjar: Done

tests/unit/commands/_utils/test_options.py
```diff
@@ -527,3 +528,20 @@ def test_skip_without_hook_package(self):
...
+class TestNextCommandSuggestions(TestCase):
+    def test_generate_next_command_recommendation(self):
+        listOfTuples = [("command1", "description1"), ("command2", "description2"), ("command3", "description3")]
+        output = generate_next_command_recommendation(listOfTuples)
+        expectedOutput = """
+Commands you can use next:
+================================
+
+[*] command1: description1
+[*] command2: description2
```
hoffa, line 542: Why change how it currently looks?

Currently if I run `sam init`, I get:

```
    Commands you can use next
    =========================
    [*] Create pipeline: cd sam-app && sam pipeline init --bootstrap
    [*] Validate SAM template: cd sam-app && sam validate
    [*] Test Function in the Cloud: cd sam-app && sam sync --stack-name {stack-name} --watch
```

The underline isn't aligned to the text width, there's an extra newline, and it's not clear to me what value the final `================================` on L545 adds.

tests/unit/commands/_utils/test_options.py
```diff
@@ -532,15 +532,15 @@ def test_skip_without_hook_package(self):
 
 class TestNextCommandSuggestions(TestCase):
     def test_generate_next_command_recommendation(self):
-        listOfTuples = [("command1", "description1"), ("command2", "description2"), ("command3", "description3")]
+        listOfTuples = [("description1", "command1"), ("description2", "command2"), ("description3", "command3")]
```
hoffa, line 535: A better test would be to reproduce what was done in the real code; that would've helped catch the ordering issue.

tests/unit/commands/_utils/test_options.py
Harleenijjar: I'll make it so that it mimics the old appearance

tests/unit/commands/_utils/test_options.py
Harleenijjar: will do!

samcli/commands/init/interactive_init_flow.py
```diff
@@ -238,14 +239,15 @@ def _generate_from_use_case(
...
     click.echo(summary_msg)
-    next_commands_msg = f"""
-    Commands you can use next
-    =========================
-    [*] Create pipeline: cd {name} && sam pipeline init --bootstrap
-    [*] Validate SAM template: cd {name} && sam validate
-    [*] Test Function in the Cloud: cd {name} && sam sync --stack-name {{stack-name}} --watch
-    """
-    click.secho(next_commands_msg, fg="yellow")
+    nextcommandsuggestions = generate_next_command_recommendation(
```
hoffa, line 242: All lowercase and no word separators becomes quickly illegible at a glance. You can use a shorter but descriptive name (e.g. maybe `command_suggestions`, since it's a suggestion it already implies it's going to be "next"), or get rid of the intermediary assignment.

samcli/commands/_utils/options.py
```diff
@@ -788,3 +789,21 @@ def _space_separated_list_func_type(value):
...
+def generate_next_command_recommendation(command_tuples: List[tuple]):
+    """
+    Generates a message containing some suggested commands to run next.
+
+    :type command_tuples: list
+    :param command_tuples: list of tuples containing the command with their respective description
+    """
+    template = """
+Commands you can use next:
+================================
```
hoffa, line 803: Match existing look:

```suggestion
Commands you can use next
=========================
```

I'm not suggesting the current UI is the most ideal, but it's less friction to do things in separate steps: one that only augments without changing existing behavior, and another that addresses UI shortcomings in a more encompassing manner across the SAM CLI.

samcli/commands/init/interactive_init_flow.py
Harleenijjar: I'll change the name of the symbol to `command_suggestions` as you suggested.

samcli/commands/_utils/options.py
Harleenijjar: That makes sense, Done!

samcli/commands/_utils/options.py
```diff
@@ -788,3 +789,21 @@ def _space_separated_list_func_type(value):
...
+    :param command_tuples: list of tuples containing the command with their respective description
+    """
+    template = """
+Commands you can use next
+=========================
+{}"""
+    command_list_txt = ""
+    for description, command in command_tuples:
+        command_list_txt += f"[*] {description}: {command}\n"
+    command_list_txt.rstrip()
```
hoffa, line 808: You simplify and avoid the `rstrip()` by joining values from a list comprehension (the inline list definition) instead:

```suggestion
    command_list_txt = "\n".join(f"[*] {description}: {command}" for description, command in command_tuples)
```

samcli/commands/_utils/options.py
Harleenijjar: Done!

samcli/commands/_utils/options.py
Harleenijjar: Thank you for the suggestion, made changes to implement this

samcli/commands/_utils/options.py
```diff
@@ -788,3 +789,18 @@ def _space_separated_list_func_type(value):
 
 
 _space_separated_list_func_type.__name__ = "LIST"
+
+
+def generate_next_command_recommendation(command_tuples: List[tuple]):
```
hoffa, line 794: A tighter type would be (needs `typing.Tuple`):

```suggestion
def generate_next_command_recommendation(command_tuples: List[Tuple[str, str]]) -> str:
```

Otherwise it'll accept accept any tuple and the command will fails.

hoffa (approved):

samcli/commands/_utils/options.py
Harleenijjar: Implemented!

tests/unit/commands/_utils/test_options.py
```diff
@@ -527,3 +528,20 @@ def test_skip_without_hook_package(self):
...
+            ("Test Function in the Cloud", "sam sync --stack-name {{stack-name}} --watch"),
+            ("Deploy", "sam deploy --guided"),
+        ]
+        output = generate_next_command_recommendation(listOfTuples)
+        expectedOutput = """
+Commands you can use next
+=========================
+[*] Validate SAM template: sam validate
+[*] Test Function in the Cloud: sam sync --stack-name {{stack-name}} --watch
+[*] Deploy: sam deploy --guided"""
```
hoffa, line 546: Does this show up correctly in the terminal?

There's no final newline character like before (previously the final `"""` was on a separate line).

Harleenijjar: Yes shows exactly the way it used to show before like shown below -  with the `"""` being in the separate line it would leave a new line, so moved it up.

```
Commands you can use next
=========================
[*] Validate SAM template: sam validate
[*] Deploy: sam deploy --guided
```

Harleenijjar: yes, done!

hawflau (approved):
LGTM!

PR #4444, opened 2022-11-25
Adding cfn-lint as optional parameter for SAM validate command

cdavidxu-hub:
To provide linting validation to SAM cli and fulfill feature and customer requests

Cfn-lint will be added as an optional parameter to SAM validate. When run, the linter will parse and provide detailed error messages and codes to the customer along with suggestions for potential fixes. 

This should not pose any side-effects either to core functionality of SAM validate or to CI/CD processes as it remains an optional parameter. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/validate/validate.py
```diff
@@ -3,9 +3,11 @@
 """
 import os
 
+import logging
 import boto3
 from botocore.exceptions import NoCredentialsError
 import click
+import cfnlint.core # type: ignore
```
sriram-mv, line 10: The number of imports at the top level should be as small as possible, if we really need it we should move it within `cli` method. Also what is the reason behind the `#type: ignore` , I'd rather not have us skip any lint checks.

samcli/commands/validate/validate.py
```diff
@@ -15,13 +17,25 @@
...
 @aws_creds_options
 @cli_framework_options
+@click.option(
+    "--lint",
+    is_flag=True,
+    is_eager=True,
+    help="Run linting validation on template through cfn-lint. "
+     "For more information, see: https://github.com/aws-cloudformation/cfn-lint",
+    cls=ClickMutex,
+    incompatible_params=["config-env", "config-file", "profile", "region"]
```
sriram-mv, line 37: why are these incompatible? you could very well have the template file be passed in through a configuration parameter within the 'validate' space such that it is `default.validate.parameters`, which has a template specified.

I see that cfn-lint has a way to pass in regions to its command line. what does that mapping mean here?

Does cfn-lint have any environment variables that it obeys? can they be used in conjunction through `sam validate --lint` ?

samcli/commands/validate/validate.py
```diff
@@ -50,29 +65,33 @@ def do_cli(ctx, template):
...
-            wrapped_from=no_region_found_e.__class__.__name__,
-        ) from no_region_found_e
-    except NoCredentialsError as e:
-        raise UserException(
-            "AWS Credentials are required. Please configure your credentials.", wrapped_from=e.__class__.__name__
-        ) from e
-
-    click.secho("{} is a valid SAM Template".format(template), fg="green")
+    if lint:
+        _lint(ctx, template)
```
sriram-mv, line 69: I'm assuming there needs to be a variety of error handling at this stage, so that we raise the appropriate sam cli related click errors.

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
...
+            click.secho("{} is a valid SAM Template".format(template), fg="green")
+        rules = cfnlint.core.get_used_rules()
+        matches_output = formatter.print_matches(matches, rules, filenames)
+
+        if matches_output:
+            if args.output_file:
+                with open(args.output_file, 'w', encoding='utf-8') as output_file:
+                    output_file.write(matches_output)
+            else:
+                print(matches_output)
```
sriram-mv, line 148: we should use click's functionality for outputting information instead of using a direct print.

sriram-mv (changes requested):

sriram-mv:
I also wonder if this should be its own command instead, i.e sam lint. Open Q

samcli/commands/validate/validate.py
```diff
@@ -15,13 +17,25 @@
 from samcli.lib.telemetry.metric import track_command
 from samcli.cli.cli_config_file import configuration_option, TomlProvider
 from samcli.lib.utils.version_checker import check_newer_version
+from samcli.commands._utils.click_mutex import ClickMutex
 
+LOGGER = logging.getLogger("cfnlint")
+LOGGER.propagate = False
```
mndeveci, line 23: Please use general way of instantiating the logger, one can be found here: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L66

And why are you setting `propagate` to `False`?

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
...
+        Path to the template file
+
+    """
+
+    try:
+        if ctx.debug:
+            (args, filenames, formatter) = cfnlint.core.get_args_filenames([template, "--debug"])
+        else:
+            (args, filenames, formatter) = cfnlint.core.get_args_filenames([template])
+        LOGGER.setLevel(logging.WARNING)
```
mndeveci, line 136: Why we need to explicitly set the logging level here?

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
...
+    ctx
+        Click context object
+    template
+        Path to the template file
+
+    """
+
+    try:
+        if ctx.debug:
+            (args, filenames, formatter) = cfnlint.core.get_args_filenames([template, "--debug"])
```
mndeveci, line 133: Can we add details why we need special handling of `--debug` flag here?

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
...
+            (args, filenames, formatter) = cfnlint.core.get_args_filenames([template])
+        LOGGER.setLevel(logging.WARNING)
+        matches = list(cfnlint.core.get_matches(filenames, args))
+        if not matches:
+            click.secho("{} is a valid SAM Template".format(template), fg="green")
+        rules = cfnlint.core.get_used_rules()
+        matches_output = formatter.print_matches(matches, rules, filenames)
+
+        if matches_output:
+            if args.output_file:
```
mndeveci, line 144: Where this `output_file` comes from and how users can set it?

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
...
+                with open(args.output_file, 'w', encoding='utf-8') as output_file:
+                    output_file.write(matches_output)
+            else:
+                print(matches_output)
+
+        return cfnlint.core.get_exit_code(matches)
+    except cfnlint.core.CfnLintExitException as e:
+        LOGGER.error(str(e))
+        return e.exit_code
```
mndeveci, line 153: Raise a `UserException` which will handle the error message and exit code, rather than manually returning the exit_code.

samcli/commands/validate/validate.py
```diff
@@ -95,3 +114,40 @@ def _read_sam_file(template):
         sam_template = yaml_parse(sam_template.read())
 
     return sam_template
+
+def _lint(ctx, template):
```
mndeveci, line 118: We are trying to add more typing into our repository, so can you please add typing for the parts that you have added?

mndeveci (changes requested):

samcli/commands/validate/validate.py
cdavidxu-hub: will look into it, sounds good

samcli/commands/validate/validate.py
cdavidxu-hub: will look into it

samcli/commands/validate/validate.py
cdavidxu-hub: Similar issue to logging issue mentioned above. Without the cfnlint logger being explicitly set to a higher level, a lot of INFO level logs will be exposed when they shouldnt be. Again, changing the base logging level in main.py could solve this but not sure what side effects that would have across sam cli.

samcli/commands/validate/validate.py
cdavidxu-hub: We need a way to pass the `--debug` flag to cfn-lint since cfn-lint also takes in a `--debug- optional parameter as well. Do you mean you would want some comments in the code explaining why?

samcli/commands/validate/validate.py
cdavidxu-hub: will look into it and add

samcli/commands/validate/validate.py
cdavidxu-hub: Will add try catch for click errors

samcli/commands/validate/validate.py
cdavidxu-hub: Hmmm, I see fair point - missed that. Would setting the template in the config-file also pass the template file path to the do_cli method? As for regions, I can include that as an option in the _lint method as well.

samcli/commands/validate/validate.py
cdavidxu-hub: So the issue here is that cfnlint has its own loggers as well - if we use the general way, the logger wont catch the cfnlint logs. SAM cli also sets its root logger in main.py with basicConfig set to INFO level. This causes logs from cfnlint to be duplicated using that base logger. Setting propagate to false will prevent the duplicate logs. Changing the level in main.py would also prevent that and generally loggers are defaulted to WARNING, but im not sure how that would affect other logging behaviour in sam cli.

samcli/commands/validate/validate.py
cdavidxu-hub: This is an cfn-lint optional parameter which we cannot specify for sam validate so it will never output to this file. Will remove in next commit. Optionally we can have it always print to an output file but we would need to have that be explicitly stated to the customer

samcli/commands/validate/validate.py
cdavidxu-hub: Have moved below to method below. It seems cfn-lint does not have a py.typed file/type hints and we likely will not be adding them before release. Is it alright to suppress for now?

samcli/commands/validate/validate.py
```diff
@@ -95,3 +107,68 @@ def _read_sam_file(template):
...
+        matches = list(cfnlint.core.get_matches(filenames, args))
+        if not matches:
+            click.secho("{} is a valid SAM Template".format(template), fg="green")
+        rules = cfnlint.core.get_used_rules()
+        matches_output = formatter.print_matches(matches, rules, filenames)
+
+        if matches_output:
+            click.secho(matches_output)
+
+    except cfnlint.core.InvalidRegionException as e:
```
cdavidxu-hub, line 160: Just to add some more context on why there are two region exceptions. Cfn-lint can catch invalid regions in commands like “samdev —lint —region us-southeast-2” and throw a cfnlint.core.InvalidRegionException. However, if the customer forgets to specify a region and then uses another optional parameter, like “samdev —lint —region —debug”, the cfn-lint help page will be outputted as it won’t detect any arguments for “—region”. Cfn-lint uses argparse and is configured to display the help page upon errors which we don’t want exposed to customers right now. I’ve added the botocore.util.validate_region_name method to check the format of the region string which will catch cases when the customer forgets to specify a region and uses another parameter. Let me know if this approach is acceptable or if there is a better way to raise these exceptions.

praneetap:
> I also wonder if this should be its own command instead, i.e sam lint. Open Q

curious why you think that Sriram. Making it a new command would make discovery an issue, and sam validate is already plugged into sam deploy, which is a  nice benefit here??

cdavidxu-hub:
> > I also wonder if this should be its own command instead, i.e sam lint. Open Q
> 
> curious why you think that Sriram. Making it a new command would make discovery an issue, and sam validate is already plugged into sam deploy, which is a nice benefit here??

Sorry, missed these comments - I think in speaking with the design team, awareness was certainly an issue and since linting is a familiar term and form of validation we decided to keep it as an optional parameter for sam validate.

sriram-mv:
@praneetap @cdavidxu-hub The thinking of `sam lint` was behind the fact we can make linting a first class citizen rather than a flag. I agree on the discoverability aspect though. `sam validate --lint` to me is unclear directly. Am I validating and linting at the same time? Am I just linting? In this case we are running cfn-lint alone if `sam validate --lint` is specified.


This goes against other commands, eg: `sam build --use-container` I'm still building (doing the same exact building logic) but within the container.

samcli/commands/validate/validate.py
```diff
@@ -50,29 +55,36 @@ def do_cli(ctx, template):
...
+                "AWS Region was not found. Please configure your region through a profile or --region option",
+                wrapped_from=no_region_found_e.__class__.__name__,
+            ) from no_region_found_e
+        except NoCredentialsError as e:
+            raise UserException(
+                "AWS Credentials are required. Please configure your credentials.", wrapped_from=e.__class__.__name__
+            ) from e
+
+        click.secho(
+            "{} can be transformed to a Cloudformation template. "
```
sriram-mv, line 84: could we still add "is a valid SAM Template" back and then continue to add the rest? There could be tools that are scrapping our output.

samcli/commands/validate/validate.py
```diff
@@ -95,3 +107,68 @@ def _read_sam_file(template):
...
+            wrapped_from=e.__class__.__name__,
+        ) from e
+    except cfnlint.core.CfnLintExitException as lint_error:
+        raise UserException(
+            lint_error,
+            wrapped_from=lint_error.__class__.__name__,
+        ) from lint_error
+    except InvalidRegionError as no_region_found_e:
+        raise UserException(
+            "AWS Region was not found. Please configure your region through the --region option",
```
sriram-mv, line 172: lets move the duplicated help text together, and raise the same exception.

sriram-mv (changes requested):

samcli/commands/validate/validate.py
cdavidxu-hub: The cfn-lint libraries make use of argparse so when specific errors occur the behaviour defaults to the argparse behaviour, like in this case with InvalidRegionExceptions.

requirements/base.txt
```diff
@@ -26,3 +26,6 @@ typing_extensions==3.10.0.0
 regex==2021.9.30
 # NOTE: tzlocal is not a direct dependency of SAM CLI, but pin to 3.0 as 4.0 break appveyor jobs
 tzlocal==3.0
+
+#Adding cfn-lint dependency for SAM validate
+cfn-lint==0.72.2
```
mndeveci, line 31: Please run `make update-reproducible-reqs` so that it can update pinned dependencies. You’ve marked as done in the conversation but I can’t see the file is updated.

We also need to update third party licenses file with this.

samcli/commands/validate/validate.py
```diff
@@ -9,37 +9,42 @@
...
 @template_option_without_build
 @aws_creds_options
 @cli_framework_options
+@click.option(
+    "--lint",
+    is_flag=True,
+    help="Run linting validation on template through cfn-lint. "
+    "For more information, see: https://github.com/aws-cloudformation/cfn-lint",
+    cls=ClickMutex,
+    incompatible_params=["profile"],
```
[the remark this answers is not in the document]
mndeveci: We should only put these checks if these parameters can’t work together. If not then let’s remove this restriction as user might have defined general default for their profile, and that shouldn’t fail to run with this option.

samcli/commands/validate/validate.py
```diff
@@ -95,3 +108,63 @@ def _read_sam_file(template):
...
+
+    """
+
+    import cfnlint.core  # type: ignore
+    import logging
+    from samcli.commands.exceptions import UserException
+    from botocore.utils import validate_region_name
+    from botocore.exceptions import InvalidRegionError
+
+    LOGGER = logging.getLogger("cfnlint")
```
mndeveci, line 137: Can we add some comments why we need these logging settings? And also change logger name to something that will hint this is related to cfn-lint? Lastly, I don’t think this should be all uppercase since it is not constant anymore, it is just another variable in a function.

samcli/commands/validate/validate.py
```diff
@@ -95,3 +108,63 @@ def _read_sam_file(template):
...
+    LOGGER.propagate = False
+
+    try:
+        validate_region_name(ctx.region)
```
mndeveci, line 141: I think this is good (I didn’t know such function exist) but I believe we should be doing this higher level, instead of manually checking it.

tests/integration/validate/test_validate_command.py
```diff
@@ -87,3 +100,23 @@ def test_validate_logs_warning_for_cdk_project(self):
...
+            ("default_yaml", TemplateFileTypes.YAML),  # project with template.yaml
+            ("default_json", TemplateFileTypes.JSON),  # project with template.json
+            ("multiple_files", TemplateFileTypes.YAML),  # project with both template.yaml and template.json
+            (
+                "with_build",
+                TemplateFileTypes.JSON,
+            ),  # project with template.json and standard build directory .aws-sam/build/template.yaml
+        ]
+    )
+    def test_lint_template(self, relative_folder: str, expected_file: TemplateFileTypes):
```
mndeveci, line 115: These are all happy path tests, can we add some which will throw error?

mndeveci (changes requested):

samcli/commands/validate/validate.py
cdavidxu-hub: Will change the name. As for comments, I had added some above the logger but pylint was reporting some formatting errors. The comments are located right below the _lint function now. Let me know if those are sufficient.

samcli/commands/validate/validate.py
cdavidxu-hub: by higher level do you mean within the do_cli method? That could be better, as I believe running "sam validate --region --debug" results in printing the stack trace

sriram-mv: I agree with @mndeveci , we should move this functionality and makes it so that its commonly used across all commands within SAM CLI.

sriram-mv (changes requested):

samcli/commands/validate/validate.py
```diff
@@ -95,3 +104,63 @@ def _read_sam_file(template):
...
+    import logging
+    from samcli.commands.exceptions import UserException
+    from botocore.utils import validate_region_name
+    from botocore.exceptions import InvalidRegionError
+
+    cfn_lint_logger = logging.getLogger("cfnlint")
+    cfn_lint_logger.propagate = False
+
+    try:
+        validate_region_name(ctx.region)
```
sriram-mv, line 137: Lets move this entire check to within the cli validator for `region`

```python
from samcli.commands.exceptions import RegionError

    def callback(ctx, param, value):
        state = ctx.ensure_object(Context)
        from botocore import exceptions, utils

        try:
            utils.validate_region_name(value)
        except exceptions.InvalidRegionError as ex:
            raise RegionError(
                message=f"Provided region: {value} doesn't match a supported format", wrapped_from=ex.__class__.__name__
            )
        state.region = value
        return value

    return click.option(
        "--region", expose_value=False, help="Set the AWS Region of the service (e.g. us-east-1).", callback=callback
    )(f)
```

at https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/options.py#L41

sriram-mv (changes requested):

requirements/reproducible-linux.txt
```diff
@@ -126,6 +131,10 @@ cffi==1.15.1 \
     --hash=sha256:fa6693661a4c91757f4412306191b6dc88c1703f780c8234035eac011922bc01 \
     --hash=sha256:fcd131dd944808b5bdb38e6f5b53013c5aa4f334c5cad0c72742f6eba4b73db0
     # via cryptography
+cfn-lint==0.72.2 \
```
mndeveci, line 134: Are there any dependencies which is dynamically imported? If so we might need to add them to hidden imports information for pyinstaller.

samcli/commands/validate/validate.py
mndeveci: @cdavidxu-hub I think it was a misunderstanding that I didn't want to keep as it is, instead I was suggesting to move this validation to a more generic place or removing it completely. Let's continue with what @sriram-mv suggested above.

requirements/reproducible-linux.txt
sriram-mv: https://github.com/aws/aws-sam-cli/pull/4476

tests/unit/commands/validate/test_cli.py
```diff
@@ -106,16 +106,6 @@ def test_lint_invalid_region_argument_fails(self, click_patch, matches_patch, ar
         with self.assertRaises(UserException):
             _lint(ctx=ctx_lint_mock(debug=False, region="region"), template=template_path)
 
-    @patch("botocore.utils.validate_region_name")
```
mndeveci: Why we have removed this test, can we add it back with more generic solution that you have implemented?

tests/unit/cli/test_main.py
```diff
@@ -26,6 +27,20 @@ def test_cli_some_command(self):
             result = runner.invoke(cli, ["local", "generate-event", "s3"])
             self.assertEqual(result.exit_code, 0)
 
+    def test_cli_with_no_region_arg_validate(self):
```
sriram-mv, line 30:
```python

    def test_cli_with_non_standard_format_region(self):
        mock_cfg = Mock()
        with patch("samcli.cli.main.GlobalConfig", mock_cfg):
            runner = CliRunner()
            for command in ["validate", "deploy"]:
                result = runner.invoke(cli, [command, "--region", "--non-standard-format"])
                self.assertEqual(result.exit_code, 1)
                self.assertIn("Error: Provided region: --non-standard-format doesn't match a supported format",
                              result.output)

```

You could do a assertRaises too.

tests/unit/cli/test_main.py
```diff
@@ -26,6 +27,20 @@ def test_cli_some_command(self):
...
+        mock_cfg = Mock()
+        with patch("samcli.cli.main.GlobalConfig", mock_cfg):
+            runner = CliRunner()
+            runner.invoke(cli, ["validate", "--region", "--debug"])
+            self.assertRaises(RegionError)
+
+    def test_cli_with_no_region_arg_deploy(self):
```
sriram-mv, line 37: I don't quite get this test.

why does deploy use use-container False?

sriram-mv (changes requested):

tests/unit/cli/test_main.py
sriram-mv: Conversely if you wanted to test that `region` being set to a standard format went ok.

```python

    @patch("samcli.commands.validate.validate.do_cli")
    def test_cli_with_valid_region(self, mock_do_cli):
        mock_cfg = Mock()
        with patch("samcli.cli.main.GlobalConfig", mock_cfg):
            runner = CliRunner()
            result = runner.invoke(cli, ["validate", "--region", "us-west-2"])
            self.assertEqual(result.exit_code, 0)
        self.assertTrue(mock_do_cli.called)
        self.assertEqual(mock_do_cli.call_count, 1)
```

tests/unit/cli/test_main.py
cdavidxu-hub: Hmm I took the parameters from a separate test, but ultimately I just want to test the region behaviour. Would it work with "deploy --region --debug"? If so, I will just remove the use-container parameter

tests/unit/cli/test_main.py
```diff
@@ -26,6 +27,20 @@ def test_cli_some_command(self):
...
+        mock_cfg = Mock()
+        with patch("samcli.cli.main.GlobalConfig", mock_cfg):
+            runner = CliRunner()
+            runner.invoke(cli, ["deploy", "use-container" "False", "--region", "--debug"])
```
mndeveci, line 41: It seems like both tests are using an empty region since another flag is provided right after `--region`? I would rather have;
- A test with a valid region
- A test with an invalid region
- A test with empty region

tests/unit/cli/test_main.py
```diff
@@ -26,6 +27,28 @@ def test_cli_some_command(self):
             result = runner.invoke(cli, ["local", "generate-event", "s3"])
             self.assertEqual(result.exit_code, 0)
 
+    def test_cli_with_no_region_argument(self):
```
sriram-mv, line 30: this is not a no region argument test. its a non standard format test.

for a no region argument, it would look like this.

```python
    def test_cli_with_empty_region(self):
        mock_cfg = Mock()
        with patch("samcli.cli.main.GlobalConfig", mock_cfg):
            runner = CliRunner()
            for command in ["validate", "deploy"]:
                result = runner.invoke(cli, [command, "--region"])
                self.assertEqual(result.exit_code, 2)
                self.assertIn("Error: Option '--region' requires an argument",
                              result.output)
```

mndeveci (approved):

sriram-mv (approved):

PR #4454, opened 2022-12-01
chore: Terraform prepare hook refactor

lucashuy:
The prepare hook file for Terraform is quite large. We had to disable the linter as a result. The test file that accompanies it is also large.

This change moves a lot of the logic into its own modules (files). This will make it easier to edit and add new tests in the future.

The logic is now spread out between five new files and two existing files:
```
samcli/hook_packages/terraform/hooks/prepare
├── constants.py (new!)
├── enrich.py (new!)
├── exceptions.py
├── hook.py (modified)
├── makefile_generator.py (new!)
├── property_builder.py (new!)
├── resource_linking.py
├── translate.py (new!)
└── types.py (modified)
```

The methods contained in the new files are all related to the name of the file (`enrich.py` contains all logic related to enriching, `makefile_generator.py` is for all Makefile and rule generation, etc). Constants were moved to `utils.py`.

The unit tests are updated in the a similar fashion. I've added a new `PrepareHookUnitBase` class that contains all the Terraform related property setup.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -314,7 +314,8 @@ ignored-classes=SQLObject
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E0201 when accessed. Python regular
 # expressions are accepted.
-generated-members=REQUEST,acl_users,aq_parent,objects,DoesNotExist,md5,sha1,sha224,sha256,sha384,sha512
+# Stub files are also missed by pylint: https://github.com/PyCQA/pylint/issues/4987
+generated-members=REQUEST,acl_users,aq_parent,objects,DoesNotExist,md5,sha1,sha224,sha256,sha384,sha512,pywintypes
 
 
 [VARIABLES]
--- a/mypy.ini
+++ b/mypy.ini
@@ -19,9 +19,6 @@ ignore_errors=True
 # below are packages/modules that do not have stubs available
 #
 
-[mypy-pywintypes]
-ignore_missing_imports=True
-
 [mypy-botocore,botocore.*]
 ignore_missing_imports=True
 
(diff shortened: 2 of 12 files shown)


samcli/hook_packages/terraform/hooks/prepare/enrich.py
```diff
@@ -0,0 +1,646 @@
+"""
+Terraform resource enrichment
```
lucashuy, line 2: We should make these a bit more clear on what the module does

samcli/hook_packages/terraform/hooks/prepare/enrich.py
```diff
@@ -0,0 +1,646 @@
...
+    AWS_LAMBDA_LAYERVERSION as CFN_AWS_LAMBDA_LAYER_VERSION,
+)
+from samcli.lib.utils.packagetype import ZIP, IMAGE
+from samcli.lib.hook.exceptions import PrepareHookException
+
+
+LOG = logging.getLogger(__name__)
+
+
+def _enrich_resources_and_generate_makefile(
```
lucashuy, line 39: Some of these methods are no longer protected/private, we should remove the underscores from the ones that are imported else where

samcli/hook_packages/terraform/hooks/prepare/makefile.py
```diff
@@ -0,0 +1,260 @@
+"""
```
lucashuy, line 1: Lets change the name of this file to be a bit more specific, something like `makefile_generator.py` or `makefile_generation.py`

mildaniel (approved):
Thanks for doing this refactoring. Looks good!

torresxb1 (approved):
🚀  Thanks Lucas!

PR #4458, opened 2022-12-02
Update pyinstaller to include modules under samcli dir as hidden imports

hawflau:

N/A

When we add modules under the "samcli" directory that's imported dynamically, it's easy to miss adding it into hidden imports in the pyinstaller script.

Now the hidden imports just include all modules under the samcli directory.

TBD - need to test the artifacts

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


lucashuy (approved):

tests/unit/cli/test_pyinstaller_imports.py
```diff
@@ -22,3 +22,42 @@ def test_hook_not_contain_self_defined_command_packages(self):
 
         for name in command_package_names:
             self.assertNotIn(name, hidden_imports.SAM_CLI_HIDDEN_IMPORTS)
+
+    def test_hook_contain_all_samcli_hook_packages(self):
```
sriram-mv, line 26: How is this test passing? I understand the imports are done because of `walk_modules`, but how are the modules present within SAM_CLI_HIDDEN_IMPORTS

installer/pyinstaller/hidden_imports.py
```diff
@@ -1,8 +1,31 @@
...
-SAM_CLI_HIDDEN_IMPORTS = _SAM_CLI_COMMAND_PACKAGES + [
-    # terraform hook
-    "samcli.hook_packages.terraform",
+
+class HasPathAndName(Protocol):
+    __path__: str
+    __name__: str
+
+
+def walk_modules(module: HasPathAndName, visited: set) -> None:
```
mndeveci, line 12: Just wondering, do we need to change anything in the hooks, or this should play well with what we already have there? https://github.com/aws/aws-sam-cli/blob/develop/installer/pyinstaller/hook-samcli.py#L10

qingchm:
Taking over this PR to test on these changes and addressing the comments

tests/unit/cli/test_pyinstaller_imports.py
qingchm: @sriram-mv it's the logic here: SAM_CLI_HIDDEN_IMPORTS = list(samcli_modules) + [...other imports]

qingchm: Every walked module will be in the list samcli_modules

installer/pyinstaller/hidden_imports.py
qingchm: Good question, this is how I understand it: for py files that are used as data files (never got statically or dynamic imported), it doesn't add any value by walking through them and adding to hidden packages (we tested this before), only when we add them as data files they will work as expected. So for any module-only additions they will go into the hidden imports. But for any terraform like projects that have also py files as data files, we still need to manually add them in hook-samcli.py as data files. Any miss in that would cause pyInstaller products to miss files. So data import and the walk through logic doesn't conflict, but we do still have this manual need to update data files with code packages automatically included

mndeveci (approved):
LGTM! Let's test changes with generated pyinstaller binaries and confirm there are no side effects.

installer/pyinstaller/hidden_imports.py
```diff
@@ -1,8 +1,31 @@
...
+            submodule = __import__(pkg.name)
+            submodule = cast(HasPathAndName, submodule)
+            walk_modules(submodule, visited)
+
+samcli_modules = set(["samcli"])
+samcli = __import__("samcli")
+samcli = cast(HasPathAndName, samcli)
+walk_modules(samcli, samcli_modules)
+
+SAM_CLI_HIDDEN_IMPORTS = list(samcli_modules) + [
```
qingchm, line 28: Raised one concern which is performance, as we know even if we only specify the parent module, the pyinstaller will also automatically pick up everything under it. Now this walking logic adds all the sub modules as separate packages, I am afraid this will create exponentially more iterations for PyInstaller. But with testing we know now this change **does not** make the PyInstaller build process slower. Just commenting on here to address similar concerns

qingchm (approved):

PR #4462, opened 2022-12-06
fix: change permissions on package universally

sriram-mv:
* Remove chance of deployment failure due to wrong permissions on zip packaging.

* To address https://aws.amazon.com/premiumsupport/knowledge-center/lambda-deployment-package-errors/

* Changes permissions more universally such that AWS Lambda is able to invoke the handler.

* Potential side affect of escalation of privileges of files to be zipped.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/package/utils.py
```diff
@@ -242,24 +242,32 @@ def make_zip(file_name, source_root):
...
-                            # Set host OS to Unix
-                            info.create_system = 3
-                            zf.writestr(info, file_bytes, compress_type=compression_type)
-                    else:
-                        zf.write(full_path, relative_path)
+                    # NOTE(sriram-mv): Keeping windows check AS-IS and setting permissions to 755,
+                    # since changing it overall would mean a regression in the file permissions.
+                    # This should be done after DUE DELIBERATION. Ideally there should only
+                    # be ONE way to zip up contents regardless of the OS.
+                    with open(full_path, "rb") as data:
```
sriram-mv, line 249: Should this be generic for just `Lambda` permissions?

samcli/lib/package/utils.py
```diff
@@ -166,9 +168,14 @@ def upload_local_artifacts(
...
 
-    # Or, pointing to a folder. Zip the folder and upload
+    # Or, pointing to a folder. Zip the folder and upload (zip_method is changed based on resource type)
     if is_local_folder(local_path):
-        return zip_and_upload(local_path, uploader, extension)
+        return zip_and_upload(
+            local_path,
+            uploader,
+            extension,
+            zip_method=make_zip_with_lambda_permissions if resource_id in LAMBDA_LOCAL_RESOURCES else make_zip,
```
sriram-mv, line 177: added a check on just lambda specific resources.

tests/unit/lib/package/test_artifact_exporter.py
```diff
@@ -1636,6 +1661,46 @@ def test_make_zip_windows(self, mock_system):
...
+                for info in zf.infolist():
+                    files_in_zip.add(info.filename)
+                    permission_bits = (info.external_attr & external_attr_mask) >> 16
+                    if not platform.system().lower() == "windows":
+                        if info.is_dir():
+                            self.assertEqual(permission_bits, 0o100755)
+                        else:
+                            self.assertEqual(permission_bits, 0o100644)
+                    else:
+                        self.assertEqual(permission_bits, 0o100755)
```
jfuss, line 1695: I might be miss reading this but looks like if you are not on windows, all permissions are 755? Shouldn't there be a directory vs file check here too?

tests/unit/lib/package/test_artifact_exporter.py
```diff
@@ -1636,6 +1661,46 @@ def test_make_zip_windows(self, mock_system):
                 os.remove(zipfile_name)
             test_file_creator.remove_all()
 
+    def test_make_zip_lambda_resources(self):
```
jfuss, line 1664: Can we have tests that verify each resource we are doing this to and another one for the unsupported ones? I would rather over test this to ensure no escalation for resources we don't intend too (now and the future).

At the same note, Can we have a test that also verifies local permissions are not escalated? The code looks correct to me but just want to have a good suite of tests surrounding this change.

jfuss (changes requested):

tests/unit/lib/package/test_artifact_exporter.py
sriram-mv: I have kept the permissions as-is because of the previous change increased the permissions overall, it did not matter if it was a directory or a file.

tests/unit/lib/package/test_artifact_exporter.py
sriram-mv: For every lambda resource: https://github.com/aws/aws-sam-cli/pull/4462/files#diff-756169fa4653144117b5d40b30eb21a08c4b4a19ab93d0029a0d69afa2e9a692R342 we test that we are only ever hitting `make_zip_with_lambda_permissions`

jfuss (approved):

samcli/lib/package/utils.py
```diff
@@ -242,28 +255,47 @@ def make_zip(file_name, source_root):
...
-                            # https://github.com/aws/aws-sam-cli/pull/2193#discussion_r513110608
-                            # Changed to 0755 due to a regression in https://github.com/aws/aws-sam-cli/issues/2344
-                            # Mimicking Unix permission bits and recommanded permission bits
-                            # in the Lambda Trouble Shooting Docs
-                            info.external_attr = 0o100755 << 16
                             # Set host OS to Unix
                             info.create_system = 3
+                            info.external_attr = (
+                                0o100755 << 16
+                                if platform.system().lower() == "windows"
```
mndeveci, line 274: Do we still need this **if windows** statement, given that we are already calling this with `755` permissions for windows at line 294?

mndeveci (approved):

PR #4469, opened 2022-12-08
chore: Update Closed Stale Issues Github Action

jfuss:
None

To reduce the load on maintainers, we will close stale issues. This keeps lingering issues to a minimum and gives maintainers better view of the repo state.

GHA will run daily, and look for stale issues. If a stale issue is found and does not have an exempt label, it will be marked "stale". Once the issue is in a stale state, it will be closed after a determined about of time (see config).

This will not closed except labels, issues with 1 or more reactions, and issues that have been recently commented on. 

This will close issues and mark them stale but should be done only after a maintainer responses first (and removes one of the exempt labels set by default). If there are issues, the issues can always be reopen, making this not something we are locked into and can get around.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -54,4 +54,4 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         loglevel: DEBUG
         # Set dry-run to true to not perform label or close actions.
-        # dry-run: true
+        dry-run: true
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -30,10 +30,10 @@ jobs:
           If you want to keep this issue open, please leave a comment below and auto-close will be canceled.
 
         # These labels are required
-        stale-issue-label: blocked/close-if-inactive
-        exempt-issue-labels: no-autoclose, stage/needs-attention, type/feature, stage/needs-triage, type/bug, stage/needs-investigation 
+        stale-issue-label: type/stale
+        exempt-issue-labels: no-autoclose, stage/needs-attention, type/feature, stage/needs-triage, type/bug, stage/needs-investigation, maintainer/need-followup, maintainer/need-response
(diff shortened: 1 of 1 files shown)


.github/workflows/close-stale-issues.yml
```diff
@@ -29,9 +31,9 @@ jobs:
 
         # These labels are required
         stale-issue-label: blocked/close-if-inactive
```
mndeveci, line 33: Should we add [blocked/more-info-needed](https://github.com/aws/aws-sam-cli/labels/blocked%2Fmore-info-needed) label as well? Assuming we asked for more information which wasn't provided for certain time period.

.github/workflows/close-stale-issues.yml
```diff
@@ -29,9 +31,9 @@ jobs:
...
-        exempt-issue-labels: no-autoclose, stage/needs-attention
+        exempt-issue-labels: no-autoclose, stage/needs-attention, type/feature, stage/needs-triage, type/bug, stage/needs-investigation
```
mndeveci, line 34: Should we include;
- [maintainer/need-followup](https://github.com/aws/aws-sam-cli/labels/maintainer%2Fneed-followup)
- [maintainer/need-response](https://github.com/aws/aws-sam-cli/labels/maintainer%2Fneed-response)

.github/workflows/close-stale-issues.yml
mndeveci: And also, [type/duplicate](https://github.com/aws/aws-sam-cli/labels/type%2Fduplicate) (which has 0 issues now but...)

jfuss: `stale-issue-label` gets added to the issues once it becomes "stale". I would prefer not to overload this too much and instead have a different label for stale-ness. So I will update this with a different label to signal the difference.

mndeveci (approved):

sriram-mv (approved):

PR #4470, opened 2022-12-09
ci: validate hidden imports during unit & integration tests

mndeveci:
N/A

If dynamic imports are used but missed in pyinstaller hidden imports configuration, it will create issues with installers that is created with pyinstaller.

If `sam` is executed in development mode (`samdev`) it replaces `import_module` with a proxy method which will analyze every dynamic import and make sure that it is defined in hidden imports.

It shouldn't make any change for `sam` and `sam-nightly` executables since proxy method is only attached if it is running as `samdev` (development mode). 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/cli/test_import_module_proxy.py
```diff
@@ -0,0 +1,30 @@
...
+from parameterized import parameterized
+
+from samcli.cli import hidden_imports
+from samcli.cli.import_module_proxy import attach_import_module_proxy, MissingDynamicImportError
+
+
+class TestImportModuleProxy(TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        attach_import_module_proxy()
```
marekaiv, line 13: Do we need to have a "detach" and call it in teardown in order to ensure that subsequent unit tests are not importing through the proxy?  I'm not sure this is needed, and I'm also not sure whether unit tests are run in parallel by the same python interpreter which would also cause this concern.  The concern is that the proxy may interfere with tests that need to load classes not under `test`.

mndeveci: Thanks for suggestions @marekaiv, I've added a detach method to remove the proxy after the unit tests are completed. I've also tested running unit tests in parallel but I haven't see any other test failing for this reason. But since this is really depend on a race condition and might happen in the future if we choose to run unit tests in parallel, I've added a comment for the test class, so that there could be some clues for investigation.

installer/pyinstaller/hook-samcli.py
```diff
@@ -1,5 +1,5 @@
 from PyInstaller.utils import hooks
-from installer.pyinstaller.hidden_imports import SAM_CLI_HIDDEN_IMPORTS
+from samcli.cli.hidden_imports import SAM_CLI_HIDDEN_IMPORTS
```
jfuss, line 2: Why do you need this under `samcli`?

samcli/cli/main.py
```diff
@@ -117,6 +117,14 @@ def cli(ctx):
     import atexit
     from samcli.lib.telemetry.metric import send_installed_metric, emit_all_metrics
 
+    # if development version of SAM CLI is used, attach module proxy
+    # to catch missing configuration for dynamic/hidden imports
+    if ctx and ctx.command_path == "samdev":
```
jfuss, line 122: Is there any risk that this can impact the production cli? 

General things look fine but just thinking of worse cases here. The way this is done it will package the "samdev" only things within the production CLI. 

Just want to ensure we thought this through.

samcli/cli/import_module_proxy.py
```diff
@@ -0,0 +1,60 @@
...
+
+def attach_import_module_proxy():
+    """
+    Attaches import_module proxy which will analyze every dynamic import and raise an error if it is not defined in
+    hidden imports configuration
+    """
+    importlib.import_module = _dynamic_import
+
+
+def detach_import_module_proxy():
```
sriram-mv, line 56: If this is used only for testing, should we move it to a testing utility?

samcli/cli/main.py
```diff
@@ -117,6 +117,14 @@ def cli(ctx):
     import atexit
     from samcli.lib.telemetry.metric import send_installed_metric, emit_all_metrics
 
+    # if development version of SAM CLI is used, attach module proxy
+    # to catch missing configuration for dynamic/hidden imports
+    if ctx and getattr(ctx, "command_path", None) == "samdev":
```
sriram-mv, line 122: can we configure the binary name, and also make it such that its injectable (maybe env var?)? especially since we have multiple binary names such as samdev, sam-nightly and now sam-dev from your GH actions PR.

sriram-mv (changes requested):

samcli/cli/import_module_proxy.py
mndeveci: Makes sense, will move this to the testing part

samcli/cli/main.py
```diff
@@ -117,6 +117,14 @@ def cli(ctx):
...
+        from samcli.cli.import_module_proxy import attach_import_module_proxy
+
+        LOG.info("Attaching import module proxy for analyzing dynamic imports")
+        attach_import_module_proxy()
```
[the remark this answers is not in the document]
mndeveci: As we discussed offline, we don't need to detach the proxy for regular execution. I will move the detach functionality into the unit test part in a follow-up commit.

samcli/cli/main.py
mndeveci: As discussed offline, we need a better way for CLI to understand which context it is operating now. Using `samdev` for now is OK but I will add a TODO here to investigate it further during release improvements in the future.

mndeveci:
Moved to https://github.com/aws/aws-sam-cli/pull/4503

PR #4479, opened 2022-12-14
fix:  Fix validate command integration tests console output missmatch and update pyyaml version requirement

qingchm:
- The validate integration tests are failing due to the expected non-lint output pattern not matching with what `sam validate` without `--lint` option will log to stdout
- Validate command will fail for using the --lint option if the pyyaml version is not >=5.4.1

- To fix the failing integration tests and resolve the console output mismatch
- To update pyyaml version in base.txt for compatible version requirement since to use --lint option, pyyaml > 5.4 is needed
  - Note: since reproducible linux already has pyyaml as 5.4.1, no changes are introduced onto reproducible-linux.txt and THIRD-PARTY-LICENSES

- Updated the related integration test and verified the tests are passing locally
- Updated base.txt under requirements folder

N.A.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -6,7 +6,7 @@ Werkzeug<2.1
 #Need to add Schemas latest SDK.
 boto3>=1.19.5,==1.*
 jmespath~=0.10.0
-PyYAML~=5.3
+PyYAML>=5.4.1,==5.*
 cookiecutter~=2.1.1
 aws-sam-translator==1.55.0
 #docker minor version updates can include breaking changes. Auto update micro version only.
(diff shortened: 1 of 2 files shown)


mndeveci (approved):

requirements/base.txt
```diff
@@ -6,7 +6,7 @@ Werkzeug<2.1
 #Need to add Schemas latest SDK.
 boto3>=1.19.5,==1.*
 jmespath~=0.10.0
-PyYAML~=5.3
+PyYAML>=5.4.1,==5.*
```
mndeveci, line 9: Do we need to re-run `make update-reproducible-reqs`?

qingchm: I ran it there are no changes

qingchm: Also therefore no changes to the third party licenses

PR #4485, opened 2022-12-14
fix: use resource type instead of resource id for deciding zip method

mndeveci:
N/A

With this PR https://github.com/aws/aws-sam-cli/pull/4462, we've changed some permissions of the files in ZIP package if the file is used for serverless function resources. But that is checking if logical id of the resource is type.

With this change, we are checking the resource type instead of resource logical id.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/sync/test_sync_adl.py
+++ b/tests/integration/sync/test_sync_adl.py
@@ -147,7 +147,7 @@ class TestSyncAdlWithWatchStartWithNoDependencies(TestSyncWatchBase):
         )
         read_until_string(
             self.watch_process,
-            "\x1b[32mFinished syncing Function Layer Reference Sync HelloWorldFunction.\x1b[0m\n",
+            "\x1b[32mFinished syncing Layer HelloWorldFunction",
             timeout=60,
         )
         self._confirm_lambda_error(lambda_functions[0])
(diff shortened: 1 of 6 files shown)


tests/unit/lib/package/test_artifact_exporter.py
```diff
@@ -340,8 +356,9 @@ def test_upload_local_artifacts_local_folder(self, zip_and_upload_mock):
 
     @patch("samcli.lib.package.utils.zip_and_upload")
     def test_upload_local_artifacts_local_folder_lambda_resources(self, zip_and_upload_mock):
-        for resource_id in LAMBDA_LOCAL_RESOURCES:
+        for resource_type in LAMBDA_LOCAL_RESOURCES:
```
sriram-mv, line 359: Looks like this was full-filling test before then.. Thanks for the fix!

sriram-mv (approved):

qingchm (approved):

tests/integration/sync/test_sync_adl.py
```diff
@@ -147,7 +147,7 @@ def test_sync_watch_code(self):
         )
         read_until_string(
             self.watch_process,
-            "\x1b[32mFinished syncing Function Layer Reference Sync HelloWorldFunction.\x1b[0m\n",
+            "\x1b[32mFinished syncing Layer HelloWorldFunction",
```
qingchm, line 150: Let's change this to function instead of layer, and update L136 to be the same

qingchm: Sorry nevermind this is actually supposed to be layer, but I'm raising a CR to update the validation on L136 to be the same as this one

PR #4486, opened 2022-12-15
fix: `pyinstaller` binaries

sriram-mv:
- With the inclusion of `cfn-lint` as a dependency. There are data files, python files that need to be included for `cfn-lint`.

- One of the dependencies brought in by `cfn-lint` is `pbr`, this in turn has a dependency on `jschema-to-python`. The way around it is to explictly set `PBR_VERSION` as per https://docs.openstack.org/pbr/latest/user/packagers.html. However this solution is brittle and will leak dependency code into AWS SAM CLI codebase, the cleaner solution is to package `jschema-to-python` metadata only for pyinstaller instead.

N/A

* To get a working pyinstaller binary.

* Built and checked a working installer binary at: https://github.com/sriram-mv/aws-sam-cli/actions/runs/3706626011

* None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/installer/pyinstaller/hidden_imports.py
+++ b/installer/pyinstaller/hidden_imports.py
@@ -12,5 +12,5 @@ SAM_CLI_HIDDEN_IMPORTS = _SAM_CLI_COMMAND_PACKAGES + [
     "pkg_resources.py2_warn",
     "aws_lambda_builders.workflows",
     "configparser",
-    "cfnlint.rules"
+    "cfnlint"
 ]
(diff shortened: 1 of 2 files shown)


mndeveci (approved):

qingchm (approved):

qingchm:
Is there a PR to add jschema_to_python into the hidden_imports list? Or do we only need the data files from jschema_to_python but not the library itself?

qingchm:
Also the same question goes for cfn-lint, do we only need cfn-lint.rules in https://github.com/aws/aws-sam-cli/blob/develop/installer/pyinstaller/hidden_imports.py or should we have the entire cfn-lint as a hidden import?

sriram-mv:
Yeah we only need the metadata files for `jschema-to-python`. the Library is a transitive dependency.

Good question on the second one, we do need the python files as well. Haven't tried playing around, I can do that tomorrow.

PR #4487, opened 2022-12-15
ci: add gh action to generate and test pyinstaller zip file

mndeveci:
Generates a pyinstaller zip file for linux & mac and runs some basic tests. ZIP file will also added as artifact so that it can be downloaded and tested with each PR.

`build-linux.sh` and `build-mac.sh` behaves differently if `CI_OVERRIDE` environment variable is set to `1`:
- Installation folder is changed as `aws-sam-cli-beta`
- Installed executable is changed as `sam-beta` (different than what we have as of now [`samdev`, `sam-nigtly`, `sam`])


installer/pyinstaller/build-linux.sh
```diff
@@ -21,9 +21,16 @@ else
     is_nightly="false"
 fi
 
+if [ "$CI_OVERRIDE" = "1" ]; then
+  build_folder="aws-sam-cli-dev"
+  build_binary_name="sam-dev"
```
qingchm, line 26: Maybe let's give it a different name to not confuse with the samdev command? I know we can name whichever binary name but just an idea for ourselves to reference easier

sriram-mv: `sam-ci-build` ? (Open to others too). Thought of `gamma` as well.

mndeveci: Updated with `sam-beta` as discussed offline

qingchm (approved):
LGTM! Let's ship it ASAP!

.github/workflows/validate_pyinstaller.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  build-for-linux:
+    name: build-pyinstaller-linux
+    runs-on: ubuntu-latest
```
sriram-mv, line 11: can we change this to be many linux? or containerized? for 1:1 parity?

mndeveci: We are now running in `quay.io/pypa/manylinux2014_x86_64:latest` image

.github/workflows/validate_pyinstaller.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/checkout@v3
+      - name: Build PyInstaller
+        run: |
+          chmod +x ./installer/pyinstaller/build-linux.sh
+          CI_OVERRIDE=1 ./installer/pyinstaller/build-linux.sh aws-sam-cli-linux-x86_64.zip
+      - name: Basic tests for PyInstaller
+        run: |
+          unzip .build/output/aws-sam-cli-linux-x86_64.zip -d sam-installation
+          ./sam-installation/install
+          sam-beta --version
```
sriram-mv, line 26: Not blocking: can we kick off another suite of tests that is bare bones across the commands, such that each command is atleast invoked once? could be a follow up commit. Could be a first step towards artifact based testing.

sriram-mv (approved):

.github/workflows/validate_pyinstaller.yml
mndeveci: I do agree with this. But I would rather have that as a separate PR and include the tests that we want to run with the pyinstaller. So that we can incrementally improve our installer test coverage.

PR #4490, opened 2022-12-16
Swap column order for changeset table to match stack events table

aidansteele:
n/a

It's not _necessary_ per se. I'll understand if this PR is closed without merging, because it's a very trivial annoyance: the logical resource ID and resource type columns are inconsistent between the two tables during a SAM deploy. See screenshot:

d327d6cd-1f94-4a4c-9b88-c266ed3094b0

Purely UI side effects. I'm not sure if there is any software out there that is _parsing_ the output of this CLI action. If it is, it could break. Maybe that risk of that outweighs the benefit of consistency in the UI.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -60,7 +60,7 @@ DESCRIBE_CHANGESET_FORMAT_STRING = "{Operation:<{0}} {ResourceType:<{1}} {Logica
 DESCRIBE_CHANGESET_DEFAULT_ARGS = OrderedDict(
     {
         "Operation": "Operation",
-        "ResourceType": "ResourceType",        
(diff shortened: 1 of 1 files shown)


qingchm:
Thanks for opening this PR! @aidansteele Please clear the checklist before we can start a review on this, for any item that's on the checklist let's verify it's addressed or doesn't apply to this PR! Thanks!

moelasmar:
Thanks @aidansteele for raising this PR. It is nice to have a consistent columns order in all our tables, but this change may break the customers who parse SAM CLI output. Since we could not confirm if there is any customer parse the SAM CLI output or not, I prefer to keep it as it is, and close this PR.

PR #4503, opened 2022-12-19
ci: validate hidden imports during unit & integration tests

mndeveci:
Original PR and discussions: https://github.com/aws/aws-sam-cli/pull/4470

N/A

If dynamic imports are used but missed in pyinstaller hidden imports configuration, it will create issues with installers that is created with pyinstaller.

If `sam` is executed in development mode (`samdev`) it replaces `import_module` with a proxy method which will analyze every dynamic import and make sure that it is defined in hidden imports.

It shouldn't make any change for `sam` and `sam-nightly` executables since proxy method is only attached if it is running as `samdev` (development mode). 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/hidden_imports.py
+++ b/samcli/cli/hidden_imports.py
@@ -1,7 +1,6 @@
 """
 Keeps list of hidden/dynamic imports that is being used in SAM CLI, so that pyinstaller can include these packages
 """
-import importlib
 import pkgutil
 from typing import cast
 from typing_extensions import Protocol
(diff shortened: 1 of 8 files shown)


sriram-mv (approved):

samcli/cli/hidden_imports.py
```diff
@@ -1,3 +1,6 @@
+"""
+Keeps list of hidden/dynamic imports that is being used in SAM CLI, so that pyinstaller can include these packages
+"""
```
qingchm, line 3: A quick question: what was the reason for us to move this file into samcli.cli?

qingchm: After discussion this change is required for the following reason: in order to access it for the new _dynamic_import logic, this file needs to be in SAM CLI code

jfuss (approved):

PR #4511, opened 2022-12-22
ci: fix canary failures

mndeveci:
- Change terraform tests to `assertIn` certain text is printed.
- Run `package` tests in parallel (which speeds up 10mins per package test group [~40mins in total])
- Run `delete` tests in parallel (which speeds up 8mins per delete test group [~32mins in total])
- Add `dateparser` to hidden imports which was caught by newly added pyinstaller checks (I am in between whether we should add this as exclusion to the dynamic import validation that we added)

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -296,7 +296,7 @@ for:
 
     test_script:
       - "pip install -e \".[dev]\""
-      - sh: "pytest -vv tests/integration/delete --json-report --json-report-file=TEST_REPORT-integration-delete.json"
+      - sh: "pytest -vv tests/integration/delete -n 4 --reruns 4 --json-report --json-report-file=TEST_REPORT-integration-delete.json"
 
   # Integ testing sync
   -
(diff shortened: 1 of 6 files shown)


jfuss (approved):

mildaniel (approved):

appveyor-ubuntu.yml
```diff
@@ -296,7 +296,7 @@ for:
 
     test_script:
       - "pip install -e \".[dev]\""
-      - sh: "pytest -vv tests/integration/delete --json-report --json-report-file=TEST_REPORT-integration-delete.json"
```
sriram-mv, line 299: we could use -n auto as well, as it will scale to number of CPUs available.

sriram-mv (approved):

appveyor-ubuntu.yml
mndeveci: True, but I didn't want to get throttled in case these are running in an 8 or more core environments.

Issue #4521, opened 2022-12-29
generate-event: Add S3 Batch Operations Event


arjunrawal4:
### Describe your idea/feature/enhancement

S3 Batch Operations is missing from `sam local generate-event`. It would be nice to have sample events for Batch Operations lambda invocations

### Proposal

Add new event to SAM.

```
{
    "invocationSchemaVersion": "1.0",
    "invocationId": "YXNkbGZqYWRmaiBhc2RmdW9hZHNmZGpmaGFzbGtkaGZza2RmaAo",
    "job": {
        "id": "f3cc4f60-61f6-4a2b-8a21-d07600c373ce"
    },
    "tasks": [
        {
            "taskId": "dGFza2lkZ29lc2hlcmUK",
            "s3Key": "customerImage1.jpg",
            "s3VersionId": "1",
            "s3BucketArn": "arn:aws:s3:us-east-1:0123456788:awsexamplebucket1"
        }
    ]
}
```

### Additional Details
https://docs.aws.amazon.com/lambda/latest/dg/services-s3-batch.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/batch-ops-invoke-lambda.html
https://github.com/awsdocs/amazon-s3-developer-guide/blob/master/doc_source/batch-ops-invoke-lambda.md

moelasmar:
Thanks @arjunrawal4 for raising this feature. I checked your PR and looks good to me, I will ask the team to provide another approval to merge it.

sriram-mv:
Closing as part of 1.69.0 release.

PR #4523, opened 2022-12-29
feat: Add support for S3 Batch Operations event (#4521)

arjunrawal4:
#4521 

When working with S3 Batch Operations lambda operations, it would be nice to have a test event

Adds a template for this event

- Docs will need to be updated at https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-generate-event.html

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


moelasmar (approved):

samcli/lib/generated_sample_events/event-mapping.json
```diff
@@ -917,6 +917,19 @@
     }
   },
   "s3": {
+    "batch-invocation": {
+      "filename": "S3BatchInvocation",
+      "help": "Generates an Amazon S3 Batch Operations Invocation Event",
+      "tags": {
+        "bucket-arn": {
+          "default": "arn:aws:s3:us-east-1:0123456788:examplebucket"
```
sriram-mv, line 925: s3 bucket arns are global, more information here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html

Would you be able to adjust based on the doc?

Thanks for the PR!

sriram-mv (changes requested):

samcli/lib/generated_sample_events/event-mapping.json
arjunrawal4: updated with proper bucket syntax, thanks!

sriram-mv (approved):

PR #4525, opened 2022-12-30
Update `sam --info` to display enriched info

hawflau:
N/A

To enrich `sam --info` to display system info. This makes it more straightforward for customers to provide diagnostic info when reporting bugs.

In addition to `version`, `sam --info` now shows more info, e.g.:
```
{
  "version": "1.67.0",
  "system": {
    "python": "3.9.12",
    "os": "macOS-12.6.1-x86_64-i386-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.12",
    "aws_cdk": "2.58.0 (build 4a577f3)",
    "terraform": "1.3.6"
  }
}
```

No

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/cli/main.py
```diff
@@ -49,11 +50,115 @@ def print_info(ctx, param, value):
...
+    Gather system info
+
+    Returns
+    -------
+    dict[str, str]
+    """
+    from platform import platform, python_version
+
+    info = {
+        "python": python_version(),
```
moelasmar, line 74: did you try to run `sam --info` for the different ways of installations, specially if there is no python installed on the machine?

moelasmar (dismissed):
in general it looks good, but can we add some integration test case to validate that `sam --info` is working fine and does not break.

samcli/cli/main.py
```diff
@@ -49,11 +50,115 @@ def print_info(ctx, param, value):
     if not value or ctx.resilient_parsing:
         return
 
-    click.echo(json.dumps({"version": __version__}, indent=2))
+    info = {
+        "version": __version__,
+        "system": _gather_system_info(),
+        "additional_dependencies": _gather_additional_dependencies_info(),
```
sriram-mv, line 56: Might be better to move "atleast" additional deps to another file.

samcli/cli/main.py
hawflau: `sam` command is run on the Python interpreter installed along with SAM CLI. So it still works even if a user hasn't installed Python on their own.

hawflau: Also confirmed we have existing integration test for `sam --info` already - https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/root/test_root_command.py#L13

samcli/cli/main.py
hawflau: updated

samcli/lib/utils/system_info.py
```diff
@@ -0,0 +1,105 @@
...
+    Gather system info
+
+    Returns
+    -------
+    dict[str, str]
+    """
+    from platform import platform, python_version
+
+    info = {
+        "python": python_version(),
```
sriram-mv, line 20: in the case of MSI and pyinstaller, I'm assuming this gives back the bundled python version information?

sriram-mv (approved):

lucashuy (approved):

samcli/lib/utils/system_info.py
hawflau: yes, that should be the python version of the python interpreter running the SAM CLI command.

PR #4542, opened 2023-01-05
Reduced the length of the cloned app template folder name

lucashuy:
This is one of the changes being made to help reduce the overall length of cloned app templates to avoid reaching the maximum path length in Windows.

Changes the folder name from `aws-sam-cli-app-templates` to `tmpl` for Windows machines.

Existing Windows installations with an `aws-sam-cli-app-templates` folder in their `%appdata%/AWS SAM` directory will have that folder remain there until it is manually deleted.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


qingchm (approved):

hawflau (approved):
Approved with question - With this change, does that mean we can/should revert PR #3950?

lucashuy:
> Approved with question - With this change, does that mean we can/should revert PR #3950?

From what I can gather, we can just leave that PR fix in place since its not changing anything about the way we digest the templates. Even though we could remove it since we are now reducing the paths of our templates, imo it doesn't hurt to leave it in so that we know that this won't happen with `git` in the future.

PR #4552, opened 2023-01-09
Add CodeQL GH Action

hawflau:
Enable Code Scan through GH Action

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/pipeline/bootstrap/stage.py
```diff
@@ -201,6 +201,10 @@ def generate_thumbprint(oidc_provider_url: Optional[str]) -> Optional[str]:
         # Create an IPV4 socket and use TLS for the SSL connection
         address = (url_for_certificate, 443)
         ctx = SSL.Context(SSL.TLS_METHOD)
+        ctx.set_options(SSL.OP_NO_TLSv1)
```
sriram-mv, line 204: should we verify this with unit tests?

sriram-mv (commented):
Is this going to be a gating workflow, preventing PR merges? What kind of actions does the unsuccessful run of codeql action entail on us?

sriram-mv (changes requested):
Changing my review to reflect requesting changes.

hawflau:
@sriram-mv 

> Is this going to be a gating workflow, preventing PR merges? What kind of actions does the unsuccessful run of codeql action entail on us?

Yes, it's run on each PR. Failure of this action should prevent the PR from being merged. And the PR author should fix me. 

The action will also run on schedule, because existing code might be subjected to newly discovered vulnerabilities. We will get notified and also can check from the "Security" tab in the repo.

samcli/lib/pipeline/bootstrap/stage.py
hawflau: Added

torresxb1 (approved):

Issue #4296, opened 2022-10-12
Bug: c# Binary response truncated for start-api


sudsy:
### Description:
When a stream contains a certain sequence of bytes, the response stream is truncated by sam local. This behaviour is inconsistent with the same function deployed to AWS. I noticed that despite the response being a binary stream sam local attaches the Content-Type': 'application/json' header while the same function deployed to AWS has the correct header 'Content-Type': 'application/octet-stream' making me think this is related to #325

Note that it is only a specific byte sequence that causes this. If I write zeroes to the stream they are not truncated. I am guessing that it is some sort of preamble that is being detected and stripped. Sadly the content that I want to send (A serialised protobuf message) starts with this sequence.

### Steps to reproduce:
The following Lambda function written in c# deployed with Runtime.DOTNET_6 using CDK demonstrates the issue:

 ```
public Stream FunctionHandler(Stream inputStream, ILambdaContext context)
    {
        var ms = new MemoryStream();
        ms.WriteByte(0x12);
        ms.WriteByte(0x10);
        ms.WriteByte(0x0A);
        ms.WriteByte(0x0B);
        ms.Position = 0;
        return ms;
    }
```

### Observed result:
aws lambda invoke --endpoint-url http://127.0.0.1:3001 ....

shows that the response is empty

2022-10-12 00:26:39,497 - MainThread - botocore.parsers - DEBUG - Response headers: {'Content-Type': 'application/json', 'Content-Length': '0', 'Server': 'Werkzeug/1.0.1 Python/3.7.10', 'Date': 'Wed, 12 Oct 2022 00:26:39 GMT'}

while the same request to aws looks like this:

2022-10-12 00:26:31,180 - MainThread - botocore.parsers - DEBUG - Response headers: {'Date': 'Wed, 12 Oct 2022 00:26:32 GMT', 'Content-Type': 'application/octet-stream', 'Content-Length': '4', 'Connection': 'keep-alive', 'x-amzn-RequestId': 'd15ffe5c-e6e8-4f45-8f67-ef7afdc67fa9', 'x-amzn-Remapped-Content-Length': '0', 'X-Amz-Executed-Version': '$LATEST', 'X-Amzn-Trace-Id': 'root=1-634609b8-53817f9f61a9b43f2b67a41e;sampled=0'}

### Expected result:

I was expecting Sam local to behave the same as AWS

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

[...]

PR #4563, opened 2023-01-11
fix: remove log & data separation from base_local_service

mndeveci:
#4296

Before switching to use [AWS Lambda RIE](https://github.com/aws/aws-lambda-runtime-interface-emulator), `sam local` commands was trying to parse the output in STDOUT and were trying to distinguish the log messages vs the response of the lambda function.

With switching to AWS Lambda RIE, function response is now gathered through an API call and written to STDOUT [here](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/container.py#L292-L297)
and function logs are written to STDERR in this [thread](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/container.py#L307). With this change, only function response is written to STDOUT and only function logging is written to STDERR.

By removing the special logic to parse STDOUT in order to distinguish the logs with actual function response.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/services/base_local_service.py
```diff
@@ -81,7 +82,7 @@ def service_response(body, headers, status_code):
 
 class LambdaOutputParser:
     @staticmethod
-    def get_lambda_output(stdout_stream):
+    def get_lambda_output(stdout_stream: io.BytesIO) -> Tuple[str, bool]:
```
jfuss, line 85: I wonder if `NamedTuple` or a Dataclass would provide better readability and typing. I would not block on this but just stood out to me with the update.

samcli/local/services/base_local_service.py
```diff
@@ -96,39 +97,18 @@ def get_lambda_output(stdout_stream):
...
-        last_line_position = stdout_data.rfind(b"\n")
-        if last_line_position >= 0:
-            # So there are multiple lines. Separate them out.
-            # Everything but the last line are logs
-            lambda_logs = stdout_data[:last_line_position]
-            # Last line is Lambda response. Make sure to strip() so we get rid of extra whitespaces & newlines around
-            lambda_response = stdout_data[last_line_position:].strip()
-
-        lambda_response = lambda_response.decode("utf-8")
+        lambda_response = stdout_stream.getvalue().decode("utf-8")
```
jfuss, line 103: Do we still need the `rstrip(b"\n")` after the `getvalue`? We had it there before but not sure if it is strictly required anymore.

jfuss (commented):
Overall seems like a good change. Just a couple small comments I would like a response on before accepting.

hnnasit (approved):
LGTM

samcli/local/services/base_local_service.py
mndeveci: I do agree it would be better, but I didn't want to introduce more changes to this PR, I've only added some typing into the functions that is been changed.

samcli/local/services/base_local_service.py
mndeveci: I've confirmed this by running local integration tests (some of them confirms the output of the command). So I believe we don't need that particular line

jfuss (approved):

PR #4565, opened 2023-01-12
Handled unhandled exception to show bug report message

hawflau:
N/A

To show a message that guides customers to create a bug report in case of unhandled exceptions

- Handle the "unhandled exception" by printing the stack trace and the message with a github url
- Include handling of `click.BadOptionUsage`, `click.BadArgumentUsage`, `click.BadParameter` and `click.UsageError` into the handling of `UserException`

Example:
```
> samdev build
.
.
.
'BuildProperties': {'Minify': False, 'Target': 'es2020', 'Sourcemap': True, 'External': ['./my-module-a.js', './my-module-b.js'], 'EntryPoints': ['app.js']}} architecture: x86_64 functions: HelloWorldFunction

Traceback:
  File "/Volumes/workplace/gh/aws-sam-cli/.venv39/lib/python3.9/site-packages/click/core.py", line 1055, in main
.
.
.
  File "/Volumes/workplace/gh/aws-sam-cli/samcli/lib/build/workflow_config.py", line 70, in get_selector
    raise UnsupportedBuilderException("'{}' does not have a supported builder".format(specified_workflow))

While trying to execute "samdev build", we encountered an unexpected error.
To create a bug report, follow the Github issue template below:
https://github.com/aws/aws-sam-cli/issues/new?template=Bug_report.md&title=samdev%20build%20-%20UnsupportedBuilderException
```

The actual exitcode of the execution in case of `UnhandledException` is still 1. This is to not break existing behavior. We only report 255 to Telemetry.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/exceptions.py
```diff
@@ -25,6 +28,41 @@ def __init__(self, message, wrapped_from=None):
...
+            file = click._compat.get_text_stderr() # pylint: disable=protected-access
+
+        tb = "".join(traceback.format_tb(self.__traceback__))
+        click.echo(f"\nTraceback:\n{tb}", file=file, err=True)
+
+        url = self.GH_BUG_REPORT_URL.format(
+            title=urllib.parse.quote(f"{self._command} - {type(self._exception).__name__}")
+        )
+        msg = (
+            f'While trying to execute "{self._command}", we encountered an unexpected error.\n'
```
sriram-mv, line 59: Let's not use language like "we". Its better to write this in 3rd person as the tool is operating on its own when run by a user.

sriram-mv (changes requested):

samcli/commands/exceptions.py
hawflau: Updated

samcli/commands/exceptions.py
```diff
@@ -25,6 +28,42 @@ def __init__(self, message, wrapped_from=None):
...
+        click.echo(f"\nTraceback:\n{tb}", file=file, err=True)
+
+        url = self.GH_BUG_REPORT_URL.format(
+            title=urllib.parse.quote(f"{self._command} - {type(self._exception).__name__}")
+        )
+        msg = (
+            f'An unexpected error was encountered while executing "{self._command}".\n'
+            "To create a bug report, follow the Github issue template below:\n"
+            f"{url}"
+        )
```
torresxb1, line 63: I feel like this could make it easier to have duplicate issues (multiple issues asking about the same problem). Could we first give them a link to search issues with the same title that this issue would be auto-populated?
For example, for the example in your PR description it would be:

https://github.com/aws/aws-sam-cli/issues?q=is%3Aissue+is%3Aopen+samdev+build+-+UnsupportedBuilderException

samcli/commands/exceptions.py
```diff
@@ -25,6 +28,42 @@ def __init__(self, message, wrapped_from=None):
         click.ClickException.__init__(self, message)
 
 
+class UnhandledException(click.ClickException):
+    """
+    Exception class to re-wrap any exception that is not a UserException.
+    Typically this means there is a bug in SAM CLI.
+    """
+
+    GH_BUG_REPORT_URL = "https://github.com/aws/aws-sam-cli/issues/new?template=Bug_report.md&title={title}"
```
torresxb1, line 37: Just to be safe, could you add a comment on the issue template that this template name shouldn't be changed, otherwise I think it would break this link:
https://github.com/aws/aws-sam-cli/blob/develop/.github/ISSUE_TEMPLATE/Bug_report.md

samcli/commands/exceptions.py
```diff
@@ -25,6 +28,42 @@ def __init__(self, message, wrapped_from=None):
...
+    exit_code = 1
+
+    def __init__(self, command: str, exception: Exception) -> None:
+        self._command = command
+        self._exception = exception
+        self.__traceback__ = self._exception.__traceback__
+
+        click.ClickException.__init__(self, type(exception).__name__)
+
+    def show(self, file: t.Optional[t.IO] = None) -> None:
```
torresxb1, line 48: This is a minor nit, but I think we usually import the specific types directly, instead of importing the whole typing 🤔

samcli/lib/telemetry/metric.py
```diff
@@ -148,16 +148,18 @@ def wrapped(*args, **kwargs):
...
-            else:
+            # TODO (hawflau): review exitcode meaning. We now understand exitcode 1 as errors fixable by users.
+            exit_code = 1
+            if hasattr(ex, "wrapped_from") and ex.wrapped_from:
                 exit_reason = ex.wrapped_from
+            else:
+                exit_reason = type(ex).__name__
         except Exception as ex:
-            exception = ex
+            command = ctx.command_path if ctx else ""
```
torresxb1, line 161: Is there any normal case when `ctx` is None? Or is this just for weird edge cases?

samcli/lib/telemetry/metric.py
```diff
@@ -148,16 +148,18 @@ def wrapped(*args, **kwargs):
             # Execute the function and capture return value. This is returned back by the wrapper
             # First argument of all commands should be the Context
             return_value = func(*args, **kwargs)
-        except UserException as ex:
+        except (UserException, click.BadOptionUsage, click.BadArgumentUsage, click.BadParameter) as ex:
```
torresxb1, line 151: why did we add more exceptions to catch?

samcli/lib/telemetry/metric.py
```diff
@@ -148,16 +148,18 @@ def wrapped(*args, **kwargs):
...
             # Capture exception information and re-raise it later so we can first send metrics.
             exception = ex
-            exit_code = ex.exit_code
-            if ex.wrapped_from is None:
-                exit_reason = type(ex).__name__
-            else:
+            # TODO (hawflau): review exitcode meaning. We now understand exitcode 1 as errors fixable by users.
+            exit_code = 1
```
torresxb1, line 155: Why did this change from `ex.exit_code` to `1`?

jfuss (changes requested):

samcli/lib/telemetry/metric.py
torresxb1: I think I understand now, but why did we add only these? Should we just catch all click exceptions? Or at least a few others that are here: https://click.palletsprojects.com/en/8.1.x/api/#exceptions

torresxb1 (commented):
Left a few questions and suggestions

samcli/lib/telemetry/metric.py
hawflau: Good catch. I missed `click.UsageError` and probably `click.NoSuchOption`. 

The idea of grouping these click exceptions together is that these exceptions should be fixable by users. Any user-fixable-exception that is thrown within the call stacks wrapped by `@track_command` should be caught by this line. I don't want to catch `click.ClickException` in general because that could be over-catching. I'd rather an exception to be an Unhandled Exception by default.

samcli/lib/telemetry/metric.py
hawflau: `click.BadOptionUsage` and other click exception caught here come with exitcode=2 by default. In our Telemetry, they were, however, classified as 255 (as we didn't catch any them). Today, we only have 1 and 255 exitcodes in our Telemetry. We should revisit the exitcode meaning someday. But for now, re-classifying them as exitcode 1 seems more correct to me.

samcli/lib/telemetry/metric.py
hawflau: > Or is this just for weird edge cases?
yep. it shouldn't happen at all

samcli/lib/telemetry/metric.py
torresxb1: Gotcha. What about `NoSuchOption`?

samcli/lib/telemetry/metric.py
torresxb1: But why not use whatever actual exit code is thrown, instead of only either 1 or 255?

samcli/lib/telemetry/metric.py
hawflau: I don't see any reason we should use `NoSuchOption` in our own code. So don't think we should include it.

torresxb1: Maybe I'm misunderstanding but wouldn't `NoSuchOption` be raised by `click` when a user provides an option that doesn't exist? Or are you saying that wouldn't even reach this code anyway?

hawflau: It won't reach this code. We should have all `@option` declared before `@track_command`

samcli/lib/telemetry/metric.py
hawflau: For now I don't want to introduce new exitcode (e.g. 2) into Telemetry, until we have clearer understandings about how we want to use the exitcode. Today, `click.BadOptionUsage` and `click.UsageError` (maybe a couple other user-fixable exceptions as well) go into Telemetry with exitcode 255, which is certainly incorrect. I just want to fix this first.

samcli/commands/exceptions.py
hawflau: Added a step in the "Unit Tests" GH Action to make sure Bug_report.md exists

samcli/commands/exceptions.py
hawflau: Updated the messaging

jfuss (approved):
Looks good to me. Thanks for handling the feedback.

samcli/commands/exceptions.py
```diff
@@ -25,6 +28,45 @@ def __init__(self, message, wrapped_from=None):
...
+
+        tb = "".join(traceback.format_tb(self.__traceback__))
+        click.echo(f"\nTraceback:\n{tb}", file=file, err=True)
+
+        encoded_title = quote(f"Bug: {self._command} - {type(self._exception).__name__}")
+        lookup_url = self.GH_ISSUE_SEARCH_URL.format(title=encoded_title)
+        create_issue_url = self.GH_BUG_REPORT_URL.format(title=encoded_title)
+        msg = (
+            f'An unexpected error was encountered while executing "{self._command}".\n'
+            "Search if there is any existing issue:\n"
```
sriram-mv, line 62: nit: "Search for an existing issue"

sriram-mv (approved):
Thanks! do we have an idea of what kind of use cases would cause Unhandled Exceptions today?

samcli/commands/exceptions.py
hawflau: Updated

hawflau:
@sriram-mv 

> do we have an idea of what kind of use cases would cause Unhandled Exceptions today?

There are two cases from my observation:
- Handled errors that are not raised as `UserException` or subclass of `UserException`
  - in somewhere we throw `click` exceptions (e.g. `click.BadOptionUsage`) but we did not handle it the same as `UserException`. This PR now includes certain `click` exceptions in the handling of `UserException`.
- Bugs that we did not catch

PR #4566, opened 2023-01-12
feat: send build in source parameter through to lambda builders library

torresxb1:
Lambda builders needs to be sent this parameter to know whether it should build directly in the source code directory or not.

Sends the parameter through to the lambda builders library from the build and sync commands. Note that it doesn't expose it to users yet (haven't added the click option), but it sets everything else up. Exposing it will be done later on.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


lucashuy (approved):
quick question nonblocking, could/should this be in a feature branch instead of going to develop or maybe draft if this is one of the last things that needs to be done for build in source?

hawflau (approved):

torresxb1:
@lucashuy this PR doesn't expose the option to customers yet so I think it's safe. The future PR to expose it to customers should indeed wait until the end.

PR #4580, opened 2023-01-16
fix: Fix the permission errors in Windows and add back the CVE-2007-4559 Patch changes

hnnasit:
Reverts aws/aws-sam-cli#4560

The CVE Security Patch changes were breaking the windows integrations tests due to a permission denied error while running --use-container for buildcmd tests. The issue was NamedTemporaryFile.name was used to open the file a second time and according to https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile, name cannot be used to open the file a second time while its still open on Windows.

The extract_tarfile function now takes in file_obj and file name as input parameters. Either option can be used to extract a tar_file.

Integration tests were run on Windows, MacOS, & Linux to confirm the new changes work. Functional tests are added to test extracting a tarfile using the file name. Integration tests that run the buildcmd with --use-container flag can be used for verifying the behavior for extracting a tarfile using fileobj.


tests/functional/lib/utils/test_tar.py
```diff
@@ -0,0 +1,34 @@
...
+from pathlib import Path
+import tarfile
+from tarfile import ExtractError
+
+from unittest import TestCase
+
+from samcli.lib.utils.tar import extract_tarfile
+
+
+class TestExtractTarFile(TestCase):
```
hawflau, line 14: Can we also add a test to call `extract_tarfile` with the `file_obj` kw arg?

hawflau (changes requested):

hawflau (approved):

qingchm (approved):
LGTM!

Issue #2813, opened 2021-04-15
Add Show Output Functionality to SAM CLI


amcquistan:
### Describe your idea/feature/enhancement

I have often thought it would be nice to be able to use the SAM CLI to reshow the output from my SAM/CloudFormation stack to lookup things like the HTTP endpoint to a Api Gateway resource to testing or sharing with other developers / end users.

For example, I can spin up the pre-canned hello world Python template like so.

```
samdev init --name sam-show-output-poc \
  --runtime python3.6 \
  --app-template hello-world \
  --package-type Zip
```

Then build & deploy and I get this very nice output for values specified in the output section.

```
cd sam-show-output-poc
samdev build --use-container
samdev deploy --stack-name sam-show-output-poc --region us-east-2 --guided
```

I'll omit / skip to only the output I'm interested in for this explanation.

```
CloudFormation outputs from deployed stack
-----------------------------------------------------------------------------------------------------------------------------
Outputs
-----------------------------------------------------------------------------------------------------------------------------
Key                 HelloWorldFunctionIamRole
Description         Implicit IAM Role created for Hello World function
Value               arn:aws:iam::accountid:role/sam-show-output-poc-HelloWorldFunctionRole-somesuffix

Key                 HelloWorldApi
Description         API Gateway endpoint URL for Prod stage for Hello World function
Value               https://6s4m40aq2l.execute-api.us-east-2.amazonaws.com/Prod/hello/

Key                 HelloWorldFunction
Description         Hello World Lambda Function ARN
Value               arn:aws:lambda:us-east-2:accountid:role:function:sam-show-output-poc-HelloWorldFunction-somesuffix
-----------------------------------------------------------------------------------------------------------------------------

Successfully created/updated stack - sam-show-output-poc in us-east-2
```

if I ever want to easily look up the HelloWorldApi URL I must login to the AWS console and click my into the CloudFormation views and then to the output tab of the "sam-show-output-poc" stack to see this again.

Instead, I'd like to be able to do the following to get the output section of my CF stack.

```
samdev show --stack-name sam-show-output-poc
```

### Proposal

I am proposing adding some variant of a `sam show` or perhaps `sam output` command that will show the stack output section from the CLI. The majority of the functionality for this is already present in the code base just need a little refactoring and reorganizing to expose it in a reusable way for the existing `sam deploy` command and shared use in this new `sam show` functionality.

### Additional Details

[...]

PR #4587, opened 2023-01-17
feat: List Command

mildaniel:

This is a new command to give some visibility into the local and deployed CloudFormation stack. This command is broken up into three sub-commands:
1. stack-outputs -> lists the stack outputs from the deployed stack
2. resources -> shows the resources in the local stack. If a stack name is provided, also shows the corresponding deployed resources and the logical ID to physical ID mapping.
3. endpoints -> shows the call-able APIGW endpoints and function URLs

The output can be deployed using tables or with json using the `--output` flag.

Kudos to @andrew-zhan139!

#### Issues Addressed
#2813

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/cli/command.py
```diff
@@ -25,6 +25,7 @@
     "samcli.commands.traces",
     "samcli.commands.sync",
     "samcli.commands.pipeline.pipeline",
+    "samcli.commands.list.list",
```
sriram-mv, line 28: we have to add this to the `_SAM_CLI_COMMAND_SHORT_HELP` below to showcase the short help text.

samcli/commands/list/cli_common/list_common_context.py
```diff
@@ -0,0 +1,33 @@
+"""
+Common context class to inherit from for sam list sub-commands
+"""
+import boto3
```
sriram-mv, line 4: let's move all boto3 imports within wherever they are required. It can incur upto 200ms extra load time.

samcli/commands/list/cli_common/options.py
```diff
@@ -0,0 +1,34 @@
...
+        type=click.STRING,
+    )
+
+
+def stack_name_option(f):
+    return stack_name_click_option()(f)
+
+
+def output_click_option():
+    return click.option(
```
sriram-mv, line 25: we should start to standardize this across commands.

sriram-mv (changes requested):

samcli/commands/validate/validate.py
```diff
@@ -2,7 +2,6 @@
 CLI Command for Validating a SAM Template
```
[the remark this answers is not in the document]
mildaniel: I think the change here is just returning the translated template instead of just validating it. The function name was renamed accordingly.

samcli/commands/list/cli_common/options.py
mildaniel: Agreed. Probably best to do in a follow-up PR to keep the scope of this one reasonable.

samcli/lib/list/endpoints/endpoints_to_table_mapper.py
```diff
@@ -0,0 +1,65 @@
...
+            to output the data in table format
+        """
+        entry_list = []
+        for endpoint in data:
+            cloud_endpoint_furl_string = endpoint.get("CloudEndpoint", "-")
+            methods_string = "-"
+            cloud_endpoint_furl_multi_list = []
+            if isinstance(endpoint.get("CloudEndpoint", "-"), list) and endpoint.get("CloudEndpoint", []):
+                cloud_endpoint_furl_string = endpoint.get("CloudEndpoint", ["-"])[0]
+                if len(endpoint.get("CloudEndpoint", [])) > 1:
```
mndeveci, line 36: Could there be an endpoint which is actually in 1 char length (like root path `/` for instance)?
If so, we could change the default value in previous line to be empty string rather than `-`.

samcli/lib/list/endpoints/endpoints_to_table_mapper.py
```diff
@@ -0,0 +1,65 @@
...
+        table_data: Dict[Any, Any]
+            Dictionary containing the information and data needed for the table consumer
+            to output the data in table format
+        """
+        entry_list = []
+        for endpoint in data:
+            cloud_endpoint_furl_string = endpoint.get("CloudEndpoint", "-")
+            methods_string = "-"
+            cloud_endpoint_furl_multi_list = []
+            if isinstance(endpoint.get("CloudEndpoint", "-"), list) and endpoint.get("CloudEndpoint", []):
```
mndeveci, line 34: Could be useful if we had some comments about what is happening in this block.

samcli/lib/list/endpoints/endpoints_producer.py
```diff
@@ -0,0 +1,472 @@
...
+from samcli.lib.list.list_interfaces import Producer
+from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
+from samcli.lib.providers.provider import Stack
+from samcli.commands._utils.template import get_template_data
+from samcli.lib.list.endpoints.endpoints_def import EndpointsDef
+from samcli.lib.list.resources.resource_mapping_producer import ResourceMappingProducer
+from samcli.lib.utils.boto_utils import get_client_error_code
+
+LOG = logging.getLogger(__name__)
+ENDPOINT_RESOURCE_TYPES = {"AWS::Lambda::Function", "AWS::ApiGateway::RestApi", "AWS::ApiGatewayV2::Api"}
```
mndeveci, line 24: We can use the static vars in here: https://github.dev/aws/aws-sam-cli/blob/e2a6e865493bee10b4aa3258a84891f2650fe009/samcli/lib/utils/resources.py#L11 (and same goes with rest of resource type strings)

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,149 @@
...
+            if get_client_error_code(e) == "ValidationError":
+                LOG.debug("Stack with id %s does not exist", self.stack_name)
+                raise StackDoesNotExistInRegionError(stack_name=self.stack_name, region=self.region) from e
+            LOG.error("ClientError Exception : %s", str(e))
+            raise SamListUnknownClientError(msg=str(e)) from e
+        except BotoCoreError as e:
+            LOG.error("Botocore Exception : %s", str(e))
+            raise SamListUnknownBotoCoreError(msg=str(e)) from e
+
+    def get_translated_dict(self, template_file_dict: Dict[Any, Any]) -> Dict[Any, Any]:
```
mndeveci, line 79: Why we needed to translate template locally?

mildaniel: We do a local translate as a way of doing a "dry run" to show what the transformed state of the stack will look like after deploy.

samcli/lib/list/endpoints/endpoints_to_table_mapper.py
mildaniel: I don't think there should be a case where the endpoint itself is only a single character (even in the case of customer domains it would still need to be in the format of `a.b`.

samcli/commands/list/cli_common/options.py
```diff
@@ -0,0 +1,34 @@
...
+        "--output",
```
qingchm, line 26: I feel that --output as a common option is too general to have default as table, I would move this into the command if this option will not be shared with other commands. If the output option will be shared with other commands, I don't feel that table would be an ideal default option, as other commands most likely have default output as console/json

mildaniel: This is only common to list options, all of which today make sense to default to table.

qingchm: Oh I see, for list options you created a new common file to store all the list only common options. My bad I misunderstood the structure

samcli/lib/list/endpoints/endpoints_producer.py
```diff
@@ -0,0 +1,483 @@
...
+        """
+        endpoints_list = []
+        local_stack = stacks[0]
+        local_stack_resources = local_stack.resources
+        seen_endpoints = set()
+        response = self.get_resources_info()
+        response_domain_dict = get_response_domain_dict(response)
+        custom_domain_substitute_dict = get_custom_domain_substitute_list(response, stacks, response_domain_dict)
+
+        for deployed_resource in response.get(STACK_RESOURCES, {}):
```
mndeveci, line 256: Same comment as I had for other one before (apologies that I missed this). I know we have some tests to validate this but it is too overloaded when someone takes a look at this for loop for the first time. Comments or more logging might help us to deep dive into an issue.

mildaniel:
> We could also use CFN endpoint to get latest deployed template. I am not sure which one would be the best in this use case, but we have an alternative there (in case we can't resolve intrinsics or have other issues).

If the stack is yet to be deployed, we still want to do a transform to see which resources are defined in the template and would be deployed without creating a change set.

mndeveci:
> > We could also use CFN endpoint to get latest deployed template. I am not sure which one would be the best in this use case, but we have an alternative there (in case we can't resolve intrinsics or have other issues).
> 
> If the stack is yet to be deployed, we still want to do a transform to see which resources are defined in the template and would be deployed without creating a change set.

Right, but we can still go with an hybrid approach where we can check deployed template first and if there is none, we can go with local transform. I am not going to block PR for this, will let you to decide whether we need or can do this as follow-up.

mndeveci (approved):

samcli/commands/list/endpoints/command.py
```diff
@@ -0,0 +1,52 @@
...
+from samcli.lib.telemetry.metric import track_command
+from samcli.commands._utils.options import template_option_without_build
+from samcli.cli.cli_config_file import configuration_option, TomlProvider
+
+
+HELP_TEXT = """
+Get a summary of the cloud endpoints in the stack.\n
+This command will show both the cloud and local endpoints that can
+be used with sam local and sam sync. Currently the endpoint resources
+are lambda functions and API Gateway API resources.
```
sriram-mv, line 20: nit: Lambda functions

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.stack_name:
+            response = self.get_resources_info()
+            for deployed_resource in response["StackResources"]:
+                resource_data = ResourcesDef(
+                    LogicalResourceId=deployed_resource["LogicalResourceId"],
+                    PhysicalResourceId=deployed_resource["PhysicalResourceId"],
+                )
+                resources_list.append(dataclasses.asdict(resource_data))
+                seen_resources.add(deployed_resource["LogicalResourceId"])
+            for local_resource in stacks[0].resources:
```
sriram-mv, line 139: I get antsy about IndexErrors everytime I see a direct access such as [0]

sriram-mv (approved):
Looks good overall!

samcli/lib/list/resources/resource_mapping_producer.py
mildaniel: Agreed, I replaced it with a more descriptive constant instead.

samcli/commands/list/endpoints/command.py
mildaniel: Good catch

PR #4612, opened 2023-01-24
Add event tracking for sam validate --lint metrics

cdavidxu-hub:
n/a

We should track the usage of the new '--lint' parameter to compare usage with the base command

Adding event tracker to lint method

#### Mandatory Checklist

- [] Write/update unit tests
- [] Write/update integration tests
- [x] `make pr` passes
- [ ] Write documentation


qingchm (approved):
LGTM

samcli/commands/validate/validate.py
```diff
@@ -131,6 +132,7 @@ def _lint(ctx: Context, template: str) -> None:
 
     cfn_lint_logger = logging.getLogger("cfnlint")
     cfn_lint_logger.propagate = False
+    EventTracker.track_event("UsedFeature", "CFNLint")
```
mndeveci, line 135: I think it is miss on our end for not using enums but using strings here. We may update as a follow-up PR

mndeveci (commented):
Thanks for the contribution! Any tests we can add to validate this?

qingchm:
> Thanks for the contribution! Any tests we can add to validate this?

I verified for other UsedFeature features there are no tests available either, we can improve by adding integration tests for all of them

mndeveci:
> > Thanks for the contribution! Any tests we can add to validate this?
> 
> I verified for other UsedFeature features there are no tests available either, we can improve by adding integration tests for all of them

I see, but we can still verify this in a unit tests. Something like; when `_lint` method is called, `CFNLint` event must be recorded.

mndeveci (approved):

PR #4614, opened 2023-01-24
ci(release): add homebrew-core bump job

### LINKED ISSUE #4607: [Announcement] Installing/Updating AWS SAM CLI through Homebrew
With [recent change to `homebrew-core` library](https://github.com/Homebrew/homebrew-core/pull/120653), AWS SAM CLI can now be installed without requiring to tap `aws/tap`. 

But this change also causes some delays to our releases. As of now, installing AWS SAM CLI through homebrew by running `brew install aws-sam-cli` installs `v1.70.0` which was released last week.

For users who wants to use up to date version, we recommend running `brew install aws/tap/aws-sam-cli` which should install latest version which was released couple of hours ago.

In the meantime, we are discussing with Hombrew maintainers for the way forward.

chenrui333:

fixes #4607 

Automate the homebrew-core formula version bump

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

---

relates to https://github.com/Homebrew/homebrew-core/pull/120653

cc @mndeveci @torresxb1

### DIFF
--- a/.github/workflows/brew.yml
+++ b/.github/workflows/brew.yml
@@ -1,0 +1,19 @@
+name: brew
+
+on:
+  release:
+    types: published
+
+jobs:
+  homebrew:
+    name: Bump homebrew-core formula
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
+        with:
+          formula-name: aws-sam-cli
+          commit-message: |
+            {{formulaName}} {{version}}
+            Created by https://github.com/mislav/bump-homebrew-formula-action
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}


chenrui333:
### Notes

Two TODOs with this PR:
- [ ] Need to create a homebrew-core fork for which PAT token can access (https://github.com/open-feature/homebrew-core)
- [ ] Need to create a repo scope token for this `COMMITTER_TOKEN` token

.github/workflows/brew.yml
```diff
@@ -0,0 +1,21 @@
+name: brew
+
+on:
+  push:
+    tags:
+      - '*'
```
torresxb1, line 6: should we trigger on a release instead? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

.github/workflows/brew.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  homebrew:
+    name: Bump homebrew-core formula
+    runs-on: ubuntu-latest
+    if: ${{ needs.release-please.outputs.release_created }}
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
```
torresxb1, line 14: I think we would need a security review for 3rd party github actions

torresxb1 (commented):
Hi @chenrui333 , thanks so much for contributing! Left a question and a note that I need to discuss with the team because I think we might need a security review

.github/workflows/brew.yml
chenrui333: Yeah, totally, let me update.

.github/workflows/brew.yml
chenrui333: Sounds good!

chenrui333:
Thanks for the review :)

.github/workflows/brew.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  release:
+    types: published
+
+jobs:
+  homebrew:
+    name: Bump homebrew-core formula
+    runs-on: ubuntu-latest
+    if: ${{ needs.release-please.outputs.release_created }}
```
torresxb1, line 11: I think this is probably not needed anymore after the on-release trigger

chenrui333: yeah, that is right!

chenrui333:
Once this gets accepted, we can do something like [this](https://github.com/bazelbuild/homebrew-tap/pull/95) to migrate the tap to core.

.github/workflows/brew.yml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  release:
+    types: published
+
+jobs:
+  homebrew:
+    name: Bump homebrew-core formula
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
```
carlocab, line 12: You should use https://github.com/dawidd6/action-homebrew-bump-formula since `aws-sam-cli` has Python resources that need updating, which `mislav/bump-homebrew-formula-action` does not update.

mndeveci:
Thanks for your efforts. As a team, we decided not to move forward with this PR and let the community to handle it. We strongly recommend our customers to use our tap when installing/updating AWS SAM CLI (`brew install aws/tap/aws-sam-cli`) since we can control the whole experience and timely releases. Appreciate your understanding.

Issue #4512, opened 2022-12-22
[Announcement] python3.6 Runtime Deprecation 


mndeveci:
### SAM CLI will stop supporting python3.6, by January 31st, 2023.

This runtime is no longer supported by [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html), and it is on the deprecation path. Users are not able to deploy/update these functions anymore.

## What is deprecation schedule?

After January 31st 2023, SAM CLI will no longer allow customers to use deprecated runtimes within the CLI.

## Am I impacted by this deprecation?

You are impacted by this if you are still using  python3.6 runtime. The recommendation is to upgrade these runtimes to a newer version which is supported by AWS Lambda.

For more information about supported runtimes by AWS Lambda, please check https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

AWS Lambda Builders will stop support using it as a library for python3.6, see https://github.com/aws/aws-lambda-builders/issues/419 for more details.

mndeveci:
Released with [v1.73.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.73.0)

PR #4635, opened 2023-01-30
fix: remove python3.6 support

mndeveci:
#4512

[`python3.6` have reached its EOL December 2021](https://endoflife.date/python) and [AWS Lambda dropped support in August 2022](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html)

By removing `python3.6` from enums and also replacing in the tests. Also added into deprecated runtimes for error handling.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/command.py
```diff
@@ -43,7 +43,7 @@
 \b
 Supported Runtimes
 ------------------
-1. Python 3.6, 3.7, 3.8 3.9 using PIP\n
+1. Python 3.7, 3.8 3.9 using PIP\n
```
lucashuy, line 46: nit

```suggestion
1. Python 3.7, 3.8, 3.9 using PIP\n
```

lucashuy (approved):

moelasmar (approved):

hnnasit (commented):
Q: Should we also remove python3.6 from the appveyor jobs or should we do that in a separate PR? For instance appveyor-ubuntu.yml installs python3.6

mndeveci:
> Q: Should we also remove python3.6 from the appveyor jobs or should we do that in a separate PR? For instance appveyor-ubuntu.yml installs python3.6

Good point, it is better to do in this PR, let me update with that

hnnasit (approved):

Issue #4605, opened 2023-01-22
Bug: Executable zipped during Terraform hook loses execute permissions when extracted back into build artifact directory


ceptoplex:
### Description:

I'm using the new Terraform hook beta feature of SAM CLI for a project with Lambda function handlers written in Go.
The Terraform HCL configuration includes a `null_resource` to build the Go function handlers into executables which are then packaged into a .zip file.
For Go, these executables need to have the executable file permission bit or flag set before being archived into the .zip file.
When the `copy_terraform_built_artifacts.py` [extracts the .zip file](https://github.com/aws/aws-sam-cli/blob/62b37bdeeea8fc3191adffc57858ce72c59f10c6/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py#L256) back into the SAM CLI build artifact directory (which defaults to `.aws-sam/build/...`), the extracted files lose there executable permission.

It seems [that this is the expected behavior of Python's `zipfile.extractall(...)`](https://stackoverflow.com/questions/39296101/python-zipfile-removes-execute-permissions-from-binaries) method?

### Steps to reproduce:

1. `git clone https://github.com/ceptoplex/aws-sam-cli-unzip-bug-poc`
2. `cd aws-sam-cli-unzip-bug-poc`
3. `cat README.md` will show all steps necessary to reproduce (it's not much!)

### Observed result:

```
~$ sam local invoke --beta-features --hook-name terraform -e events/example.json

Experimental features are enabled for this session.
Visit the docs page to learn more about the AWS Beta terms https://aws.amazon.com/service-terms/.

Skipped prepare hook. Current application is already prepared.
Invoking example (go1.x)
Skip pulling image and use local one: public.ecr.aws/sam/emulation-go1.x:rapid-1.70.0-x86_64.

Mounting /home/***/aws-sam-cli-zip-extraction-poc/.aws-sam/build/AwsLambdaFunctionThis*** as /var/task:ro,delegated inside runtime container
START RequestId: *** Version: $LATEST
fork/exec /var/task/example: permission denied: PathError
null
END RequestId: ***
REPORT RequestId: ***	Init Duration: 0.10 ms	Duration: 5.28 ms	Billed Duration: 6 ms	Memory Size: 128 MB	Max Memory Used: 128 MB	
{"errorMessage":"fork/exec /var/task/example: permission denied","errorType":"PathError"}
```

### Expected result:

```
~$ sam local invoke --beta-features --hook-name terraform -e events/example.json

[...]

PR #4640, opened 2023-01-31
fix: fix the copy terraform script to maintain the files permissions while extracting the ZIP files

moelasmar:
https://github.com/aws/aws-sam-cli/issues/4605

The current copy terraform built artifacts script loses all the file permissions while extracting ZIP files, and this cause some issues to the customers while invoking the lambda functions locally.

The current script is using the standard ZIP library to unzip which has [this bug](https://bugs.python.org/issue15795) and switch to use our ZIP module.
I had to split our ZIP module into two modules as it contains some functions to read a file from a URL, and require some libraries that does not exist in the standard library, and so it could not be used with the copy terraform built artifacts script.

I tested generating the PyInstaller package, and tested that I can execute the sam commands on a terraform project.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/makefile_generator.py
```diff
@@ -88,6 +89,12 @@ def generate_makefile(
     )
     shutil.copy(copy_terraform_built_artifacts_script_path, output_directory_path)
 
+    # copy zip.py script into output directory
+    ZIP_UTILS_MODULE_script_path = os.path.join(
+        Path(os.path.dirname(__file__)).parent.parent.parent.parent, "local", "lambdafn", ZIP_UTILS_MODULE
```
torresxb1, line 94: the `.parent.parent.parent.parent` seems a bit weird. Is there a better way of doing this?

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
```diff
@@ -51,6 +51,14 @@
 import zipfile
 import logging
 
+try:
+    # this case will work only while executing the copy terraform command from the makefile
+    from zip import unzip  # type: ignore
+except ImportError:
+    # this way to fix the hidden import unit testing
+    from samcli.local.lambdafn.zip import unzip
```
torresxb1, line 59: why do we use something different for unit tests than for the actual use? Doesn't this remove the confidence we can have in our unit tests?

samcli/local/lambdafn/remote_files.py
```diff
@@ -0,0 +1,53 @@
...
+
+import requests
+
+from samcli.lib.utils.progressbar import progressbar
+from samcli.local.lambdafn.zip import unzip
+
+LOG = logging.getLogger(__name__)
+
+
+def unzip_from_uri(uri, layer_zip_path, unzip_output_dir, progressbar_label):
```
torresxb1, line 17: was this just moved from the other file below or did something change?

torresxb1 (approved):
left comments above, but will change to approve so I don't block

.pylintrc
```diff
@@ -9,7 +9,7 @@
 
 # Add files or directories to the ignore list. They should be base names, not
 # paths.
-ignore=compat.py, __main__.py
+ignore=compat.py, __main__.py, zip.py
```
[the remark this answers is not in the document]
moelasmar: good catch, I should remove it. Originally I added a link to the zip module to avoid the try catch while importing the zip module, but I found it was not working fine in windows. I will update it.

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
moelasmar: The issue is the script will run in  a different way than we do in our unit testing (the hidden import test case) .. let me check with Wing how can we remove it from this test case.

samcli/local/lambdafn/remote_files.py
moelasmar: yes, refactored it to a separate module.

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
hawflau: Which unit test does it break?

hawflau: my instinct is we should fix the test

samcli/local/lambdafn/remote_files.py
```diff
@@ -0,0 +1,53 @@
...
+            with progressbar(file_length, progressbar_label) as p_bar:
+                # Set the chunk size to None. Since we are streaming the request, None will allow the data to be
+                # read as it arrives in whatever size the chunks are received.
+                for data in get_request.iter_content(chunk_size=None):
+                    local_layer_file.write(data)
+                    p_bar.update(len(data))
+
+        # Forcefully set the permissions to 700 on files and directories. This is to ensure the owner
+        # of the files is the only one that can read, write, or execute the files.
+        unzip(layer_zip_path, unzip_output_dir, permission=0o700)
```
sriram-mv, line 47: we want to be very careful with directly setting permissions anywhere, whatever we do should be additive.

sriram-mv (changes requested):

samcli/local/lambdafn/remote_files.py
moelasmar: the unzipped files will be used only for local testing, and will not be packaged or deployed using sam cli, so I think there is no risk

samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
moelasmar: I skipped the `samcli.hook_packages.terraform.copy_terraform_built_artifacts` from the `test_import_should_succeed_for_a_defined_hidden_package` test case, as this script will be only executed as a standalone script and does not depend on any non standard modules

samcli/hook_packages/terraform/hooks/prepare/makefile_generator.py
moelasmar: I defined the samcli root path in a separate variable to make it more readable

samcli/hook_packages/terraform/hooks/prepare/makefile_generator.py
```diff
@@ -88,6 +89,12 @@ def generate_makefile(
     )
     shutil.copy(copy_terraform_built_artifacts_script_path, output_directory_path)
 
+    samcli_root_path = Path(os.path.dirname(__file__)).parent.parent.parent.parent
+
+    # copy zip.py script into output directory
+    ZIP_UTILS_MODULE_script_path = os.path.join(samcli_root_path, "local", "lambdafn", ZIP_UTILS_MODULE)
```
hawflau, line 95: nit: use all lowercase for variable name - https://peps.python.org/pep-0008/#function-and-variable-names

hawflau (approved):

Issue #4622, opened 2023-01-26
Feature request: ZIP_STORED instead of ZIP_DEFLATED


Manouchehri:
### Describe your idea/feature/enhancement

I'm not sure there's a huge benefit to using DEFLATE compression for things like AWS Lambda packages. They're already very small.

### Proposal

I'd like there to be a flag to disable (or choose) compression.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model) - No.

### Additional Details

https://github.com/aws/aws-sam-cli/blob/ee23953806be066ebfb72dd42e188de4866483ae/samcli/lib/package/utils.py#L238

Manouchehri:
If I were an AWS employee, I'd also consider adding zipalign later. This would probably get slightly faster Lambda startups by bypassing the zip decompression stage; you can just mmap things straight out of the zip file, like modern Android does. 

This optimization is out-of-scope for this ticket, just a bit of free advice. =)

qingchm:
Thanks for opening this feature request! We will take a look at the request and if it makes sense we will put it into our backlogs

jfuss:
See response in PR: https://github.com/aws/aws-sam-cli/pull/4644#issuecomment-1570592936

Closing as we are electing to not support this at the current time. You can provide your own Zip file within the template, which will be uploaded if you still want tight control over how the artifact is zipped.

PR #4644, opened 2023-01-31
feat: allow disabling zip compression when packaging (#4622).

Manouchehri:
https://github.com/aws/aws-sam-cli/issues/4622

Previously, DEFLATE compression was hardcoded in.

By introducing a new optional `--no-compression` flag.

Should be none, as `--no-compression` is disabled by disable.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation

### DIFF
--- a/samcli/commands/package/command.py
+++ b/samcli/commands/package/command.py
@@ -17,8 +17,13 @@ from samcli.commands._utils.options import (
     use_json_option,
     force_upload_option,
     resolve_s3_option,
+    no_compression_option,
 )
-from samcli.commands._utils.options import metadata_option, template_click_option, no_progressbar_option
+from samcli.commands._utils.options import (
+    metadata_option,
+    template_click_option,
+    no_progressbar_option,
+)
 from samcli.lib.utils.resources import resources_generator
 from samcli.lib.bootstrap.bootstrap import manage_stack
 from samcli.lib.telemetry.metric import track_command, track_template_warnings
@@ -74,6 +79,7 @@ The following resources and their property locations are supported.
 @metadata_option
 @signing_profiles_option
 @no_progressbar_option
+@no_compression_option
 @common_options
 @aws_creds_options
 @image_repository_validation
@@ -97,6 +103,7 @@ def cli(
     no_progressbar,
     metadata,
     signing_profiles,
+    no_compression,
     resolve_s3,
     config_file,
     config_env,
@@ -119,6 +126,7 @@ def cli(
         no_progressbar,
         metadata,
         signing_profiles,
+        no_compression,
         ctx.region,
         ctx.profile,
         resolve_s3,
@@ -138,6 +146,7 @@ def do_cli(
     no_progressbar,
     metadata,
     signing_profiles,
+    no_compression,
     region,
     profile,
     resolve_s3,
@@ -169,5 +178,6 @@ def do_cli(
         region=region,
         profile=profile,
         signing_profiles=signing_profiles,
+        no_compression=no_compression,
     ) as package_context:
         package_context.run()
--- a/samcli/lib/package/utils.py
+++ b/samcli/lib/package/utils.py
@@ -125,6 +125,7 @@ def upload_local_artifacts(
     parent_dir: str,
     uploader: S3Uploader,
     extension: Optional[str] = None,
+    no_compression: Optional[bool] = False,
 ) -> str:
     """
     Upload local artifacts referenced by the property at given resource and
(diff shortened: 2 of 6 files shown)


Manouchehri:
In my _opinion_ compression should be disabled by default for Lambdas (since they have to get unpacked anyway), but I don't think I'm the person who can be making that call.

Manouchehri:
Could I have some help with the linter on this one? =)

samcli/lib/package/utils.py
```diff
@@ -166,9 +168,14 @@ def upload_local_artifacts(
 
     local_path = make_abs_path(parent_dir, local_path)
 
+    compression_type = zipfile.ZIP_DEFLATED
+    if no_compression[0]:
```
mildaniel, line 172: Why is this condition checking for the 0 index of no_compression? It's typed as a bool in the function definition but turned into a tuple earlier in the control flow. I'm not completely following the logic here.

samcli/commands/package/command.py
```diff
@@ -169,5 +178,6 @@ def do_cli(
         region=region,
         profile=profile,
         signing_profiles=signing_profiles,
+        no_compression=no_compression,
```
mildaniel, line 181: The `PackageContext` is used in other commands as well and those calls need to be updated too. This is likely why the linter is complaining.

mildaniel (changes requested):
Overall, the pull request looks good. Thanks for the contribution!

Just a few points:
1. I wonder if it's better to make this property more generic from the get-go to accept a compression type, where no compression is a potential option instead of having to generalize later with a new option.
2. Let's keep the default as is to prevent breaking anyone.
3. Can you add some unit and integration tests too to validate the logic.

I'll bring this up to the team too to discuss the best path forward here.

samcli/lib/package/utils.py
Manouchehri: Turning it into a tuple was unintentional, wasn't sure how to make it a bool instead. =(

Manouchehri:
> * I wonder if it's better to make this property more generic from the get-go to accept a compression type, where no compression is a potential option instead of having to generalize later with a new option.

You're not wrong. =) I thought about this too. The reason I did not, is because I'm not sure what decompression optimizations AWS might be doing internally, and I don't want to break anything by allowing users to introduce a possibly unsupported compression method. (I have yet to see a tool that doesn't correctly extract uncompressed ZIP files, but I have seen some that fail at things like LZMA.)

> * Let's keep the default as is to prevent breaking anyone.

Agreed. (I think a nice future feature improvement would be only compressing large files by default.)

> * Can you add some unit and integration tests too to validate the logic.

Sure, but let's fix the linter fix? =)

samcli/commands/package/command.py
Manouchehri: I'm very new to AWS SAM (just started using it in the past week); do those other commands actually zip things?

mildaniel: Yes, in the sense that `sam package` is used under the hood for other commands like `sam deploy` and `sam sync`.

sriram-mv:
Thanks for the PR! First time contributor yay! 🎉 

Few things that we as a team need to check before going forward with the PR.

* History behind using DEFLATE and why was it chosen? This is on us.
* The issue mentions that the zip packages are generally small, however the limit is upto 250MB and it could be different per runtime. This requires more clarity. What kind of runtimes are you using and where do you see a perceived speed up?
* Checking with AWS Lambda team on what are the required attributes for the zip file that `aws-sam-cli` creates.

Manouchehri:
> History behind using DEFLATE and why was it chosen? This is on us.

It's not the _worst_ idea in the world. But it's also not a great idea either. (The main reason I noticed it, was because I've read the ZIP specification for more hours than I'd like to admit, and spent a lot of time working with APKs.)

There's a reason why the Lambda team enforces a limit on the uncompressed size: because that's what they have to extract, run, and cache. (If I was a fly on the wall, I'd take a guess that ZIP compression is supported on Lambda packages for developer convenience, not performance.)

> The issue mentions that the zip packages are generally small, however the limit is upto 250MB and it could be different per runtime. This requires more clarity. 

The limit is 250MB unzipped on the AWS server-side of things, this doesn't change that; the size that SAM compresses the ZIP down to on the client-side doesn't impact the limit. https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html

> What kind of runtimes are you using and where do you see a perceived speed up?

This patch speeds up the ZIP packaging on the client side, regardless of runtime. (So I guess the answer is sorta all runtimes? =P ) 

(If zipalign is added too later, you _could_ get a performance speed up on the server side too by mmap'ing files instead of extracting them. This would a project for another AWS employee though, not me.)

> Checking with AWS Lambda team on what are the required attributes for the zip file that aws-sam-cli creates.

Uncompressed definitely works (which is all this patch allows), I've tried it. =)

Issue #4304, opened 2022-10-14
Bug: Sam sync adds an extra nodejs directory to the path of the lambda layer


RobertMarton1985:
### Description:
We have a sam project with lambda functions and layers written in nodejs14.x.
We are currently working on using sam sync to speed up development. When we start the project with sam sync it starts properly but the lambdas fail to use the functions provided in the layers.
We started investigating the issue and downloaded the layer's code with the following command:
aws lambda get-layer-version --layer-name #LAYER_NAME# --version-number #LAYER_VERSION# --query Content.Location --output text

It turned out that when we use sam sync it adds an extra nodejs directory into the path of the layer.
Layer content when deployed with sam deploy:
layer.zip/nodejs/node_modules/...

Layer content when deployed with sam sync:
layer.zip/nodejs/nodejs/node_modules/...

### Steps to reproduce:
Create a layer with AWS::Serverless::LayerVersion and set the ContentUri to a local folder

### Observed result:
sam sync adds an extra nodejs directory to the path of a nodejs14.x lambda layer

### Expected result:
sam sync should create the same layer as sam deploy.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10
2. `sam --version`: SAM CLI, version 1.60.0
3. AWS region: eu-west-1

lucashuy:
Hi thanks for raising this issue! Can you provide snippet of your template with the `AWS::Serverless::LayerVersion` resource while we investigate?

RobertMarton1985:
This is the problematic Layer:

```yaml
  DummyLayer:
    Type: AWS::Serverless::LayerVersion
    Properties:
      LayerName: dummy-layer
      Description: Dummy functions
      ContentUri: src/common/dummy/
    Metadata:
      BuildMethod: nodejs14.x
```

We don't build the layer, but if the Metadata part is not there then sam sync throws an error even though it should be optional:

`
Syncing Layer DummyLayer...
Layer <samcli.lib.providers.sam_function_provider.SamFunctionProvider object at 0x000002026967FD90> is missing BuildMethod Metadata.
Code sync encountered an error.
Traceback (most recent call last):
  File "C:\Program Files\Amazon\AWSSAMCLI\runtime\lib\site-packages\samcli\lib\sync\sync_flow_executor.py", line 342, in _sync_flow_execute_wrapper
    dependent_sync_flows = sync_flow.execute()
[...]

PR #4651, opened 2023-02-01
fix: handle special cases for layer sync

mndeveci:
#4304

If a layer resource doesn't have `BuildMethod` or marked with `SkipBuild: True`, SAM CLI should skip build step and directly zip contents of the target folder. 

By checking layer details and assigning different sync flows depending on the layer details.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/flows/layer_sync_flow.py
```diff
@@ -258,6 +259,30 @@ def _get_dependent_functions(self) -> List[Function]:
         return dependent_functions
 
 
+class LayerSyncFlowSkipBuildDirectory(LayerSyncFlow):
+    """
+    LayerSyncFlow special implementation that will skip build step and zip contents of CodeUri
+    """
+
+    def gather_resources(self) -> None:
+        zip_file_path = os.path.join(tempfile.gettempdir(), f"data-{uuid.uuid4().hex}")
```
mildaniel, line 268: nit: Should we try use `Pathlib` where we can instead of `os.path`?

samcli/lib/sync/sync_flow_factory.py
```diff
@@ -170,8 +175,26 @@ def _create_lambda_flow(
...
+        self, resource_identifier: ResourceIdentifier, resource: Dict[str, Any]
+    ) -> Optional[SyncFlow]:
+        layer = self._build_context.layer_provider.get(str(resource_identifier))
+        if not layer:
+            LOG.warning("Can't find layer resource with %s logical id", str(resource_identifier))
+            return None
+        layer_sync_flow_type = LayerSyncFlow
+
+        if not layer.build_method or layer.skip_build:
+            LOG.debug("Creating LayerSyncFlowNoBuildDirectory for %s resource", resource_identifier)
```
mildaniel, line 188: nit: Should this be moved inside the condition below?

mildaniel (approved):

samcli/lib/sync/flows/layer_sync_flow.py
mndeveci: We could, but I just wanted to follow what we already have in this file. We could do it in another PR for refactoring purposes.

samcli/commands/build/build_context.py
```diff
@@ -591,7 +591,7 @@ def _collect_single_function_and_dependent_layers(
             return
 
         resource_collector.add_function(function)
-        resource_collector.add_layers([l for l in function.layers if l.build_method is not None and not l.skip_build])
+        resource_collector.add_layers([l for l in function.layers if BuildContext.is_layer_buildable(l)])
```
jfuss, line 594: Interesting. I wonder if there was a specific reason for this. Also looks like the `function` part doesn't filter. I assume that is likely a bug? If so, just want to note it for us to fix later (outside this pr).

samcli/lib/sync/flows/layer_sync_flow.py
```diff
@@ -258,6 +259,30 @@ def _get_dependent_functions(self) -> List[Function]:
...
+        self._local_sha = file_checksum(cast(str, self._zip_file), hashlib.sha256())
+
+
+class LayerSyncFlowSkipBuildZipFile(LayerSyncFlow):
+    """
+    LayerSyncFlow special implementation, that will skip build and upload zip file which is defined in CodeUri directly
+    """
+
+    def gather_resources(self) -> None:
+        self._zip_file = os.path.join(tempfile.gettempdir(), f"data-{uuid.uuid4().hex}")
```
jfuss, line 280: How are we cleaning up these temp dirs for this?

samcli/lib/sync/sync_flow_factory.py
```diff
@@ -170,14 +175,45 @@ def _create_lambda_flow(
...
+                str(resource_identifier),
+                self._build_context,
+                self._deploy_context,
+                self._physical_id_mapping,
+                self._stacks,
+            )
+
+        if is_zip_file(layer.codeuri):
+            LOG.debug("Creating LayerSyncFlowSkipBuildZipFile for '%s' resource", resource_identifier)
+            return LayerSyncFlowSkipBuildZipFile(
```
jfuss, line 207: Nit: This is how I read the name, which led to some confusion on my end. LayerSyncFlow + SkipBuild + Zip + File. Reading it this way, made me think we were zipping the file/folder but doesn't seem like that is the ended thing. Maybe we should rename or add in some "_" to signal to the reader the right thing from the name?

jfuss (changes requested):
No integration tests?

samcli/lib/sync/flows/layer_sync_flow.py
mndeveci: It is cleaned up here in parent class: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/sync/flows/layer_sync_flow.py#L98

samcli/commands/build/build_context.py
mndeveci: I am not sure. Existing logic was pretty same with `is_layer_buildable`. Only difference is, `is_layer_buildable` additionally checks if `codeuri` of the layer points to a ZIP file or not.

For function collection part, even if function is set `SkipBuild: True`, then running `sam build {logical-id}` will build the function. But if you run `sam build` to build entire stack, it will skip that function to build. So it feels like this is not the expected behavior.

samcli/lib/sync/sync_flow_factory.py
mndeveci: Naming is hard and I am also not very happy with these names, open for any suggestion.

jfuss (approved):

PR #4654, opened 2023-02-02
feat: Collect Lambda Authorizers under the Auth property for Serverless resources

lucashuy:
This change updates the SAM API provider logic to add and collect Lambda Authorizers that are defined under the `Auth` and `Authorizers` section in the properties of Serverless resources.

Updates existing logic to also search in the above keys for `AWS::Serverless::Api`, `AWS::Serverless::HttpApi`, and `AWS::Serverless::Function`

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
         collector.stage_variables = stage_variables
         collector.cors = cors
 
+        auth = properties.get(SamApiProvider._AUTH, {})
+        if auth:
+            if auth.get(SamApiProvider._DEFAULT_AUTHORIZER):
+                collector.set_default_authorizer(logical_id, auth.get(SamApiProvider._DEFAULT_AUTHORIZER))
+
+            self._extract_authorizers_from_props(logical_id, auth, collector, Route.API)
```
mndeveci, line 123: nit: early exit might improve the readability (getting rid of one of the nested if statements). There is also one below with very similar logic, we can apply there as well if we decide to.
```suggestion
        auth = properties.get(SamApiProvider._AUTH, {})
        if not auth:
            return
        
        if auth.get(SamApiProvider._DEFAULT_AUTHORIZER):
            collector.set_default_authorizer(logical_id, auth.get(SamApiProvider._DEFAULT_AUTHORIZER))

        self._extract_authorizers_from_props(logical_id, auth, collector, Route.API)
```

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
...
+                # but the V2 still expects "$"
+                prefix = HTTP_IDENTITY_SOURCE_PREFIX if event_type == Route.HTTP else ""
+
+                for context in identity_object.get(SamApiProvider._IDENTITY_CONTEXT, []):
+                    identity_sources.append(f"{prefix}context.{context}")
+
+                for stage_variable in identity_object.get(SamApiProvider._IDENTITY_STAGE, []):
+                    identity_sources.append(f"{prefix}stageVariables.{stage_variable}")
+
+                payload_version = identity_object.get(SamApiProvider._AUTHORIZER_PAYLOAD, "1.0")
```
mndeveci, line 182: nit: I saw V1 was added as constant in this or previous PR, can we use that?

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
         collector.stage_variables = stage_variables
         collector.cors = cors
 
+        auth = properties.get(SamApiProvider._AUTH, {})
+        if auth:
+            if auth.get(SamApiProvider._DEFAULT_AUTHORIZER):
```
moelasmar, line 120: why not to make it in one if statement?
```
if auth and auth.get(SamApiProvider._DEFAULT_AUTHORIZER):
```

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
...
+        authorizers: Dict[str, Authorizer] = {}
+
+        for auth_name, auth_props in auth.get(SamApiProvider._AUTHORIZERS, {}).items():
+            authorizer_type = auth_props.get(SamApiProvider._FUNCTION_TYPE, LambdaAuthorizer.TOKEN).lower()
+            identity_object = auth_props.get(SamApiProvider._IDENTITY, {})
+
+            function_arn = auth_props.get(SamApiProvider._FUNCTION_ARN)
+
+            if not function_arn:
+                LOG.info("Authorizer '%s' is currently unsupported (must be a Lambda Authorizer), skipping", auth_name)
```
moelasmar, line 154: Is it appropriate to change the log statement from an info log to a debug log? My concern is that if the project has multiple Lambda functions, the log output could become overwhelming.

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
...
+
+            if not function_name:
+                LOG.warning("Unable to parse the Lambda ARN for Authorizer '%s', skipping", auth_name)
+                continue
+
+            if authorizer_type == LambdaAuthorizer.REQUEST or event_type == Route.HTTP:
+                identity_sources = []
+
+                for query_string in identity_object.get(SamApiProvider._IDENTITY_QUERY, []):
+                    identity_sources.append(f"{prefix}request.querystring.{query_string}")
```
moelasmar, line 167: in my opinion, the following way is more readable. 
```
identity_sources = [f"{prefix}request.querystring.{query_string}" for query_string in identity_object.get(SamApiProvider._IDENTITY_QUERY, [])]
```

samcli/lib/providers/sam_api_provider.py
lucashuy: Yup, the constant was added in a previous PR but I can copy that over here/wait for the previous one to go in.

samcli/lib/providers/sam_api_provider.py
lucashuy: Yeah, the same concern/questions were brought up in a previously merged in PR about too many/too few log statements. One idea we could do, is to change these over to debug statements and add some logic to just do a single info log at the end to let them know there are unsupported authorizers in the template.

samcli/lib/providers/sam_api_provider.py
```diff
@@ -143,6 +254,13 @@ def _extract_from_serverless_http(
         collector.stage_variables = stage_variables
         collector.cors = cors
 
+        auth = properties.get(SamApiProvider._AUTH, {})
+        if auth:
+            if auth.get(SamApiProvider._DEFAULT_AUTHORIZER):
```
moelasmar, line 259: Please combine these in one if condition.

moelasmar (approved):
it looks good to me with some minor comments.

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +115,100 @@ def _extract_from_serverless_api(
...
+                LOG.info("Authorizer '%s' is currently unsupported (must be a Lambda Authorizer), skipping", auth_name)
+                continue
+
+            function_name = LambdaUri.get_function_name(function_arn)
+
+            if not function_name:
+                LOG.warning("Unable to parse the Lambda ARN for Authorizer '%s', skipping", auth_name)
+                continue
+
+            if authorizer_type == LambdaAuthorizer.REQUEST or event_type == Route.HTTP:
```
moelasmar, line 163: can we extract the if/else bodies to functions to increase the readability

samcli/lib/providers/sam_api_provider.py
```diff
@@ -98,6 +117,179 @@ def _extract_from_serverless_api(
...
+            The type of API this is (API or HTTP API)
+
+        Returns
+        -------
+        LambdaAuthorizer
+            The request based Lambda Authorizer object
+        """
+        payload_version = properties.get(SamApiProvider._AUTHORIZER_PAYLOAD)
+
+        if payload_version is not None and not isinstance(payload_version, str):
```
sriram-mv, line 157: I wonder for validation overall, for this PR and other PRs, it might be easier if we have a validation decorator with a certain set of rules that we can attach to a given function. This gives a cleaner delineation between validation logic and the business logic.

Looks to me, most of the validation is around the args themselves.

sriram-mv (approved):
Approving.

1/ Being thoughtful about logging is important as its our signal to users on what's happening on the process.
2/ Validation logic could be made as a decorator with `x` rules as input, I like that its a separate class in the other PR.

samcli/lib/providers/sam_api_provider.py
lucashuy: Thats a good callout about validation vs business logic. Its not ideal since we in each instance (authorizers under `Properties`, as a CFN resource, under Swagger) very slight variations in the keys and some slight variation in the expected type/format.

Moving this validation logic out is probably best done once that other PR is merged in to reuse that class.

samcli/lib/providers/sam_api_provider.py
lucashuy: Opted to change some of the LOGs (leaving the one about not being to parse ARN) to debugs. Since we also have an authorizer linking step, the user will be alerted there if someone is wrong, so debug output in this PR is suitable.

PR #4655, opened 2023-02-02
fix: file/directory permissions within the zip while packaging artifacts

sriram-mv:
* Additive permissions to add execute on directories and read on files.

* Re-adds following commits https://github.com/aws/aws-sam-cli/pull/4488 , https://github.com/aws/aws-sam-cli/pull/4462 and https://github.com/aws/aws-sam-cli/pull/4485

* This change makes the permissions additive on files and directories.

* Only adds permissions and does not take away existing ones.

* N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss (approved):
Probably out of scope for this PR but I think it would be a good idea to add integration tests for a compiled and interpreted language that package, deploy, and then invoke the function. I will leave that up to you but at the very least, I think we should add that before this is released.

qingchm (approved):
LGTM

samcli/lib/package/permissions.py
```diff
@@ -0,0 +1,153 @@
...
+
+        Returns
+        -------
+        zipfile.ZipInfo
+            modified object with `external_attr` set to member permissions.
+
+        """
+        if platform.system().lower() != "windows":
+            return zip_info
+        zip_info.external_attr = self.permissions << 16 if not zip_info.is_dir() else zip_info.external_attr
```
mndeveci, line 41: nit: inline if statement can be moved up for skipping no-op assignment here.

samcli/lib/package/utils.py
```diff
@@ -242,28 +263,55 @@ def make_zip(file_name, source_root):
...
+                    with open(full_path, "rb") as data:
+                        file_bytes = data.read()
+                        info = zipfile.ZipInfo(relative_path)
+                        # Context: Nov 2020
+                        # Set external attr with Unix 0755 permission
+                        # Originally set to 0005 in the discussion below
+                        # https://github.com/aws/aws-sam-cli/pull/2193#discussion_r513110608
+                        # Changed to 0755 due to a regression in https://github.com/aws/aws-sam-cli/issues/2344
+                        # Final PR: https://github.com/aws/aws-sam-cli/pull/2356/files
+                        if any(permission_mappers):
```
mndeveci, line 275: We could get rid of this if statement by checking permission_mappers in the beginning of method. Something like; `permission_mappers = permission_mappers or []`.

samcli/lib/package/utils.py
```diff
@@ -242,28 +263,55 @@ def make_zip(file_name, source_root):
...
+)
+# Context: Jan 2023
+# NOTE(sriram-mv): Modify permissions regardless of the Operating system
+# to add 111 for directories and 444 for files in addition to existing permissions.
+# No overriding explicit permissions are set.
+# Extended Attributes are preserved.
+make_zip_with_lambda_permissions = functools.partial(
+    make_zip_with_permissions,
+    permission_mappers=[
+        WindowsFilePermissionMapper(permissions=0o100755) if platform.system().lower() == "windows" else None,
```
mndeveci, line 307: If we are adding the `WindowsFilePermissionMapper` by checking the OS first, then if statement in the mapper itself seems redundant. Can we get rid of one to make it easy to read?

And also, I would recommend not adding `None` into array (with also supporting it type checking) so that during the execution of these mapper would be simpler, we won't need to have the `if` statement above when we are running the mapper itself.

mndeveci (commented):
Thanks for the change! I've requested some small changes to make it more readable. I also agree for adding an integration test to verify this behavior which would also help us to not break this in the future.

samcli/lib/package/utils.py
sriram-mv: Agree, I was refactoring on the fly, and it can be made better.

mndeveci (approved):

Issue #4649, opened 2023-02-01
Bug: sam validate --lint always return 0, even on failure.


andyloughran:
### Description:
When running sam validate --lint, it always exits with 0 on linting failure.

### Steps to reproduce:

- Start with a perfect template.
- Run `sam validate --lint` and see a success.
- Run `echo $?` to get the return code, it shows 0.
- Create a deliberate linting mistake in your template (such as adding a conflicting property to a resource)
- Run `sam validate --lint` and see a failure.
- Run `echo $?` and see the return code is still 0.

### Observed result:

A return code of 0 on a linting failure

### Expected result:
A return code of greater than 0 for a linting failure, to demonstrate error.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

MacOS

```{
  "version": "1.70.0",
  "system": {
    "python": "3.11.1",
    "os": "macOS-13.2-arm64-arm-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.22",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  }
}```

mildaniel:
Thanks, @andyloughran. Let me try to reproduce and look further into this issue.

mildaniel:
I was able to reproduce and confirm the behaviour. I will raise a PR shortly to address this issue.

PR #4657, opened 2023-02-02
fix: Return non-zero exit code on lint failures

mildaniel:
#4649

Currently, `sam validate --lint` returns a 0 return code even if lint errors occurred.

Raises an exception if any lint rules matched, which will propagate up and will exit the process with a return code of 1. 

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/validate/validate.py
+++ b/samcli/commands/validate/validate.py
@@ -12,6 +12,7 @@ from samcli.cli.context import Context
 from samcli.cli.main import pass_context, common_options as cli_framework_options, aws_creds_options, print_cmdline_args
 from samcli.commands._utils.cdk_support_decorators import unsupported_command_cdk
 from samcli.commands._utils.options import template_option_without_build
+from samcli.commands.exceptions import LinterRuleMatchedException
 from samcli.lib.telemetry.event import EventTracker
 from samcli.lib.telemetry.metric import track_command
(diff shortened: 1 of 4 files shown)


samcli/commands/validate/validate.py
```diff
@@ -146,12 +147,18 @@ def _lint(ctx: Context, template: str) -> None:
...
             click.secho("{} is a valid SAM Template".format(template), fg="green")
+            return
+
         rules = cfnlint.core.get_used_rules()
         matches_output = formatter.print_matches(matches, rules, filenames)
 
         if matches_output:
             click.secho(matches_output)
 
+        raise LinterRuleMatchedException(
```
sriram-mv, line 158: If there are 0 matches, do we know that we are in exit code 1 state by default?

mildaniel: The opposite actually. Line 148 checks for any matches and logs a success message if none are found
```
if not matches:
     click.secho("{} is a valid SAM Template".format(template), fg="green")
```
Otherwise there are some rules matched (meaning the linter found some issues) and we're in exit code 1 state.

sriram-mv (approved):

moelasmar (approved):

Issue #4658, opened 2023-02-03
Bug: sam list stack-outputs - KeyError


Tak1wa:
### Description:

sam list is a nice feature.
I was trying sam list stack-outputs on various stacks.
I noticed an error in some of the CloudFormation stacks.

### Steps to reproduce:

- Follow the steps below to deploy the base stack to use the AWS Toolkit for .NET Refactoring.
  - https://docs.aws.amazon.com/tk-dotnet-refactoring/latest/userguide/roles-and-policies.html
- Run `sam list stack-outputs --stack-name refactoringtoolkit`

### Observed result:

```
iwasa.takahito@xxx hoge0204sam-dotnet-serverless % sam list stack-outputs --stack-name refactoringtoolkit --d
ebug
2023-02-04 06:55:14,894 | Config file location: /Users/iwasa.takahito/work/hoge0204sam/hoge0204sam-dotnet-serverless/samconfig.toml
2023-02-04 06:55:14,894 | Config file '/Users/iwasa.takahito/work/hoge0204sam/hoge0204sam-dotnet-serverless/samconfig.toml' does not exist
2023-02-04 06:55:14,922 | Using config file: samconfig.toml, config environment: default
2023-02-04 06:55:14,922 | Expand command line arguments to:
2023-02-04 06:55:14,922 | --stack_name=refactoringtoolkit --output=table 
2023-02-04 06:55:15,268 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2023-02-04 06:55:15,268 | Template is not provided in context, skip adding project type metric
[...]

PR #4659, opened 2023-02-04
fix: Make stack outputs description optional

mildaniel:

#4658

Don't assume CFN stack outputs will have a description since it's optional.

Use dictionary `.get()` method with default value where applicable to avoid getting a `KeyError` when a value doesn't exist.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


jfuss (approved):

samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -132,11 +132,11 @@ def produce(self):
...
             for deployed_resource in response["StackResources"]:
                 resource_data = ResourcesDef(
-                    LogicalResourceId=deployed_resource["LogicalResourceId"],
-                    PhysicalResourceId=deployed_resource["PhysicalResourceId"],
+                    LogicalResourceId=deployed_resource.get("LogicalResourceId"),
+                    PhysicalResourceId=deployed_resource.get("PhysicalResourceId"),
                 )
                 resources_list.append(dataclasses.asdict(resource_data))
-                seen_resources.add(deployed_resource["LogicalResourceId"])
+                seen_resources.add(deployed_resource.get("LogicalResourceId"))
```
hnnasit, line 139: If `deployed_resource.get("LogicalResourceId")` returns None, None gets added to the seen_resources set. Should we avoid adding None here and check it with an if condition instead?

hnnasit (changes requested):

samcli/lib/list/resources/resource_mapping_producer.py
mildaniel: We could to be extra safe, but in this case if a resource is already deployed it must have a logical ID and a Physical ID, so we should be okay.

hnnasit: Right makes sense, in that case should we keep the [] brackets for getting the LogicalResourceId and PhysicalResourceId if these keys are always assumed to be present?

mildaniel: I do think it's still better practice to use `.get()` as it's a safer operation here. If downstream there's an issue we can deal with it there, but at least we won't be getting a `KeyError` here.

hnnasit: Fair enough

hnnasit (approved):

Issue #4453, opened 2022-12-01
Feature request: Support dotnet lambda container builds


1davidmichael:
### Describe your idea/feature/enhancement

All other language run-times support the `--use-container` option for `sam build` except dotnet runtimes. With the addition of [build images for dotnet](https://github.com/aws/aws-sam-build-images/pull/25) it should now be possible to do this.

I've tested this with a [quick hack here](https://github.com/aws/aws-sam-cli/commit/34a3fd8859759ed7e8bec568276fee97bc2b5b03) and it appears to work. I was able to successfully build and publish a `dotnet6` runtime lambda via the following commands:

```bash
sam build -u
sam deploy --config-file samconfig.toml
```

However, I am guessing the change from read-only to read-write for the manifest dir is not ideal.

### Proposal

Add support for dotnet lambda builds via docker containers to align with other runtime supported builds.

Things to consider:
1. May require updates to the docker build logic to support how it is handled via dotnet

torresxb1:
Thanks for the feature request! Write permissions would indeed require some caution here. Let me bring this up to the team for discussion.

torresxb1:
Hi @1davidmichael, wanted to give an update after discussing with the team: this is something that was on our map already, and we're in the process of prioritizing it, but we can't give any timelines at the moment.

1davidmichael:
@torresxb1 awesome to here. Thanks!

PR #4665, opened 2023-02-07
feat: support dotnet lambda container builds

mrkdeng:

#4453

- This change adds support for .NET lambda container builds. It allows mounting source directory with write permissions when building in container because .NET builds write output to the source directory.
- In most cases on posix system, Docker Daemon is running as root. This change solves this problem by running docker as a non-root user.

- Add a new sam build CLI argument `--mount-with [READ|WRITE]` that enables mounting source code directory with READ/WRITE permissions. Then run Docker as a non-root user (unless current user is root) if mount with write, which gives user the least privilege 
- .NET lambda container builds always need `--mount-with WRITE` to be specified. If building without explicit `--mount-with`, a confirmation dialog pops up and asks write permissions to source code directory
- This change is backward compatible and not breaking since container builds for .NET was not supported. For other runtimes, they mount source code as READ only by default, unless with intently using `--mount-with WRITE`

There is permission issue when running .NET image as non-root user [#7868](https://github.com/dotnet/core/issues/7868). This change relies on docker images to set environment variables such as `DOTNET_CLI_HOME`, `XDG_DATA_HOME`

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


samcli/local/docker/lambda_build_container.py
```diff
@@ -104,8 +130,23 @@ def __init__(  # pylint: disable=too-many-locals
...
             entrypoint=entry,
             env_vars=env_vars,
+            mount_with_write=mount_with_write,
         )
 
+    def make_host_tmp_dirs(self):
+        """
+        Make tmp dirs on host
+        """
+        if self._host_tmp_base_dir and not os.path.exists(self._host_tmp_base_dir):
```
Beau-Gosse-dev, line 140: So we won't create it if it already exists, but then we'll delete it afterwards? Can we not use a common path like /tmp/samcli ? A user might actually use that for something. Can we just use a GUID or something for the path? This is on the host machine, right? It doesn't have to match the path inside the container, right?

Beau-Gosse-dev (changes requested):

samcli/local/docker/container.py
```diff
@@ -135,15 +139,15 @@ def create(self):
         _volumes = {}
 
         if self._host_dir:
-            LOG.info("Mounting %s as %s:ro,delegated inside runtime container", self._host_dir, self._working_dir)
+            mount_mode = "rw" if self._mount_with_write else "ro,delegated"
```
jfuss, line 142: Should this still have `,delegated` on the `rw`?

samcli/local/docker/container.py
```diff
@@ -157,6 +161,16 @@ def create(self):
...
 
+        # Get effective user when mounting with write permissions
+        if self._mount_with_write:
+            effective_user = get_posix_effective_user()
+            if effective_user:
+                user_id = effective_user["user_id"]
+                # skip if user id is 0 (root)
+                if user_id != 0:
+                    group_id = effective_user["group_id"]
+                    kwargs["user"] = f"{user_id}:{group_id}" if group_id else user_id
```
jfuss, line 172: Can you explain what this is for? I am having a hard time understanding the implications here.

samcli/local/docker/lambda_build_container.py
```diff
@@ -42,7 +46,16 @@ def __init__(  # pylint: disable=too-many-locals
...
+        mount_with_write: bool = False,
     ):
+        # tmp dirs on host that will be mounted to allow write permissions to `/tmp/samcli/` on container
+        # only when `mount_with_write` is True
+        if mount_with_write:
+            self._host_tmp_dir = os.path.join(source_dir, "tmp")
+            self._host_tmp_base_dir = os.path.join(source_dir, "tmp", "samcli")
+        else:
+            self._host_tmp_dir = None
+            self._host_tmp_base_dir = None
```
jfuss, line 58:
```
self._host_temp_dir = None
self._host_tmp_base_dir = None

if mount_with_write:
            self._host_tmp_dir = os.path.join(source_dir, "tmp")
            self._host_tmp_base_dir = os.path.join(source_dir, "tmp", "samcli")
```

Is a much cleaner way to write this

samcli/local/docker/lambda_build_container.py
```diff
@@ -249,21 +295,27 @@ def _convert_to_container_dirs(host_paths_to_convert, host_to_container_path_map
...
+            The build method specified in metadata
 
         Returns
         -------
         str
             valid image name
         """
-        return f"{LambdaBuildContainer._IMAGE_URI_PREFIX}-{runtime}:" + LambdaBuildContainer.get_image_tag(architecture)
+        # a list of build methods need to get image based on build method itself instead of runtime
+        get_image_from_build_method = ["dotnet7"]
```
jfuss, line 315: Can we move this logic to before the calling this method? Seems weird to put this here and just override the runtime. Also hardcodes specific `dotnet7` things in "random" places. Can we make this more abstract to support similar things for future cases?

For example, If `build_method` is provided, we use that over the runtime directly?

tests/testing_utils.py
```diff
@@ -66,7 +66,7 @@ def run_command(command_list, cwd=None, env=None, timeout=TIMEOUT) -> CommandRes
         raise
 
 
-def run_command_with_input(command_list, stdin_input, timeout=TIMEOUT, cwd=None) -> CommandResult:
+def run_command_with_input(command_list, stdin_input, timeout=TIMEOUT, cwd=None, env=None) -> CommandResult:
```
jfuss, line 69: `env` is unused here.

jfuss (changes requested):
Did a first pass with some feedback. 

Thank you for adding integration tests but please make sure you are also adding unit tests to cover your changes.

samcli/commands/build/command.py
```diff
@@ -136,6 +136,15 @@
     help="Enabled parallel builds. Use this flag to build your AWS SAM template's functions and layers in parallel. "
     "By default the functions and layers are built in sequence",
 )
+@click.option(
+    "--mount-with-write",
```
mndeveci, line 140: For boolean options, we usually provide both values. So we can rename this to; `--mount-with-write/--mount-with-read`? Default value would be `--mount-with-read` which will resolve as `mount_with_write=False` down below.

samcli/lib/build/app_builder.py
```diff
@@ -946,6 +961,7 @@ def _build_function_on_container(
             # "/." is a Docker thing that instructions the copy command to download contents of the folder only
             result_dir_in_container = response["result"]["artifacts_dir"] + "/."
             container.copy(result_dir_in_container, artifacts_dir)
+            container.remove_host_tmp_dirs()
```
mndeveci, line 964: This should be called in finally block

samcli/local/docker/container.py
```diff
@@ -16,7 +16,7 @@
 from samcli.lib.utils.tar import extract_tarfile
 from .exceptions import ContainerNotStartableException
 
-from .utils import to_posix_path, find_free_port, NoFreePortsError
+from .utils import to_posix_path, find_free_port, NoFreePortsError, get_posix_effective_user
```
mndeveci, line 19: nit: Please use absolute imports

samcli/local/docker/utils.py
```diff
@@ -154,3 +155,22 @@ def get_docker_platform(architecture: str) -> str:
...
     return f"linux/{get_image_arch(architecture)}"
+
+def get_posix_effective_user() -> Dict[str, str]:
+    """
+    Returns the posix effective user and group id
+
+    :returns Dictionary containing user and group id, or None for non posix systems
+    """
+    if os.name.lower() == 'posix':
+        effective_user = {}
```
mndeveci, line 166: Can we use `@dataclass` instead of dictionary here? It will be easy to use in other places (no need to hardcode its keys for instance).

mndeveci (changes requested):

mrkdeng:
@Beau-Gosse-dev @jfuss @mndeveci thank you for reviewing this PR. Push two commits with one addressing comments, another one updating/adding unit tests

samcli/local/docker/lambda_build_container.py
mrkdeng: Done. I changed the name to `specified_workflow` to be more consistent with `workflow_config.py`

samcli/lib/build/workflow_config.py
```diff
@@ -258,6 +254,47 @@ def _key(c: CONFIG) -> str:
     return True, None
 
 
+def supports_specified_workflow(specified_workflow: str) -> bool:
```
jfuss, line 257: I am having an hard time understand what this is actually for. Could you describe the use-case for this?

Note: This is probably just a gap in my understanding.

samcli/local/docker/lambda_build_container.py
```diff
@@ -142,12 +162,14 @@ def _make_request(
                         "dependency_manager": dependency_manager,
                         "application_framework": application_framework,
                     },
+                    "base_dir": container_dirs["base_dir"],
```
jfuss, line 165: This json is used to call AWS Lambda Builders in an RPC mode within the container. `base_dir`, `specified_workflow`, and `mount_with_write` are not apart of that: https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/__main__.py#L117

Was this meant to be added here?

jfuss (changes requested):

samcli/lib/build/app_builder.py
```diff
@@ -7,15 +7,14 @@
...
-
 import docker
 import docker.errors
 from aws_lambda_builders import (
     RPC_PROTOCOL_VERSION as lambda_builders_protocol_version,
 )
 from aws_lambda_builders.builder import LambdaBuilder
 from aws_lambda_builders.exceptions import LambdaBuilderError
-
+from samcli.commands.build.utils import prompt_user_to_enable_mount_with_write
```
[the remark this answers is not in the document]
mrkdeng: Yes, we could do it in the `build_context`. Currently we don't prompt click until we know `_mount_with_write` is a must-have. I might be wrong - if move it to `command`, we will need to introduce the functionality in `workflow_config.get_workflow_config` to determine if `_mount_with_write` is needed. Doing so does move the UX part out of the `lib`, but not separate those components ideally.

samcli/lib/build/workflow_config.py
mrkdeng: Regarding [comment](https://github.com/aws/aws-sam-cli/pull/4665#discussion_r1099260481) to keep `_get_image` clean. If this method returns True, the `specified_workflow` (`BuildMethod` in `.yaml`) will overwrite `runtime` to get docker image. One concern is, `dotnet7` is not the only one that has `specified_workflow`, but we know for sure it is safe because container builds for `dotnet` wasn't supported. For other runtimes consuming `provided.al2`, it's hard to say it wouldn't break. So I think this check is necessary before passing down `specified_workflow`. Lmk if there's a better way to do this.

samcli/local/docker/lambda_build_container.py
mrkdeng: Remove if that's the only consumer.

samcli/commands/build/command.py
```diff
@@ -136,6 +136,16 @@
     help="Enabled parallel builds. Use this flag to build your AWS SAM template's functions and layers in parallel. "
     "By default the functions and layers are built in sequence",
 )
+@click.option(
+    "--mount-with-write/--mount-with-read",
```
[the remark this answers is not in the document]
mrkdeng: Regarding [comment](https://github.com/aws/aws-sam-cli/pull/4665#discussion_r1100488846). Lmk `--mount-with-read` or `--no-mount-with-write` which makes more sense.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1038,10 +1038,158 @@ def test_with_dotnetcore(self, runtime, code_uri, mode, architecture="x86_64"):
...
+        [
+            ("dotnetcore3.1", "Dotnetcore3.1", None),
+            ("dotnet6", "Dotnet6", None),
+            ("dotnetcore3.1", "Dotnetcore3.1", "debug"),
+            ("dotnet6", "Dotnet6", "debug"),
+            ("provided.al2", "Dotnet7", None),
+        ]
+    )
+    @pytest.mark.flaky(reruns=3)
+    def test_with_dotnetcore_in_container_mount_with_write_explicit(
```
lucashuy, line 1051: Can we also add the Docker skip decorator to this test?
```python
@skipIf(SKIP_DOCKER_TESTS or SKIP_DOCKER_BUILD, SKIP_DOCKER_MESSAGE)
```

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1038,10 +1038,158 @@ def test_with_dotnetcore(self, runtime, code_uri, mode, architecture="x86_64"):
...
+                str(self.default_build_dir),
+            ),
+        )
+
+        expected = "{'message': 'Hello World'}"
+        if not SKIP_DOCKER_TESTS:
+            self._verify_invoke_built_function(
+                self.built_template, self.FUNCTION_LOGICAL_ID, self._make_parameter_override_arg(overrides), expected
+            )
+        self.verify_docker_container_cleanedup(runtime)
```
lucashuy, line 1112: Similarly, since this is already a Docker related test, we can remove this check and just call the methods.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1038,10 +1038,158 @@ def test_with_dotnetcore(self, runtime, code_uri, mode, architecture="x86_64"):
...
+        [
+            ("dotnetcore3.1", "Dotnetcore3.1", None),
+            ("dotnet6", "Dotnet6", None),
+            ("dotnetcore3.1", "Dotnetcore3.1", "debug"),
+            ("dotnet6", "Dotnet6", "debug"),
+            ("provided.al2", "Dotnet7", None),
+        ]
+    )
+    @pytest.mark.flaky(reruns=3)
+    def test_with_dotnetcore_in_container_mount_with_write_interactive(
```
lucashuy, line 1124: Same feedback regarding skipping Docker test decorator and `if` statement check at the end of this test

samcli/lib/build/workflow_config.py
```diff
@@ -245,11 +245,7 @@ def _key(c: CONFIG) -> str:
     # A "workflow config" is like a primary key to identify the workflow. So we use the config as a key in the
     # map to identify which workflows can support building within a container.
 
-    unsupported = {
-        _key(DOTNET_CLIPACKAGE_CONFIG): "We do not support building .NET Core Lambda functions within a container. "
-        "Try building without the container. Most .NET Core functions will build "
-        "successfully.",
-    }
+    unsupported: Dict[str, str] = {}
```
mndeveci, line 248: Why we are keeping empty `unsupported` dict here? It was `dotnet` builders before, but I can't understand what are we setting it now.

samcli/local/docker/container.py
```diff
@@ -157,6 +168,15 @@ def create(self):
             "use_config_proxy": True,
         }
 
+        # Get effective user when building lambda and mounting with write permissions
+        # Pass effective user to docker run CLI as "--user" option in the format of uid[:gid]
+        # to run docker as current user instead of root
+        # Skip if current user is root on posix systems or non-posix systems
+        _root_user_id = "0"
+        effective_user = EffectiveUser.get_current_effective_user().to_effective_user_str()
+        if self._mount_with_write and effective_user and effective_user != _root_user_id:
```
mndeveci, line 177: Can we add some debug logs here to investigate any possible issue in the future?

samcli/local/docker/container.py
```diff
@@ -256,6 +276,12 @@ def delete(self):
             if not removal_in_progress:
                 raise ex
             LOG.debug("Container removal is in progress, skipping exception: %s", msg)
+        finally:
+            # Remove tmp dir on the host
+            if self._host_tmp_dir:
+                host_tmp_dir_path = pathlib.Path(self._host_tmp_dir)
+                if host_tmp_dir_path.exists():
+                    shutil.rmtree(self._host_tmp_dir)
```
mndeveci, line 284: Same, let's add some debug logging

samcli/local/docker/container.py
```diff
@@ -277,6 +303,10 @@ def start(self, input_data=None):
         if not self.is_created():
             raise RuntimeError("Container does not exist. Cannot start this container")
 
+        # Make tmp dir on the host
+        if self._mount_with_write and self._host_tmp_dir and not os.path.exists(self._host_tmp_dir):
+            os.mkdir(self._host_tmp_dir)
```
mndeveci, line 308: Same, let's add some debug logging

mndeveci (changes requested):

samcli/commands/build/command.py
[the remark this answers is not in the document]
jfuss: I can sync with @mndeveci. I would rather this be an option with enum values. I find this being a flag confusing and will make `samconfig.toml` more confusing than it is today.

samcli/lib/build/app_builder.py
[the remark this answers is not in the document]
jfuss: I don't have anything against prompting but it shouldn't be done within the lib. If the customer is passing in "mount with write", we can check that in the `samcli/commands/build` module. It doesn't need to be tied to a specific workflow (that will just cause more issues in the long run).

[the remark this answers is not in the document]
jfuss: I don't have anything against prompting but it shouldn't be done within the lib. If the customer is passing in "mount with write", we can check that in the samcli/commands/build module. It doesn't need to be tied to a specific workflow (that will just cause more issues in the long run).

samcli/commands/build/command.py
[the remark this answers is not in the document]
jfuss: I can sync with @mndeveci. I would rather this be an option with enum values. I find this being a flag confusing and will make samconfig.toml more confusing than it is today.

samcli/lib/build/workflow_config.py
sriram-mv: This makes sense, let's remove if we don't need it.

samcli/commands/build/build_context.py
```diff
@@ -296,7 +300,6 @@ def run(self):
             BuildError,
             BuildInsideContainerError,
             UnsupportedBuilderLibraryVersionError,
-            ContainerBuildNotSupported,
```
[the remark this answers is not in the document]
mrkdeng: `dotnet` is the only runtime that throws this exception

samcli/lib/build/app_builder.py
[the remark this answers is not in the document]
mrkdeng: Moved prompting to `samcli.build.build_context`. I import `samcli.lib.build.work_config` and assume it's ok since there are already imports from `samlic.lib.build` to `samcli.build.build_context`

samcli/commands/build/utils.py
```diff
@@ -1,13 +1,65 @@
...
+        function = functions[0]
+        runtime = function.runtime
+        code_uri = function.codeuri
+        # get specified_workflow if metadata exists
+        metadata = function.metadata
+        specified_workflow = metadata.get("BuildMethod", None) if metadata else None
+    elif len(layers) > 0:
+        layer = layers[0]
+        specified_workflow = layer.build_method
+        code_uri = layer.codeuri
```
mndeveci, line 50: Why we are checking just one function and layer? Shouldn't this loop through all of them to see if any of them satisfy the need for the prompt?

samcli/commands/build/command.py
```diff
@@ -275,6 +287,7 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
         aws_region=click_ctx.region,
         hook_name=hook_name,
         build_in_source=build_in_source,
+        mount_with=mount_with,
```
mndeveci, line 290: I would set the default (READ) here in case it is `None` to make the rest of the implementation easier (we don't need to check if it is `None` or any other value.)

samcli/lib/build/app_builder.py
```diff
@@ -682,7 +682,7 @@ def _build_function(  # pylint: disable=R1710
                     # None represents the global build image for all functions/layers
                     global_image = self._build_images.get(None)
                     image = self._build_images.get(function_name, global_image)
-
+                    build_workflow = specified_workflow if supports_specified_workflow(specified_workflow) else None
```
mndeveci, line 685: Shouldn't we set this higher up at line 650?

samcli/commands/build/command.py
```diff
@@ -136,6 +136,15 @@
     help="Enabled parallel builds. Use this flag to build your AWS SAM template's functions and layers in parallel. "
     "By default the functions and layers are built in sequence",
 )
+@click.option(
+    "--mount-with",
+    "-mw",
+    type=click.Choice(["READ", "WRITE"], case_sensitive=False),
```
mndeveci, line 142: I would recommend writing an enum and comparing its value rather than keeping constant string (`"read", "write"`) in the code.

mndeveci (changes requested):

samcli/commands/build/command.py
```diff
@@ -136,6 +137,16 @@
     help="Enabled parallel builds. Use this flag to build your AWS SAM template's functions and layers in parallel. "
     "By default the functions and layers are built in sequence",
 )
+@click.option(
+    "--mount-with",
+    "-mw",
+    type=click.Choice(MountMode.values(), case_sensitive=False),
+    default=MountMode.READ.value,
+    help="Optional. Specify mount mode for building functions/layers inside container. "
+    "If mount with write permissions, some files in source code directory may be changed/added by the build process. "
```
mndeveci, line 146: Might be a better help text? 🤔 
```suggestion
    "If it is mounted with write permissions, some files in source code directory may be changed/added by the build process. "
```

samcli/commands/build/command.py
```diff
@@ -175,6 +186,7 @@ def cli(
     config_env: str,
     hook_name: Optional[str],
     skip_prepare_infra: bool,
+    mount_with="READ",
```
mndeveci, line 189: Still using string, let's use the enum

samcli/commands/build/command.py
```diff
@@ -230,6 +243,7 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     exclude: Optional[Tuple[str, ...]],
     hook_name: Optional[str],
     build_in_source: Optional[bool],
+    mount_with="READ",
```
mndeveci, line 246: Same as above, please use enum rather than hardcoded string

samcli/commands/build/utils.py
```diff
@@ -0,0 +1,106 @@
...
+        if code_uri:
+            runtime = function.runtime
+            code_dir = str(pathlib.Path(base_dir, code_uri).resolve())
+            # get specified_workflow if metadata exists
+            metadata = function.metadata
+            specified_workflow = metadata.get("BuildMethod", None) if metadata else None
+            config = get_workflow_config(runtime, code_dir, base_dir, specified_workflow)
+            # at least one function needs mount with write, return with prompting
+            if config.must_mount_with_write_in_container:
+                return prompt(config, code_dir)
```
mndeveci, line 66: If you use early-exit in if statements, we can remove some indentation here to remove some complexity while reading this code (see: https://medium.com/swlh/return-early-pattern-3d18a41bba8)
```suggestion
        if not code_uri:
            continue
        runtime = function.runtime
        code_dir = str(pathlib.Path(base_dir, code_uri).resolve())
        # get specified_workflow if metadata exists
        metadata = function.metadata
        specified_workflow = metadata.get("BuildMethod", None) if metadata else None
        config = get_workflow_config(runtime, code_dir, base_dir, specified_workflow)
        # at least one function needs mount with write, return with prompting
        if not config.must_mount_with_write_in_container:
            continue
        return prompt(config, code_dir)
```

Same can be done with the other loop for layer below

samcli/lib/build/app_builder.py
mndeveci: Then can we re-use `specified_workflow` name or something else? Since `build_workflow` is too generic comparing to where it is coming from.

samcli/lib/build/workflow_config.py
```diff
@@ -222,40 +222,25 @@ def get_workflow_config(
...
     -------
-    tuple(bool, str)
-        True, if this workflow can be built inside a container. False, along with a reason message if it cannot be.
+    bool
+        True, if this workflow is supported, can be used to overwrite runtime and get docker image
     """
 
-    def _key(c: CONFIG) -> str:
-        return str(c.language) + str(c.dependency_manager) + str(c.application_framework)
+    supported_specified_workflow = ["dotnet7"]
```
mndeveci, line 241: Isn't this return `False` for every other runtime? In that case, it looks like we might be breaking their build workflow

mndeveci (changes requested):

samcli/lib/build/workflow_config.py
mrkdeng: This wouldn't break others. The last parameter passed to `_build_function_on_container` and downstream is newly added. It allows builds such as `dotnet7` with `provided.al2` as runtime and a specified workflow to use the specified workflow to get docker image. This method returns `False` for others, so no specified workflow being passed down and they will use `runtime` to get docker image as previous.

samcli/commands/build/build_context.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         locate_layer_nested: bool = False,
         hook_name: Optional[str] = None,
         build_in_source: Optional[bool] = None,
+        mount_with=MountMode.READ.value,
```
mndeveci, line 81:
```suggestion
        mount_with: MountMode = MountMode.READ
```

samcli/commands/build/build_context.py
```diff
@@ -235,6 +239,19 @@ def run(self):
 
         self._stacks = self._handle_build_pre_processing()
 
+        # boolean value indicates if mount with write or not, defaults to READ ONLY
+        mount_with_write = False
+        if self._use_container:
+            if self._mount_with.lower() == MountMode.WRITE.value.lower():
```
mndeveci, line 245:
```suggestion
            if self._mount_with == MountMode.WRITE:
```

samcli/commands/build/command.py
```diff
@@ -136,6 +137,16 @@
     help="Enabled parallel builds. Use this flag to build your AWS SAM template's functions and layers in parallel. "
     "By default the functions and layers are built in sequence",
 )
+@click.option(
+    "--mount-with",
+    "-mw",
+    type=click.Choice(MountMode.values(), case_sensitive=False),
```
mndeveci, line 143: Please check how this is done in other places, example: https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/local/cli_common/options.py#L230

samcli/commands/build/command.py
```diff
@@ -175,6 +186,7 @@ def cli(
     config_env: str,
     hook_name: Optional[str],
     skip_prepare_infra: bool,
+    mount_with=MountMode.READ.value,
```
mndeveci, line 189: Please add typing and  default value might be redundant since click sets this up automatically

samcli/commands/build/command.py
```diff
@@ -230,6 +243,7 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
     exclude: Optional[Tuple[str, ...]],
     hook_name: Optional[str],
     build_in_source: Optional[bool],
+    mount_with=MountMode.READ.value,
```
mndeveci, line 246: Default might come from click setting above, so we can add typing and remove default from here as well

samcli/commands/build/utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @classmethod
+    def values(cls) -> List[str]:
+        """
+        A getter to retrieve the accepted value list for mount mode
+
+        Returns: List[str]
+            The accepted mount mode list
+        """
+        return [e.value for e in cls]
```
mndeveci, line 30: If we can use `_members_` for click option, then we might remove this function here as well

samcli/local/docker/effective_user.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def to_effective_user_str(self) -> Optional[str]:
+        """
+        Return String representation of the posix effective user, or None for non posix systems
+        """
+        if not self.user_id:
+            # Return None for non-posix systems
+            return None
+
+        root_user_id = "0"
```
mndeveci, line 22: can we move this to a constant?

samcli/local/docker/lambda_build_container.py
```diff
@@ -83,16 +88,25 @@ def __init__(  # pylint: disable=too-many-locals
...
             # Manifest is mounted separately in order to support the case where manifest
             # is outside of source directory
-            manifest_dir: {"bind": container_dirs["manifest_dir"], "mode": "ro"}
+            manifest_dir: {"bind": container_dirs["manifest_dir"], "mode": mount_mode}
         }
 
+        host_tmp_dir = None
+        if mount_with_write:
+            # Mounting tmp dir on the host as ``/tmp/samcli`` on container, which gives current user write permissions
+            host_tmp_dir = os.path.join(source_dir, str(uuid4()))
```
mndeveci, line 107: Let's use utility functions to create tmp directory rather than creating one in the source folder. See example utility here: https://github.com/aws/aws-sam-cli/blob/3a7a4c39d62456b623d4183857c6e253e5236cef/samcli/lib/package/local_files_utils.py#L14

mrkdeng: Instead of `source_dir`, now using `../.aws-sam/build/`

samcli/commands/build/build_context.py
mrkdeng: Click.Choice takes [str] only. Workaround: Convert str to enum in BuildContext.__init__() https://github.com/aws/aws-sam-cli/blob/280b16030ddee5e1b8ff1121f39e29892a53186e/samcli/commands/build/build_context.py#L178

mndeveci (approved):
Thanks for addressing all the comments, looking forward to ship this feature soon! 🥳

jfuss (approved):
LGTM.

Thanks @mrkdeng for handling all the feedback along the way. I know that can be a lot and time consuming but I think we ended up in a good state as a result.

PR #4668, opened 2023-02-08
feat: Collect Lambda Authorizers found in Cloudformation resources

lucashuy:
Adds the ability to parse `AWS::ApiGateway::Authorizer` and `AWS::ApiGatewayV2::Authorizer` resources to collect Lambda authorizers.

Updates the parsing of `AWS::ApiGateway::Method` and `AWS::ApiGatewayV2::Route` resources to save their reference to an authorizer.

Updates the main `cfn_api_provider.py` by adding two new static methods for the authorizers, and modifies the existing methods for `Method` and `Route`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/cfn_api_provider.py
```diff
@@ -75,6 +95,208 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
...
+            raise InvalidSamTemplateException(
+                f"Lambda Authorizer '{logical_id}' of type TOKEN, must have "
+                f"'{CfnApiProvider._AUTHORIZER_IDENTITY_SOURCE}' of type string defined."
+            )
+
+        if identity_source_template is not None and not isinstance(identity_source_template, str):
+            raise InvalidSamTemplateException(
+                f"Lambda Authorizer '{logical_id}' contains an invalid '{CfnApiProvider._AUTHORIZER_IDENTITY_SOURCE}', "
+                "it must be a comma-separated string."
+            )
```
lucashuy, line 175: For authorizers of type `REQUEST`, identity sources are optional in API Gateway V1 depending on if caching is enabled or not (if caching is enabled, must provide identity sources). I've made the choice to not validate this.

samcli/commands/local/lib/swagger/parser.py
```diff
@@ -100,7 +100,9 @@ def get_authorizers(self, event_type: str = Route.API) -> Dict[str, Authorizer]:
                 continue
 
             authorizer_type = authorizer_object.get(SwaggerParser._AUTHORIZER_TYPE, "").lower()
-            payload_version = authorizer_object.get(SwaggerParser._AUTHORIZER_PAYLOAD_VERSION, "1.0")
+            payload_version = authorizer_object.get(
+                SwaggerParser._AUTHORIZER_PAYLOAD_VERSION, LambdaAuthorizer.PAYLOAD_V1
+            )
```
lucashuy, line 105: Moving to constant

samcli/lib/list/endpoints/endpoints_producer.py
```diff
@@ -25,9 +25,9 @@
     AWS_APIGATEWAY_V2_API,
     AWS_LAMBDA_FUNCTION_URL,
     AWS_APIGATEWAY_BASE_PATH_MAPPING,
-    AWS_APIGATEWAY_v2_BASE_PATH_MAPPING,
+    AWS_APIGATEWAY_V2_BASE_PATH_MAPPING,
     AWS_APIGATEWAY_V2_DOMAIN_NAME,
-    AWS_APIGATWAY_DOMAIN_NAME,
+    AWS_APIGATEWAY_DOMAIN_NAME,
```
lucashuy, line 30: Fixing spelling/formatting mistake

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -66,6 +83,9 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
                 if resource_type == AWS_APIGATEWAY_METHOD:
                     self._extract_cloud_formation_method(stack.stack_path, resources, logical_id, resource, collector)
 
+                if resource_type == AWS_APIGATEWAY_AUTHORIZER:
```
sriram-mv, line 86: Nothing specific to this PR, but I would want to us to move to a dispatch pattern whenever there are multiple if branches.

Something like:
```python
for resource_extractor in resource_extractors:
     resource_extract.extract(*args, **kwargs) # pass in the args deemed necessary.
```

might be easier to read and the behavior is apparent.

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -75,6 +95,208 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
...
+        if identity_source_template is not None and not isinstance(identity_source_template, str):
+            raise InvalidSamTemplateException(
+                f"Lambda Authorizer '{logical_id}' contains an invalid '{CfnApiProvider._AUTHORIZER_IDENTITY_SOURCE}', "
+                "it must be a comma-separated string."
+            )
+
+        # split and parse out identity sources
+        identity_source_list = []
+
+        if identity_source_template:
```
sriram-mv, line 180: should there be validation on it being a comma separated list?

lucashuy: Thats a good idea. I think we can further extend this to validating the actual identity source itself, likely using a set of regular expressions.

I make that change in a different PR since that also applies to the other methods of collecting authorizers, and also address the empty string comparison that is done when linking.

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -75,6 +95,208 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
...
+        ----------
+        logical_id: str
+            The logical ID of the Authorizer
+        resource: dict
+            The attributes for the Authorizer
+        collector: ApiCollector
+            ApiCollector to save Authorizers into
+        """
+        properties = resource.get("Properties", {})
+        authorizer_type = properties.get(CfnApiProvider._AUTHORIZER_TYPE, "").lower()
```
mndeveci, line 116: Just curious, why we are calling `lower` here?

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -75,6 +95,208 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
...
+                logical_id,
+                authorizer_type,
+            )
+            return
+
+        if not authorizer_uri:
+            raise InvalidSamTemplateException(
+                f"Authorizer '{logical_id}' is missing the '{CfnApiProvider._AUTHORIZER_AUTHORIZER_URI}' "
+                "property, a valid Lambda ARN must be provided."
+            )
```
mndeveci, line 152: I feel like this part can be extracted as validation method to to improve the readability. Which can also be combined what @sriram-mv suggested above (like different types might have their own implementation for a base interface or an abstract class).

samcli/lib/providers/cfn_api_provider.py
mndeveci: It is great to read these feedbacks, but I wonder if we should write some of them in the code so that we can understand the decision later? 🤔

samcli/lib/providers/cfn_api_provider.py
lucashuy: When defining the type in `Properties`, the expected values are `TOKEN` and `REQUEST`, but if defined in swagger, then its `token` or `request`. To make things consistent, I elected to make things lowercase once it makes its way into SAM CLI.

lucashuy: Oh, a good point here would be to also consider validating this behaviour.

sriram-mv (approved):

mndeveci (approved):

PR #4671, opened 2023-02-08
chore: fix sam local and sam sync integration tests

mndeveci:
Fixes recent issues with integration tests;
- With removal of `python3.6`, we bumped some local tests to use newer python versions but they have different output if an error happens. Updated those tests with expected error message (also verified with error message of actually deployed lambda [see below]).
```
{
    "errorMessage": "Lambda is raising an exception", 
    "errorType": "Exception", 
    "requestId": "6698a3f0-09b6-4ae9-93cb-27882edd5046",
    "stackTrace": ["  File \"/var/task/lambda_function.py\", line 4, in lambda_handler\n    raise Exception(\"Lambda is raising an exception\")\n"]
}
```
- Recently added `sam sync` tests which is related to layers, have added new functions in the test template, which broke the existing tests. Those tests have been moved to its own template and test case.
- Increased default time out for waiting the process output, which should resolve the issues in recent windows tests.

### DIFF
--- a/tests/integration/testdata/invoke/template.yml
+++ b/tests/integration/testdata/invoke/template.yml
@@ -129,7 +129,7 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       Handler: main.raise_exception
-      Runtime: python3.9
+      Runtime: python3.7
       CodeUri: .
       Timeout: 600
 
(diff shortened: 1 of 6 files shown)


tests/integration/testdata/invoke/template.yml
```diff
@@ -129,7 +129,7 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       Handler: main.raise_exception
-      Runtime: python3.9
+      Runtime: python3.7
```
jfuss, line 132: Is there a reason we downgraded this?

jfuss (changes requested):

jfuss (approved):

hnnasit (approved):

PR #4677, opened 2023-02-09
feat: Pass sync context into all sync flows

qingchm:
N.A.

In order to pass sync context information into all sync flows, which enables more information to be obtained when each sync flow executes

Pass in sync context object into all sync flow construction.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/command.py
```diff
@@ -348,14 +348,28 @@ def do_cli(
...
                         if watch:
                             execute_watch(
-                                template_file, build_context, package_context, deploy_context, dependency_layer, code
+                                template_file,
+                                build_context,
+                                package_context,
+                                deploy_context,
+                                dependency_layer,
+                                code,
+                                sync_context,
```
mndeveci, line 362: nit: can we combine context objects together (for other places as well) for readability?

mndeveci (commented):
Thanks for the change, LGTM! Left a small comment for readability

samcli/commands/sync/command.py
qingchm: Updated

mndeveci (approved):
Thanks for addressing the feedback, can you check why CI jobs are failing?

hnnasit (approved):
LGTM too

PR #4679, opened 2023-02-09
feat: implementation on the root command help text design

sriram-mv:
Screen Shot 2023-02-13 at 5 39 31 PM

N/A

This gives the tool better UX to showcase the lifecycles of the process and how one should view the process and how to get started.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/command.py
+++ b/samcli/cli/command.py
@@ -96,10 +96,13 @@ class BaseCommand(click.MultiCommand):
 
         return commands
 
-    def format_options(self, ctx, formatter):
-        # Re-order options so that they come after the commands.
+    def format_options(self, ctx: click.Context, formatter: RootCommandHelpTextFormatter):  # type: ignore
+        # NOTE(sriram-mv): `ignore` is put in place here for mypy even though it is the correct behavior,
+        # as the `formatter_class` can be set in subclass of Command. If ignore is not set,
+        # mypy raises argument needs to be HelpFormatter as super class defines it.
+        # NOTE(sriram-mv): Re-order options so that they come after the commands.
         self.format_commands(ctx, formatter)
-        opts = [RowDefinition(text="\n")]
+        opts = [RowDefinition(name="", text="\n")]
         for param in self.get_params(ctx):
             row = param.get_help_record(ctx)
             if row is not None:
@@ -114,6 +117,7 @@ class BaseCommand(click.MultiCommand):
             formatter.write_rd(
                 [
                     RowDefinition(
+                        name="",
                         text="\n",
                     ),
                     RowDefinition(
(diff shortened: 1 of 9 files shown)


samcli/cli/command.py
```diff
@@ -104,6 +96,117 @@ def _set_commands(package_names):
 
         return commands
 
+    def format_options(self, ctx, formatter):
+        # Re-order options so that they come after the commands.
+        self.format_commands(ctx, formatter)
+        opts = [RowDefinition(text="\n")]
+        for param in self.get_params(ctx):
+            rv = param.get_help_record(ctx)
```
mildaniel, line 104: nit: Can we be a bit more descriptive with the name? From first glance it wasn't 100% intuitive to me what `rv` meant.

samcli/cli/row_modifiers.py
```diff
@@ -0,0 +1,56 @@
+"""
+Row Definition and Row Modifiers to be used with `click.HelpFormatter write_dl methods`.
+"""
+
+from typing import NamedTuple, List
+
+from click import style
+
+
+class RowModifier:
```
mildaniel, line 10: For completeness, should `RowModifier` be defined as an abstract class with an `apply` method?

samcli/cli/formatters.py
```diff
@@ -0,0 +1,66 @@
...
+class RootCommandHelpTextFormatter(HelpFormatter):
+
+    # Picked an additive constant that gives an aesthetically pleasing look.
+    ADDITIVE_JUSTIFICATION = 10
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # NOTE(sriram-mv): Do not left justify for more than half the width.
+        self.left_justification_length = min(
+            max([len(command) for command, _ in SAM_CLI_COMMANDS.items()]) + self.ADDITIVE_JUSTIFICATION,
```
mildaniel, line 22: I'm probably missing something, but not completely following the logic here for determining the justification length.

mildaniel (approved):
Left a few minor comments/questions. Curious on your thoughts about how we can integration test these more UI-specific elements going forward.

samcli/cli/formatters.py
sriram-mv: We want to add justification length based on the command that has the longest length. eg: `sam sync` is shorter than `sam pipeline`, so the command help text should start after the longest command name, otherwise the alignment would not look right.

samcli/cli/row_modifiers.py
sriram-mv: I thought about it, went with the current implementation to provide flexibility while writing.

samcli/cli/command.py
sriram-mv: renaming to `row`

samcli/cli/command.py
```diff
@@ -104,6 +96,117 @@ def _set_commands(package_names):
 
         return commands
 
+    def format_options(self, ctx, formatter):
```
mndeveci, line 99: Can we add some typing for the parameters here?

samcli/cli/command.py
```diff
@@ -104,6 +96,117 @@ def _set_commands(package_names):
...
+                        name="Get Started:",
+                        text=click.style(f"${ctx.command_path} init"),
+                        extra_row_modifiers=[ShowcaseRowModifier()],
+                    ),
+                ],
+            )
+
+    def format_commands(self, ctx, formatter):
+        with formatter.section("Commands"):
+            with formatter.section("Create an App"):
```
mndeveci, line 129: nit: I am in between whether each section here should be a method/class of its own and handle how it writes the commands section. Since this won't be touched too much I am OK with this approach as well. 🤔

mndeveci (commented):
Looks good in general, 2 minor comments there.

samcli/cli/command.py
sriram-mv: I thought about this as well, but left it this way as its apparent to the user and it won't be touched as often, we can refactor as we fit later.

mndeveci (approved):

PR #4683, opened 2023-02-09
feat: Added identity source validation and removed empty string state

lucashuy:
This adds a method to validate the identity sources for both versions of API Gateway. It also removes the empty string state that was used before to represent "use default authorizer".

Adds a new class/static method to check if the identity source is valid. Updated the `Route` object to now use a boolean to determine whether we use the default authorizer or not.

n/a

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/identity_source_validator.py
```diff
@@ -0,0 +1,57 @@
...
+            The identity source to validate
+        event_type: str
+            The type of API Gateway to validate against (API or HTTP)
+
+        Returns
+        -------
+        bool
+            True if the identity source is valid
+        """
+        for regex in IdentitySourceValidator.API_GATEWAY_VALIDATION_LIST[event_type]:
```
sriram-mv, line 53: ```return any([re.match(..) for regex in ..)]``` might be succint but might not be as performant. maybe with a generator it might be.

samcli/commands/local/lib/swagger/parser.py
```diff
@@ -305,12 +317,11 @@ def get_routes(self, event_type=Route.API) -> List[Route]:
                 payload_format_version = self._get_payload_format_version(method_config)
 
                 authorizers = method_config.get(SwaggerParser._SWAGGER_SECURITY, None)
+
                 authorizer_name = None
+                use_default_authorizer = True
```
sriram-mv, line 322: Do we enforce having an authorizer by default?

sriram-mv (approved):

samcli/commands/local/lib/swagger/parser.py
lucashuy: Cloudformation behaviour is if there is a default authorizer defined (this is defined at the root swagger level or explicitly if defined a Serverless API resource is being used), then all routes will use that authorizer unless explicitly told not to (either by setting `security` to an empty list in swagger, or setting `NONE` if using Cloudformation/Serverless resources.

So I guess that yes, we are "enforcing" a default authorizer, just by extension of replicating CFN. This would be breaking in the sense that once this feature is released, customers will now find that their Lambda authorizers are being invoked.

samcli/commands/local/lib/validators/identity_source_validator.py
```diff
@@ -0,0 +1,57 @@
...
+            The identity source to validate
+        event_type: str
+            The type of API Gateway to validate against (API or HTTP)
+
+        Returns
+        -------
+        bool
+            True if the identity source is valid
+        """
+        for regex in IdentitySourceValidator.API_GATEWAY_VALIDATION_LIST[event_type]:
```
sriram-mv, line 53: for perf reasons it might be better to compile the regex. But I don't see it being a massive perf bottleneck though.

sriram-mv (approved):

moelasmar (approved):

samcli/commands/local/lib/identity_source_validator.py
lucashuy: Comparing this vs `any` using `dis.dis(IdentitySourceValidator`), you are correct in that using `any` introduces a couple more instructions than if we just had a regular `for` loop. How much that will actually change things is to be determined, however I think this is fine as is.

PR #4673, opened 2023-02-08
Expose `rustcargolambda` BuildMethod

hawflau:
#### Which issue(s) does this change fix?

N/A

#### Why is this change necessary?
To expose `"rustcargolambda"` as a `BuildMethod` option for `provided.al2` runtime.

#### How does it address the issue?
Expose it as a build method. We aim to support two use cases right now:

**1/ Single Lambda Function/binary**
Example project structure:
```
.
├── Cargo.lock
├── Cargo.toml
├── src
│   └── main.rs
└── template.yaml
```

Example `AWS::Serverless::Function` definition:
```yaml
  MyFunction:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: rustcargolambda
    Properties:
      CodeUri: ./             # points to the dir where Cargo.toml is located
      Handler: bootstrap      # must be bootstrap
      Runtime: provided.al2
```

Example Cargo.toml:
```toml
[package]
name = "my_handler"
version = "0.1.0"
edition = "2021"

[dependencies]
aws-config = "0.49.0"
aws-sdk-dynamodb = "0.19.0"
aws-smithy-client = { version = "0.49.0", features = ["test-util"] }
lambda_http = "0.6.1"
tokio = { version = "1.21.1", features = ["full"] }

[dev-dependencies]
aws-smithy-http = "0.49.0"
aws-types = { version = "0.49.0", features = ["hardcoded-credentials"] }
http = "0.2.8"
```

**2/ Multiple Lambda Functions/binaries**
Example project structure:
```
.
├── Cargo.lock
├── Cargo.toml
├── src
│   ├── function_a.rs
│   └── function_b.rs
└── template.yaml
```

Example `AWS::Serverless::Function` definition:
```yaml
FunctionA:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: rustcargolambda
      BuildProperties:
        Binary: function_a   # specify which binary (as defined in Cargo.toml) to build
    Properties:
      CodeUri: ./            # points to dir of Cargo.toml
      Handler: bootstrap     # must be bootstrap
      Runtime: provided.al2
```

Example Cargo.toml:
```toml
[package]
name = "test-handler"
version = "0.1.0"
edition = "2021"

[dependencies]
lambda_runtime = "0.6.0"
serde = "1.0.136"
tokio = { version = "1", features = ["macros"] }
tracing = { version = "0.1", features = ["log"] }
[...]

PR #4687, opened 2023-02-10
Add integration tests for Rust BuildMethod

hawflau:

#4673 

Adding integration tests for Rust build method

Add integration tests for two use cases:
- single funtion
- multi binaries

No

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


appveyor-ubuntu.yml
```diff
@@ -23,6 +23,7 @@ environment:
   NODE_VERSION: "14.17.6"
   AWS_S3: 'AWS_S3_TESTING'
   AWS_ECR: 'AWS_ECR_TESTING'
+  CARGO_LAMBDA_VERSION: "v0.17.1"
```
jfuss, line 26: Should we run the cargo related tests in GHA instead?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1122,6 +1123,65 @@ def test_go_must_fail_with_unknown_architecture(self, runtime, code_uri, archite
         self.assertEqual(process_execute.process.returncode, 1)
 
 
+@skipIf(
+    ((IS_WINDOWS and RUNNING_ON_CI) and not CI_OVERRIDE),
+    "Skip build tests on windows when running in CI unless overridden",
+)
+@parameterized_class(
```
jfuss, line 1130: nice!

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1122,6 +1123,65 @@ def test_go_must_fail_with_unknown_architecture(self, runtime, code_uri, archite
...
+        ),
+        (
+            "template_build_method_rust_binary.yaml",
+            "Rust/multi-binaries",
+            "function_b",
+            {"req_id": "99", "msg": "Hello FunctionB"},
+        ),
+    ],
+)
+class TestBuildCommand_Rust(BuildIntegRustBase):
```
jfuss, line 1153: Is Rust buildable in a container? If so, should we have a test for that too?

jfuss (changes requested):

appveyor-ubuntu.yml
hawflau: That's a good idea. We might still want to keep cargo tests running in the Canary though (i.e. appveyor-ubuntu.yml and appveyor-windows.yml). But moving those running per PR will prevent each PR from taking longer for the checks.

tests/integration/buildcmd/test_build_cmd.py
hawflau: It's buildable in a container but customers will need to provide their own build image.
For now, our `provided.al2` build image doesn't have Rust toolchain installed. We have a plan to provide individual `provided.al2` build images with specific toolchain for each of Go, Rust, but it's not proritized yet.

appveyor-ubuntu.yml
hawflau: Made a change to use GH Actions to run Rust integration tests but there are some caveats:
- The GH Actions now can only run on `ubuntu-latest` but not Windows. Reason is I couldn't find a way to switch Windows Docker to use Linux Containers (unlike AppVeyor).
- Therefore, we still need AppVeyor to run these tests in Windows, which is defined in `appvyeor-windows-build-rust.yml`
- We still run the Rust integration tests in Canary (defined in appveyor-ubuntu.yml and appveyor-windows.yml) to keep the completeness of tests in Canary
- The result is - we only remove running Rust tests in the Appveyor make pr check. But the benefit is significant - we avoid adding an extra 22minutes of time to the Appveyor make pr check.
- Downside - it is a bit confusing as tests are spread across in different CI scripts. I'm a bit worried about the possibility of us missing some tests by having none of the CI scripts to pick up these tests.

Tagging @moelasmar as well as he is working on SAM CLI release improvement. For this PR, I'm in favor of running the Rust integration tests in GH Actions to not add extra time in AppVeyor make pr check. But let's keep the downside in mind and be sure to address it when we revisit this topic in SAM CLI release improvement

jfuss: > it is a bit confusing as tests are spread across in different CI scripts. I'm a bit worried about the possibility of us missing some tests by having none of the CI scripts to pick up these tests

Could we make it a single command (within `Makefile`) to make this easier?

jfuss:
@hawflau Once this is merged, let's make sure `continuous-integration/appveyor/pr/windows-build-rust` job get's marked as Required and same for any of the Github Actions for testing.

moelasmar (approved):
It looks good to me. Just a question shall we add package and deployment integration test cases as per the last action item we decide based on the last issue we faced for Go ?

appveyor-windows-build-rust.yml
```diff
@@ -0,0 +1,81 @@
+version: 1.0.{build}
```
mndeveci, line 1: I think we need this since GH actions are not supporting some stuff for windows?

hawflau: yes, unfortunately GH Actions' Windows runner doesn't have the Docker support (switch to Linux containers) that we need. At least I couldn't figure out how to do that as of now.

mndeveci (approved):

PR #4697, opened 2023-02-10
test: add sam sync integration tests for building in source with makefile

torresxb1:
Adds tests with makefiles that create new files, and asserts that those new files are/aren't created in the source directory depending on whether we're using the build-in-source option.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/sync/test_sync_build_in_source.py
```diff
@@ -0,0 +1,72 @@
...
+from parameterized import parameterized
+
+from tests.integration.sync.sync_integ_base import SyncIntegBase
+from tests.testing_utils import run_command_with_input
+from samcli.lib.utils.resources import AWS_LAMBDA_FUNCTION
+
+LOG = logging.getLogger(__name__)
+
+
+@skip("Building in source option is not exposed yet. Stop skipping once it is.")
```
torresxb1, line 15: skipping since the click option to build in source is not exposed yet. Made sure tests passed on my machine when I exposed the option and didn't skip the tests.

tests/integration/sync/test_sync_build_in_source.py
```diff
@@ -0,0 +1,72 @@
...
+class TestSyncInfra_BuildInSource_Makefile(SyncIntegBase):
+    dependency_layer = False
+
+    def setUp(self):
+        super().setUp()
+        self.new_files_in_source = [
+            self.test_data_path.joinpath(
+                "code/before/makefile_function_create_new_file/file-created-from-makefile-function"
```
mndeveci, line 23: Will `/` work in windows?

tests/integration/sync/test_sync_build_in_source.py
```diff
@@ -0,0 +1,72 @@
...
+                os.remove(path)
+
+    @parameterized.expand(
+        [
+            (True, True),  # build in source
+            (False, False),  # don't build in source
+            (None, False),  # use default for workflow (don't build in source)
+        ]
+    )
+    def test_sync_builds_and_deploys_successfully(self, build_in_source, new_file_should_be_in_source):
```
torresxb1, line 41: @mndeveci do you think tests with code, watch, ADL are also necessary? I was thinking they probably shared the same building logic, but let me know what you think.

mndeveci: Good point, let's add them as parameterized, we don't want to miss an edge case there. You can check some existing ones, they should have parameters on top of class definition.

tests/integration/sync/test_sync_build_in_source.py
torresxb1: seems like windows tests are passing, and I think I did this because I saw it was done in other places:
https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/sync/test_sync_watch.py#L559

tests/integration/sync/test_sync_build_in_source.py
torresxb1: done, added sync --code tests

tests/integration/sync/test_sync_build_in_source.py
```diff
@@ -0,0 +1,141 @@
...
+        super().tearDown()
+        for path in self.new_files_in_source:
+            if os.path.isfile(path):
+                os.remove(path)
+
+    @parameterized.expand(
+        [
+            (True, True),  # build in source
+            (False, False),  # don't build in source
+            (None, False),  # use default for workflow (don't build in source)
```
sriram-mv, line 40: why None? seems odd since everything else is bool.

tests/integration/testdata/sync/code/after/makefile_layer_create_new_file/Makefile
```diff
@@ -0,0 +1,7 @@
+build-HelloWorldLayer:
```
sriram-mv, line 1: does this work on windows too?

tests/integration/testdata/sync/code/before/template-makefile-create-new-file.yaml
```diff
@@ -0,0 +1,29 @@
...
+  HelloWorldLayer:
+    Type: AWS::Serverless::LayerVersion
+    Properties:
+      ContentUri: makefile_layer_create_new_file/
+      CompatibleRuntimes:
+        - python3.7
+    Metadata:
+      BuildMethod: makefile
+
+
```
sriram-mv, line 29: some extra lines hanging out here.

sriram-mv (commented):
Overall, it looks good. Will approve once questions are addressed.

tests/integration/sync/test_sync_build_in_source.py
torresxb1: In Lambda Builders, each workflow has a different directory they use to build in by default (some use scratch, some artifacts, some source dir). So there's the option to build in source (build-in-source is True), not build in source (build-in-source is False), or use the workflow's default (build-in-source is None). We could use a False value to indicate that we should use the workflow's default, but it would be weird for someone to say "don't build in source" (setting it to False), and then the default for the workflow is to build in source so it would indeed end up building in source

tests/integration/sync/test_sync_build_in_source.py
```diff
@@ -0,0 +1,141 @@
...
+        ]
+    )
```
mndeveci, line 105: Should this be parameterized on top of the class, since build-in-source might change the initial deployment there?

tests/integration/sync/sync_integ_base.py
```diff
@@ -44,6 +44,7 @@ def setUp(self):
         self.stacks = []
         self.s3_prefix = uuid.uuid4().hex
         self.dependency_layer = True if self.dependency_layer is None else self.dependency_layer
+        self.basic_capabilities = ["CAPABILITY_IAM", "CAPABILITY_AUTO_EXPAND"]
```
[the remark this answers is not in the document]
torresxb1: ah I just noticed that most tests just copy pasted those so I thought about replacing them with a variable, but then I think I decided not to refactor (too avoid making many changes)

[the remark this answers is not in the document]
torresxb1: done

tests/integration/sync/test_sync_build_in_source.py
torresxb1: done

tests/integration/testdata/sync/code/before/template-makefile-create-new-file.yaml
torresxb1: done

tests/integration/testdata/sync/code/after/makefile_layer_create_new_file/Makefile
torresxb1: good question, have not addressed this one yet because I'm not able to access my Windows VM at the moment

torresxb1: done; addressed by changing `touch` to instead echo into a file (should be supported in windows too)

mndeveci (approved):

sriram-mv (approved):

Issue #681, opened 2018-09-26
Support for context.function_name?


aniecki:
**Description:**
I want a lambda to call itself. I am using an expression inside a python programme 
FunctionName=context.function_name

yet, it returns ResourceNotFoundException error as there is no such a function, actually somehow (why?) named test..
with the arn (as example) arn:aws:lambda::eu-west-1:104260451092:function:test.

If I set  an enviroment variable: AWS_LAMBDA_FUNCTION_NAME inside template, it will work "almost" ok, but it will call aws account lambda, not the local one which I test.

So how to reference (the function name) of the locally invoked lambda function, so it will call itself ?

I thought it  a name as a copy of a main handler file - in my case name (name.py) 

but no, as again -  An error occurred (ResourceNotFoundException) when calling the Invoke operation: Function not found: arn:aws:lambda:eu-west-1:163157437355:function:main

**Output of `sam --version`**:
SAM CLI, version 0.6.0

Hope there is still a way to pass a function name to the local lambda, not to call the aws one...

jfuss:
@aniecki You can call Lambda's locally through the `sam local start-lambda` command. You can find some details on how to configure boto3 (other SDK's follow similar patterns) [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-automated-tests.html). 

If you are using Windows or Mac, you will need to configure the endpoint to be `host.docker.internal:3001`. This will allows docker to talk to the localhost endpoint on your machine. Docker for Linux does not yet support this. If you are on that system, you can either attach the container to the `host` network, which will allow you to directly use `localhost:3001` (or the port `sam local start-lambda` is configured to). 

`sam local start-lambda` only supports the `FunctionName` being the LogicalId of the function in the template.

jfuss:
Also it is not clear to me (can't remember) the reason why we don't add the [FunctionName to the env var](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/local/lambdafn/env_vars.py#L162). This could be due to we don't support or understand the `Name` Property of the function. But this should't stop what you are trying to do.

otherview:
@jfuss Any change you can expand on this ?

I'm running aws-sam-cli in debug-api mode, and trying to make one lambda call another lambda.

Here's my Template:

```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  derp-serverless

[...]

PR #4699, opened 2023-02-10
feat(invoke): Ensure context object is function name if defined in template

jfuss:
Solves: #681

#681

Prior to this change, the `context.function_name` only ever returned the LogicalId. This updates to be the FunctionName if defined in the template.

See above.

There shouldn't be any.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jfuss:
I need to validate what happens when the FunctionName in the template is an unresolvable intrinsic or an intrinsic in general. I think it just works but want to validate and only thought about it when I pushed and was creating this pr.

jfuss:
Moving this to draft. With an unresolved intrinsic, I am getting the Intrinsic name. I need to spend more time figuring out how to handle this correctly.

PR #4713, opened 2023-02-14
chore: update deploy tests to run in 4 parallel threads

mndeveci:
- Create temporary directory for each test case, so they can run in isolated folder
- Update current working directory to the directory above for every command that is run in the tests
- Provide `--s3-prefix` option so that any uploaded resource won't clash with each other
- Update appveyor configurations to run these tests in 4 parallel threads

Ran with a benchmark tool with 3 runs, results are;
```
  Time (mean ± σ):     1744.245 s ± 46.524 s    [User: 230.183 s, System: 32.357 s]
  Range (min … max):   1703.782 s … 1795.080 s    3 runs
```
This should bring down deploy tests from 2hrs to around 30mins.


appveyor-ubuntu.yml
```diff
@@ -269,7 +269,7 @@ for:
 
     test_script:
       - "pip install -e \".[dev]\""
-      - sh: "pytest -vv tests/integration/deploy --json-report --json-report-file=TEST_REPORT-integration-deploy.json"
+      - sh: "pytest -vv tests/integration/deploy -n 4 --reruns 4 --json-report --json-report-file=TEST_REPORT-integration-deploy.json"
```
sriram-mv, line 272: How do we ensure that this won't result in throttling?

sriram-mv (approved):

moelasmar (approved):

PR #4719, opened 2023-02-15
feat: Add local hash for all sync flows and add local hash comparison

qingchm:
N.A.

To enable local hash calculation mechanism for all sync flows. In this way we can store the sync status each time locally, if the hash remains the same, we can skip the code sync.

Added a sync state identifier field for all sync flows, which is unique for each sync flow to be mapped to. When the hash value is the same between the stored value and the calculated value, we skip sync.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -61,6 +62,11 @@ def __init__(
         self._function_identifier = function_identifier
         self._alias_name = alias_name
         self._lambda_client = None
+        # Sync state is the unique identifier for each sync flow
+        # In sync state toml file we will store
+        # Key as AliasVersionSyncFlow:FunctionLogicalId:AliasName
+        # Value as alias version number
+        self._sync_state_identifier = self.__class__.__name__ + ":" + self._function_identifier + ":" + self._alias_name
```
mndeveci, line 69: I wonder if we could do this as a method and calculate this on the fly (to have less fields and more straightforward constructor)? 🤔

samcli/lib/sync/flows/generic_api_sync_flow.py
```diff
@@ -67,6 +68,8 @@ def __init__(
     def gather_resources(self) -> None:
         self._definition_uri = self._get_definition_file(self._api_identifier)
         self._swagger_body = self._process_definition_file()
+        if self._swagger_body:
```
mndeveci, line 71: What happens if `self._swagger_body` is `None`, are we skipping this sync flow if that is the case?

samcli/lib/sync/sync_flow.py
```diff
@@ -315,7 +340,7 @@ def execute(self) -> List["SyncFlow"]:
         LOG.debug("%sGathering Resources", self.log_prefix)
         self.gather_resources()
         LOG.debug("%sComparing with Remote", self.log_prefix)
-        if not self.compare_remote():
+        if (not self.compare_local()) and (not self.compare_remote()):
```
mndeveci, line 343: For now, `this.compare_local()` will return `False` until we start to store this information, right?

qingchm: Correct, since compare local is local sha == sync state hash, since sync state hash will always be none for now, this will compare local will always be false

samcli/lib/sync/flows/generic_api_sync_flow.py
qingchm: This is only for mypy compliance for Optional type, we do not allow none type for this (this is only None if the definition file is None), as we will throw exception during sync

samcli/lib/sync/flows/alias_version_sync_flow.py
qingchm: We can make this a property

mndeveci: You mean function with `@property` annotation on top of it? If so I like that idea!

qingchm: yep!

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -62,19 +63,34 @@ def __init__(
...
+    @property
+    def sync_state_identifier(self) -> str:
+        """
+        Sync state is the unique identifier for each sync flow
+        In sync state toml file we will store
+        Key as AliasVersionSyncFlow:FunctionLogicalId:AliasName
+        Value as alias version number
+        """
+        identifier = self.__class__.__name__ + ":" + self._function_identifier + ":" + self._alias_name
+        return identifier.replace("/", ":")
```
hnnasit, line 75: Q: Is the replacement of "/" here intended for nested stack names? If so sync_context already replaces "/" with "-" before storing it in the syncstate.

mndeveci: I would assume that should be the job of `SyncContext` (or `SyncState`) to keep functionality in one place. Otherwise, we might need to do it in every `SyncFlow` definition.

samcli/lib/sync/flows/layer_sync_flow.py
```diff
@@ -202,6 +201,16 @@ def __init__(
...
 
+    @property
+    def sync_state_identifier(self) -> str:
+        """
+        Sync state is the unique identifier for each sync flow
+        In sync state toml file we will store
+        Key as LayerSyncFlow:LayerLogicalId
+        Value as layer ZIP hash
+        """
+        return self.__class__.__name__ + ":" + self._layer_identifier
```
mndeveci, line 212: This could be done in `AbstractLayerSyncFlow` so that we don't have to repeat it for other instances (since they are the same).

samcli/lib/sync/sync_flow.py
```diff
@@ -93,13 +97,40 @@ def set_up(self) -> None:
...
+        Simply return False if there is no comparison needed.
+        Ex: Comparing local Lambda function artifact with stored SHA256
+
+        Returns
+        -------
+        bool
+            Return True if current resource and cached are in sync. Skipping rest of the execution.
+            Return False otherwise.
+        """
+        stored_sha = self._sync_context.get_resource_latest_sync_hash(self.sync_state_identifier)
```
mndeveci, line 129: For being safe, can we return `False` for now?

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -62,19 +63,33 @@ def __init__(
...
+        return self.__class__.__name__ + ":" + self._function_identifier + ":" + self._alias_name
+
     def set_up(self) -> None:
         super().set_up()
         self._lambda_client = self._boto_client("lambda")
 
     def gather_resources(self) -> None:
         pass
 
+    def compare_local(self) -> bool:
```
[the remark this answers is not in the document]
qingchm: SyncFlow class already has its own compare_local method, to me since for alias version sync flow they will always have no local sha value during comparison, we can just override the method to return False. But removing this method will have the same effect, so I'm fine with removal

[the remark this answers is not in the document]
mndeveci: Ah OK it makes sense now, thanks for explaining!

samcli/lib/sync/sync_flow.py
```diff
@@ -93,13 +97,41 @@ def set_up(self) -> None:
     def _boto_client(self, client_name: str):
         return get_boto_client_provider_from_session_with_config(cast(Session, self._session))(client_name)
 
+    # Sync state is the unique identifier for each sync flow
+    # We store the identifier in sync state toml file as key
```
mndeveci, line 101: This is already explained below the function, why we needed to write here too?

samcli/lib/sync/flows/zip_function_sync_flow.py
```diff
@@ -70,10 +69,19 @@ def __init__(
...
 
+    @property
+    def sync_state_identifier(self) -> str:
+        """
+        Sync state is the unique identifier for each sync flow
+        In sync state toml file we will store
+        Key as ZipFunctionSyncFlow:FunctionLogicalId
+        Value as function ZIP hash
+        """
+        return self.__class__.__name__ + ":" + self._function_identifier
```
mndeveci, line 83: This method can be moved to `FunctionSyncFlow` class which should have the `self._function_identifier` field.

hnnasit (approved):

mndeveci (approved):

PR #4723, opened 2023-02-16
feat: Create default samconfig.toml

mildaniel:

N/A

Make it easier to understand the way `samconfig.toml` works by providing a default configuration with best practices for each generated init template.

Created a post-processing step after `sam init` is run to generate a `samconfig.toml` file into the newly created app template directory. Config file is dynamically generated based on input parameter.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/init/default_samconfig.py
```diff
@@ -0,0 +1,111 @@
...
+        Write global default properties to samconfig.toml
+        """
+        defaults = [Default(writing_type=WritingType.Both, key="stack_name", value=f"sam-app-{self._project_name}")]
+        self._write(defaults=defaults, command=["global"])
+
+    def _write_build(self) -> None:
+        """
+        Write build default properties to samconfig.toml
+        """
+        defaults = [
```
sriram-mv, line 74: Wonder if we could do a singular list of defaults with the command and env baked into it. This way we could remove all the `_write_{command}` methods.

samcli/lib/init/__init__.py
```diff
@@ -131,6 +134,8 @@ def generate_project(
 
     _enable_application_insights(application_insights, output_dir, name)
 
+    _create_default_samconfig(package_type, output_dir, name)
```
sriram-mv, line 137: TODO: we need to address these kind of post-processing steps with hooks in the future.

samcli/lib/init/default_samconfig.py
```diff
@@ -0,0 +1,111 @@
...
+        Parameters
+        ----------
+        defaults: List[Default]
+            A list of default properties for a specific command to be written to the samconfig file
+        command: List[str]
+            List of strings representing the command to be added to the samconfig file
+        """
+        for default in defaults:
+            if default.writing_type == WritingType.Both or default.writing_type.value == self._package_type:
+                self._config.put(cmd_names=command, section="parameters", key=default.key, value=default.value)
```
mndeveci, line 61: I wonder if we should write the other package type as commented out so that they can uncomment if they add the other package into their template? 🤔

sriram-mv: hmm, maybe but how do we showcase to customers to go back and uncomment them? especially if they are switching between image and zip on the same function.

sriram-mv: I am going to leave a TODO in there.

samcli/lib/init/__init__.py
```diff
@@ -145,3 +150,30 @@ def _enable_application_insights(application_insights: bool, output_dir: str, na
         template_modifier = ApplicationInsightsTemplateModifier(template_file_path)
         template_modifier.modify_template()
         EventTracker.track_event(EventName.USED_FEATURE.value, UsedFeature.INIT_WITH_APPLICATION_INSIGHTS.value)
+
+
+def _create_default_samconfig(package_type: str, output_dir: str, name: str) -> None:
+    """
+    Init post-processing function to create a default samconfig.toml
+    file and place it into the cookie cutter project.
```
moelasmar, line 158: I am confused, are we generating the toml file in the generated sam app project or in the cookie cutter project ?

mndeveci (approved):

samcli/lib/init/default_samconfig.py
```diff
@@ -0,0 +1,80 @@
...
+        self._config = SamConfig(path)
+        self._defaults = [
+            Default(
+                writing_type=WritingType.Both,
+                key="stack_name",
+                value=f"{self._project_name}",
+                command=[f"{DEFAULT_GLOBAL_CMDNAME}"],
+            ),
+            Default(writing_type=WritingType.ZIP, key="cached", value=True, command=["build"]),
+            Default(writing_type=WritingType.Both, key="parallel", value=True, command=["build"]),
```
moelasmar, line 43: not sue if we should make parallel, and cached as default values for sam build?

moelasmar (commented):
in general it looks Ok, I just left some minor comments/questions

samcli/lib/init/__init__.py
sriram-mv: we generate a file and add it to the generated project, post cookie-cutter.

samcli/lib/init/__init__.py
```diff
@@ -145,3 +150,30 @@ def _enable_application_insights(application_insights: bool, output_dir: str, na
...
+        string representing the output directory of the template
+    name:
+        string representing the name of the application
+    """
+    # NOTE(sriram-mv): If this is coming from a `--location` without a corresponding project name,
+    # do not attempt to insert a samconfig.toml for a straight cookie-cutter init.
+    if not name:
+        return
+    project_root_path = Path(output_dir, name) if name else Path(output_dir)
+    if Path(project_root_path, "samconfig.toml").is_file():
```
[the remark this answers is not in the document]
sriram-mv: that is a fair call out, it can be changed.

samcli/lib/init/default_samconfig.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, path, package_type, project_name):
+        self._path = path
+        self._package_type = package_type
+        self._project_name = project_name
+        self._config = SamConfig(path)
+        self._defaults = [
+            Default(
+                writing_type=WritingType.Both,
+                key="stack_name",
+                value=f"{self._project_name}",
```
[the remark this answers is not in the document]
sriram-mv: this is usually the default prompt which is `sam-app` and it will be allowed. We need better prompt handling for sure though, it can be a separate PR.

samcli/lib/init/default_samconfig.py
sriram-mv: Generally should be good. Users typically build off the init templates and add new functionality as see they fit. So this will ensure performance is good as they iterate.

moelasmar (approved):

samcli/lib/init/__init__.py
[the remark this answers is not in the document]
sriram-mv: I thought about this some more. As long its not overwriting an existing file we should be okay. And if customers are using another .toml file they are already having to specify the that file since its no longer a default. Might be a good way for customers to learn about this and adopt some patterns if they choose to adopt it.

Issue #4505, opened 2022-12-20
Bug: Fail sam local gracefully for cases with inline code


praneetap:
sam local should fail gracefully in cases where lambda code is defined inline, since it doesnt support it. Look at this issue comment https://github.com/aws/aws-sam-cli/issues/790#issuecomment-1157759305

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS:
2. `sam --version`:
3. AWS region:

`Add --debug flag to command you are running`

moelasmar:
@praneetap ... do you mean only if all functions in the project defined with inline code?

praneetap:
@moelasmar just updated the issue with more details. Here is the issue comment this issue is referencing - https://github.com/aws/aws-sam-cli/issues/790#issuecomment-1157759305

hawflau:
Released with [1.74.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.74.0). Closing.

PR #4725, opened 2023-02-16
fix: Failing sam local commands for functions with inline code gracefully

hnnasit:

#4505 

Invocation of sam local commands for functions with Inline Code is not supported.

This PR adds error handling logic for when functions with inline code are invoked. It also warns the developers if inline code is detected in the template when sam local commands are run.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -270,6 +271,15 @@ def __enter__(self) -> "InvokeContext":
         if self._containers_initializing_mode == ContainersInitializationMode.EAGER:
             self._initialize_all_functions_containers()
 
+        for func in self._function_provider.get_all():
+            if func.packagetype == ZIP and func.inlinecode:
+                LOG.info(
```
mndeveci, line 276: Shouldn't this be `LOG.warning`?

samcli/local/apigw/service_error_responses.py
```diff
@@ -9,6 +9,7 @@ class ServiceErrorResponses:
     _MISSING_AUTHENTICATION = {"message": "Missing Authentication Token"}
     _LAMBDA_FAILURE = {"message": "Internal server error"}
 
+    HTTP_STATUS_CODE_501 = 501
```
mndeveci, line 12: Just curious, why we went with 501 error code?

hnnasit: Since the functionality of invoking sam local commands with inline code is not supported on our end we use 501 as the error code.

samcli/commands/local/cli_common/invoke_context.py
hnnasit: yeah LOG.warn would make sense here, I can change that.

samcli/local/apigw/service_error_responses.py
mndeveci: I was more of looking for why we choose 501 out of other 5xx error codes. But I saw that 501 is not implemented, which makes sense for this one.

samcli/commands/local/cli_common/invoke_context.py
hnnasit: Noticed LOG.warn is a deprecated method, changed it to LOG.warning just like your suggestion.

mndeveci (approved):

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -270,6 +271,15 @@ def __enter__(self) -> "InvokeContext":
         if self._containers_initializing_mode == ContainersInitializationMode.EAGER:
             self._initialize_all_functions_containers()
 
+        for func in self._function_provider.get_all():
+            if func.packagetype == ZIP and func.inlinecode:
+                LOG.warning(
+                    "Warning: Inline code found for function %s."
+                    " Invocation of inline code is not supported for sam local commands.",
```
sriram-mv, line 278: let's not reference the binary here directly. as it would be samdev, sam-beta, sam-nightly. You could get the cli context and try to figure out the command name (could be a useful helper function).

samcli/commands/local/lib/local_lambda.py
```diff
@@ -120,6 +121,11 @@ def invoke(
 
         LOG.debug("Found one Lambda function with name '%s'", function_identifier)
         if function.packagetype == ZIP:
+            if function.inlinecode:
+                raise UnsupportedInlineCodeError(
+                    "Inline code is not supported for sam local commands."
```
sriram-mv, line 126: Same here

samcli/local/apigw/local_apigw_service.py
```diff
@@ -361,6 +363,10 @@ def _request_handler(self, **kwargs):
             self.lambda_runner.invoke(route.function_name, event, stdout=stdout_stream_writer, stderr=self.stderr)
         except FunctionNotFound:
             return ServiceErrorResponses.lambda_not_found_response()
+        except UnsupportedInlineCodeError:
+            return ServiceErrorResponses.not_implemented_locally(
+                "Inline code is not supported for sam local commands. Please write your code in a separate file."
```
sriram-mv, line 368: Same here

samcli/local/lambda_service/local_lambda_invoke_service.py
```diff
@@ -166,6 +167,10 @@ def _invoke_request_handler(self, function_name):
         except FunctionNotFound:
             LOG.debug("%s was not found to invoke.", function_name)
             return LambdaErrorResponses.resource_not_found(function_name)
+        except UnsupportedInlineCodeError:
+            return LambdaErrorResponses.not_implemented_locally(
+                "Inline code is not supported for sam local commands. Please write your code in a separate file."
```
sriram-mv, line 172: Same here

samcli/commands/local/cli_common/invoke_context.py
hnnasit: I agree, we distribute the binaries sam and sam-nightly but hard code it to sam in the print statements. Same goes for CI jobs where we use sam-beta for the unit/integ tests. I can create a separate PR for that helper function and keep this PR clean as this change would require updating the hard coded binary name for the related unit/integration tests for all the commands as well.

sriram-mv (approved):

Issue #4681, opened 2023-02-09
Bug: Key Error when waiting for changeset/Max attempts exceeded


tirumerla:
### Description:

Hello Folks,

We are running into issues during "wait_for_changeset". Our changeset creation takes `~20 minutes` and we started seeing this issue from past 2 months mainly because we added more services as part of our nested stacks which increased the changeset creation time. Based on the previous bug reports here thought it was some syntactical issues in our templates but no issues were raised with `cfn lint` also changeset creation doesn't fail and will eventual gets to "CREATE_COMPLETE" so definitely not syntax issues in the templates.

This issue is mainly due to the [delay that was set](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/deploy/deployer.py#L311) to `5 seconds` and [assuming the changeset creation is quick](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/deploy/deployer.py#L310). With delay set to `5 seconds` and max attempts set to `120` if the changeset creation takes more than `> 10 mins` waiting on changeset will fail. 

I could be missing something/may be wrong here but that's what i'm thinking.

Prior to sam cli version `v1.72.0` we were seeing "Max attempts exceeded"

```
Waiting for changeset to be created..
Uploading to testing/5e2a0b45a2ccfed7bf77d88d96d85453.template  207618 / 207618  (100.00%)[4470] Failed to execute script '__main__' due to unhandled exception!

Traceback (most recent call last):
  File "samcli/lib/deploy/deployer.py", line 313, in wait_for_changeset
  File "botocore/waiter.py", line 54, in wait
  File "botocore/waiter.py", line 366, in wait
botocore.exceptions.WaiterError: Waiter ChangeSetCreateComplete failed: Max attempts exceeded

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "samcli/__main__.py", line 12, in <module>
  File "click/core.py", line 1130, in __call__
  File "click/core.py", line 1055, in main
  File "click/core.py", line 1657, in invoke
  File "click/core.py", line 1404, in invoke
  File "click/core.py", line 760, in invoke
  File "samcli/lib/cli_validation/image_repository_validation.py", line 92, in wrapped
  File "click/decorators.py", line 84, in new_func
  File "click/core.py", line 760, in invoke
  File "samcli/lib/telemetry/metric.py", line 175, in wrapped
  File "samcli/lib/telemetry/metric.py", line 150, in wrapped
[...]

PR #4729, opened 2023-02-17
Allow re-use of poll delay env variable on wait changeset

tirumerla:
#4681 

- Currently `wait_for_changeset` is successful if takes `<~10mins` ( since Delay is set to `5 secs` ). If there are any changesets `>10 mins` they will fail ( which is the case in our scenario ).

- Allow usage of `SAM_CLI_POLL_DELAY` env variable to set more delay than what is right now.

- If users set `SAM_CLI_POLL_DELAY` accidentally on a small changeset it will still take that many seconds to wait before it can continue.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/deploy/deployer.py
```diff
@@ -307,8 +307,13 @@ def wait_for_changeset(self, changeset_id, stack_name):
 
         # Wait for changeset to be created
         waiter = self._client.get_waiter("change_set_create_complete")
-        # Poll every 5 seconds. Changeset creation should be fast
+        # If there is a client sleep, use it as the delay or use the default 5 seconds
         waiter_config = {"Delay": 5}
+        if ( 
+            self.client_sleep 
+            and self.client_sleep > waiter_config["Delay"]
```
jfuss, line 314: Curious on why you decided to default to it this way. `client_sleep` is set by 
```
try:
  self.client_sleep = float(client_sleep)
except ValueError:
  self.client_sleep = DEFAULT_CLIENT_SLEEP
if self.client_sleep <= 0:
  self.client_sleep = DEFAULT_CLIENT_SLEEP
```

So why not just set it to `self.client_sleep` here?

jfuss (changes requested):

tests/unit/lib/deploy/test_deployer.py
```diff
@@ -342,6 +348,13 @@ def test_wait_for_changeset(self):
         self.deployer._client.get_waiter = MagicMock(return_value=MockChangesetWaiter())
         self.deployer.wait_for_changeset("test-id", "test-stack")
 
+    @patch("os.environ", {**os.environ, "SAM_CLI_POLL_DELAY": 10})
+    def test_wait_for_changeset_client_sleep(self):
+        deployer = Deployer(MagicMock().client("cloudformation"), client_sleep=os.getenv("SAM_CLI_POLL_DELAY", 0.5))
+        deployer._client.get_waiter = MagicMock(return_value=MockChangesetWaiter())
+        deployer.wait_for_changeset("test-id", "test-stack")
+        self.assertEqual(deployer.client_sleep, 10)
+
```
lucashuy, line 357: Can we also add a test to test that the `Delay` value in the waiter config is set correctly? This can be done by checking if the waiter's `.wait` call was called with the correct parameters.

lucashuy (dismissed):

samcli/lib/deploy/deployer.py
tirumerla: Thanks @jfuss for the reviewing it quickly. I'm basically reading the `self.client_sleep` whatever the value that's getting it into `wait_for_changeset` and verifying with the previously set waiter_config delay value ( 5 seconds ) i think 5 seconds is reasonable for polling and if we use `DEFAULT_CLIENT_SLEEP` of `0.5` it may be too short and might exhaust max attempts quickly. Does this makes sense to you?

tests/unit/lib/deploy/test_deployer.py
tirumerla: Thanks @lucashuy for the review. Yeah definitely makes sense. I will push that change.

tests/unit/lib/deploy/test_deployer.py
```diff
@@ -90,6 +91,11 @@ def test_deployer_init_default_sleep(self):
         deployer = Deployer(MagicMock().client("cloudformation"))
         self.assertEqual(deployer.client_sleep, 0.5)
 
+    @patch("os.environ", {**os.environ, "SAM_CLI_POLL_DELAY": 30})
+    def test_deployer_sam_cli_poll_delay(self):
+        deployer = Deployer(MagicMock().client("cloudformation"), client_sleep=os.getenv("SAM_CLI_POLL_DELAY", 0.5))
+        self.assertEqual(deployer.client_sleep, 30)
+
```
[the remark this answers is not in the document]
tirumerla: @lucashuy makes sense. I removed it from here and made a minor change in command unit to make sure we are capturing the os env variable since it already captures the tests related to `SAM_CLI_POLL_DELAY`

samcli/lib/deploy/deployer.py
jfuss: Sounds reasonable but I think we should just default to `self.client_sleep`. Even if the customer set this to shorter, we should respect that. It can be confusing to someone, if we put a lower limit here. It's likely people will only ever increase but I would rather not lock them into a "not smaller than" time period.

tirumerla: Makes sense to me. Just updated to use `self.client_sleep` by default.

jfuss (approved):
Thanks for handling all the feedback! Can't wait to see this ship.

jfuss:
@tirumerla Looking at the `make-pr` build, seems like files are not formatted. Could you run `make black` and push an update? That should auto-fix the formatting.

tirumerla:
> @tirumerla Looking at the `make-pr` build, seems like files are not formatted. Could you run `make black` and push an update? That should auto-fix the formatting.

Thanks just fixed it!

tirumerla:
@jfuss do you think we can merge this?

lucashuy:
Thanks for raising this PR! Everything looks good, I just had a general concern that by changing the default wait time from 5 to `client_sleep` (which defaults to 0.5 if not defined) might introduce breaking behaviour to existing customers since we go from a max of 10 minutes to 1 minute. There is a way out (passing the env variable) but customers may not know about this variable. Mainly raising this since something similar happened with `sam delete`.

tirumerla:
> Thanks for raising this PR! Everything looks good, I just had a general concern that by changing the default wait time from 5 to `client_sleep` (which defaults to 0.5 if not defined) might introduce breaking behaviour to existing customers since we go from a max of 10 minutes to 1 minute. There is a way out (passing the env variable) but customers may not know about this variable. Mainly raising this since something similar happened with `sam delete`.

Thanks @lucashuy. Yep this is something i thought of initially see - https://github.com/aws/aws-sam-cli/pull/4729#discussion_r1113493913 but after @jfuss comment it made sense to allow user to explicitly set the env. variable. If we explicitly specify this breaking change ( updating documentation etc.. ) i think it should be ok. 
What do you think?

sriram-mv (approved):

jfuss:
@tirumerla Any idea what happened to the file diff? There are tons of whitespace changes. Just trying to sort that out. Maybe this was part of us moving to ruff and this is reverting somethings (accidentally)?

Also apologize on the delay, I lost track of this in the shuffle of other things I was focusing on.

tirumerla:
> @tirumerla Any idea what happened to the file diff? There are tons of whitespace changes. Just trying to sort that out. Maybe this was part of us moving to ruff and this is reverting somethings (accidentally)?

Yeah these are the formatting changes i see when i did `make black` locally


> Also apologize on the delay, I lost track of this in the shuffle of other things I was focusing on.

Not a problem.

jfuss:
> Yeah these are the formatting changes i see when i did make black locally

Weird. I wouldn't expect additional files to be formatted because we pin `black`. Everything pass though, so maybe something happened along the way.

jfuss:
@tirumerla Thanks for submitting the patch and reporting the issue. This will go out with our next release. We don't have strict release schedules so I am not sure when the patch will release. We do releases pretty often, so it shouldn't be too long.

PR #4730, opened 2023-02-17
Default to ARM64 when running on Arm hosts

AGSaidi:
If the function architecture isn't specified and we're running on an Arm platform assume we should be trying to run the function for ARM64

Fixes #4708

#4708

Today if the architecture isn't set on an Arm based platform sam-cli tries to use the x86 runner which won't work

If the architecture isn't set and the host in Arm (linux-aarch64 or mac os-arm64) it assumes arm64 is the intent

When running local invoke on a Arm platform if the function architecture isn't specified it defaults to the architecture of host

#### Mandatory Checklist

- [ NA] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [NA ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write/update unit tests
- [NA] Write/update functional tests if needed
- [X] `make pr` passes
- [NA] Write documentation


jfuss:
Left a comment on the Discussion post, but I don't think this is the right direction for SAM CLI to be taking.

lucashuy:
Thanks for raising this PR, the default behaviour for Cloudformation/Lambda is to set the architecture to x86 if none was provided. Changing the default behaviour to be set depending on the host machine's architecture might introduce some breaking changes for customers. We'll have to discuss this with the team to see where to go with this.

mndeveci:
Thanks for your PR!

As it is been discussed above an in the discussion thread, setting Lambda's architecture has no relation between developer machine's architecture. Moreover, this might lead to an invalid state where your local invocation will succeed but your actual deployment would fail due to different architectures that is been changed on the fly for `sam local` commands.

We would recommend using `sam sync` until the actual root cause have been identified and fixed.

Appreciate your understanding!

PR #4731, opened 2023-02-17
If a function is created without an architecture on an arm host assume arm

AGSaidi:
If someone is running the sam cli on an Arm host, assume they actually want an arm-based function if none is provided

N/A

It's not necessary, but today if someone runs the sam cli and doesn't specify the architecture they can create a function they can't locally test. 

It detects the architecture of the host and defaults the cli to create a function of that type if none is specified

It changes behavior of the command line options defaults which previously always defaulted to x86

#### Mandatory Checklist

- [NA] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [NA] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write/update unit tests
- [NA] Write/update functional tests if needed
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/init_flow_helpers.py
+++ b/samcli/commands/init/init_flow_helpers.py
@@ -4,8 +4,9 @@ Init flow based helper functions
 import logging
 import functools
 import re
+from platform import machine
 
-from samcli.lib.utils.architecture import X86_64
+from samcli.lib.utils.architecture import X86_64, ARM64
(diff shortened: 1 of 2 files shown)


jfuss:
SAM CLI defaults to `x86` because that is now AWS Lambda behaves. If you do not provide `Architecture` AWS Lambda defaults to `x86`. Please open an issue first before implementing features.

Closing

PR #4732, opened 2023-02-17
ruff: bye-bye pylint

sriram-mv:
N/A

Get developer time back, ruff is insanely fast. Takes less than a second to run compared to pylint which is in the order of minutes.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


pyproject.toml
```diff
@@ -2,6 +2,14 @@
 requires = ["setuptools", "wheel"]  # PEP 508 specifications.
 
 
+[tool.ruff]
```
jfuss, line 5: You may want to add https://github.com/aws/aws-lambda-builders/blob/develop/pyproject.toml#L4 here too. Pylint is off by default.

sriram-mv: Good to know!

aahung: 😎  sam-t is away ahead of cli regarding how many rules it comply: https://github.com/aws/serverless-application-model/blob/develop/ruff.toml#L4

Make.ps1
```diff
@@ -83,6 +83,7 @@ param (
 )
 
 function Init {
+    pip install -e '.[pre-dev]'
```
torresxb1, line 86: curious about this change. I think our Makefile doesn't install the pre-dev ones: https://github.com/aws/aws-sam-cli/blob/e0d02edad389e4933fb550144ed9e9e3cc81e5db/Makefile#L6

torresxb1, line 86: curious about this change. I think our Makefile doesn't install the pre-dev ones: https://github.com/aws/aws-sam-cli/blob/e0d02edad389e4933fb550144ed9e9e3cc81e5db/Makefile#L6

pyproject.toml
```diff
@@ -1,6 +1,26 @@
...
+
+[tool.ruff.pylint]
+max-branches = 25
+max-returns = 8
+max-statements = 80
+
+[tool.ruff.per-file-ignores]
+"__init__.py" = ["F401", "E501"]
+"integration_uri.py" = ["E501"] # ARNs are long.
+"app.py" = ["E501"] # Doc links are long.
```
torresxb1, line 23: I feel like we should probably do per-line ignores if possible, rather than the whole file

requirements/pre-dev.txt
```diff
@@ -1 +1 @@
-pylint>=2.13.0,<3
+ruff
```
aahung, line 1: Ruff is not very stable and keeps adding rules. We should try to ping the version otherwise once it adds a new rule the build might fail

torresxb1 (approved):

jfuss (approved):

requirements/pre-dev.txt
sriram-mv: Any build failures you have experienced? I'm gonna pin to 0.0.247 in that case.

aahung: they added `too-many-arguments` in the past few weeks and when we upgrade there were many violations in sam-t codebase

Make.ps1
sriram-mv: I don't believe we have as many windows devs, so this might have gotten missed out, so I brought back the pre-dev step in there.

pyproject.toml
sriram-mv: Ack. its just those files that presented it here, so went with those files.

PR #4739, opened 2023-02-21
feat: Add template comparison logic as infra sync utility

qingchm:
N.A.

In order to make infra sync faster, a template comparison utility can help with checking if the template has changed from previous deployment and conditionally skip infra sync.

It add a infra sync executor class with a compare template method, which compares templates with the removal some fields in comparison

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Parameters
+        ----------
+        resource_logical_id: str
+            Logical ID of the resource
+        resource_type: str
+            Resource type
+        resource_dict: Dict
+            The resource level dict containing Properties field
+        processed_resources: set
```
hnnasit, line 226: Curious why we pass in processed_resources parameter for this function? Does it make sense to update the processed_resources set in `_remove_unnecessary_fields` function and change the return type of `_remove_resource_field` function to be logical_id of the removed resource? This way processed_resources variable doesn't need to be passed to `_remove_resource_field` function.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+    AWS_LAMBDA_FUNCTION: {"Code": ["ImageUri", "S3Bucket", "S3Key", "S3ObjectVersion"]},
+    AWS_SERVERLESS_LAYERVERSION: ["ContentUri"],
+    AWS_LAMBDA_LAYERVERSION: ["Content"],
+    AWS_SERVERLESS_API: ["DefinitionBody"],
+    AWS_APIGATEWAY_RESTAPI: ["BodyS3Location"],
+    AWS_SERVERLESS_HTTPAPI: ["DefinitionUri"],
+    AWS_APIGATEWAY_V2_API: ["BodyS3Location"],
+    AWS_SERVERLESS_STATEMACHINE: ["DefinitionUri"],
+    AWS_STEPFUNCTIONS_STATEMACHINE: ["DefinitionS3Location"],
+}
```
mndeveci, line 44: I would recommend keeping same types of values in a dictionary. This is hard to maintain and use.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+        deploy_context : DeployContext
+            DeployContext
+        """
+        self._build_context = build_context
+        self._package_context = package_context
+        self._deploy_context = deploy_context
+
+        session = Session(profile_name=self._deploy_context.profile, region_name=self._deploy_context.region)
+        self._cfn_client = session.client("cloudformation")
+        self._s3_client = session.client("s3")
```
mndeveci, line 74: Please use utility methods to create boto clients: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/boto_utils.py#L44

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+            if local_template_path.startswith("https://"):
+                parsed_s3_location = re.search(r"https:\/\/[^/]*\/([^/]*)\/(.*)", local_template_path)
+                if parsed_s3_location:
+                    s3_bucket = parsed_s3_location.group(1)
+                    s3_key = parsed_s3_location.group(2)
+                    s3_object = self._s3_client.get_object(Bucket=s3_bucket, Key=s3_key)
+
+                    streaming_body = s3_object.get("Body")
+                    if streaming_body:
+                        current_template = yaml_parse(streaming_body.read().decode("utf-8"))
```
mndeveci, line 105: Do we still need to pull a template from remote given that we moved comparison logic before package and after build?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+                if resource_type == "AWS::CloudFormation::Stack" or resource_type == "AWS::Serverless::Application":
+                    stack_resource_detail = self._cfn_client.describe_stack_resource(
+                        StackName=stack_name, LogicalResourceId=resource_logical_id
+                    )
+
+                    # If the nested stack is of type AWS::CloudFormation::Stack,
+                    # The template location will be under TemplateURL property
+                    # If the nested stack is of type AWS::Serverless::Application,
+                    # the template location will be under Location property
+                    template_field = "TemplateURL" if resource_type == "AWS::CloudFormation::Stack" else "Location"
```
mndeveci, line 143: Same, please use static variables rather than hardcoded string

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,253 @@
...
+                    resource_dict,
+                    processed_resources,
+                    linked_resources,
+                )
+
+            # Remove SamResourceId metadata since this metadata does not affect any cloud behaviour
+            resource_dict.get("Metadata", {}).pop("SamResourceId", None)
+            if not resource_dict.get("Metadata"):
+                resource_dict.pop("Metadata", None)
+            LOG.debug(f"Sanitizing the Metadata for resource {resource_logical_id}")
```
mndeveci, line 203: Do we know why `SamResourceId` is creating problems, can we find where it is been added and change our behaviour regarding to that? My worry is, if there are some other variables like that, we might hit the same problem later.

mndeveci (changes requested):

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,295 @@
...
+            Returns False if two templates are different
+        """
+        try:
+            # If the customer template uses a nested stack with location/template URL in S3
+            if local_template_path.startswith("https://"):
+                parsed_s3_location = re.search(r"https:\/\/[^/]*\/([^/]*)\/(.*)", local_template_path)
+                s3_bucket = parsed_s3_location.group(1)
+                s3_key = parsed_s3_location.group(2)
+                s3_object = self._s3_client.get_object(Bucket=s3_bucket, Key=s3_key)
+                current_template = yaml_parse(s3_object.get("Body").read().decode("utf-8"))
```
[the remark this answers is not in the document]
qingchm: Yes, we do. This is needed because the nested template location can be a non-local path. Even though we can ensure that the parent template path is local, we still need this logic for nested templates

samcli/lib/sync/infra_sync_executor.py
qingchm: What do you mean by this? You mean reversing the keys and values? That makes iterations harder

samcli/lib/sync/infra_sync_executor.py
qingchm: Yes we do, again nested stack location can be remote before we even build

mndeveci: I've started reviewing this early in the morning and somehow this comment made it to afternoon session even though it wasn't there when I was publishing my review. Apologies for duplication.

samcli/lib/sync/infra_sync_executor.py
qingchm:
```
        metadata_copied.pop(SAM_RESOURCE_ID_KEY, "")
        metadata_copied.pop(SAM_IS_NORMALIZED, "")
```
Seems like a value that's to be popped, but in our code base there are many hard coded references to "SamResourceId", haven't looked at which exact reference is causing the difference

samcli/lib/sync/infra_sync_executor.py
qingchm: Good call out, we can have one less parameter since we will only add one resource ID each time, so we only need to return a string

samcli/lib/sync/infra_sync_executor.py
mndeveci: `REMOVAL_MAP` dictionary has `str` as its keys, but it has either `Dict` or `List` for its values. Having different types as the values could be confusing, so I would recommend keeping one type there.

qingchm: But this is due to lambda function having one additional layer here, under code we are storing the fields as list as well. So if I'm understanding correctly you are suggesting to change the lists into empty value dicts, which will help with the type formality, but also concerns me being not readable

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,295 @@
...
+            # If the template location is local
+            else:
+                current_template = get_template_data(local_template_path)
+
+            try:
```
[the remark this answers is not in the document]
qingchm: Addressed to prevent nested try catch, only wrapped the client calls

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,295 @@
...
+            self._remove_unnecessary_fields(last_deployed_template_dict, sanitized_resources)
+
+            if last_deployed_template_dict != current_template:
+                return False
+
+            # The recursive template check for Nested stacks
+            for resource_logical_id in current_template.get("Resources", {}):
+                resource_dict = current_template.get("Resources", {}).get(resource_logical_id, {})
+                resource_type = resource_dict.get("Type")
+                if resource_type == "AWS::CloudFormation::Stack" or resource_type == "AWS::Serverless::Application":
```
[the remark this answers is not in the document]
qingchm: Addressed

samcli/lib/sync/infra_sync_executor.py
qingchm: Addressed

samcli/lib/sync/infra_sync_executor.py
qingchm: Addressed

samcli/lib/sync/infra_sync_executor.py
qingchm: Currently changed all list notations to set notation, however it is still inconsistent in terms of typing.

qingchm: I think the best way to format this would be:
```
REMOVAL_MAP = {
    AWS_SERVERLESS_FUNCTION: {"CodeUri": [], "ImageUri": []},
    AWS_LAMBDA_FUNCTION: {"Code": ["ImageUri", "S3Bucket", "S3Key", "S3ObjectVersion"]},
    AWS_SERVERLESS_LAYERVERSION: {"ContentUri": []},
    AWS_LAMBDA_LAYERVERSION: {"Content": []},
    AWS_SERVERLESS_API: {"DefinitionBody": []},
    AWS_APIGATEWAY_RESTAPI: {"BodyS3Location": []},
    AWS_SERVERLESS_HTTPAPI: {"DefinitionUri": []},
    AWS_APIGATEWAY_V2_API: {"BodyS3Location": []},
    AWS_SERVERLESS_STATEMACHINE: {"DefinitionUri": []},
    AWS_STEPFUNCTIONS_STATEMACHINE: {"DefinitionS3Location": []},
}
```
Opinions?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,284 @@
...
+        package_context : PackageContext
+            PackageContext
+        deploy_context : DeployContext
+            DeployContext
+        """
+        self._build_context = build_context
+        self._package_context = package_context
+        self._deploy_context = deploy_context
+
+        self._session = Session(profile_name=self._deploy_context.profile, region_name=self._deploy_context.region)
```
mndeveci, line 78: If we are instantiating the clients and not using `self._session` anymore, we don't need to keep it as a class field here.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,284 @@
...
+
+        Returns
+        -------
+        bool
+            Returns True if two templates are identical
+            Returns False if two templates are different
+        """
+        current_template = None
+        # If the customer template uses a nested stack with location/template URL in S3
+        if local_template_path.startswith("https://"):
```
mndeveci, line 116: Would there be a case where template will start with `s3://` notation?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,284 @@
...
+                streaming_body = s3_object.get("Body")
+                if streaming_body:
+                    current_template = yaml_parse(streaming_body.read().decode("utf-8"))
+
+        # If the template location is local
+        else:
+            current_template = get_template_data(local_template_path)
+
+        if not current_template:
+            return False
```
mndeveci, line 136: Please add some debugging log here.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,284 @@
...
+            LOG.debug("Stack with name %s does not exist on CloudFormation", stack_name, exc_info=ex)
+            return False
+
+        last_deployed_template_dict = yaml_parse(last_deployed_template_str)
+
+        sanitized_resources = self._remove_unnecessary_fields(current_template)
+        self._remove_unnecessary_fields(last_deployed_template_dict, sanitized_resources)
+
+        if last_deployed_template_dict != current_template:
+            return False
```
mndeveci, line 152: Please add some debugging log here.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,284 @@
...
+            LOG.debug("Sanitizing the Metadata for resource %s", resource_logical_id)
+
+        return sorted(list(processed_resources))
+
+    def _remove_resource_field(
+        self,
+        resource_logical_id: str,
+        resource_type: str,
+        resource_dict: Dict,
+        linked_resources: List[str] = [],
```
mndeveci, line 242: To improve the performance, this `linked_resources` can be a `Set[str]` instead of `List[str]`

mndeveci (changes requested):

samcli/lib/sync/infra_sync_executor.py
qingchm: I'll take a look at this field, if notation with s3:// is supported, I will add parsing logic here as well

qingchm: With experimentation, we found that there are three formats that the Location field can take: 1. local template path, which we will package in package step. 2. s3 URL, which is a https:// location. 3. ApplicationLocationObject, which is a dict that contains SAR application ID. I have added new logic that handles the 3rd scenario, now all possible values are covered

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,295 @@
...
+                    LOG.debug("The provided template location %s can not be found", local_template_path, exc_info=ex)
+                    return False
+
+                streaming_body = s3_object.get("Body")
+                if streaming_body:
+                    current_template = yaml_parse(streaming_body.read().decode("utf-8"))
+
+        # If the template location is local
+        else:
+            current_template = get_template_data(local_template_path)
```
mndeveci, line 135: This part until here can be moved to a separate function to improve the readability.

samcli/lib/sync/infra_sync_executor.py
qingchm: Next PR: if we decide that the comparison happens before packaging, then this logic needs to be improved because nested stack resources are not sanitized, and it's possible for a early False return due to the template comparison here before going down to the recursive check below, which defeats partially the purpose below.

qingchm: The comparison will happen after packaging, but have addressed the above concern by making a deepcopy for sanitization, also having nested stack resources as a part of sanitized resources

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,370 @@
...
+        """Constructs the sync for infra executor.
+
+        Parameters
+        ----------
+        build_context : BuildContext
+            BuildContext
+        package_context : PackageContext
+            PackageContext
+        deploy_context : DeployContext
+            DeployContext
```
mndeveci, line 77: nit: we can remove the parameter comments if we don't have anything to explain them :)

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,370 @@
...
+        except ClientError as ex:
+            LOG.debug("Stack with name %s does not exist on CloudFormation", stack_name, exc_info=ex)
+            return False
+
+        last_deployed_template_dict = yaml_parse(last_deployed_template_str)
+
+        sanitized_current_template = copy.deepcopy(current_template)
+        sanitized_last_template = copy.deepcopy(last_deployed_template_dict)
+
+        sanitized_resources = self._remove_unnecessary_fields(sanitized_current_template)
```
mndeveci, line 142: nit: should we update the method name to something related to `sanitation` given that we changed the variable names?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,370 @@
...
+            resource_dict = current_template.get("Resources", {}).get(resource_logical_id, {})
+            resource_type = resource_dict.get("Type")
+
+            if resource_type in CODE_SYNCABLE_RESOURCES:
+                last_resource_dict = last_deployed_template_dict.get("Resources", {}).get(resource_logical_id, {})
+                if resource_type == AWS_LAMBDA_FUNCTION:
+                    if not resource_dict.get("Properties", {}).get("Code", None) == last_resource_dict.get(
+                        "Properties", {}
+                    ).get("Code", None):
+                        self._code_sync_resources.add(resource_logical_id)
```
mndeveci, line 160: I think we should also check if they have local code path, otherwise we shouldn't trigger a code sync for them. Are we planning to check it in other place?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,370 @@
...
+        -------
+        Optional[str]
+            The processed resource ID
+        """
+        processed_logical_id = None
+
+        if resource_type == AWS_LAMBDA_FUNCTION:
+            for field in LAMBDA_FUNCTION_REMOVAL_MAP.get(resource_type, {}).get("Code", []):
+                if (
+                    is_local_path(resource_dict.get("Properties", {}).get("Code", {}).get(field, None))
```
mndeveci, line 291: Somewhat related to previous comment, does `is_local_path` work as expected, if we run this block after running package context?

samcli/lib/sync/infra_sync_executor.py
qingchm: I see, then we will need to use different templates, both package.template_file and package.output_template_file, one for confirming code sync, one for comparison

samcli/lib/sync/infra_sync_executor.py
qingchm: Yes, it will not make sense if we use package.output_template_file anymore, I will update this section using a two template logic, using both the built template and the packaged template

samcli/lib/sync/infra_sync_executor.py
qingchm: Updated implementation to include one additional template file

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,404 @@
...
+        session: boto3.Session
+            The session created using customer config
+
+        Returns
+        -------
+        Service client instance
+        """
+        return get_boto_client_provider_from_session_with_config(session)(client_name)
+
+    def _auto_skip_infra(
```
mndeveci, line 104: nit: let's complete the function name to `_auto_skip_infra_sync`

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,404 @@
...
+        Optional[str]
+            The processed resource ID
+        """
+        processed_logical_id = None
+
+        if resource_type == AWS_LAMBDA_FUNCTION:
+            for field in LAMBDA_FUNCTION_REMOVAL_MAP.get(resource_type, {}).get("Code", []):
+                # We sanitize only if the provided resource is local
+                if (
+                    built_resource_dict
```
mndeveci, line 323: Why we are having `built_resource_dict` in the `if` sentences, given that it can't be `None` (its type in the method definition is not `Optional`). Can we remove that to simplify the statement?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,404 @@
...
+        template_dict: Dict
+            The unprocessed template dictionary
+        linked_resources: List[str]
+            The corresponding resources in the other template that got processed
+        built_template_dict: Optional[Dict]
+            The built template dict that the paths didn't get replaced with packaged links yet
+
+        Returns
+        -------
+        List[str]
```
mndeveci, line 251: Docstring is out of date. It should be `Set[str]`

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,404 @@
...
+                    LOG.debug("Sanitized %s resource %s", resource_type, resource_logical_id)
+                    processed_resources.add(processed_resource)
+
+            # Remove SamResourceId metadata since this metadata does not affect any cloud behaviour
+            resource_dict.get("Metadata", {}).pop("SamResourceId", None)
+            if not resource_dict.get("Metadata"):
+                resource_dict.pop("Metadata", None)
+            LOG.debug("Sanitizing the Metadata for resource %s", resource_logical_id)
+
+        return cast(Set, sorted(processed_resources))
```
mndeveci, line 286: Is this `cast` redundant?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,404 @@
...
+
+            if resource_type in CODE_SYNCABLE_RESOURCES:
+                last_resource_dict = last_deployed_template_dict.get("Resources", {}).get(resource_logical_id, {})
+                resource_resolved_id = nested_prefix + resource_logical_id if nested_prefix else resource_logical_id
+
+                if resource_type == AWS_LAMBDA_FUNCTION:
+                    if not resource_dict.get("Properties", {}).get("Code", None) == last_resource_dict.get(
+                        "Properties", {}
+                    ).get("Code", None):
+                        self._code_sync_resources.add(resource_resolved_id)
```
mndeveci, line 173: Shouldn't we check that resource have a local path to add it into `self._code_sync_resources`?

qingchm: Local path is already ensured in the removal logic, if not local, we will not sanitize at all, therefore returning value None for resource_resolved_id

samcli/lib/sync/infra_sync_executor.py
qingchm: No, sorted method makes the typing to recognize as List

samcli/lib/sync/infra_sync_executor.py
qingchm: It can be none, 
```
                processed_resource = self._remove_resource_field(
                    resource_logical_id,
                    resource_type,
                    resource_dict,
                    linked_resources,
                    built_resource_dict if built_template_dict else None,
                )
```
When sanitizing for the deployed template we do not pass in the build resource dict

qingchm: built_resource_dict: Optional[Dict] = None, this is what's in the method definition

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -0,0 +1,401 @@
...
+                    continue
+                # For other scenarios, template location will be a string (local or s3 URL)
+                elif not self._auto_skip_infra_sync(
+                    resource_dict.get("Properties", {}).get(template_field),
+                    current_built_template.get("Resources", {})
+                    .get(resource_logical_id, {})
+                    .get("Properties", {})
+                    .get(template_field),
+                    stack_resource_detail.get("StackResourceDetail", {}).get("PhysicalResourceId", ""),
+                    resource_logical_id + "/",
```
hnnasit: `nested_prefix` should be included here as well for stacks with more than 2 levels of nesting.

hnnasit (approved):
Looks good to me 🎉 . Thanks for addressing all the comments patiently.

mndeveci (approved):
Really appreciate efforts to address all feedback and answer all questions!

Issue #4735, opened 2023-02-17
sam-validate fails when date is unquoted in `AWSTemplateFormatVersion: 2010-09-09`


ljacobsson:
### Description
![image](https://user-images.githubusercontent.com/7579097/219804835-1d298431-b788-42a0-bb07-2481bd0a2b4a.png)

When using third party tools, like sam-patterns-cli, evb-cli, etc to modify the SAM templates there's a conversion happening between YAML->JSON->YAML. This leaves the `2010-09-09` part unquoted.

`sam build and sam deploy` works, but `sam validate` fails with: 
```
Template schema validation reported the following errors: [AWSTemplateFormatVersion] datetime.date(2010, 9, 9) is not of type 'string', [AWSTemplateFormatVersion] datetime.date(2010, 9, 9) is not one of ['2010-09-09']
```

I'd expect sam validate to be successful if the deploy is successful

hnnasit:
This issue has been addressed by the changes released in SAM CLI 1.75.0. Please check if the fix works and feel free to open a new issue if the problem still exists.

Issue #2476, opened 2020-12-11
"sam package --use-json" throws TypeError if `AWSTemplateFormatVersion` value is not in quotes.


brysontyrrell:
### Description:

This is a default behavior of PyYAML.

If the template format version is defined as:

```yaml
AWSTemplateFormatVersion: 2010-09-09
```

And we attempt to use the `--use-json` option of `sam package`:

```text
TypeError: Object of type date is not JSON serializable
```

PyYAML loads it as a date object. The workaround is to wrap the value in quotes:

```yaml
AWSTemplateFormatVersion: "2010-09-09"
```

I've tested a fix that overrides the constructor for timestamps to strings.

```python
# samcli/yamlhelper.py

import yaml
import yaml.constructor
from yaml.resolver import ScalarNode, SequenceNode

TAG_STR = "tag:yaml.org,2002:str"

yaml.constructor.SafeConstructor.yaml_constructors['tag:yaml.org,2002:timestamp'] = \
    yaml.constructor.SafeConstructor.yaml_constructors[TAG_STR]
```

I can open a PR if this seems like a valid solution.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS 11.0
2. `sam --version`: 1.13.2

`Add --debug flag to command you are running`

moelasmar:
I can reproduce this issue using the following steps
- initiate any new project using `sam init`
- Update the template to be something like the following
```
AWSTemplateFormatVersion: 2010-09-09
Transform: AWS::Serverless-2016-10-31

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello_world/
      Handler: app.lambda_handler
      Runtime: python3.7
```
(`AWSTemplateFormatVersion: 2010-09-09`) is a valid YAML format
- run `sam package --use-json --resolve-s3`
```
% sam package --resolve-s3

	Looking for resources needed for deployment: Found!

		Managed S3 bucket: aws-sam-cli-managed-default-samclisourcebucket-iucm54o5o7dc
		A different default S3 bucket can be set in samconfig.toml
		Or by specifying --s3-bucket explicitly.
AWSTemplateFormatVersion: 2010-09-09
Transform: AWS::Serverless-2016-10-31
Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
[...]

PR #4746, opened 2023-02-21
Override the YAML constructor for timestamps to output as strings and not datetime objects

brysontyrrell:
Addresses an issue with `--use-json` throws "TypeError: Object of type date is not JSON serializable" if `AWSTemplateFormatVersion` value is not quoted.

#2476 #4735 

If the value for `AWSTemplateFormatVersion: 2010-09-09` is not in quotes the YAML parser will interpret it as a date and output a datetime object which will throw a `TypeError` when the `--use-json` argument is used.

This adds an additional constructor to `yaml_parse()` that forces any `tag:yaml.org,2002:timestamp` to output to a string.

Date processing doesn't appear to be in use internally in SAM so there does not appear to be any other impact.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


sriram-mv:
Thanks for your contribution! would you mind adding unit tests as well?

brysontyrrell:
> Thanks for your contribution! would you mind adding unit tests as well?

Unit test added.

samcli/yamlhelper.py
```diff
@@ -118,6 +118,9 @@ def yaml_parse(yamlstr) -> Dict:
         # json parser.
         return cast(Dict, json.loads(yamlstr, object_pairs_hook=OrderedDict))
     except ValueError:
+        yaml.constructor.SafeConstructor.yaml_constructors[
+            "tag:yaml.org,2002:timestamp"
```
jfuss, line 122: Nit: Can we make this a Constant like we did for `TAG_STR`?

tests/unit/test_yamlhelper.py
```diff
@@ -161,3 +163,22 @@ def test_unroll_yaml_anchors(self):
...
+            "      Handler: main.handler\n"
+            "      Runtime: python3.7\n"
+            "      CodeUri: .\n"
+            "      Timeout: 600\n"
+        )
+
+        output = yaml_parse(input_template)
+        # Raises a `TypeError` if an unquoted `AWSTemplateFormatVersion` value has been parsed to a
+        # `datetime` object and not a string by `yaml_parse` when using `--use-json` argument.
+        json.dumps(output)
```
jfuss, line 184: Nit: Can we validate the expected date is converted to a string? Not throwing an error is probably enough but would like to ensure we are at least asserting the behavior.

jfuss (changes requested):
Mostly just Nits/small asks. 

The one question that came up for me is how does CloudFormation handle this? For example, if I had a random date somewhere in my template, does CloudFormation convert it to string always or lets yaml cover to date or something more conditionally?

I ask because when I was thinking about the test, I realized this may convert all dates to a string and wonder if that is the right thing. Just want to make sure we don't convert something incorrectly and have a wider impact of this change. Generally I think it's good, just want to ensure we understand completely to avoid having to revert later on due to something unexpected. :)

brysontyrrell:
> The one question that came up for me is how does CloudFormation handle this? For example, if I had a random date somewhere in my template, does CloudFormation convert it to string always or lets yaml cover to date or something more conditionally?
> 
> I ask because when I was thinking about the test, I realized this may convert all dates to a string and wonder if that is the right thing. Just want to make sure we don't convert something incorrectly and have a wider impact of this change. Generally I think it's good, just want to ensure we understand completely to avoid having to revert later on due to something unexpected. :)

You're correct that this change casts all dates to strings. I don't know how CloudFormation handles this, but dates as an object or type don't exist in it as far as I know.

jfuss:
> You're correct that this change casts all dates to strings. I don't know how CloudFormation handles this, but dates as an object or type don't exist in it as far as I know.

That is probably the right thing here, just being generally cautious. I can't find any docs or details on this sadly. Which means we either need to test the behavior or ask someone on the team (which would be on us).

jfuss (approved):

sriram-mv (approved):
Thanks for adding the tests! Appreciate it :D

PR #4760, opened 2023-02-23
feat: update local hash of the each sync flow

mndeveci:
N/A

To support local hashing of the each sync flow. This information will be used to compare if a sync flow needs to be executed.

By storing the local hashes of each sync flow, once its execution is completed.
This also updates the `_local_sha` of the image sync flow to read its hash value from docker client with latest locally built image hash.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/flows/image_function_sync_flow.py
```diff
@@ -51,20 +49,21 @@ def __init__(
...
-        super().set_up()
-        self._ecr_client = self._boto_client("ecr")
+    def _get_docker_client(self) -> DockerClient:
         if not self._docker_client:
             self._docker_client = docker.from_env()
+        return self._docker_client
+
+    def _get_ecr_client(self) -> Any:
+        if not self._ecr_client:
+            self._ecr_client = self._boto_client("ecr")
```
sriram-mv, line 65: https://github.com/aws/aws-sam-cli/blob/93462eea9b82d45e386c334490df87a653912594/samcli/lib/utils/boto_utils.py#L15

Should we pass in a config? Also worth noting that we may be missing it in other places.

mndeveci: We are already generating a client with config, see call trace below;
- [`self._boto_client`](https://github.com/mndeveci/aws-sam-cli/blob/update_local_hash/samcli/lib/sync/sync_flow.py#L96)
- [`get_boto_client_provider_from_session_with_config`](https://github.com/mndeveci/aws-sam-cli/blob/update_local_hash/samcli/lib/utils/boto_utils.py#L44)

sriram-mv (approved):

hnnasit (approved):
LGTM!

samcli/lib/sync/sync_flow.py
```diff
@@ -111,6 +111,13 @@ def gather_resources(self) -> None:
         """
         raise NotImplementedError("gather_resources")
 
+    def _update_local_hash(self) -> None:
```
qingchm, line 114: Are we missing unit test here?

qingchm: NIT: docstring for method

samcli/lib/sync/flows/image_function_sync_flow.py
qingchm: NIT: docstring for methods `_get_ecr_client` and `_get_docker_client`

samcli/lib/sync/flows/image_function_sync_flow.py
```diff
@@ -82,7 +81,14 @@ def gather_resources(self) -> None:
         )
         self._image_name = builder.build().artifacts.get(self._function_identifier)
         if self._image_name:
-            self._local_sha = str_checksum(self._image_name)
+            self._local_sha = self._get_local_image_id(self._image_name)
+
+    def _get_local_image_id(self, image: str) -> Optional[str]:
+        """Returns the local hash of the image"""
+        docker_img = self._get_docker_client().images.get(image)
```
qingchm, line 88: Note for other reviewers: we do not need exception handling here because a failed execution of the sync flow is wrapped by an exception handler that will automatically trigger infra sync, and will not break --watch process by design

samcli/lib/sync/sync_flow.py
```diff
@@ -111,6 +111,13 @@ def gather_resources(self) -> None:
...
+        if not self._local_sha:
+            LOG.debug("%sNo local hash is configured, skipping to update local hash", self.log_prefix)
+            return
+
+        self._sync_context.update_resource_sync_state(self.sync_state_identifier, self._local_sha)
```
qingchm, line 119: Think I missed this in the previous review, but I think the sync context class needs a class-wide thread lock on toml reads and writes to make things atomic, so that we don't skip by accident due to race condition

tests/unit/lib/sync/test_sync_flow.py
```diff
@@ -1,4 +1,7 @@
 from pathlib import Path
+
```
qingchm, line 2: Let's remove line 2 to 4, seems like an unneeded auto-import from IDE

mndeveci: Good catch, addressed!

qingchm (approved):

PR #4762, opened 2023-02-23
feat: Added identity source validation in request handling

lucashuy:

N/A

Adds identity source validation when visiting the route URL if the route has a Lambda Authorizer linked.

Adds a check after CORS, but before creating Lambda events to validate that the identity sources are valid/present in the headers, query parameters, context, and stage variables.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -272,6 +238,9 @@ def _add_catch_all_path(self, methods, path, route):
                 payload_format_version=route.payload_format_version,
                 is_default_route=True,
                 stack_path=route.stack_path,
+                authorizer_name=route.authorizer_name,
+                authorizer_object=route.authorizer_object,
+                use_default_authorizer=route.use_default_authorizer,
```
lucashuy, line 243: Previously, the catch all (default) path did not copy over the authorizers, this fixes that.

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -506,6 +512,57 @@ def test_get_current_route_keyerror(self):
...
+        route = self.api_gateway_route
+        route.authorizer_object = LambdaAuthorizer("", "", "", [], "")
+
+        mocked_id_source_obj = Mock()
+        mocked_id_source_obj.is_valid = Mock(return_value=is_valid)
+        route.authorizer_object.identity_sources = [mocked_id_source_obj]
+
+        # create a dummy Flask app to populate the request object with testing data
+        with flask.Flask(__name__).test_request_context():
+            self.assertEqual(self.api_service._valid_identity_sources(route), is_valid)
```
lucashuy, line 562: This populates the `request` global object with testing/dummy data that Flask sets. This shouldn't change any behaviour since the `_valid_identity_sources()` does not call any methods that use the `request` global, nor does it actually use any variables in the request (the calls are mocked out).

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @abstractmethod
+    def is_valid(self, **kwargs) -> bool:
+        """
+        Validates if the identity source is present
+        """
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, IdentitySource):
+            return NotImplemented
```
moelasmar, line 31: why not to return false ?

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -0,0 +1,254 @@
...
+            "$stageVariables.": StageVariableIdentitySource,
+        }
+
+        self._identity_sources_raw = identity_sources
+        self._identity_sources = []
+
+        for identity_source in self._identity_sources_raw:
+            for prefix, identity_source_object in identity_source_type.items():
+                if identity_source.startswith(prefix):
+                    # get the stuff after the prefix
```
moelasmar, line 240: what should we do if the provided identity source is not one of the known Identity sources?

samcli/local/apigw/authorizers/lambda_authorizer.py
lucashuy: I had actually folded this into a single statement on the line just below this one but changed it to what `mypy` suggested to do instead (return NotImplemented). Not sure what I did to trigger that, but looks like if I add that back to the return at the end of this method it is fine now.

Will add this back to the other checks.

samcli/local/apigw/authorizers/lambda_authorizer.py
lucashuy: As of right now, it doesn't do anything if there is an identity source that doesn't match the above table. In the instances where we create Lambda Authorizers, identity sources are already validated before being passed in, however this method could be a bit more defensive.

I'll add some checks to validate that the incoming identity sources are valid.

samcli/local/apigw/authorizers/lambda_authorizer.py
sriram-mv: we could this without ABCs too if it makes it simpler.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -315,6 +284,77 @@ def _construct_error_handling(self):
...
+
+        if not isinstance(lambda_auth, LambdaAuthorizer):
+            return False
+
+        identity_sources = lambda_auth.identity_sources
+
+        endpoint = PathConverter.convert_path_to_api_gateway(request.endpoint)
+        method = request.method
+
+        if lambda_auth.payload_version == LambdaAuthorizer.PAYLOAD_V1:
```
sriram-mv, line 311: the if else could just be two functions, esp if one branch is v1 and other is v2.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -315,6 +284,77 @@ def _construct_error_handling(self):
...
+            ).to_dict()
+
+        kwargs = {
+            "headers": request.headers,
+            "querystring": request.query_string.decode("utf-8"),
+            "context": context,
+            "stageVariables": self.api.stage_variables,
+        }
+
+        for validator in identity_sources:
```
sriram-mv, line 352: I like this pattern :)

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -0,0 +1,254 @@
...
+                    # and create the corresponding identity source object
+                    property = identity_source[len(prefix) :]
```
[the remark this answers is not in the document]
lucashuy: I don't have any immediate ideas on making this less magic. It would be easy to explode on a period and grab the last element, however the identity source itself could have a period too (`my.cool.header` is technically valid)

samcli/local/apigw/route.py
```diff
@@ -0,0 +1,85 @@
+"""
+Route definition for local start-api
+"""
```
lucashuy, line 3: Moved the `Route` class definition since it didn't need to be in the `local_apigw_service.py` file and this helps with refactoring the file.

sriram-mv (approved):

moelasmar (approved):

PR #4781, opened 2023-02-28
fix: Remove the logic that stores datetime information into packaged ZIP

qingchm:
With the latest make ZIP with permission change, we improved ZIP information storage with an additional field `date_time` into the ZIP file. Without this field, the contents of the lambda package will have a default last modified date of 1980. This is the default behaviour from python's zipfile.ZipInfo library: https://docs.python.org/3/library/zipfile.html. However, adding this field is breaking our ZIP hash comparison that sync flow hashing is based upon.

With the date_time, every time we make a package ZIP for lambda functions, the ZIP will have a different hash from the previous package even if the contents of the ZIP did not change. This means the sync flow hash comparison logic for ZIP functions will not work as each time we will generate a ZIP with a new hash.

By removing the `date_time` field, the sync flow compare remote caching logic will be restored, we can again skip sync flows that do not change the lambda package ZIP.

The downloaded package will again have some files showing last modified date of 1980.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/package/utils.py
+++ b/samcli/lib/package/utils.py
@@ -281,7 +281,7 @@ def make_zip_with_permissions(file_name, source_root, permission_mappers: List[P
                                 info = permission_mapper.apply(info)
                             # ZIP date time can be set to the last time the zip content was modified using this logic.
                             # info.date_time = time.localtime()[0:6]
-            
+
                             # If the date time above is added, the caching logic that compares ZIP files sha will break.
                             # However, without this field, contents of the zip file will have a last modified date 1980
                             # because python's zipfile.ZipInfo is set to: https://docs.python.org/3/library/zipfile.html.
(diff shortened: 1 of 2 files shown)


samcli/lib/package/utils.py
```diff
@@ -280,8 +279,12 @@ def make_zip_with_permissions(file_name, source_root, permission_mappers: List[P
                             info.external_attr = os.stat(full_path).st_mode << 16
                             for permission_mapper in permission_mappers:
                                 info = permission_mapper.apply(info)
-                            # Set current time to be the last time the zip content was modified.
-                            info.date_time = time.localtime()[0:6]
+                            # ZIP date time can be set to the last time the zip content was modified using this logic.
+                            # info.date_time = time.localtime()[0:6]
+
+                            # If the date time above is added, the caching logic that compares ZIP files sha will break.
```
sriram-mv, line 285: Can we be more specific on the caching logic and where it is used.

sriram-mv (approved):

samcli/lib/package/utils.py
sriram-mv: Could we also add a unit test here?

mndeveci (approved):

samcli/lib/package/utils.py
qingchm: Updated with a new unit test

PR #4788, opened 2023-03-01
test: end-to-end test suite

mildaniel:
We currently have no visibility into what happens post-deploy, and how commands work together. If there is an issue such as invalid file permissions with the deployed Lambda package, we wouldn't be aware through our current tests.

The corresponding CI/CD configuration will be completed in an upcoming PR.
Adds an end-to-end testing suite that tests command chaining and post-deploy Lambda invocation.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/end_to_end/end_to_end_base.py
```diff
@@ -0,0 +1,127 @@
...
+            ]
+            self._run_tests(stages)
+
+    def default_sync_workflow(self):
+        stack_name = self._method_to_stack_name(self.id())
+        with EndToEndTestContext(self.app_name) as e2e_context:
+            self.template_path = e2e_context.template_path
+            init_command_list = self._get_init_command(e2e_context.working_dir)
+            sync_command_list = self._get_sync_command(stack_name)
+            delete_command_list = self._get_delete_command(stack_name)
```
sriram-mv, line 61: complete nit: move delete_command_list after stack_outputs_command_list to showcase sequence of events.

tests/end_to_end/test_runtimes_e2e.py
```diff
@@ -0,0 +1,31 @@
...
+    [
+        ("go1.x", "mod"),
+        ("python3.9", "pip"),
+    ],
+)
+class TestHelloWorldDefaultEndToEnd(EndToEndBase):
+    app_template = "hello-world"
+
+    def test_go_hello_world_default_workflow(self):
+        self.default_workflow()
```
sriram-mv, line 17: overall looks good, for purpose this function having an interface since this its something we call within a test, should we pass in a list/dict of assertions?

As-is, all the decisions are handed over to the helper functions, which manages stages, running them and asserting them.

sriram-mv: Some raw notes:

Maybe we can pass a configuration that is a list of things which knows commands, what kind of assertion to be done, what to assert on and the actual assertion value.

Super raw.

```
Config(
[command="", "assertEquals", "returncode", 0, extraction_function],

extraction function = lambda x: x.process.returncode
```

How to use config:

(getattr(self, attr).getattr(self, attr2), 0)

Discussion points:

* What if we used just one lambda/validation function with custom logic on assertions.

tests/end_to_end/test_runtimes_e2e.py
```diff
@@ -0,0 +1,106 @@
...
+    DefaultInitStage,
+    DefaultRemoteInvokeStage,
+    DefaultDeleteStage,
+    EndToEndBaseStage,
+    DefaultSyncStage,
+)
+from tests.testing_utils import CommandResult
+
+
+class HelloWorldValidators(TestCase):
```
sriram-mv, line 21: nit: These are not really `HelloWorldValidators` though, they are more generic than that.

tests/end_to_end/test_runtimes_e2e.py
```diff
@@ -0,0 +1,106 @@
...
+    def default_command_output_validator(self, command_result: CommandResult):
+        self.assertEqual(command_result.process.returncode, 0)
+
+    def validate_init(self, command_result: CommandResult):
```
sriram-mv, line 25: I wonder if we should split them up in to their own classes.

Example: InitValidator, BuildValidator. This way the method name could be validate and it will be consistent across all validators. And we have a base level validator which always checks for return code.

sriram-mv: It also maps with the stage validator concept that you have already.

sriram-mv (commented):
Getting a lot closer! Loving how its evolving.

tests/end_to_end/test_stages.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, validator: BaseValidator, command_list: Optional[List[str]] = None):
+        super().__init__()
+        self.validator = validator
+        self.command_list = command_list
+
+    def run_stage(self) -> CommandResult:
+        return run_command(self.command_list)
+
+    def validate(self, command_result: CommandResult):
+        self.validator.validate(command_result)
```
sriram-mv, line 28: ❤️

sriram-mv (approved):
Thanks for patiently addressing feedback!

lucashuy (approved):

mndeveci (approved):

Issue #4638, opened 2023-01-31
Feature request: Avoid Archiving Before Uploading Lambda if Already a Zip


jhamet:
### Describe your idea/feature/enhancement

I wish SAM CLI (`sam deploy`) would not package my Serverless function if it's already a ZIP file post `sam build`.

If this already exists (was unable to find a CLI flag that indicated this and it appeared the behavior was to archive my zip file rather than just uploading directly), would love any pointers.

### Proposal

I utilize a Makefile such that `sam build` delegates to a Make target so I can utilize our Python build system (Pants) to be responsible for building Lambda deployments. In this case, Pants will actually build the Zip file and then subsequent builds will be lightning fast given the build cache. The challenge is when it comes to deploys, SAM is actually zipping the already zipped deployment which fails at runtime after deployment. It would be ideal if SAM could avoid this if it detects a Zip file already existing in the artifacts to deploy.

torresxb1:
Thanks for bringing this to our attention! SAM CLI indeed expects to zip the contents of the artifacts directory. As a workaround, you could add a step in your make rule to unzip and delete the zip file. I'll bring it up with the team, as it would be nice to avoid having to unzip.

jhamet:
Thanks @torresxb1 . Yep that's what I do at the moment, just becomes the performance bottleneck with the duplicate efforts.

PR #4795, opened 2023-03-01
Upload ZIP archive if one already exists in deployment folder

jhamet:
#4638 

A popular build system, [Pants](https://www.pantsbuild.org/docs/awslambda-python, creates zipped Lambda deployments automatically. We currently use this in conjunction with a custom `BuildMethod` via a `Makefile` for SAM for performant builds. The challenge right now is in order for `sam deploy` to function correctly, an unarchived directory needs to exist for SAM to then archive and upload. Overall, this results in the need to unzip in the `Makefile` implementation after the build just to get rezipped again which is the bottleneck in our build/deployments (especially as the number of Lambdas grow in a single SAM template).

If there's a directory with a single zip file inside, avoid zipping.

None (that I am aware of).

#### Mandatory Checklist

- [ ✅ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jhamet:
> Seems reasonable. Question on the Makefile, if the `makefile` is editable, could we change it such that it dumps artifacts in the ARTIFACTS_DIR and then we let `sam package` handle the zipping?

The problem in my case is the [Pants build system](https://www.pantsbuild.org/docs/awslambda-python) builds a zip for deployment (not configurable to be unzipped). Thus, users of this build system need to unzip into ARTIFACTS_DIR  if they want to use SAM for deployment (which then subsequently rezips upon package/deploy).

samcli/lib/package/utils.py
```diff
@@ -176,8 +175,15 @@ def upload_local_artifacts(
...
-    # Or, pointing to a folder. Zip the folder and upload (zip_method is changed based on resource type)
     if is_local_folder(local_path):
+        files = os.listdir(local_path)
+        # Custom build methods may automatically produce a
+        # Zip.
+        if len(files) == 1:
+            archive_path = os.path.join(local_path, files[0])
+            if is_zip_file(archive_path):
+                return uploader.upload_with_dedup(archive_path)
+        # Or, pointing to a folder. Zip the folder and upload (zip_method is changed based on resource type)
```
mndeveci, line 186: Thanks for your contribution! Can we add some logging here to track this down later?

hawflau (changes requested):
Thanks for the contribution. Could you please address the import format issues, and also add some integration test cases? Thanks

jhamet:
> Thanks for the contribution. Could you please address the import format issues, and also add some integration test cases? Thanks

Yep sorry for the delay will get on it.

lucashuy:
Hi, sorry about the delayed, looks like `shutils` was removed during the refactoring of imports, which is why the tests are failing. Did you need any help with adding an integration test to test this new case?

jhamet:
> Hi, sorry about the delayed, looks like `shutils` was removed during the refactoring of imports, which is why the tests are failing. Did you need any help with adding an integration test to test this new case?

Apologies for the delay. Added back that mistake.

I am curious for some feedback on how to add an integration test for this new case. I took a gander at how the current set is constructed. Would it be https://github.com/aws/aws-sam-cli/blob/8ac861cad28bee327d2521a0ae5e591eae78c0ea/tests/integration/buildcmd/test_build_cmd.py#L1775 for example ? I presume I can "fake" a build such that it produces a Zip in build directory to test the functionality (otherwise I would need to hoist the Pants build system to the container running the test).

jfuss:
@jhamet Are you still interesting in contributing this? If not, that is ok but would like to close it then. Happy to review an update if you are still interested in wrapping this up.

hnnasit:
Hi @jhamet, since this PR has been inactive for a while I am going to close it. If you are still interested in contributing and wrapping the remaining changes, feel free to open a new PR with the changes requested and we will be happy to review the updates.

PR #4796, opened 2023-03-02
feat: Skip first infra sync if template doesn't change from previous deployment

qingchm:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/command.py
```diff
@@ -373,31 +374,49 @@ def do_cli(
...
+                                    dependency_layer,
+                                )
 
 
 def execute_infra_contexts(
     build_context: "BuildContext",
     package_context: "PackageContext",
     deploy_context: "DeployContext",
-) -> None:
+) -> Optional[Set[ResourceIdentifier]]:
```
mndeveci, line 398: I would recommend returning empty set rather than `Optional[Set]` to make implementation easier on the other side.

samcli/commands/sync/command.py
```diff
@@ -373,31 +374,49 @@ def do_cli(
...
+                            if code_sync_resources:
+                                resource_ids = [str(resource) for resource in code_sync_resources]
+
+                                LOG.info("The following resources will be code synced for an update: %s", resource_ids)
+                                execute_code_sync(
+                                    template_file,
+                                    build_context,
+                                    deploy_context,
+                                    sync_context,
+                                    resource_ids,  # type: ignore
```
mndeveci, line 388: Why we need to ignore type here?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -101,6 +101,39 @@ def _boto_client(self, client_name: str, session: Session):
...
+            Returns True if infra sync got executed
+            Returns False if infra sync got skipped
+        """
+        self._build_context.set_up()
+        self._build_context.run()
+        self._package_context.run()
+
+        # Will not combine the comparisons in order to save operation cost
+        if first_sync:
+            if self._auto_skip_infra_sync(
```
mndeveci, line 125: Should we wrap this method with try/except so that any unexpected failure will be handled by running the InfraSync?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -319,12 +352,24 @@ def _remove_resource_field(
...
                     resource_dict.get("Properties", {}).get("Code", {}).pop(field, None)
                     processed_logical_id = resource_logical_id
+                # SAM templates also accepts local paths for AWS::Lambda::Function's Code property
+                # Which will be transformed into a dict containing S3Bucket and S3Key after packaging
+                if (
+                    built_resource_dict
+                    and isinstance(built_resource_dict.get("Properties", {}).get("Code"), str)
+                    and is_local_path(built_resource_dict.get("Properties", {}).get("Code"))
+                ):
+                    resource_dict.get("Properties", {}).get("Code", {}).pop("S3Bucket", None)
```
mndeveci, line 370: nit: Bucket information might stay though 🤔

mndeveci (commented):
Super excited for this! 🥳

samcli/lib/sync/infra_sync_executor.py
qingchm: When having local path as "Code", we should remove this field so that we do proper code sync for this scenario. I think even if the bucket is different, we can still leave this for code sync to handle instead of triggering an infra sync

samcli/commands/sync/command.py
qingchm: Because click returns a type of Tuple[str], that's what we defined in the execute_code_sync, and resource_ids is a list as defined above.

samcli/commands/sync/command.py
qingchm: Having None vs an empty set can be helpful to determine if it means the sync infra executed vs. no code sync updates are required.

samcli/lib/sync/infra_sync_executor.py
qingchm: auto_skip_infra_sync already has sufficient error handling, but for safety purpose adding another layer of try/catch can be safer

samcli/commands/sync/command.py
mndeveci: I see, but in that case, I would recommend returning explicit field to indicate that. Hiding such information behind being `None` or `non-None` would make it hard to maintain in the future.

samcli/lib/sync/infra_sync_executor.py
mndeveci: Yeap. Maybe we can move some of the error handling stuff here so that we can remove some redundant ones from method below.

samcli/lib/sync/infra_sync_executor.py
mndeveci: Oh got it, yeap makes sense now.

samcli/commands/sync/command.py
mndeveci: In that case, we should convert into the type that we are willing to use instead of ignoring the type checking (which might lead into problems in the future).

qingchm: We have sufficient unit tests to prevent such risks, I really don't see the benefit of having an additional operation here for the type conversion

samcli/commands/sync/command.py
qingchm: AFAIK this is a common behaviour in our code base to check empty list/dict vs. None for such three-way possible returns, for better readability a can return an infra sync response object here, which has fields indicating such differences

samcli/lib/sync/infra_sync_executor.py
qingchm: One refactoring that I can think of that moves everything outside would be to catch one type of most probable exception outside and provide a useful logging message, such as stack does not exist exception. However the only disadvantage of such would be that tracing down other uncommon exceptions would be a bit harder as we will only print out the exact exception message without additional information

PR #4798, opened 2023-03-02
feat: Event construction refactor

lucashuy:

N/A

The `local_apigw_service.py` file is too large. 

Moved the event generation logic to it's own module.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/event_constructor.py
+++ b/samcli/local/apigw/event_constructor.py
@@ -1,0 +1,327 @@
+"""
+Lambda event construction and generation
+"""
+
+import base64
+import json
+import logging
+from datetime import datetime
+from time import time
+
+from samcli.local.apigw.path_converter import PathConverter
+from samcli.local.events.api_event import (
+    ApiGatewayLambdaEvent,
+    ApiGatewayV2LambdaEvent,
+    ContextHTTP,
+    ContextIdentity,
+    RequestContext,
+    RequestContextV2,
+)
+
+LOG = logging.getLogger(__name__)
+
+
+class EventConstructor:
+    @staticmethod
+    def v1_event(flask_request, port, binary_types, stage_name=None, stage_variables=None, operation_name=None) -> str:
+        """
+        Helper method that constructs the Event to be passed to Lambda
+
+        :param request flask_request: Flask Request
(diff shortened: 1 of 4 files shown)


samcli/local/apigw/local_apigw_service.py
```diff
@@ -205,6 +201,58 @@ def _construct_error_handling(self):
...
+
+            return EventConstructor.v2_event_http(
+                flask_request=flask_request,
+                port=self.port,
+                binary_types=self.api.binary_media_types,
+                stage_name=self.api.stage_name,
+                stage_variables=self.api.stage_variables,
+                route_key=route_key,
+            )
+        elif route.event_type == Route.API:
```
lucashuy, line 244: This can be changed to an `if` since we already return in the block above

sriram-mv (approved):

samcli/local/apigw/local_apigw_service.py
sriram-mv: if there are multiple ifs and if the number of cases could increase in the future, a dispatch model would be great to adopt.

```python
for x in _:
    artifact = x.apply(artifact, *args, **kwargs)
```

samcli/local/apigw/event_constructor.py
```diff
@@ -0,0 +1,327 @@
...
+    ContextHTTP,
+    ContextIdentity,
+    RequestContext,
+    RequestContextV2,
+)
+
+LOG = logging.getLogger(__name__)
+
+
+class EventConstructor:
```
torresxb1, line 24: question for thought, but not necessarily requiring any changes:
if all of the methods of this class are static methods, does this need to be a class (vs. just a module with functions)?

torresxb1 (approved):

samcli/local/apigw/event_constructor.py
lucashuy: Initially stuffed it into a class to reduce the amount of importing, but looking at it again we can just make these standalone functions since only two of these methods are exposed in other parts of the CLI

PR #4804, opened 2023-03-02
test: Configure end-to-end tests in canaries

mildaniel:
Building on #4788, this PR configures the Appveyor jobs to run the newly created end-to-end tests with the existing canaries.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -14,6 +14,7 @@ configuration:
   - DeleteIntegTesting
   - SyncIntegTesting
   - LocalIntegTesting
+  - EndToEndTesting
   # other Integration testing, Dev, regression and smoke testing
   - OtherTesting
 
@@ -337,6 +338,17 @@ for:
 
       - "pip install -e \".[dev]\""
       - sh: "pytest -vv tests/integration/local --json-report --json-report-file=TEST_REPORT-integration-local.json"
+
+  # End-to-end testing
+  -
+    matrix:
+      only:
+        - image: Ubuntu
+          configuration: EndToEndTesting
+
+    test_script:
+      - "pip install -e \".[dev]\""
+      - sh: "pytest -vv tests/end_to_end --json-report --json-report-file=TEST_REPORT-end-to-end.json"
 
   # Other testing
   -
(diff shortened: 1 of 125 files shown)


lucashuy (approved):

appveyor-ubuntu.yml
```diff
@@ -338,6 +339,17 @@ for:
...
+  # End-to-end testing
+  -
+    matrix:
+      only:
+        - image: Ubuntu
+          configuration: EndToEndTesting
+
+    test_script:
+      - "pip install -e \".[dev]\""
+      - sh: "pytest -vv tests/end_to_end --json-report --json-report-file=TEST_REPORT-end-to-end.json"
```
mndeveci, line 351: Any chance we can run them in parallel (with providing `--dist loadscope`, which will queue tests from same class into same thread)?

mndeveci (approved):

appveyor-ubuntu.yml
mildaniel: Does it require any additional configuration other than providing that one flag?

mndeveci: We will need to replace the line above with something like;
```
pytest -vv -n N --reruns M --dist loadscope tests/end_to_end --json-report --json-report-file=TEST_REPORT-end-to-end.json
```
`N` is number of threads running in parallel, and `M` will be number of retries that we need to do for each test case in case they fail. From their documentation (https://pytest-xdist.readthedocs.io/en/stable/distribution.html);
> `--dist loadscope`: Tests are grouped by module for test functions and by class for test methods. Groups are distributed to available workers as whole units. This guarantees that all tests in a group run in the same process. This can be useful if you have expensive module-level or class-level fixtures. Grouping by class takes priority over grouping by module.

mildaniel: This took some tinkering with the init stage since the tests write to a shared resource. I was able to get it to run in parallel by locking the `~/.aws-sam/` directory using `Filelock`.

mndeveci (approved):

PR #4805, opened 2023-03-02
chore: Add class tear down for cleaning up custom credentials

qingchm:
N/A

When running init integration tests, if a specific custom credential test errors out, the custom credentials will remain and affect the other tests.

Add a class teardown method to reinforce the credential restoration.

None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/init/schemas/schemas_test_data_setup.py
+++ b/tests/integration/init/schemas/schemas_test_data_setup.py
@@ -55,6 +55,10 @@ eb-app-maven
             runner = CliRunner()
             runner.invoke(init_cmd, ["--output-dir", temp], input=user_input)
 
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls._tear_down_custom_config()
+
     def _init_custom_config(self, profile, region):
         self.config_dir = tempfile.mkdtemp()
         env = os.environ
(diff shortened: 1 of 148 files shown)


tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -55,17 +55,43 @@ def setUpClass(cls):
...
+            cls.original_cred_file = env[AWS_SHARED_CREDENTIALS_FILE]
         if AWS_PROFILE in env:
-            self.original_profile = env[AWS_PROFILE]
+            cls.original_profile = env[AWS_PROFILE]
         if AWS_DEFAULT_REGION in env:
-            self.original_region = env[AWS_DEFAULT_REGION]
+            cls.original_region = env[AWS_DEFAULT_REGION]
+
+    @classmethod
+    def tearDownClass(cls) -> None:
```
mndeveci, line 69: Can we remove `_tear_down_custom_config` or call it from here given that they have pretty similar logic?

qingchm: _tear_down_custom_config helps keep each test case's configuration clean, so I don't really want to touch that, one refactoring that I can do tho, is to apply this tear down as a test case tearDown method into every test case instead of a class method, some extra resets, but we can remove _tear_down_custom_config that way

hawflau (approved):

hnnasit (approved):

PR #4807, opened 2023-03-03
feat: Event constructors for authorizers

lucashuy:

N/A

Adds event generation for Lambda authorizers.

Reuse some of the existing constructing logic and added some authorizer specific properties to them.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/authorizers/lambda_authorizer.py
+++ b/samcli/local/apigw/authorizers/lambda_authorizer.py
@@ -3,7 +3,7 @@ Custom Lambda Authorizer class definition
 """
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import Any, List, Optional, Tuple
 from urllib.parse import parse_qsl
 
 from samcli.commands.local.lib.validators.identity_source_validator import IdentitySourceValidator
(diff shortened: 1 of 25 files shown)


samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -50,18 +66,17 @@ def is_valid(self, **kwargs) -> bool:
...
         -------
-        bool
-            True if the header is present and not empty
+        Any
+            The header if it is found, otherwise None
         """
         headers = kwargs.get("headers", {})
-
-        return bool(headers.get(self.identity_source, None))
+        return headers.get(self.identity_source, None)
```
hawflau, line 73: nit: `dict.get(some_key)` returns `None` by default if the key doesn't exists

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -50,18 +66,17 @@ def is_valid(self, **kwargs) -> bool:
...
 
 
 class QueryIdentitySource(IdentitySource):
-    def is_valid(self, **kwargs) -> bool:
+    def find_identity_value(self, **kwargs) -> Any:
```
hawflau, line 77: Can we annotate the return type with `Optional[Something]` instead of `Any`?
where `Something` is the type

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -24,10 +24,26 @@ def __init__(self, identity_source: str):
...
+
+        Returns
+        -------
+        bool:
+            True if the identity source is present
+        """
+        return self.find_identity_value(**kwargs) is not None
+
+    @abstractmethod
+    def find_identity_value(self, **kwargs) -> Any:
```
sriram-mv, line 44: why Any?

sriram-mv (approved):
Overall looks good, needs typing fixes.

samcli/local/apigw/authorizers/lambda_authorizer.py
lucashuy: Was initially done since in theory, these values could be anything, however in practice they are suppose to be string values, or missing. Will update the typing in this class

hawflau (approved):

Issue #4277, opened 2022-10-04
Please update the hash function to SHA1 or pass in  `usedforsecurity=False` flag


ARMAAN7139:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/package/packageable_resources.py", line 127, in export
    self.do_export(resource_id, resource_dict, parent_dir)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/package/packageable_resources.py", line 149, in do_export
    uploaded_url = upload_local_artifacts(
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/package/utils.py", line 171, in upload_local_artifacts
    return zip_and_upload(local_path, uploader, extension)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/package/utils.py", line 188, in zip_and_upload
    with zip_folder(local_path) as (zip_file, md5_hash):
  File "/usr/lib64/python3.9/contextlib.py", line 119, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/package/utils.py", line 210, in zip_folder
    md5hash = dir_checksum(folder_path, followlinks=True)
  File "/usr/local/lib/python3.9/site-packages/samcli/lib/utils/hash.py", line 63, in dir_checksum
    hash_generator = hashlib.md5()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
Error: Unable to upload artifact src/ referenced by CodeUri parameter of resource.
[digital envelope routines: EVP_DigestInit_ex] disabled for FIPS

sriram-mv:
Thanks for the issue! could you give us some more details such as the platform, version of aws sam cli. 

Where do you want us to pass in the flag? Additional details would be appreciated.

ARMAAN7139:
I am using RedHat8 with SAM CLI, version 1.41.0

mndeveci:
@ARMAAN7139 do you know if this will change the output of the hash function?

ARMAAN7139:
I do not think so.

jfuss:
I was just looking into this but we still support 3.7 and 3.8: https://github.com/aws/aws-sam-cli/blob/develop/setup.py#L57 So we either have to wait to drop them or do some python version checks. 

The md5 is not used for anything security related. It's used to compute a hash of the function code that we can rely on to allow us to only upload code that has changed to s3 (so optimization for caching).

[...]

PR #4812, opened 2023-03-03
fix: use sha256 where it is possible & update md5 call to include usedforsecurity=False

mndeveci:
#4277


samcli/lib/utils/hash.py
```diff
@@ -3,9 +3,21 @@
...
 import hashlib
 import os
+import sys
 from typing import Any, List, Optional, cast
 
 BLOCK_SIZE = 4096
+# earliest python version to support usedforsecurity option for hashlib.md5 is 3.9
+# https://docs.python.org/3/library/hashlib.html#hash-algorithms
+_MAJOR_PYTHON_VERSION = 3
+_MINOR_PYTHON_VERSION = 9
```
hawflau, line 13: question - Should we bump the python version we used to building installers?

tests/unit/lib/utils/test_hash.py
```diff
@@ -122,3 +123,11 @@ def test_dir_cyclic_links(self):
...
         checksum = str_checksum("Hello, World!")
         self.assertEqual(checksum, "65a8e27d8879283831b664bd8b7f0ad4")
+
+    @patch("samcli.lib.utils.hash.hashlib")
+    def test_md5_instantiation(self, patched_hashlib):
+        str_checksum("dummy-data")
+        if sys.version_info.major >= 3 & sys.version_info.minor >= 9:
+            patched_hashlib.md5.assert_called_with(usedforsecurity=False)
+        else:
+            patched_hashlib.md5.assert_called_with()
```
qingchm, line 133: Seems like python 3.9 is not compatible with this logic?

hawflau: Interesting, this branch is executed instead of above in Py3.9

lucashuy: Looks like the condition is using the bitwise `&` instead of `and`

mndeveci: Good catch, writing in different languages have messed my mind :)

qingchm (approved):

hnnasit (approved):

samcli/lib/utils/hash.py
jfuss: Most likely. In the ideal world, we are always on the latest stable version. It will cause less headaches for us in the long term for various reason. @mndeveci and I chatted about this, we likely need a task in our backlog.

jfuss (approved):

PR #4814, opened 2023-03-03
chore: Added dependabot to the internal PR label workflow

lucashuy:

N/A

dependabot PRs are being labelled as external.

Adds dependabot to our PR label workflow so that it applies the correct label to it.

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hawflau:
> dependabot PRs are being labelled as external.

Should that be the right thing?

lucashuy:
I guess that sort of depends on what we wanted to do with these PRs and whether we wanted to treat them as something we should resolve within our SLAs like other externally created PRs. If we wanted to resolve and triage these within our SLAs, then we can close this. Don't have a particular preference, so I can go either way.

lucashuy:
Closing PR to deal with dependabot PRs under SLA

PR #4815, opened 2023-03-03
feat: Skip first infra sync if template doesn't change from previous deployment

qingchm:
N/A.

This will help with sync efficiency by not executing the unnecessary infrastructure re-deployments

Use template comparison logic to check if a template can be skipped for infrastructure sync by comparing to the previous deployed version, if skipping we only queue up the necessary code syncs.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/command.py
```diff
@@ -374,31 +375,44 @@ def do_cli(
...
                             )
                         else:
-                            execute_infra_contexts(build_context, package_context, deploy_context)
+                            infra_sync_result = execute_infra_contexts(build_context, package_context, deploy_context)
+                            code_sync_resources = infra_sync_result.code_sync_resources
+
+                            if code_sync_resources:
+                                resource_ids = [str(resource) for resource in code_sync_resources]
+
+                                LOG.info("The following resources will be code synced for an update: %s", resource_ids)
```
mndeveci, line 384: Should we convert this into debug, in the case where there would be a lot of them?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -55,6 +55,27 @@
...
+
+    @property
+    def infra_sync_executed(self) -> bool:
+        """Returns a boolean indicating whether infra sync executed"""
+        return self._infra_sync_executed
+
+    @property
+    def code_sync_resources(self) -> Set[ResourceIdentifier]:
+        """Returns a set of resource identifiers that need an infra sync"""
+        return self._code_sync_resources
```
mndeveci, line 76: We could annotate this class with `@dataclass` and remove all the methods which is been added here

samcli/lib/sync/watch_manager.py
```diff
@@ -227,10 +233,34 @@ def _execute_infra_sync(self) -> None:
...
+        Parameters
+        ----------
+        resource_ids_with_code_sync: Set[ResourceIdentifier]
+            The set of resource IDs to be synced
+        """
+        for resource_id in resource_ids_with_code_sync:
+            if self._sync_flow_factory:
+                sync_flow = self._sync_flow_factory.create_sync_flow(resource_id)
+                if sync_flow:
+                    self._sync_flow_executor.add_delayed_sync_flow(sync_flow)
```
mndeveci, line 262: nit: Let's use early exit if statements to improve readability?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -55,6 +55,27 @@
 }
 
 
+class InfraSyncResult:
+    """Data class for storing infra sync result"""
+
+    _infra_sync_executed: bool
+    _code_sync_resources: Set[ResourceIdentifier]
+
+    def __init__(self, executed: bool, code_sync_resources: Set[ResourceIdentifier] = set()) -> None:
```
hnnasit, line 64: nit: add docstring for init.

samcli/lib/sync/infra_sync_executor.py
qingchm: But we need to encapsulate the fields to only be readable after instantiation, so I don't think we can remove any methods here

samcli/lib/sync/watch_manager.py
qingchm: This logic here is consistent with on_code_change() on line 278, consistency improves readability

samcli/commands/sync/command.py
qingchm: We only print this once, resource_ids is a list. We have gone through UX review with such message

qingchm: It's logical to tell customers what happens after we skip infra sync -> we queue up relevant code syncs

mndeveci: My concern is, if there would be 20 items or more, how that will be useful to read this line. And this will go away very quickly with other outputs from sync flows. That is why I thought using a debug might be useful for us to see which items are been queued up.

qingchm: Maybe we can just mention we are queuing up necessary code syncs without listing them, since the following log flows will include each sync flow's execution, and move the list print out to debug

samcli/lib/sync/watch_manager.py
mndeveci: I agree consistency improves readability, but it is not same at line 278. There is no `for` loop there (which is only required here), there is no `if self._sync_flow_factory`. These extra indentations are making this one harder to read.

qingchm: I agree we can early exit on if not self._sync_flow_factory, but for the second if since we have a for loop it doesn't really make a difference since we will need to continue on the for loop anyways

samcli/commands/sync/command.py
mndeveci: That makes sense. And we can still keep list of sync flows that we are queuing up in a debug message.

qingchm: Updated ;)

PR #4817, opened 2023-03-04
feat: only clone init templates when required

mildaniel:

N/A

AWS SAM CLI stores the cloned app templates in a local directory, but doesn't use it as a cache. The app templates repository doesn't need to be cloned every time if it is already up-to-date.

Checks if the app templates repository needs to be cloned before cloning based on latest commit hash.

The `.git` directory in the app templates folder will now be copied to the cache directory whereas before it was excluded.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/init/init_templates.py
```diff
@@ -102,6 +106,17 @@ def clone_templates_repo(self):
                 if expected_previous_clone_local_path.exists():
                     self._git_repo.local_path = expected_previous_clone_local_path
 
+    def _check_upsert_templates(self, shared_dir, cloned_folder_name):
+        cache_dir = shared_dir / cloned_folder_name
+        git_executable = self._git_repo.get_git_executable()
+        command = [git_executable, "rev-parse", "--verify", "HEAD"]
```
sriram-mv, line 112: Something to keep in mind when using capabilities is to ensure when they were released so that we don't inadvertently create a min version requirement for git.

However, for this particular command I do not believe it's recent.

sriram-mv (commented):
I like where this is going and have been thinking of doing this same improvement for a while.

It should improve perf on `sam init` especially on not so great internet connections.

samcli/commands/init/init_templates.py
mildaniel: This is part of the reason why we don't fail here if we fail to run the command. We'll just keep going with existing behaviour.

samcli/commands/init/init_templates.py
```diff
@@ -87,13 +88,16 @@ def clone_templates_repo(self):
...
+
             try:
-                self._git_repo.clone(
-                    clone_dir=shared_dir,
-                    clone_name=cloned_folder_name,
-                    replace_existing=True,
-                    commit=APP_TEMPLATES_REPO_COMMIT,
-                )
+                if to_upsert:
+                    self._git_repo.clone(
```
mndeveci, line 95: nit: too many indentations here, any possibility to early exit and reduce them?

samcli/commands/init/init_templates.py
```diff
@@ -102,6 +106,37 @@ def clone_templates_repo(self):
...
+        cloned_folder_name: str
+            Name of the directory into which the app templates will be copied
+
+        Returns
+        -------
+        bool
+            True if the cache should be updated, False otherwise
+
+        """
+        cache_dir = shared_dir / cloned_folder_name
```
mndeveci, line 127: Should we use `Path` or `os.path` to join the paths according to the different operating systems?

samcli/commands/init/init_templates.py
hawflau: so if `to_upsert` is `False`, nothing will happen in the try block?

samcli/commands/init/init_templates.py
hawflau: if `shared_dir` and `cloned_folder_name` are `Path` object, the `/` operator should support all platforms already.

samcli/commands/init/init_templates.py
```diff
@@ -102,6 +106,37 @@ def clone_templates_repo(self):
...
+        git_executable = self._git_repo.get_git_executable()
+        command = [git_executable, "rev-parse", "--verify", "HEAD"]
+        try:
+            existing_hash = check_output(command, cwd=cache_dir, stderr=STDOUT).decode("utf-8").strip()
+        except CalledProcessError:
+            LOG.debug("Unable to check existing cache hash")
```
hawflau, line 133: I wonder if we can show more info in this log about the `CalledProcessError`. Is there any message from the `CalledProcessError` that we can wrap into the debug log. Could be a useful info to us even if we don't actually throw an error here.

samcli/lib/utils/git_repo.py
```diff
@@ -70,6 +70,9 @@ def _ensure_clone_directory_exists(clone_dir: Path) -> None:
             LOG.warning("WARN: Unable to create clone directory.", exc_info=ex)
             raise
 
+    def get_git_executable(self):
```
hawflau, line 73: is this method just to "public"-tize `_git_executable`?

tests/integration/init/test_init_command.py
```diff
@@ -838,6 +839,62 @@ def test_interactive_init_default_runtime(self):
...
+            first_modification_time = self._last_modification_time(cache_dir)
+            self.assertTrue(project_directory.is_dir())
+            shutil.rmtree(project_directory)
+
+            # Run init again, get the second modification time
+            self._run_init(temp)
+            second_modification_time = self._last_modification_time(cache_dir)
+            self.assertTrue(project_directory.is_dir())
+
+            self.assertEqual(first_modification_time, second_modification_time)
```
hawflau, line 859: 👍

samcli/commands/init/init_templates.py
mndeveci: Thanks @hawflau, I didn't know this before. Looking at typing, it is a `Path` object, so it should work as expected then.

samcli/lib/utils/git_repo.py
mildaniel: Yes.. I did this while testing to avoid updating other references to `_git_executable` but I can just change that method to be public.

samcli/commands/init/init_templates.py
mildaniel: `Pathlib` overloads the `/` operator so it should handle paths correctly. This might not be the most read-able solution though so I'll make it more explicit.

samcli/commands/init/init_templates.py
mildaniel: Yeah, we can just avoid the try block all together if we don't need to be upserting the app templates repo

hawflau (approved):

mndeveci (approved):

sriram-mv (approved):

Issue #4585, opened 2023-01-17
Bug: `--on-failure DELETE` causes `Replacement type updates not supported on stack with disable-rollback`


eugene-bright:
### Description:

We are running sam with option `--on-failure DELETE` to prevent rollback failures to prevent stack from being stuck in `ROLLBACK_COMPLETE` status (we prefer automatic stack deletion). But because of that the stack is being marked with property `"DisableRollback": true` and it causes the following errors down the road:

```
Replacement type updates not supported on stack with disable-rollback. 
```

I'm pretty sure there is a way to correctly implement `--on-failure DELETE` without setting `"DisableRollback": true` .

Related to: #4164

### Steps to reproduce:

1. `sam deploy --on-failure DELETE ...`
2. change the template in a way that causes resource replacement and repeat deployment

### Observed result:

The option `"DisableRollback": true`  is set.

### Expected result:

IMHO, the option `DisableRollback` mustn't be set. 

### Additional environment details

1. OS: x86_64.ubuntu.22 (GitHub CI runner)
3. `sam --version`: SAM CLI, version 1.70.0
4. AWS region: us-east-1

torresxb1:
Thanks for bringing this to our attention! I was able to replicate this issue. Let me check with the team if this was intentional (i.e. if there is a reason or necessity for why we disable the rollback in this case), and whether we can adjust the behaviour without breaking customers.

jfuss:
@eugene-bright I am looking into this issue and from what I understand, this is the correct behavior you are seeing.

So `--on-failure` is only for Stack Creation not on update [boto3 docs on create-stack](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation/client/create_stack.html). The docs for `on-failure` both in SAM CLI and CloudFormation state: " Provide an action to determine what will happen when a stack fails to create.". So `on-failure` does nothing on `update-stack`, which is what you are doing. 

In terms of the `DisableRollback`, CloudFormation blocks updates to replacement actions. You can disable this through `--disable-rollback`.

eugene-bright:
The question I have is why `sam --on-failure DELETE` sets `DisableRollback` property? It looks irrelevant and harmful for my workflow.
Thanks for clarification!

eugene-bright:
We don't use `samconfig.toml`.
My previous observations show that  `--on-failure DELETE` sets `"DisableRollback": true`

jfuss:
[...]

PR #4834, opened 2023-03-08
bug: Don't set DisableRollback with OnFailure

jfuss:
#4585 

`--on-failure DELETE` would set `disable-rollback` to True in the `execute-change-set` API call. However, this causes issues when you have replacement updates within your CloudFormation template. Instead, we remove setting `disable-rollback` unless that is explicitly passed in and then allow the deletion of the stack to happen based on the end status of the stack. In talking with with @lucashuy, this seems to align with the original intent.

Does what I said above.

In looking at this in depth, I removed the "UPDATE_FAILED" status to trigger a deletion. This doesn't seem like a valid end state in CloudFormation for a stack but assuming a Stack Ends up in this state, it should be recoverable from. The idea behind this was to not accidentally delete a Stack if we are not in one of the states that force a stack deletion (which the others in the list are, see changes for full list).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/deploy/deploy_context.py
+++ b/samcli/commands/deploy/deploy_context.py
@@ -267,14 +267,8 @@ class DeployContext:
                     if not click.confirm(f"{self.MSG_CONFIRM_CHANGESET}", default=False):
                         return
 
-                # Stop the rollback in the case of DO_NOTHING or if this is a new stack and DELETE is specified
-                # DO_NOTHING behaves the same disable_rollback, they both preserve the current state of the stack
-                do_disable_rollback = (
-                    self.on_failure in [FailureMode.DO_NOTHING, FailureMode.DELETE] or disable_rollback
-                )
-
-                self.deployer.execute_changeset(result["Id"], stack_name, do_disable_rollback)
-                self.deployer.wait_for_execute(stack_name, changeset_type, do_disable_rollback, self.on_failure)
+                self.deployer.execute_changeset(result["Id"], stack_name, disable_rollback)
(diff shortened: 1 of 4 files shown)


samcli/commands/deploy/deploy_context.py
```diff
@@ -267,14 +267,8 @@ def deploy(
...
-                # Stop the rollback in the case of DO_NOTHING or if this is a new stack and DELETE is specified
-                # DO_NOTHING behaves the same disable_rollback, they both preserve the current state of the stack
-                do_disable_rollback = (
-                    self.on_failure in [FailureMode.DO_NOTHING, FailureMode.DELETE] or disable_rollback
-                )
-
-                self.deployer.execute_changeset(result["Id"], stack_name, do_disable_rollback)
-                self.deployer.wait_for_execute(stack_name, changeset_type, do_disable_rollback, self.on_failure)
+                self.deployer.execute_changeset(result["Id"], stack_name, disable_rollback)
+                self.deployer.wait_for_execute(stack_name, changeset_type, disable_rollback, self.on_failure)
```
sriram-mv, line 271: From this my takeaway is we promote explicit behavior overall?

sriram-mv (approved):

mndeveci (approved):

lucashuy (approved):

samcli/commands/deploy/deploy_context.py
lucashuy: The `--disable-rollback` option will now have to be explicitly defined for it to stop rollbacks in a change set like before, since this can affect resource replacements and we don't want to infer that from a different option.

PR #4839, opened 2023-03-08
feat: Auto infra sync after 7 days

hnnasit:

N/A

To minimize the drift in CloudFormation due to only running sync flows to update the code resources. 

The logic will store the last sync time after every infra sync and run an infra sync if it detects an infra sync hasn't been done in a week.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/sync_context.py
```diff
@@ -39,11 +40,29 @@ class ResourceSyncState:
 @dataclass
 class SyncState:
     dependency_layer: bool
+    latest_infra_sync_time: Optional[datetime]
     resource_sync_states: Dict[str, ResourceSyncState]
 
     def update_resource_sync_state(self, resource_id: str, hash_value: str):
```
qingchm, line 46:
```suggestion
    def update_resource_sync_state(self, resource_id: str, hash_value: str) -> None:
```

samcli/commands/sync/sync_context.py
```diff
@@ -39,11 +40,29 @@ class ResourceSyncState:
...
+        Parameters
+        -------
+        resource_id: str
+            The resource identifier of the resource
+        hash_value: str
+            The logical ID identifier of the resource
+        """
         self.resource_sync_states[resource_id] = ResourceSyncState(hash_value, datetime.utcnow())
 
+    def update_infra_sync_time(self):
```
qingchm, line 60:
```suggestion
    def update_infra_sync_time(self) -> None:
```

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -179,8 +196,18 @@ def execute_infra_sync(self, first_sync: bool = False) -> InfraSyncResult:
             #     )
             pass
 
+        # Will be added with the sync infra skip is ready for release
+        # if days_since_last_infra_sync > AUTO_INFRA_SYNC_DAYS:
```
mndeveci, line 200: I wonder if we can simplify this double if blocks over same condition. Not blocking now but something for GA PR.

mndeveci (approved):

samcli/lib/sync/infra_sync_executor.py
hnnasit: Yeah that makes sense. This if loop can be switched to an elif condition with the above if condition.

qingchm (approved):
Overall looks good, made small comments on debug message improvements and dataclass field order which the next commit will address

PR #4840, opened 2023-03-09
feat: Invoke Lambda authorizer

lucashuy:

N/A

This adds the logic required to invoke a Lambda authorizer before the route's Lambda function and passes any context from the authorizer to the route. We also validate the response from the authorizer to make sure that the request is indeed valid and allowed.

Updates our Lambda invocation to be a method so that we can reuse it for both authorizers and routes. Adds some event manipulation to append the required keys to pass along context.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/authorizers/lambda_authorizer.py
+++ b/samcli/local/apigw/authorizers/lambda_authorizer.py
@@ -1,11 +1,11 @@
 """
 Custom Lambda Authorizer class definition
 """
-import json
 import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Tuple
+from json import JSONDecodeError, loads
+from typing import Any, Dict, List, Optional, Tuple, Type
 from urllib.parse import parse_qsl
 
 from samcli.commands.local.lib.validators.identity_source_validator import IdentitySourceValidator
@@ -18,6 +18,7 @@ _RESPONSE_CONTEXT = "context"
 _RESPONSE_POLICY_DOCUMENT = "policyDocument"
 _RESPONSE_IAM_STATEMENT = "Statement"
 _RESPONSE_IAM_EFFECT = "Effect"
+_RESPONSE_IAM_EFFECT_ALLOW = "Allow"
 _RESPONSE_IAM_ACTION = "Action"
 _RESPONSE_IAM_RESOURCE = "Resource"
 _SIMPLE_RESPONSE_IS_AUTH = "isAuthorized"
@@ -305,6 +306,8 @@ class LambdaAuthorizer(Authorizer):
         ----------
         response: str
             JSON string containing the output from a Lambda authorizer
+        method_arn: str
+            The method ARN of the route that invoked the Lambda authorizer
 
         Returns
         -------
@@ -312,8 +315,8 @@ class LambdaAuthorizer(Authorizer):
             True if the request is properly authenticated
         """
         try:
-            json_response = json.loads(response)
-        except ValueError:
+            json_response = loads(response)
+        except (ValueError, JSONDecodeError):
             raise InvalidLambdaAuthorizerResponse(
                 f"Authorizer {self.authorizer_name} return an invalid response payload"
             )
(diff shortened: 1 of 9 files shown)


samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -284,3 +296,156 @@ def _parse_identity_sources(self, identity_sources: List[str]) -> None:
...
+            raise InvalidLambdaAuthorizerResponse(
+                f"Authorizer {self.authorizer_name} return an invalid response payload"
+            )
+
+        if self.payload_version == LambdaAuthorizer.PAYLOAD_V2 and self.use_simple_response:
+            return self._validate_simple_response(json_response)
+
+        return self._validate_iam_response(json_response, method_arn)
+
+    def _validate_iam_response(self, response: dict, method_arn: str) -> bool:
```
lucashuy, line 326: Example responses can be found here as a reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html#http-api-lambda-authorizer.payload-format-response

samcli/local/apigw/local_apigw_service.py
```diff
@@ -564,6 +570,31 @@ def _valid_identity_sources(self, route: Route) -> bool:
 
         return True
 
+    def _invoke_lambda_function(self, lambda_function_name: str, event: dict) -> str:
```
lucashuy, line 573: Moved invocation logic to its own method so that we can reuse it.

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -284,3 +296,166 @@ def _parse_identity_sources(self, identity_sources: List[str]) -> None:
...
+        method_arn: str
+            The method ARN of this request
+
+        Returns
+        -------
+        bool
+            True if the request is authorized
+        """
+        # validate there exists a principalId (unique ID representing the authenticated request)
+        if not response.get(_RESPONSE_PRINCIPAL_ID):
```
sriram-mv, line 343: All the 'ifs' here make it seem that its ripe for a refactor. Thought: how could this be more maintainable?

sriram-mv (changes requested):

samcli/local/apigw/authorizers/lambda_authorizer.py
lucashuy: Yeah, I've thought about that too. We could revisit the same design that was used to validate identity sources (`for validator in list_of_validators: validator.check()`) but we would have the same kind of bloat, however it would be cleaner to read and modify in the future.

lucashuy: This may be out of scope for the purposes of this PR, but we could also look towards writing a JSON schema validator, or using an existing library, to replace any of the JSON schema parsing that happens in our local commands (or anywhere for that matter)

samcli/local/apigw/local_apigw_service.py
```diff
@@ -604,29 +636,42 @@ def _request_handler(self, **kwargs):
...
         except UnicodeDecodeError as error:
             LOG.error("UnicodeDecodeError while processing HTTP request: %s", error)
             return ServiceErrorResponses.lambda_failure_response()
 
-        stdout_stream = io.BytesIO()
-        stdout_stream_writer = StreamWriter(stdout_stream, auto_flush=True)
-
         try:
-            self.lambda_runner.invoke(route.function_name, event, stdout=stdout_stream_writer, stderr=self.stderr)
+            if lambda_authorizer:
```
mildaniel, line 654: Just my opinion, but I think we can move some of this specific logic to a separate function:
```
if lambda_authorizer:
      do_authorizer_logic(context)
... 
```
just so these functions don't keep getting so bulky.

samcli/local/apigw/local_apigw_service.py
```diff
@@ -228,7 +237,7 @@ def _create_method_arn(self, flask_request: Request, event_type: str) -> str:
 
         return (
             f"arn:aws:execute-api:us-east-1:{context.account_id}:"  # type: ignore
-            f"{context.api_id}/{self.api.stage_name}/{method}/{endpoint}"
```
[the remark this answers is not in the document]
lucashuy: Yup, previously I had thought that the endpoint did not contain the leading `/`, however the Flask requests includes it. While working on this task, this was discovered when testing.

samcli/local/apigw/event_constructor.py
```diff
@@ -159,9 +159,9 @@ def construct_v2_event_http(
         stage_variables=stage_variables,
     )
 
-    event_str = json.dumps(event.to_dict())
-    LOG.debug("Constructed String representation of Event Version 2.0 to invoke Lambda. Event: %s", event_str)
-    return event_str
+    event_dict = event.to_dict()
+    LOG.debug("Constructed Event Version 2.0 to invoke Lambda. Event: %s", event_dict)
+    return event_dict  # type: ignore
```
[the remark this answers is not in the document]
lucashuy: mypy seems to think that `event_dict` is of type `Any`, while `event.to_dict()` returns a dictionary. I've updated the return typing on the `.to_dict()` method and it appears to fix this, thanks for bringing this up

samcli/local/apigw/authorizers/lambda_authorizer.py
```diff
@@ -284,3 +297,235 @@ def _parse_identity_sources(self, identity_sources: List[str]) -> None:
...
+        method_arn: str
+            The method ARN of this request
+
+        Returns
+        -------
+        bool
+            True if the request is authorized
+        """
+
+        def is_resource_authorized(response: dict, method_arn: str) -> bool:
```
mildaniel, line 346: Question: why define this as a nested function instead of just another class method?

mildaniel (approved):
Left a couple of non-blocking comments. This looks great, thanks for addressing the comments!

samcli/local/apigw/authorizers/lambda_authorizer.py
lucashuy: Initial choice to leave this as an inner function was that since it was only being used here, moving it to a static class method didn't make that much sense to me. I can move this out though to make the tests easier to read/write.

lucashuy: Oh this class is instantiated, scratch what I said about static class method 😆

sriram-mv (approved):
Thanks for looking into all the comments! ❤️ There is scope for further refactor, but I will not block the PR on that front.

Issue #4824, opened 2023-03-06
Bug: sam cli with Terraform.  sam local invoke MemorySize is always 128MB even though I change it in terraform resource.


dedosmedia:
### Description:
I am following the article: https://aws.amazon.com/es/blogs/compute/better-together-aws-sam-cli-and-hashicorp-terraform/
running any of the samples provided.

### Steps to reproduce:
```bash
git clone https://github.com/aws-samples/aws-sam-terraform-examples
cd aws-sam-terraform-examples/zip_based_lambda_functions/lambda-example
```

Edit main.tf and add a memory_size to the lambda function resource
```terraform
resource "aws_lambda_function" "hello-terraform" {
     ...
    memory_size = 1024
     ...
}
```
```bash
terraform init
terraform apply
sam build --hook-name terraform --beta-features
sam local invoke  --beta-features 
```

### Observed result:
The lambda logs report:
```
END RequestId: 09b77c0c-e5b7-4913-b1d5-5b427c7f1bfd
REPORT RequestId: 09b77c0c-e5b7-4913-b1d5-5b427c7f1bfd  Init Duration: 0.03 ms  Duration: 111.26 ms     Billed Duration: 112 ms
Memory Size: 128 MB     Max Memory Used: 128 MB
```
Notice the reported Memory Size for the function remains as the 128MB default value

### Expected result:

```
END RequestId: 09b77c0c-e5b7-4913-b1d5-5b427c7f1bfd
REPORT RequestId: 09b77c0c-e5b7-4913-b1d5-5b427c7f1bfd  Init Duration: 0.03 ms  Duration: 111.26 ms     Billed Duration: 112 ms
Memory Size: 1024 MB     Max Memory Used: 128 MB
```
It should report the memory_size configured in terraform resource block, 1024MB in this case. It's killing my function because the max memory is more than 128MB even though my desired Memory Size is 1024MB. 

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

**OS:** 
Linux Mint 5.15.0-56-generic #62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**sam --info**
```json
{
  "version": "1.76.0",
  "system": {
    "python": "3.7.10",
    "os": "Linux-5.15.0-56-generic-x86_64-with-debian-bookworm-sid"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.22",
[...]

PR #4846, opened 2023-03-09
bug: adding missing MemorySize field to terraform hook

dedosmedia:

#4824 

When working with Sam CLI and the new Terraform hook, I noticed that when I create a Lambda function and specify the memory_size in the Terraform resource, this value is not passed to the Lambda function when running locally with sam local invoke. As a result, the Lambda function always runs with the default memory size of 128MB. This causes issues when testing functions that require more than 128MB, as they are killed when the memory usage exceeds that value.

To address this issue, I added the missing MemorySize field to the hook property builder. This allows the memory_size value defined in the terraform resource to be passed to the Lambda function, so I can increase the MemorySize as needed.

Not sure

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/property_builder.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/property_builder.py
@@ -204,6 +204,7 @@ AWS_LAMBDA_FUNCTION_PROPERTY_BUILDER_MAPPING: PropertyBuilderMapping = {
     "Runtime": _get_property_extractor("runtime"),
     "Layers": _get_property_extractor("layers"),
     "Timeout": _get_property_extractor("timeout"),
+    "MemorySize": _get_property_extractor("memory_size"),
     "ImageConfig": _build_lambda_function_image_config_property,
 }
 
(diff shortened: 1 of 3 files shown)


samcli/hook_packages/terraform/hooks/prepare/property_builder.py
```diff
@@ -204,6 +204,7 @@ def _check_image_config_value(image_config: Any) -> bool:
     "Runtime": _get_property_extractor("runtime"),
     "Layers": _get_property_extractor("layers"),
     "Timeout": _get_property_extractor("timeout"),
+    "MemorySize": _get_property_extractor("memory_size"),
```
mndeveci, line 207: Will there be `memory_size` property defined all the time, or should we add this conditionally? 
cc: @moelasmar

moelasmar: it is Optional for the customer, but has a default value. The generated plan will always contains a value.

moelasmar (approved):

hnnasit (approved):

mndeveci (approved):

PR #4857, opened 2023-03-13
feat: Add skip-deploy-sync option for sync command

hnnasit:

N/A

This change allows the skip infra sync feature to be used when a new sam sync option is used.

By adding the new option skip-deploy-sync for sam sync.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/command.py
```diff
@@ -496,6 +508,8 @@ def execute_watch(
     skip_infra_syncs: bool
         Boolean flag to determine if only ececute code syncs.
     """
+    # Note: skip_infra_sync is different from skip_infra_syncs, skip_infra_syncs completely skips infra syncs and
```
qingchm, line 511: Let's use another variable name that completely separates the two things

qingchm (dismissed):
Let's either refactor the other variable skip_infra_syncs to use a different name or rename the new flag to have a name that is clearly distinct. Currently this is still confusing even with comments

samcli/commands/sync/command.py
hnnasit: Agreed, will change the variable that only executes code syncs to `disable-infra-syncs` for clarity.

mndeveci (approved):

lucashuy (approved):

samcli/commands/sync/command.py
```diff
@@ -145,6 +145,15 @@
     is_flag=True,
     help="Separate dependencies of individual function into a Lambda layer for improved performance.",
 )
+# Reminder: remove hidden flag once the feature is ready for release.
+@click.option(
+    "--skip-deploy-sync/--no-skip-deploy-sync",
+    default=False,  # Reminder: Change this to True when this feature is released
+    is_flag=True,
+    hidden=True,  # Reminder: Change this to False when this feature is released
+    help="This option will skip the initial infrastructure sync if it is not required"
```
mndeveci, line 154:
```suggestion
    help="This option will skip the initial infrastructure deployment if it is not required"
```

samcli/commands/sync/command.py
```diff
@@ -365,17 +377,17 @@ def do_cli(
...
-                                sync_context,
-                                dependency_layer,
-                                code,
+                                template=template_file,
+                                build_context=build_context,
+                                package_context=package_context,
+                                deploy_context=deploy_context,
+                                sync_context=sync_context,
+                                auto_dependency_layer=dependency_layer,
+                                disable_infra_syncs=code,
```
mndeveci, line 390: Parameter is wrong? Shouldn't this be `skip_deploy_sync`?

mndeveci (changes requested):

samcli/commands/sync/command.py
hnnasit: This is when both code and watch are provided to sync. This will completely disable the infra sync right?

mndeveci: Sorry, my bad! I mismatched with the new option that we are adding since they are very similar

mndeveci (approved):

samcli/commands/sync/command.py
hnnasit: np! Have a note added already to avoid confusion in the future. https://github.com/aws/aws-sam-cli/blob/143e0b654c1b83062f4f318979ab059f81646b87/samcli/commands/sync/command.py#L524-L526

Issue #4858, opened 2023-03-13
Use faster `get_managed_policy_map` when calling SAM transform locally


hoffa:
The SAM transform library now supports faster local transforms by only loading managed policies when not already [bundled within the transform library](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/internal/data/aws_managed_policies.json) (see https://github.com/aws/serverless-application-model/pull/2839), instead of loading them every time. It also minimizes annoying credential expiry issues.

To migrate:

1. Set [`managed_policy_map`](https://github.com/aws/serverless-application-model/blob/e2dd6f63d00b60ccce5427196abab3f7b60ce9d9/samtranslator/translator/translator.py#L48) to `None`.
2. Create a function `get_managed_policy_map` that loads managed policies and caches the result:

    ```python
    @lru_cache(maxsize=None)
    def get_managed_policy_map() -> Dict[str, str]:
        return managed_policy_loader.load()
    ```

4. Pass `get_managed_policy_map` to the new [`get_managed_policy_map`](https://github.com/aws/serverless-application-model/blob/e2dd6f63d00b60ccce5427196abab3f7b60ce9d9/samtranslator/translator/translator.py#L109) argument.

See [`transform.py`](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/translator/transform.py) for an example.

hawflau:
Released in AWS SAM CLI v1.77.0. Closing

PR #4859, opened 2023-03-13
feat: Update translator to use new managed policy map

mildaniel:

#4858 

Makes use of the new SAM-T feature in version 1.61.0 to call Translator with `get_managed_policy_map` parameter and cache policies, speeding up any calls to the transform.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/translate/sam_template_validator.py
+++ b/samcli/lib/translate/sam_template_validator.py
@@ -7,6 +7,7 @@ import logging
 from boto3.session import Session
 from samtranslator.parser import parser
 from samtranslator.public.exceptions import InvalidDocumentException
+from samtranslator.translator.managed_policy_translator import ManagedPolicyLoader
 from samtranslator.translator.translator import Translator
 
 from samcli.commands.validate.lib.exceptions import InvalidSamDocumentException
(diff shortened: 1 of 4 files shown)


hnnasit (approved):

samcli/lib/translate/sam_template_validator.py
```diff
@@ -67,14 +68,29 @@ def get_translated_template_if_valid(self):
...
+        """
+        Helper function for getting managed policies and caching them.
+        Used by the transform for loading policies.
+
+        Returns
+        -------
+        Dict[str, str]
+            Dictionary containing the policy map
+        """
+        return cast(Dict[str, str], self.managed_policy_loader.load())
```
hoffa, line 92: Is the `cast` necessary? `load()` should already be typed.

mildaniel: Even though it's typed, mypy seems to think it returns `Any`.

aahung (commented):
Note that after this change, SAM CLI `sam validate` won't work with previous versions of SAM-T  as `get_managed_policy_map` argument doesn't exist in older version of SAM-T. Hope customers are not using `pip install`.

Issue #4813, opened 2023-03-03
Bug: sam deploy - KeyError


NathanCFoster:
### Description:

When deploying using GitHub actions, a `Error: Status` pops is caught. We are deploying by adding [Python](https://github.com/actions/setup-python) and [SAM](https://github.com/marketplace/actions/setup-aws-sam-cli), then [configuring AWS credentials](https://github.com/marketplace/actions/configure-aws-credentials-action-for-github-actions)

### Steps to reproduce:

Create a basic GitHub actions workflow, with these actions:
```
    - name: Setup python
      uses: actions/setup-python@v4
      with:
        python-version: ${{ inputs.python_v }}
    - name: Setup sam
      uses: aws-actions/setup-sam@v2
      with:
        use-installer: true
    - name: Get Credentials
      uses: aws-actions/configure-aws-credentials@v1
      with:
        aws-region: ${{ inputs.region }}
        role-to-assume: ${{ inputs.role_to_assume }}
    - name: Deploy sam
      shell: bash
      working-directory: ${{ inputs.location }}
      run: |
        if [[ ${{ inputs.sam_config }} == "true" ]]
        then
        sam build && sam deploy
        else
        sam package ${{ inputs.sam_package_config }}
        sam deploy ${{ inputs.sam_config }}
        fi
```
then run the workflow and you should see an error.

### Observed result:

```
Waiting for changeset to be created..


Error: 'Status'
Traceback:
  File "click/core.py", line 1055, in main
  File "click/core.py", line 1657, in invoke
  File "click/core.py", line 1404, in invoke
  File "click/core.py", line 760, in invoke
  File "samcli/lib/cli_validation/image_repository_validation.py", line 92, in wrapped
  File "click/decorators.py", line 84, in new_func
  File "click/core.py", line 760, in invoke
  File "samcli/lib/telemetry/metric.py", line 183, in wrapped
  File "samcli/lib/telemetry/metric.py", line 148, in wrapped
  File "samcli/lib/utils/version_checker.py", line 42, in wrapped
  File "samcli/cli/main.py", line 92, in wrapper
  File "samcli/commands/_utils/cdk_support_decorators.py", line 39, in wrapped
  File "samcli/commands/_utils/command_exception_handler.py", line 93, in wrapper_command_exception_handler
  File "samcli/commands/_utils/command_exception_handler.py", line 73, in wrapper_command_exception_handler
  File "samcli/commands/deploy/command.py", line 223, in cli
[...]

PR #4868, opened 2023-03-15
Handle non changeset error while waiting for chageset

hawflau:

#4813 

Error handling was not catered for non-changeset error

re-raise the waiter exception if it doesn't contain info about the changeset. so the generic error handler will handle it as SDKError

No

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -314,22 +314,22 @@ class Deployer:
             waiter.wait(ChangeSetName=changeset_id, StackName=stack_name, WaiterConfig=waiter_config)
         except botocore.exceptions.WaiterError as ex:
             resp = ex.last_response
-            if "Status" in resp and "StatusReason" in resp:
-                status = resp["Status"]
-                reason = resp["StatusReason"]
+            status = resp.get("Status")
+            reason = resp.get("StatusReason")
 
-                if (
-                    status == "FAILED"
-                    and "The submitted information didn't contain changes." in reason
-                    or "No updates are to be performed" in reason
-                ):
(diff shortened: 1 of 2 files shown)


samcli/lib/deploy/deployer.py
```diff
@@ -314,19 +314,22 @@ def wait_for_changeset(self, changeset_id, stack_name):
...
-                or "No updates are to be performed" in reason
-            ):
-                raise deploy_exceptions.ChangeEmptyError(stack_name=stack_name)
-
-            raise ChangeSetError(
-                stack_name=stack_name, msg="ex: {0} Status: {1}. Reason: {2}".format(ex, status, reason)
-            ) from ex
+            if "Status" in resp and "StatusReason" in resp:
+                status = resp["Status"]
+                reason = resp["StatusReason"]
```
jfuss, line 319: If `Status` is not in the response, what does that indicate? 

I ask because I typically push to avoid checks like this, especially on dicts. It's a better practice to `resp.get("Status", <sane default>)`, which allows for lest nesting and easier to read/maintain code.

jfuss (changes requested):

samcli/lib/deploy/deployer.py
hawflau: If `Status` and `StatusReason` are missing, it means the response is not a valid CloudFormation `DescribeChangeSet` response, and could be something else like Credentials error. 

Some form of branching is still needed to handle the error differently IMO. But let me try to write it in a less nested manner.

mndeveci (approved):

jfuss (approved):

PR #4881, opened 2023-03-16
Add docs sub-commands

mildaniel:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/docs/command.py
```diff
@@ -13,27 +21,114 @@
...
+
+class DocsSubcommand(click.MultiCommand):
+    def __init__(self, command: Optional[List[str]] = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.docs_command = DocsCommand()
+        self.command = command or self.docs_command.sub_commands
+        self.command_string = self.docs_command.sub_command_string
+        self.command_callback = self.docs_command.command_callback
+
+    def get_command(self, ctx: Context, cmd_name: str) -> Command:
```
sriram-mv, line 48: how does --help text look like?

mildaniel: I used the new formatters to format the help text. Looks like this for the base command and all sub-commands.

image

samcli/lib/docs/documentation_links.json
```diff
@@ -0,0 +1,21 @@
+{
+  "config": "https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-config.html",
```
mildaniel, line 2: There are a lot of repetitive URLs here but I left it this way since we have no way of guaranteeing that new URLs will have the same pattern. It also just keeps the configuration easy instead of trying to parse patterns for different types of URLs in our documentation.

samcli/lib/docs/documentation.py
```diff
@@ -23,3 +36,27 @@ def open_docs(self):
...
+        """
+        Opens the configuration file and returns the contents
+
+        Returns
+        -------
+        Dict[str, Any]
+            A dictionary containing commands and their corresponding docs URLs
+        """
+        with open(Path(__file__).parent / DOCS_CONFIG_FILE) as f:
+            return dict(json.load(f))
```
lucashuy, line 62: Thoughts about having the links as constants in code instead of a JSON file to avoid reading a file?

samcli/lib/docs/documentation.py
```diff
@@ -1,17 +1,30 @@
...
+DOCS_CONFIG_FILE = "documentation_links.json"
+LANDING_PAGE = "https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/what-is-sam.html"
+
 
 class Documentation:
-    def __init__(self, browser: BrowserConfiguration, url: str):
+    def __init__(self, browser: BrowserConfiguration, command: str):
         self.browser = browser
-        self.url = url
+        self.command = command
```
lucashuy, line 20: nit: Doc string for the arguments

samcli/lib/docs/documentation.py
mildaniel: We can do either. I had initially played with some different JSON structures to store the data instead of just a simple KV pair which is why I went with JSON, though I don't think reading a file here is overly expensive.

samcli/lib/docs/documentation.py
mildaniel: I don't think we typically add docstrings for constructors but I don't see why not. I'll add it in!

lucashuy: Yeah I see that in some places we do, but then in other places there isn't a docstring but it is nice to have one

samcli/commands/docs/command_context.py
```diff
@@ -0,0 +1,122 @@
...
+            raise InvalidDocsCommandException(
+                f"Command not found. Try using one of the following available commands:{os.linesep}"
+                f"{os.linesep.join([command for command in self.docs_command.get_complete_command_paths()])}"
+            )
+        browser = BrowserConfiguration()
+        documentation = Documentation(browser=browser, command=self.command)
+        try:
+            documentation.open_docs()
+        except BrowserConfigurationError:
+            echo(ERROR_MESSAGE.format(URL=documentation.url))
```
moelasmar, line 120: do we need to consider that the `sam docs` command failed in this case?

samcli/commands/docs/core/command.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+from typing import List, Optional
+
+from click import Command, Context, MultiCommand, style
+
+from samcli.cli.row_modifiers import RowDefinition
+from samcli.commands.docs.command_context import COMMAND_NAME, DocsCommandContext
+from samcli.commands.docs.core.formatter import DocsCommandHelpTextFormatter
+
+HELP_TEXT = "NEW! Open the documentation in a browser."
```
moelasmar, line 12: shall we remove `NEW!` word after some time :)

samcli/commands/docs/core/command.py
```diff
@@ -0,0 +1,151 @@
...
+        formatter: DocsCommandHelpTextFormatter
+            A formatter instance to use for formatting the help text
+        """
+        with formatter.indented_section(name="Description", extra_indents=1):
+            formatter.write_rd(
+                [
+                    RowDefinition(
+                        text="",
+                        name=DESCRIPTION
+                        + style("\n  This command does not require access to AWS credentials.", bold=True),
```
moelasmar, line 52: is it better to use `os.linesep` instead of `\n` ?

samcli/lib/docs/documentation_links.json
```diff
@@ -0,0 +1,21 @@
+{
```
moelasmar, line 1: do we need to keep the keys in this doc in any specific order, or it does not matter ?

moelasmar (approved):
It looks good to me. I just left some minor comments and questions, but I do not think it should block the PR

samcli/commands/docs/core/command.py
mildaniel: Yup, we'll remove it in a few months.

samcli/lib/docs/documentation_links.json
mildaniel: Shouldn't matter

lucashuy (approved):

PR #4882, opened 2023-03-17
fix: move mount with write into try/except block and don't prompt for Image type lambda functions

mndeveci:
N/A

- `prompt_user_to_enable_mount_with_write_if_needed` errors are not handled properly
- `prompt_user_to_enable_mount_with_write_if_needed` runs for `Image` type functions which is not needed
- `TestBuildCommand_PythonFunctions_CDK_0_cdk_v1_synthesized_template_zip_image_functions_json::test_cdk_app_with_default_requirements` is failing in Windows due to timeout.

- Move `prompt_user_to_enable_mount_with_write_if_needed` into try/except block so that errors will be handled for it
- Don't prompt for the `Image` type functions, since this is only needed for `Zip` ones
- Increase `Timeout` for the function in CDK tests so that timeout will be fixed for Windows tests

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -302,8 +302,11 @@ class BuildContext:
                 )
 
                 click.secho(msg, fg="yellow")
+        except FunctionNotFound as function_not_found_ex:
+            raise UserException(
+                str(function_not_found_ex), wrapped_from=function_not_found_ex.__class__.__name__
+            ) from function_not_found_ex
         except (
-            FunctionNotFound,
             UnsupportedRuntimeException,
             BuildError,
             BuildInsideContainerError,
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -312,6 +312,7 @@ class BuildContext:
             BuildInsideContainerError,
             UnsupportedBuilderLibraryVersionError,
             InvalidBuildGraphException,
+            ResourceNotFound,
         ) as ex:
             click.secho("\nBuild Failed", fg="red")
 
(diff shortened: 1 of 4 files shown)


samcli/commands/build/build_context.py
```diff
@@ -305,8 +302,8 @@ def run(self):
                 )
 
                 click.secho(msg, fg="yellow")
-
         except (
+            FunctionNotFound,
```
hawflau, line 306: Should `FunctionNotFound` be in its own `except`? 
because previously it didn't include `click.secho("\nBuild Failed", fg="red")`

mndeveci: 🤔 makes sense. I was thinking to merge everything, but `FunctionNotFound` is not build failure, since we can't run any build at all. Will update it

samcli/commands/build/build_context.py
```diff
@@ -236,20 +236,20 @@ def run(self):
...
+            # boolean value indicates if mount with write or not, defaults to READ ONLY
+            mount_with_write = False
+            if self._use_container:
+                if self._mount_with == MountMode.WRITE:
+                    mount_with_write = True
+                else:
+                    # if self._mount_with is NOT WRITE
+                    # check the need of mounting with write permissions and prompt user to enable it if needed
+                    mount_with_write = prompt_user_to_enable_mount_with_write_if_needed(
+                        self.get_resources_to_build(),
```
lucashuy, line 249: Question, the call to `self.get_resources_to_build()` makes another call that looks like it could raise a `ResourceNotFound` exception, should this also be caught?

mndeveci: Good catch. It was already called in `ApplicationBuilder`'s constructor and we were printing the full stack trace. Now we are gracefully ending the execution there.

hawflau (approved):

lucashuy (approved):

PR #4910, opened 2023-03-22
feat: Validate headers against validation expression property

lucashuy:
#### Which issue(s) does this change fix?

N/A

#### Why is this change necessary?
Adds a missing check to validate incoming headers against the validation expression property defined in the template file.

#### How does it address the issue?
Modifies the existing header validator by overriding the parent validation method to also check the expression.

#### What side effects does this change have?
None

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [ ] Write/update integration tests
- [ ] Write/update functional tests if needed
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #4916, opened 2023-03-23
fix: Fix APIGW V2 context passing

lucashuy:

N/A

APIGW payload format version `2.0` passes the context under `{authorizer: {lambda: ...context...}}` instead of `{authorizer: ...context...}`

Adds a check for the `2.0` payload format version.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/local/apigw/test_local_apigw_service.py
+++ b/tests/unit/local/apigw/test_local_apigw_service.py
@@ -872,6 +872,20 @@ class TestApiGatewayService(TestCase):
         self.http_service._invoke_parse_lambda_authorizer(auth, {}, route_event, self.http_v2_payload_route)
         self.assertEqual(route_event, {"requestContext": {"authorizer": {"lambda": mock_get_context}}})
 
+    @patch.object(LocalApigwService, "_invoke_lambda_function")
+    @patch.object(LocalApigwService, "_create_method_arn")
+    def test_lambda_authorizer_pass_context_api(self, method_arn_mock, mock_invoke):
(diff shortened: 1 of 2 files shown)


tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -858,6 +858,20 @@ def test_lambda_auth_unauthorized_response(
...
+    def test_lambda_authorizer_pass_context_http(self, method_arn_mock, mock_invoke):
+        mock_get_context = Mock()
+        route_event = {}
+
+        auth = LambdaAuthorizer(Mock(), Mock(), "auth_lambda", [], Mock(), Mock(), Mock())
+        auth.is_valid_response = Mock(return_value=True)
+        auth.get_context = Mock(return_value=mock_get_context)
+        self.http_v2_payload_route.authorizer_object = auth
+
+        self.http_service._invoke_parse_lambda_authorizer(auth, {}, route_event, self.http_v2_payload_route)
```
moelasmar, line 872: I am not able to find a test case that cover the V1 route. Is there a test case that cover this case?

lucashuy: There doesn't appear to be a case for the V1 route, thanks for catching that

mildaniel (commented):
Same as #4910, I'll defer to the testing comments.

mildaniel (approved):

PR #4924, opened 2023-03-25
unpin regex dependency

gruebel:
the `regex` package was pinned to version `2021.9.30` for over a year, even the issue was already fixed and the package `dateparser`, which is actually using it, is not pinning it.

I'm not judging, but I highly recommend to rethink your practice of pinning all these packages, which makes it almost impossible to install anything else in the same environment. You should rather start to embrace wildcard dependencies or at least upper/lower bounds for all of them. If you find an incompatible one, then you can easily exclude it, like I did in my PR.
The SAM CLI is pre-installed on CodeBuild via `pip` so your strict dependencies block any other packages version dependencies or dictate typically very old versions.
ex.
```
docker~=4.2.0 (from 06/2020, current latest 6.0.1)
jmespath~=0.10.0 (from 05/2020, current latest 1.0.1)
PyYAML>=5.4.1,==5.* (from 01/2021, current latest 6.0)
tzlocal==3.0 from 08/2021, current latest 4.3) <- dateparser doesn't pin this dependency, so probably worth to test an upgrade again
watchdog==2.1.2 (from 05/2021, current latest 3.0.0)
```

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


requirements/base.txt
```diff
@@ -21,8 +21,8 @@ pyopenssl==23.0.0
 # Needed for supporting Protocol in Python 3.7, Protocol class became public with python3.8
 typing_extensions~=4.4.0
 
-# NOTE: regex is not a direct dependency of SAM CLI, but pin to 2021.9.30 due to 2021.10.8 not working on M1 Mac - https://bitbucket.org/mrabarnett/mrab-regex/issues/399/missing-wheel-for-macosx-and-the-new-m1
-regex==2021.9.30
+# NOTE: regex is not a direct dependency of SAM CLI, exclude version 2021.10.8 due to not working on M1 Mac - https://github.com/mrabarnett/mrab-regex/issues/399
+regex!=2021.10.8
```
moelasmar, line 25: in my opinion it is better to pin to a specific version, so the SAM CLI behaviour will be the same for everyone. Can you change this to pin to the latest stable version.

moelasmar: good point

mildaniel (approved):

Issue #4860, opened 2023-02-28
Bug: Duplicate `Globals` sections in `hello-world-powertools-dotnet` 


ljacobsson:
### Description:
If I select 'y' on both xray and applications insights prompts, then the tool generates two `Globals` sections:

```
$ sam init --app-template hello-world-powertools-dotnet --name sam-app --package-type Zip --runtime dotnet6

Based on your selections, the only Template available is Hello World Example.
We will proceed to selecting the Template as Hello World Example.

Based on your selections, the only dependency manager available is cli-package.
We will proceed copying the template using cli-package.

Would you like to enable X-Ray tracing on the function(s) in your application?  [y/N]: y
X-Ray will incur an additional cost. View https://aws.amazon.com/xray/pricing/ for more details

Would you like to enable monitoring using CloudWatch Application Insights?
For more info, please view https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-application-insights.html [y/N]: y
AppInsights monitoring may incur additional cost. View https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/appinsights-what-is.html#appinsights-pricing for more details

Cloning from https://github.com/aws/aws-sam-cli-app-templates (process may take a moment)

    -----------------------
    Generating application:
    -----------------------
    Name: sam-app
    Runtime: dotnet6
    Architectures: x86_64
    Dependency Manager: cli-package
    Application Template: hello-world-powertools-dotnet
    Output Directory: .
    Configuration file: sam-app/samconfig.toml
    
    Next steps can be found in the README file at sam-app/README.md
        

Commands you can use next
=========================
[*] Create pipeline: cd sam-app && sam pipeline init --bootstrap
[*] Validate SAM template: cd sam-app && sam validate
[*] Test Function in the Cloud: cd sam-app && sam sync --stack-name {stack-name} --watch


Error: while constructing a mapping
  in "./sam-app/template.yaml", line 1, column 1
found duplicate key "Globals" with value "ordereddict([('Function', ordereddict([('Tracing', 'Active')])), ('Api', ordereddict([('TracingEnabled', True)]))])" (original value: "ordereddict([('Function', ordereddict([('Timeout', 10), ('MemorySize', 256), ('Runtime', 'dotnet6')])), ('Api', ordereddict([('TracingEnabled', True)]))])")
  in "./sam-app/template.yaml", line 16, column 1

To suppress this check see:
[...]

PR #4940, opened 2023-03-28
fix: Allow better comment handling in XrayTracingTemplateModifier

jfuss:
#4860 

Fixes a bug in which the XrayTracingTemplateModifier was duplicating Globals.

Reworks the XrayTracingTemplateModifier to be closer to the ApplicationInsights in using ruamel to parse and keep comments.

None. But I did remove unnecessary code and tests that we no longer needed.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
```diff
@@ -2,100 +2,77 @@
...
 
 from samcli.lib.init.template_modifiers.cli_template_modifier import TemplateModifier
 
 LOG = logging.getLogger(__name__)
 
 
 class XRayTracingTemplateModifier(TemplateModifier):
+    import ruamel.yaml
+    from ruamel.yaml import YAML
+    from ruamel.yaml.comments import CommentedMap
```
mndeveci, line 15: - I like that you put imports under class so that they won't be loaded once the module is imported. But this module has only this class, which means this will be imported whenever the class is used. Should we move the imports to the top into others?
- Line 13 imports entire `ruamel.yaml` but then next lines imports individual elements from `ruamel.yaml` again. Shouldn't we use one way or the other (either import everything or import only used stuff one by one)?

tests/unit/lib/init/test_cli_template_modifier.py
```diff
@@ -62,77 +55,80 @@ def test_must_update_template_fields(self, get_template_patch):
...
-            "    api_field: field_value\n",
-            "\n",
-            "Resources:\n",
-            "  HelloWorldFunction:\n",
-            "    Type: AWS::Serverless::Function\n",
-            "    Properties:\n",
-            "      CodeUri: hello_world/\n",
-            "      Handler: app.lambda_handler\n",
-        ]
+        get_template_patch.return_value = YAML().load(
```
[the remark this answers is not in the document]
jfuss: My understanding is the default is `safe` in ruamel: https://stackoverflow.com/questions/71298807/is-it-safe-to-use-the-default-load-in-ruamel-yaml

samcli/lib/init/template_modifiers/cli_template_modifier.py
```diff
@@ -26,85 +25,16 @@ def modify_template(self):
         CFN yaml
         """
         self._update_template_fields()
-        self._write(self.template)
+        template_to_write = self.template
         if not self._sanity_check():
-            self._write(self.copy_of_original_template)
+            template_to_write = self.copy_of_original_template
+
+        self._write(template_to_write)
```
jfuss, line 32: Changed this so we only every write once.

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
```diff
@@ -2,100 +2,78 @@
...
 
     def _update_template_fields(self):
         """
         Add new field to SAM template
         """
-        global_section_position = self._section_position(self.GLOBALS)
+        if self.template.get(self.GLOBALS):
+            template_globals = self.template.get(self.GLOBALS)
+
+            function_gloabls = template_globals.get(self.FUNCTION, {})
```
hnnasit, line 54: nit: typo in variable name

hnnasit (approved):
Thanks Jacob, left a small comment. Overall LGTM.

hawflau (approved):

mndeveci (approved):

samcli/lib/init/template_modifiers/xray_tracing_template_modifier.py
jfuss: Thanks. Updated in the latest revision

PR #4942, opened 2023-03-28
feat: Add metrics for using the Lambda authorizer feature

lucashuy:

N/A

This adds a new event to our event tracker so that we can upload Lambda authorizer invocation events to our telemetry.

Adds a new enum under `UsedFeature` and makes a call to the event tracking method to add that to the list of events to send once the command finishes execution.

This also adds a new property to the event metric: `exception_name`.
The exception name property will let us pass in the name of an exception that could have happened when using the feature.

This PR also updates the existing Lambda authorizer invocation tests to mock out the telemetry call and adds another one to specifically test if the event tracker was called with the correct parameters.

Lastly, this PR also now property catches the invalid response exception from the lambda authorizer.

Event related metrics will now have an `exception_type` property going forward.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel (dismissed):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -701,6 +702,8 @@ def _invoke_parse_lambda_authorizer(
         route: Route
             The route that is being called
         """
+        EventTracker.track_event(EventName.USED_FEATURE.value, UsedFeature.CUSTOM_LAMBDA_AUTHORIZERS.value)
```
moelasmar, line 705: why we do not track if the lambda authorizer works successfully or not ?

lucashuy: I decided to track all uses of authorizers, whether or not they were successful or not since we can filter the results out. If we typically track after success, then I can change the call to be after, no issues with either approach

moelasmar (approved):

lucashuy:
Holding off on merging for now, pending further discussion on tracking this event.

samcli/commands/local/lib/swagger/parser.py
```diff
@@ -136,15 +136,17 @@ def get_authorizers(self, event_type: str = Route.API) -> Dict[str, Authorizer]:
                     auth_name,
                 )
 
-            enable_simple_response = authorizer_object.get(SwaggerParser._AUTHORIZER_SIMPLE_RESPONSES, False)
+            enable_simple_response = authorizer_object.get(SwaggerParser._AUTHORIZER_SIMPLE_RESPONSES)
             if event_type != Route.HTTP or payload_version != LambdaAuthorizer.PAYLOAD_V2 and enable_simple_response:
-                enable_simple_response = False
+                # only display message if simple response is defined
```
mildaniel, line 141: Is this change for something unrelated to metrics?

samcli/lib/telemetry/event.py
```diff
@@ -108,6 +111,7 @@ def to_json(self):
             "event_value": self.event_value,
             "thread_id": self.thread_id,
             "time_stamp": self.time_stamp,
+            "exception_name": self.exception_name,
```
mildaniel, line 114: I assume this is the structure that gets published to Athena, right? Is it safe to publish `None` for the exception type and how does it appear in the database when we need to query it?

samcli/lib/telemetry/event.py
```diff
@@ -195,10 +202,25 @@ def clear_trackers():
...
         send_thread = threading.Thread(target=EventTracker._send_events_in_thread)
         send_thread.start()
         return send_thread
 
+    @staticmethod
+    def _set_session_id() -> None:
+        """
+        Get the session ID from click and save it locally.
+        """
+        if not EventTracker._session_id:
```
mildaniel, line 216: I don't see why there would be, but wanted to confirm that there won't be any collisions or race-conditions between retrieving and setting the `session_id` from click since it was moved from a previously locked state out.

mildaniel (commented):
Thanks Lucas. Looks good overall and I like the change. Left a few minor questions.

samcli/commands/local/lib/swagger/parser.py
lucashuy: This appears to be old code from a completely different PR, will remove.

samcli/lib/telemetry/event.py
lucashuy: This is a good catch. I can move the function call back to the locked state so that we don't accidentally undo anything that was caught before.

samcli/lib/telemetry/event.py
lucashuy: Yup you're correct that this is what gets published to the database. Publishing `None` is safe, this gets interpreted as `null` in the database (`[{...., exception_name=null, ....}]`), so a query would consist of a null check.

samcli/lib/telemetry/event.py
```diff
@@ -134,7 +138,9 @@ class EventTracker:
     MAX_EVENTS: int = 50  # Maximum number of events to store before sending
 
     @staticmethod
-    def track_event(event_name: str, event_value: str):
+    def track_event(
+        event_name: str, event_value: str, session_id: Optional[str] = None, exception_name: Optional[str] = None
```
moelasmar, line 142: why do we need to send the `session_id` is there any case that this parameter will be different than the `session_id` retrieved from the command context ?

samcli/local/apigw/local_apigw_service.py
```diff
@@ -92,6 +94,19 @@ def __init__(
         self._dict_of_routes: Dict[str, Route] = {}
         self.stderr = stderr
 
+        self._click_session_id = None
```
moelasmar, line 97: I think this session_id should be similar to the default session_id we already use in the Event Tracker. can we remove this part here, and just depend on the default session_id used in the event tracker

moelasmar (commented):
Can you create some task to update the other features that use the event tracking to include the new exception_name idea?

samcli/local/apigw/local_apigw_service.py
lucashuy: Addressed in above comment thread ^^

moelasmar (approved):

mildaniel (approved):

Issue #4937, opened 2023-03-28
Bug: sam deploy - AttributeError


mikegrima:
### Description:
Just upgraded to the SAM CLI version 1.78.0, and am now seeing this error on deployment:

```
% sam deploy --config-env TEST      

Error: 'str' object has no attribute 'value'
Traceback:
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/samcli/lib/cli_validation/image_repository_validation.py", line 92, in wrapped
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.78.0/libexec/lib/python3.11/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4945, opened 2023-03-29
fix: parsing capabilities attribute as array or string value in config toml file

moelasmar:
[](https://github.com/aws/aws-sam-cli/issues/4937)

The tomlkit version upgrade causes regression for customers who uses toml config file that contains the capabilities attribute defined as array value.

revert the tomlkit version upgrade [pr](https://github.com/aws/aws-sam-cli/pull/4757)

#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


requirements/base.txt
```diff
@@ -14,7 +14,7 @@ dateparser~=1.1
 requests==2.28.2
 serverlessrepo==0.1.10
 aws_lambda_builders==1.28.0
-tomlkit==0.11.6
+tomlkit==0.7.2
```
hawflau, line 17: why downgrading it?

moelasmar: I just reverted the upgrade PR https://github.com/aws/aws-sam-cli/pull/4757

mildaniel (approved):
It makes sense to me to revert back with additional tests that cover this case, then try to update the version after with safeguards in place. Thanks Mohamed.

requirements/base.txt
moelasmar: I changed the PR to upgrade tomlkit to `0.11.7`

tests/integration/deploy/test_deploy_command.py
```diff
@@ -1579,3 +1579,39 @@ def test_deploy_with_language_extensions(self):
         )
         deploy_process_execute = self.run_command(deploy_command_list)
         self.assertEqual(deploy_process_execute.process.returncode, 0)
+
+    @parameterized.expand([("aws-serverless-function.yaml", "samconfig-capabilities-list.toml")])
+    def test_deploy_with_valid_config_capabilities_list(self, template_file, config_file):
```
hawflau, line 1584: nit and non-blocking: I wonder if it makes more sense to add functional tests rather than integration tests. It seems to me that what we actually want to test is that the list in the toml config file is parsed correctly.

PR #4968, opened 2023-04-04
feat: Add usage disclaimer when starting API with authorizers

lucashuy:

N/A

Adds a disclaimer message to the output of `local start-api` if there are any authorizers that are being used in the template.

Screen Shot 2023-04-11 at 11 49 51 AM

This change also fixes some unnecessary log output if there are no authorizers defined in the template by checking to see if a route had an authorizer set in the first place.

Adds a log statement at the end of the API collection phase and checks to the warning log for undefined authorizers.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/api_collector.py
+++ b/samcli/lib/providers/api_collector.py
@@ -4,10 +4,12 @@ routes in a standardized format
 """
 
 import logging
+import os
 from collections import defaultdict
 from typing import Dict, Iterator, List, Optional, Set, Tuple, Union
 
 from samcli.lib.providers.provider import Api, Cors
+from samcli.lib.utils.colors import Colored
 from samcli.local.apigw.authorizers.authorizer import Authorizer
 from samcli.local.apigw.route import Route
 
@@ -102,7 +104,10 @@ class ApiCollector:
                         route.authorizer_name,
                         route.path,
                     )
-                else:
+
+                    continue
+
+                if not authorizer_object and authorizer_name_lookup:
                     route.authorizer_name = None
 
                     LOG.info(
@@ -184,6 +189,14 @@ class ApiCollector:
         api.stage_name = self.stage_name
         api.stage_variables = self.stage_variables
         api.cors = self.cors
+
+        for authorizers in self._authorizers_per_resources.values():
+            if len(authorizers):
+                message = f"{os.linesep}<sample warning message goes here>{os.linesep}"
+                LOG.warning(Colored().yellow(message))
+
+                break
+
         return api
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


samcli/lib/providers/api_collector.py
```diff
@@ -184,6 +189,16 @@ def get_api(self) -> Api:
         api.stage_name = self.stage_name
         api.stage_variables = self.stage_variables
         api.cors = self.cors
+
+        for authorizers in self._authorizers_per_resources.values():
+            if len(authorizers):
+                message = f"""{os.linesep}AWS SAM CLI cannot guarantee one-to-one behaviour of Lambda authorizers
```
sriram-mv, line 195: Rewording a bit:
```
AWS SAM CLI cannot guarantee 100% Fidelity between authorizers locally vs authorizers deployed on AWS. Any application critical behavior should be validated before deploying to production.
```

sriram-mv (changes requested):

samcli/lib/providers/api_collector.py
mildaniel: Should we include a snippet about testing in the cloud with `sam sync` as an option for "True" validation? Thinking it could be a way to bring some attention to what `sam sync` can do.

lucashuy: Thats a good call out, we can amend the prompt to remove personal pronouns if the Daniel's suggestion sounds good to everyone else.

samcli/lib/providers/api_collector.py
```diff
@@ -184,6 +189,18 @@ def get_api(self) -> Api:
         api.stage_name = self.stage_name
         api.stage_variables = self.stage_variables
         api.cors = self.cors
+
+        for authorizers in self._authorizers_per_resources.values():
+            if len(authorizers):
+                message = f"""{os.linesep}AWS SAM CLI does not guarantee 100% fidelity between authorizers locally 
+vs authorizers deployed on AWS. Any application critical behavior should
```
sriram-mv, line 196: I know I made the suggestion, but on re-reading. we could replace "vs" with "and".

samcli/lib/providers/api_collector.py
```diff
@@ -184,6 +189,18 @@ def get_api(self) -> Api:
...
+be validated thoroughly before deploying to production.
+
+Testing application behaviour against authorizers deployed to AWS can be done using the sam sync command.{os.linesep}"""
```
sriram-mv, line 199: deployed "on" AWS?

Issue #4161, opened 2022-08-28
Bug: CORS Support in start-api


sravimohan:
## "sam local start-api" does not return cors headers in the lambda responses.

This issue was marked as fixed and closed in [https://github.com/aws/aws-sam-cli/issues/323](https://github.com/aws/aws-sam-cli/issues/323).  but seems to have regressed and is no longer working any more.

The options request correctly returns the CORS headers.  But the lambda response does not include "access-control-allow-origin" header.

The only workaround is to also manually include "access-control-allow-origin" header in the lambda responses.  

Pl. note its works fine when deployed without any workaround.

```
$ sam --version
SAM CLI, version 1.55.0
```

minimal repo for reproducing issue
[https://github.com/sravimohan/bug-aws-sam-cli-4161](https://github.com/sravimohan/bug-aws-sam-cli-4161)

template.yml
```
AWSTemplateFormatVersion: "2010-09-09"
Transform: AWS::Serverless-2016-10-31

Globals:
  Function:
    Timeout: 3
    Runtime: nodejs16.x
    Architectures:
      - x86_64

  HttpApi:
    CorsConfiguration:
      AllowHeaders:
        - Content-Type
        - Authorization
        - Access-Control-Allow-Origin
      AllowMethods:
        - HEAD
        - OPTIONS
        - GET
        - POST
      AllowOrigins:
        - http://localhost:3010

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello-world/
      Handler: app.lambdaHandler
      Events:
        HelloWorld:
          Type: HttpApi
          Properties:
            Path: /hello
            Method: GET
```

function
```
let response;

exports.lambdaHandler = async (event, context) => {
    try {
        response = {
            'statusCode': 200,
            // work around for cors issue
            // headers: {
            //     "Access-Control-Allow-Origin": "http://localhost:3010",
            // },
            'body': JSON.stringify({
                message: 'hello world',
            })
        }
    } catch (err) {
        console.log(err);
        return err;
    }

    return response
};
```

jfuss:
@sravimohan According to API Gateway docs, the Lambda is responsible for adding the `access-control-allow-origin` headers.

See https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html
[...]

PR #4991, opened 2023-04-10
Bug: CORS Support in start-api

sleepwithcoffee:

#4161

To address a bug raised in ticket of local API gateway that does not properly return CORS headers. Currently, a workaround of using external plugin is required.

Inject CORS response headers when returning to client. There are generated based on request Origin header.

I would expect none since this is a bug fix and an improvement. With this, user would not need to use external plugin to enable CORS for the local stack.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation (as below)

Current behavior:
* Request header "Access-Control-Allow-Origin" is passed on as response header "Access-Control-Allow-Origin" without processing
New behavior:
* Get request header "Access-Control-Allow-Origin" and use it (either single domain or multiple domains or *) to match with Origin request header
  * If a match was found, only return that single matched domain as response "Access-Control-Allow-Origin" 
  * If no match was found, reject the request by not returning according response header "Access-Control-Allow-Origin" 
* If there is no "Origin" header, straightout reject the request

Some consideration:
* I didn't put the new logic under Cors class since this logic is supposed to be contextual of incoming requests
* I named the new function _response_cors_headers since it is a private function, made it static for unit testing
* The code prefers being restrictive over permissive due to the nature of this handling


samcli/local/apigw/local_apigw_service.py
```diff
@@ -413,6 +414,31 @@ def _get_current_route(self, flask_request):
 
         return route
 
+    @staticmethod
+    def _response_cors_headers(flask_request, cors_headers):
+        if "Access-Control-Allow-Origin" not in cors_headers:
+            return cors_headers
+
+        cors_origins = cors_headers["Access-Control-Allow-Origin"]
+        # unset this header due to restrictive manner
```
jfuss, line 423: Why unset this? It looks like it should be removed in the case that `response_allowed_origin` is not set. If so, I would move this to be closer to line 437. I think it just make it a little easier to read since the mutation of the `cors_headers` is all in one spot.

jfuss (changes requested):
Is it possible to have integration tests that could help us validate the behavior?

samcli/local/apigw/local_apigw_service.py
sleepwithcoffee: Hi @jfuss , thank you for taking your time to review this

we unset and set again if a match is found (restrictive manner)

if multiple domains are allowed, we only send back 1 allowed domain in our response header
if all domains are allowed, we also only send back  1 allowed domain in our response header

we do not return this header (implying a deny) if no matches is found

samcli/local/apigw/local_apigw_service.py
```diff
@@ -303,6 +303,7 @@ def _request_handler(self, **kwargs):
 
         route = self._get_current_route(request)
         cors_headers = Cors.cors_to_headers(self.api.cors)
+        cors_headers = self._response_cors_headers(request, cors_headers)
```
lucashuy, line 306: Thanks for creating this PR! Did we need to restrict this behaviour to only run for HTTP APIs? From the docs and the linked issue, this seems to be something that the Lambda function deals with on for REST, but we would need to manage for HTTP.

sleepwithcoffee: hi @lucashuy 

During the implementation, I referred to [MDN web docs](https://developer.mozilla.org/en-US/docs/web/http/headers/access-control-allow-origin) for reference:

image

I also referred to the implementation of [middy cors middleware](https://github.com/middyjs/middy/blob/main/packages/http-cors/index.js)

below piece of code will basically parse and return a single origin if matched:
image

it is explained if looking deeper into the `getOrigin()` function:

image

Hope these help explain the reasoning behind my code.

Thanks,

samcli/local/apigw/local_apigw_service.py
sleepwithcoffee: hi @jfuss please also refer to my below comment:
https://github.com/aws/aws-sam-cli/pull/4991#discussion_r1174612400

sleepwithcoffee:
merged and resolved conflicts

hawflau (approved):
LGTM!

moelasmar:
@sleepwithcoffee .. We had to revert this PR as it cause some regression, and some of the integration test cases are failing. Please check the revert PR https://github.com/aws/aws-sam-cli/pull/5200 to get more details.

PR #4992, opened 2023-04-10
fix: run managed stack tests sequentially & suppress rustup output

mndeveci:
- Tests in `test_managed_stack_deploy.py` clashing each other when we run them in parallel. Switched to use `--dist=loadgroup` and marked the tests with `@pytest.mark.xdist_group(name="managed_stack")` so that each test will run in the same thread sequentially.
- Updated same tests to only run in `python3.8` instead of `python3.7`
- Suppressed `rustup` output to fix issue which causes workers to stall.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -99,7 +99,7 @@ install:
   - sh: "terraform -version"
 
   # install Rust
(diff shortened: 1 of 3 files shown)


sriram-mv (approved):
I learned something new today!

hawflau (approved):

PR #4993, opened 2023-04-10
sam init command help

sriram-mv:
N/A

* Clean up the help text for sam init, especially its the first command we ask to run after looking at the `sam --help` text.
* Split up examples section to showcase what each section is about, we should do this for other commands as well.

init

N/A

* Ran the unit and integration tests for the init suite.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/command.py
+++ b/samcli/commands/init/command.py
@@ -183,7 +183,7 @@ def non_interactive_validation(func):
     type=click.Choice(SUPPORTED_DEP_MANAGERS),
     default=None,
     help="Dependency manager for Lambda runtime."
-    + click.style(f"\n\nDependency managers: {' '.join(SUPPORTED_DEP_MANAGERS)}", bold=True),
+    + click.style(f"\n\nDependency managers: {', '.join(SUPPORTED_DEP_MANAGERS)}", bold=True),
     required=False,
     cls=ClickMutex,
     replace_help_option="--dependency-manager DEPENDENCY_MANAGER",
--- a/samcli/commands/init/command.py
+++ b/samcli/commands/init/command.py
@@ -10,6 +10,7 @@ import click
 from samcli.cli.cli_config_file import TomlProvider, configuration_option
 from samcli.cli.main import common_options, pass_context, print_cmdline_args
 from samcli.commands._utils.click_mutex import ClickMutex
+from samcli.commands.init.core.command import InitCommand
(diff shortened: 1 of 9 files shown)


samcli/commands/init/command.py
```diff
@@ -238,12 +218,12 @@ def wrapped(*args, **kwargs):
...
     "--tracing/--no-tracing",
     default=None,
-    help="Enable AWS X-Ray tracing for your lambda functions",
+    help="Enable AWS X-Ray tracing for application.",
 )
 @click.option(
     "--application-insights/--no-application-insights",
     default=None,
-    help="Enable CloudWatch Application Insights monitoring for your application",
+    help="Enable CloudWatch Application Insights monitoring for application.",
```
mndeveci, line 226: nit: I remember seeing another PR which removed the trailing dots in a help text, we might need to align which approach we are taking

mndeveci (approved):

samcli/commands/init/command.py
sriram-mv: Oh interesting, do we remember the rationale for that? Agree on adopting one vs other in totality for consistency.

mndeveci: I think that is more of a personal preference, not sure if CLI recommendations favor one or the other. I am OK with either but just wanted to make sure we will have a persistent help text through the CLI. :)

samcli/commands/init/core/command.py
```diff
@@ -0,0 +1,116 @@
...
+            with formatter.indented_section(name=option_heading, extra_indents=1):
+                formatter.write_rd(options.get("extras", [RowDefinition()]))
+                formatter.write_rd(
+                    [RowDefinition(name="", text="\n")]
+                    + [
+                        opt
+                        for options in zip(opts, [RowDefinition(name="", text="\n")] * (len(opts)))
+                        for opt in options
+                    ],
+                )
```
mildaniel, line 116: Is there a reason for not using `CoreCommand._format_options()` here?
https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/core/command.py#L38

mildaniel (commented):
Thanks Sriram, I really like the sectioned examples. Left one small questions.

mildaniel (approved):

PR #4994, opened 2023-04-10
fix: Only print console message if enableSimpleResponses is defined

lucashuy:

N/A

Previously, the "simple response only available on HTTP APIs" message was being logged to the terminal even if simple responses was never defined in the template.

Encases the `if` statement in a `None` check to only print if the property was defined. This PR also adds a typing check to the property, defaulting to `False` if an invalid type was defined.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/lib/swagger/parser.py
+++ b/samcli/commands/local/lib/swagger/parser.py
@@ -137,14 +137,20 @@ class SwaggerParser:
                 )
 
             enable_simple_response = authorizer_object.get(SwaggerParser._AUTHORIZER_SIMPLE_RESPONSES, False)
-            if event_type != Route.HTTP or payload_version != LambdaAuthorizer.PAYLOAD_V2 and enable_simple_response:
+
+            if (
+                event_type != Route.HTTP
+                or payload_version != LambdaAuthorizer.PAYLOAD_V2
+                or not isinstance(enable_simple_response, bool)
+            ):
                 enable_simple_response = False
 
-                LOG.warning(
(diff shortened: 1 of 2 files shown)


samcli/commands/local/lib/swagger/parser.py
```diff
@@ -137,14 +137,20 @@ def get_authorizers(self, event_type: str = Route.API) -> Dict[str, Authorizer]:
...
+                or not isinstance(enable_simple_response, bool)
+            ):
                 enable_simple_response = False
 
-                LOG.warning(
-                    "Simple responses are only available on HTTP APIs with payload version "
-                    "2.0, ignoring for Lambda authorizer '%s'",
-                    auth_name,
-                )
+                if authorizer_object.get(SwaggerParser._AUTHORIZER_SIMPLE_RESPONSES) is not None:
```
sriram-mv, line 148: minor nit: `if authorizer_object.get(SwaggerParser._AUTHORIZER_SIMPLE_RESPONSES, None)` might be enough and more pythonic. Not blocking.

sriram-mv (approved):

samcli/commands/local/lib/swagger/parser.py
lucashuy: Yup, the default return value is `None`, so I decided to leave that unpopulated. As for the `is not None` check, this is here since the customer can provide `0` or `""` and the statement would be false, so `None` is explicitly checked here.

mildaniel (approved):

PR #4998, opened 2023-04-11
feat: logs powered by `rich`

sriram-mv:
better UX

After:

Screen Shot 2023-04-10 at 11 12 31 PM

* Better UX overall and opens the pathway to better UX experiences powered by `rich`

#### Testing
```
sam local invoke
sam init
sam logs
sam traces
sam build
sam build --use-container
```

* N/A, only setting `rich` if its a tty otherwise we go back to a StreamHandler.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/installer/assets/THIRD-PARTY-LICENSES
+++ b/installer/assets/THIRD-PARTY-LICENSES
@@ -670,7 +670,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 The above BSD License Applies to all code, even that also covered by Apache 2.0.
 
 --------------------------------------------------------------------------------
-** pygemnts; version 2.15.0 -- https://pypi.org/project/pygemnts/
+** pygments; version 2.15.0 -- https://pypi.org/project/pygemnts/
 
 Copyright (c) 2006-2022 by the respective authors (see AUTHORS file).
 
--- a/installer/assets/THIRD-PARTY-LICENSES
+++ b/installer/assets/THIRD-PARTY-LICENSES
@@ -669,7 +669,36 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 The above BSD License Applies to all code, even that also covered by Apache 2.0.
 
+--------------------------------------------------------------------------------
+** pygemnts; version 2.15.0 -- https://pypi.org/project/pygemnts/
+
+Copyright (c) 2006-2022 by the respective authors (see AUTHORS file).
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice,
+      this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice,
+      this list of conditions and the following disclaimer in the documentation
+      and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 ------
+
 
 ** binaryornot; version 0.4.4 -- https://pypi.org/project/binaryornot/
 Copyright (c) 2013, Audrey Roy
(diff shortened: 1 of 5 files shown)


installer/assets/THIRD-PARTY-LICENSES
```diff
@@ -1991,6 +1991,7 @@ Copyright (c) 2019-2021 Anthon van der Neut, Ruamel bvba
 ** cfn-lint; version 0.72.2 -- https://pypi.org/project/cfn-lint/
 ** jschema-to-python; version 1.2.3 -- https://pypi.org/project/jschema-to-python/
 ** junit-xml; version 1.9 -- https://pypi.org/project/junit-xml/
+** rich; version 13.3.3 -- https://pypi.org/project/rich/
```
jfuss, line 1994: IIRC, we need to include all dependencies not just direct.

samcli/lib/utils/sam_logging.py
```diff
@@ -28,9 +32,12 @@ def configure_logger(logger, formatter, level):
         if handlers:
             log_stream_handler = handlers[0]
         else:
-            log_stream_handler = logging.StreamHandler()
+            log_stream_handler = (
+                RichHandler(console=Console(stderr=True), show_time=False, show_path=False, show_level=False)
+                if sys.stderr.isatty()
```
jfuss, line 37: Why only `stderr`?

requirements/base.txt
```diff
@@ -16,6 +16,7 @@ serverlessrepo==0.1.10
 aws_lambda_builders==1.29.0
 tomlkit==0.11.7
 watchdog==2.1.2
+rich~=13.3.3
```
mndeveci, line 19: Any python version requirement that we need to be careful about?

samcli/lib/utils/sam_logging.py
```diff
@@ -28,9 +32,12 @@ def configure_logger(logger, formatter, level):
         if handlers:
             log_stream_handler = handlers[0]
         else:
-            log_stream_handler = logging.StreamHandler()
+            log_stream_handler = (
+                RichHandler(console=Console(stderr=True), show_time=False, show_path=False, show_level=False)
```
mndeveci, line 36: Can we define as many `Console` instances as possible, or should we define one and use it during the whole execution?

samcli/lib/utils/sam_logging.py
sriram-mv: logging is to stderr on the stream handler.

requirements/base.txt
sriram-mv: rich is supported for 3.7+ which maps to AWS SAM CLI.

samcli/lib/utils/sam_logging.py
sriram-mv: we re-use once its configured AFAIK within the scope of the process, that's the existing methodology. But good to check.

sriram-mv:
@jfuss , Good catch, we can check for `NO_COLOR`.

samcli/lib/utils/sam_logging.py
mndeveci: I see, but how can someone access `Console` instance from somewhere else? Let's say we wanted to add progress bars to build/package/deploy steps, where they can find the previously instantiated `Console` instance?

sriram-mv: At this point you can't. since the console instance is exclusive to log_stream_handler. however in a future PR we could move the console instance to thread local storage? thoughts?

samcli/lib/utils/sam_logging.py
```diff
@@ -28,9 +35,12 @@ def configure_logger(logger, formatter, level):
...
+                if sys.stderr.isatty() and os.getenv(LOGGING_COLOR_ENV_VAR) != "1"
```
jfuss, line 40: Shouldn't this be an `or`?

samcli/lib/utils/sam_logging.py
```diff
@@ -2,13 +2,20 @@
...
+from rich.console import Console
+from rich.logging import RichHandler
 
 SAM_CLI_FORMATTER = logging.Formatter("%(message)s")
 SAM_CLI_FORMATTER_WITH_TIMESTAMP = logging.Formatter("%(asctime)s | %(message)s")
 
 SAM_CLI_LOGGER_NAME = "samcli"
 LAMBDA_BULDERS_LOGGER_NAME = "aws_lambda_builders"
 
+LOGGING_COLOR_ENV_VAR = "NO_COLOR"
```
jfuss, line 17: Should the Env Var be SAM CLI specific?

jfuss (changes requested):

samcli/lib/utils/sam_logging.py
sriram-mv: It has be "and". You could be on an interactive terminal and have NO_COLOR set to 1, if that is the case you get the normal stream logger initialized.

installer/assets/THIRD-PARTY-LICENSES
```diff
@@ -669,8 +669,37 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 The above BSD License Applies to all code, even that also covered by Apache 2.0.
 
+--------------------------------------------------------------------------------
+** pygemnts; version 2.15.0 -- https://pypi.org/project/pygemnts/
```
mndeveci, line 673: typo for the package name;
```suggestion
** pygments; version 2.15.0 -- https://pypi.org/project/pygemnts/
```

samcli/lib/utils/sam_logging.py
```diff
@@ -28,9 +38,19 @@ def configure_logger(logger, formatter, level):
...
             log_stream_handler = handlers[0]
         else:
-            log_stream_handler = logging.StreamHandler()
+            log_stream_handler = (
+                RichHandler(console=Console(stderr=True), show_time=False, show_path=False, show_level=False)
+                if sys.stderr.isatty()
+                and not any(
+                    [
+                        os.getenv(LOGGING_COLOR_ENV_VAR),
+                        os.getenv(SAM_LOGGING_COLOR_ENV_VAR),
```
mndeveci, line 47: Should this environment variables called `NO_LOGGING..` and `SAM_NO_LOGGING..` to make it more readable?

mndeveci (changes requested):

installer/assets/THIRD-PARTY-LICENSES
sriram-mv: Done

samcli/lib/utils/sam_logging.py
sriram-mv: Done

mndeveci (approved):

Issue #5019, opened 2023-04-14
Bug: local start-api breaks under Podman


com4:
### Description:

Running `sam local start-api` while using podman will load the first request to an endpoint, but return a 502 for all subsequent calls.

### Steps to reproduce:

```bash
sam local start-api

curl http://127.0.0.1:3000/hello
# {"message": "hello world"}
curl http://127.0.0.1:3000/hello
# {"message":"Internal server error"}
```

### Observed result:

[...]

PR #5020, opened 2023-04-14
fix: Favor a stale image when daemon doesn't support registry query

com4:

#5019

Podman doesn't support the distribution API. This is used when querying to see if a docker image is up-to-date.

containers/podman#17726

Instead of crashing this treats a stale image as good enough.

Users of podman might test locally on an outdated base image.

#### Mandatory Checklist

- [n/a] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [n/a] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [x] `make pr` passes
- [n/a] `make update-reproducible-reqs` if dependencies were changed
- [x] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_image.py
+++ b/samcli/local/docker/lambda_image.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import Optional
 
 import docker
+from docker.errors import NotFound as DockerNotFound
 
 from samcli.commands.local.cli_common.user_exceptions import ImageBuildException
 from samcli.commands.local.lib.exceptions import InvalidIntermediateImageError
@@ -447,7 +448,23 @@ class LambdaImage:
         if self.skip_pull_image or self.force_image_build:
             return
 
-        if self.is_base_image_current(image_name):
+        # if we can't check to see if the image is up-to-date, warn but don't crash.
+        try:
+            base_image_is_current = self.is_base_image_current(image_name)
+        except DockerNotFound as e:
+            # Assume the base image is the latest because we got a 404 on the API from
+            # the Docker daemon (This is different than the image is missing)
+            # https://github.com/containers/podman/issues/17726
+            LOG.warning(
+                "Cannot check if base image is current because a 404 error was returned from "
+                "the Docker daemon. This might be due to an incompatible Docker engine clone. "
+                "Proceeding with possibly stale image."
+            )
+            LOG.debug("Error 404 response from Docker Engine", exc_info=e)
+            self.skip_pull_image = True
+            return
+
+        if base_image_is_current:
             self.skip_pull_image = True
             LOG.info("Local image is up-to-date")
         else:
--- a/tests/unit/local/docker/test_lambda_image.py
+++ b/tests/unit/local/docker/test_lambda_image.py
@@ -5,7 +5,7 @@ from unittest import TestCase
 from unittest.mock import patch, Mock, mock_open, ANY, call
 from parameterized import parameterized
 
-from docker.errors import ImageNotFound, BuildError, APIError
+from docker.errors import ImageNotFound, BuildError, APIError, NotFound
 from parameterized import parameterized
 
 from samcli.commands.local.lib.exceptions import InvalidIntermediateImageError
@@ -737,3 +737,14 @@ class TestLambdaImage(TestCase):
         lambda_image._check_base_image_is_current("image_name")
         self.assertEqual(lambda_image.skip_pull_image, expected_skip_pull_image)
         self.assertEqual(lambda_image.force_image_build, expected_force_image_build)
+
+    def test_check_base_image_is_current_404_error_from_daemon(self):
+        # Verify that we skip pulling the image when we get a 404 from the Docker daemon's API.
+        # This happens when a Docker engine clone, such as podman hasn't implemented the
+        # necessary APIs.
+        # See issue #5019
+        lambda_image = LambdaImage("layer_downloader", False, False, docker_client=Mock())
+        lambda_image.is_base_image_current = Mock(side_effect=NotFound("404"))
+        lambda_image._check_base_image_is_current("image_name")
+        self.assertEqual(lambda_image.skip_pull_image, True)
+        self.assertEqual(lambda_image.force_image_build, False)


samcli/local/docker/lambda_image.py
```diff
@@ -447,7 +448,23 @@ def _check_base_image_is_current(self, image_name: str) -> None:
...
+            # Assume the base image is the latest because we got a 404 on the API from
+            # the Docker daemon (This is different than the image is missing)
+            # https://github.com/containers/podman/issues/17726
+            LOG.warning(
+                "Cannot check if base image is current because a 404 error was returned from "
+                "the Docker daemon. This might be due to an incompatible Docker engine clone. "
+                "Proceeding with possibly stale image."
+            )
+            LOG.debug("Error 404 response from Docker Engine", exc_info=e)
+            self.skip_pull_image = True
```
jfuss, line 464: If we can't tell if the image is current, shouldn't we pull the image unless `--skip-pull-image` was set on the command line?

jfuss (changes requested):

samcli/local/docker/lambda_image.py
com4: unfortunately image freshness is checked on every reload of the endpoint so simply pulling when we can't check means it pulls the image on every refresh. 

would it work to pull a new image when we can't check the base image version the first time and then `skip_pull_image` on reloads?

```diff
diff --git a/samcli/local/docker/lambda_image.py b/samcli/local/docker/lambda_image.py
index 27ac817a..443fd9e5 100644
--- a/samcli/local/docker/lambda_image.py
+++ b/samcli/local/docker/lambda_image.py
@@ -142,20 +142,23 @@ class LambdaImage:
             The image to be used (REPOSITORY:TAG)
         """
         base_image = None
+        # It's possible with some Docker engine clones that we can't check the latest
+        # image version. This flag marks this case so the image is only pulled once.
+        cant_check_image_version = False
         tag_prefix = ""
 
         if packagetype == IMAGE:
@@ -185,20 +188,32 @@ class LambdaImage:
         image_not_found = False
 
         # If we are not using layers, build anyways to ensure any updates to rapid get added
         try:
             self.docker_client.images.get(rapid_image)
             # Check if the base image is up-to-date locally and modify build/pull parameters accordingly
             self._check_base_image_is_current(base_image)
         except docker.errors.ImageNotFound:
             LOG.info("Local image was not found.")
             image_not_found = True
+        except DockerNotFound as e:
+            # Assume the base image is not the latest because we got a 404 on the API from
+            # the Docker daemon (This is different than the image is missing)
+            # https://github.com/containers/podman/issues/17726
+            LOG.warning(
+                "Cannot check if base image is current because a 404 error was returned from "
+                "the Docker daemon. This might be due to an incompatible Docker engine clone. "
+                "Pulling image."
+            )
+            LOG.debug("Error 404 response from Docker Engine", exc_info=e)
+            cant_check_image_version = True
+            image_not_found = True
 
         # If building a new rapid image, delete older rapid images
         if image_not_found and rapid_image == f"{image_repo}:{tag_prefix}{RAPID_IMAGE_TAG_PREFIX}-{architecture}":
             if tag_prefix:
                 # ZIP functions with new RAPID format. Delete images from the old ecr/sam repository
                 self._remove_rapid_images(f"{self._SAM_INVOKE_REPO_PREFIX}-{runtime}")
             else:
                 self._remove_rapid_images(image_repo)
 
         if (
@@ -207,20 +222,23 @@ class LambdaImage:
                 image if image else base_image, rapid_image, downloaded_layers, architecture, stream=stream_writer
             )
 
+        if cant_check_image_version:
+            # we're unable to check for the latest image and we've already pulled once
+            # so skip pulling new image from now on.
+            LOG.warning("Skipping base image updates until the daemon is restarted.")
+            self.skip_pull_image = True
+
         return rapid_image
 
     def get_config(self, image_tag):
```

com4: @jfuss just following up. does this, or another alternative work for when to pull a fresh base image when we can't detect the latest version so we aren't pulling an image each time an api is called during local development?

jfuss: @com4 Sorry for the longer response. I am not sure if we do the freshness check on each invoke for a specific reason or now. @mndeveci or @moelasmar might have a better understanding. 

I think if we pull a fresh image once, that is probably more acceptable. The situation I want to try to avoid is you are running a 3 month old version of the container locally but doesn't match the latest version of Lambda. I think this is just a balance between always downloading a fresh image each time vs once a command run (which seems more appropriate).

com4: @jfuss understood.

the patch in the comment would keep the old image if the user didn't restart `sam local start-api` for 3 months.

of note, this does leave the current functionality the same for docker desktop users. that is, each time they hit the endpoint sam will continue checking to see if there is a new base image to install and install it. when podman implements the distribution api, podman users will also have an image version check (and auto-update when needed) on each api request with no additional changes to the code. 

i'm not a big fan of the idea personally, but as an alternative we could change the bool to a counter so every Xth request blindly pulls a new base image when it can't verify the version.  for example: every 100th request if we can't verify we're on the latest base image, just download the latest and use that.

com4: > If docker throws: (ImageNotFound),  something else that extends NotFound, this may cause unexpected issues. 

Good point.

> I think where we are leaning is to handle the failure but not enable skip_pull_image and let customers define this through the command line.

I feel that querying the registry on every request is over the top to avoid running a "very" outdated version which is why these solutions are setting the `skip_pull_image` flag. In my workflow (so naturally everyone else's too ;-)) I'd only risk running an outdated version for maybe a day. I don't know that I have ever left a development server running for a week or more. 

With that said, would removing the skip toggle from the code and suggesting the user re-run the daemon with `--skip-pull-image` to stop it from pulling an image on every refresh suffice?.  So something like this: 

```python
except DockerNotFound as e:
    # See Also: https://github.com/containers/podman/issues/17726
    LOG.warning(
        "Cannot check if base image is current because an unknown 404 error was returned. "
        "This might be due to an incompatible Docker engine clone. Consider rerunning "
        "with the --skip-pull-image argument. Pulling image."
    )
    LOG.debug("Error unknown 404 response from Docker Engine", exc_info=e)
```

Without the flag a broken docker daemon would pull the image on every request slightly irritating the user with slow response times and battery drain. 

Including the --skip-pull-image suggestion would of course run the risk of the user unintentionally including that flag as they hit up in the shell to rerun the command resulting in a very old image being used so I'm also open to replacing the suggestion for the flag with a suggestion that "If this problem persists try reinstalling the official Docker engine" or something similar.

> I think at the very least, we need to check what kind of status code it is...

From the traceback in the bug, what we're getting is `docker.errors.NotFound: 404 Client Error: Not Found ("b'Not Found'")`. I think the 404 status code will be the same for an ImageNotFound from the registry and the base NotFound we're getting from the daemon so there's not a good way to differentiate between the two types of errors. `is_client_error` will return `True` for both without some upstream work.

An alternative to checking status code would be to check that the exception caught by `except DockerNotFound` is actually that class instead of a subclass. All together it'd look something like this:

```python
except DockerNotFound as e:
    if not hasattr(e, "__class__") or e.__class__ is not DockerNotFound:
        # This is most likely a subclass of the base exception we're trying to catch
        # so raise the error that way it can be handled.
        raise
        
    # Otherwise we aren't able to check for new images so pull a fresh one to be safe.
    # See Also: https://github.com/containers/podman/issues/17726
    LOG.warning(
        "Cannot check if base image is current because an unknown 404 error was returned. "
        "This might be due to an incompatible Docker engine clone. Consider rerunning "
        "with the --skip-pull-image argument. Pulling image."
    )
    LOG.debug("Error unknown 404 response from Docker Engine", exc_info=e)
```

com4: > I think we should wrap this into a "DockerDistibutionAPIError" error 

Happy to do this. Should `DockerDistributIonAPIError` be added to `samcli.commands.local.cli_common.user_exceptions` along side `ImageBuildException` or somewhere else? It looks like there are a few exception modules.

> we don't want the full traceback thrown into the console

Do you also want me to remove the LOG.debug line? 

---
All together with a bit of extra context:
```python
        # If we are not using layers, build anyways to ensure any updates to rapid get added
        try:
            self.docker_client.images.get(rapid_image)
            # Check if the base image is up-to-date locally and modify build/pull parameters accordingly
            self._check_base_image_is_current(base_image)
        except docker.errors.ImageNotFound:
            LOG.info("Local image was not found.")
            image_not_found = True
        except docker.errors.APIError as e:
            if e.__class__ is docker.errors.NotFound:
                # A generic "NotFound" is raised when we aren't able to check the image version
                # for example when the docker daemon's api doesn't support this action.
                #
                # See Also: https://github.com/containers/podman/issues/17726
                LOG.warning(
                    "Cannot check if base image is current because an unknown 404 error was returned. "
                    "This might be due to an incompatible Docker engine clone. Consider rerunning "
                    "with the --skip-pull-image argument to improve speed but may cause you running outdated containers locally."
                )
                image_not_found = True
            else:
                raise DockerDistributionAPIError("Unknown API error received from docker") from e

        # If building a new rapid image, delete older rapid images
        if image_not_found and rapid_image == f"{image_repo}:{tag_prefix}{RAPID_IMAGE_TAG_PREFIX}-{architecture}":
```

This results in this output from the log (on each refresh):
```
Invoking app.lambda_handler (python3.10)
Cannot check if base image is current because an unknown 404 error was returned. This might be due to
an incompatible Docker engine clone. Consider rerunning with the --skip-pull-image argument to improve
speed but may cause you running outdated containers locally.
Removing rapid images for repo public.ecr.aws/sam/emulation-python3.10
Building image..............................
Using local image: public.ecr.aws/lambda/python:3.10-rapid-arm64.

Mounting /--snip--/sam-hello-world/hello_world as /var/task:ro,delegated, inside runtime container
START RequestId: aa3c1005-07a8-40c2-9bcf-2d00cac999a0 Version: $LATEST
END RequestId: aa3c1005-07a8-40c2-9bcf-2d00cac999a0
REPORT RequestId: aa3c1005-07a8-40c2-9bcf-2d00cac999a0	Init Duration: 0.19 ms	Duration: 45.95 ms	Billed Duration: 46 ms	Memory Size: 128 MB	Max Memory Used: 128 MB	
No Content-Type given. Defaulting to 'application/json'.
2023-04-26 11:25:44 127.0.0.1 - - [26/Apr/2023 11:25:44] "GET /hello HTTP/1.1" 200 -
```

jfuss: > Happy to do this. Should DockerDistributIonAPIError be added to samcli.commands.local.cli_common.user_exceptions along side ImageBuildException or somewhere else? It looks like there are a few exception modules.

Seems reasonable to me. At least matches others we have.

> Do you also want me to remove the LOG.debug line?

Ya. I think it's valuable.

com4:
> ```
> Unknown 404 - Unable to check if base image is current.
> 
> Possible incompatible Docker engine clone employed. Consider `--skip-pull-image` for improved speed, the tradeoff being not running the latest image.
> ```

I've updated the PR to use this error message. Log now looks like:

```
Invoking app.lambda_handler (python3.10)
Unknown 404 - Unable to check if base image is current.

Possible incompatible Docker engine clone employed. Consider `--skip-pull-image` for improved speed, the tradeoff being not running the latest image.
Removing rapid images for repo public.ecr.aws/sam/emulation-python3.10
Building image..............................
Using local image: public.ecr.aws/lambda/python:3.10-rapid-arm64.

Mounting /--snip--/sam-hello-world/hello_world as /var/task:ro,delegated, inside runtime container
START RequestId: 8a25bb51-d230-4c34-8c4f-af5c91e7ad94 Version: $LATEST
START RequestId: 8a25bb51-d230-4c34-8c4f-af5c91e7ad94 Version: $LATEST
END RequestId: 8a25bb51-d230-4c34-8c4f-af5c91e7ad94
```

jfuss (approved):
@com4 Thanks for the discussion and the patch!

com4:
> @com4 Thanks for the discussion and the patch!

Thanks again for your help @jfuss 

Just to close the loop, running the following command gives me the UX that i prefer but happy to chat more about it in discussions if you'd like.

```bash
# pull the latest image by invoking a function once, then reuse that image for the duration of the api 
# session
sam local invoke HelloWorldFunction && sam local start-api --skip-pull-image
```

Hopefully it won't be too much longer until podman supports the API.

sriram-mv (approved):

PR #5023, opened 2023-04-14
feat: help text for all local suite.

sriram-mv:
N/A

Better UX across local suite of commands

Better Option Grouping and examples. Tests pending.

##### Validate 
validate

##### Local
local
lambda
api

event

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/options.py
+++ b/samcli/commands/local/cli_common/options.py
@@ -46,7 +46,7 @@ def local_common_options(f):
             "--shutdown",
             is_flag=True,
             default=False,
-            help="If set, will emulate a shutdown event after the invoke completes, "
+            help="Emulate a shutdown event after invoke completes, "
             "in order to test extension handling of shutdown behavior.",
         ),
         click.option(
@@ -54,9 +54,9 @@ def local_common_options(f):
             default="localhost",
             show_default=True,
             help="Host of locally emulated Lambda container. "
-            "This option is useful when the container runs on a different host than SAM CLI. "
-            "For example, if you want to run SAM CLI in a Docker container on macOS, "
-            "use this option with host.docker.internal",
+            "This option is useful when the container runs on a different host than AWS SAM CLI. "
+            "For example, if one wants to run AWS SAM CLI in a Docker container on macOS, "
+            "this option could specify `host.docker.internal`",
         ),
         click.option(
             "--container-host-interface",
@@ -72,11 +72,11 @@ def local_common_options(f):
             required=False,
             multiple=True,
             help="Container image URIs for invoking functions or starting api and function. "
-            "You can specify the image URI used for the local function invocation "
+            "One can specify the image URI used for the local function invocation "
             "(--invoke-image public.ecr.aws/sam/build-nodejs14.x:latest). "
-            "You can specify for each individual function with "
+            "One can also specify for each individual function with "
             "(--invoke-image Function1=public.ecr.aws/sam/build-nodejs14.x:latest). "
-            "If a function does not have invoke image specified, the default SAM CLI "
+            "If a function does not have invoke image specified, the default AWS SAM CLI "
             "emulation image will be used.",
         ),
     ]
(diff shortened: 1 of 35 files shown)


samcli/commands/local/cli_common/options.py
```diff
@@ -173,16 +173,16 @@ def invoke_common_options(f):
                 "--docker-volume-basedir",
                 "-v",
                 envvar="SAM_DOCKER_VOLUME_BASEDIR",
-                help="Specifies the location basedir where the SAM file exists. If the Docker is running on "
-                "a remote machine, you must mount the path where the SAM file exists on the docker machine "
-                "and modify this value to match the remote machine.",
+                help="Specify the location basedir where the SAM template exists. If Docker is running on "
```
mildaniel, line 176: Thinking out loud question: should we keep specifying "Docker" in our help text/docs or should we move away from that and start using terminology like "Container engine"?

samcli/commands/local/cli_common/options.py
```diff
@@ -173,16 +173,16 @@ def invoke_common_options(f):
...
+                "a remote machine, Path of the SAM template must be mounted on the docker machine "
```
mildaniel, line 177:
```suggestion
                "a remote machine, Path of the SAM template must be mounted on the Docker machine "
```

samcli/commands/local/invoke/cli.py
```diff
@@ -21,23 +22,27 @@
...
-\b
-Invoking a Lambda function without an input event
-$ sam local invoke "HelloWorldFunction"\n
-\b
-Invoking a Lambda function using an event file
-$ sam local invoke "HelloWorldFunction" -e event.json\n
-\b
-Invoking a Lambda function using input from stdin
-$ echo '{"message": "Hey, are you there?" }' | sam local invoke "HelloWorldFunction" --event - \n
+    Invoke AWS serverless functions locally.
```
mildaniel, line 25: I know we use serverless functions interchangeably with lambda functions, but I'm wondering if the description here should list both in case customers are confused about it working with `AWS::Severless::..` resources versus non serverless resources.

I might be thinking too much into this and it might be overly nit-picky, especially considering the description below. Feel free to ignore this if you think it's not necessary.

samcli/commands/local/invoke/core/options.py
```diff
@@ -0,0 +1,85 @@
...
+)
+
+OPTIONS_INFO: Dict[str, Dict] = {
+    "Required Options": {"option_names": {opt: {"rank": idx} for idx, opt in enumerate(REQUIRED_OPTIONS)}},
+    "Template Options": {"option_names": {opt: {"rank": idx} for idx, opt in enumerate(TEMPLATE_OPTIONS)}},
+    "AWS Credential Options": {
+        "option_names": {opt: {"rank": idx} for idx, opt in enumerate(AWS_CREDENTIAL_OPTION_NAMES)}
+    },
+    "Container Options": {"option_names": {opt: {"rank": idx} for idx, opt in enumerate(CONTAINER_OPTION_NAMES)}},
+    "Beta Options": {"option_names": {opt: {"rank": idx} for idx, opt in enumerate(BETA_OPTION_NAMES)}},
```
hawflau, line 69: Does "Beta Options" has to be before "Artifact Location Options", "Extension Options" and "Configuration Options"? 

Asking because in another PR #5062, we intend to move "Beta Options" to part of the common options along with "Other Options". And they will be always at the bottom.

tests/unit/commands/local/start_lambda/core/test_options.py
```diff
@@ -0,0 +1,11 @@
...
+
+from samcli.commands.local.start_lambda.cli import cli
+from samcli.commands.local.start_lambda.core.options import ALL_OPTIONS
+
+
+class TestOptions(TestCase):
+    def test_all_options_formatted(self):
+        command_options = [param.human_readable_name for param in cli.params]
+        command_options = [command_option for command_option in command_options if command_option is not None]
+        self.assertEqual(sorted(ALL_OPTIONS), sorted(command_options))
```
hawflau, line 11: I'm a bit not sure if this is the right thing to assert equal. `cli.params` does not include options that's automatically added by click (e.g. `--help`). `cli.pararms` also includes arguments, which I believe would not be added to `ALL_OPTIONS`

hawflau (commented):
LGTM. Just want to clarify with two questions before approving.

tests/unit/commands/local/start_lambda/core/test_options.py
sriram-mv: Open to a better testing strategy overall.

samcli/commands/local/invoke/core/options.py
sriram-mv: It's in that place because the beta options were closely tied to container options at that point. I do not have concerns moving it down. Should we then indicate with UX markings that a particular command has beta options up top somewhere before listing the actual options?

My concern is that users do not see them if we move them down.

tests/unit/commands/local/start_lambda/core/test_options.py
hawflau: I think we can just simply compare these two:
1. set of `ALL_OPTIONS` excluding "help"
2. set of `cli.params` excluding `click.core.Argument` type

samcli/commands/local/invoke/core/options.py
hawflau: The fact that `OPTIONS_INFO` is a dict guarantees the keys' iteration order are always the same as the order they were inserted. So if we have "Beta Options" here at an upper position, the `add_common_options_info` function in the other PR won't change the ordering. We can still keep "Beta Options" as it is if we prefer it at an upper position.

> Should we then indicate with UX markings that a particular command has beta options up top somewhere before listing the actual options?
Not sure what it looks like, but open to it.

> My concern is that users do not see them if we move them down.
IMO it's easier for users to see if it's at bottom, because you usually see the bottom of previous command run when you type the new command. But I guess what we want is a consensus. I personally don't have a strong preference.

samcli/commands/local/cli_common/options.py
sriram-mv: Good call, I'd keep that to separate PR to address those concerns.

mildaniel (approved):

hawflau (approved):

PR #5039, opened 2023-04-18
design: Generalized resource linker design

mildaniel:

N/A

For Terraform to support additional resources, it needs to link resources together.

Creates a design to generalize the existing Function-Layer linking logic to support any two resources.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/designs/resource_linking_generalized.md
```diff
@@ -0,0 +1,153 @@
...
+`_process_resolved_resources` &larr; `_process_resolved_layers` 
+
+To generalize these functions: 
+- All instances of hard-coded properties need to be replaced with the corresponding values
+defined in the `ResourceLinkingPair` instance
+- Exception messages should be updated to be more generic
+- The `_update_mapped_parent_resource_with_resolved_child_resources` method should be updated to support writing other intrinsic types.
+
+
+### Adding New Resource Links
```
moelasmar, line 149: we need also to change the linking exceptions we throw (mainly the limitations exceptions) for each linking pair. I think it can be done by adding a new Exception Type, and in the new `ResourceLinkingPair` instance we can set the Exceptions type to be used.

moelasmar (changes requested):
In general it looks good to me, but we need to add some part of the limitation exceptions to be customized for each linking pairs.

samcli/hook_packages/terraform/designs/resource_linking_generalized.md
mildaniel: The only Function/Layer specific exception I can see is `OneLambdaLayerLinkingLimitationException`. Are there any others?

lucashuy (approved):

moelasmar (approved):

Issue #5037, opened 2023-04-18
sam 1.80 authorizer support doesn't work then the authorizer is defined with a FunctionArn


ylilarry:
### Description:

The new `sam local start-api` authorizer support works only if the authorizer function is defined in the same template. It breaks if the function is defined by a FunctionArn.

For example:

      Auth: # Uses the UserAppTokenAuthorizer to identify the session user.
        ApiKeyRequired: false
        DefaultAuthorizer: UserAppTokenAuthorizer
        Authorizers:
          UserAppTokenAuthorizer:
            FunctionPayloadType: REQUEST
            FunctionArn: arn:aws:lambda:us-east-1:887279901853:function:UserAppTokenAuthorizer
            Identity:
              ReauthorizeEvery: 300
              Headers: ["X-FAST-API-KEY"] # This is important

Breaks with an error saying `SAM: UserAppTokenAuthorizer not found`.
    
        Traceback (most recent call last):
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
            response = self.full_dispatch_request()
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/flask/app.py", line 1518, in full_dispatch_request
            rv = self.handle_user_exception(e)
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/flask/app.py", line 1516, in full_dispatch_request
            rv = self.dispatch_request()
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/flask/app.py", line 1502, in dispatch_request
            return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/samcli/local/apigw/local_apigw_service.py", line 675, in _request_handler
            self._invoke_parse_lambda_authorizer(lambda_authorizer, auth_lambda_event, route_lambda_event, route)
          File "/Users/yu/lab/FastchargeAPI/.venv/lib/python3.9/site-packages/samcli/local/apigw/local_apigw_service.py", line 745, in _invoke_parse_lambda_authorizer
            lambda_auth_response = self._invoke_lambda_function(lambda_authorizer.lambda_name, auth_lambda_event)
[...]

PR #5040, opened 2023-04-18
fix: Catches exception raised if a function is not found

lucashuy:

#5037

Lambda authorizers with function ARNs referencing a remotely deployed Lambda function will still try to be invoked, even though the Lambda function is not defined or created locally.

Handles the `FunctionNotFound` exception by logging it and continuing with execution.

Customers that have deployed Lambda functions referenced in their templates when running `sam local start-api` will experience the same behaviour as before 1.80.0: Lambda authorizers will not end up validating requests and instead will continue to the route's function.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/apigw/local_apigw_service.py
```diff
@@ -742,6 +754,7 @@ def _invoke_parse_lambda_authorizer(
         route: Route
             The route that is being called
         """
+        raise ValueError()
```
mndeveci, line 757: Left over from testing?

mndeveci (approved):

samcli/local/apigw/local_apigw_service.py
lucashuy: Removed

sriram-mv (approved):

samcli/local/apigw/local_apigw_service.py
```diff
@@ -690,8 +701,9 @@ def _request_handler(self, **kwargs):
             )
 
             if lambda_authorizer_exception:
-                LOG.error("Lambda authorizer failed to invoke successfully: %s", lambda_authorizer_exception.message)
+                LOG.error("Lambda authorizer failed to invoke successfully: %s", exception_name)
```
qingchm, line 704: Is it by design that no return calls are made for this condition?

lucashuy: There is a separate condition to return Flask service errors if they were created during the catching of other exceptions in this try block, but we don't re-raise these exceptions since these would likely indicate a user related exception (eg. unauthorized request, invalid response format)

Issue #4909, opened 2023-03-22
Bug: `sam list resources` command only works for first 100 resources.


awood45:
### Description:

When I run `sam list resources` on a template with > 100 resources (~140 in this case) a number of resources have "-" as the PhysicalResourceId in the resulting map.

I strongly wager that this is an issue with pagination, because I believe that `DescribeStackResources` returns 100 resources per page, and you need to look for a response truncation and continue to iterate if the response is truncated (has a NextToken value or similar).

### Steps to reproduce:

1. Create a SAM/CFN template with more than 100 resources.
2. Deploy it by `sam deploy` or any other means.
3. Run `sam list resources` against that template, and note that "-" is present for `num_resources - 100` resources.

### Observed result:

Described above in detail.

### Expected result:

All resources with valid `PhysicalResourceId` values should have them present in the response.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOS
4. `sam --version`: `1.75.0`
5. AWS region: `eu-south-1`

```
# Paste the output of `sam --info` here
{
  "version": "1.75.0",
  "system": {
    "python": "3.11.2",
    "os": "macOS-12.6.2-x86_64-i386-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "Not available",
    "aws_cdk": "2.69.0 (build 60a5b2a)",
    "terraform": "Not available"
  }
}
```

`Add --debug flag to command you are running`

moelasmar:
thanks @awood45 .

PR #5097, opened 2023-05-01
fix(list): Handle Stacks with >100 resources

jfuss:
Previously, we were using `describe_stack_resources` to get all the Resources from a Stack. However, this will only ever return the first 100 resources (per API docs). Instead, we move to using the `list_stack_resources` paginator (per API docs). This allows us to get all the resources within a stack. The response is different from `describe_stack_resources` and `list_stack_resources` but the difference is not meaningful for the current implementation.

As apart of this, I updated the unit tests. Previously, we were using another class to get a CFN Client to mock. Instead of mocking out a file that is not used within the `resource_mapping_producer.py`, I remove that and mocked out the `cfn_client` directly. This makes the tests easier to follow and removes unneeded mocking of classes/methods that are not used within the implementation.

#4909

Fixes github issues linked.

See above commit details

None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/list/resources/resource_mapping_producer.py
+++ b/samcli/lib/list/resources/resource_mapping_producer.py
@@ -62,10 +62,15 @@ class ResourceMappingProducer(Producer):
         """
 
         try:
-            response = self.cloudformation_client.describe_stack_resources(StackName=self.stack_name)
-            if "StackResources" not in response:
-                return {"StackResources": []}
-            return response
+            all_resources = []
+            cfn_paginator = self.cloudformation_client.get_paginator("list_stack_resources")
+            for response in cfn_paginator.paginate(StackName=self.stack_name):
+                if "StackResourceSummaries" not in response:
+                    continue
+
+                all_resources.extend(response.get("StackResourceSummaries", []))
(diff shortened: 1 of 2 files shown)


samcli/lib/list/resources/resource_mapping_producer.py
```diff
@@ -62,10 +62,15 @@ def get_resources_info(self):
...
 
         try:
-            response = self.cloudformation_client.describe_stack_resources(StackName=self.stack_name)
-            if "StackResources" not in response:
-                return {"StackResources": []}
-            return response
+            all_resources = []
+            cfn_paginator = self.cloudformation_client.get_paginator("list_stack_resources")
+            for response in cfn_paginator.paginate(StackName=self.stack_name):
+                if "StackResourceSummaries" not in response:
```
sriram-mv, line 68: nit: StackResourceSummaries to a constant.

sriram-mv (approved):

hawflau (approved):

PR #5107, opened 2023-05-02
fix: switch to `rich` for color logging.

sriram-mv:
- maintain click backcompatability with interface.
- tests added

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/sync/command.py
+++ b/samcli/commands/sync/command.py
@@ -284,7 +284,7 @@ def do_cli(
             return
         global_config.set_accelerate_opt_in_stack(template_file, stack_name)
     else:
-        LOG.info(Colored().color_log(SYNC_INFO_TEXT, "yellow"), extra=dict(markup=True))
+        LOG.info(Colored().color_log(msg=SYNC_INFO_TEXT, color="yellow"), extra=dict(markup=True))
 
     s3_bucket_name = s3_bucket or manage_stack(profile=profile, region=region)
 
--- a/samcli/commands/sync/command.py
+++ b/samcli/commands/sync/command.py
@@ -284,7 +284,7 @@ def do_cli(
             return
         global_config.set_accelerate_opt_in_stack(template_file, stack_name)
     else:
-        LOG.info(Colored().yellow(SYNC_INFO_TEXT))
+        LOG.info(Colored().color_log(SYNC_INFO_TEXT, "yellow"), extra=dict(markup=True))
 
     s3_bucket_name = s3_bucket or manage_stack(profile=profile, region=region)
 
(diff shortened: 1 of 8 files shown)


samcli/commands/sync/command.py
```diff
@@ -284,7 +284,7 @@ def do_cli(
             return
         global_config.set_accelerate_opt_in_stack(template_file, stack_name)
     else:
-        LOG.info(Colored().yellow(SYNC_INFO_TEXT))
+        LOG.info(Colored().color_log(msg=SYNC_INFO_TEXT, color="yellow"), extra=dict(markup=True))
```
mildaniel, line 287: nit and not blocking: Let's think about creating some constants that might make it a little easier to keep consistency and allow us to define the colors to use for each scenario. Something like having

```
Colors.Success = "green"
Colors.Warning = "yellow"
Colors.Message = "whatever we decide"
...
```

mildaniel (approved):
Thanks Sriram, looks good. Left one minor comment.

samcli/lib/utils/colors.py
```diff
@@ -42,6 +48,9 @@ def __init__(self, colorize=True):
         colorize : bool
             Optional. Set this to True to turn on coloring. False will turn off coloring
         """
+        self.rich_logging = any(
```
mndeveci, line 51: Non blocking for this one, somewhat related to my feedback in the original pr, being able to find if rich is enabled throughout execution would be valuable in places like this.

samcli/lib/build/app_builder.py
```diff
@@ -641,7 +641,7 @@ def _build_function(  # pylint: disable=R1710
                     f"update to a newer supported runtime. For more information please check AWS Lambda Runtime "
                     f"Support Policy: https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html"
                 )
-                LOG.warning(self._colored.yellow(message))
+                LOG.warning(self._colored.color_log(msg=message, color="yellow"), extra=dict(markup=True))
```
mndeveci, line 644: Can we make this inside the colored instance? So that we don’t need to have string param as yellow and can use existing methods for colorful printing?

mndeveci: I would rather keep things in the utility methods (or in other ways) where we don't need to re-write same parameters (like `extra` param) for logging in different colors.

Maybe we should use `click` methods to write colorful output, and logging should be pure text.

mndeveci (approved):
Approving to unblock the things, but we might need to re-think about how the colorful outputs should be used and work in general.

PR #5108, opened 2023-05-03
chore: GHA for Auto Closing issues that were released

jfuss:
This GHA closes any issue that is marked "stage/waiting-for-release". This may close issues that are linked to PRs that were merged into develop but not yet released. There was no easy way to get all issues that are linked to a PR closed between two commits. Instead of doing long queries to get the exact issues, we will rely on the labels being updated and handle false positives as/if they happen.

None

Make releases less manual

See above commit comments

none

Example: https://github.com/jfuss/aws-sam-cli/issues/10

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/close-issue-on-release.yml
+++ b/.github/workflows/close-issue-on-release.yml
@@ -1,0 +1,21 @@
+name: Close issues on release cut
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  run-workflow:
+    permissions:
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+    - name: Close issues marked
+      env:
+        REPO : ${{ github.repository }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        release_url=$(gh release view ${{ github.ref_name }} --repo $REPO --json url --jq ".url")
+        for issue_number in $(gh issue list -l "stage/waiting-for-release" --repo $REPO --json number --jq ".[].number"); do
+          gh issue close $issue_number -c "Patch is released in [${{ github.ref_name }}]($release_url). Closing" -r completed --repo $REPO
+        done


.github/workflows/close-issue-on-release.yml
```diff
@@ -0,0 +1,21 @@
...
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+    - name: Close issues marked
+      env:
+        REPO : ${{ github.repository }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        release_url=$(gh release view ${{ github.ref_name }} --repo $REPO --json url --jq ".url")
+        for issue_number in $(gh issue list -l "stage/waiting-for-release" --repo $REPO --json number --jq ".[].number"); do
```
hawflau, line 19: One edge case to consider - if I merge a PR while a release is in progress, and label the corresponding issue with "stage/waiting-for-release", it seems this GHA will close the issue but the patch will actually be shipped in next release. 

While we can avoid merging PRs and labeling issues while a release is in progress, can/should this edge case be handled in this GHA?

super132 (approved):

hawflau (approved):

.github/workflows/close-issue-on-release.yml
jfuss: > While we can avoid merging PRs and labeling issues while a release is in progress, can/should this edge case be handled in this GHA?

I covered this in the commit message above. There is no easy way to get all the issues linked from a PR. So we would have to get all the commit messages, filter to get the pr number in the commit message, query for each pr, process the description/look for an issue, and then close that linked issue. 

I attempted this and become more overhead that I think was really needed. We might close some that are not meant to but I think that is generally ok. Being 100% accurate isn't super needed (imo). 

> Should we also have a GHA to mark the release number for issues with stage/waiting-for-release label once SAM CLI version bump PR is merged?

We could. Right now we add a message to the closed issues with the release version. My only worry about adding a label for the version is that will likely explode our label usage. Maybe there is a good reason we want to quickly find issues in a given version. Typically these are doing manually if we need to. Though I wonder if we should enforcing issues numbers being linked in the commit message 🤔

mndeveci (approved):

Issue #4670, opened 2023-02-08
Bug: Cloud Endpoints doesn't show API Gateway domain name


kaihendry:
### Description:

In terms of developer experience, if you don't use Route53, it's very cumbersome to see the endpoint to configure your DNS provider with! https://twitter.com/praneetaprakash/status/1621327393804914696

Currently you need to dig it out the AWS console:

<img src="https://s.natalian.org/2023-02-08/custom-domain-name-endpoint-configuration.png">

### Steps to reproduce:

https://github.com/kaihendry/helloworld update the DOMAINNAME and ACMCERTIFICATEARN in https://github.com/kaihendry/helloworld/blob/master/Makefile

### Observed result:

https://s.natalian.org/2023-02-08/sam-list-endpoints.txt

### Expected result:

`sam list endpoints` to show the actual **API Gateway domain name**, in my case d-908ajvn73h.execute-api.ap-southeast-1.amazonaws.com

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
# Paste the output of `sam --info` here
{
  "version": "1.72.0",
  "system": {
    "python": "3.7.10",
    "os": "Linux-6.1.9-arch1-2-x86_64-with-arch"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.23",
    "aws_cdk": "Not available",
    "terraform": "1.2.9"
  }
}
```

https://s.natalian.org/2023-02-08/sam-list-endpoints.txt

jfuss:
@kaihendry Thanks for the issue. I am still trying to understand our code as there seems to be logic targeting this use case: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/list/endpoints/endpoints_producer.py#L392. There is some interesting output from your command run too: `2023-02-08 15:11:48,032 | There is no customer defined id or cdk path defined for resource ApiGatewayDomainNameV260ef7d66a0, so we will use the resource logical id as the resource id`. So it looks like it got ignored "somewhere". I need to dig a little deeper and sync up with someone on the team that is more familiar here than I am.

[...]

PR #5117, opened 2023-05-05
feat: Added parameter overrides to sam list endpoints and resources

lucashuy:

https://github.com/aws/aws-sam-cli/issues/4670

Previously, parameters could not be overridden in the template for these sub commands, causing local transformation of the templates to not resolve the resources correctly. This caused some of the hashes for implicitly defined resources to be generated incorrectly, using the wrong values.

Adds parameter overrides to the `sam list endpoints` and `sam list resources` subcommands so that the template can be resolved correctly.

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel (commented):
Thanks Lucas, looks good. Can we add an integration test case too?

samcli/commands/list/endpoints/endpoints_context.py
```diff
@@ -18,7 +18,13 @@ class EndpointsContext(ListContext):
...
 
     def __init__(
-        self, stack_name: str, output: str, region: Optional[str], profile: Optional[str], template_file: Optional[str]
+        self,
+        stack_name: str,
+        output: str,
+        region: Optional[str],
+        profile: Optional[str],
+        template_file: Optional[str],
+        parameter_overrides: dict = {},
```
mndeveci, line 27: Python has a weird behaviour when it comes to default collections. Read for more info: https://rules.sonarsource.com/python/RSPEC-5717

So it is better to define the default as `None` here and during assignment, you can do some thing like;
```
self.parameter_overrides = parameter_overrides or {}
```

mndeveci (changes requested):

samcli/commands/list/endpoints/endpoints_context.py
lucashuy: TIL, will make that change

mildaniel (approved):

sriram-mv (approved):

PR #5120, opened 2023-05-07
feat: update terraform prepare hook to use linking fields values to link the applied resources.

moelasmar:

- Refactoring the linking algorithm in terraform prepare hook to move the CFN resources update to a call back function defined by each resource, and passed to the linking algorithm. Each callback function can map the destination resources and determine the correct intrinsic function to be used.
- Added a change to the linking algorithm to split the source resources based on the dependent resources (which depend on applied resources, and which depend on at least one non-applied resource). The applied resource will be linked using the linking fields approach, and the non applied resources will use the current liking approach (using the Terraform Config approach).

The linking fields approach in summary determine the value of the linking field in the source resource, and map it to all the destination resources that have linking field with the same value. 

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -118,14 +191,66 @@ def _handle_linking(self, source_tf_resource: TFResource, cfn_source_resources:
...
+        # build map between the destination linking field property values, and resources' logical ids
+        child_resources_linking_attributes_logical_id_mapping = {
+            self._resource_pair.destination_resource_tf.get(logical_id, {})
+            .get("values", {})
+            .get(self._resource_pair.tf_destination_attribute_name): logical_id
+            for logical_id in self._resource_pair.destination_resource_tf.keys()
+            if self._resource_pair.destination_resource_tf.get(logical_id, {})
+            .get("values", {})
+            .get(self._resource_pair.tf_destination_attribute_name)
+        }
```
mildaniel, line 232: Not completely following this. If we know that the reference is a constant value arn, won't 
```
self._resource_pair.destination_resource_tf.get(logical_id, {})
            .get("values", {})
            .get(self._resource_pair.tf_destination_attribute_name)
```
also yeild an arn? But we later create a `LogicalIdReference` if there is something in this map.

mildaniel (commented):
Thanks Mohamed for getting this across the finish line!

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: let me rewrite it using loops to be clear. I am creating a map between Layers `ARN` values (for example) and the Layers logical Ids, then use this map to search for the layers that has the same `ARN` values exist in the lambda functions.layers

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -31,9 +30,19 @@
 COMPILED_REGULAR_EXPRESSION = re.compile(r"\[[^\[\]]*\]")
 
 
-class LinkerIntrinsics(Enum):
-    Ref = "Ref"
-    GetAtt = "GetAtt"
+@dataclass
+class ReferenceType:
```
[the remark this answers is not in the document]
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -47,11 +56,11 @@ class ResourceLinkingPair:
...
     source_resource_tf_config: Dict[str, TFResource]
     destination_resource_tf: Dict[str, Dict]
-    intrinsic_type: LinkerIntrinsics
-    cfn_intrinsic_attribute: Optional[str]
-    source_link_field_name: str
+    tf_destination_attribute_name: str  # arn or id
     terraform_link_field_name: str
+    cfn_link_field_name: str
     terraform_resource_type_prefix: str
+    cfn_resource_update_call_back_function: Callable
```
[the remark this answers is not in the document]
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: Done

samcli/hook_packages/terraform/hooks/prepare/translate.py
```diff
@@ -343,13 +344,36 @@ def _translate_properties(
     return cfn_properties
 
 
+def _link_lambda_functions_to_layers_call_back(
+    function_cfn_resource: Dict, referenced_layers_logical_ids: List[ReferenceType]
```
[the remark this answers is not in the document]
moelasmar: done

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -94,6 +120,68 @@ def _handle_linking(self, source_tf_resource: TFResource, cfn_source_resources:
...
+            linking_field_value = cfn_resource.get("Properties", {}).get(self._resource_pair.cfn_link_field_name)
+
+            # if customer uses any non-applied resources to define the source resource property, Terraform will set
+            # the property value to unknown in the Terraform plan.
+            if linking_field_value is None:
+                non_applied_cfn_resources.append(cfn_resource)
+            else:
+                applied_cfn_resources.append(cfn_resource)
+
+        if applied_cfn_resources:
```
sriram-mv, line 155: do we need these ifs?

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -306,7 +361,10 @@ def _process_reference_resource_value(
                 )
 
             tf_dest_res_name = resolved_destination_resource.value[
-                len(self._resource_pair.terraform_resource_type_prefix) : -len(".arn")
+                len(self._resource_pair.terraform_resource_type_prefix) : -len(
```
sriram-mv, line 364: lets add a note on why, otherwise this looks like magic.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -94,6 +120,68 @@ def _handle_linking(self, source_tf_resource: TFResource, cfn_source_resources:
...
+            # if customer uses any non-applied resources to define the source resource property, Terraform will set
+            # the property value to unknown in the Terraform plan.
+            if linking_field_value is None:
+                non_applied_cfn_resources.append(cfn_resource)
+            else:
+                applied_cfn_resources.append(cfn_resource)
+
+        if applied_cfn_resources:
+            LOG.debug(
+                "Link resource configuration %s that has these applied instances %s using linking fields approach.",
```
sriram-mv, line 157: these logs appear to be the same for both cases.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: I think the part that's confusing me a little is this 
```
dest_resources = [
      LogicalIdReference(child_resources_linking_attributes_logical_id_mapping[value])
      if value in child_resources_linking_attributes_logical_id_mapping
      else ExistingResourceReference(value)
      for value in values
]
```
I guess we can have a list that is a combination of `LogicalIdReferences` and `ExistingReferences`, but I thought that if 

```
linking_attribute_value = destination_resource.get("values", {}).get(
      self._resource_pair.tf_destination_attribute_name
)
```
is not none it has to be an arn. So when we create `dest_resources` it will only create `LogicalIdReference`.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: copy/past error :D .. fixed

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: we need to determine which approach will be used based on the filter output .. so either we can do it here, or in each approach function.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: done

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
sriram-mv: if the lists are empty, we won't go through the loop. so we can remove those ifs.

mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mildaniel: Discussed offline. This value should in fact be a logical id reference even if we have a static value for better supporting `sam local` commands. If the value is a static arn, AWS SAM CLI will try to pull it instead of using the locally defined resource.

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
moelasmar: this can work for the applied approach, and I update that, but for the non applied, we do some other logic before we loop on the CFN resource to do the mapping, and I do not want to do some non required processing. I can move the other if condition to be done in the beginning of the `_link_using_terraform_config`

PR #5121, opened 2023-05-07
feat: add API Gateway Method, and Rest API resources linking

moelasmar:
Update the Terraform prepare hook to add the linking logic between the Gateway Method, and Rest API resources.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


moelasmar:
This PR is build on top of this [PR](https://github.com/aws/aws-sam-cli/pull/5120). Check this [link](https://github.com/moelasmar/aws-sam-cli/compare/linking_fields_for_applied_resources...moelasmar:aws-sam-cli:aws_api_gateway_method_and_aws_api_gateway_rest_api_linking) to get the differences.

samcli/hook_packages/terraform/hooks/prepare/types.py
```diff
@@ -129,12 +131,19 @@ def collect(self, properties: ResourceTranslationProperties):
...
 
+        In parent class we collect all the required properties. This logic can be customized in the children classes.
+
         Parameters
         ----------
         properties: ResourceTranslationProperties
             Properties acquired specific to an AWS resource when iterating through a Terraform module
         """
-        raise NotImplementedError
+        resolved_config_address = get_configuration_address(properties.resource_full_address)
```
mildaniel, line 141: Was just about to raise a PR doing the same thing, beat me to it 😄 

Should we also remove the implementations of `Collect` from Lambda function and Layer and let them inherit from the parent?

samcli/hook_packages/terraform/hooks/prepare/translate.py
```diff
@@ -369,11 +404,75 @@ def _link_lambda_functions_to_layers(
...
+    """
+    # if the destination rest api list contains more than one element, so we have an issue in our linking logic
+    if len(referenced_rest_apis_values) > 1:
+        raise InvalidResourceLinkingException("Could not link multiple Rest APIs to one Gateway method resource")
+
+    ref_list = [
+        {"Ref": logical_id.value} if isinstance(logical_id, LogicalIdReference) else logical_id.value
+        for logical_id in referenced_rest_apis_values
+    ]
+    gateway_method_cfn_resource["Properties"]["RestApiId"] = ref_list[0]
```
mildaniel, line 441: nit: maybe slightly more readable since we already know there is only one item in the list
```suggestion
    logical_id = referenced_rest_apis_values[0]
    gateway_method_cfn_resource["Properties"]["RestApiId"] = {"Ref": logical_id.value} if isinstance(logical_id, LogicalIdReference) else logical_id.value
```

mildaniel (approved):
Thanks Mohamed, minor, non-blocking comments

samcli/hook_packages/terraform/hooks/prepare/types.py
moelasmar: yes we can do that. We need it for Lambda function any way, as it will be destination resource in one of our mappings

samcli/hook_packages/terraform/hooks/prepare/translate.py
moelasmar: done

samcli/hook_packages/terraform/hooks/prepare/types.py
moelasmar: done

samcli/hook_packages/terraform/hooks/prepare/translate.py
```diff
@@ -403,6 +414,69 @@ def _link_lambda_functions_to_layers(
     ResourceLinker(resource_linking_pair).link_resources()
 
 
+def _link_gateway_methods_to_gateway_rest_apis_call_back(
```
mildaniel, line 417: Can we rename this function and the first argument to be a little more generic. This callback function is actually usable by any APIGW resource that links on the `RestApiId` field.

mildaniel (approved):

lucashuy (approved):

samcli/hook_packages/terraform/hooks/prepare/translate.py
moelasmar: done

PR #5125, opened 2023-05-08
fix: fix some issues in integration testing

moelasmar:
Resolve some integration testing issues. I added comments to the PR for more details.

### DIFF
--- a/tests/integration/local/invoke/test_invoke_terraform_applications.py
+++ b/tests/integration/local/invoke/test_invoke_terraform_applications.py
@@ -184,7 +184,7 @@ class TestInvokeTerraformApplicationWithoutBuild(InvokeTerraformApplicationInteg
             "You can also enable this beta feature with 'sam local invoke --beta-features'."
         )
         self.assertNotRegex(stdout.decode("utf-8"), terraform_beta_feature_prompted_text)
-        self.assertTrue(stderr.decode("utf-8").startswith(Colored().yellow(EXPERIMENTAL_WARNING)))
+        self.assertTrue(Colored().yellow(EXPERIMENTAL_WARNING) in stderr.decode("utf-8"))
 
         response = json.loads(stdout.decode("utf-8").split("\n")[-1])
         expected_response = json.loads('{"statusCode":200,"body":"{\\"message\\": \\"hello world\\"}"}')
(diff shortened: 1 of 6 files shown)


tests/integration/local/invoke/test_invoke_terraform_applications.py
```diff
@@ -184,7 +184,7 @@ def test_invoke_function_with_beta_feature_expect_warning_message(self):
             "You can also enable this beta feature with 'sam local invoke --beta-features'."
         )
         self.assertNotRegex(stdout.decode("utf-8"), terraform_beta_feature_prompted_text)
-        self.assertTrue(stderr.decode("utf-8").startswith(Colored().yellow(EXPERIMENTAL_WARNING)))
+        self.assertTrue(Colored().yellow(EXPERIMENTAL_WARNING) in stderr.decode("utf-8"))
```
moelasmar, line 187: We start printing some log statement in case if `samdev` is used, so the Warning message is not the first printed message any more.

https://github.com/aws/aws-sam-cli/blob/ad9a5986b434bab26159719a95ad33ba8f602d49/samcli/cli/main.py#L138-L146

tests/integration/local/start_api/test_start_api.py
```diff
@@ -34,8 +34,12 @@ class TestServiceHTTP10(StartApiIntegBaseClass):
 
     def setUp(self):
         self.url = "http://127.0.0.1:{}".format(self.port)
+        self.current_svn_str = HTTPConnection._http_vsn_str
         HTTPConnection._http_vsn_str = "HTTP/1.0"
```
moelasmar, line 38: This line cause that the S3 `PutObject` API call fail, and so some of the test cases were failing with the following error, if they got executed after the test cases defined in this class.
```
botocore.exceptions.ClientError: An error occurred (RequestTimeout) when calling the PutObject operation (reached max retries: 4): Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed.
```

tests/integration/testdata/sync/code/after/dotnet_function/src/HelloWorld/HelloWorld.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
moelasmar, line 4: we could not create new lambda function using DotNet 3.1 runtime

mndeveci (approved):

sriram-mv (approved):

Issue #5104, opened 2023-05-02
Bug: Method ARN sent to local authorizer and used for checking against authorized resource inconsistent with API Gateway


rhynix:
### Description:

When using a local authorizer, there are two issues with the method ARN used:

* The `methodArn` in the request to the authorizer lambda contains the Flask endpoint instead of the actual path of the request.
* The resources returned by the authorizer are compared to an ARN containing the same Flask endpoint instead of the actual path of the request.

This only leads to issues when the endpoint is parameterized.

### Steps to reproduce:

1. Use the code from this gist: [AWS SAM local method ARN inconsistency](https://gist.github.com/rhynix/27d6d1c163b9ae80f3ca315dfb513604)
2. Start the local API Gateway with `sam local start-api`
3. Fire the following requests:
   1. ```sh
      curl http://localhost:3000/projects/1 --header "Authorization: Bearer token"
      ```
   2. ```sh
      curl http://localhost:3000/projects/4 --header "Authorization: Bearer token"
      ```

### Observed result:

1. The responses to these two requests are respectively:
   1. ```json
      {"message":"User is not authorized to access this resource"}
      ```
   2. ```json
      {"message":"User is not authorized to access this resource"}
      ```
2. These requests lead to the following logs for the authorizer:
   1. ```methodArn: arn:aws:execute-api:us-east-1:123456789012:1234567890/dev/GET/projects/<id>```
   2. ```methodArn: arn:aws:execute-api:us-east-1:123456789012:1234567890/dev/GET/projects/<id>```

### Expected result:

This is the behavior when this template is deployed to AWS using `sam deploy`.

1. The responses to these two requests are respectively:
   1. ```json
      {"message":"ok"}
      ```
   2. ```json
      {"message":"User is not authorized to access this resource"}
      ```
2. These requests lead to the following logs for the authorizer:
   1. ```methodArn: arn:aws:execute-api:us-east-1:123456789012:1234567890/dev/GET/projects/1```
   2. ```methodArn: arn:aws:execute-api:us-east-1:123456789012:1234567890/dev/GET/projects/4```

### Additional environment details

```
{
  "version": "1.81.0",
  "system": {
    "python": "3.8.16",
    "os": "Linux-6.1.21_1-aarch64-with-glibc2.34"
  },
  "additional_dependencies": {
    "docker_engine": "23.0.3",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  }
}
```

rhynix:
[...]

PR #5135, opened 2023-05-08
Fix authorizer method ARN for parameterized endpoint

rhynix:
This patch fixes a wrong method ARN for endpoints that are parameterized. Instead of using the path for constructing the ARN, the endpoint would be used. The endpoint contains placeholder names of parameters instead of their values for the request.

This caused two issues:

* The method ARN sent to the authorizer would be wrong.
* The resources in the policy statement returned by the authorizer would be compared to the wrong ARN. This means authorization did not work properly in these cases.

#5104

The method ARN for the local authorizer is built in a way that is inconsistent with API Gateway. It uses the flask endpoint instead of the path.

Use the path of the flask request instead of the endpoint.

Implementations using this wrong behavior might break. This would break when deploying to API Gateway.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


moelasmar (approved):
LGTM .. Thanks @rhynix for your contribution.

rhynix:
The updated unit test is failing due to a deprecation warning from Werkzeug. It looks like it only happens when Werkzeug happens to resolve to version >=2.3.0. I believe there are three options we have here:

1. Pin Werkzeug to <2.3.0, which does not have the deprecation warning yet. It is now a transitive dependency which is not pinned.
2. Update Flask to >=2.2.4. This removes the use of the deprecated function.
3. Ignore the deprecation warning in pytest.ini.

My preference would go to option 1. Am I missing any options? Are there any things I should take into consideration here?

tests/unit/local/apigw/test_local_apigw_service.py
```diff
@@ -918,15 +917,20 @@ def test_lambda_auth_unauthorized_response(
...
 
-        result = self.api_service._request_handler()
+        flask_app = flask.Flask("test_flask_app")
+
+        with flask_app.test_request_context() as context:
+            context.request.method = "test"
+            context.request.path = "/test"
+
+            result = self.api_service._request_handler()
+
```
lucashuy, line 933: To address the deprecated call (the werkzeug call happens within the `flask_app.test_request_context()` like mentioned), we should be able to patch the call to `_create_method_arn` since we aren't concerned/testing the output of this method.

The reason why `werkzeug` is no longer pinned is to resolve a CVE. We could also upgrade Flask, however this may have other cascading side effects.

rhynix: Done.

lucashuy (approved):

lucashuy:
The changes to the ARN look good to me! Thanks for your contributions

PR #5139, opened 2023-05-09
feat: Link APIGW Resources to APIGW REST APIs

mildaniel:

N/A

To support `sam local start-api` for Terraform projects, AWS SAM CLI needs to be able to read the APIGW Terraform resources and then link related resources together. 

This PR adds the logic for linking all `AWS::ApiGateway::Resource` resources to their corresponding `AWS::ApiGateway::RestApi` resources linked on the `RestApiId` field.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/translate.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/translate.py
@@ -14,7 +14,9 @@ from samcli.hook_packages.terraform.hooks.prepare.constants import (
 from samcli.hook_packages.terraform.hooks.prepare.enrich import enrich_resources_and_generate_makefile
 from samcli.hook_packages.terraform.hooks.prepare.exceptions import (
     FunctionLayerLocalVariablesLinkingLimitationException,
+    OneGatewayResourceToRestApiLinkingLimitationException,
     OneLambdaLayerLinkingLimitationException,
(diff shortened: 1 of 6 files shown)


samcli/hook_packages/terraform/hooks/prepare/translate.py
lucashuy, line 1: Not anything specifically wrong about this in the case of this PR, but going forward (maybe in a different PR) is there anyway to reduce the amount of repeat code for linking two resources? We have it pretty modular right now which is great, but as we link more attributes together it looks like we might be left with a lot of modular bits that have roughly the same code

moelasmar (approved):
Thanks Daniel

lucashuy (approved):

samcli/hook_packages/terraform/hooks/prepare/translate.py
lucashuy: Yup, had the same idea for the snippet you linked where we could just iterate through the methods with the two constants. As for the actual linking methods themselves, we might be able to do something similar where we had some sort of list of attributes to link to properties (thinking in the scope of Lambda authorizers: RestApiId, the function URI).

Just looking ahead since for instance the method has 3 possible references to other resources (Lambda authorizer, integration, rest API) :smile:

PR #5152, opened 2023-05-10
fix: catch all ClientError from Boto

sriram-mv:
- regardless of error code in exception message - DO NOT Crash.

Per https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html we should be catching all `ClientError`

Screen Shot 2023-05-10 at 12 22 27 PM

* Catch all ClientError and transform them to click exceptions.
Screen Shot 2023-05-10 at 2 05 41 PM

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/exceptions.py
+++ b/samcli/commands/exceptions.py
@@ -68,9 +68,9 @@ class UnhandledException(click.ClickException):
         click.secho(msg, file=file, err=True, fg="yellow")
 
 
-class CredentialsError(UserException):
+class AWSServiceClientError(UserException):
     """
-    Exception class when credentials that have been passed are invalid.
+    Exception class when there are errors calling any AWS services via Boto.
     """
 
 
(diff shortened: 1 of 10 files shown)


samcli/commands/exceptions.py
```diff
@@ -68,9 +68,9 @@ def show(self, file: Optional[IO] = None) -> None:
         click.secho(msg, file=file, err=True, fg="yellow")
 
 
-class CredentialsError(UserException):
+class AWSServiceClientError(UserException):
```
jfuss, line 71: How does this end up looking in our telemetry system? I know we capture errors that are thrown. My concern in changing this to be generic is we loose the ability to understand what errors we are throwing. Capturing "CredentialError" vs "AWSServiceClientError" for missing/invalid creds is vastly different.

sriram-mv: I can rename it back to CredentialError for the sake of consistency, but ClientError covers a large suite of errors.

jfuss: This was more of a meta point. I am ok either way.

jfuss (approved):

moelasmar (approved):

lucashuy (approved):

PR #5153, opened 2023-05-10
feat: Add support for translating and linking gateway integrations

mildaniel:

N/A

To support sam local start-api for Terraform resources, we need to read in the APIGW methods and convert them to the corresponding CFN resources.

Adds a new "Internal" resource type for translating resources that exist in Terraform but not in CFN. These resources are stored temporarily to handle necessary linking and translation logic, but are not written to the metadata file.

Creates the first internal resource `Internal::ApiGateway::Method::Integration`

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/resources/internal.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/resources/internal.py
@@ -10,5 +10,6 @@ class InternalApiGatewayIntegrationProperties(ResourceProperties):
     Collect resource properties for translating and linking an aws_api_gateway_integration
     to the AWS::ApiGateway::Method resource
     """
+
     def __init__(self):
         super(InternalApiGatewayIntegrationProperties, self).__init__()
--- a/samcli/hook_packages/terraform/hooks/prepare/resources/internal.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/resources/internal.py
@@ -5,6 +5,7 @@ INTERNAL_PREFIX = "Internal::"
 INTERNAL_API_GATEWAY_INTEGRATION = f"{INTERNAL_PREFIX}ApiGateway::Method::Integration"
 
 
+# TODO(mladan): Add a mechanism for gating internal resources from being added to the metadata file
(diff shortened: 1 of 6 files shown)


samcli/hook_packages/terraform/hooks/prepare/resources/internal.py
```diff
@@ -0,0 +1,15 @@
+"""Module containing logic specific to internal resources handling during the prepare hook execution"""
+from samcli.hook_packages.terraform.hooks.prepare.types import ResourceProperties
+
+INTERNAL_PREFIX = "Internal::"
```
sriram-mv, line 4: is there a way where we can prevent accidental injection of this temporary resource into metadata?

mildaniel: Other than this condition https://github.com/aws/aws-sam-cli/blob/1d57965b88139076d686d8e804818876b4fe577c/samcli/hook_packages/terraform/hooks/prepare/translate.py#L225 I'm not sure what else we can do. Maybe a post-processing step to ensure we didn't write anything we weren't supposed to? Though even if we did write it to the file, I don't think it should break local behaviour.

moelasmar: I am not sure if we do validate the Template somewhere in SAM CLI while processing it or not, if we do so adding this internal type by mistake will cause issues.

Anyway, we only have one place now where we add the resources to Metadata file, and we guard it with this condition, we may face this issue if we start updating the Metadata file from different places. I think for now we can keep it simple.

sriram-mv: Agree not blocking. 

My concern is that this resource should never accidentally make it to the template itself as it may fail certain aspects? Maybe add a TODO note.

mildaniel: I added a TODO and also confirmed that if we accidentally do add an internal resources to the metadata file then the local test suite still works.

PR #5158, opened 2023-05-11
chore: Cleanup the linking functions

mildaniel:

N/A

Cleans up a bunch of repeated function calls with the same signature to be done by iterating over a loop of resources that need to be linked.

- Creates a dataclass for storing the source resource, destination resource and the function to link the two.
- Creates a list of the newly defined dataclass in a new module containing all of the existing supported links
- Updates the translate function to remove the repeated function calls to be handled in a separate function that iterates over the list and calls the linking function for each pair.
- Moves all relevant linking logic from the `translate.py` module to the `link_resources.py` module
- Updates unit tests accordingly

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar (approved):

sriram-mv (approved):
Overall looks good! Thanks for the refactor!

PR #5174, opened 2023-05-15
feat: Link API Gateway Methods to Integrations 

mildaniel:

N/A

To support `sam local start-api` for Terraform resources, we need to read in the Terraform API Gateway resources and convert them to their corresponding CFN resources.

Adds support for linking API Gateway methods to their corresponding API Gateway integrations. Additional handling needed to be done for these resources since they don't map 1:1 between Terraform and CFN. In Terraform, integrations are their own resources whereas in CFN, they are a field under the `AWS::ApiGateway::Method` resource.

1. Added support for mapping internal integration resources to the corresponding REST API and Gateway Resource resources on the `RestApiId` and `ResourceId` fields, respectively.
2. Linked all internal integration resources to the Lambda functions they are proxied to on the `Uri` field of the integration.
3. Added additional handling during the translation to iterate over all methods, find the corresponding integration and map the fields from the integration to the `ApiGatewayMethod.Properties.Integration` field.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/resources/apigw.py
```diff
@@ -11,6 +12,20 @@
     ResourceTranslationValidator,
     TFResource,
 )
+from samcli.lib.utils.resources import AWS_APIGATEWAY_METHOD
+
+LOG = logging.getLogger(__name__)
+
+INVOKE_ARN_FORMAT = (
+    "arn:aws:apigateway:${{AWS::Region}}:lambda:path/2015-03-31/functions/${{{function_logical_id}.Arn}}/invocations"
```
sriram-mv, line 20: aws can be also be replaced with aws partition.

```${{AWS::Partition}}```

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html

**Review by sriram-mv — APPROVED** [review]
Overall looks good to me. Added one thing to be changed.

File: samcli/hook_packages/terraform/hooks/prepare/resource_linking.py [hunk 2]
```diff
@@ -1239,7 +1247,149 @@ def _link_gateway_method_to_gateway_resource(
...
+    ----------
+    gateway_integrations_config_resources: Dict[str, TFResource]
+        Dictionary of configuration Gateway Integrations
+    gateway_integrations_config_address_cfn_resources_map: Dict[str, List]
+        Dictionary containing resolved configuration addresses matched up to the cfn Gateway Integration
+    lambda_function_terraform_resources: Dict[str, Dict]
+        Dictionary of all actual terraform Lambda function resources (not configuration resources).
+        The dictionary's key is the calculated logical id for each resource.
+    """
+
```
moelasmar, line 1379: can we filter the dictionary `gateway_integrations_config_resources` for only the resources that has `type` = `AWS_PROXY` , and if there is noting we can skip linking these resources.

moelasmar: My point is to avoid throwing limitation exceptions, and it is not required to do this linking.

samcli/hook_packages/terraform/hooks/prepare/resources/apigw.py
```diff
@@ -91,3 +106,105 @@ class ApiGatewayStageProperties(ResourceProperties):
...
+      Type: AWS::ApiGateway::Method
+      Properties:
+        Integration:
+          Uri: Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${function.Arn}/invocations
+          Type: AWS_PROXY
+
+    Parameters
+    ----------
+    cfn_dict: dict
+        Resultant CFN dict that will be mutated to append integrations to each API Gateway method
```
moelasmar, line 130: we can send only the method CFN resources. It should be part of the ApiGateway method resources collected [here](https://github.com/aws/aws-sam-cli/blob/a070af6cb944b55fbd8f8da53846c54040fcb36e/samcli/hook_packages/terraform/hooks/prepare/resources/apigw.py#L69)

samcli/hook_packages/terraform/hooks/prepare/resources/apigw.py
```diff
@@ -91,3 +106,105 @@ class ApiGatewayStageProperties(ResourceProperties):
...
+        Dict containing the resource properties that can be used to uniquely identify and API Gateway integration
+
+    Returns
+    -------
+        Returns a set comprised of unique identifiers of an API Gateway integration
+
+    """
+    return {
+        resource_properties.get("RestApiId", {}).get("Ref", ""),
+        resource_properties.get("ResourceId", {}).get("Ref", ""),
```
moelasmar, line 189: we have a case that these values can be strings if customer used static values for Apigateway that is defined outside the TF project.

moelasmar (approved):
Thanks Daniel

PR #5178, opened 2023-05-16
Thread id bug

Leo10Gama:
N/A

Previously, Python's `threading.thread_id()` method did not provide an accurate representation of distinct Events running on the same thread. This change allows for a more accurate look at which methods are running on the same thread, and helps group long Events more accurately.

The previous `Event.thread_id` was changed from an `int` to a `UUID` to act as a better identifier for Events run on the same thread.

The `thread_id` will now be a UUID, and its representation in JSON and `__repr__` formats will be its hex value, as a string.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/telemetry/event.py
```diff
@@ -144,7 +150,11 @@ class EventTracker:
 
     @staticmethod
     def track_event(
-        event_name: str, event_value: str, session_id: Optional[str] = None, exception_name: Optional[str] = None
+        event_name: str,
+        event_value: str,
+        session_id: Optional[str] = None,
+        thread_id: Optional[UUID] = None,
```
hawflau, line 156: does putting `thread_id` before `exception_name` as positional argument affect any existing `track_event` call?

hawflau (commented):
overall LGTM. Just one question - can you also check if some change is needed for the `track_event` calls in samcli/lib/sync/infra_sync_executor.py?

samcli/lib/telemetry/event.py
Leo10Gama: Nope it does not, just verified in the code base. The only call right now that makes use of the `session_id` property is the call in `local_apigw_service.py` on line 701, which calls it by keyword.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -174,8 +175,9 @@ def execute_infra_sync(self, first_sync: bool = False) -> InfraSyncResult:
             days_since_last_infra_sync = (current_time - last_infra_sync_time).days
 
         # Will not combine the comparisons in order to save operation cost
+        thread_id = uuid4()
         if self._sync_context.skip_deploy_sync and first_sync and (days_since_last_infra_sync <= AUTO_INFRA_SYNC_DAYS):
-            EventTracker.track_event("SyncFlowStart", "SkipInfraSyncExecute")
+            EventTracker.track_event("SyncFlowStart", "SkipInfraSyncExecute", thread_id=thread_id)
```
mndeveci, line 180: I think you missed to add `thread_id` to line 191

mndeveci (changes requested):
LGTM in general, left some comments there!

samcli/lib/telemetry/event.py
```diff
@@ -83,14 +84,19 @@ class Event:
...
 
-    def __init__(self, event_name: str, event_value: str, exception_name: Optional[str] = None):
+    def __init__(
+        self, event_name: str, event_value: str, thread_id: Optional[UUID] = None, exception_name: Optional[str] = None
+    ):
         Event._verify_event(event_name, event_value)
         self.event_name = EventName(event_name)
         self.event_value = event_value
+        if not thread_id:
+            thread_id = uuid4()
```
mndeveci, line 98: Should we still use `threading.get_ident()` for default value or other fields will same for that purpose?
@hawflau what do you think?

samcli/lib/sync/infra_sync_executor.py
Leo10Gama: Yup I missed that, good catch

samcli/lib/telemetry/event.py
hawflau: I think `uuid4()` is good. From what I've seen empirically, `session_id` and `threading.get_ident()` are producing the same grouping effects.

mndeveci (approved):

hawflau (approved):

PR #5115, opened 2023-05-05
feat: add remote invoke factories

mndeveci:
#### Which issue(s) does this change fix?
https://github.com/aws/aws-sam-cli/discussions/5018

This PR adds some library implementations for Remote Invoke feature which contains;
- `RemoteInvokeExecutorFactory` implementation which will then be used to instantiate `RemoteInvokeExecutor` instance depending on the resource type. It is mapping is empty now, as implementations will be added later.
- `RemoteInvokeExecutionInfo` object which holds request and response information as well as any exception that is been raised during the execution.
- Abstract `RemoteInvokeRequestResponseMapper` definition, which then be used to implement different mappers for different remote invoke use case.
  - `ResponseObjectToJsonStringMapper` implements mapping an object into JSON string.
- Abstract `BotoActionExecutor` definition, which then be extended by different remote invoke instances, by implementing `_execute_action` and `_execute_action_file` methods.
- `RemoteInvokeExecutor` definition which is a container of holding request mappers, response mappers and `BotoActionExecutor` instance, which can be used from remote invoke context later.

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [ ] Write/update integration tests
- [ ] Write/update functional tests if needed
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #5179, opened 2023-05-16
feat: Add lambda invoke boto API setup for remote invoke

hnnasit:

This PR adds the lambda invoke boto API setup to the Executor factory that was created in [PR](https://github.com/aws/aws-sam-cli/pull/5115) which contains the following:

- `LambdaInvokeExecutor` class which contains functions for validating boto parameters and executes the lambda invoke API. The supported boto parameters are `InvocationType`, `LogType`, and `Qualifier`.
- New method in the remote_invoke_executor_factory to create `RemoteInvokeExecutor` for lambda with the above boto_executor.
- Adds input mapper called DefaultConvertToJSON which is used to convert input payloads into JSON string before calling the invoke API.
- Adds 2 output mappers specific to Lambda. `LambdaResponseConverter` decodes the response Payload which is returned as a Stream and `LambdaResponseOutputFormatter` formats the response output which will be printed on the screen. The `output_format` variable takes 2 values `default` and `original-boto-response`

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,143 @@
...
+    def validate_action_parameters(self, parameters: dict):
+        """
+        Validates the input boto parameters and prepares the parameters for calling the API.
+
+        :param parameters: Boto parameters provided as input
+        """
+        for parameter_key, parameter_value in parameters.items():
+            parameter_allowed_values = self._SUPPORTED_PARAMETERS.get(parameter_key, [])
+            if parameter_key == "FunctionName":
+                LOG.info("FunctionName is defined using the value provided for --resource-id option.")
```
mndeveci, line 50: Should we fail for these cases or print a warning?

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,143 @@
...
+            elif parameter_key == "Payload":
+                LOG.info("Payload is defined using the value provided for either --payload or --payload-file options.")
+            elif parameter_key not in self._SUPPORTED_PARAMETERS:
```
mndeveci, line 53: I am not sure if we want to have supported parameters here. Can't we just pass these parameters to the boto3 calls, and let boto3 to handle if it is a valid parameter or not?

samcli/lib/remote_invoke/exceptions.py
```diff
@@ -0,0 +1,11 @@
+"""
+Exceptions used by remote invoke executors
+"""
+
+
+class InvalidResourceBotoParameterException(Exception):
```
mndeveci, line 6: If this is catch in some other place then we can extend it from `Exception`. But if this will be thrown back to user, then it should extend from `UserException` which is handled by click and execution is ended gracefully (without printing stack trace)

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,143 @@
...
+    be written to stdout.
+    """
+
+    def map(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> RemoteInvokeExecutionInfo:
+        """
+        Maps the lambda response output to the type of output format specified by customer.
+        If output_format is original-boto-response, write the original boto repsonse
+        to stdout.
+        """
+        if remote_invoke_input.output_format == "default":
```
mndeveci, line 130: Should these output formats be an enum so that we don't need to copy/paste string constants in different places?

samcli/lib/remote_invoke/lambda_invoke_executors.py
hnnasit: We can print a warning and continue the execution since these parameters would be provided using resource_id and payload/payload-file options as input.

samcli/lib/remote_invoke/exceptions.py
```diff
@@ -0,0 +1,12 @@
+"""
+Exceptions used by remote invoke executors
+"""
+from samcli.commands.exceptions import UserException
+
+
+class InvalidResourceBotoParameterException(UserException):
```
mndeveci, line 7: Apologies for confusion, but any exception that we are going to use in `lib` module should not be extended from `click` library (`UserException` does extend from `click.Exception`). For that reason, we can revert this change and handle this exception seperately once we implement the command or its context.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,144 @@
...
+        self.request_parameters = {"InvocationType": "RequestResponse", "LogType": "Tail"}
+
+    def validate_action_parameters(self, parameters: dict) -> None:
+        """
+        Validates the input boto parameters and prepares the parameters for calling the API.
+
+        :param parameters: Boto parameters provided as input
+        """
+        for parameter_key, parameter_value in parameters.items():
+            if parameter_key == "FunctionName":
```
mndeveci, line 46: Let's move these string constants (`FunctionName` and `Payload`) at the top as they were used in multiple places.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            response = self._lambda_client.invoke(**self.request_parameters)
+        except ParamValidationError as param_val_ex:
+            raise InvalidResourceBotoParameterException(
+                f"Invalid parameter key provided."
+                f" {str(param_val_ex).replace('FunctionName, ', '').replace('Payload, ', '')}"
+            ) from param_val_ex
+        except ClientError as client_ex:
+            validation_err_str = "ValidationException"
+            if boto_utils.get_client_error_code(client_ex) == validation_err_str:
```
mndeveci, line 73: I would suggest to move `ValidationException` as a constant at the top, or use the string directly in the line below.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class LambdaResponseConverter(RemoteInvokeRequestResponseMapper):
+    """
+    This class helps to convert response from lambda service. Normally lambda service
+    returns 'Payload' field as stream, this class converts that stream into string object
+    """
+
+    def map(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> RemoteInvokeExecutionInfo:
+        assert isinstance(remote_invoke_input.response, dict), "Invalid response type received, expecting dict"
```
mndeveci, line 110: Not sure how `assert` keyword will behave here, can we instead verify with something else and raise an exception?

samcli/lib/remote_invoke/remote_invoke_executors.py
```diff
@@ -25,15 +35,27 @@ class RemoteInvokeExecutionInfo:
...
+        payload_file: Optional[TextIOWrapper],
+        parameters: dict,
+        output_format: RemoteInvokeOutputFormat,
+    ):
         self.payload = payload
         self.payload_file = payload_file
+        self.parameters = parameters
+        self.output_format = output_format
         self.response = None
+        self.stderr_str = None
```
mndeveci, line 58: Let's rename `stderr_str` to something else (like `log_output`)

samcli/lib/remote_invoke/remote_invoke_executors.py
```diff
@@ -177,10 +207,12 @@ def __init__(
     def execute(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> RemoteInvokeExecutionInfo:
         """
         First runs all mappers for request object to get the final version of it.
-        Then invokes the BotoActionExecutor to get the result
+        Then validates all the input boto parameters and invokes the BotoActionExecutor to get the result
         And finally, runs all mappers for the response object to get the final form of it.
         """
         remote_invoke_input = self._map_input(remote_invoke_input)
+        remote_invoke_output = None
```
mndeveci, line 214: Just curious, do we need to define `remote_invoke_output` as `None` here, can't we just let line 216 to set it?

samcli/lib/remote_invoke/exceptions.py
hnnasit: Oh no that makes sense, I forgot to change it back after our offline discussion yesterday. Updated in the new commit.

samcli/lib/remote_invoke/lambda_invoke_executors.py
hnnasit: Addressed in the new commit.

samcli/lib/remote_invoke/remote_invoke_executors.py
hnnasit: Addressed in the new commit.

samcli/lib/remote_invoke/lambda_invoke_executors.py
hnnasit: Addressed in the new commit.

samcli/lib/remote_invoke/remote_invoke_executors.py
hnnasit: Nope it doesn't need to be defined. Addressed in the new commit.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,147 @@
+"""
+Remote invoke executor implementation for Lambda
+"""
+import base64
+import json
+import logging
+from json import JSONDecodeError
+from typing import Any, Dict, cast
+
+from botocore.exceptions import ClientError, ParamValidationError
```
sriram-mv, line 10: One thing about any boto imports is that they add latency, so be mindful about that as its a large lib.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,147 @@
...
+        self.request_parameters = {"InvocationType": "RequestResponse", "LogType": "Tail"}
+
+    def validate_action_parameters(self, parameters: dict) -> None:
+        """
+        Validates the input boto parameters and prepares the parameters for calling the API.
+
+        :param parameters: Boto parameters provided as input
+        """
+        for parameter_key, parameter_value in parameters.items():
+            if parameter_key == FUNCTION_NAME:
```
sriram-mv, line 48: can this ever expand apart from the two terms? we don't want this to become multiple if, elifs.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -0,0 +1,147 @@
...
+            raise InvalidResourceBotoParameterException(
+                f"Invalid parameter key provided."
+                f" {str(param_val_ex).replace('{FUNCTION_NAME}, ', '').replace('{PAYLOAD}, ', '')}"
+            ) from param_val_ex
+        except ClientError as client_ex:
+            if boto_utils.get_client_error_code(client_ex) == "ValidationException":
+                raise InvalidResourceBotoParameterException(
+                    f"Invalid parameter value provided. {str(client_ex).replace('(ValidationException) ', '')}"
+                ) from client_ex
+            raise client_ex
```
sriram-mv, line 78: I would urge that we re-catch and throw all exceptions as `click` exceptions. Maybe we are already doing that, but leaving it as a note in any case. :)

hnnasit: Yup thats the plan. These exceptions would be handled in the context.py or command.py files.

samcli/lib/remote_invoke/lambda_invoke_executors.py
hnnasit: Other than those 2 parameters, the additional parameters are directly passed to the boto API without any changes. If we do decide to process any other parameter, it would be added in here. For instance, the `ClientContext` parameter is required to be base64-encoded string. If we want to encode this before passing to the API, that logic would be added here.

samcli/lib/remote_invoke/exceptions.py
```diff
@@ -1,10 +1,17 @@
...
-from samcli.commands.exceptions import UserException
 
 
-class InvalidResourceBotoParameterException(UserException):
+class InvalidResourceBotoParameterException(Exception):
+    msg: str
+
+    def __init__(self, msg: str) -> None:
+        self.msg = msg
+        super().__init__(self.msg)
```
mndeveci, line 11: nit: I think we don't need to have this part if you just want to keep `msg` field there.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -107,12 +108,14 @@ class LambdaResponseConverter(RemoteInvokeRequestResponseMapper):
...
-            remote_invoke_input.response["Payload"] = cast(StreamingBody, payload_field).read().decode("utf-8")
+        if not isinstance(remote_invoke_input.response, dict):
+            raise InvalideBotoResponseException("Invalid response type received from Lambda service, expecting dict")
+        else:
+            payload_field = remote_invoke_input.response.get(PAYLOAD)
+            if payload_field:
+                remote_invoke_input.response[PAYLOAD] = cast(StreamingBody, payload_field).read().decode("utf-8")
 
-        return remote_invoke_input
+            return remote_invoke_input
```
mndeveci: nit: for readability, no need to have this `else` block here (just to reduce the indentation), something like;
```
        if not isinstance(remote_invoke_input.response, dict):
            raise InvalideBotoResponseException("Invalid response type received from Lambda service, expecting dict")
        
        payload_field = remote_invoke_input.response.get(PAYLOAD)
        if payload_field:
            remote_invoke_input.response[PAYLOAD] = cast(StreamingBody, payload_field).read().decode("utf-8")

        return remote_invoke_input
```

mndeveci (approved):
Thanks for addressing all previous comments, approved with some additional minor comments.

samcli/lib/remote_invoke/remote_invoke_executors.py
```diff
@@ -4,13 +4,23 @@
...
 
 LOG = logging.getLogger(__name__)
 
 
+class RemoteInvokeOutputFormat(Enum):
+    """
+    Types of output formats used to by remote invoke
+    """
+
+    DEFAULT = "default"
```
sriram-mv, line 20: This feels like it could be more pythonic.

sriram-mv (approved):

samcli/lib/remote_invoke/remote_invoke_executors.py
hnnasit: Discussed offline this is not a blocker. Keeping this enum consistent with the other enums we have where the string is assigned directly to the constant.

PR #5181, opened 2023-05-16
feat: Terraform apply message to populate unresolved values

lucashuy:

N/A

There exists a current limitation in the Terraform implementation where some values in the planned file do not exist, or are not populated, until the project is applied.

Adds a warning message after the project is planned and again when a limitation is reached to let the customer know that they can apply their project.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar (commented):
can we update the [unsupported integration test cases](https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/buildcmd/test_build_terraform_applications_other_cases.py#L253) to do a check for this warning message?

moelasmar (approved):

samcli/hook_packages/terraform/hooks/prepare/exceptions.py
```diff
@@ -54,7 +64,7 @@ def __init__(self, local_variable_reference, dest_resource_list):
             issue_link=LOCAL_VARIABLES_SUPPORT_ISSUE_LINK,
             line_sep=os.linesep,
         )
-        UserException.__init__(self, msg)
+        ApplyLimitationException.__init__(self, msg)
```
hawflau, line 67: just curious and non-blocking - why use `ApplyLimitationException.__init__(self, msg)` instead of `super().__init__(self, msg)`?

lucashuy: No particular reason, the rest of the exceptions in this file reference the class instead of using `super()`, so instead of changing everything over to `super`, I just kept the existing behaviour. We could go either way since we are only inheriting from one parent class.

samcli/hook_packages/terraform/hooks/prepare/translate.py
```diff
@@ -67,6 +68,74 @@
...
+        A tuple of the current module's planned values and TFModule representation of configuration values
+    """
+    queue = [(root_module, root_tf_module)]
+
+    while queue:
+        modules = queue.pop(0)
+
+        yield modules
+
+        _add_child_modules_to_queue(modules[0], modules[1], queue)
```
hawflau, line 94: nit: this also works:
```
_add_child_modules_to_queue(*modules, queue)
```

hawflau (approved):
LGTM

PR #5182, opened 2023-05-17
fix: Relaxed OpenApi authorizer definition

lucashuy:

N/A

Request based authorizers can have no identity sources defined if caching is disabled. We don't check for caching, but we can make the change to allow empty identity sources.

Additionally, the root level authorizer check was incorrectly written to not apply to Rest Apis, however this is not the case.

Relaxes the check only for request based authorizers to allow empty identity sources. Also updates the validation to accept root default authorizers for Rest Apis when using the 3.x version specification.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/lib/swagger/parser.py
```diff
@@ -152,9 +152,9 @@ def get_authorizers(self, event_type: str = Route.API) -> Dict[str, Authorizer]:
                         auth_name,
                     )
 
-            if not identity_sources:
+            if not identity_sources and authorizer_type == LambdaAuthorizer.TOKEN:
```
moelasmar, line 155: can you add a comment to explain this condition

moelasmar (approved):
Thanks Lucas, it looks good to me. Just left a small comment.

hawflau (approved):
LGTM

PR #5183, opened 2023-05-17
feat: remote invoke context initial impl

mndeveci:
Adds `RemoteInvokeContext` class implementation with following validations;

- If no `stack_name` or `resource_id` is provided, it fails with `InvalidRemoteInvokeParameters`
- If only `stack_name` is provided;
  - If there is only one resource with supported type, continues execution
  - If there is more than one resource with supported type, fails with `AmbiguousResourceForRemoteInvoke`
  - If there is no resource definition with supported type, fails with `NoResourceFoundForRemoteInvoke`
- If only `resource_id` is provided;
  - If `resource_id` is a valid ARN from supported resource, continues execution
  - If `resource_id` is a valid `physical_resource_id` in a deployed stack, continues execution
  - If `resource_id` is an invalid `physical_resource_id` (where it can't be find by CFN API call), fails with `AmbiguousResourceForRemoteInvoke`

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+            raise AmbiguousResourceForRemoteInvoke(
+                f"{self._stack_name} contains more than one resource that could be used with remote invoke, "
+                f"please provide --resource-id to resolve ambiguity."
+            )
+
+        raise NoResourceFoundForRemoteInvoke(
+            f"{self._stack_name} stack has no resources that can be used with remote invoke."
+        )
+
+    def _get_from_physical_resource_id(self) -> CloudFormationResourceSummary:
```
hnnasit, line 95: nit: docstring for method.

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+        self._populate_resource_summary()
+        return self
+
+    def __exit__(self, *args) -> None:
+        pass
+
+    def run(self):
+        pass
+
+    def _populate_resource_summary(self) -> None:
```
hnnasit, line 59: nit: docstring for method

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+            for logical_id, resource_summary in resource_summaries.items():
+                LOG.debug("Using %s resource for remote invocation (%s)", logical_id, resource_summary)
+                return resource_summary
+        elif len(resource_summaries) > 1:
+            raise AmbiguousResourceForRemoteInvoke(
+                f"{self._stack_name} contains more than one resource that could be used with remote invoke, "
+                f"please provide --resource-id to resolve ambiguity."
+            )
+
+        raise NoResourceFoundForRemoteInvoke(
```
sriram-mv, line 91: nit: should we do a `if not resource_summaries` here first.

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+import logging
+from typing import Optional, cast
+
+from samcli.commands.remote_invoke.exceptions import (
+    AmbiguousResourceForRemoteInvoke,
+    InvalidRemoteInvokeParameters,
+    NoResourceFoundForRemoteInvoke,
+    UnsupportedServiceForRemoteInvoke,
+)
+from samcli.lib.pipeline.bootstrap.resource import ARNParts
```
jfuss, line 13: Should this be refactored to a more common place?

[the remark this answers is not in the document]
mndeveci: You mean moving `ARNParts` into a common place rather than directly using from `samcli.lib.pipeline.bootstrap.resource`?

samcli/commands/remote_invoke/remote_invoke_context.py
mndeveci: Since we already exhausted all other options above, I was just raising the exception here. If you think that would be useful, I can add a comment on top of it to state this is the case where we don't have any resource summaries at all.

jfuss (changes requested):

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+SUPPORTED_SERVICES = {"lambda": AWS_LAMBDA_FUNCTION}
+
+
+class RemoteInvokeContext:
+
+    _boto_client_provider: BotoProviderType
+    _boto_resource_provider: BotoProviderType
+    _stack_name: Optional[str]
+    _resource_id: Optional[str]
+    _resource_summary: Optional[CloudFormationResourceSummary]
```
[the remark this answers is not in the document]
mndeveci: This makes it more explicit about all the fields and their types that will/may be instantiated in this class. Mypy doesn't need to implicitly find their types from assignments. I know it is more verbose but I personally find this way more readable :)

samcli/commands/remote_invoke/remote_invoke_context.py
mndeveci: That is clear now :) Will move this class into common place

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __exit__(self, *args) -> None:
+        pass
+
+    def run(self):
+        pass
+
+    def _populate_resource_summary(self) -> None:
+        if not self._stack_name and not self._resource_id:
+            raise InvalidRemoteInvokeParameters("Either --stack-name or --resource-id parameter should be provided")
```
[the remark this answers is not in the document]
mndeveci: 🤔 agreed on this one, this validation can be moved up to `command.py` through `click`. @hnnasit what do you think?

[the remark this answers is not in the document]
hnnasit: I think that makes sense. We will also have another validation for `payload` and `payload-file` in command.py. We can move the above validation to cli as well.

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -0,0 +1,150 @@
...
+            self._resource_summary = self._get_from_physical_resource_id()
+
+    def _get_single_resource_from_stack(self) -> CloudFormationResourceSummary:
+        """
+        Queries all resources from stack with its type,
+        and returns its information if stack has only one resource from that type (including nested stacks)
+        """
+        LOG.debug(
+            "Trying to get single resource with %s type in % stack since no resource id is provided",
+            AWS_LAMBDA_FUNCTION,
```
hnnasit, line 90: Should we change this to `SUPPORTED_SERVICES.values()` as well? When we add support for more services, we can have a priority of which resource gets invoked when there are single resources of different services.

jfuss (approved):

hnnasit (approved):

samcli/commands/remote_invoke/remote_invoke_context.py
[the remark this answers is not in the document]
mndeveci: Regarding to this one, I feel weird when removing this validation since typing system can't guarantee one or the other field should have a value. After discussing with @hnnasit we decided to keep it here as well as adding it to the `command.py` implementation later.

PR #5208, opened 2023-05-26
feat: Abstract SamConfig

Leo10Gama:
This change decouples the logic of reading/writing TOML documents from the `SamConfig` class, allowing for the future implementation of additional `samconfig` file types. It also allows the class to better follow the Single Responsibility Principle, as it handles the logic of storing and representing the config data, rather than also implementing the reading and writing.

The `SamConfig` has been changed such that `document` is now no longer a TOML document, but a dictionary object, making it easier to interface with. In addition, two new classes, the abstract `FileManager` and concrete child `TomlFileManager` have been created to handle the logic of reading and writing to TOML files.

The `SamConfig` class now more accurately represents and encapsulates solely the configuration options, and any reading and writing to TOML files has now been generalized to a designated `FileManager`, assigned based on the provided filename, and set to be TOML by default to ensure backwards compatibility.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/config/samconfig.py
```diff
@@ -39,11 +84,15 @@
             Optional. Name of the configuration file. It is recommended to stick with default so in the future we
             could automatically support auto-resolving multiple config files within same directory.
         """
+        self.document = {}
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
+        self.file_manager = self.FILE_MANAGER_MAPPER.get(self.filepath.suffix[1:], TomlFileManager)  # default to TOML
+        if not self.file_manager:
+            LOG.warning(f"The config file extension '{self.filepath.suffix[1:]}' is not supported. TOML will be used.")
```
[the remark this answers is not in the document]
lucashuy: Not an issue, this returns a suffix of a file if there is one in the user's terminal.

samcli/lib/config/samconfig.py
```diff
@@ -39,11 +40,16 @@ def __init__(self, config_dir, filename=None):
             Optional. Name of the configuration file. It is recommended to stick with default so in the future we
             could automatically support auto-resolving multiple config files within same directory.
         """
+        self.document = {}
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
+        self.file_manager = self.FILE_MANAGER_MAPPER.get(self.filepath.suffix[1:], None)  # default to TOML
+        if not self.file_manager:
+            LOG.warning(f"The config file extension '{self.filepath.suffix[1:]}' is not supported. TOML will be used.")
```
lucashuy, line 47: Open question: should we explicitly require the file format instead of assuming TOML by default?

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,99 @@
...
+        -------
+        dict
+            A Python dictionary representation of the contents of the TOML file at the provided location.
+        """
+        document: dict = {}
+        try:
+            txt = filepath.read_text()
+            document = dict(tomlkit.loads(txt))
+        except OSError:
+            document = {}
```
lucashuy, line 78: Can we either add a debug log message here when we catch this `OSError` so that we have some trace of this happening when debugging issues?

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Parameters
+        ----------
+        document: dict
+            The object to write.
+        filepath: Path
+            The final location for the TOML file to be written.
+        """
+        if not document:
+            return
```
lucashuy, line 97: Same comment about logging message as above.

samcli/lib/config/samconfig.py
lucashuy: I personally think this should be changed to explicitly require a file extension. If I was to pass in a file format (such as `myconfig.xml`) that isn't supported, it'll try to parse it as TOML. The first part of that message is clear in reasoning/intent (not supported), but then from a UX stand point, it wouldn't quite make sense to try to parse it as TOML anyways.

Leo10Gama: Yup, as per our conversation on Friday, being explicit is probably better in this situation. I'll have it implemented in the next commit 👍

samcli/lib/config/samconfig.py
```diff
@@ -39,11 +40,18 @@ def __init__(self, config_dir, filename=None):
             Optional. Name of the configuration file. It is recommended to stick with default so in the future we
             could automatically support auto-resolving multiple config files within same directory.
         """
+        self.document = {}
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
+        self.file_manager = self.FILE_MANAGER_MAPPER.get(self.filepath.suffix[1:], None)  # default to TOML
+        if not self.file_manager:
+            LOG.warning(f"The config file extension '{self.filepath.suffix[1:]}' is not supported.")
+            raise SamConfigFileReadException(
+                f"The config file {self.filepath} uses an unsupported extension, and cannot be read."
```
hnnasit, line 49: Should we also list the supported extensions for users to see what is supported?

Leo10Gama: We could, though if the number of supported formats grows, it could be a large list. But for now with 3 (4 if you want to count .yml and .yaml as separate) supported formats, it would probably be a good idea yes. I'll add that in 👍

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,104 @@
...
+            A Python dictionary representation of the contents of the TOML file at the provided location.
+        """
+        document: dict = {}
+        try:
+            txt = filepath.read_text()
+            document = dict(tomlkit.loads(txt))
+        except OSError as e:
+            LOG.debug(f"OSError occurred while reading TOML file: {str(e)}")
+            document = {}
+        except tomlkit.exceptions.TOMLKitError as e:
```
mildaniel, line 83: How come for some exceptions we return an empty dict, and for others we raise an exception?

samcli/lib/config/samconfig.py
```diff
@@ -144,30 +149,29 @@ def put_comment(self, comment):
...
 
         Raises
         ------
-        tomlkit.exceptions.TOMLKitError
+        SamConfigFileReadException
             If the data is invalid
 
         """
         self._write()
+        print(self.document)
```
hnnasit, line 166: remove any print statements

samcli/lib/config/samconfig.py
```diff
@@ -70,22 +81,21 @@ def get_all(self, cmd_names, section, env=DEFAULT_ENV):
...
-        if isinstance(self.document, dict):
-            toml_content = self.document.get(env, {})
-            params = toml_content.get(self._to_key(cmd_names), {}).get(section, {})
-            if DEFAULT_GLOBAL_CMDNAME in toml_content:
-                global_params = toml_content.get(DEFAULT_GLOBAL_CMDNAME, {}).get(section, {})
-                global_params.update(params.copy())
-                params = global_params.copy()
-            return params
-        return {}
+        self.document = self.file_manager.read(self.filepath)
```
hnnasit, line 90: Any reason why `self._read()` is not called here directly?

samcli/lib/config/samconfig.py
```diff
@@ -39,11 +40,18 @@ def __init__(self, config_dir, filename=None):
             Optional. Name of the configuration file. It is recommended to stick with default so in the future we
             could automatically support auto-resolving multiple config files within same directory.
         """
+        self.document = {}
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
+        self.file_manager = self.FILE_MANAGER_MAPPER.get(self.filepath.suffix[1:], None)  # default to TOML
```
mildaniel, line 45: I agree that if there are multiple configs and one isn't specified we should default to TOML. But if a customer only has one configuration file that isn't the default, should we just use that one?

samcli/lib/config/samconfig.py
```diff
@@ -102,20 +112,15 @@ def put(self, cmd_names, section, key, value, env=DEFAULT_ENV):
...
         ------
-        tomlkit.exceptions.TOMLKitError
+        SamConfigFileReadException
             If the data is invalid
         """
-
-        if self.document is None:
-            # Checking for None here since a TOMLDocument can include a
-            # 'body' property but still be falsy without a 'value' property
-            self._read()
```
hnnasit, line 118: Asking for my understanding, why is this `_read` method call removed? If this method call needs to be removed, is there anything else that raises `SamConfigFileReadException` exception as mentioned in line 121?

samcli/lib/config/samconfig.py
```diff
@@ -144,30 +146,29 @@ def put_comment(self, comment):
...
 
         Raises
         ------
-        tomlkit.exceptions.TOMLKitError
+        SamConfigFileReadException
             If the data is invalid
 
         """
         self._write()
+        print(self.document)
```
mildaniel, line 166: Residual debug print?

samcli/lib/config/samconfig.py
```diff
@@ -144,30 +146,29 @@ def put_comment(self, comment):
         comment: str
             A comment to write to the samconfg file
         """
-        if self.document is None:
-            self._read()
 
-        self.document.add(tomlkit.comment(comment))
+        self.document.update({"__comment__": comment})
```
mildaniel, line 153: Did you verify that the TOML library knows to handle this line as a comment?

mildaniel (commented):
Awesome work, this looks great! Mainly left a few minor questions.

I was also wondering if there was any thought put into changing the default `samconfig.toml` that we generate during `sam init` into a different supported format, or to provide customers a flag to do so?

samcli/lib/config/samconfig.py
Leo10Gama: That's a good catch, calling `self._read()` would also probably be faster here since `self.document` would likely already be stored and we wouldn't have to do any I/O. I'll update accordingly.

samcli/lib/config/samconfig.py
Leo10Gama: Formerly, the `_read` method was being called in every function that used `self.document`, but now it performs the read once upon initialization, so we don't have to scatter function calls like this any time we want to look through the config.

samcli/lib/config/file_manager.py
Leo10Gama: In this case, my logic was that if the file itself couldn't be found, then we simply continue execution as if there is no configuration set. Meanwhile, if the file does exist and is formatted incorrectly, we raise the exception and let the customer know. We could throw an exception for the file not being found as well if that's preferred, but that's my reasoning for not throwing one in these instances

samcli/lib/config/samconfig.py
Leo10Gama: I think that comment is outdated from previously, where the `file_manager` would default to `TomlFileManager` if the given config file isn't supported, which I resolved in a previous commit. If the customer has a config file, it'll be set and passed into the constructor via `cli_config_file.py::TomlProvider::__call__`, which passes in the `config_path` which is defined if there's an existing config. Only if that isn't defined will it use the default name

samcli/lib/config/samconfig.py
Leo10Gama: I assumed the TOML library would, but looking through, I think it may have to be manually commented in through the `TomlFileManager.write()` method. Will update accordingly 👍

Leo10Gama:
@mildaniel 
> I was also wondering if there was any thought put into changing the default `samconfig.toml` that we generate during `sam init` into a different supported format, or to provide customers a flag to do so?

I believe there were talks about migrating off of TOML as the default, and if we do decide to do that, it would be as simple as changing `DEFAULT_CONFIG_FILE_EXTENSION` in samconfig.py to be whatever we would want the default file extension to be, like yaml or json. In terms of backwards compatibility, as long as customers already have existing samconfig.toml files, altering this default value shouldn't affect them.

samcli/lib/config/samconfig.py
hnnasit: Understood :+1: , can we update the doc-string if `SamConfigFileReadException` is not raised in this method?

samcli/lib/config/samconfig.py
Leo10Gama: As discussed in our 1:1 earlier this morning, that logic will be put into `SamConfig` after all the other `FileManager` subclasses are created. Expected behaviour for that is that it will first check if a config file is provided and exists, then if a TOML exists, then YAML, then JSON, and if none of those exist, default to TOML.

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        if not document:
+            LOG.debug("No document given for TomlFileManager to write.")
+            return
+
+        doc_no_comments = {k: v for k, v in document.items() if k != "__comment__"}
+        toml_doc = tomlkit.document()
+
+        if document.get("__comment__", None):  # Comment appears at the top of doc
+            toml_doc.add(tomlkit.comment(document["__comment__"]))
```
lucashuy, line 108: (from discussion): What happens if we have more than one comment in the file, will those disappear? I wonder what the current behaviour is with regard to non top level comments in a samconfig.toml file

Leo10Gama: Previously, `tomlkit` was able to store the comment information somewhere in its structure. But in this refactor to change `SamConfig.document` from a TOML document to a dictionary, we lose the functionality of these inline comments. Implementing these changes now will essentially get rid of any customer comments in their config files, but if this is functionality we want to preserve, we could alter the code to keep a copy of the information as a `tomlkit.document()` as well as a dictionary.

Leo10Gama: Update: multiple comments will now be preserved thanks to recent changes, adding in the `SamConfig.file_document` attribute, which currently holds the `tomlkit.TOMLDocument`. When the file is read using the `TomlFileManager`, it stores the document there, and when writing, it checks if it exists, and writes using the appropriate methods.

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,187 @@
...
+            The final location for the TOML file to be written.
+        """
+        if not document:
+            LOG.debug("No document given for TomlFileManager to write.")
+            return
+
+        doc_no_comments = {k: v for k, v in document.items() if k != "__comment__"}
+        toml_doc = tomlkit.document()
+
+        if document.get("__comment__", None):  # Comment appears at the top of doc
```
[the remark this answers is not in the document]
lucashuy: Sounds good, I think that if there was a case where we converted from TOML/YAML to JSON, all the inline comments and nested comments would be lost (only the root level comment is preserved), however thats just a limitation of regular JSON

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,187 @@
...
+        document: dict
+            The object to write.
+        filepath: Path
+            The final location for the TOML file to be written.
+        """
+        if not document:
+            LOG.debug("No document given for TomlFileManager to write.")
+            return
+
+        doc_no_comments = {k: v for k, v in document.items() if k != "__comment__"}
```
sriram-mv, line 141: lets move up all constants.

samcli/lib/config/samconfig.py
```diff
@@ -39,11 +40,21 @@ def __init__(self, config_dir, filename=None):
             Optional. Name of the configuration file. It is recommended to stick with default so in the future we
             could automatically support auto-resolving multiple config files within same directory.
         """
+        self.document = {}
         self.filepath = Path(config_dir, filename or DEFAULT_CONFIG_FILE_NAME)
+        self.file_manager = self.FILE_MANAGER_MAPPER.get(self.filepath.suffix[1:], None)
```
sriram-mv, line 45: is there no native file extension api?

samcli/commands/deploy/guided_config.py
```diff
@@ -26,13 +27,17 @@ def get_config_ctx(self, config_file=None):
...
         msg = (
             "Syntax invalid in samconfig.toml; save values "
             "through sam deploy --guided to overwrite file with a valid set of values."
         )
+        try:
+            _, samconfig = self.get_config_ctx(config_file)
+        except SamConfigFileReadException:
+            raise GuidedDeployFailedError(msg)
+
+        status = "Found" if samconfig.exists() else "Not found"
```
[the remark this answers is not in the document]
Leo10Gama: I think letting the `status` be an enum might be a bit of overkill, since it's only used in this one method, and only really exists to be appended to the following click message. We could make it an enum, but I don't think it's necessary in this case.

samcli/lib/config/samconfig.py
sriram-mv: if not, we can move the the filemanager mapper, to just have .toml, .yaml and so-on as the keys. having to do string splicing, makes me queasy on how things can break.

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,187 @@
...
+        document: dict
+            The object to write.
+        filepath: Path
+            The final location for the document to be written.
+        """
+        raise NotImplementedError("Write method not implemented.")
+
+    @staticmethod
+    @abstractmethod
+    def write_document(document: Any, filepath: Path):
```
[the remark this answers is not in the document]
Leo10Gama: We could change them to `write_dict()` and `write_document()` if that adds more clarity. I'm thinking that `write_dict()` will write a generic dictionary to its respective document type, while `write_document()` writes the file document object, which may have its own special formatting. In the case of TOML, it would take in a `tomlkit.TOMLDocument` and write it, which would have support for comments.

samcli/lib/config/samconfig.py
Leo10Gama: The closest thing we can get with `pathlib.Path` is suffix, which returns the extension in the format of ".toml", with the period. In this file, extensions are usually referred to without the period, such as the case with `DEFAULT_CONFIG_FILE_EXTENSION`. Hence, anywhere I map extensions, such as the `FILE_MANAGER_MAPPER`, I use the extension name without the period.

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,187 @@
...
+
+        if document.get("__comment__", None):  # Comment appears at the top of doc
+            toml_doc.add(tomlkit.comment(document["__comment__"]))
+        for k, v in doc_no_comments.items():
+            toml_doc.add(k, v)
+
+        filepath.write_text(tomlkit.dumps(toml_doc))
+
+    @staticmethod
+    def write_document(document: Any, filepath: Path):
```
hnnasit, line 152: Can we use specific typing instead of `Any`, `document: TOMLDocument`? Makes it easier to for type checking to catch any errors.

Leo10Gama: Yup, in my changes just pushed, we set it to be generic `dict` since the `TOMLDocument` is a subclass of dictionaries.

Leo10Gama: We could change to be the specific `TOMLDocument`, but then the method signature differs from the abstract class it inherits from

mildaniel (approved):

lucashuy (approved):
Thanks for working with the comments and feedback

samcli/commands/deploy/guided_config.py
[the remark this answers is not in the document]
sriram-mv: Having strings be status seems counter-intuitive to me, won't block on it. but we should leave a todo at the very least.

samcli/lib/config/file_manager.py
```diff
@@ -0,0 +1,161 @@
...
+
+import tomlkit
+
+from samcli.lib.config.exceptions import FileParseException
+
+LOG = logging.getLogger(__name__)
+COMMENT_KEY = "__comment__"
+
+
+class FileManager(ABC):
```
sriram-mv, line 19: Nice, I don't see write_document anymore

sriram-mv (approved):

Issue #5207, opened 2023-05-26
Bug: Hardcoded stage count in console output


Detectronic-PB:
### Description:
When using a custom template with more than 2 stages, the output still shows:

> 2 stage(s) were detected, matching the template requirements.

This is due to the hardcoded value on [this line](https://github.com/aws/aws-sam-cli/blob/de9da32eeb4300074d9357649bdf8fb1bcbe491f/samcli/commands/pipeline/init/interactive_init_flow.py#L259).

### Steps to reproduce:

1. Copy the contents of [two-stage-pipeline-template](https://github.com/aws/aws-sam-cli-pipeline-init-templates/tree/main/AWS-CodePipeline/two-stage-pipeline-template) into a new repository
2. Modify metadata.json to: `{
  "number_of_stages": 3
}`
3. Commit changes
4. Run `sam pipeline init`
5. Select '2 - Custom Pipeline Template Location'
6. Enter repository path

### Observed result:
```
sam pipeline init generates a pipeline configuration file that your CI/CD system
can use to deploy serverless applications using AWS SAM.
We will guide you through the process to bootstrap resources for each stage,
then walk through the details necessary for creating the pipeline config file.

Please ensure you are in the root folder of your SAM application before you begin.

Select a pipeline template to get started:
        1 - AWS Quick Start Pipeline Templates
        2 - Custom Pipeline Template Location
Choice: 2
Template Git location: https://github.com/xxxxxxx/test-aws-pipeline-template

Cloning from https://github.com/xxxxxxx/test-aws-pipeline-template (process may take a moment)
You are using the 3-stage pipeline template.
 _________    _________    _________
|         |  |         |  |         |
| Stage 1 |->| Stage 2 |->| Stage 3 |
|_________|  |_________|  |_________|

Checking for existing stages...

2 stage(s) were detected, matching the template requirements. If these are incorrect, delete .aws-sam/pipeline/pipelineconfig.toml and rerun
```

### Expected result:
```
sam pipeline init generates a pipeline configuration file that your CI/CD system
can use to deploy serverless applications using AWS SAM.
We will guide you through the process to bootstrap resources for each stage,
then walk through the details necessary for creating the pipeline config file.

Please ensure you are in the root folder of your SAM application before you begin.

Select a pipeline template to get started:
        1 - AWS Quick Start Pipeline Templates
        2 - Custom Pipeline Template Location
Choice: 2
[...]

PR #5210, opened 2023-05-26
fix: fix the hardcoded number of stages printed in logs.

moelasmar:
https://github.com/aws/aws-sam-cli/issues/5207

Removed the hardcoded number, and replaced it with the number of stages.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/pipeline/init/interactive_init_flow.py
+++ b/samcli/commands/pipeline/init/interactive_init_flow.py
@@ -256,7 +256,7 @@ class InteractiveInitFlow:
                 continue
             click.echo(
                 Colored().yellow(
-                    "2 stage(s) were detected, matching the template requirements. "
+                    f"{number_of_stages} stage(s) were detected, matching the template requirements. "
                     "If these are incorrect, delete .aws-sam/pipeline/pipelineconfig.toml and rerun"
                 )
             )


samcli/commands/pipeline/init/interactive_init_flow.py
```diff
@@ -256,7 +256,7 @@ def _generate_from_pipeline_template(
                 continue
             click.echo(
                 Colored().yellow(
-                    "2 stage(s) were detected, matching the template requirements. "
+                    f"{number_of_stages} stage(s) were detected, matching the template requirements. "
```
mndeveci, line 259: Can we update existing unit tests to confirm this message from now on?

mndeveci (approved):
Thanks for fixing this! It is really straight forward but I wonder if we can update existing unit tests to verify the behavior.

ssenchenko (approved):

samcli/commands/pipeline/init/interactive_init_flow.py
moelasmar: let me check the existing unit test cases.

PR #5218, opened 2023-05-29
fix: fix failing Terraform integration test cases

moelasmar:
#### Changes:
 - Removed the `skip_get_ec2_platforms` flag from the sample Terraform testing project, AS Hashicorp pushed a new AWS Provider and removed this flag.
 - Fix the resource address while accessing the module config resources
 - Fix checking the experimental log integration test cases

#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/translate.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/translate.py
@@ -111,15 +111,21 @@ def _check_unresolvable_values(root_module: dict, root_tf_module: TFModule) -> N
         # iterate over resources for current module
         for resource in curr_module.get("resources", []):
             resource_type = resource.get("type")
+            resource_name = resource.get("name")
+            resource_mode = resource.get("mode")
 
             resource_mapper = RESOURCE_TRANSLATOR_MAPPING.get(resource_type)
             if not resource_mapper:
                 continue
 
             resource_values = resource.get("values")
-            resource_full_address = resource.get("address")
+            resource_address = (
+                f"data.{resource_type}.{resource_name}"
+                if resource_mode == "data"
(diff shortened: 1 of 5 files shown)


lucashuy (approved):

samcli/hook_packages/terraform/hooks/prepare/translate.py
```diff
@@ -111,15 +111,21 @@ def _check_unresolvable_values(root_module: dict, root_tf_module: TFModule) -> N
...
             resource_type = resource.get("type")
+            resource_name = resource.get("name")
+            resource_mode = resource.get("mode")
 
             resource_mapper = RESOURCE_TRANSLATOR_MAPPING.get(resource_type)
             if not resource_mapper:
                 continue
 
             resource_values = resource.get("values")
-            resource_full_address = resource.get("address")
```
mndeveci, line 120: We are not using `address` field anymore and generating its value with the information that we have? If so, do you think there might be some edge cases if they change the way to construct that field, should we keep using it and prepend `data.` for the case below?

mndeveci (commented):
Thanks for the fix @moelasmar, posted a question to understand more about the implementation and the problem.

samcli/hook_packages/terraform/hooks/prepare/translate.py
moelasmar: Yes, the reason is this address contains the full address of the resource starting from the root module, but when we built the module configuration data structure we used the address in the module (which is <resource type>.<resource name> in case of resources, and data.<resource type>.<resource name> in case of data resource). This is because the same module configuration can be used multiple times to create actual modules

mndeveci (approved):

PR #5220, opened 2023-05-29
feat: Return early during linking if no destination resources are found

lucashuy:

N/A

If there are no destinations found during linking, the call back method to link the properties together still run.

Adds a check to exit early in the linking handler, in addition to the existing checks in the call back methods. As a result, this will not call the call back methods.

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
+++ b/samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
@@ -231,9 +231,12 @@ class ResourceLinker:
             )
         if not dest_resources:
             LOG.debug(
-                "There are destination resources defined for for the source resource %s",
+                "There are destination resources defined for for the source resource %s, skipping linking.",
                 source_tf_resource.full_address,
             )
+
+            return
+
         for cfn_resource in cfn_resources:
             self._resource_pair.cfn_resource_update_call_back_function(cfn_resource, dest_resources)  # type: ignore
 
@@ -284,7 +287,12 @@ class ResourceLinker:
             else ExistingResourceReference(value)
             for value in values
         ]
-        LOG.debug("The value of the source resource linking field after mapping $s", dest_resources)
+
+        if not dest_resources:
+            LOG.debug("Skipping linking call back, no destination resources discovered.")
+            return
+
+        LOG.debug("The value of the source resource linking field after mapping %s", dest_resources)
         self._resource_pair.cfn_resource_update_call_back_function(cfn_resource, dest_resources)  # type: ignore
 
     def _process_resolved_resources(
--- a/tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
+++ b/tests/unit/hook_packages/terraform/hooks/prepare/test_resource_linking.py
@@ -1124,6 +1124,21 @@ class TestResourceLinker(TestCase):
             linking_exceptions=self.linker_exceptions,
         )
 
+    def test_applied_empty_destination_skip_call_back(self):
+        resource_linker = ResourceLinker(self.sample_resource_linking_pair)
+        resource_linker._link_using_linking_fields({"Properties": {"Layers": []}})
+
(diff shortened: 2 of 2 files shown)


samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -231,9 +231,12 @@ def _link_using_terraform_config(self, source_tf_resource: TFResource, cfn_resou
             )
         if not dest_resources:
             LOG.debug(
-                "There are destination resources defined for for the source resource %s",
+                "There are destination resources defined for for the source resource %s, skipping linking.",
                 source_tf_resource.full_address,
             )
+
+            return
```
mndeveci, line 238: LGTM in general. I am just wondering why having 2 destinations is throwing an error (at line 229 above) but defining none is skipping this step.

lucashuy: [This comment](https://github.com/aws/aws-sam-cli/pull/5196#discussion_r1206143196) explains it in a little more detail, but essentially we don't want to block the execution of some testing cases where the customer may have a partially defined API (eg. defined gateway, but missing some parts of method or integration).

If they were looking to only test their Lambda function using `sam local invoke`, blocking here on missing links would require them to fully define an API Gateway to test their Lambda function.

mndeveci (approved):

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
mndeveci: If this method linking layers to functions and also APIGW methods/paths to functions, should we have distinct ways?

Like different part of our logic can call this function (either linking layers to functions or apigw methods/paths to functions), then this method can do whatever it supposed to do. The caller function then can decide whether the case is valid for that purpose or not. I feel like this method is been overloaded with different purposes, and this might lead to a bug in the future. (Of course there is a chance that I mis-understood everything wrong, in that case please ignore my comment :D )

moelasmar: This function is a generic function that we used to link different resources. The logic of what to be linked is maintained here https://github.com/aws/aws-sam-cli/blob/develop/samcli/hook_packages/terraform/hooks/prepare/resources/resource_links.py#L29-L81

moelasmar: for your comment regarding the exception we raise in line 229, this is related to the Terraform limitations we have, at this point we could not differentiate if the customer is really linking to 2 different resources, or using some conditional expression to choose one of them. As, we do not know what to be done in this case, shall we use the 2 destination resources, or choose one of them, so, we decided to raise an exception to the customer that we do not support this case, and they can resolve this issue by running `terraform apply` command, so we can use a different linking approach where we do not have any issues.

But, for no destination resources found at all, this means that there is no destinations at all to be linked, and should not be an issue (this like if the customer does not have any layers defined for a lambda function), so we do not raise exception here. Sometimes, it can be an issue in the customer terraform project, but we decided not to raise the exception here, and leave this decision to each SAM CLI command to decide if it is an issue or not (Like the customer define a Rest API method, but does not link it to a Lambda function)

samcli/hook_packages/terraform/hooks/prepare/resource_linking.py
```diff
@@ -231,9 +231,12 @@ def _link_using_terraform_config(self, source_tf_resource: TFResource, cfn_resou
             )
         if not dest_resources:
             LOG.debug(
-                "There are destination resources defined for for the source resource %s",
+                "There are destination resources defined for for the source resource %s, skipping linking.",
```
moelasmar, line 234: can you update the log message to be There are **no** destination resources

moelasmar (approved):
left a minor comment

moelasmar (approved):
left a minor comment

PR #5222, opened 2023-05-30
fix: handle edge cases with function sync flow in sam sync command

mndeveci:
#5034

When customers provide pre-built ZIP file for their functions, `sam sync` would fail since it always tries to build the function. Same happens when customers denote their functions with `SkipBuild: true`.

When creating `ZipFunctionSyncFlow`, it checks whether function's `CodeUri` is pointing to a `.zip` file, or it has `SkipBuild: True` in function metadata. If so then it instantiates different sync flows.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/build_context.py
```diff
@@ -650,34 +642,6 @@ def _collect_single_buildable_layer(
 
         resource_collector.add_layer(layer)
 
-    @staticmethod
```
mndeveci, line 653: This functionality is moved to `Function` class and `FunctionBuildInfo` enum below.

samcli/lib/sync/sync_flow_factory.py
```diff
@@ -153,13 +157,33 @@ def _create_lambda_flow(
         resource: Dict[str, Any],
```
hnnasit, line 157: Can this parameter be removed since this is replaced with the function from the build_context?

hnnasit (commented):
Overall LGTM, left some small comments.

samcli/lib/sync/sync_flow_factory.py
mndeveci: Unfortunately not, this is called from below (line 373) with generic function signature, if we remove it then it will break that call.

samcli/commands/build/build_context.py
```diff
@@ -5,44 +5,40 @@
...
-    BuildError,
-    UnsupportedBuilderLibraryVersionError,
-    ApplicationBuildResult,
-)
-from samcli.commands._utils.constants import DEFAULT_BUILD_DIR
-from samcli.lib.build.workflow_config import UnsupportedRuntimeException
 from samcli.local.lambdafn.exceptions import FunctionNotFound
-from samcli.commands._utils.template import move_template
-from samcli.lib.build.exceptions import InvalidBuildGraphException
+from samcli.local.lambdafn.exceptions import ResourceNotFound
```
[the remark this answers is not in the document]
mndeveci: Weird that ruff didn't bark at any of this, will update them in next revision 👍🏼

samcli/lib/sync/sync_flow_factory.py
hnnasit: Seems like only `_create_lambda_flow` factory method was using that parameter and the other methods had to pass it due to generic signature. Could that parameter be removed from all the methods or would it be useful to keep that parameter for any future resources added?

mndeveci: Himm actually you are right and that is a very good catch! Let's see if I can remove it completely.

samcli/lib/providers/provider.py
```diff
@@ -138,6 +161,36 @@ def architecture(self) -> str:
             )
         return str(arch_list[0])
 
+    @property
+    def build_info(self) -> FunctionBuildInfo:
```
jfuss, line 165: Nit: Should this be a dynamic property or something we could set once on creation?

jfuss (approved):

hnnasit (approved):

Issue #5229, opened 2023-05-31
Bug: sam local invoke still print the response in latin-1


guanlinz:
<!-- Make sure we don't have an existing Issue that reports the bug you are seeing (both open and closed). 
If you do find an existing Issue, re-open or add a comment to that Issue instead of creating a new one. -->

### Description:
<!-- Briefly describe the bug you are facing.-->
 Passing the event in Chineses to sam local invoke produces results like unsupported character.

This issue only hapeens in STDOUT with `sam local invoke`, I can get correct and human readable Chineses character via API Gateway

I go through the closed issue and got this: https://github.com/aws/aws-sam-cli/issues/2494, seems like the issue is fixed in 2021, but in my testing result, it didn't.


### Steps to reproduce:
<!-- Provide detailed steps to replicate the bug, including steps from third party tools (CDK, etc.) -->

SAM Template:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  helloword-app

  Sample SAM Template for helloword-app

# More info about Globals: https://github.com/awslabs/serverless-application-model/blob/master/docs/globals.rst
Globals:
  Function:
    Timeout: 3
    MemorySize: 128

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
    Properties:
      CodeUri: hello_world/
      Handler: app.lambda_handler
      Runtime: python3.9
      Architectures:
        - x86_64
      Events:
        HelloWorld:
          Type: Api # More info about API Event Source: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#api
          Properties:
            Path: /hello
            Method: get

Outputs:
  # ServerlessRestApi is an implicit API created out of Events key under Serverless::Function
  # Find out more about other implicit resources you can reference within SAM
  # https://github.com/awslabs/serverless-application-model/blob/master/docs/internals/generated_resources.rst#api
  HelloWorldApi:
    Description: "API Gateway endpoint URL for Prod stage for Hello World function"
    Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/hello/"
  HelloWorldFunction:
    Description: "Hello World Lambda Function ARN"
    Value: !GetAtt HelloWorldFunction.Arn
[...]

PR #5232, opened 2023-05-31
fix(invoke): Write in UTF-8 string instead of bytes.

jfuss:
It appears that we were using sys.stdout.buffer to support python2 and python3 at the same time. Switching to just write to sys.stdout allows us to write a utf-8 encoding string. When using sys.stdout.buffer, we can only write bytes and I couldn't get the correct UTF8 encoded string to print correctly.

#5229

Should allow us to write utf8 correctly to the terminal.

See above git commit message

We may need to change how we write in more places but my system python is all messed up so going to rely on our tests to validate in the PR. I did validate this locally with the example in the Github Issue.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hawflau (approved):
LGTM

samcli/lib/utils/stream_writer.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, stream, auto_flush=False):
     def stream(self):
         return self._stream
 
-    def write(self, output, encode=False):
+    def write(self, output, encode=False, write_to_buffer=True):
```
mndeveci, line 25: Just wondering, are there other places where this is called with `write_to_buffer=True`?

mndeveci (approved):

jfuss:
Added integration test for this. I ran both of develop and my branch. Results below:

develop:
```
$ SAM_CLI_DEV=1 pytest tests/integration -k test_invoke_returns_utf8
...

tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_0::test_invoke_returns_utf8 FAILED                                                                                                                                                                                                                                     [ 50%]
tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_1::test_invoke_returns_utf8 FAILED
...
FAILED tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_0::test_invoke_returns_utf8 - AssertionError: '{"key1": "안녕", "key2": "😍", "key3": "ḧ̸͈̟o̸̡̧͒ẉ̴̌ ̶[67 chars]̛͈"}' != '{"key1": "\\uc548\\ub155", "key2": "\\ud83d\\ude0d"[512 chars]48"}'
FAILED tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_1::test_invoke_returns_utf8 - AssertionError: '{"key1": "안녕", "key2": "😍", "key3": "ḧ̸͈̟o̸̡̧͒ẉ̴̌ ̶[67 chars]̛͈"}' != '{"key1": "\\uc548\\ub155", "key2": "\\ud83d\\ude0d"[512 chars]48"}'
```

jfuss:issue/5229:
```
$ SAM_CLI_DEV=1 pytest tests/integration -k test_invoke_returns_utf8
...

tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_0::test_invoke_returns_utf8 PASSED                                                                                                                                                                                                                                     [ 50%]
tests/integration/local/invoke/test_integrations_cli.py::TestSamPython37HelloWorldIntegration_1::test_invoke_returns_utf8 PASSED
```

samcli/lib/utils/stream_writer.py
jfuss: I added `write_to_buffer` to keep all the previous code paths the same. I was seeing different behavior when writing to `sys.stdout.buffer` and `sys.stdout`. I wanted to write to `sys.stdout` for the output of the invoke but leave everything else writing to the buffer. So I added this option to the write call, defaulted to the previous behavior of writing to the buffer, and then passing is `False` for my specific case. 

Ideally, I would have wanted to change everything but that was a much bigger change and coming from all the docker upgrade issues, I went the least impactful route.

hawflau (approved):

PR #5238, opened 2023-06-01
feat: Add click command for cloud invoke command

hnnasit:
https://github.com/aws/aws-sam-cli/discussions/5018

This PR adds the click command for the new feature called `cloud invoke`. The formatting of the UX will be done in a separate PR.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/_utils/options.py
```diff
@@ -571,6 +587,21 @@ def image_repositories_option(f):
...
 
 
+def remote_invoke_parameter_click_option():
+    return click.option(
+        "--parameter",
+        multiple=True,
+        type=RemoteInvokeBotoAPIParameterType(),
+        callback=remote_invoke_boto_parameter_callback,
+        required=False,
+        help="Additional parameters for the boto API call",
```
mndeveci, line 597: Should we add name of the APIs so that customers can find other keys of the specific API?

samcli/cli/types.py
```diff
@@ -421,3 +421,23 @@ def convert(self, value, param, ctx):
...
+    def convert(self, value, param, ctx):
+        # Split by first "=" as some values could have multiple "=" For e.g. base-64 encoded ClientContext for Lambda
+        key_value_pair = value.split("=", 1)
+        if len(key_value_pair) < self.MIN_KEY_VALUE_PAIR_LENGTH:
+            raise click.BadParameter(
+                f"{param.opts[0]} is not a valid format, it needs to be of the form parameter_key=parameter_value"
+            )
+        key = key_value_pair[0]
+        _value = key_value_pair[1]
+        return {key: _value}
```
mndeveci, line 443: Some debug logging might be useful

samcli/commands/remote_invoke/cloud.py
```diff
@@ -0,0 +1,19 @@
+"""
+Command group for "cloud" suite for commands. It provides common CLI arguments, template parsing capabilities,
+setting up stdin/stdout etc
```
mndeveci, line 3: I don't understand what do we mean by "template parsing capabilities"

samcli/commands/remote_invoke/invoke/cli.py
```diff
@@ -0,0 +1,146 @@
...
+        with RemoteInvokeContext(
+            boto_client_provider=boto_client_provider,
+            boto_resource_provider=boto_resource_provider,
+            stack_name=stack_name,
+            resource_id=resource_id,
+        ) as remote_invoke_context:
+
+            if not resource_id and not stack_name:
+                LOG.error("Atleast 1 of --stack-name or --resource-id parameters should be provided")
+                return
```
mndeveci, line 121: Let's validate this via click option configuration so we don't need to manually check here.

samcli/commands/remote_invoke/invoke/cli.py
```diff
@@ -0,0 +1,146 @@
...
+
+            remote_invoke_result = remote_invoke_context.run(remote_invoke_input=remote_invoke_input)
+
+            if remote_invoke_result.is_succeeded():
+                if remote_invoke_result.log_output:
+                    remote_invoke_context.stderr.write(remote_invoke_result.log_output.encode())
+                output_response = cast(str, remote_invoke_result.response)
+                remote_invoke_context.stdout.write(output_response.encode())
+            else:
+                raise cast(Exception, remote_invoke_result.exception)
```
mndeveci, line 144: Should we raise whatever is been raised from the call, or should we wrap it into specific user exception and raise that instead?

mndeveci (changes requested):

samcli/commands/remote_invoke/cloud.py
hnnasit: That's on me, I copied the text from `local.py` and missed updating the text there.

samcli/commands/remote_invoke/invoke/cli.py
hnnasit: So any exception that occurs in `/lib` is set to `remote_invoke_result.exception` parameter. Line 144 raises that exception which is caught on line [145](https://github.com/aws/aws-sam-cli/blob/08f296126e2d3d1793bea36056e6db31712283f6/samcli/commands/remote_invoke/invoke/cli.py#L145) and raised as `UserException`.  Do you mean we should name the specific UserExceptions being raised in line 145?

mndeveci: Oh, if we are already sorting that out in the inner layers, than we should be good for now. Thanks for explaining it.

samcli/cli/types.py
```diff
@@ -421,3 +424,40 @@
...
+        # Split by first "=" as some values could have multiple "=" For e.g. base-64 encoded ClientContext for Lambda
+        key_value_pair = value.split("=", 1)
+        if len(key_value_pair) < self.MIN_KEY_VALUE_PAIR_LENGTH:
+            raise click.BadParameter(
+                f"{param.opts[0]} is not a valid format, it needs to be of the form parameter_key=parameter_value"
+            )
+        key = key_value_pair[0]
+        _value = key_value_pair[1]
+        converted_dict = {key: _value}
+        LOG.debug("Converting provided parameter %s to dict %s", value, converted_dict)
```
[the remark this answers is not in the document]
hnnasit: Nothing sensitive, this logs boto API parameter values provided  to the users terminal

samcli/commands/remote_invoke/cloud.py
mndeveci, line 1: I think we need to move this file under `samcli/commands/cloud/cloud.py` or `samcli/commands/cloud/command.py`. And then `invoke` would be a sub-command under that folder.

samcli/lib/cli_validation/remote_invoke_options_validations.py
```diff
@@ -0,0 +1,94 @@
...
+                "Please check that you don't have both specified in the command or in a configuration file",
+            ),
+        )
+
+        validator.validate()
+
+        # if no event nor event_file arguments are given, read from stdin
+        if not event and not event_file:
+            LOG.info("Neither --event nor --event-file options have been provided, reading from stdin")
+            kwargs["event_file"] = cast(TextIOWrapper, sys.stdin)
```
mndeveci, line 54: why not use the `Validator` pattern above?

samcli/commands/remote_invoke/invoke/cli.py
```diff
@@ -0,0 +1,139 @@
...
+Invoke or send an event to cloud resources in your CFN stack
+"""
+SHORT_HELP = "Invoke a deployed resource in the cloud"
+
+
+@click.command("invoke", help=HELP_TEXT, short_help=SHORT_HELP)
+@configuration_option(provider=TomlProvider(section="parameters"))
+@click.option("--stack-name", required=False, help="Name of the stack to get the resource information from")
+@click.option("--resource-id", required=False, help="Name of the resource that will be invoked")
+@click.option(
```
sriram-mv, line 33: we need similarly named options to have homogenity between local and cloud invoke?

samcli/commands/remote_invoke/remote_invoke_context.py
```diff
@@ -187,3 +189,29 @@ def _get_from_physical_resource_id(self) -> CloudFormationResourceSummary:
                     f"Please provide full resource ARN or --stack-name to resolve the ambiguity."
                 )
             return resource_summary
+
+    @property
+    def stdout(self) -> StreamWriter:
```
sriram-mv, line 194: Should this be more common than here?

samcli/lib/cli_validation/remote_invoke_options_validations.py
hnnasit: Discussed offline that `Validator` raises an exception when the validation function succeeds. For the use case when `event` and `event-file` are not provided, this will read from `stdin`. For this reason, we will keep it as it is and change the log level to `debug`.

samcli/commands/remote_invoke/invoke/cli.py
hnnasit: Yeah, this is something we will be discussing with the UX team next week. I will raise a follow up PR for the new UX changes to the help text and can add any named option changes there.

samcli/commands/remote_invoke/remote_invoke_context.py
hnnasit: I added it to the context because if we decide to add another handle, for instance log_file option that `sam local` has, we can add that here directly after passing it to the context.

mndeveci (approved):
Thanks for addressing all the feedback. Left a comment about a class naming, I think we need to rename all modules & classes once we decide the command naming.

samcli/cli/types.py
```diff
@@ -421,3 +424,57 @@ def convert(self, value, param, ctx):
         if not is_ecr_url(_value):
             raise click.BadParameter(f"{param.opts[0]} needs to have valid ECR URI as value")
         return {key: _value}
+
+
+class RemoteInvokeBotoApiParameterType(click.ParamType):
```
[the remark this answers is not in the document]
hnnasit: Yup that's right, left it as remote invoke here to keep it consistent with other places that use remote invoke. We can refactor all modules/classes/variable names once we confirm the name to be used.

PR #5250, opened 2023-06-02
Artifact export for GraphQLApi

ssenchenko:
Adding `sam package` support for SAM `GraphQLApi` resource. 
`GraphQLApi` has 2 properties which can use local paths for code: `SchemaUri` and `CodeUri`.

`SchemaUri` is not nested and has a single instance in the resource, so it's handled the same way as the rest of the paths resources, without any code change.

`CodeUri` can be present in both `Resolvers` and `Functions` (see the resource example below). `GraphQLApi` may have many instances of `CodeUri`. In addition, `CodeUri` is nested in the dictionaries with keys which cannot be standardized because they are resolvers' and functions' names. Since `botocore.set_value_from_jmespath`, which is used to set a property to S3 URI after upload, supports only dotted lookups and no wildcards, `GraphQLApi`'s `CodeUri` requires special treatment. I had to write a custom method to look up all the paths together with the values, because `jmespath` library, which is used by the base class, doesn't allow to "trace" the paths.

Here is a small example of `GraphQLApi` resource:
```yaml
  SuperCoolAPI:
    Type: AWS::Serverless::GraphQLApi
    Properties:
      SchemaUri: "schema.graphql"
      Defaults:
        Runtime:
          Name: APPSYNC_JS
          Version: "1.0.0"
        Functions:
          DataSource: PostsDataSource
      Auth:
        Type: API_KEY
      ApiKeys:
        MyApiKey: {}
      DataSources:
        DynamoDb:
          PostsDataSource:
            TableName: !Ref DynamoDBPostsTable
      Functions:
        formatPostItem:
          DataSource: NONE
          CodeUri: "functions/formatPostItem.js"
        createPostItem:
          CodeUri: "functions/createPostItem.js"
        getPostFromTable:
          CodeUri: "functions/getPostFromTable.js"
      Resolvers:
        Mutation:
          addPost:
            Pipeline:
            - formatPostItem
            - createPostItem
        Query:
          getPost:
            Pipeline:
            - getPostFromTable
```

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


hawflau (commented):
LGTM overall. Two questions:
1. Does this change depend on an upcoming SAM-T version?
2. Can you add some integration tests?

mildaniel (commented):
LGTM too. Same comment as Wing to add some integration test cases.

ssenchenko:
I'll integ tests as a separate PR

hawflau (approved):

ssenchenko:
I don’t use any specific functionality from SAM-T package. So I guess it doesn’t depend on any specific SAM-T version

mildaniel (approved):

PR #5253, opened 2023-06-02
feat: Add YAML config file option

Leo10Gama:
With this addition, users will also be able to define their config files in YAML, adding additional support for those who may be unfamiliar with TOML.

A YAML `FileManager` child class has been added, which handles reading and writing to and from YAML files. It is used by `SamConfig` to read and write in config options.

Config files in TOML will not be affected, but any YAML config files can be passed in and read as well.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/config/file_manager.py
```diff
@@ -159,3 +161,90 @@ def put_comment(document: dict, comment: str) -> Any:
...
+        document = YamlFileManager._to_yaml(document)
+        document.yaml_set_start_comment(comment)
+        return document
+
+    @staticmethod
+    def _to_yaml(document: dict) -> Any:
+        """Ensure a dictionary-like object is a YAML document."""
+        stream = StringIO()
+        YamlFileManager.yaml.dump(document, stream)
+        return YamlFileManager.yaml.load(stream.getvalue())
```
lucashuy, line 250: Can we update the doc string here for the method?
Additionally, it looks like this opens an I/O stream to capture the string, can we close this stream after reading the value in, whether that be with a `.close()` or inside of a context?

Leo10Gama: How were you thinking to update the docstring? I checked through the `ruamel.yaml` docs and the return value seems to be `Any` for the parsed documents, and if it's set to something else, the linter complains about certain methods not being available.

lucashuy: Even though it is a pretty minimal method, the `document` parameter. As for the return value, though I'm not the biggest fan of `Any`, we could mention that the method returns a YAML document that is derived from an `dict` (or I guess more specifically, an `OrderedDict`)

lucashuy (approved):
Thanks for raising the PR

hnnasit (approved):
Looks good to me!

PR #5264, opened 2023-06-05
Add JsonFileManager

Leo10Gama:
With this addition, users will also be able to define their config files in JSON, adding additional support for those who may be unfamiliar with TOML.

A JSON `FileManager` child class has been added, which handles reading and writing to and from JSON files. It is used by `SamConfig` to read and write in config options.

Config files in TOML will not be affected, but any JSON config files can be passed in and read as well.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/config/file_manager.py
+++ b/samcli/lib/config/file_manager.py
@@ -3,6 +3,7 @@ Class to represent the parsing of different file types into Python objects.
 """
 
 
+import json
 import logging
 from abc import ABC, abstractmethod
 from pathlib import Path
@@ -37,8 +38,7 @@ class FileManager(ABC):
         Returns
         -------
         Any
-            A dictionary-like representation of the contents at the filepath location, along with a specialized
-            representation of the file that was read, if there is a specialization of it.
+            A dictionary-like representation of the contents at the filepath location.
         """
         raise NotImplementedError("Read method not implemented.")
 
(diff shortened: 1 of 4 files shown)


samcli/lib/config/file_manager.py
```diff
@@ -260,3 +260,74 @@ def _to_yaml(document: dict) -> Any:
...
+            The JSON object to write.
+        filepath: Path
+            The final location for the document to be written.
+        """
+        if not document:
+            LOG.debug("No document given to JsonFileManager to write.")
+            return
+
+        with filepath.open("w") as file:
+            json.dump(document, file)
```
lucashuy, line 313: Could we also specify `indent=2` to the dump method so that the file will be more easily editable?

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -16,7 +16,7 @@
 import logging
 from parameterized import parameterized
 from samcli.lib.config.exceptions import SamConfigFileReadException
-from samcli.lib.config.file_manager import YamlFileManager
+from samcli.lib.config.file_manager import JsonFileManager, YamlFileManager
 
 from samcli.lib.config.samconfig import SamConfig, DEFAULT_ENV, TomlFileManager
```
lucashuy, line 21: Must have missed this from before, but is the `TomlFileManager` importing from `samconfig` correct? I'm assuming that we should import this from `file_manager` instead.

tests/unit/lib/samconfig/test_file_manager.py
```diff
@@ -179,3 +180,98 @@ def test_yaml_put_comment(self):
...
+        config_path = Path(config_dir, "samconfig.json")
+
+        config_doc = JsonFileManager.read(config_path)
+
+        self.assertEqual(config_doc, {})
+
+    def test_write_json(self):
+        config_dir = tempfile.gettempdir()
+        config_path = Path(config_dir, "samconfig.json")
+        yaml = {
```
lucashuy, line 228: `json_config` (or something similar :) )

samcli/lib/config/file_manager.py
Leo10Gama: Now that you mention it, would it be better to specify it as a local parameter for `JsonFileManager`? I'm thinking that way we can tweak indentation values if need be, and prevent magic numbers.

tests/unit/commands/samconfig/test_samconfig.py
Leo10Gama: Yeah I didn't notice that at first, must've selected the wrong option in my IDE for resolving the import. It should be importing correctly this way, since all the tests pass and functionality is as expected, but I'll fix it regardless. Nice catch!

samcli/lib/config/file_manager.py
lucashuy: That sounds like a good improvement to have it in a variable

tests/unit/lib/samconfig/test_file_manager.py
```diff
@@ -186,17 +186,12 @@ class TestJsonFileManager(TestCase):
     def test_read_json(self):
         config_dir = tempfile.gettempdir()
         config_path = Path(config_dir, "samconfig.json")
+        indent_spaces = " " * JsonFileManager.INDENT_SIZE
```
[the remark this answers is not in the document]
Leo10Gama: Ah, it's not. Originally I was going to use it when making the temp JSON to test, but ended up using a different approach, and this is a remnant of that. Will remove 👍

lucashuy (approved):
Thanks for addressing the changes

hnnasit (approved):
Overall LGTM!

Issue #5243, opened 2023-06-01
Receiving "Error: request() got an unexpected keyword argument 'chunked'" after upgrading to SAM 1.85.0


donovan-middleton:
After upgrading to AWS SAM CLI from version 1.84.0 to version 1.85.0 using the `AWS_SAM_CLI_64_PY3.msi` installer, I receive the following error message when attempting to do a `sam build` command:
`Error: request() got an unexpected keyword argument 'chunked'`

And in writing up this Bug Report, also found I get the following error when running `sam --info`:
`TypeError: request() got an unexpected keyword argument 'chunked'`

### Steps to reproduce:
On Windows 11 Pro, upgrade from SAM 1.84.0 to SAM 1.85.0.
Using AWS SAM in a PowerShell 7.3.4 terminal within Visual Studio Code.
Running `sam build --template 'IL4SS-template.yml' --parallel --use-container` from within the folder the template is in.

#### With SAM 1.85.0:
```
PS C:\Users\donov\OneDrive\Documents\AFPOA\Lambda\Lambda Functions\SQS-to-Mail-Out> sam build --template 'IL4SS-template.yml' --parallel --use-container --debug
2023-06-01 17:52:21,805 | Config file location: C:\Users\donov\OneDrive\Documents\AFPOA\Lambda\Lambda Functions\SQS-to-Mail-Out\samconfig.toml
2023-06-01 17:52:21,809 | Config file 'C:\Users\donov\OneDrive\Documents\AFPOA\Lambda\Lambda Functions\SQS-to-Mail-Out\samconfig.toml' does not exist
2023-06-01 17:52:21,825 | Using config file: samconfig.toml, config environment: default
2023-06-01 17:52:21,827 | Expand command line arguments to:
2023-06-01 17:52:21,828 | --template_file=C:\Users\donov\OneDrive\Documents\AFPOA\Lambda\Lambda Functions\SQS-to-Mail-Out\IL4SS-template.yml --parallel --use_container --mount_with=READ     
--build_dir=.aws-sam\build --cache_dir=.aws-sam\cache
2023-06-01 17:52:22,240 | 'build' command is called
2023-06-01 17:52:22,241 | Starting Build inside a container
2023-06-01 17:52:22,248 | No Parameters detected in the template
2023-06-01 17:52:22,269 | There is no customer defined id or cdk path defined for resource Lambda, so we will use the resource logical id as the resource id
2023-06-01 17:52:22,271 | 0 stacks found in the template
[...]

PR #5279, opened 2023-06-07
chore: Upgrade Docker-py/ Support Py3.11 for running tests

jfuss:
#5243

Re-adds:
1. docker upgrade
2. getting SAM CLI dev dependencies installable in py3.11

This also fixes issues we found that `docker.from_env()` fails when docker is not running.

In order for the docker.from_env() not to fail when docker is not installed/running,
we force the min version on client creation. This was the default behavior in 4.X of
docker-py but not longer in the latest version.

None that I know of

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ lint:
 	# Linter performs static analysis to catch latent bugs
 	ruff samcli
 	# mypy performs type check
-	mypy --no-incremental setup.py samcli tests
+	mypy --exclude /testdata/ --exclude /init/templates/ --no-incremental setup.py samcli tests
 
 # Command to run everytime you make changes to verify everything works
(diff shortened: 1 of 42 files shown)


jfuss:
Marking as a draft, as I am working on adding a GHA that will validate at least one test that runs locally that creates a docker client but docker is not running. `sam build` without `-u` should be enough here.

sriram-mv:
https://github.com/aws/aws-sam-cli/pull/5289 is in relation to this PR.

sriram-mv (approved):

mildaniel (approved):

PR #5290, opened 2023-06-07
test: GHA to Execute Test without Docker Running

mildaniel:
To ensure that executing AWS SAM CLI commands while Docker isn't running succeeds.
Guard against regressions.

Creates a GitHub action that runs on PRs to ensure that a build test succeeds after the Docker daemon is stopped.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -194,6 +194,8 @@ jobs:
     name: Docker-disabled Tests / ${{ matrix.os }}
     if: github.repository_owner == 'aws'
     runs-on: ${{ matrix.os }}
+    env:
+      SAM_CLI_DEV: "1"
     strategy:
       matrix:
         os:
@@ -211,7 +213,6 @@ jobs:
             3.7
       - name: Init samdev
         run: make init
-
       - name: Stop Docker Linux
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: sudo systemctl stop docker
@@ -219,7 +220,6 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' }}
         shell: pwsh
         run: stop-service docker
-
       - name: Check Docker not Running
         run: docker info
         id: run-docker-info
@@ -227,6 +227,5 @@ jobs:
       - name: Report failure
         if: steps.run-docker-info.outcome == 'success'
         run: exit 1
-
       - name: Run tests without Docker
-        run: SAM_CLI_DEV=1 pytest -vv tests/integration/buildcmd/test_build_cmd.py -k TestBuildCommand_PythonFunctions_WithoutDocker
+        run: pytest -vv tests/integration/buildcmd/test_build_cmd.py -k TestBuildCommand_PythonFunctions_WithoutDocker
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,12 +205,13 @@ jobs:
         with:
           # These are the versions of Python that correspond to the supported Lambda runtimes
           python-version: |
-            3.7
-            3.8
-            3.9
             3.10
+            3.9
+            3.8
+            3.7
       - name: Init samdev
         run: make init
+
       - name: Stop Docker Linux
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: sudo systemctl stop docker
@@ -218,6 +219,7 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' }}
         shell: pwsh
         run: stop-service docker
+
       - name: Check Docker not Running
         run: docker info
         id: run-docker-info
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -187,3 +187,19 @@ jobs:
...
+
+  docker-disabled:
+    name: docker-disabled-test
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.7" # The integration test below needs python3.7 be to configured on the PATH
```
sriram-mv, line 199: aside: need to update the python version on the templates in general.

mildaniel: I'll update to a newer version, just want to figure out the GHA stuff first 🙃

.github/workflows/build.yml
```diff
@@ -187,3 +189,30 @@ jobs:
...
+        run: sudo systemctl stop docker
+      - name: Check Docker not Running
+        run: docker info
+        id: run-docker-info
+        continue-on-error: true
+      - name: Report failure
+        if: steps.run-docker-info.outcome == 'success'
+        run: exit 1
+      - name: Run tests without Docker
+        run: pytest -vv tests/integration/buildcmd/test_build_cmd.py -k TestBuildCommand_PythonFunctions_WithoutDocker
```
jfuss, line 218: 1. Should this be a make command at all?
2. Should we be running this with `SAM_CLI_DEV=1` like https://github.com/aws/aws-sam-cli/blob/develop/Makefile#L20?

.github/workflows/build.yml
```diff
@@ -187,3 +189,30 @@ jobs:
...
+
+  docker-disabled:
+    name: docker-disabled-test
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: |
```
jfuss, line 201: Can we comment this is runtime supported versions not SAM CLI installation versions?

Also worth making sure our "adding/removing runtime support" docs are updated so we remember this needs to be updated.

jfuss (changes requested):

.github/workflows/build.yml
mildaniel: 1. Not too sure what you mean, it's not currently a make command.
2. 👍

lucashuy (approved):

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -398,33 +398,64 @@ def _validate_skipped_built_function(
...
+    overrides = True
+    runtime = "python3.9"
+    codeuri = "Python"
+    use_container = "use_container"
     check_function_only = False
 
     @pytest.mark.flaky(reruns=3)
     def test_with_default_requirements(self):
-        if self.use_container and (SKIP_DOCKER_TESTS or SKIP_DOCKER_BUILD):
+        if SKIP_DOCKER_TESTS or SKIP_DOCKER_BUILD:
```
mndeveci, line 458: If the whole class would be skipped (since all container related tests moved here), we can add `@skip` at the top of class definition which will also not select these cases anymore.

.github/workflows/build.yml
jfuss: On one: I was asking if this should be a make command. Not a big deal.

sriram-mv (approved):

.github/workflows/build.yml
mildaniel: Gotcha. Don't think it needs to be. We have other tests in our actions we run directly through pytest. Don't think we need to abstract them all with make.

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -398,34 +398,64 @@ def _validate_skipped_built_function(
...
+        ("template.yaml", "Function", True, "python3.7", "Python", False, "CodeUri"),
+        ("template.yaml", "Function", True, "python3.8", "Python", False, "CodeUri"),
+        ("template.yaml", "Function", True, "python3.9", "Python", False, "CodeUri"),
+        ("template.yaml", "Function", True, "python3.10", "Python", False, "CodeUri"),
+        ("template.yaml", "Function", True, "python3.7", "PythonPEP600", False, "CodeUri"),
+        ("template.yaml", "Function", True, "python3.8", "PythonPEP600", False, "CodeUri"),
     ],
 )
-class TestBuildCommand_PythonFunctions(BuildIntegPythonBase):
+class TestBuildCommand_PythonFunctions_WithoutDocker(BuildIntegPythonBase):
```
super132, line 413: Do we skip all the builds tests without docker?

mildaniel: These are tests that don't use Docker, so we don't need to skip them.

PR #5291, opened 2023-06-07
feat: Added OpenApi body integration testing and updated property builder

lucashuy:

N/A

To test OpenApi based Apis properly, they need to be deployed first. We'll need to add some setup to run `terraform apply` to deploy the project.

Additionally, the Api body is a string dump of a JSON object. This needs to be parsed before it gets handed off to the `local start-api` logic so that it ends up getting parsed correctly.

Adds a new base class that is capable of running the setup and tear down to deploy a Terraform project. Also updates the property mapping for the API Gateway to instead load the string as a JSON object.

The API Gateway property mapping for `Body` now gets the JSON object, instead of a string.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/hook_packages/terraform/hooks/prepare/property_builder.py
```diff
@@ -234,7 +268,7 @@ def _check_image_config_value(image_config: Any) -> bool:
 
 AWS_API_GATEWAY_REST_API_PROPERTY_BUILDER_MAPPING: PropertyBuilderMapping = {
     "Name": _get_property_extractor("name"),
-    "Body": _get_property_extractor("body"),
+    "Body": _get_json_body,
```
moelasmar, line 271: good catch

tests/integration/local/start_api/test_start_api_with_terraform_application.py
```diff
@@ -54,3 +97,38 @@ def test_successful_request(self):
...
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json(), {"message": "hello world"})
+
+
+@skipIf(
+    not CI_OVERRIDE,
+    "Skip Terraform test cases unless running in CI",
+)
+@pytest.mark.flaky(reruns=3)
+class TestStartApiTerraformApplicationV1OpenApiAuthorizer(TerraformStartApiIntegrationApplyBase):
```
moelasmar, line 107: I believe it is better to avoid using V1 on our names, I believe for V2 testing we can just modify the current projects and add some V2 resources for testing.

moelasmar (approved):
it looks good to me, I just left minor comment.

sriram-mv (approved):

PR #5294, opened 2023-06-07
GraphQLApi support for `sam deploy`

ssenchenko:
Similar to [this PR](https://github.com/aws/aws-sam-cli/pull/5250) for `sam package`, `sam build` has to support updating relative code paths for `GraphQLApi` resource with multiple properties (in different sub-dictionaries) with the same name `CodeUri`.

I also did a bit of refactoring, moving the function which traverses `GraphQLApi` into `lib/utils` out of the `GraphQLApiCodeResource` class.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/package/packageable_resources.py
```diff
@@ -599,7 +600,7 @@ def get_property_value(self, resource_dict):
 
 class GraphQLApiSchemaResource(ResourceZip):
     RESOURCE_TYPE = AWS_SERVERLESS_GRAPHQLAPI
-    PROPERTY_NAME = RESOURCES_WITH_LOCAL_PATHS[RESOURCE_TYPE][0]
+    PROPERTY_NAME = graphql_api.SCHEMA_ARTIFACT_PROPERTY
```
mildaniel, line 603: nit: Can we import anything we need from `graphql_api` directly instead of importing the module?

mildaniel (commented):
Thanks Slava, LGTM! Can we add an integration test case, in this or separate PR?

mildaniel (approved):

samcli/lib/package/packageable_resources.py
ssenchenko: Is the style I used is against SAM CLI practice? I import everything from that file anyway. But the main reason, I think that `graphql_api.SCHEMA_ARTIFACT_PROPERTY` and `graphql_api.find_all_paths_and_values(self.PROPERTY_NAME, resource_dict)` is easier to read then simply `SCHEMA_ARTIFACT_PROPERTY` and `find_all_paths_and_values` which show less "belonging". But if it's against the style, I can remove it. It's up to you

hawflau: I think it's fine since `graphql_api` is not a huge module. And +1 for the point of "belonging".

hawflau (approved):

PR #5295, opened 2023-06-07
fix: Update Arn parsing logic and fix some edge cases/bug fixes for remote invoke

hnnasit:
The Arn format for Lambda functions [arn:aws:lambda:us-west-2:123456789012:function:my-function](https://docs.aws.amazon.com/lambda/latest/dg/lambda-api-permissions-ref.html) is not captured with the current logic. The new changes matches the Arn based on the formats provided [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html). This PR also changes how the ECR Arn was parsed for sam pipeline where `repository/${Repository-name}` was considered as the `resource-id`. 

For the case when `sam cloud invoke` is run with `--stack-name` option and the stack does not exist, it throws a traceback error. Added an error check to throw `UserException`.

Rename command from `cloud` to `remote` and update file/variable names.

Described above

No side effects, integration tests were run for `sam pipeline` locally

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/arn_utils.py
```diff
@@ -31,11 +32,47 @@ class ARNParts:
...
+            # https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html
+            arn_pattern = (
+                r"arn:([a-zA-Z0-9_-]+):"  # Pattern for partition
+                r"([a-zA-Z0-9_-]+):"  # Pattern for service
+                r"([a-zA-Z0-9_-]*):"  # Pattern for region
+                r"([a-zA-Z0-9_-]*)"  # Pattern for account_id
+                r"(?::([a-zA-Z0-9_-]+))?"  # Pattern for resource_type if it exists
+                r"(?::(.+))"  # Pattern for resource_id if it exists
+            )
+            print(arn_pattern)
```
mndeveci, line 50: left over debugging?

samcli/lib/utils/arn_utils.py
```diff
@@ -31,11 +32,47 @@ class ARNParts:
...
+                split_resource_type_and_id = matched_arn.group(6).split("/", 1)
+                self.resource_type = split_resource_type_and_id[0]
+                self.resource_id = split_resource_type_and_id[1]
+            else:
+                # This handles the Arns of services with the format:
+                # arn:partition:service:region:account-id:resource-id
+                self.resource_type = ""
+                self.resource_id = matched_arn.group(6) if matched_arn.group(6) else ""
+
         except ValueError as ex:
```
mndeveci, line 77: Is this exception valid anymore, or should this be replaced something else? I am not sure what will happen if a random string is passed here?

samcli/commands/cloud/remote_invoke_context.py
```diff
@@ -105,20 +108,28 @@ def _populate_resource_summary(self) -> None:
...
+            if not self._stack_name:
+                # no stack name provided, resource id should be physical id so that we can use it
+                self._resource_summary = self._get_from_physical_resource_id()
+                return
+
+            self._resource_summary = get_resource_summary(
+                self._boto_resource_provider, self._boto_client_provider, self._stack_name, self._resource_id
+            )
+        except ClientError as ex:
+            error_code = get_client_error_code(ex)
```
sriram-mv, line 127: do we care about any other errors?

sriram-mv (commented):
Looks good otherwise apart from existing comments.

samcli/lib/utils/arn_utils.py
hnnasit: That's True, we can raise the `InvalidArnValue` on line 53 instead and remove line the try/except here. This will be caught in remote_invoke_context.py and a `UserException` will be thrown with the message `Can't find exact resource information with given {self._resource_id}. "
f"Please provide full resource ARN or --stack-name to resolve the ambiguity.`

samcli/commands/cloud/remote_invoke_context.py
hnnasit: I think `ValidationError` should be enough here.

mndeveci (approved):

lucashuy (approved):

PR #5306, opened 2023-06-08
chore: cleanup appveyor definitions for not running jobs which is already run with GHA & add docker info/version commands

mndeveci:
N/A

Some of the checks that we do in GHA is also run with Appveyor jobs and they have custom configuration rather than using the ones from `Makefile`. 

- Removing the tests/checks from Appveyor that we already run with GHA
- Cleaning up some Appveyor setup where it was repeated couple of times

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -61,7 +61,8 @@ install:
 
   - sh: "source ${HOME}/venv${PYTHON_VERSION}/bin/activate"
   - sh: "rvm use 2.7.2"
-  - sh: "docker --version"
+  - sh: "docker info"
+  - sh: "docker version"
   - sh: "nvm install ${NODE_VERSION}"
   - sh: "npm install npm@7.24.2 -g"
   - sh: "npm -v"
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -149,7 +149,8 @@ install:
 
 
 build_script:
-      - "python -c \"import sys; print(sys.executable)\""
+    - "python -c \"import sys; print(sys.executable)\""
+    - "make init"
 
 # Final clean up no matter success or failure
 on_finish:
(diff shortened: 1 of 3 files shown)


appveyor-windows.yml
```diff
@@ -135,6 +136,10 @@ install:
 
   - "IF DEFINED BY_CANARY ECHO Logging in Public ECR && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws"
 
+build_script:
+  - "docker system prune -a -f"
```
super132, line 140: Is it a Windows specific step? Why do we need to prune on Windows?

appveyor-ubuntu.yml
```diff
@@ -206,10 +200,6 @@ for:
         - configuration: LocalZipTerraformBuildInContainerIntegTesting
 
     test_script:
-      # Pre-dev Tests
-      - "pip install -e \".[pre-dev]\""
```
jfuss, line 210: I think we need this or we need to move `ruff` into the `requirements/dev.txt`. I don't know why we have a "pre-dev" file.

jfuss (changes requested):

sriram-mv:
Assuming that linting is covered through `Makefile` targets being invoked with this removal.

appveyor-ubuntu.yml
mndeveci: As far as I understand that `pre-dev.txt` is imported at the top of the `dev.txt` https://github.com/aws/aws-sam-cli/blob/develop/requirements/dev.txt#L1

appveyor-windows.yml
super132: Do we need to do the same for Ubuntu then?

appveyor-ubuntu.yml
jfuss: Well would you look at that... lol

jfuss (approved):

sriram-mv (approved):

PR #5309, opened 2023-06-08
test: Test the unsupported limitations for local start api

moelasmar:
Adding integration test cases to test the unsupported limitations while running `sam local start-api` , and testing them after running `terraform apply`

#### Mandatory Checklist

- [ ] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


tests/integration/local/start_api/test_start_api_with_terraform_application.py
```diff
@@ -99,6 +108,89 @@ def test_successful_request(self):
...
+        {
+            "terraform_application": "terraform-api-simple-local-variables-limitation",
+            "expected_error_message": "Error: AWS SAM CLI could not process a Terraform project that uses local "
+            "variables to define linked resources.",
+        },
+    ]
+)
+class TestStartApiTerraformApplicationLimitations(TerraformStartApiIntegrationBase):
+    @classmethod
+    def setUpClass(cls):
```
lucashuy, line 137: question: how come we are setting up the class here again?

moelasmar: Sorry I did not get your question. I am overriding the setupClass function defined in parent class. This will be called instead of the one in the parent class

moelasmar: I do not want to call 'sam local start-api' in the setup class, as it is supposed to fail. So I override this method to skip executing sam local command

lucashuy: That makes sense, since the parent will fail before it ever gets to the test, thanks!

tests/integration/local/start_api/test_start_api_with_terraform_application.py
```diff
@@ -99,6 +108,89 @@ def test_successful_request(self):
...
+        cls._remove_generated_directories()
+
+    def test_unsupported_limitations(self):
+        apply_disclaimer_message = "Unresolvable attributes discovered in project, run terraform apply to resolve them."
+
+        process = self._run_command(self.command_list, check=False)
+
+        LOG.info(process.stderr)
+        output = process.stderr.decode("utf-8")
+        self.assertNotEqual(process.returncode, 0)
```
mildaniel, line 163: nit: Should we assert an exit code of 1 instead of not 0? I don't think we want any of these to exit on 255, right?

mildaniel (approved):
Thanks Mohamed, LGTM. Left one non-blocking question.

tests/integration/local/start_api/test_start_api_with_terraform_application.py
moelasmar: yes, let me update it

moelasmar: done

PR #5315, opened 2023-06-09
feat: Track config file extension

Leo10Gama:
These changes allow us to keep track of which config file extensions users are using. As such, it will provide us insights into how many people may prefer to use one file type over another, which could aid in the potential deprecation of less-used extensions, or the potential to support newer ones.

A new `EventType` has been created for `EventName.CONFIG_FILE_EXTENSION`, which is called upon the creation of a `SamConfig` object. As such, whenever someone runs a command that reads in a config file, we will receive an Event.

In order to prevent a cyclic import, the `FILE_MANAGER_MAPPER` has been moved to `samconfig/file_manager.py` from `samconfig/samconfig.py`. This change makes sense, as the logic of pairing file extensions to `FileManager`s is encapsulated in this file anyway, and it reduces the number of imports made by `samconfig.py`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/config/samconfig.py
+++ b/samcli/lib/config/samconfig.py
@@ -5,11 +5,12 @@ Class representing the samconfig.toml
 import logging
 import os
 from pathlib import Path
-from typing import Any, Dict, Iterable, Type
+from typing import Any, Iterable
 
 from samcli.lib.config.exceptions import FileParseException, SamConfigFileReadException, SamConfigVersionException
-from samcli.lib.config.file_manager import FileManager, JsonFileManager, TomlFileManager, YamlFileManager
+from samcli.lib.config.file_manager import FILE_MANAGER_MAPPER
 from samcli.lib.config.version import SAM_CONFIG_VERSION, VERSION_KEY
+from samcli.lib.telemetry.event import EventTracker
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


tests/integration/telemetry/test_installed_metric.py
```diff
@@ -24,7 +24,9 @@ def test_send_installed_metric_on_first_run(self):
             self.assertIn(EXPECTED_TELEMETRY_PROMPT, stderrdata.decode())
 
             all_requests = server.get_all_requests()
-            self.assertEqual(2, len(all_requests), "There should be exactly two metrics request")
+            self.assertEqual(
+                3, len(all_requests), "There should be exactly two metrics request"
```
lucashuy, line 28: nit: three metrics

samcli/lib/config/samconfig.py
```diff
@@ -46,16 +40,18 @@ def __init__(self, config_dir, filename=None):
...
-                f"The config file extension '{self.filepath.suffix}' is not supported. "
-                f"Supported formats are: [{'|'.join(self.FILE_MANAGER_MAPPER.keys())}]"
+                f"The config file extension '{file_extension}' is not supported. "
+                f"Supported formats are: [{'|'.join(FILE_MANAGER_MAPPER.keys())}]"
             )
             raise SamConfigFileReadException(
                 f"The config file {self.filepath} uses an unsupported extension, and cannot be read."
             )
         self._read()
+        EventTracker.track_event("SamConfigFileExtension", file_extension)
```
lucashuy, line 54: Could we add a simple unit test to check that the expected file extension with that event name is passed in?

lucashuy (commented):
Thanks for making the changes thus far, have one more comment to leave about testing!

hnnasit (approved):
Overall looks good to me!

lucashuy (approved):

PR #5359, opened 2023-06-13
refactor: make remote invoke reactive to display results as soon as t…

mndeveci:
…hey are available

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/remote/remote_invoke_context.py
+++ b/samcli/commands/remote/remote_invoke_context.py
@@ -2,6 +2,7 @@
 Context object used by `sam remote invoke` command
 """
 import logging
+from dataclasses import dataclass
 from typing import Optional, cast
 
 from botocore.exceptions import ClientError
@@ -15,7 +16,12 @@ from samcli.commands.remote.exceptions import (
     UnsupportedServiceForRemoteInvoke,
 )
 from samcli.lib.remote_invoke.remote_invoke_executor_factory import RemoteInvokeExecutorFactory
-from samcli.lib.remote_invoke.remote_invoke_executors import RemoteInvokeExecutionInfo
+from samcli.lib.remote_invoke.remote_invoke_executors import (
+    RemoteInvokeConsumer,
+    RemoteInvokeExecutionInfo,
+    RemoteInvokeLogOutput,
+    RemoteInvokeResponse,
+)
 from samcli.lib.utils import osutils
 from samcli.lib.utils.arn_utils import ARNParts, InvalidArnValue
 from samcli.lib.utils.boto_utils import BotoProviderType, get_client_error_code
@@ -61,7 +67,7 @@ class RemoteInvokeContext:
     def __exit__(self, *args) -> None:
         pass
 
-    def run(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> RemoteInvokeExecutionInfo:
+    def run(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> None:
         """
         Instantiates remote invoke executor with populated resource summary information, executes it with the provided
         input & returns its response back to the caller. If no executor can be instantiated it raises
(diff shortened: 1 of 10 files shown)


samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -86,41 +94,60 @@ def _execute_action(self, payload: str):
...
         )
-        return cast(dict, self._lambda_client.invoke(**self.request_parameters))
+        lambda_response = self._execute_boto_call(self._lambda_client.invoke)
+        if self._remote_output_format == RemoteInvokeOutputFormat.RAW:
+            yield RemoteInvokeResponse(lambda_response)
+        if self._remote_output_format == RemoteInvokeOutputFormat.DEFAULT:
+            yield RemoteInvokeResponse(cast(StreamingBody, lambda_response.get(PAYLOAD)).read().decode("utf-8"))
+            log_result = lambda_response.get(LOG_RESULT)
+            if log_result:
+                yield RemoteInvokeLogOutput(base64.b64decode(log_result).decode("utf-8"))
```
mildaniel, line 120: Feels a little weird to me that these two cases are handled differently here when we've already created different formatters to handle the different output types. I guess I'm wondering how come the outputs themselves are different and not just the format of the outputs?

mildaniel (approved):
LGTM! Left a question about the different outputs.

samcli/commands/remote/remote_invoke_context.py
```diff
@@ -61,7 +67,7 @@ def __enter__(self) -> "RemoteInvokeContext":
     def __exit__(self, *args) -> None:
         pass
 
-    def run(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> RemoteInvokeExecutionInfo:
+    def run(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> None:
```
hnnasit, line 70: nit: The doc string needs to be updates as this function doesn't returns None.

samcli/lib/remote_invoke/lambda_invoke_executors.py
```diff
@@ -86,41 +94,60 @@ def _execute_action(self, payload: str):
...
             self.request_parameters,
         )
-        return cast(dict, self._lambda_client.invoke(**self.request_parameters))
+        lambda_response = self._execute_boto_call(self._lambda_client.invoke)
+        if self._remote_output_format == RemoteInvokeOutputFormat.RAW:
+            yield RemoteInvokeResponse(lambda_response)
+        if self._remote_output_format == RemoteInvokeOutputFormat.DEFAULT:
+            yield RemoteInvokeResponse(cast(StreamingBody, lambda_response.get(PAYLOAD)).read().decode("utf-8"))
+            log_result = lambda_response.get(LOG_RESULT)
+            if log_result:
```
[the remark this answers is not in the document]
mndeveci: We can flush the streams once the event is consumed there, and we can also yield the log event first and then the payload, I will update the PR with this. But technically, they should be utilizing stdout and stderr streams differently to get their expected output.

samcli/lib/remote_invoke/lambda_invoke_executors.py
mndeveci: If you mean why a log event is different then the response, is because they are written into different streams. Actual response (aka Payload) is written into `stdout` where log events are printed into `sdterr`, so that customers can pipe them to their other command line utilities. That is why we have two different events to distinguish them so that each can be consumed by a different consumer.

samcli/lib/remote_invoke/lambda_invoke_executors.py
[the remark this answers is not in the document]
mndeveci: Btw we can do this for regular invoke, but it is not possible for streaming as log result received as last chunk.

hnnasit (approved):

samcli/lib/remote_invoke/lambda_invoke_executors.py
mildaniel: My question was more about the output itself. It seems a little strange to me that the output for RAW and DEFAULT is different, it's not just the format that changes between the two output modes.

hnnasit: The DEFAULT output format is a more readable output format that prints the returned payload and LogResult if any. The RAW output format prints the full API boto response for users to parse.

Issue #3621, opened 2022-02-03
sam delete needs to care about profile defined in the samconfig


organom:
### Description:
With an environment defined in the samconfig.saml, that requires a profile different from the default aws profile, running `sam delete --config-env ENV_FROM_SAMCONFIG --stack-name STACK_TO_DELETE` will fail because the profile is not taken into account.
If running with `AWS_PROFILE=XXX` before the above sam delete command, the environment is found and sam can proceed.
`sam deploy` behaves correctly in the system.

### Steps to reproduce:
On macOS Monterey 12.2  with sam installed over brew (v 1.37.0)
Use an environment from samconfig.saml that requires a profile (different from the AWS default profile)
Run `sam delete --config-env ENV_FROM_SAMCONFIG --stack-name STACK_TO_DELETE`

### Observed result:
`Error: The input stack STACK_TO_DELETE does not exist on Cloudformation in the region REGION_FROM_ENV`

### Expected result:
Same behavior as when adding AWS_PROFILE=XXXX  before running the sam delete command.

terusaku:
Hello.
Probably, almost same issue happened to me.
When `sam delete` was executed with `--profile` and no samconfig.toml, region variable seems to be loaded implicitly without profile.

### pre-requirement:
- SAM CLI, version 1.72.0
- no `samconfig.toml`.
- executing `sam deploy/delete` command with `--profile` explicitly
- my profile is below.
```
    [profile poweruse]
    sso_start_url = https://d-###.awsapps.com/start
    sso_region = ap-northeast-1
    sso_account_id = #####
    sso_role_name = PowerUserAccess
    region = ap-northeast-1
    output = json
```

### reproduce detail:
- `sam deploy` works with `--profile` option,  but `sam delete` with `--profile` cause missing of region variable defined in ~/.aws/config.
    - Stack delete is always executing on us-east-1.
- command example: `sam deploy`
```
$ sam deploy --stack-name ${STACK} --s3-bucket ${SAM_BUCKET} --s3-prefix ${STACK} --capabilities CAPABILITY_IAM --role-arn ${DEPLOY_ROLE} --profile poweruse
```
- command example: `sam delete`
```
$ sam delete --stack-name ${STACK} --s3-bucket ${SAM_BUCKET} --s3-prefix ${STACK} --profile poweruse --debug             
```
- when deleting, debug stdout is below.
    - if added `--region` options, above `sam delete` was succesfully completed.
```
[...]

PR #5360, opened 2023-06-13
refactor init_clients in sam delete

hawflau:
#5248 #3621 

See the issues above for the observed behaviors.
Looking into the code, it seems like the `init_clients` method repeatedly set region and profile into the command context, which is already done from the root command, which then would update the boto3 default session  (see https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/context.py#L88-L106 and https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/context.py#L196). In fact, creating the boto3 clients will fail if region or credentials is invalid. We can just make use of this fact to reduce unnecessary checking. In Python, it is common and ["easier to ask for forgiveness than permissions"](https://docs.python.org/3/glossary.html#term-EAFP)

- remove resetting region and profile into command context
- just init the clients and catch exceptions if raised
- remove defaulting region to us-east-1 in "prompt" mode (IMO we should not default a region)

Not that I know of

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/delete/delete_context.py
```diff
@@ -108,38 +109,27 @@ def init_clients(self):
...
-        cloudformation_client = boto3.client(
-            "cloudformation", region_name=self.region if self.region else None, config=boto_config
-        )
-
-        s3_client = boto3.client("s3", region_name=self.region if self.region else None, config=boto_config)
-        ecr_client = boto3.client("ecr", region_name=self.region if self.region else None, config=boto_config)
+        try:
+            cloudformation_client = boto3.client("cloudformation", config=boto_config)
+            s3_client = boto3.client("s3", config=boto_config)
+            ecr_client = boto3.client("ecr", config=boto_config)
```
mndeveci, line 117: There is a utility method which can generate the client with default configuration, I would recommend to use that to streamline the client generation https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/boto_utils.py#L65

samcli/commands/delete/delete_context.py
```diff
@@ -108,38 +109,27 @@ def init_clients(self):
         """
         Initialize all the clients being used by sam delete.
         """
-        if not self.region:
-            if not self.no_prompts:
-                session = boto3.Session()
-                region = session.region_name
-                self.region = region if region else "us-east-1"
```
mndeveci, line 115: If customer is running this command with `no_prompts` then we are setting region to `us-east-1` by default, is it OK to remove it with this change?

mndeveci (changes requested):

samcli/commands/delete/delete_context.py
hawflau: My question is why should we default to us-east-1? That doesn't sound reasonable to me. 🤔

samcli/commands/delete/delete_context.py
hawflau: Via `_refresh_session` on this [line](https://github.com/aws/aws-sam-cli/blob/develop/samcli/cli/context.py#L196) we have configured boto3 to update its default session with the provided region and profile. This results in `boto3.client` would use the new default Session. 

On the other hand, `get_boto_client_provider_with_config` in boto_util.py requires me to pass in the region and the profile again. I don't have the context of why `get_boto_client_provider_with_config` was created. It seems to me that `_refresh_session` is enough and is a better mechanism to make sure SAM CLI honors `--region` and `--profile`. Because I don't need another mechanism to ensure other contributors to use `get_boto_client_provider_with_config`.

Happy to discuss further. Maybe I have missed something that `_refresh_session` was insufficient.

samcli/commands/delete/delete_context.py
mndeveci: I do agree this is different than other commands, my vote would be keeping the commands in the same way but since we released `sam delete` with this default, I am not sure if we need to make an announcement or need to keep the backwards compatibility.

samcli/commands/delete/delete_context.py
hawflau: After a discussion, we agreed to switch to use `get_boto_client_provider_with_config` for the reason that it's more verbose and doesn't require deep knowledge about how region and profile are set into the default session.

samcli/commands/delete/delete_context.py
```diff
@@ -108,38 +108,29 @@ def init_clients(self):
...
-        if self.region:
-            Context.get_current_context().region = self.region
-
-        boto_config = get_boto_config_with_user_agent()
-
-        # Define cf_client based on the region as different regions can have same stack-names
-        cloudformation_client = boto3.client(
-            "cloudformation", region_name=self.region if self.region else None, config=boto_config
+        client_provider = get_boto_client_provider_with_config(
+            region=self.region if self.region else None, profile=self.profile if self.profile else None
```
mndeveci, line 112: nit: assuming these fields are instantiated as `None` default, we may simplify this line (not sure if that has some side effects)
```suggestion
            region=self.region, profile=self.profile
```

mndeveci (approved):

Issue #5285, opened 2023-06-07
Feature request: Optional flag to disable authorizer for "sam local start-api"


mikepianka:
### Describe your idea/feature/enhancement

Being able to test API Gateway authorizers locally when running `sam local start-api` is a great new feature. However, I think it would be helpful to be able to pass an optional flag at the command line to disable the authorizer for local testing.

### Proposal

Add an optional flag to disable API Gateway authorizers like `sam local start-api --disable-authorizer`

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)

mildaniel:
Thanks for the suggestion @mikepianka, I like the idea! We welcome contributions if you'd like tackle this yourself, if not, I'll take it back to the team for prioritization.

hnnasit:
Closing as the PR has been merged and released.

PR #5364, opened 2023-06-14
add --disable-authorizer flag to local start-api

squizlee:

This is a possible implementation of #5285. I haven't written unit tests, just need confirmation on go ahead before writing them. Leaving in draft for now :) 

Get past having to generate an auth token everytime I wish to test my api locally. 

Add flag and if statements. Ensure the extraction of authorizers is ignored at the parsing level before mounting the local server.

None that I can see.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


lucashuy:
Hi, thanks for opening this issue! Just a general question, but have we consider disabling the parsing of authorizers too? Authorizers are first read/parsed from the template, then they are invoked as part of the Flask route handler.

lucashuy:
Hi, just checking in. Did you have time to contribute to this new command line option? Just wanted to address the comment above regarding when to disable before we continued

squizlee:
@lucashuy hey mate, I'll have a look into the parsing logic and see if we can disable there. Apologies, I was on a trip

squizlee:
@lucashuy going through the parsing flow, it looks like I would have to go to each of the providers and add conditional branches there to ignore the adding of authorizers. The parsing stage only bulk reads the yaml file using `yaml_parse()` in `template.py`. Seems simpler to skip the invocation of the authorizer after parsing. What are your thoughts/preferences?

lucashuy:
Yup, thats correct that if we had wanted to disable the authorizers at the parsing level, something would have to be passed down to the providers to get them to not read authorizers.

While skipping the invocation without skipping parsing might be more simple to implement (as you've already done), it might flow better in terms of the overall UX if we skipped parsing too. If a user were to have an incomplete authorizer definition, passing `--disable-authorizer` while still parsing would result in validation errors.

squizlee:
Great points, I'll implement changes as described - killing it at parsing

lucashuy:
Hi, just a friendly check in to see if there was any blockers or questions you might have regarding this change.

squizlee:
Hey mate, I've pushed through a look at what the implementation looks like when disabling at parser level. Unit tests can commence once I get your thoughts on it

samcli/commands/local/start_api/cli.py
```diff
@@ -61,6 +61,12 @@
     default="public",
     help="Any static assets (e.g. CSS/Javascript/HTML) files located in this directory " "will be presented at /",
 )
+@click.option(
+    "--disable-authorizer",
+    is_flag=True,
+    default=False,
+    help="Disable custom authorizer from being invoked. This is useful for local testing."
```
lucashuy, line 68:
```suggestion
    help="Disable custom Lambda Authorizers from being parsed and invoked."
```
Thoughts about leaving the local testing line out? Mainly since this is `local start-api`, so "local testing of xyz" is sort of what is expected out of running this command :)

samcli/commands/local/lib/local_api_service.py
```diff
@@ -63,6 +63,7 @@ def start(self):
             port=self.port,
             host=self.host,
             stderr=self.stderr_stream,
+            disable_authorizer=self.disable_authorizer
```
lucashuy, line 66: Is `self.disable_authorizer` defined? It looks like it was passed into the constructor, but wasn't saved anywhere.

samcli/lib/providers/cfn_base_api_provider.py
```diff
@@ -81,16 +82,16 @@ def extract_swagger_route(
...
         routes = parser.get_routes(event_type)
-
         LOG.debug("Found '%s' APIs in resource '%s'", len(routes), logical_id)
-        LOG.debug("Found '%s' authorizers in resource '%s'", len(authorizers), logical_id)
 
         collector.add_routes(logical_id, routes)
-        collector.add_authorizers(logical_id, authorizers)
 
         if default_authorizer:
             collector.set_default_authorizer(logical_id, default_authorizer)
```
lucashuy, line 97: `default_authorizer` is undefined here if the above `if not disable_authorizer` statement evaluates to `False`.

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -268,6 +269,7 @@ def _extract_cloud_formation_method(
         logical_id: str,
         method_resource: Dict,
         collector: ApiCollector,
+        disable_authorizer: Optional[bool] = None
```
lucashuy, line 272: Default `False`?

samcli/lib/providers/sam_api_provider.py
```diff
@@ -114,13 +114,16 @@ def _extract_from_serverless_api(
...
+        CfnBaseApiProvider.extract_swagger_route(stack_path, logical_id, body, uri, binary_media, collector, cwd=cwd, disable_authorizer=disable_authorizer)
         collector.stage_name = stage_name
         collector.stage_variables = stage_variables
         collector.cors = cors
 
         auth = properties.get(SamApiProvider._AUTH, {})
-        if not auth:
+        if not auth or disable_authorizer:
+            LOG.debug(
+                "authorizer not found or disabled, returning early"
```
lucashuy, line 125: nit: `Authorizer` instead of `authorizer`

lucashuy (commented):
Thanks for taking the time to implement this! The approach looks good and tackles what we wanted. Left some early comments for consideration while tests are being worked on. Let us know if you needed any pointers as to where tests should go.

samcli/lib/providers/cfn_base_api_provider.py
squizlee: Great catch!

squizlee:
Thanks for the comments! I'll start with the unit and integration tests. Was there any other type of test you had in mind that I should touch?

lucashuy:
Namely just making sure that the existing unit tests pass (eg. updating any parameters that might need to be passed in). As for new unit tests, we'd probably want to test that if the disable authorizer flag was passed into a method, or inside a method that access the disable authorizer flag, the logic does not actually add an authorizer.

Normally when authorizers are detected, SAM CLI prints a disclaimer message in the console. We can use this fact to check if the message is printed or not when passing in the disable flag (it ideally shouldn't).

I'm going to move this PR to the review stage if that all sounds good to you :)
As always, let us know if you needed any help or pointers as to where things are located.

moelasmar (changes requested):
could you please run `make schema` to add the new parameter to our samconfig file schema.

moelasmar (dismissed):
in general the logic looks good to me, could you please update the unit testing to add testing when `disable-authorizer` is true, and make sure that there are no authorizers linked to the collected APIs. Also, add integration test case for using `--disable-authorizer` to make sure that SAM CLI is actually skipping running the authorization logic, and it directly call the lambda function

samcli/lib/providers/api_provider.py
```diff
@@ -13,7 +13,7 @@
 
 
 class ApiProvider(AbstractApiProvider):
-    def __init__(self, stacks: List[Stack], cwd: Optional[str] = None):
+    def __init__(self, stacks: List[Stack], cwd: Optional[str] = None, disable_authorizer: Optional[bool] = False):
```
moelasmar, line 16: could you update the function docs to add the new parameter!

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -49,7 +49,13 @@ class CfnApiProvider(CfnBaseApiProvider):
     _METHOD_AUTHORIZER_ID = "AuthorizerId"
     _ROUTE_AUTHORIZER_ID = "AuthorizerId"
 
-    def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: Optional[str] = None) -> None:
+    def extract_resources(
```
moelasmar, line 52: could you update the function docs to add the new parameter!

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -268,6 +295,7 @@ def _extract_cloud_formation_method(
         logical_id: str,
         method_resource: Dict,
         collector: ApiCollector,
+        disable_authorizer: Optional[bool] = False,
```
moelasmar, line 298: could you update the function docs to add the new parameter!

samcli/lib/providers/cfn_api_provider.py
```diff
@@ -400,6 +440,7 @@ def _extract_cfn_gateway_v2_route(
         logical_id: str,
         route_resource: Dict,
         collector: ApiCollector,
+        disable_authorizer: Optional[bool] = False,
```
moelasmar, line 443: could you update the function docs to add the new parameter!

samcli/lib/providers/sam_api_provider.py
```diff
@@ -46,7 +46,13 @@ class SamApiProvider(CfnBaseApiProvider):
     _API_IDENTITY_SOURCE_PREFIX = "method."
     _HTTP_IDENTITY_SOURCE_PREFIX = "$"
 
-    def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: Optional[str] = None) -> None:
+    def extract_resources(
```
moelasmar, line 49: could you update the function docs to add the new parameter!

samcli/lib/providers/sam_api_provider.py
```diff
@@ -69,14 +75,27 @@ def extract_resources(self, stacks: List[Stack], collector: ApiCollector, cwd: O
...
 
     def _extract_from_serverless_api(
-        self, stack_path: str, logical_id: str, api_resource: Dict, collector: ApiCollector, cwd: Optional[str] = None
+        self,
+        stack_path: str,
+        logical_id: str,
+        api_resource: Dict,
+        collector: ApiCollector,
+        cwd: Optional[str] = None,
+        disable_authorizer: Optional[bool] = False,
```
moelasmar, line 98: could you update the function docs to add the new parameter!

samcli/lib/providers/sam_api_provider.py
```diff
@@ -295,7 +317,13 @@ def _extract_authorizers_from_props(logical_id: str, auth: dict, collector: ApiC
...
 
     def _extract_from_serverless_http(
-        self, stack_path: str, logical_id: str, api_resource: Dict, collector: ApiCollector, cwd: Optional[str] = None
+        self,
+        stack_path: str,
+        logical_id: str,
+        api_resource: Dict,
+        collector: ApiCollector,
+        cwd: Optional[str] = None,
+        disable_authorizer: Optional[bool] = False,
```
moelasmar, line 326: could you update the function docs to add the new parameter!

moelasmar:
Thanks @squizlee for your contribution, please let us know if you need any help to update the code or add integration test cases.

squizlee:
Hi there! 

With the integration test is there a nice command to run an individual integration test file? I'm not sure how to verify locally without running the whole suite of tests.

For the unit tests, which test files should I go for? I want to make sure I'm touching the correct files.

mndeveci:
> Hi there!
> 
> With the integration test is there a nice command to run an individual integration test file? I'm not sure how to verify locally without running the whole suite of tests.

You can run individual integration tests by providing the folder/file that you want to run.
```
pytest -vv tests/integration/my/folder
```

```
pytest -vv tests/integration/my/file.py
```


> For the unit tests, which test files should I go for? I want to make sure I'm touching the correct files.

Our unit test files are following the original file path usually, you can look for test files for the ones that is been changed here. For example; `samcli/lib/delete/cfn_utils.py` -> `tests/unit/lib/delete/test_cfn_utils.py`.

squizlee:
I've added unit tests for cfn api provider, investigating the integration test in the coming days. Please let me know if there any smells with the tests as far.

samcli/lib/providers/api_provider.py
```diff
@@ -29,11 +29,14 @@ def __init__(self, stacks: List[Stack], cwd: Optional[str] = None):
             List of stacks apis are extracted from
         cwd : str
             Optional working directory with respect to which we will resolve relative path to Swagger file
+        disable_authorizer : str
```
lucashuy, line 32: nit: type `Optional[bool]` instead of `str`

tests/unit/commands/local/lib/test_cfn_api_provider.py
```diff
@@ -1186,6 +1186,86 @@ def test_empty_integration_array(self):
...
+                },
+            }
+        }
+        mock_get_route_function_name.return_value = ("Dummy", "Values")
+
+        provider = CfnApiProvider()
+        mock_collector = Mock()
+        mock_collector.add_routes = Mock()
+        provider._extract_cfn_gateway_v2_route(stack_path="stack_path", resources={}, logical_id="MyRoute", route_resource=template["Resources"]["MyRoute"], collector=mock_collector, disable_authorizer=True)
+        assert mock_collector.add_routes.call_args.args[1][0].authorizer_name is None
```
lucashuy, line 1213: nit: can we change the asserts to use `unittest`'s assertions? The method call to make here would be `<mocked object>.assert_called_with` (ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with

tests/unit/commands/local/lib/test_cfn_api_provider.py
```diff
@@ -1186,6 +1186,86 @@ def test_empty_integration_array(self):
...
+        }
+        provider = CfnApiProvider()
+        mock_collector = Mock()
+        mock_collector.add_routes = Mock()
+
+        provider._extract_cloud_formation_method(stack_path="", resources={}, logical_id="MyMethod", method_resource=template["MyMethod"], collector=mock_collector, disable_authorizer=True)
+        assert mock_collector.add_routes.call_args.args[1][0].authorizer_name is None
+        provider._extract_cloud_formation_method(stack_path="", resources={}, logical_id="MyMethod", method_resource=template["MyMethod"], collector=mock_collector, disable_authorizer=False)
+        assert mock_collector.add_routes.call_args.args[1][0].authorizer_name is "AuthorizerId"
+
```
lucashuy, line 1268: Similar comment as above (using `unittest`'s assertions). Could we also parameterize this test case? There are some examples on how to parameterize tests here: https://github.com/aws/aws-sam-cli/blob/09f5fdc3c299d68197853a8dca5e12b0044d60b7/tests/unit/local/apigw/test_lambda_authorizer.py

tests/unit/commands/local/lib/test_cfn_base_provider.py
```diff
@@ -0,0 +1,28 @@
...
+class TestDisableAuthorizerFlagInExtractSwaggerRoute(TestCase):
+
+    @patch("samcli.lib.providers.cfn_base_api_provider.LOG.debug")
+    @patch("samcli.lib.providers.cfn_base_api_provider.SwaggerParser")
+    @patch("samcli.lib.providers.cfn_base_api_provider.SwaggerReader")
+    def test_disable_authorizer_flag_when_enabled_skips_authorizer_extraction(self, mock_reader: Mock, mock_parser: Mock, mock_debug_log: Mock):
+        mock_collector = Mock()
+
+        CfnBaseApiProvider.extract_swagger_route(stack_path="", logical_id="", body={}, uri="", binary_media=None, collector=mock_collector, cwd="", event_type="HttpApi", disable_authorizer=True)
+        assert "Found '%s' authorizers in resource" not in mock_debug_log.call_args[0]
```
lucashuy, line 17: Could we update this test to instead check that the `mock_collector.add_authorizers` was not called? This might be a more concrete way of testing if the collection logic was not called instead of checking for log calls.

tests/unit/commands/local/lib/test_cfn_base_provider.py
```diff
@@ -0,0 +1,28 @@
...
+    @patch.object(SwaggerParser, "get_authorizers")
+    @patch("samcli.lib.providers.cfn_base_api_provider.LOG.debug")
+    @patch("samcli.lib.providers.cfn_base_api_provider.SwaggerParser")
+    @patch("samcli.lib.providers.cfn_base_api_provider.SwaggerReader")
+    def test_disable_authorizer_flag_when_disabled_extracts_authorizers(self, mock_reader: Mock, mock_parser: Mock, mock_debug_log: Mock, mock_get_authorizers: Mock):
+        mock_collector = Mock()
+        mock_get_authorizers.return_value = [{}, {}, {}]
+
+        CfnBaseApiProvider.extract_swagger_route(stack_path="", logical_id="", body={}, uri="", binary_media=None, collector=mock_collector, cwd="", event_type="HttpApi", disable_authorizer=False)
+        mock_debug_log.assert_any_call("Found '%s' authorizers in resource '%s'", 0, '')
```
lucashuy, line 28: Same comment about testing the collector vs logger

lucashuy (commented):
Thanks for adding tests to the PR! Left some comments about testing method calls

tests/unit/commands/local/lib/test_cfn_api_provider.py
squizlee: Doesn't seem like there is a clean way of asserting something was called with a partial matcher. I would have to create a Route object that matches exactly - thinking it's better to extract and assert the auth name than adding noise to the test. Thumbs this up if you still want me to create the route obj

squizlee:
Done another touch up of the tests, I think that covers all the changes. Integration test remaining...

mndeveci:
> Done another touch up of the tests, I think that covers all the changes. Integration test remaining...

Thanks for the updates @squizlee! You mean integration tests are done or are you planning to add more there?

squizlee:
> > Done another touch up of the tests, I think that covers all the changes. Integration test remaining...
> 
> Thanks for the updates @squizlee! You mean integration tests are done or are you planning to add more there?

Hi sorry for not being clearer! I've added the unit tests and an integration test as remarked above. Let me know if anything else is necessary please. I'll clean up any linting issues with `make pr` locally and get this ready for merge.

tests/unit/commands/local/lib/test_cfn_api_provider.py
lucashuy: Since the only part of that assertion that is changing is the name (which depends on if the disable flag was passed), this could be parameterized in a way where the test cases could be:
- `(True, None)`
- `(False, "AuthorizerId)`

It'll be a bit more setup code to get it to this state, but would potentially be easier to modify this test in the future if there were other checks that we wanted to make/behaviours that we wanted to updated.

tests/unit/commands/local/lib/test_cfn_api_provider.py
```diff
@@ -1187,6 +1187,127 @@ def test_empty_integration_array(self):
...
+        )
+
+        if disable_authorizer:
+            route = mock_collector.add_routes.call_args[0][1][0]
+            assert route.authorizer_name is None
+
+        if not disable_authorizer:
+            mock_collector.add_routes.assert_called()
+            route = mock_collector.add_routes.call_args[0][1][0]
+            assert route.authorizer_name is "AuthorizerId"
```
lucashuy, line 1234: Could we utilize `unittest`'s assertions here instead of using Python's `assert`?

tests/unit/commands/local/lib/test_cfn_api_provider.py
```diff
@@ -1187,6 +1187,127 @@ def test_empty_integration_array(self):
...
+        assert mock_collector.add_routes.call_args.args[1][0].authorizer_name is None
+        provider._extract_cloud_formation_method(
+            stack_path="",
+            resources={},
+            logical_id="MyMethod",
+            method_resource=template["MyMethod"],
+            collector=mock_collector,
+            disable_authorizer=False,
+        )
+        assert mock_collector.add_routes.call_args.args[1][0].authorizer_name is "AuthorizerId"
```
lucashuy, line 1308: Similar comment about using `unittest`'s asserts instead of Python's.

tests/unit/commands/local/lib/test_sam_api_provider.py
```diff
@@ -1872,6 +1872,111 @@ def test_extract_serverless_api_extracts_default_authorizer(
...
+            disable_authorizer=disable_authorizer,
+        )
+
+        if disable_authorizer:
+            mock_debug_log.assert_called_with("authorizer not found or disabled, returning early")
+            mock_collector.set_default_authorizer.assert_not_called()
+            mock_extract_authorizers_from_props.assert_not_called()
+
+        if not disable_authorizer:
+            mock_debug_log.assert_not_called()
```
lucashuy, line 1914: Thoughts about asserting log output? I think the previous comment I had noted that we could test the logic methods being called instead of debug logs that way we are testing the implementation.

tests/unit/commands/local/lib/test_sam_api_provider.py
```diff
@@ -1872,6 +1872,111 @@ def test_extract_serverless_api_extracts_default_authorizer(
...
+            disable_authorizer=disable_authorizer,
+        )
+
+        if disable_authorizer:
+            mock_debug_log.assert_called_with("Authorizer not found or disabled, returning early")
+            mock_collector.set_default_authorizer.assert_not_called()
+            mock_extract_authorizers_from_props.assert_not_called()
+
+        if not disable_authorizer:
+            mock_debug_log.assert_not_called()
```
lucashuy, line 1957: Same comment regarding the logs

tests/unit/commands/local/lib/test_sam_api_provider.py
```diff
@@ -1872,6 +1872,111 @@ def test_extract_serverless_api_extracts_default_authorizer(
...
+            "Auth": {"Authorizer": "Authorizer"},
+        }
+        _, route = SamApiProvider._convert_event_route(
+            stack_path="",
+            lambda_logical_id="MyLambda",
+            event_properties=event_properties,
+            event_type="Api",
+            disable_authorizer=disable_authorizer,
+        )
+        assert route.authorizer_name is expected
```
lucashuy, line 1978: Same comment about using `unittest`'s assertions.

lucashuy (commented):
Sorry about the delay in response, left some follow up comments

tests/unit/commands/local/lib/test_cfn_api_provider.py
squizlee: hey matey, do you mean self.assertEqual sort of assertions?

lucashuy: Yup, more so utilizing the `.assert_called_with(<args>)` method (ref: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with).

squizlee: is there a way to partial match a route object, or do I have create a route object to match exactly?

lucashuy: `unittest` provides an `ANY` object (https://docs.python.org/3/library/unittest.mock.html#any) that can be used in place of a specific argument check

squizlee: but what about for object properties e.g.  mock_collector.add_routes.assert_called_once_with(ANY, [mock_route(ANY, authorizer_name="AuthorizerId")])

doesn't seem to have the intended effect.

lucashuy: Ah I see, sorry for the initial misunderstanding. Getting the object with `call_args` should be fine in this case then, using `assert_called_once_with` would involve mocking out a `Route` object. We can go with your initial suggestion of `self.assertEqual`

squizlee:
@lucashuy resolved your comments, was there anything else that needed modifications before you guys can merge it?

lucashuy (approved):
Thanks for addressing the comments! I'll try to arrange for another reviewer to give your changes a look through. It just looks like `make format` needs to be run to fix one of the test files and the PR should be in a good shape.

squizlee:
Thanks you guys have been awesome, I've just pushed up make format

squizlee:
@lucashuy hey matey, anything else required for this to be merged in?

jysheng123 (approved):
Sorry for the late response, I looked over the changes and LGTM! The testing looks well covered here so I see no issues with it :)

PR #5374, opened 2023-06-16
Remove unneeded test cases

hawflau:

N/A

To remove two unneeded test cases due to PR #5360 

More context: changes in PR #5360 now requires region to be specified via `--region`, env var or aws profile. As such, unit test has been updated to make sure error with right messages is thrown in case no region or no creds can be resolved by boto3. Thus there is no need to keep the two integration test cases in this PR.

Remove two test cases

no

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hnnasit (approved):

lucashuy (approved):

sriram-mv:
@hawflau : could we be explicit and add more information in both commit message and PR. What? Why? how? and so on.

PR #5382, opened 2023-06-19
test: Integration tests for remote invoke on regular lambda functions

hnnasit:
Changes:
- Add remote invoke base class for integration tests
- Add integration tests for invoking  regular lambda functions

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/remote/invoke/remote_invoke_integ_base.py
```diff
@@ -0,0 +1,105 @@
...
+    RUNNING_ON_CI,
+    RUNNING_TEST_FOR_MASTER_ON_CI,
+    RUN_BY_CANARY,
+)
+from tests.integration.deploy.deploy_integ_base import DeployIntegBase
+
+from samcli.lib.utils.boto_utils import get_boto_resource_provider_with_config, get_boto_client_provider_with_config
+from samcli.lib.utils.cloudformation import get_resource_summaries
+
+SKIP_REMOTE_INVOKE_TESTS = RUNNING_ON_CI and RUNNING_TEST_FOR_MASTER_ON_CI and not RUN_BY_CANARY
```
moelasmar, line 17: why do we need to skip running the remote test cases in case of master branch?

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -0,0 +1,284 @@
...
+from parameterized import parameterized
+
+from tests.integration.remote.invoke.remote_invoke_integ_base import RemoteInvokeIntegBase
+from tests.testing_utils import run_command
+
+from pathlib import Path
+import pytest
+
+
+@pytest.mark.xdist_group(name="sam_remote_invoke_single_lambda_resource")
```
moelasmar, line 14: what is the value of this mark, are we going to run the remote invoke test cases in parallel?

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -0,0 +1,284 @@
...
+class TestSingleResourceInvoke(RemoteInvokeIntegBase):
+    template = Path("template-single-lambda.yaml")
+
+    @classmethod
+    def tearDownClass(cls):
```
moelasmar, line 19: shall we move the tear down method to be in the parent class?

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -0,0 +1,284 @@
...
+        # Delete the deployed stack
+        cls.cfn_client.delete_stack(StackName=cls.stack_name)
+
+    @classmethod
+    def setUpClass(cls):
```
moelasmar, line 24: also for the setup method, should it be moved to the parent class?

moelasmar (approved):
It looks good to me, I left some minor comments/questions

tests/integration/remote/invoke/test_remote_invoke.py
mndeveci: Any new tests added into "OtherTests" group will run the tests in parallel, by grouping them with the `xdist_group` above. With this added, pytest will make sure that the tests will be executed in the same group so that we can deploy -> run tests -> delete stack.

If this wasn't provided, the tests will be split into different workers, causing re-deployment of this stack and eventually will took longer time than it should.

mndeveci:
LGTM in general, left some clarifying questions.

tests/integration/remote/invoke/test_remote_invoke.py
hnnasit: I inherited this method for each child class to set the `stack-name` based on the class name. This might help debugging failed tests easier.

tests/integration/remote/invoke/remote_invoke_integ_base.py
hnnasit: Seems like this check might not be required anymore since we moved the AppVeyor tests that run on each PR to GH actions. I can remove this for `remote invoke` tests first and update the other commands in a separate PR.

mndeveci (approved):

PR #5386, opened 2023-06-20
fix: Handle BROKEN_PIPE_ERROR

lucashuy:
N/A

When a customer on Windows hits `control+c` to stop their `local start-api`, or `local start-lambda` commands, in `--warm-containers=EAGER` mode, they will encounter a `BROKEN_PIPE_ERROR`.

Adds a handler to capture this exception. Since this happens after all containers are deleted and cleaned up, we can safely ignore this exception.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/file_observer.py
```diff
@@ -264,15 +266,33 @@ def __init__(self, on_change: Callable) -> None:
         self._lock: Lock = threading.Lock()
 
     def _watch_images_events(self):
```
sriram-mv, line 268: Thought: should we add this exception handling as a decorator to the _watch_images_events func?

lucashuy: Hmm, that's a good call out. This would let us apply this handling elsewhere (if we needed to for some reason) and would look nicer since we would be wrapping a function call, instead of everything in a try catch.

samcli/lib/utils/file_observer.py
```diff
@@ -24,6 +25,7 @@
 from samcli.local.lambdafn.config import FunctionConfig
 
 LOG = logging.getLogger(__name__)
+BROKEN_PIPE_ERROR = 109
```
mndeveci, line 28: Can we add some comment around what is this and how it is been used below?

mndeveci (approved):

samcli/lib/utils/file_observer.py
```diff
@@ -263,6 +299,7 @@ def __init__(self, on_change: Callable) -> None:
         self._images_observer_thread: Optional[Thread] = None
         self._lock: Lock = threading.Lock()
 
+    @broken_pipe_handler
```
sriram-mv, line 302: Not going to block on this. Might be even better if we can inject the behavior into the decorator.

```python
@os_exception_handler(platform="windows", exception=BROKEN_PIPE_ERROR)
```

sriram-mv (approved):

samcli/lib/utils/file_observer.py
lucashuy: Interesting idea, though with this it might need a bit more rewriting/logic to be more generic. As of right now it has a pretty specific check for a particular error code vs a more generic callback/handler argument. I'll add a note in the code that this could be abstracted if something like this happens again

Issue #5026, opened 2023-04-17
Bug: `sam build` doesn't return an error when it should


aidansteele:
### Description:
I am trying to build and deploy an image-based Lambda function using the SAM CLI. `sam build` succeeds even when the image failed to build, which causes a lot of confusion. 

I think the cause is the underlying docker-py library not raising an exception when it should. I filed an issue in that repo too: https://github.com/docker/docker-py/issues/3108. The code in question hasn't changed between 4.2.0 (which SAM uses) and the head of that repo.

### Steps to reproduce:

Have the following template:
```yaml
Transform: AWS::Serverless-2016-10-31

Resources:
  Repro:
    Type: AWS::Serverless::Function
    Metadata:
      DockerContext: example
      Dockerfile: Dockerfile
    Properties:
      PackageType: Image
```

Make sure there is an `example` directory without a `Dockerfile` inside it. Now run `sam build`. This should fail. It fails (as expected) on Linux, but it succeeds on MacOS.

### Observed result:
```
aidan@Aidans-MacBook-Pro samcli-issue-repro % sam build --debug
2023-04-17 15:20:56,189 | Config file location: /Users/aidan/dev/tmp/samcli-issue-repro/samconfig.toml
2023-04-17 15:20:56,189 | Config file '/Users/aidan/dev/tmp/samcli-issue-repro/samconfig.toml' does not exist
2023-04-17 15:20:56,190 | Using SAM Template at /Users/aidan/dev/tmp/samcli-issue-repro/template.yml
2023-04-17 15:20:56,236 | Using config file: samconfig.toml, config environment: default
2023-04-17 15:20:56,236 | Expand command line arguments to:
2023-04-17 15:20:56,236 | --template_file=/Users/aidan/dev/tmp/samcli-issue-repro/template.yml --mount_with=READ --build_dir=.aws-sam/build --cache_dir=.aws-sam/cache 
2023-04-17 15:20:56,587 | 'build' command is called
2023-04-17 15:20:56,588 | No Parameters detected in the template
2023-04-17 15:20:56,609 | There is no customer defined id or cdk path defined for resource Repro, so we will use the resource logical id as the resource id
2023-04-17 15:20:56,609 | 0 stacks found in the template
2023-04-17 15:20:56,609 | No Parameters detected in the template
2023-04-17 15:20:56,618 | There is no customer defined id or cdk path defined for resource Repro, so we will use the resource logical id as the resource id
[...]

PR #5399, opened 2023-06-22
fix: use images.build rather than low level build API to catch errors

mndeveci:
#5026

Today, if there are issues during `sam build` for Image type lambda functions, SAM CLI exits with zero code even though the docker build fails. This is due to fact that we are using a low level API which doesn't return the status of the build operation.

This PR changes that process to use `images.build` which uses the earlier API under the hood but raises any issues if docker build fails.

Verified by running build tests manually, no issues have been identified.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


moelasmar (approved):

samcli/lib/build/app_builder.py
```diff
@@ -407,14 +407,18 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
...
             "platform": get_docker_platform(architecture),
             "rm": True,
         }
         if docker_build_target:
             build_args["target"] = cast(str, docker_build_target)
 
-        build_logs = self._docker_client.api.build(**build_args)
+        try:
+            (build_image, build_logs) = self._docker_client.images.build(**build_args)
+            LOG.debug("%s image is built for %s function", function_name, build_image)
```
hnnasit, line 418: Does `function_name` and `build_image` need to be swapped here?

hnnasit (commented):
Overall LGTM, left a small comment.

samcli/lib/build/app_builder.py
mndeveci: Good catch, updated!

PR #5409, opened 2023-06-23
feat: GraphQL init flow

mildaniel:

N/A

To support other types of app templates, such as the AppSync template, the interactive init flow needs to be modified since the current one doesn't make sense for if a template doesn't contain a Lambda function.

Creates an interface for implementing an "override workflow" that will be executed instead of the default workflow for app templates with certain use cases.

New AppSync Workflow:
image

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/override_init_flows.py
+++ b/samcli/commands/init/override_init_flows.py
@@ -113,7 +113,7 @@ class GraphQlInitFlow(OverrideFlow):
             output_dir=self._output_dir,
             name=self._name,
             app_template=self._use_case,
-            architecture=NOT_APPLICABLE,
+            architecture=[NOT_APPLICABLE],
         )
         click.echo(summary_msg)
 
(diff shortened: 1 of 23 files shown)


samcli/commands/init/override_init_flows.py
```diff
@@ -0,0 +1,171 @@
...
+        output_dir: Optional[str],
+    ):
+        self._name = name
+        self._templates = templates
+        self._use_case = use_case
+        self._preprocessed_options = preprocessed_options
+        self._output_dir = output_dir
+
+    @abstractmethod
+    def generate_project(self):
```
sriram-mv, line 42: Is override a short-circuit process? is it able to adapt to adding more questions/prompts if need be?

mildaniel: It's a short circuit process. My thinking was that any additional questions beyond the initial use case would be handled by the overriding flow. Happy to hear other ideas for the approach

mildaniel:
Closing this PR as we are going to address this issue in a different way by reworking the existing init flow logic.

PR #5410, opened 2023-06-23
chore: remove the CDK SAM CLI integration testing and depend on the same test cases defined in CDK repo

moelasmar:
Remove the CDK integration testing Appveyor job and depend on the testing we added to the CDK repository https://github.com/aws/aws-cdk/pull/18875 


mndeveci (approved):

.github/workflows/cdk-sam-cli-integration-testing.yml
```diff
@@ -0,0 +1,56 @@
...
+          - "3.11"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          # set last version as the one in matrix to make it default
+          python-version: |
+            3.7
+            3.8
+            3.11
```
mndeveci, line 38: These are only required if we have lambda functions using these python versions in the tests. Do we have lambda functions with all these runtimes, otherwise we can remove some of them here?

That is also valid for the runtimes below, let's remove the ones that we don't use in the tests (if there are any).

moelasmar: I do not remember .. let me check the lambda functions defined in the testing project

moelasmar: done

.github/workflows/cdk-sam-cli-integration-testing.yml
```diff
@@ -0,0 +1,54 @@
+name: Test the integration between SAM CLI and Latest CDK version
+
+on:
+  schedule:
+    - cron: "0 13 * * 1-5" # run 1PM UTC (6 am PDT) in every weekday
```
jfuss, line 5: Why run on a schedule instead of on pr?

moelasmar: we already have the CDK integration test cases but we test against the synthesized templates. This integration was created to validate that every new CDK version is generating the template SAM CLI expects it, that is why we originally created the appveyor job to run daily instead of running it with each PR.

hawflau (approved):

PR #5427, opened 2023-06-28
fix(invoke): Write in UTF-8 string instead of bytes

jfuss:
Revert fix issue #5229

It appears that we were using sys.stdout.buffer to support python2 and python3 at the same time. Switching to just write to sys.stdout allows us to write a utf-8 encoding string. When using sys.stdout.buffer, we can only write bytes and I couldn't get the correct UTF8 encoded string to print correctly.

This commit adds additional typing to help ensure things are working as expected using static analysis by mypy.

#5229

Should allow us to write utf8 correctly to the terminal.

See above git commit message

Shouldn't have any

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -291,6 +291,27 @@ def test_invoke_returns_expected_result_when_no_event_given(self):
         self.assertEqual(process.returncode, 0)
         self.assertEqual("{}", process_stdout.decode("utf-8"))
 
+    # @pytest.mark.flaky(reruns=3)
```
lucashuy, line 294: nit: left over from debugging I assume

samcli/local/docker/manager.py
```diff
@@ -168,16 +168,16 @@ def pull_image(self, image_name, tag=None, stream=None):
...
-            stream_writer.write("\nFetching {}:{} Docker container image...".format(image_name, tag))
+            stream_writer.write_str("\nFetching {}:{} Docker container image...".format(image_name, tag))
 
             # Each line contains information on progress of the pull. Each line is a JSON string
             for _ in result_itr:
                 # For every line, print a dot to show progress
-                stream_writer.write(".")
-                stream_writer.flush()
+                stream_writer.write_str(".")
+                # stream_writer.flush()
```
lucashuy, line 177: nit: left over from debugging I assume

samcli/local/docker/container.py
```diff
@@ -430,10 +446,27 @@ def _write_container_output(output_itr, stdout=None, stderr=None):
...
-                    stderr.write(stderr_data)
+                    if isinstance(stderr, StreamWriter):
+                        stderr.write_bytes(stderr_data)
+                        stderr.flush()
+
+                    if isinstance(stderr, io.BytesIO):
+                        stderr.write(stderr_data)
+
+                    if isinstance(stderr, io.TextIOWrapper):
+                        stderr.buffer.write(stderr_data)
```
lucashuy, line 468: Worth it to move this to a helper method since the same logic happens with `stdout`?

hawflau (approved):

tests/integration/local/invoke/test_integrations_cli.py
jfuss: Good catch. It made tests runs faster but forgot to adjust.

samcli/local/docker/manager.py
jfuss: Ya. I don't think flush is required here but will leave it. Shouldn't be hurting anything.

samcli/local/docker/container.py
jfuss: done

lucashuy (approved):

samcli/commands/remote/remote_invoke_context.py
```diff
@@ -242,7 +242,7 @@ class DefaultRemoteInvokeResponseConsumer(RemoteInvokeConsumer[RemoteInvokeRespo
     _stream_writer: StreamWriter
 
     def consume(self, remote_invoke_response: RemoteInvokeResponse) -> None:
-        self._stream_writer.write(cast(str, remote_invoke_response.response).encode())
+        self._stream_writer.write_bytes(cast(str, remote_invoke_response.response).encode())
```
mndeveci, line 245: nit: could this be a method in `StreamWriter` where it encodes the given string and writes as bytes (just a QOL improvement so that we don't need to encode every time calling it)

edit: I saw that there is also `write_str` method below, why we are not using it here? (trying to understand the reasoning)

samcli/local/docker/container.py
```diff
@@ -344,11 +347,21 @@ def wait_for_result(self, full_path, event, stdout, stderr, start_timer=None):
         # start the timer for function timeout right before executing the function, as waiting for the socket
         # can take some time
         timer = start_timer() if start_timer else None
-        self.wait_for_http_response(full_path, event, stdout)
+        response = self.wait_for_http_response(full_path, event, stdout)
```
mndeveci, line 350: Should `response` be printed into logs (as a debug)  maybe?  We weren't using it before, but now it is assigned but not used (as far as I can see) 🤔

mndeveci (commented):
LGTM in general, just asking questions to understand some of the changes in the code.

samcli/commands/remote/remote_invoke_context.py
jfuss: This mostly comes down to time and understanding. I didn't want to do a complete refactor within this commit but instead add typing to help myself get this right and solve the bug attached to this commit. 

I could have gone through and update all of this but it's not clear to me why we are writing both `str` and `bytes` to the stream. This might be legacy at this point but I wanted to keep the paths the same as much as I could for now. Accepting `bytes` likely stems from docker returning logs as `bytes`. I vaguely recall my past struggles with this area during the Py2 to Py3 migration but that was so long ago now, those struggles were replaced 😊. My preference here is to do this migration to only `str` slowly to hopefully not affect the content we write to the console.

samcli/local/docker/container.py
jfuss: Used on line 357

mndeveci: Oh that removal of function definition (`wait_for_logs`) confused me since it has different indentation.

mndeveci (approved):

PR #5430, opened 2023-06-29
fix: remove circular dependency by moving parse_s3 method to its own util file

mndeveci:
N/A

Fixing circular dependency which happens during E2E tests;
```
E   ImportError: cannot import name 'S3Uploader' from partially initialized module 'samcli.lib.package.s3_uploader' (most likely due to a circular import) 
```

By moving `parse_s3_url` to a new module under `utils`.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


sriram-mv (approved):
LGTM, we have a proliferation of utils happening across, I wonder if we should consolidate them into another library.

hnnasit (approved):

PR #5441, opened 2023-07-04
feat(sync): support build-image option

elvishp2006:
To use --build-image on sync command

Adding support to --build-image on sync command

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/build/command.py
```diff
@@ -94,21 +95,7 @@
...
-    help="Container image URIs for building functions/layers. "
-    "You can specify for all functions/layers with just the image URI "
-    "(--build-image public.ecr.aws/sam/build-nodejs18.x:latest). "
-    "You can specify for each individual function with "
-    "(--build-image FunctionLogicalID=public.ecr.aws/sam/build-nodejs18.x:latest). "
-    "A combination of the two can be used. If a function does not have build image specified or "
-    "an image URI for all functions, the default SAM CLI build images will be used.",
-    cls=ContainerOptions,
-)
+@build_image_option(cls=ContainerOptions)
```
mndeveci, line 98: Just curiosity, this `ContainerOptions` are the same type for both build and sync, why do we need to parameterize it?

mndeveci (commented):
Thanks a lot for your contribution, LGTM in general! Left a small question for clarification.

samcli/commands/build/command.py
elvishp2006: It's because if I import this class on options.py will be a circular import, this class is present also on build file.

mndeveci (approved):

hnnasit (approved):
Thanks for your contributions @elvishp2006. Overall LGTM, we will need to update our docs https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-sync.html which we can do on our side.

hnnasit (commented):
Hey @elvishp2006, apologies I missed this before but we will also need to add the new option in this [list](https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/sync/core/options.py#L16-L28) for it to be visible in `sam sync --help` command.

PR #5445, opened 2023-07-04
feat: Retrieve SAM CLI command parameters

Leo10Gama:
To begin with generating a JSON schema, we must retrieve the parameters as dictionary objects to make them simpler to work with.

Functions have been added, along with a new `schema` folder, to retrieve the parameters within commands of the SAM CLI.

A new folder with the schema logic has been added to the root directory.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/schema.py
```diff
@@ -0,0 +1,84 @@
...
+        A dictionary representation of the JSON schema.
+    """
+    schema = {}
+    commands = {}
+    params = set()  # NOTE(leogama): Currently unused due to some params having different help values
+    # TODO: Populate schema with relevant attributes
+    for package_name in _SAM_CLI_COMMAND_PACKAGES:
+        new_command = retrieve_command_structure(package_name)
+        commands.update(new_command)
+        command_params = [param for param_list in new_command.values() for param in param_list]
```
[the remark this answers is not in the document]
Leo10Gama: Yup, for the sake of readability we can definitely have it expanded out

schema/schema.py
```diff
@@ -0,0 +1,84 @@
...
+            formatted_param["choices"] = param.type.choices
+
+        return formatted_param
+
+    if isinstance(module.cli, click.core.Group):  # command has subcommands (e.g. local invoke)
+        for subcommand in module.cli.commands.values():
+            params = [
+                param
+                for param in subcommand.params
+                if param.name != None and isinstance(param, click.core.Option)  # exclude None and non-Options
```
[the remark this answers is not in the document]
Leo10Gama: It might be, but my initial logic behind it is that every command seems to have an `<Option None>` that doesn't seem to have any other parameters. Not sure if it's a remnant of something else, but since it doesn't show up in any of the parameter lists, I'm assuming it just shouldn't be there anyway, so I filter it out.
The second instance check is to make sure there are no `Argument`s being passed in as well, since those can't be set in the config AFAIK.

schema/schema.py
```diff
@@ -0,0 +1,84 @@
...
+            ]
+            formatted_params = {param.name: format_param(param) for param in params}
+            command.update({SamConfig._to_key([command_name, subcommand.name]): formatted_params})
```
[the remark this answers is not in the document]
Leo10Gama: Yup, I wasn't sure why it was "protected" in the first place since it seems like a useful function to have, so we can open it up since it's only used in a few other places anyway

schema/schema.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if param.default:
+            formatted_param["default"] = param.default
+
+        if param.type.name == "choice":
+            formatted_param["choices"] = param.type.choices
+
+        return formatted_param
+
+    if isinstance(module.cli, click.core.Group):  # command has subcommands (e.g. local invoke)
```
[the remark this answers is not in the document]
Leo10Gama: We could, though it would involve writing out a whole separate function for parsing one click command. Still though, it would also address some of the code duplication, so it would definitely be a good addition 👍

schema/schema.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    module = importlib.import_module(package_name)
+    command = {}
+
+    def get_params_from_command(cli, main_command_name: str = "") -> dict:
+        """Given a command CLI, return its parameters."""
+        params = [
+            param
+            for param in cli.params
+            if param.name is not None and isinstance(param, click.core.Option)  # exclude None and non-Options
```
hnnasit, line 46: nit: This can be simplified to `if param.name`

schema/schema.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Returns
+    -------
+    dict
+        A dictionary that maps the name of the command to its relevant click options.
+    """
+    module = importlib.import_module(package_name)
+    command = {}
+
+    def get_params_from_command(cli, main_command_name: str = "") -> dict:
```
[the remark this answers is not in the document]
Leo10Gama: That's a good catch. My idea was to have it append the name of the main command for handling the naming of subcommands, but I forgot to actually put it to use. Will fix that 👍

schema/schema.py
lucashuy, line 1: nit: could we update the file with some parameter doc strings and return values?

schema/schema.py
```diff
@@ -0,0 +1,86 @@
...
```
lucashuy, line 41: question: any particular reasoning as to why we have this as an inner method? No particular issue with this aside from making it easier to test if this was located outside the retrieve method.

Leo10Gama: My line of reasoning was that it was probably only going to be used within the context of this method as a helper function, but we could bring it outside if that's preferred

lucashuy: Let's move it outside, that way when it comes time to test things, we can isolate this method to make it easier to test

Makefile
```diff
@@ -34,15 +34,15 @@ smoke-test:
 
 lint:
 	# Linter performs static analysis to catch latent bugs
-	ruff samcli
+	ruff samcli schema
 	# mypy performs type check
 	mypy --exclude /testdata/ --exclude /init/templates/ --no-incremental setup.py samcli tests
```
lucashuy, line 39: sorry, one last change: lets add the new schema folder to the mypy check

Makefile
```diff
@@ -34,15 +34,15 @@ smoke-test:
...
 
 # Command to run everytime you make changes to verify everything works
 dev: lint test
 
 black:
-	black setup.py samcli tests
+	black setup.py samcli tests schema
 
 black-check:
 	black --check setup.py samcli tests
```
lucashuy, line 48: and same comment here, lets add the schema folder to this make rule

lucashuy (approved):

hnnasit (approved):

PR #5458, opened 2023-07-06
feat: Added remote invoke implementation for step functions

hnnasit:
This PR adds Step functions support for remote invoke.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/remote/remote_invoke_context.py
+++ b/samcli/commands/remote/remote_invoke_context.py
@@ -31,13 +31,13 @@ from samcli.lib.utils.cloudformation import (
     get_resource_summary,
     get_resource_summary_from_physical_id,
 )
-from samcli.lib.utils.resources import AWS_LAMBDA_FUNCTION
+from samcli.lib.utils.resources import AWS_LAMBDA_FUNCTION, AWS_STEPFUNCTIONS_STATEMACHINE
 from samcli.lib.utils.stream_writer import StreamWriter
 
 LOG = logging.getLogger(__name__)
 
 
-SUPPORTED_SERVICES = {"lambda": AWS_LAMBDA_FUNCTION}
+SUPPORTED_SERVICES = {"lambda": AWS_LAMBDA_FUNCTION, "step-functions": AWS_STEPFUNCTIONS_STATEMACHINE}
 
 
 class RemoteInvokeContext:
--- a/samcli/lib/remote_invoke/remote_invoke_executor_factory.py
+++ b/samcli/lib/remote_invoke/remote_invoke_executor_factory.py
@@ -20,10 +20,12 @@ from samcli.lib.remote_invoke.remote_invoke_executors import (
     RemoteInvokeResponse,
     ResponseObjectToJsonStringMapper,
 )
-from samcli.lib.utils.cloudformation import CloudFormationResourceSummary
-from samcli.lib.utils.resources import (
-    AWS_LAMBDA_FUNCTION,
+from samcli.lib.remote_invoke.stepfunctions_invoke_executors import (
+    SfnDescribeExecutionResponseConverter,
+    StepFunctionsStartExecutionExecutor,
 )
+from samcli.lib.utils.cloudformation import CloudFormationResourceSummary
+from samcli.lib.utils.resources import AWS_LAMBDA_FUNCTION, AWS_STEPFUNCTIONS_STATEMACHINE
 
 LOG = logging.getLogger(__name__)
 
@@ -127,6 +129,41 @@ class RemoteInvokeExecutorFactory:
             log_consumer=log_consumer,
         )
 
+    def _create_stepfunctions_boto_executor(
+        self,
+        cfn_resource_summary: CloudFormationResourceSummary,
+        remote_invoke_output_format: RemoteInvokeOutputFormat,
+        response_consumer: RemoteInvokeConsumer[RemoteInvokeResponse],
+        log_consumer: RemoteInvokeConsumer[RemoteInvokeLogOutput],
+    ):
+        """Creates a remote invoke executor for Step Functions resource type based on
+        the boto action being called.
+
+        :param cfn_resource_summary: Information about the Lambda resource
+        :param remote_invoke_output_format: Response output format
+        :param response_consumer: Consumer for writing responses to stdout
+        :param log_consumer: Log consumer for writing logs to stderr
+
+        :return: Returns the created remote invoke Executor
+        """
+        LOG.info(f"Invoking Step Function {cfn_resource_summary.logical_resource_id}")
+        sfn_client = self._boto_client_provider("stepfunctions")
+        mappers = []
+        if remote_invoke_output_format == RemoteInvokeOutputFormat.JSON:
+            mappers = [
+                SfnDescribeExecutionResponseConverter(),
+                ResponseObjectToJsonStringMapper(),
+            ]
+        return RemoteInvokeExecutor(
+            request_mappers=[DefaultConvertToJSON()],
+            response_mappers=mappers,
+            boto_action_executor=StepFunctionsStartExecutionExecutor(
+                sfn_client, cfn_resource_summary.physical_resource_id, remote_invoke_output_format
+            ),
+            response_consumer=response_consumer,
+            log_consumer=log_consumer,
+        )
+
     # mapping definition for each supported resource type
     REMOTE_INVOKE_EXECUTOR_MAPPING: Dict[
         str,
@@ -142,4 +179,5 @@ class RemoteInvokeExecutorFactory:
         ],
     ] = {
         AWS_LAMBDA_FUNCTION: _create_lambda_boto_executor,
+        AWS_STEPFUNCTIONS_STATEMACHINE: _create_stepfunctions_boto_executor,
     }
--- a/samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
+++ b/samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
@@ -1,0 +1,115 @@
+import logging
+import time
+from typing import Any
+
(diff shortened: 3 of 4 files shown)


samcli/lib/remote_invoke/remote_invoke_executor_factory.py
```diff
@@ -142,4 +179,5 @@ def _create_lambda_boto_executor(
         ],
     ] = {
         AWS_LAMBDA_FUNCTION: _create_lambda_boto_executor,
+        AWS_STEPFUNCTIONS_STATEMACHINE: _create_stepfunctions_boto_executor,
```
mndeveci, line 182: I think we are going to remove this until we release this feature, right?

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,115 @@
...
+            payload,
+            self._state_machine_arn,
+        )
+        try:
+            start_execution_response = self._stepfunctions_client.start_execution(**self.request_parameters)
+            execution_arn = start_execution_response.get(EXECUTION_ARN)
+
+            describe_execution_response = self._stepfunctions_client.describe_execution(executionArn=execution_arn)
+            while describe_execution_response["status"] == RUNNING:
+                describe_execution_response = self._stepfunctions_client.describe_execution(executionArn=execution_arn)
```
mndeveci, line 73: nit: might be good to put a debug log message here

samcli/lib/remote_invoke/remote_invoke_executor_factory.py
hnnasit: Yes that's right, added this here so that Jacob can play around with the draft PR. Can update this before merging.

hnnasit: I have removed this in the new commits.

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,119 @@
...
+                execution_arn = describe_execution_response["executionArn"]
+                LOG.debug(f"ExecutionArn: {execution_arn}, status: {execution_status}")
+                time.sleep(SFN_EXECUTION_WAIT_TIME)
+
+            if self._remote_output_format == RemoteInvokeOutputFormat.JSON:
+                yield RemoteInvokeResponse(describe_execution_response)
+            if self._remote_output_format == RemoteInvokeOutputFormat.TEXT:
+                output_data = describe_execution_response.get("output", "")
+                error_data = describe_execution_response.get("error", "")
+                yield RemoteInvokeResponse(output_data)
```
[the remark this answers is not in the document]
hnnasit: This follows from the pattern which is used for Lambda functions invoke where the response generated is consumed right away. For lambda response stream invoke, the outputs are generated before the execution is completed. Since the setup [here](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/remote_invoke/remote_invoke_executors.py#L252-L256) is the same, we use yield for SFN as well.

samcli/lib/remote_invoke/remote_invoke_executor_factory.py
```diff
@@ -127,6 +129,41 @@ def _create_lambda_boto_executor(
...
+        the boto action being called.
+
+        :param cfn_resource_summary: Information about the Lambda resource
+        :param remote_invoke_output_format: Response output format
+        :param response_consumer: Consumer for writing responses to stdout
+        :param log_consumer: Log consumer for writing logs to stderr
+
+        :return: Returns the created remote invoke Executor
+        """
+        LOG.info(f"Invoking Step Function {cfn_resource_summary.logical_resource_id}")
```
hawflau, line 149: nit and feel free to ignore: personally I prefer using `%s` over f-string for logging. 
There's a pylint rule [W1203](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html) that catches using f-string in logging, even though I think we didn't enable it.
Discussion about why `%s` is preferred for logging over f-string: https://stackoverflow.com/questions/54367975/python-3-7-logging-f-strings-vs

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,119 @@
...
+SFN_EXECUTION_WAIT_TIME = 2
+
+
+class StepFunctionsStartExecutionExecutor(BotoActionExecutor):
+    """
+    Calls "start_execution" method of "stepfunctions" service with given input.
+    If a file location provided, the file handle will be passed as Payload object
+    """
+
+    _stepfunctions_client: Any
```
hawflau, line 35: Would suggest to use specific type rather than `Any`. We have `boto3-stub` as dev dependency, so we should use it to provide more accurate type hint.

samcli/lib/remote_invoke/remote_invoke_executor_factory.py
mndeveci: I agree using `%s` rather than f-strings. Most of the logging frameworks have improvements using their formatting since they can decide whether the string needs to be formatted or not. If you use f-strings, the string will be formatted no matter the log will be printed or not (depending on the logging level).

samcli/commands/remote/remote_invoke_context.py
```diff
@@ -92,8 +92,8 @@ def run(self, remote_invoke_input: RemoteInvokeExecutionInfo) -> None:
             DefaultRemoteInvokeResponseConsumer(self.stdout),
             DefaultRemoteInvokeLogConsumer(self.stderr),
         )
-        if not remote_invoke_executor:
-            raise NoExecutorFoundForRemoteInvoke(
+        if not remote_invoke_executor or self._resource_summary.resource_type not in SUPPORTED_SERVICES.values():
```
mndeveci, line 95: nit: do we need second part of the `if` statement here? If there is no executor then we should be already failing here.

samcli/lib/remote_invoke/remote_invoke_executor_factory.py
```diff
@@ -127,6 +129,41 @@ def _create_lambda_boto_executor(
...
+    ):
+        """Creates a remote invoke executor for Step Functions resource type based on
+        the boto action being called.
+
+        :param cfn_resource_summary: Information about the Lambda resource
+        :param remote_invoke_output_format: Response output format
+        :param response_consumer: Consumer for writing responses to stdout
+        :param log_consumer: Log consumer for writing logs to stderr
+
+        :return: Returns the created remote invoke Executor
```
mndeveci, line 147: We use nupy style docstrings (I think we missed some other places in this file)

https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#code-conventions

We should be having that as a style check, until then can you update docstrings in the changed files to align with our dev-guide?

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,119 @@
+import logging
```
mndeveci, line 1: module level docstring please

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,119 @@
...
+            if parameter_key == "stateMachineArn":
+                LOG.warning("stateMachineArn is defined using the value provided for resource_id argument.")
+            elif parameter_key == "input":
+                LOG.warning("input is defined using the value provided for either --event or --event-file options.")
+            else:
+                self.request_parameters[parameter_key] = parameter_value
+
+        if not self.request_parameters.get("name"):
+            timestamp = str(int(time.time() * 1000))
+            self.request_parameters["name"] = f"sam_remote_invoke_{timestamp}"
```
mndeveci, line 57: nit: timestamp is nice, iso formatted string would be even better for readability

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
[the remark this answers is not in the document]
hawflau: IMO, using `return` would be better if the method only produces a single output value.
Using `yield` gives a feeling that the method is producing a series of values. Will using `return` break the code?

samcli/lib/remote_invoke/stepfunctions_invoke_executors.py
```diff
@@ -0,0 +1,119 @@
...
+        LOG.debug("Mapping Step Function execution response to string object")
+        if not isinstance(remote_invoke_input.response, dict):
+            raise InvalideBotoResponseException(
+                "Invalid response type received from Step Functions service, expecting dict"
+            )
+
+        start_date_field = remote_invoke_input.response.get("startDate")
+        stop_date_field = remote_invoke_input.response.get("stopDate")
+        if start_date_field:
+            remote_invoke_input.response["startDate"] = start_date_field.strftime("%m/%d/%Y, %H:%M:%S")
```
hawflau, line 116: Should we also include the timezone in the formatted datetime?

samcli/commands/remote/remote_invoke_context.py
hnnasit: :thinking_face: I was thinking with this check, we can add StepFunctions to the `REMOTE_INVOKE_EXECUTOR_MAPPING` dict in the factory and only maintain the supported_services dict until SFN support is released. But I agree this check here is redundant, I can remove SFN from  `REMOTE_INVOKE_EXECUTOR_MAPPING` and the 2nd part of the check.

hawflau (approved):

jfuss (approved):

PR #5468, opened 2023-07-07
feat: Parse parameters into JSON schema objects

Leo10Gama:
In order for SAM CLI parameters to be read correctly in the schema, they must be parsed in a certain format.

Instead of formatting into an arbitrary dictionary, the parameters are parsed into JSON schema keywords. In addition, some other schema generation methods have been set up.

The parameters that are parsed use different keys.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/schema/schema.py
+++ b/schema/schema.py
@@ -34,12 +34,12 @@ def format_param(param: click.core.Option) -> Dict[str, Any]:
 
     # NOTE: Params do not have explicit "string" type; either "text" or "path".
     #       All choice options are from a set of strings.
-    if param.type.name in ["text", "path", "choice", "filename", "directory"]:
+    if param.type.name.lower() in ["text", "path", "choice", "filename", "directory"]:
         formatted_param["type"] = "string"
-    elif param.type.name == "LIST":
+    elif param.type.name.lower() == "list":
         formatted_param["type"] = "array"
     else:
-        formatted_param["type"] = param.type.name or "string"
+        formatted_param["type"] = param.type.name.lower() or "string"
 
     if param.default:
         formatted_param["default"] = list(param.default) if isinstance(param.default, tuple) else param.default
--- a/schema/schema.py
+++ b/schema/schema.py
@@ -2,7 +2,9 @@
 
 
 import importlib
-from typing import Dict, List, Optional, Union
+import json
+from enum import Enum
+from typing import Any, Dict
 
 import click
 
@@ -10,7 +12,13 @@ from samcli.cli.command import _SAM_CLI_COMMAND_PACKAGES
 from samcli.lib.config.samconfig import SamConfig
 
 
-def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], List[str]]]:
+class SchemaKeys(Enum):
+    SCHEMA_FILE_NAME = "samcli.json"
+    SCHEMA_DRAFT = "http://json-schema.org/draft-04/schema"
+    TITLE = "AWS SAM CLI samconfig schema"
+
+
+def format_param(param: click.core.Option) -> Dict[str, Any]:
     """Format a click Option parameter to a dictionary object.
 
     A parameter object should contain the following information that will be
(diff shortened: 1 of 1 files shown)


schema/schema.py
```diff
@@ -91,15 +101,34 @@ def generate_schema() -> dict:
...
         params.update(command_params)
+    # TODO: Generate schema for each of the commands
     return schema
 
 
+def write_schema():
+    """Generate the SAM CLI JSON schema and write it to file."""
+    schema = generate_schema()
+    with open(SchemaKeys.SCHEMA_FILE_NAME.value, "w+", encoding="utf-8") as outfile:
+        json.dump(schema, outfile)
```
lucashuy, line 130: Can we dump with some indenting (if this doesn't do this already by default)

schema/schema.py
```diff
@@ -22,20 +30,22 @@ def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], Lis
...
+    if param.type.name in ["text", "path", "choice", "filename", "directory"]:
         formatted_param["type"] = "string"
+    elif param.type.name == "LIST":
+        formatted_param["type"] = "array"
     else:
         formatted_param["type"] = param.type.name
 
     if param.default:
-        formatted_param["default"] = str(param.default)
+        formatted_param["default"] = list(param.default) if isinstance(param.default, tuple) else param.default
```
lucashuy, line 45: Just a question, what are some instances of a default being a tuple of something?

schema/schema.py
```diff
@@ -22,20 +30,22 @@ def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], Lis
...
 
     # NOTE: Params do not have explicit "string" type; either "text" or "path".
     #       All choice options are from a set of strings.
-    if param.type.name in ["text", "path", "choice"]:
+    if param.type.name in ["text", "path", "choice", "filename", "directory"]:
         formatted_param["type"] = "string"
+    elif param.type.name == "LIST":
+        formatted_param["type"] = "array"
     else:
         formatted_param["type"] = param.type.name
```
lucashuy, line 42: Should we be worried about casing here? Just commenting since the list check above looks for all uppercase.

Leo10Gama: We should, yes. When I went looking through the parameter types to make sure everything was accurate, it was returned exactly like that. I thought it weird as well, since everything else is lowercase, but it's the only other instance that shows up.
image

schema/schema.py
Leo10Gama: The only one I was able to find was `sam sync --capabilities`, which has a default value of `('CAPABILITY_NAMED_IAM', 'CAPABILITY_AUTO_EXPAND')` (as a tuple). Even though it's just the one instance, JSON schema doesn't accept tuples, so I cast them to list, which will have the same effect in the schema.

schema/schema.py
```diff
@@ -22,20 +30,22 @@ def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], Lis
                        a list of those allowed options
     * default - The default option for that parameter
     """
-    formatted_param: Dict[str, Union[Optional[str], List[str]]] = {"name": param.name, "help": param.help}
+    formatted_param: Dict[str, Any] = {"title": param.name, "description": param.help}
```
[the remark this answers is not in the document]
Leo10Gama: It was done partly for linting, but mostly just to improve readability, since I realized the value of those keys could be more than just strings populating that field, especially for the `"default"` key (i.e. `local start-api --port` has a default value of `3000` as an integer, `sync --dependency-layer` has a default value of `True` as a boolean). To include every possible data type would become more verbose than just saying `Any`, and would also make the linter throw errors if we were to introduce additional default values for other parameter types (e.g. if we added a default `number` parameter, which would then need to be included)

schema/schema.py
```diff
@@ -22,20 +30,22 @@ def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], Lis
...
 
     # NOTE: Params do not have explicit "string" type; either "text" or "path".
     #       All choice options are from a set of strings.
-    if param.type.name in ["text", "path", "choice"]:
+    if param.type.name in ["text", "path", "choice", "filename", "directory"]:
         formatted_param["type"] = "string"
+    elif param.type.name == "LIST":
```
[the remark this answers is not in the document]
Leo10Gama: We could, though since we are only using the string in this one situation, and the logic is localized to this function, I don't see much of a need for having it separated from this logic.
I wouldn't be opposed to making it constant, I just don't see much reason to right now besides adding abstracting it.

hnnasit (approved):

schema/schema.py
lucashuy: Since that's the case, we should consider converting the parameter type names to lowercase (or uppercase) before doing a comparison against values we check in case anything changes in the future (`LIST` gets updated to `list`)

Leo10Gama: Yup that sounds reasonable, will add those changes

lucashuy (approved):
LGTM, left with a nit comment

schema/schema.py
```diff
@@ -22,20 +30,22 @@ def format_param(param: click.core.Option) -> Dict[str, Union[Optional[str], Lis
...
     # NOTE: Params do not have explicit "string" type; either "text" or "path".
     #       All choice options are from a set of strings.
-    if param.type.name in ["text", "path", "choice"]:
+    if param.type.name.lower() in ["text", "path", "choice", "filename", "directory"]:
         formatted_param["type"] = "string"
+    elif param.type.name.lower() == "list":
+        formatted_param["type"] = "array"
     else:
-        formatted_param["type"] = param.type.name
+        formatted_param["type"] = param.type.name.lower() or "string"
```
lucashuy, line 42: nit: can this be a variable instead of using `.lower()` three times? That way if something were to change, we can just change one line.

PR #5469, opened 2023-07-07
feat: Support YAML file formats for config files

lucashuy:


samcli/cli/cli_config_file.py
```diff
@@ -240,14 +304,22 @@ def decorator(f):
...
+    Returns
+    -------
+    Callable
+        A Callback function
     """
-    config_file_attrs = {}
+    config_file_attrs: Dict[str, Any] = {}
     config_file_param_decls = ("--config-file",)
     config_file_attrs["help"] = "Configuration file containing default parameter values."
     config_file_attrs["default"] = "samconfig.toml"
```
hnnasit, line 325: Maybe this got missed in previous reviews but should we use `DEFAULT_CONFIG_FILE_NAME` constant here so that this does not get missed if we decide to change the default?

samcli/cli/cli_config_file.py
```diff
@@ -258,14 +330,22 @@ def decorator_customize_config_file(f):
...
+    Returns
+    -------
+    Callable
+        A Callback function
     """
-    config_env_attrs = {}
+    config_env_attrs: Dict[str, Any] = {}
     config_env_param_decls = ("--config-env",)
     config_env_attrs["help"] = "Environment name specifying default parameter values in the configuration file."
     config_env_attrs["default"] = "default"
```
hnnasit, line 351: Same as above for consistency for the constant `DEFAULT_ENV`?

samcli/lib/config/samconfig.py
```diff
@@ -261,6 +244,40 @@ def _deduplicate_global_parameters(self, cmd_name_key, section, key, env=DEFAULT
...
+        Returns
+        -------
+        str
+            The name of the config file found, if it exists. In the case that it does not exist, the default config
+            file name is returned instead.
+        """
+        config_files_found = 0
+        config_file = DEFAULT_CONFIG_FILE_NAME
+
+        for extension in reversed(list(FILE_MANAGER_MAPPER.keys())):
```
sriram-mv, line 265: Q: why reversed?

tests/unit/lib/samconfig/test_samconfig.py
```diff
@@ -221,3 +232,84 @@ def test_write_config_file_will_create_the_file_if_not_exist(self):
...
+
+        with self.assertRaises(SamConfigFileReadException):
+            SamConfig(config_path, filename="samconfig")
+
+    def test_file_manager_unsupported(self):
+        config_dir = tempfile.gettempdir()
+        config_path = Path(config_dir, "samconfig.jpeg")
+
+        with self.assertRaises(SamConfigFileReadException):
+            SamConfig(config_path, filename="samconfig.jpeg")
```
sriram-mv, line 291: Steganography 👀

tests/unit/lib/samconfig/test_file_manager.py
```diff
@@ -0,0 +1,277 @@
...
+    def test_write_json_file_bad_path(self):
+        config_path = Path("path/to/some", "file_that_doesnt_exist.json")
+
+        with self.assertRaises(FileNotFoundError):
+            JsonFileManager.write({"key": "value"}, config_path)
+
+    def test_json_put_comment(self):
+        json_doc = {"version": 0.1, "config_env": {"topic1": {"parameters": {"word": "clarity"}}}}
+
+        json_doc = JsonFileManager.put_comment(json_doc, "This is a comment")
```
sriram-mv, line 274: maybe this test already exists, do we preserve comments when we read and write back to the config file?

samcli/lib/config/samconfig.py
Leo10Gama: For priority. The code goes through each possible extension, checks if the file exists, and if it does, sets that to be the config file to use. Iterating backwards means that the first extension in the mapper (in this case `.toml`, which is the default) will be used, even if other extensions like `.yaml` are found.

tests/unit/lib/samconfig/test_file_manager.py
Leo10Gama: For this case of JSON, the only way we can store comments is as dictionary keys, since it doesn't natively support comments. This test asserts that the comment can be written by the `put_comment()` method, and `test_write_json()` asserts that the comment will read back in correctly (line 253). Either way, comments will be preserved.

samcli/lib/config/samconfig.py
Leo10Gama: While not stated here, it is [left as a comment on the `FILE_MANAGER_MAPPER` itself](https://github.com/aws/aws-sam-cli/blob/e0860ce3615296a17447e868afd521453eefd726/samcli/lib/config/file_manager.py#L337).

sriram-mv (approved):
overall looks good

samcli/lib/config/samconfig.py
sriram-mv: that dict is ordered by default since python 3.x so that's okay, but the naming does not imply what priority we actually want. Its hidden behind the file manager mapper which makes it less obvious. I would still say make it more explicit.


I approved the PR, but let's look to change it.

PR #5478, opened 2023-07-11
chore: fix the v2 gateway integration to route resources linking

moelasmar:
Updating the Linking algorithm to add a function to parse the linking property value to extract the destination resource id so we can use it to get the destination resource logical id. This function will be used in processing the applied resources, and for the resources that the linking properties values does not equal the destination resource like the following case:
```
resource "aws_apigatewayv2_integration" "example" {
  api_id           = aws_apigatewayv2_api.example.id
  integration_type = "HTTP_PROXY"

  integration_method = "ANY"
  integration_uri    = "https://example.com/{proxy}"
}

resource "aws_apigatewayv2_route" "example" {
  api_id    = aws_apigatewayv2_api.example.id
  route_key = "ANY /example/{proxy+}"

  target = "integrations/${aws_apigatewayv2_integration.example.id}"
}
```
the new function will extract the integration id from the `target` property value.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


mildaniel (approved):
Good catch, thanks Mohamed.

lucashuy (approved):

PR #5484, opened 2023-07-11
feat: Implement JSON schema generation logic

Leo10Gama:
In order to implement auto-generation of a JSON schema, we must include a general framework to accommodate for future reviews, making the process of reviewing individual command schemas easier. In addition, we can write functions that automatically take this parsed information and write it to a JSON file.

The `write_schema()` function will write a dictionary to a JSON file. In addition, the `generate_schema()` function has been revamped to include more relevant schema information, such as titles and tooltips (descriptions) that will make writing samconfig files easier.

Now, when the `schema/schema.py` file is run, it will write the JSON schema to `schema/samcli.json`.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/schema.py
```diff
@@ -51,15 +52,14 @@ def format_param(param: click.core.Option) -> Dict[str, Any]:
...
+def get_params_from_command(cli) -> Dict[str, dict]:
     """Given a command CLI, return it in a dictionary, pointing to its parameters as dictionary objects."""
     params = [
         param
         for param in cli.params
         if param.name and isinstance(param, click.core.Option)  # exclude None and non-Options
     ]
-    cmd_name = SamConfig.to_key([main_command_name, cli.name]) if main_command_name else cli.name
-    return {cmd_name: {param.name: format_param(param) for param in params}}
+    return {param.name: format_param(param) for param in params if param.name}
```
lucashuy, line 62: Could we refactor this to be a single loop instead of two? Or if should stay, why we couldn't move the parameter collection and parsing to a single loop

schema/schema.py
```diff
@@ -109,18 +108,65 @@ def generate_schema() -> dict:
...
     }
     schema["required"] = ["version"]
     schema["additionalProperties"] = False
     # Iterate through packages for command and parameter information
     for package_name in _SAM_CLI_COMMAND_PACKAGES:
         new_command = retrieve_command_structure(package_name)
         commands.update(new_command)
-        for param_list in new_command.values():
+    for cmd_properties in commands.values():
+        for param_list in cmd_properties["parameters"].values():
```
lucashuy, line 120: Lets use `.get()` instead so that we can specify an empty dictionary as the default just as an extra safe guard

schema/schema.py
lucashuy, line 1: Thoughts about using Python `dataclass` objects to represent some of the objects here? This could likely reduce the amount of dictionary "gets" with something more concrete, for both commands and parameters.

schema/schema.py
```diff
@@ -109,18 +108,65 @@ def generate_schema() -> dict:
...
+            + "\n* ".join([f"{param['title']}:\n{param['description']}" for param in cmd_items["parameters"].values()])
+            if not exclude_params
+            else ""
+        )
+        param_description = f"Available parameters for the {formatted_cmd_name} command:\n{formatted_params_list}"
+        schema["patternProperties"][SchemaKeys.ENVIRONMENT_REGEX.value]["properties"].update(
+            {
+                cmd_name: {
+                    "title": f"{' '.join(full_cmd_name).title()} command",
+                    "description": cmd_items["description"] or "",
```
lucashuy, line 157: I've noticed that some of the descriptions in the skeleton schema have tabs in them, will this affect anything in how the information is displayed in the various code editors/IDEs? (this isn't the location where the description is set, but wanted to call it out here)

Also `.get()` to access this (this change is dependent on the outcome of the comment at the top of this file) :)

schema/schema.py
```diff
@@ -132,4 +178,4 @@ def write_schema():
 
 
 if __name__ == "__main__":
-    generate_schema()
+    write_schema()
```
lucashuy, line 181: Could we make this more modular? That way in the future when tests are added/considered, we could test each component a little more easily.

schema/schema.py
Leo10Gama: In my testing using VSCode, it doesn't seem to be an issue for readability. I thought it was odd that there's inconsistency among the descriptions, with some starting with newlines or being indented more than others, but at least with VSCode it's still legible. 
I could look into potentially reformatting some of the help text to be more consistent with everything, but I'm worried that may have impact across the CLI itself (i.e. things are formatted the way they are for readability in the CLI, and fixing for the description could decrease readability there). Alternatively, we could write a cleaner function to do things like prevent more than two spaces between words, and prevent more than one or two newlines.

schema/schema.py
Leo10Gama: I agree, putting things in dataclasses will probably make the code easier to comprehend as well. I can get on that for the next commit 👍

schema/schema.py
lucashuy: Just to pull in an example from the build skeleton: https://github.com/aws/aws-sam-cli/pull/5484/files#diff-8ad631c7b37df828ce338f907037667605bd5eaad2e91c772c073936cd90a5f6R52

If its pulling the descriptions right from the commands, then the formatting is likely there to make it easier to read when running something like `sam build --help`. We could also consider just trimming the start and ends of the descriptions make sure that they don't get parsed by other extensions or editors outside of VSCode.

schema/samcli.json
```diff
@@ -0,0 +1,338 @@
...
+    "version"
+  ],
+  "additionalProperties": false,
+  "patternProperties": {
+    "^.+$": {
+      "title": "Environment",
+      "properties": {
+        "init": {
+          "title": "Init command",
+          "description": "",
```
hnnasit, line 22: Noticed the `init` command description is not picked up. Is that expected?

schema/schema.py
Leo10Gama: Trimming off newlines and extra spaces from the ends would probably be best in that case. If there are other weird formatting issues we come across afterwards, we can address them as they come up, such as with the specific command schemas.

schema/samcli.json
Leo10Gama: Yup. From what I've found, there is no "help" parameter for the init command. 
Looking a little deeper, though, I can see there are other fields in that area that might be just as good for giving a description. I'll see if I can implement one of them to make sure all description fields are populated 👍

lucashuy (approved):

hnnasit (approved):

PR #5485, opened 2023-07-11
fix: use StringIO instead of BytesIO with StreamWriter

mndeveci:
Related to https://github.com/aws/aws-sam-cli/pull/5427
N/A

With recently changed the way to write into stream to fix some issues related to encoding (see PR above). That changed caused some issues with ECRUpload class, due to discrepancy of the `write` method between `BytesIO` vs `StringIO`. `BytesIO` instance accepts byte array where `StringIO` accepts string.

This resolves issue by using `StringIO` stream instead of `BytesIO`. This PR also adds some typing for the inner `stream` instance of the `StreamWrapper` class to check for other usages in the code base.

Manually running `sam local` tests to validate the change.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -7,7 +7,7 @@ import logging
 import os
 from enum import Enum
 from pathlib import Path
-from typing import IO, Any, Dict, List, Optional, Tuple, Type, cast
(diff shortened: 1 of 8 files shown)


jfuss (approved):
Thank you!!!!!!

mildaniel (approved):

PR #5500, opened 2023-07-12
chore: add reproducible-win

mndeveci:
Adds `update-reproducible-win-reqs` Makefile target as well as `requirements/reproducible-win.txt` which will be used to generate MSI installer during the release.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hnnasit (commented):
Should we also update our GH action that runs this?

mndeveci:
> Should we also update our GH action that runs this?

Yeah I've checked that and it already runs a parent target (`update-reproducible-reqs`) which is also updated in this PR.

hnnasit (approved):

hawflau (approved):
feels like the reproducible files should be categorized by Python version instead of OS. But categorizing by OS is also fine. I guess the benefit of categorizing by OS is when we upgrade Python version in the future, it's easier for us to do it per OS.

requirements/reproducible-win.txt
```diff
@@ -0,0 +1,785 @@
+#
```
mildaniel, line 1: Is this file different from the MacOS file? Looks like the command to create it is the same. Does it need to be run on Windows?

mndeveci:
Abandoning this PR now until we figure out that we need to run this on a Windows environment.

requirements/reproducible-win.txt
mildaniel: Also, if this should be run on a Windows machine, maybe we need to move it to a GHA?

mndeveci:
> feels like the reproducible files should be categorized by Python version instead of OS. But categorizing by OS is also fine. I guess the benefit of categorizing by OS is when we upgrade Python version in the future, it's easier for us to do it per OS.

I thought about this one as well, especially for build images, it might be better to generate these files per python version so that we can use it seamlessly in other places.

PR #5502, opened 2023-07-12
test: Step functions remote invoke integ tests

hnnasit:
This PR adds integration tests for step functions for the `remote invoke` command. The tests are skipped for now as the resource is not supported yet.
- `TestSFNPriorityInvoke` contains a single Step function for testing `remote invoke` commands when the `resource_id` is not provided and the resource is invoked by default.
- Tests are added to `TestMultipleResourcesInvoke` and `TestNestedTemplateResourcesInvoke` with different combinations of commands and resources.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -131,26 +202,38 @@ def setUpClass(cls):
         cls.stack_name = f"{TestMultipleResourcesInvoke.__name__}-{uuid.uuid4().hex}"
         cls.create_resources_and_boto_clients()
 
-    def test_invoke_empty_event_provided(self):
-        command_list = self.get_command_list(stack_name=self.stack_name, resource_id="EchoEventFunction")
+    @parameterized.expand([("EchoEventFunction", {})])
```
mndeveci, line 205: We are going to add SFN into here once it is enabled, is that a right understanding?

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -245,32 +378,46 @@ def setUpClass(cls):
...
         remote_invoke_result_stdout = json.loads(remote_invoke_result.stdout.strip().decode())
-        self.assertEqual(remote_invoke_result_stdout, {"message": "Hello world"})
+        self.assertEqual(remote_invoke_result_stdout, expected_response)
 
-    def test_invoke_with_only_event_provided(self):
+    @parameterized.expand(
+        [
+            ("ChildStack/HelloWorldFunction", '{"key1": "Hello", "key2": "world"}', {"message": "Hello world"}),
+            # Skip remote invoke Step function integration tests
+            # ("ChildStack/HelloWorldStateMachine", '{"key1": "Hello", "key2": "world"}', "World"),
```
mndeveci, line 400: Just thinking out loud, can we use the dictionary we use in factory implementation to decide whether certain test group should be run or skipped? In that sense we don't need to track to uncomment these lines.

mndeveci (commented):
LGTM in general, left couple of questions to see if we can automate test skipping.

tests/integration/remote/invoke/test_remote_invoke.py
hnnasit: Initially I wanted to add that here, but the SFN resource used here requires passing in an event to run the Lambda functions. I can remove the `parameterized.expand` as it's not required here.

tests/integration/remote/invoke/test_remote_invoke.py
hnnasit: Good point, let me think  about changing it to a `skipIf` so that it's easier to maintain.

tests/integration/remote/invoke/remote_invoke_integ_base.py
```diff
@@ -58,6 +58,8 @@ def create_resources_and_boto_clients(cls):
         }
         cls.cfn_client = get_boto_client_provider_with_config()("cloudformation")
         cls.lambda_client = get_boto_client_provider_with_config()("lambda")
+        cls.stepfunctions_client = get_boto_client_provider_with_config()("stepfunctions")
+        cls.xray_client = get_boto_client_provider_with_config()("xray")
```
hawflau, line 62: why do we need to call `get_boto_client_provider_with_config()` 4 times?

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -12,13 +14,13 @@
...
 @pytest.mark.xdist_group(name="sam_remote_invoke_single_lambda_resource")
-class TestSingleResourceInvoke(RemoteInvokeIntegBase):
+class TestSingleLambdaInvoke(RemoteInvokeIntegBase):
     template = Path("template-single-lambda.yaml")
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.stack_name = f"{TestSingleResourceInvoke.__name__}-{uuid.uuid4().hex}"
+        cls.stack_name = f"{TestSingleLambdaInvoke.__name__}-{uuid.uuid4().hex}"
```
hawflau, line 23: is it better to use `cls` instead of `TestSingleLambdaInvoke`? (considering the amount of code needed to change if you change the class name)
```suggestion
        cls.stack_name = f"{cls.__name__}-{uuid.uuid4().hex}"
```

tests/integration/remote/invoke/remote_invoke_integ_base.py
hnnasit: Good callout, can create a variable and create a single call.

tests/integration/remote/invoke/test_remote_invoke.py
hnnasit: I can update the other test class names as well.

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -116,6 +118,75 @@ def test_invoke_response_json_output_format(self):
...
 
+@skip("Skip remote invoke Step function integration tests")
+@pytest.mark.xdist_group(name="sam_remote_invoke_sfn_resource_priority")
+class TestSFNPriorityInvoke(RemoteInvokeIntegBase):
+    template = Path("template-step-function-priority.yaml")
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.stack_name = f"{TestSFNPriorityInvoke.__name__}-{uuid.uuid4().hex}"
```
hawflau, line 129:
```suggestion
        cls.stack_name = f"{cls.__name__}-{uuid.uuid4().hex}"
```

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -131,26 +202,38 @@ def setUpClass(cls):
...
+                '{"key1": "Hello", "key2": "serverless", "key3": "world"}',
+                {"message": "Hello world"},
+            ),
+            ("EchoCustomEnvVarFunction", '{"key1": "Hello", "key2": "serverless", "key3": "world"}', "MyOtherVar"),
+            (
+                "EchoEventFunction",
+                '{"key1": "Hello", "key2": "serverless", "key3": "world"}',
+                {"key1": "Hello", "key2": "serverless", "key3": "world"},
+            ),
+            # Skip remote invoke Step function integration tests
```
hawflau, line 227: is this not needed or will be un-commented in the future?

hnnasit: This will be un-commented once step functions is supported for `remote invoke`.  Trying to think of a way to not have to comment them and skip them based on the resource-type.

tests/integration/remote/invoke/test_remote_invoke.py
hnnasit: Addressed this in the new commit, used the factory dict so that only that needs to be updated when new resources support is added.

tests/integration/remote/invoke/test_remote_invoke.py
```diff
@@ -396,11 +392,12 @@ def test_invoke_empty_event_provided(self, resource_id, expected_response):
...
         [
             ("ChildStack/HelloWorldFunction", '{"key1": "Hello", "key2": "world"}', {"message": "Hello world"}),
-            # Skip remote invoke Step function integration tests
-            # ("ChildStack/HelloWorldStateMachine", '{"key1": "Hello", "key2": "world"}', "World"),
+            ("ChildStack/HelloWorldStateMachine", '{"key1": "Hello", "key2": "world"}', "World"),
         ]
     )
     def test_invoke_with_event_provided(self, resource_id, event, expected_response):
+        if self.stack_resource_summaries[resource_id].resource_type not in self.supported_resources:
+            pytest.skip()
```
mndeveci, line 400: Thanks for addressing this, can we also add message into `skip` call why this is been skipped. If you can't find it with `pytest.skip` there should be also one with `self.skip` (or something like that).

hnnasit: Yes can do  :+1:

mndeveci (approved):

hawflau (approved):

PR #5504, opened 2023-07-12
feat: Generate `sam deploy` schema

Leo10Gama:
With this change, the schema for `sam deploy` will be added to the JSON schema. It will serve to validate the parameters in samconfig for the deploy command.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
...k.\n* tags:\nList of tags to associate with the stack.\n* parameter_overrides:\nString that contains AWS CloudFormation parameter overrides encoded as key=value pairs.\n* signing_profiles:\nA string that contains Code Sign configuration parameters as FunctionOrLayerNameToSign=SigningProfileName:SigningProfileOwner Since signing profile owner is optional, it could also be written as FunctionOrLayerNameToSign=SigningProfileName\n* no_progressbar:\nDoes not showcase a progress bar when uploading artifacts to S3 and pushing docker images to ECR\n* capabilities:\nList of capabilities that one must specify before AWS Cloudformation can create certain stacks.\n\nAccepted Values: CAPABILITY_IAM, CAPABILITY_NAMED_IAM, CAPABILITY_RESOURCE_POLICY, CAPABILITY_AUTO_EXPAND.\n\nLearn more at: https://docs.aws.amazon.com/serverlessrepo/latest/devguide/acknowledging-application-capabilities.html\n* profile:\nSelect a specific profile from your credential file to get AWS credentials.\n* region:\nSet the AWS Region of the service. (e.g. us-east-1)\n* beta_features:\nEnable/Disable beta features.\n* debug:\nTurn on debug logging to print debug message generated by AWS SAM CLI and display timestamps.",
```
moelasmar, line 146: why do we add the command parameters here in the command description?

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                },
+                "disable_rollback": {
+                  "title": "disable_rollback",
+                  "type": "boolean",
+                  "description": "Preserves the state of previously provisioned resources when an operation fails."
+                },
+                "on_failure": {
+                  "title": "on_failure",
+                  "type": "string",
+                  "description": "Provide an action to determine what will happen when a stack fails to create. Three actions are available:\n\n- ROLLBACK: This will rollback a stack to a previous known good state.\n\n- DELETE: The stack will rollback to a previous state if one exists, otherwise the stack will be deleted.\n\n- DO_NOTHING: The stack will not rollback or delete, this is the same as disabling rollback.\n\nDefault behaviour is ROLLBACK.\n\n\n\nThis option is mutually exclusive with --disable-rollback/--no-disable-rollback. You can provide\n--on-failure or --disable-rollback/--no-disable-rollback but not both at the same time.",
```
moelasmar, line 196: do we have an idea on how this long description will look like in the customers IDEs?

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                  "type": "string",
+                  "description": "Environment name specifying default parameter values in the configuration file.",
+                  "default": "default"
+                },
+                "config_file": {
+                  "title": "config_file",
+                  "type": "string",
+                  "description": "Configuration file containing default parameter values.",
+                  "default": "samconfig.toml"
+                },
```
moelasmar, line 160: can the customer set these parameters in the config file, and how it will work if they can set them ??

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                  "description": "Automatically create and delete ECR repositories for image-based functions in non-guided deployments. A companion stack containing ECR repos for each function will be deployed along with the template stack. Automatically created image repositories will be deleted if the corresponding functions are removed."
+                },
+                "metadata": {
+                  "title": "metadata",
+                  "type": "string",
+                  "description": "Map of metadata to attach to ALL the artifacts that are referenced in the template."
+                },
+                "notification_arns": {
+                  "title": "notification_arns",
+                  "type": "array",
```
moelasmar, line 266: can we set items section to the schema here to let customer knows that the values of the list items should be string?

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                  "description": "A string that contains Code Sign configuration parameters as FunctionOrLayerNameToSign=SigningProfileName:SigningProfileOwner Since signing profile owner is optional, it could also be written as FunctionOrLayerNameToSign=SigningProfileName"
+                },
+                "no_progressbar": {
+                  "title": "no_progressbar",
+                  "type": "boolean",
+                  "description": "Does not showcase a progress bar when uploading artifacts to S3 and pushing docker images to ECR"
+                },
+                "capabilities": {
+                  "title": "capabilities",
+                  "type": "array",
```
moelasmar, line 291: can we set items section to the schema here to let customer knows that the values of the list items should be string?

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                "guided": {
+                  "title": "guided",
+                  "type": "boolean",
+                  "description": "Specify this flag to allow SAM CLI to guide you through the deployment using guided prompts."
+                },
+                "template_file": {
+                  "title": "template_file",
+                  "type": "string",
+                  "description": "AWS SAM template which references built artifacts for resources in the template. (if applicable)",
+                  "default": "template.[yaml|yml|json]"
```
hawflau, line 170: Is `template.[yaml|yml|json]` really the default value? or a symbol representing that it will look up these three files in certain order?

schema/samcli.json
Leo10Gama: When it appears as a "description", it also comes up as a tooltip in some IDEs (like VSCode). At least from testing around with it, I found it helpful to include them here so I could know which parameters I was allowed to have under each command

schema/samcli.json
Leo10Gama: Some IDEs like VSCode seem to make smaller, scrollable windows pop up so that things can be scrolled through. This parameter specifically, in VSCode, looks like this:
image

schema/samcli.json
Leo10Gama: Yup, I was unaware of that parameter for arrays, we can include that as well 👍

schema/samcli.json
Leo10Gama: It is, yes. I thought it weird, but that's how the default value is stored in the command, I assume so it's more readable under the CLI as options.

schema/samcli.json
```diff
@@ -143,9 +143,175 @@
...
+                  "title": "confirm_changeset",
+                  "type": "boolean",
+                  "description": "Prompt to confirm if the computed changeset is to be deployed by SAM CLI."
+                },
+                "disable_rollback": {
+                  "title": "disable_rollback",
+                  "type": "boolean",
+                  "description": "Preserves the state of previously provisioned resources when an operation fails."
+                },
+                "on_failure": {
```
lucashuy, line 193: Some of the parameters are incompatible with each other, should we hint towards this at the config level? (we might have already had this conversation before too)

Leo10Gama: We did have this conversation, yea. I think leaving it as is should be fine, both because if the user tries to run the command with those params anyway, the CLI will throw them an error saying they can't, and they can resolve it on their own; and also because that logic is more tied to the logic of the command itself, rather than being something we can find in the click option.

schema/samcli.json
Leo10Gama: After some testing, it seems like those two parameters will get ignored anyways in the config file, though the debug menu still indicates that it's being run with the provided values. I'll include a check to disallow those parameters, since they'll otherwise cause more confusion about what's being read from where

schema/schema.py
```diff
@@ -136,7 +139,8 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
         formatted_param_type = param_type or "string"
 
     formatted_param: SamCliParameterSchema = SamCliParameterSchema(
-        param.name or "", formatted_param_type, clean_text(param.help or "")
+        param.name or "", formatted_param_type, clean_text(param.help or ""),
+        items="string" if formatted_param_type == "array" else None
```
moelasmar: will it be always an array of Strings ?

Leo10Gama: Yes. From looking through the parameters, it seems like all parameters that can be lists (only `notification_arns` and `capabilities`) have their items as strings.

lucashuy (approved):

schema/schema.py
moelasmar: I will not block the PR on this, but we should think on a away to make it generic, as in future we can add new parameters that accept an array of different types than Strings.

moelasmar (approved):

schema/schema.py
Leo10Gama: Yup, will look into investigating click a little more to see if we can pull those data types dynamically 👍

PR #5525, opened 2023-07-14
fix: Convert paths to POSIX paths before checking in esbuild handling logic

lucashuy:
Windows paths contain an escape backslash (`\\`). This needs to be turned into a POSIX path before being parsed/saved into the template file.

Makes sure that the string path from the helper method is a POSIX path (`/` instead of `\\`) before parsing.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/lib/build_module/test_bundler.py
+++ b/tests/unit/lib/build_module/test_bundler.py
@@ -1,10 +1,11 @@
 from pathlib import Path
-from unittest import TestCase
+from unittest import TestCase, skipIf
 from unittest.mock import patch, Mock
 
 from parameterized import parameterized
 
 from samcli.lib.build.bundler import EsbuildBundlerManager
+from tests.testing_utils import IS_WINDOWS
 from tests.unit.commands.buildcmd.test_build_context import DummyStack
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/lib/build_module/test_bundler.py
```diff
@@ -240,3 +241,22 @@ def test_update_function_handler(self):
...
+        result_path = EsbuildBundlerManager(Mock())._get_path_and_filename_from_handler(input_path)
+
+        self.assertEqual(result_path, expected_path)
+
+    @parameterized.expand(
+        [
+            ("\\opt\\my\\path\\handler.handler", "/opt/my/path/handler.js"),
+        ]
+    )
+    @skipIf(not IS_WINDOWS, "Skipping POSIX converting logic since WindowsPath is not available on unix systems")
```
lucashuy, line 258: unix systems don't have the `WindowsPath` object, making things a bit difficult to test. Instead, I just skipped this unit test if its not running on Windows.

moelasmar: I got confused .. are we going to skip running this test case in windows or we want to run it only on windows. this condition `not IS_WINDOWS` is confusing me.

lucashuy: We'd want to only run if it is Windows. The `IS_WINDOWS` constant when used with a `not` in the `skipIf` does make it a little weird to read out.

mndeveci (approved):

moelasmar (approved):

Issue #5432, opened 2023-06-30
Bug: sam sync - InvalidSamDocumentException


CoshUS:
### Description:

sam sync --watch does not handle InvalidSamDocumentException gracefully during infra sync.

### Steps to reproduce:

1. Run sync watch on a template containing AWS::Serverless::Api and has StageName specified.
2. Wait for sync watch to complete initial infra sync
3. Update template and remove StageName from template
4. Save changes and trigger infra sync
5. Watch errors out and stops execution with `InvalidSamDocumentException`

### Observed result:

Code trigger
```
sam sync --watch --no-dependency-layer

The SAM CLI will use the AWS Lambda, Amazon API Gateway, and AWS StepFunctions APIs to upload your code without
performing a CloudFormation deployment. This will cause drift in your CloudFormation stack.
**The sync command should only be used against a development stack**.

Queued infra sync. Waiting for in progress code syncs to complete...
Starting infra sync.
Manifest is not changed for (IngestionFunction, VersionCheckFunction), running incremental build
Building codeuri: E:\...\lambda runtime: python3.8 metadata: {} architecture: x86_64 functions: IngestionFunction, VersionCheckFunction
Running PythonPipBuilder:CopySource
Running PythonPipBuilder:CopySource

Build Succeeded

Successfully packaged artifacts and wrote output template to file C:\Users\oo0SP\AppData\Local\Temp\tmpzonf0myr.
Execute the following command to deploy the packaged template
sam deploy --template-file C:\Users\...\AppData\Local\Temp\tmpzonf0myr --stack-name <YOUR STACK NAME>

Template haven't been changed since last deployment, skipping infra sync...
CodeTrigger not created as CodeUri or DefinitionUri is missing for ApiGatewayApi.
Skipped infra sync as the local template is in sync with the cloud template.
Required code syncs are queued up.
Syncing Lambda Function IngestionFunction...
Syncing Lambda Function VersionCheckFunction...
Finished syncing Lambda Function VersionCheckFunction.
Finished syncing Lambda Function IngestionFunction.
Queued infra sync. Waiting for in progress code syncs to complete...
Starting infra sync.
Failed to sync infra. Code sync is paused until template/stack is fixed.
Traceback (most recent call last):
  File "C:\Program Files\Amazon\AWSSAMCLI\runtime\lib\site-packages\samcli\lib\samlib\wrapper.py", line 73, in run_plugins
[...]

PR #5527, opened 2023-07-14
fix: don't run sam transform when getting (nested) stack information

mndeveci:
#5432

In `sam sync`, when we are getting whole stack information by calling `stacks = SamLocalStackProvider.get_stacks(self._template)[0]`, which runs SAMT automatically.

With this change, we are adding `use_sam_transform` to `SamLocaStackProvider.get_stacks` so that we can list all stack information without running transform.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -32,4 +32,4 @@ regex!=2021.10.8
 tzlocal==3.0
 
 #Adding cfn-lint dependency for SAM validate
-cfn-lint~=0.77.9
+cfn-lint~=0.78.1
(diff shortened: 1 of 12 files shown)


hnnasit (commented):
Can the SAM-T transform be skipped for this function call as well https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/sync/watch_manager.py#L115?

hnnasit (approved):

mndeveci:
> Can the SAM-T transform be skipped for this function call as well https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/sync/watch_manager.py#L115?

Thanks for catching that out, updated!

mildaniel (approved):

Issue #5346, opened 2023-06-12
Bug: SAM CLI crashes when receiving error from Docker


mildaniel:
### Description:

AWS SAM CLI crashes when starting up a debugging session on a port that is already in use.

Related to: https://github.com/aws/aws-sam-cli/issues/5292#issuecomment-1582602340

### Steps to reproduce:

1. Start a debugging session, leave it running.
2. Attempt to start another debugging session

### Observed result:

Crash and dumped stack trace.

### Expected result:

AWS SAM CLI should fail gracefully with an appropriate message

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOS
3. `sam --version`: 1.86.1
4. AWS region: N/A

PR #5554, opened 2023-07-18
fix: Graceful error handling for port already in use

hnnasit:

#5346 

When debugging port is provided as `-d` option and the port is already in use, `sam cli` throws a trace error.

This PR handles the above exception gracefully with the following message `Ports are not available: exposing port TCP 127.0.0.1:<port_num> -> 0.0.0.0:0: listen tcp 127.0.0.1:<port_num>: bind: address already in use.`

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel (approved):
Thanks for this!

mndeveci (approved):

PR #5558, opened 2023-07-18
feat: Generate `sam local` schema

Leo10Gama:
With this change, the schema for `sam local` subcommands will be added to the JSON schema. It will serve to validate the parameters in samconfig for the `local` subcommands.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/samcli.json
```diff
@@ -187,9 +187,140 @@
...
+                "log_file": {
+                  "title": "log_file",
+                  "type": "string",
+                  "description": "File to capture output logs."
+                },
+                "layer_cache_basedir": {
+                  "title": "layer_cache_basedir",
+                  "type": "string",
+                  "description": "Specify the location basedir where the lambda layers used by the template will be downloaded to.",
+                  "default": "/Users/leogama/.aws-sam/layers-pkg"
```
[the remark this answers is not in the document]
Leo10Gama: It looks like the default value for it gets computed using `click.get_app_dir()` in [this section of the codebase](https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/local/cli_common/options.py#L12-L35). Since it's always calling `.joinpath()` with that last folder, **layers-pkg**, we could either parse that out in the generation as the default, or get rid of the default value altogether. Either way, this will probably need to be explicitly excepted as a special case.

lucashuy (approved):

schema/schema.py
```diff
@@ -145,7 +145,10 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
         items="string" if formatted_param_type == "array" else None,
     )
 
-    if param.default:
+    params_to_exclude_default = [
```
lucashuy, line 148: Would it be possible to centralize this special parameter handling somewhere so that we don't need to have snippets of custom handling everywhere?

Not blocking on this since its something we can change later if we encounter more cases/need to refactor.

Leo10Gama: Yup that's a good point, keeping everything centralized will definitely help 👍

schema/samcli.json
```diff
@@ -202,9 +332,160 @@
...
+                  "description": "Optional. Specifies how AWS SAM CLI manages \ncontainers for each function.\nTwo modes are available:\nEAGER: Containers for all functions are \nloaded at startup and persist between \ninvocations.\nLAZY:  Containers are only loaded when each \nfunction is first invoked. Those containers \npersist for additional invocations.",
+                  "enum": [
+                    "EAGER",
+                    "LAZY"
+                  ]
+                },
+                "debug_function": {
+                  "title": "debug_function",
+                  "type": "string",
+                  "description": "Optional. Specifies the Lambda Function logicalId to apply debug options to when --warm-containers is specified.  This parameter applies to --debug-port, --debugger-path, and --debug-args."
```
hnnasit, line 444: nit: Extra space between `specified.` and `This parameter`

hnnasit (approved):
Overall LGTM!

PR #5559, opened 2023-07-18
feat: Generate `sam validate` and `sam package` schema

Leo10Gama:
With this change, the schema for `sam validate` and `sam package` will be added to the JSON schema. It will serve to validate the parameters in samconfig for the `validate` and `package` commands.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/samcli.json
```diff
@@ -247,9 +279,96 @@
...
+                },
+                "force_upload": {
+                  "title": "force_upload",
+                  "type": "boolean",
+                  "description": "Indicates whether to override existing files in the S3 bucket. Specify this flag to upload artifacts even if they match existing artifacts in the S3 bucket."
+                },
+                "resolve_s3": {
+                  "title": "resolve_s3",
+                  "type": "boolean",
+                  "description": "Automatically resolve AWS S3 bucket for non-guided deployments. Enabling this option will also create a managed default AWS S3 bucket for you. If one does not provide a --s3-bucket value, the managed bucket will be used. Do not use --guided with this option."
```
lucashuy, line 334: > Do not use --guided with this option.

I'm guessing this comes from the deploy command (or rather the option was written with the deploy command in mind and we are reusing it here)

lucashuy (approved):

schema/samcli.json
Leo10Gama: Yup, that seems to be the case. It uses a decorator method to produce the `resolve_s3` click option, and the help text seems specifically catered to deploy. We could change that text if need be, to be more specific about "If running with deploy, do not use --guided", but I don't think it's completely necessary.

hnnasit (approved):

PR #5560, opened 2023-07-18
feat: Generate `sam init` schema

Leo10Gama:
With this change, the schema for `sam init` will be added to the JSON schema. It will serve to validate the parameters in samconfig for the init command.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


schema/samcli.json
```diff
@@ -23,9 +23,149 @@
...
+              "properties": {
+                "no_interactive": {
+                  "title": "no_interactive",
+                  "type": "boolean",
+                  "description": "Disable interactive prompting for init parameters. (fail if any required values are missing)"
+                },
+                "architecture": {
+                  "title": "architecture",
+                  "type": "string",
+                  "description": "Architectures for Lambda functions.\u001b[1m\n\nArchitectures: ['arm64', 'x86_64']\u001b[0m",
```
lucashuy, line 37: Will the unicode characters show up correctly in tool tips?

Leo10Gama: It shows up like this in VSCode: 
image
Based on the command line output, it seems like it's supposed to be bolded, but it's not appearing that way. For readability sake, and because different IDEs could process markdown differently, we could opt to remove instances of the `[1m` and `[0m` altogether?

lucashuy: Lets strip out the ANSI encoding from the text, not sure if there is a built in way of doing that so that we can avoid having manually replace things.

hnnasit (approved):

schema/schema.py
```diff
@@ -120,6 +123,8 @@ def clean_text(text: str) -> str:
     """Clean up a string of text to be formatted for the JSON schema."""
     if not text:
         return ""
+    ansi_regex = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
```
lucashuy, line 126: Could this be a constant at the file root instead? I'm sure that Python will optimize this out so its not calling compile on the regular expression every time this method is called, but just wanted to be safe

schema/schema.py
```diff
@@ -120,7 +127,9 @@ def clean_text(text: str) -> str:
     """Clean up a string of text to be formatted for the JSON schema."""
     if not text:
         return ""
-    return text.replace("\b", "").strip("\n").strip()
+    for char_to_delete in CHARS_TO_CLEAN:
+        text = text.replace(char_to_delete, "")
+    return text.strip("\n").strip()
```
lucashuy, line 132: NIce! Makes this easier to expand in the future.

lucashuy (approved):

Issue #5566, opened 2023-07-19
Bug: sam remote invoke - ModuleNotFoundError


cm-suzuki-junya:
### Description:
execute `sam remote invoke`.

an error occurred `Error: No module named 'mypy_boto3_lambda'`

### Steps to reproduce:
1. Execute `sam init`
2. Execute `sam deploy`
3. Execute `sam remote invoke`

sam init message result
```
    -----------------------
    Generating application:
    -----------------------
    Name: sam-app-remote
    Runtime: python3.9
    Architectures: x86_64
    Dependency Manager: pip
    Application Template: hello-world
    Output Directory: .
    Configuration file: sam-app-remote/samconfig.toml
```

### Observed result:
```
$ sam remote invoke --debug
2023-07-19 19:04:36,898 | Config file location: /xxxx/sam-app-remote/samconfig.toml
2023-07-19 19:04:36,899 | Loading configuration values from [default.['remote', 'invoke'].parameters] (env.command_name.section) in config file at '/xxxx/sam-app-remote/samconfig.toml'...
2023-07-19 19:04:36,900 | Configuration values successfully loaded.
2023-07-19 19:04:36,900 | Configuration values are: {'stack_name': 'sam-app-remote'}
2023-07-19 19:04:36,900 | Converting provided --output option value to Enum
2023-07-19 19:04:36,935 | Using config file: samconfig.toml, config environment: default
2023-07-19 19:04:36,936 | Expand command line arguments to:
2023-07-19 19:04:36,936 | --stack_name=sam-app-remote --output=RemoteInvokeOutputFormat.TEXT
2023-07-19 19:04:36,937 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2023-07-19 19:04:36,984 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2023-07-19 19:04:36,986 | Unable to find Click Context for getting session_id.
2023-07-19 19:04:36,986 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': '467cc95e-b86a-4fa8-ac84-1afb86a1556e', 'installationId': '4e91b0f1-0a87-4da7-95f9-c9c6301a0dc0', 'sessionId': 'a02ab33d-bd73-4527-83ed-02333e1f4cf2',
[...]

PR #5570, opened 2023-07-19
fix: Add type checking on condition for mypy-boto3

hnnasit:

#5566 

`mypy_boto3_lambda` which is used for type checking purposes breaks the command. `boto3-stubs` is a dev dependency and not included in the prod dependencies.

Adds a condition to only import this for type checking.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/remote_invoke/lambda_invoke_executors.py
+++ b/samcli/lib/remote_invoke/lambda_invoke_executors.py
@@ -4,6 +4,7 @@ Remote invoke executor implementation for Lambda
 import base64
 import json
 import logging
+import typing
 from abc import ABC, abstractmethod
 from json import JSONDecodeError
 from typing import cast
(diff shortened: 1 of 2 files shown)


hawflau (commented):
why don't we just move the boto3-stub requirement to base.txt?

hnnasit (commented):
Yeah, we could add this to `base.txt`.

hnnasit (commented):
There are some issues with updating the reproducible files after adding the dependency to `base.txt` due to the `pyyaml` issues. We will address moving the dependency to `base.txt` in a separate PR to unblock `remote invoke` regression.

hawflau (approved):

mndeveci (approved):

Issue #5543, opened 2023-07-17
Bug: Error installing `PyYAML==5.4.1` as a dependency for `aws-sam-cli`


realFranco:
### Description:

I am trying to install AWS SAM CLI, and I am encountering an error when trying to install the PyYAML dependency. Specifically, when I try to install version 5.4.1 of PyYAML.

Some references:

- https://github.com/yaml/pyyaml/issues/724.

### Steps to reproduce:

```
# Just install the package `aws-sam-cli` on a host that currently has `python3` and `pip3`.

python3 -V
Python 3.10.12

pip3 -V
pip 23.2 from /usr/lib/python3.10/site-packages/pip (python 3.10)

pip3 install --no-cache-dir \
        aws-sam-cli==1.82.0
```

I pick that specific version becuase of business requirements, but will affect all the versions that has this constraint:

`PyYAML>=5.4.1,==5.*`
  - see: https://github.com/aws/aws-sam-cli/blob/develop/requirements/base.txt#L10

### Observed result:

```
# pip3 install --no-cache-dir aws-sam-cli==1.82.0


Collecting aws-sam-cli==1.82.0
  Downloading aws_sam_cli-1.82.0-py3-none-any.whl (5.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.4/5.4 MB 5.5 MB/s eta 0:00:00
Collecting chevron~=0.12 (from aws-sam-cli==1.82.0)
  Downloading chevron-0.14.0-py3-none-any.whl (11 kB)
Collecting click~=8.0 (from aws-sam-cli==1.82.0)
  Obtaining dependency information for click~=8.0 from https://files.pythonhosted.org/packages/22/b3/1da4ea0efa2e5ae410a347be614162ca08bd24a84059938aa5122d1e751b/click-8.1.5-py3-none-any.whl.metadata
  Downloading click-8.1.5-py3-none-any.whl.metadata (3.0 kB)
Collecting Flask<2.1 (from aws-sam-cli==1.82.0)
  Downloading Flask-2.0.3-py3-none-any.whl (95 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 95.6/95.6 kB 4.5 MB/s eta 0:00:00
[...]

PR #5573, opened 2023-07-19
chore: Bump pyyaml to version 6, revert pyyaml pins

mildaniel:

https://github.com/aws/aws-sam-cli/issues/5543

Fix an issue where there are multiple dependency conflicts causing issues with pip install.

We've removed our dependency on `serverlessrepo` so that we can bump `pyyaml`.

N/A


moelasmar (approved):

mndeveci (approved):

requirements/reproducible-mac.txt
```diff
@@ -695,17 +704,12 @@ sarif-om==1.0.4 \
     --hash=sha256:539ef47a662329b1c8502388ad92457425e95dc0aaaf995fe46f4984c4771911 \
     --hash=sha256:cd5f416b3083e00d402a92e449a7ff67af46f11241073eea0461802a3b5aef98
     # via cfn-lint
-serverlessrepo==0.1.10 \
```
realFranco, line 698: Hello, thank you for your contribution.

Some days ago, I was thinking:

- "why `aws-sam-cli` is utilizing `serverlessrepo` considering the current status (only-read | public archive)?"

I have some questions now:

 - How do you make sure that the package `serverlessrepo` is not been in use?;
 - Is there any other package with this same behaviour? (unused).
 
 Seems to me that another non-used package soon or later "could" lead to another breaking. 👋

mildaniel: Hi @realFranco, the `serverlessrepo` package was in fact being used by the `sam publish` command. We brought it in to be vendored by us yesterday https://github.com/aws/aws-sam-cli/pull/5572. All other dependencies are being used.

realFranco: Nice, thank you for the clarification.

PR #5584, opened 2023-07-20
chore: Move boto3-stubs dependency from dev.txt to base.txt

hnnasit:
`boto3-stubs` is a dev dependency and to use the boto3 objects for type checking, they need to be imported conditionally. If the import is not conditional, it can break production installers as the testing uses dev-dependencies and this can get missed.

This PR moves `boto3-stubs` to base.txt for making the boto3 type checks more sustainable. I built a mac binary with the new changes and tested the functionality for `remote invoke` and `sam deploy` companion stack and it works as expected.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -8,7 +8,6 @@ pytest-cov==4.1.0
 # mypy adds new rules in new minor versions, which could cause our PR check to fail
 # here we fix its version and upgrade it manually in the future
 mypy==1.3.0
-boto3-stubs[apigateway,cloudformation,ecr,iam,lambda,s3,schemas,secretsmanager,signer,stepfunctions,sts,xray]==1.28.2
 types-pywin32==306.0.0.2
 types-PyYAML==6.0.12
 types-chevron==0.14.2.4
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -17,7 +17,7 @@ types-Pygments==2.15.0.1
 types-colorama==0.4.15.11
 types-dateparser==1.1.4.9
 types-docutils==0.20.0.1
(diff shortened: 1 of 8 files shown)


mndeveci (approved):
LGTM, can we confirm generated artifacts with E2E tests that it doesn't have any side effects?

hnnasit:
> LGTM, can we confirm generated artifacts with E2E tests that it doesn't have any side effects?

I did some testing around `remote invoke` and `sam deploy` commands after generating an artifact with the new changes. I can run the E2E integ tests to verify as well before merging.

requirements/dev.txt
```diff
@@ -8,7 +8,6 @@ pytest-cov==4.1.0
 # mypy adds new rules in new minor versions, which could cause our PR check to fail
 # here we fix its version and upgrade it manually in the future
 mypy==1.3.0
-boto3-stubs[apigateway,cloudformation,ecr,iam,lambda,s3,schemas,secretsmanager,signer,stepfunctions,sts,xray]==1.28.7
```
hawflau, line 11: not an immediate issue but we should think about it - I wonder if we should move all typing stubs to base.txt, considering that the same could happen for any typing stubs we are using. That also yield another questions:
- How do we test if we've included all dependencies used for type hints into base.txt?
- Should we consider striping all type hints as part of release process after passing type checks?

hawflau (approved):

requirements/dev.txt
hnnasit: I think that's a good callout and both are valid questions worth thinking.

PR #5604, opened 2023-07-24
Fix generation logic

Leo10Gama:
Previously, generation logic was not working due to the file handling the generation being named identically to the folder, causing import errors. In addition, some custom parameter types were failing in the newly-implemented checks for type names.

The file housing the generation logic has been renamed to `make_schema.py`. A new constant has been added to map the custom-typed parameters to their JSON schema representation type. This includes `parameter_overrides`, which can be provided as an array or a string.

Some of the types in the schema have been updated to better reflect their actual typing.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/schema/make_schema.py
+++ b/schema/make_schema.py
@@ -20,6 +20,15 @@ PARAMS_TO_EXCLUDE = [
 PARAMS_TO_OMIT_DEFAULT_FIELD = [
     "layer_cache_basedir"  # sets default to root directory to that of machine the schema is generated on
 ]
+EXCEPTION_OPTION_CLASS_MAPPER = {  # for params without proper class names, map them uniquely
+    "parameter_overrides": ["array", "string"],
+    "image_repository": "string",
+    "image_repositories": "array",
+    "metadata": "string",
+    "signing_profiles": "string",
+    "tags": "array",
+    "parameter": "array",
+}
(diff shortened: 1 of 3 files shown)


schema/make_schema.py
```diff
@@ -147,14 +156,16 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
...
     if not param:
         raise SchemaGenerationException("Expected to format a parameter that doesn't exist")
-    if not param.type.name:
-        raise SchemaGenerationException(f"Parameter {param} passed without a type")
+    if not param.type.name and param.name not in EXCEPTION_OPTION_CLASS_MAPPER.keys():
+        raise SchemaGenerationException(f"Parameter {param} passed without a type:\n{param.type}")
 
     param_type = param.type.name.lower()
-    formatted_param_type = ""
+    formatted_param_type: Any = None
```
lucashuy, line 163: Since this type can now take the type of a list, would it be valid for the JSON schema to have a single element list (eg. `["string"]`)? If a single element list is valid, then we might just be able to have things as lists of type(s). This way the `items` if check is more simple.

lucashuy (approved):

hnnasit (approved):

PR #5610, opened 2023-07-25
feat: Generate `sam delete` and `sam traces` schema

Leo10Gama:
With this change, the schema for `sam delete` and `sam traces` will be added to the JSON schema. It will serve to validate the parameters in samconfig for the `delete` and `traces` commands.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


hnnasit (approved):

schema/samcli.json
```diff
@@ -1206,9 +1247,60 @@
...
+                  "type": "string",
+                  "description": "Fetch specific trace by providing its id"
+                },
+                "output": {
+                  "title": "output",
+                  "type": "string",
+                  "description": "The formatting style of the command output. Following options are available:\n\nTEXT: Prints information as regular text with some formatting (default option)\n\nJSON: Prints each line as JSON without formatting",
+                  "enum": [
+                    "json",
+                    "text"
```
lucashuy, line 1264: question: Is this case sensitive? The description is upper case, but then the enum is lower case. If it doesn't matter, then I think that this is fine as is.

schema/samcli.json
```diff
@@ -1206,9 +1247,60 @@
...
+                  ]
+                },
+                "tail": {
+                  "title": "tail",
+                  "type": "boolean",
+                  "description": "Tail events. This will ignore the end time argument and continue to fetch events as they become available. If option --tail without a --name will pull from all possible resources"
```
lucashuy, line 1270: Does this mean that this command is suppose to have a `--name` parameter 🤔 ?

schema/samcli.json
Leo10Gama: Testing it quickly, it seems like case sensitivity doesn't affect it; it works with all of `["json", "JSON", "jSoN"]` when I tested.

schema/samcli.json
Leo10Gama: Not sure. In the CLI, it's only bunched with `[--start-time, --end-time, --output]`, and the only command that uses `--name` is `sam logs`. Should the help text be updated to remove this line? Or turn the `--tail` option into a standalone constructor so that the `logs` command can still have the extra description on it?

lucashuy: We should update this so that each command has it's own option, since the description here only applies to `logs`, and not `traces`. This might mean some other things might also need to change, but we can just capture that here.

lucashuy (approved):

PR #5619, opened 2023-07-26
fix: cors for local start-api

runk:

Alternative attempt similar to #4991. This solves https://github.com/aws/aws-sam-cli/issues/4161.

During `sam local start-api`, for the `AWS::Serverless::HttpApi` with configured CORS there's two issues.

- CORS headers currently not being returned to other than OPTIONS requests (when Origin header set)
- Multiple origins in template are not supported

Behaviour of `sam local start-api` is different from the production when it comes to handling CORS configuration of the `AWS::Serverless::HttpApi`. With this change local behaviour mimics what users experience in prod.

- Add CORS headers to all responses (if setup)
- Parse comma-separated list of allowed origins and do basic matching

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/providers/provider.py
```diff
@@ -494,8 +494,21 @@ def cors_to_headers(cors: Optional["Cors"]) -> Dict[str, Union[int, str]]:
         """
         if not cors:
             return {}
+
+        # cors.allow_origin can be either a single origin or comma separated list of origins
```
runk, line 498: I'm not 100% sure this is the right place to do matching. Would appreciate guidance if this has to go somewhere else.

runk:
Hi @jfuss @moelasmar, could you please review this PR and unblock CI pipeline?

moelasmar:
thanks @runk for raising this PR, could you please add integration test cases to cover this update.

runk:
> thanks @runk for raising this PR, could you please add integration test cases to cover this update.

Hey mate, can you please point me to "how to" guide or something? I really struggle to figure how to work with integration tests. My changes do not seem to be picked up by tests.. :( Feels like a cache or something holding state

runk:
@moelasmar hey bud, I believe this is ready for review - added integration tests.

samcli/lib/providers/provider.py
```diff
@@ -481,7 +481,7 @@ def binary_media_types(self) -> List[str]:
 
 class Cors(_CorsTuple):
     @staticmethod
-    def cors_to_headers(cors: Optional["Cors"]) -> Dict[str, Union[int, str]]:
+    def cors_to_headers(cors: Optional["Cors"], request_origin: str) -> Dict[str, Union[int, str]]:
```
lucashuy, line 484: nit: Can we have the optional parameter come after the required parameters?

samcli/lib/providers/provider.py
```diff
@@ -494,8 +494,28 @@ def cors_to_headers(cors: Optional["Cors"]) -> Dict[str, Union[int, str]]:
...
+
+        # Resource processing start here.
+        # The following code is based on the following spec:
+        # https://www.w3.org/TR/2020/SPSD-cors-20200602/#resource-processing-model
+
+        if not request_origin:
+            return {}
+
+        # cors.allow_origin can be either a single origin or comma separated list of origins
+        allowed_origins = cors.allow_origin.split(",") if cors.allow_origin else list()
```
lucashuy, line 506: Is it possible to have both of these strings (`"origin1,origin2"` and `"origin1, origin2"` with a space) be passed in as valid origins?

samcli/lib/providers/provider.py
lucashuy: This looks like a good place for this to live :)

tests/integration/local/start_api/test_start_api.py
```diff
@@ -1521,22 +1521,32 @@ class TestServiceCorsSwaggerRequests(StartApiIntegBaseClass):
...
 
-    @pytest.mark.flaky(reruns=3)
-    @pytest.mark.timeout(timeout=600, method="thread")
-    def test_cors_swagger_options(self):
-        """
-        This tests that the Cors are added to option requests in the swagger template
-        """
-        response = requests.options(self.url + "/echobase64eventbody", timeout=300)
-
+    def assertCors(self, response):
```
lucashuy, line 1524: nit: `snake_case`, the unittest stuff uses camel case, however the rest of our code use snake case.

tests/integration/local/start_api/test_start_api.py
```diff
@@ -1549,20 +1559,92 @@ class TestServiceCorsSwaggerRequestsWithHttpApi(StartApiIntegBaseClass):
...
+    """
+
+    template_path = "/testdata/start_api/template-cors.yaml"
+    binary_data_file = "testdata/start_api/binarydata.gif"
+
+    def setUp(self):
+        self.url = "http://127.0.0.1:{}/test".format(self.port)
+
+    def assertPresence(self, response):
+        expected_origin = response.request.headers.get("Origin")
```
lucashuy, line 1603: Would it be possible to pass `expected_origin` as a parameter instead of grabbing it from the `response.request` object? My thinking here was that this way, we can pass around concrete values to test against instead of testing the expected origin via the `request` object

lucashuy (commented):
Thanks for raising this PR! Left some comments

samcli/lib/providers/provider.py
runk: I do not think so, as this is coming from yaml parsers (providers?) in this format. I'll add a trim to be safe 👍

samcli/lib/providers/provider.py
runk: `request_origin` is actually optional as well - will fix

runk:
Hi @lucashuy 👋 Addressed your comments, please have another look.

samcli/lib/providers/provider.py
```diff
@@ -504,6 +504,7 @@ def cors_to_headers(cors: Optional["Cors"], request_origin: str) -> Dict[str, Un
 
         # cors.allow_origin can be either a single origin or comma separated list of origins
         allowed_origins = cors.allow_origin.split(",") if cors.allow_origin else list()
+        allowed_origins = [s.strip() for s in allowed_origins]
```
lucashuy, line 507: nit: Can we rename `s` to avoid single character variable names (even though its just used in a loop that is being assigned back to itself)

lucashuy (commented):
Thanks for making these changes! I left a minor comment and restarted the jobs

runk:
@lucashuy can you please unblock the pipeline

runk:
@lucashuy looks like it's good to go with CI pipeline finally green

runk:
@moelasmar can you please review it one more time?

samcli/lib/providers/provider.py
```diff
@@ -481,7 +481,7 @@ def binary_media_types(self) -> List[str]:
 
 class Cors(_CorsTuple):
     @staticmethod
-    def cors_to_headers(cors: Optional["Cors"]) -> Dict[str, Union[int, str]]:
+    def cors_to_headers(cors: Optional["Cors"], request_origin: Optional[str]) -> Dict[str, Union[int, str]]:
```
moelasmar, line 484: can you update the function docs, to add the new parameter

moelasmar (approved):
it looks good to me, thanks @runk for your contribution

lucashuy (approved):
Thanks for making these changes, and addressing comments

PR #5620, opened 2023-07-26
fix: Add type names to custom parameter types

Leo10Gama:
Since the SAM CLI uses custom types without names, we had to create a constant to map the name of the parameter to its set of possible values. This creates additional overhead when creating new custom parameter types.

The `name` fields of each of these types have now been populated, and as such, they are able to be read by the schema parser. This means we can include the expected type (or types) of the parameter to be read.

The `name` field of each SAM CLI custom type has been populated instead of being an empty string (not that it was being used to begin with).

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/schema/make_schema.py
+++ b/schema/make_schema.py
@@ -166,6 +166,9 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
             formatted_param_types.append("array")
         else:
             formatted_param_types.append(param_name or "string")
+    formatted_param_types = [
+        x for i, x in enumerate(formatted_param_types) if x not in formatted_param_types[:i]
+    ]  # deduplicate
 
     formatted_param: SamCliParameterSchema = SamCliParameterSchema(
         param.name or "",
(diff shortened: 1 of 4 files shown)


lucashuy (approved):

schema/make_schema.py
```diff
@@ -156,21 +147,25 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
     """
     if not param:
         raise SchemaGenerationException("Expected to format a parameter that doesn't exist")
-    if not param.type.name and param.name not in EXCEPTION_OPTION_CLASS_MAPPER.keys():
+    if not param.type.name:
         raise SchemaGenerationException(f"Parameter {param} passed without a type:\n{param.type}")
 
-    param_type = param.type.name.lower()
+    param_type = []
```
hnnasit, line 153: Do we need to update unit tests to test these changes?

Leo10Gama: Including them may be beneficial, yea, since we can verify the behaviour of having multiple allowable types as well 👍

schema/make_schema.py
```diff
@@ -166,6 +166,9 @@ def format_param(param: click.core.Option) -> SamCliParameterSchema:
             formatted_param_types.append("array")
         else:
             formatted_param_types.append(param_name or "string")
+    formatted_param_types = [
```
hnnasit, line 169: Discussed offline to use sets and calling sorted here for simplicity and consistency.

hnnasit (approved):

PR #5627, opened 2023-07-26
fix: Prevent stack trace when incorrect filetype is passed as config

Leo10Gama:
Previously, if a config file is provided without a supporting extension, a stack trace is thrown rather than just an error. This prompts the user to open an issue, which they should not do, since it is not something we can immediately address.

The `SamConfigFileReadException` class has been changed from inheriting from `Exception` to inheriting from `UserException`. That way, it is caught properly and is passed through the correct channels.

Any `SamConfigFileReadException` exceptions that are raised no longer prompt the user to open an issue about it

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/config/exceptions.py
+++ b/samcli/lib/config/exceptions.py
@@ -3,6 +3,9 @@ Exceptions to be used by samconfig.py
 """
 
 
+from samcli.commands.exceptions import UserException
+
+
 class SamConfigVersionException(Exception):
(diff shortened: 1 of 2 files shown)


mildaniel (commented):
Let's add an integration test that validates the exit code from the process is 1 instead of 255. Otherwise LGTM!

mildaniel (approved):

moelasmar (approved):

PR #5654, opened 2023-07-31
feat: Add schema validation tests

Leo10Gama:
To ensure the generated schema is correct, we need to test it against various config files. This will ensure that tests that should validate against the schema will validate, while tests that shouldn't validate against the schema won't.

A new test file, `test_schema_validation.py` has been created to validate various config files against the generated schema. In addition, a `testdata` folder has been added, with various hand-written config files to test each command, as well as prevent invalid schemas.

New tests have been added, to test that config files will and won't validate against the schema.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/unit/schema/test_schema_validation.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Read in and assert all files in passing_tests pass
+        for config_file_path in passing_tests_dir.iterdir():
+            config_file = FILE_MANAGER_MAPPER[config_file_path.suffix].read(Path(str(passing_tests_dir), config_file_path))
+            self.assertNotEqual(config_file, {}, f"Config file {config_file_path} should be read correctly")
+
+            try:
+                jsonschema.validate(config_file, self.schema)
+            except jsonschema.ValidationError as e:
+                tests_not_passing.append(f"<{config_file_path}: {e.message}>")
```
lucashuy, line 37: Should we stop the test immediately instead of appending to an error list (failing fast)?

tests/unit/schema/testdata/failing_tests/buildcmd_noparams.toml
lucashuy, line 1: Can we also include a YAML version? Doesn't have to be for all the commands, but just want to include one instance of a YAML config file without a `parameters`

tests/unit/schema/testdata/passing_tests/buildcmd.toml
```diff
@@ -0,0 +1,36 @@
...
+beta_features = false
+debug = false
+profile = "Profile"
+region = "Region"
+
+[other.build.parameters]
+skip_prepare_infra = true
+use_container = true
+parallel = true
+mount_with = "WRITE"
```
lucashuy, line 31: Would it be possible to test the typing of the parameters here (eg. `mount_with` is a string, not a list or an array)?

tests/unit/schema/test_schema_validation.py
Leo10Gama: Yup we definitely could. I implemented it this way so we could see if multiple tests were failing for the same or different reasons, but keeping things quicker may be ideal instead 👍

tests/unit/schema/testdata/passing_tests/buildcmd.toml
Leo10Gama: Yup, definitely. The only reason I didn't earlier is deciding on which parameters to test, since testing all of them could be tedious

hnnasit (approved):

tests/unit/schema/test_schema_validation.py
```diff
@@ -0,0 +1,53 @@
...
+        for config_file_path in os.listdir(passing_tests_dir):
+            config_file = FILE_MANAGER_MAPPER[Path(config_file_path).suffix].read(
+                Path(str(passing_tests_dir), config_file_path)
+            )
+            self.assertNotEqual(config_file, {}, f"Config file {config_file_path} should be read correctly")
+
+            try:
+                jsonschema.validate(config_file, self.schema)
+            except jsonschema.ValidationError as e:
+                self.assertTrue(False, f"File {config_file_path} not validating: {e.message}")
```
lucashuy, line 36: nit:
```suggestion
                self.fail(f"File {config_file_path} not validating: {e.message}")
```

lucashuy (approved):

lucashuy (approved):

tests/unit/schema/test_schema_validation.py
Leo10Gama: I wasn't aware of that method 👀

PR #5677, opened 2023-08-02
chore: auto update reproducible requirements when there is a dependency change

mndeveci:
N/A

Automation for updating pinned dependencies as well as adding pinned dependencies for MSI.

- Adds a new Makefile target for updating reproducible dependencies for Windows MSI.
- Adds a new GHA, which will only run if there is a change in `requirements/base.txt`, and that GHA will update our pinned dependencies in `reproducible-{os}.txt`.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


.github/workflows/update-reproducibles.yml
```diff
@@ -0,0 +1,42 @@
+name: Update reproducible requirements
+on:
+  pull_request:
+    branches: [develop]
+    paths:
+      - requirements/base.txt # run this GHA only if requirements file is changed
```
moelasmar, line 6: just to make sure, we do not need this GHA to be executed in case of the dependencies got updated, or their dependencies got updated. Dependabot should handle that, right?

.github/workflows/update-reproducibles.yml
```diff
@@ -0,0 +1,42 @@
+name: Update reproducible requirements
+on:
+  pull_request:
+    branches: [develop]
```
moelasmar, line 4: are we going to update the PR itself, or update the develop branch, then we need to merge the PR branch from the develop branch?

.github/workflows/update-reproducibles.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+      - run: make ${{ matrix.target }}
+      - name: Push changes
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}
+          git commit -am "Update reproducibles: ${{ matrix.target }}" || echo "nothing to commit"
```
lucashuy, line 41: Just a general comment, butI think I had the same conversation with a team member about auto generating a JSON schema file, and whether or not this should be a message instead to manually run the generation rules, or something that Github Actions will do for us in the same PR.

The wrench that is thrown here is that Windows repro generation would in theory need to happen on a Windows machine.

.github/workflows/update-reproducibles.yml
mndeveci: This will update the PR itself (see this commit for example in this PR: https://github.com/aws/aws-sam-cli/pull/5677/commits/32b350a69da0e7ea78b379bfc4487026a9392024)

But this GHA will only run if `requirements/base.txt` is updated (see line 6)

.github/workflows/update-reproducibles.yml
mndeveci: I wanted to keep our existing flow where we update these reproducibles when we update our requirements (for example when we bump SAMT version). If we run this for every PR there is a chance that we will have updates every day or so (for minor version updates), which might make harder for those PRs getting merged, and I will leave those updates to dependabot as it is been currently handled.

.github/workflows/update-reproducibles.yml
mndeveci: The reason we put these updates into GHA is purely needing a Windows machine to update reproducibles. If these were not OS specific (like we can generate one and use it for every installer), then I would keep the existing flow where we will have a Makefile target which can be run on any OS so that contributors can run it after they update any dependency that we have.

moelasmar (approved):

lucashuy (approved):

PR #5683, opened 2023-08-02
feat: Add schema generation to GitHub Actions

Leo10Gama:
With the new schema generation logic, we want to ensure that the script runs in such a way as to keep the schema up to date with the current changes.

A new GitHub Actions workflow has been created to run the schema generation script once per weekday, check if the schema has been altered in any way, and if it has, open a PR to make the necessary changes.

At least once every weekday, if the generated schema is changed in any way, a PR will be raised to update it.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


.github/workflows/build.yml
```diff
@@ -64,6 +66,27 @@ jobs:
...
+    permissions:
+      pull-requests: write
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        name: Install Python 3.9
+        with:
+          python-version: 3.9
```
lucashuy, line 81: nit: could we use the latest version of Python we support (3.11)?

lucashuy (approved):

hnnasit (approved):

Issue #5533, opened 2023-07-15
SAM error puts stack into indeterminate, undeletable state


garretwilson:
I'm using SAM version 1.88.0 on Windows 10.

If something goes wrong with a deployment (for example, the SAM bug in aws/serverless-application-model#3264), SAM leaves the stack in some `REVIEW_IN_PROGRESS` state that isn't deployed and but isn't deleted either. If I try to delete the stack using `sam delete`, it says:

> `Error: The input my-stack does not exist on Cloudformation in the region None`

I'm forced go into the CloudFormation console and manually delete the stack. I should never be forced to go into the console to do something because SAM put something into an indeterminate state. At the very least I should be able to delete the thing and start over from the command line.

(I haven't tried deleting from CloudFormation, but in any case I should be able to delete a failed deployment stack from SAM.)

hawflau:
@garretwilson Thanks for raising the issue. Marking this issue as a bug. We will prioritize fixing it.

PR #5687, opened 2023-08-02
fix: Delete stacks in REVIEW_IN_PROGRESS

lucashuy:

#5533

Stacks in `REVIEW_IN_PROGRESS` were not being deleted since the stack does not yet have a template. This meant that any stacks stuck in this state, were not delete-able with SAM CLI since we need that stack to find the file in S3 to delete.

Change sets contain the stack we can use to delete template files. This change uses the change set's template file to find the one in the S3 bucket to delete.

The helper method to return a template response, now returns the template body as a string (previous was returning a dictionary).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/delete/cfn_utils.py
```diff
@@ -20,8 +26,21 @@ def has_stack(self, stack_name: str) -> bool:
...
+            Name or ID of the stack
+
+        Returns
+        -------
+        bool
+            True if stack exists. False otherwise
+
+        Raises
+        ------
+        DeleteFailedError
```
moelasmar, line 41: nit: the exception name looks weird to be raised from `has_stack` function.

moelasmar: can we change the function name to be something like `can_delete_stack`

samcli/lib/delete/cfn_utils.py
```diff
@@ -121,11 +184,50 @@ def wait_for_delete(self, stack_name):
...
+        try:
+            change_sets: dict = self._client.list_change_sets(StackName=stack_name)
+        except (ClientError, BotoCoreError) as ex:
+            LOG.debug("Failed to perform boto call to fetch change sets")
+            raise FetchChangeSetError(stack_name=stack_name, msg=str(ex)) from ex
+
+        change_sets = change_sets.get("Summaries", [])
+
+        if len(change_sets) > 0:
+            change_set = change_sets[0]
```
moelasmar, line 226: is the change set in index 0 always the latest, do we need to add any sorting to the service call?

moelasmar (approved):
it looks good to me, I left some minor comment and a question.

samcli/lib/delete/cfn_utils.py
```diff
@@ -56,27 +71,56 @@ def has_stack(self, stack_name: str) -> bool:
...
             # If there are credentials, environment errors,
             # catch that and throw a delete failed error.
 
             LOG.error("Failed to fetch template for the stack : %s", str(e))
             raise FetchTemplateFailedError(stack_name=stack_name, msg=str(e)) from e
-
+        except FetchChangeSetError as ex:
+            raise FetchTemplateFailedError(stack_name=stack_name, msg=str(ex)) from ex
+        except NoChangeSetFoundError as ex:
+            msg = f"Failed to find a change set for stack {stack_name} to fetch the template"
```
hawflau, line 122: The stack name part in the msg might seem a bit redundant as the exception will also use the `stack_name` attr in formatting the final message.

hawflau (approved):

samcli/lib/delete/cfn_utils.py
lucashuy: Sure, we can update this method to better represent what could happen

samcli/lib/delete/cfn_utils.py
lucashuy: Good catch, will remove the dup stack name

samcli/lib/delete/cfn_utils.py
lucashuy: Just double checked, the change sets are sorted in ascending order if there are more than one. However, for the purposes of this PR we are going to delete the first change set and the template associated with it.

The others will be left to the user as we'll want to make sure the stack is deleted so that the customer is unblocked.

I'll make a change to add a log message to let them know that there may be lingering artifacts.

PR #5716, opened 2023-08-04
Pass function_name to lambda_builders

jysheng123:
5369

Not necessary, but a QOL change for developers running -parallel builds so they can understand logs easier

Adds the function_name at the start of each log for lambda_builder so developer can differentiate between each log.

Logs from Lambda-Builder will look slightly different in the future

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jysheng123:
No need to pass in value, existing build_logical_id serves the purpose needed

PR #5719, opened 2023-08-04
test: AppVeyor Jobs Running Against Nightly Binaries

mildaniel:

N/A

To run integration tests against production binaries.

Creates new AppVeyor configurations that download the nightly binaries on Windows and Linux and then run the integration test suite against them.

N/A

#### Mandatory Checklist

- [] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


appveyor-linux-binary.yml
```diff
@@ -0,0 +1,295 @@
...
+  # Install AWS CLI
+  - sh: "virtualenv aws_cli"
+  - sh: "./aws_cli/bin/python -m pip install awscli"
+  - sh: "PATH=$(echo $PWD'/aws_cli/bin'):$PATH"
+
+  # Install pytest
+  - sh: "pip3 install -r requirements/pre-dev.txt"
+  - sh: "pip3 install -r requirements/dev.txt"
+  - sh: "pip3 install -r requirements/base.txt"
+  - sh: "pytest --version"
```
moelasmar, line 77: I think it is better if we installed these libraries in a virtual environment to make sure that we isolate our test cases from any libraries that got installed here. I also think if just install the libraries that are required for the integration testing only.

moelasmar (approved):

appveyor-linux-binary.yml
mildaniel: I changed it to download the dependencies into a virtual env, but we require a bunch of these dependencies to run the integration tests. Might not be worth trying to narrow it down right now.

appveyor-linux-binary.yml
```diff
@@ -0,0 +1,296 @@
...
+    if [[ -n $BY_CANARY ]];
+      then echo Logging in Public ECR; aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws;
+    fi"
+
+
+build_script:
+  - curl -L https://github.com/aws/aws-sam-cli/releases/download/sam-cli-nightly/aws-sam-cli-linux-x86_64.zip -o aws-sam-cli-linux-x86_64.zip
+  - unzip aws-sam-cli-linux-x86_64.zip -d sam-installation
+  - sudo ./sam-installation/install
+  - sudo mv /usr/local/bin/sam-nightly /usr/local/bin/sam
```
mndeveci, line 155: nit: worth adding some comment that we are overriding executable name (`sam-nightly`) with `sam` here.

appveyor-linux-binary.yml
```diff
@@ -0,0 +1,296 @@
+version: 1.0.{build}
```
mndeveci, line 1: There are too many `appveyor*` files at the root level, we might want to move them under `.appveyor` folder.

And there is some level of duplication between these files, I wonder if we can use some sort of "import" and move common configuration into separate files.

mndeveci (approved):
Thanks for working on this, left some unblocking comments.

Issue #2934, opened 2021-06-06
Skip Docker Lambda image builds when ImageUri is a valid ECR location


alexisfacques:
### Describe your idea/feature/enhancement

When configuring a lambda / serverless function of which the `PackageType` is `Zip`, SAM preemptively checks whether the resource `CodeUri` is a S3 location or not.

```
  # This function is already packaged and will to be ignored by SAM
  Function:
    Type: AWS::Serverless::Function
    Properties:
      PackageType: Zip
      CodeUri: s3://bucket/function.zip
      ...
```

If so, SAM will not attempt to build said lambda, and will show a warning to the user: 

```
% sam build
The resource AWS::Serverless::Function 'Function' has specified S3 location for CodeUri. It will not be built and SAM CLI does not support invoking it locally.
```

I would expect SAM to do the same when attempting to build a `PackageType: Image` lambda function; checking whether or not the resource property `ImageUri` is a valid ECR URL before attempting to build said images locally, yet this feature is not supported by SAM today:

```
  # SAM should not attempt to build this Lambda
  Function:
    Type: AWS::Serverless::Function
    Properties:
      PackageType: Image
      ImageUri: 123456789012.dkr.ecr.eu-west-1.amazonaws.com/function:latest
```

Instead, attempting to do so will most likely output the following exception (as the `DockerContext` resource metadata, required for building Docker Images using SAM, will most likely be unset, [error traced here](https://github.com/aws/aws-sam-cli/blob/54644dd72c7b227683566ef1d07def5ad343fbf3/samcli/lib/build/app_builder.py#L322)):

```
% sam build
Building image for Function function
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
[...]

PR #5729, opened 2023-08-08
fix: wrongful IMAGE function options validation (aws#2934)

alexisfacques:

Issue #2934

- `sam deploy` fails with error `Error: Missing option '--image-repositories', '--image-repository', '--resolve-image-repos'`, even when only valid ECRs `IMAGE` Lambdas are configured;
- Finally resolves issues from #2934 , introduced in PR #2935;

- Reuses the newly introduced `_is_all_image_funcs_provided` function, which validates that any buildable IMAGE Lambda has a valid ECR location provided; instead of only checking for any IMAGE lambdas in the template.
- Error prompt should only show if: no `--image-repositories` option is set and no ` '--image-repository'` or no `'--resolve-image-repos'` is set if the `_is_all_image_funcs_provided` function returns a falsy result
  
- Extensive refactoring to `samconfig` and `cli_validation` unit tests;
- Change in CLI options validation for many commands including `package`, `deploy`, `sync` (Both `ZIP` and `Image` lambdas);

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/cli_validation/image_repository_validation.py
+++ b/samcli/lib/cli_validation/image_repository_validation.py
@@ -5,7 +5,6 @@ This is to be run last after all CLI options have been processed.
 import click
 
 from samcli.commands._utils.option_validator import Validator
-from samcli.commands._utils.template import get_template_artifacts_format
 from samcli.lib.providers.provider import (
     ResourceIdentifier,
     get_resource_full_path_by_id,
@@ -49,16 +48,6 @@ def image_repository_validation(support_resolve_image_repos=True):
                 or ctx.params.get("template", False)
             )
 
-            # Check if `--image-repository`, `--image-repositories`, or `--resolve-image-repos` are required by
-            # looking for resources that have an IMAGE based packagetype.
-
-            required = any(
-                [
-                    _template_artifact == IMAGE
-                    for _template_artifact in get_template_artifacts_format(template_file=template_file)
-                ]
-            )
-
             available_options = "'--image-repositories', '--image-repository'"
             if support_resolve_image_repos:
                 available_options += ", '--resolve-image-repos'"
@@ -84,16 +73,6 @@ def image_repository_validation(support_resolve_image_repos=True):
                 ),
                 Validator(
                     validation_function=lambda: not guided
-                    and not (image_repository or image_repositories or resolve_image_repos)
-                    and required,
-                    exception=click.BadOptionUsage(
-                        option_name="--image-repositories",
-                        ctx=ctx,
-                        message=f"Missing option {available_options}",
-                    ),
-                ),
-                Validator(
-                    validation_function=lambda: not guided
                     and (
                         image_repositories
                         and not resolve_image_repos
@@ -101,6 +80,17 @@ def image_repository_validation(support_resolve_image_repos=True):
                     ),
                     exception=click.BadOptionUsage(
                         option_name="--image-repositories", ctx=ctx, message=image_repos_error_msg
+                    ),
+                ),
+                Validator(
+                    validation_function=lambda: not guided
+                    and not image_repositories
+                    and not _is_all_image_funcs_provided(template_file, {}, parameters_overrides)
+                    and not (image_repository or resolve_image_repos),
+                    exception=click.BadOptionUsage(
+                        option_name="--image-repositories",
+                        ctx=ctx,
+                        message=f"Missing option {available_options}",
                     ),
                 ),
             ]
--- a/tests/unit/lib/cli_validation/test_image_repository_validation.py
+++ b/tests/unit/lib/cli_validation/test_image_repository_validation.py
@@ -33,11 +33,10 @@ class TestImageRepositoryValidation(TestCase):
 
     @patch("samcli.lib.cli_validation.image_repository_validation.click")
     @patch("samcli.lib.cli_validation.image_repository_validation._is_all_image_funcs_provided")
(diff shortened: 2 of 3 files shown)


samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -85,7 +85,8 @@ def wrapped(*args, **kwargs):
                 Validator(
                     validation_function=lambda: not guided
                     and not (image_repository or image_repositories or resolve_image_repos)
-                    and required,
+                    and required
+                    and not _is_all_image_funcs_provided(template_file, image_repositories, parameters_overrides),
```
lucashuy, line 89: Could we add a quick unit test just to confirm that if all the conditions are met (not guided, no image repo options passed, not required, no images set in template) that we indeed get that exception?

lucashuy (commented):
Thanks for following up on this issue after all this time! Left a comment about testing

alexisfacques:
Hi there @lucashuy;
While implementing unit tests, turns out there was a little more to this story.

The `required` flag seems somewhat redundant now with the introduction of `_is_all_image_funcs_provided`. This is because if not all image functions are provided **then** options become `required`.
I've proposed a refactoring that not only eliminates the use of this flag, which in turns, removes a module dependency, leading to a more extensive set of changes.

I've made adjustments to the unit tests, specifically to the cli_validation, to encompass various scenarios:

- Given that all `Image` functions are provided (whether they consist of only `ZIP` lambdas functions, or `ECR` Image lambdas), no option should be required.
- If `--image-repositories` is set but not all image functions are provided, an error indicating incomplete repositories should always be prompted to favour to the missing options one.
- If **not** all image functions are provided &  `--image-repository` is set, no error should occur.
- If **not** all image functions are provided & `--image-repository` is unset, then only should a missing option error occur.

These changes could potentially impact the CLI validation behavior across multiple sam commands, notably including `package`, `deploy`, and `sync`; so lets make sure we don't forget anything if we move forward.

tests/unit/lib/cli_validation/test_image_repository_validation.py
```diff
@@ -90,12 +87,31 @@ def test_image_repository_validation_success_IMAGE_image_repositories(
...
+            "123456789012.dkr.ecr.us-east-1.amazonaws.com/test1",
+            False,
+            False,
+            None,
+            MagicMock(),
+        ]
+        mock_click.get_current_context.return_value = mock_context
+        self.foobar()
+
+    @patch("samcli.lib.cli_validation.image_repository_validation.click")
```
lucashuy, line 108: Duplicate test as the one on line 49 (`test_image_repository_validation_success_IMAGE_image_repository`) of this file now that the refactor happened?

lucashuy (commented):
Thanks for that catch! That is a good call out regarding the other commands that also use this decorator/validation. Since we are still checking if all the Lambdas that are image based, have ECR URLs, I don't imagine that it would be an issue to update with the new method call, however will keep an eye out on the tests.

tests/unit/lib/cli_validation/test_image_repository_validation.py
alexisfacques: Good call, the change was already included in that line 49 test; must have been a left over of my first PR version.
Edited and rebased.

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -103,6 +82,17 @@ def wrapped(*args, **kwargs):
                         option_name="--image-repositories", ctx=ctx, message=image_repos_error_msg
                     ),
                 ),
+                Validator(
+                    validation_function=lambda: not guided
+                    and not image_repositories
+                    and not _is_all_image_funcs_provided(template_file, {}, parameters_overrides)
+                    and not (image_repository or resolve_image_repos),
```
moelasmar, line 89: why not to check all parameters in one condition like and `not (image_repositories or image_repository or resolve_image_repos)` instead of having them in 2 conditions?

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -103,6 +82,17 @@ def wrapped(*args, **kwargs):
                         option_name="--image-repositories", ctx=ctx, message=image_repos_error_msg
                     ),
                 ),
+                Validator(
+                    validation_function=lambda: not guided
+                    and not image_repositories
+                    and not _is_all_image_funcs_provided(template_file, {}, parameters_overrides)
```
moelasmar, line 88: I think we still need to update `_is_all_image_funcs_provided` function logic. The current logic of this function is to check if all Image type functions defined in the template are exist in the `image_repositories` list. It does not exclude the functions that has an ECR url value.

alexisfacques: To my understanding, it actually does (?)

To acquire the list of Lambda functions from the template, it relies on [SamFunctionProvider](https://github.com/aws/aws-sam-cli/blob/98f8f3b738c5a98e159e04a899f0784a567daa43/samcli/lib/providers/sam_function_provider.py#L179) which ignores IMAGE Lambdas with ECR registries

moelasmar: my bad, I did not check the Function Provider.

moelasmar:
Thanks @alexisfacques for your contribution, please let us know if you need any help to update the code or add integration test cases.

mndeveci:
Closing this PR due to staleness. Please re-open it once you will be able to address the previous comments. Thanks!

PR #5734, opened 2023-08-09
feat: Terraform Custom Plan File Support

mildaniel:

N/A

Support users who use Terraform Cloud by allowing them to pass a custom plan file instead of generating one.

Creates a new command line option for passing a plan file and using that one in the Terraform hook instead of generating a plan file.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/_utils/options.py
```diff
@@ -789,6 +789,16 @@ def use_container_build_option(f):
     return use_container_build_click_option()(f)
 
 
+def plan_file_click_option():
+    return click.option(
+        "--plan-file", type=click.Path(), help="Used for passing a custom plan file when executing the Terraform hook."
```
moelasmar, line 794: do we need to make the parameter name more specific to terraform like `--terraform-plan-file` ?

samcli/commands/build/core/options.py
```diff
@@ -71,5 +74,6 @@
             ),
         ],
     },
+    "Hook Options": {"option_names": {opt: {"rank": idx} for idx, opt in enumerate(HOOK_OPTIONS)}},
```
moelasmar, line 77: make it `Terraform Hook Options`, so in future if we decided to add more hooks, or we support 3rd party hooks we can differentiate between each hook options.

moelasmar: apply this comment to other commands as well.

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -167,3 +141,56 @@ def _update_resources_paths(cfn_resources: Dict[str, Any], terraform_application
                 original_path = resource.get("Properties", {}).get(attribute)
                 if isinstance(original_path, str) and not os.path.isabs(original_path):
                     resource["Properties"][attribute] = str(Path(terraform_application_dir).joinpath(original_path))
+
+
+def _generate_plan_file(skip_prepare_infra: bool, terraform_application_dir: str) -> dict:
```
moelasmar, line 146: should we move the try/except block as well to this function, as all the errors should be coming from executing Terraform commands

mildaniel: I moved the `CalledProcessError` and `LoadingPatternError` exceptions to the function but left the `OSError` in the parent function since `open()` will throw this if it fails.

moelasmar: thanks Daniel :)

moelasmar (approved):

lucashuy (approved):

lucashuy:
Should we consider adding a check if hook name is provided, if it isn't then we raise a bad click option exception?

Issue #4724, opened 2023-02-16
Bug: sam build command doesn't support terraform's nested directory structure


shazi57:
### Description:
Relative paths provided in the source for modules throw errors when sam build command is run.

### Steps to reproduce:
```
  {root}/environments/dev/main.tf

  module "proxy_integrated_lambda" {
  source = "../../modules/lambda"
}
```
```
in {root}/modules/lambda

resource "aws_lambda_function" "terraform_lambda_func" {
  function_name                  = "${var.service_name}-${var.filename}-${var.env}"
  role                           = "${var.role}"
  handler                        = "${var.filename}.handler"
  s3_bucket                      = "${var.bucket_name}"
  s3_key                         = "${var.service_name}.zip"
  s3_object_version              = "${var.version_id}"
  publish                        = true
  runtime                        = "nodejs16.x"
  environment {
    variables = var.envvars
  }
}

resource "null_resource" "sam_metadata_aws_lambda_function_terraform_lambda_func" {
    triggers = {
        resource_name = "aws_lambda_function.terraform_lambda_func"
        resource_type = "ZIP_LAMBDA_FUNCTION"
        original_source_code = "./src"
        built_output_path = "./src/${var.service_name}.zip"
    }
}
```

- cli command
sam build --hook-name terraform --beta-features

### Observed result:
```
Running Prepare Hook to prepare the current application
Executing prepare hook of hook "terraform"
Initializing Terraform application
.....
Creating terraform plan and getting JSON output
.............
Generating metadata file
Finished generating metadata file. Storing in /Users/tryu2/terraform-practice/environments/dev/.aws-sam-iacs/iacs_metadata/template.json
Prepare hook completed and metadata file generated at: /Users/tryu2/terraform-practice/environments/dev/.aws-sam-iacs/iacs_metadata/template.json
Building codeuri: /Users/tryu2/terraform-practice/environments/dev/src runtime: nodejs16.x metadata: {'SkipBuild': False, 'BuildMethod': 'makefile', 'ContextPath': '/Users/tryu2/terraform-practice/environments/dev/.aws-sam-iacs/iacs_metadata', 'WorkingDirectory': '/Users/tryu2/terraform-practice/environments/dev', 'ProjectRootDirectory': '/Users/tryu2/terraform-practice/environments/dev'} architecture: x86_64 functions: module.proxy_integrated_lambda.aws_lambda_function.terraform_lambda_func
Running CustomMakeBuilder:CopySource
[...]

PR #5745, opened 2023-08-10
feat: support terraform nested directories project structure

moelasmar:
#4724

Currently, customers could not use `sam build` command on Terraform projects that has a nested directories structure.

Customers can pass a new parameter `--terraform-project-root-path` and pass the project root directory path that contains all the nested structures.
for example for this scenario:
```
/project
  /infrastructure
    envrionment/
        dev/
            dev.tfvars
        test/
            test.tfvars
    main.tf
    providers.tf
  /src
    ... (lambda code) ...
  /tests
    ... (lambda tests) ...
```
customers can execute `sam build` command from this path `/project/infrastructure` where the Terraform root module configuration files exist, and pass the value of the new parameter `--terraform-project-root-path` as `./..` or `/project`.

SAM CLI will use the value passed in the new parameter to set the `ProjectRootDirectory` property in all the lambda functions/layers metadata properties, and so Lambda Builders will be able to determine which directory that will be copied to the Temporary directory, and then run the building logic.

N/A

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
```diff
@@ -42,13 +42,20 @@ def prepare(params: dict) -> dict:
...
         raise PrepareHookException("OutputDirPath was not supplied")
 
-    LOG.debug("Normalize the project root directory path %s", terraform_application_dir)
+    LOG.debug("Normalize the terraform application root module directory path %s", terraform_application_dir)
     if not os.path.isabs(terraform_application_dir):
         terraform_application_dir = os.path.normpath(os.path.join(os.getcwd(), terraform_application_dir))
-        LOG.debug("The normalized project root directory path %s", terraform_application_dir)
+        LOG.debug("The normalized terraform application root module directory path %s", terraform_application_dir)
+
+    LOG.debug("Normalize the project root directory path %s", terraform_application_dir)
```
mildaniel, line 55: nit: Should this be `project_root_dir`?

mildaniel (approved):

samcli/hook_packages/terraform/hooks/prepare/hook.py
moelasmar: yes, I will update it

lucashuy (approved):

PaulMillen-Which:
Hi @toddtarsi This doesn't appear to be an option as of 1.108.0 (not listed in `-h`), although I can see it was released at the end of Aug.  Still getting `Error: No such option: --terraform-project-root-path` and it doesn't work as a `samconfig.yaml` setting either.

Issue #2590, opened 2021-02-05
HttpApi lambda error mismatch with sam local


iRoachie:
### Description:

Hey team, 

Found a discrepancy between `sam local start-api` and running in an AWS account. 

While running in an AWS account, unhandled errors thrown from AWS lambda connected with API gateway return a 500 internal server error.

While running locally via `sam local start-api` these same errors return a 200:

```json
{
    "errorType": "Error",
    "errorMessage": "",
    "trace": [
        "Error",
        "    at Runtime.exports.index [as handler] (/var/task/app.js:13:9)",
        "    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)"
    ]
}
```

### Steps to reproduce:

I've provided a repo https://github.com/iRoachie/http-api-example.

- Clone the repo 

##### For local:
- `sam local start-api`
- `curl -i localhost:3000`

```
HTTP/1.0 200 OK
Content-Type: application/json
Content-Length: 192
Server: Werkzeug/1.0.1 Python/3.8.7
Date: Fri, 05 Feb 2021 16:23:11 GMT

{"errorType": "Error", "errorMessage": "", "trace": ["Error", "    at Runtime.exports.index [as handler] (/var/task/app.js:13:9)", "    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)"]}%
```

##### For AWS
- `sam deploy --guided`
- See outputs `ApiURL`
- `curl -i {ApiURL}`

```
HTTP/2 500 
date: Fri, 05 Feb 2021 16:34:30 GMT
content-type: application/json
content-length: 35
apigw-requestid: aR7_khtxoAMEVlQ=

{"message":"Internal Server Error"}%    
```

### Expected result:

`sam local start-api` should have the same output as running on AWS

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: mac os 10.15.7
2. `sam --version`: 1.17.0
3. AWS region: us-east-1

PR #5749, opened 2023-08-10
Local Invoke Lambda Errors now display 500 code

jysheng123:
#2590 

Previously, there was a mismatch on http response codes on the same commands on AWS and local

Add's the exception testing for when the lambda function fails locally, so the output is the same as on AWS

Updates current lambda failure response code to 500 from 502 previously. May need to verify that this is correct instead of creating a new response code

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jysheng123:
I did make a change which I would like input on, the current lambda failure response is 502, which I changed to 500, which now matches the initial internal server error message as well now. Just want to make sure this was a bug and not intentional, as there are other functions that use this response code

tests/unit/local/apigw/test_service_error_responses.py
```diff
@@ -16,7 +16,7 @@ def test_lambda_failure_response(self, jsonify_patch, make_response_patch):
         self.assertEqual(response, {"Some Response"})
 
         jsonify_patch.assert_called_with({"message": "Internal server error"})
-        make_response_patch.assert_called_with({"json": "Response"}, 502)
+        make_response_patch.assert_called_with({"json": "Response"}, 500)
```
jfuss, line 19: I don't think this should change. Looking at this test, it seems like the Lambda function raises an error vs the issues comments about a logical issue in the code.

It would be helpful to write an integration test for the case detailed in the Github Issue that proves what you intend happens and does not have side affects to other areas of the codebase.

samcli/local/apigw/service_error_responses.py
```diff
@@ -51,7 +52,7 @@ def lambda_failure_response(*args):
         :return: A Flask Response
         """
         response_data = jsonify(ServiceErrorResponses._LAMBDA_FAILURE)
-        return make_response(response_data, ServiceErrorResponses.HTTP_STATUS_CODE_502)
+        return make_response(response_data, ServiceErrorResponses.HTTP_STATUS_CODE_500)
```
jfuss, line 55: How did we validate this? The issue states the current code returns 200, so why are we adjusting the 502 to a 500 error?

jfuss (changes requested):

samcli/local/apigw/service_error_responses.py
jysheng123: I changed this part because the lambda_failure message attached to this error is Internal Server Error, however when I look at http codes for 502 it's bad gateway which I interpreted as there was an issue here with the service code/message not coordinating properly. Let me know what you think

samcli/local/apigw/service_error_responses.py
```diff
@@ -51,7 +52,7 @@ def lambda_failure_response(*args):
         :return: A Flask Response
         """
         response_data = jsonify(ServiceErrorResponses._LAMBDA_FAILURE)
-        return make_response(response_data, ServiceErrorResponses.HTTP_STATUS_CODE_502)
```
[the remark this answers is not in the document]
jysheng123: No, but maybe we need to make a new one for it

samcli/local/apigw/service_error_responses.py
jysheng123: The change I made before updated it so lambda internal error doesn't return 200 it throws a internal server error, however I saw even after throwing it it returned a 502 error which is separate from what AWS throws which is 500, I saw this discrepancy in the error codes, which i labeled above.

tests/unit/local/apigw/test_service_error_responses.py
jysheng123: We have some integration tests already testing the functionality that local start api returns an error for invoking, as for the integration tests comparing local and aws, that is something we should look at in the future, because there is a difference between the error codes used there so once we fix that we can create an integ test for it.

jysheng123:
Note: this change will fix the issue so that in the scenario for local invoke, it will throw an error, however in the existing codebase for local, the error code throws a 502 error instead of the previous 200 and unless some structural changes are made, can't throw a 500 error or a lot of existing integration tests will fail. However the client sees a 500 error when run on aws, so some investigation should be made on why there are 2 different errors. That should be for a different scope, this ticket is just for making an error be thrown when it should be for lambda parsing errors.

tests/integration/testdata/start_api/main.py
```diff
@@ -73,6 +73,8 @@ def invalid_v2_respose_returned(event, context):
 def invalid_hash_response(event, context):
     return {"foo": "bar"}
 
+def invalid_body_response(event, context):
+    return {"errorType": "Error", "errorMessage": ""}
```
jfuss, line 77: Why mock this vs making the exception happen by throwing an exception. From the issue: that is what the customer is doing? https://github.com/iRoachie/http-api-example/blob/master/hello-world/app.js

Nothing strictly wrong with this but could become out of date. Will let you decide if this is enough of a reason to update.

jfuss (approved):

tests/integration/testdata/start_api/main.py
jysheng123: I think it is fine to leave the testing like this, this is mocking the what the client is calling on python instead of javascript, because this is the formatting for error on lambda https://docs.aws.amazon.com/lambda/latest/dg/nodejs-exceptions.html#nodejs-exceptions-syntax. So unless aws lambda changes the error formatting, It shouldn't become out of date.

lucashuy (approved):

PR #5753, opened 2023-08-11
Fix: Remove unneeded directory "schema"

forzagreen:

It fixes a bug introduced in [v1.95.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.95.0):
When installing `aws-sam-cli`, it creates 2 folders in `site-packages`: `samcli` and `schema`.

The folder `schema` should not be published. It pollutes the environment, and enters in conflict with third-party library [schema](https://pypi.org/project/schema/)

Remove the directory `schema` before publishing the package.

#### Mandatory Checklist

- [x] `make pr` passes


forzagreen:
I was not sure whether to remove the directory here:
https://github.com/aws/aws-sam-cli/blob/6fd338993aefdcb0261698e2b88f58a37ee1bf2e/installer/pyinstaller/build-linux.sh#L51-L56
or to exclude it here:
https://github.com/aws/aws-sam-cli/blob/6fd338993aefdcb0261698e2b88f58a37ee1bf2e/setup.py#L54

lucashuy:
Thanks for raising this PR! We can remove this folder from our packages, however they shouldn't affect the PKG installers since those are isolated from the rest of the system, and thus from any other Python packages. We can exclude this folder anyways since the code that is in there is not used by the rest of the tool.

We can remove the folder in the `setup.py` file like you've noted so that it doesn't get included in the package that is uploaded to pypi.

forzagreen:
@lucashuy  thank you for your response. Feel free to review/update this PR, or to close it and open a fix in a separate PR.

lucashuy:
We should be able to reuse this existing PR. We can add that `schema` folder to the `setup.py`'s exclude, and we can also `rm` that folder for the `build-mac.sh` script too. That way the behaviour is the same across our repo's installation scripts.

forzagreen:
@lucashuy , done. 
`build-mac.sh` was missing `rm` of other folders, so I added the missing deletions to have the same behaviour as `build-linux.sh`

forzagreen:
Reverted adding `schema` to the `setup.py`'s exclude, because it breaks `make schema`

hawflau (changes requested):
I believe the changes in setup.py are needed. `build-linux.sh` and `build-mac.sh` are used to create our installers. If you install AWS SAM CLI using our installers, I don't think you will hit the issue you saw. My guess is that you used `pip install`. But then if setup.py is not updated, using `pip install` will still hit the same issue.

I think the easiest way is to change the "schema" directory name to another name that's more unlikely to conflict with existing packages on PyPI.

forzagreen:
@hawflau , thank you for reviewing.
Indeed we got the issue when using `pip install`.

Do we really want to deploy the `schema` folder ? Isn't it used only for development purposes ?
If `schema` must be deployed ? why not just putting it inside `samcli` ? otherwise, I would suggest renaming it `awscli-schema`

lucashuy:
After some offline discussion, we can still exclude the folder and it's contents like was previously done in `setup.py`. To fix the issue of the file not being found when being installed in dev mode (like being done in the PR jobs), we can just change the import statements for the schema generation file itself.

This involves changing the follow line: https://github.com/aws/aws-sam-cli/blob/8bfda3287c617d3fadd0bf670d75bb16e8d07dd6/schema/make_schema.py#L14 to just an `import exceptions` and updating the two references to `SchemaGenerationException` in the same file, to `exceptions.SchemaGenerationException`. This should solve the issue of a missing parent module when excluding this folder from the setup file.

forzagreen:
`import exceptions` brings other issues in typing and unit tests. I simply modified `PYTHONPATH` in `make schema` @lucashuy @hawflau

forzagreen:
Fixed an issue with Windows build in eeb968b750bad74b6e8e29b1fb98e91cc994ce07 . I run `make_schema.py` with `-m` instead of setting `PYTHONPATH`.

CI tested in https://github.com/forzagreen/aws-sam-cli/actions/runs/5936100836

lucashuy:
> Fixed an issue with Windows build in [eeb968b](https://github.com/aws/aws-sam-cli/commit/eeb968b750bad74b6e8e29b1fb98e91cc994ce07) . I run `make_schema.py` with `-m` instead of setting `PYTHONPATH`.
> 
> CI tested in https://github.com/forzagreen/aws-sam-cli/actions/runs/5936100836

Thanks for checking this out, this looks to be a cleaner fix!

lucashuy (approved):

forzagreen:
@lucashuy @hawflau Thank you for the review. Feel free to merge.

hawflau (approved):
Thanks for your contribution!

Issue #5361, opened 2023-06-13
Bug: DockerDistributionAPIError suppresses error information


Patman12:
### Description:
When `sam local invoke --debug` fails due to an unknown Docker APIError, the only output you see is `Error: Unknown API error received from docker`. Since this error handling was [added in v1.83.0](https://github.com/aws/aws-sam-cli/commit/6d02c36565189247284d0e75fa8a605746a1781f), I reverted to v1.82.0 to investigate and noticed the error was instead `Error: 500 Server Error: Internal Server Error ("Get "https://public.ecr.aws/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)")`.

### Steps to reproduce:
Not sure how to consistently reproduce the issue, but any non-404 APIError from docker will cause `sam local invoke` to fail with this error

### Observed result:
Only output relevant to the failure is `Error: Unknown API error received from docker` even with `--debug` set.

### Expected result:
The original APIError returned from Docker should be shown in the output

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS Monterey
2. `sam --version`: 1.86.1
3. AWS region: N/A

```{
  "version": "1.82.0",
  "system": {
    "python": "3.11.4",
    "os": "macOS-12.6.6-x86_64-i386-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.2",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  }
}```

PR #5757, opened 2023-08-11
Display full error in DockerDistributionAPI error

jysheng123:
#5361 

Adds useful information for developer encountering this error

Adds the full error when error is raised

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/docker/lambda_image.py
```diff
@@ -211,7 +211,7 @@ def build(self, runtime, packagetype, image, layers, architecture, stream=None,
                 )
                 image_not_found = True
             else:
-                raise DockerDistributionAPIError("Unknown API error received from docker") from e
+                raise DockerDistributionAPIError("Error: " + str(e)) from e
```
hnnasit, line 214: I think `str(e)` should be enough here. `Error: ` gets added when the exception is thrown by default.

hnnasit (approved):

moelasmar (approved):

Issue #5777, opened 2023-08-16
Bug: Local lambda authorizer: wildcards in resource ARNs do not match an empty string


ethanmills:
### Description:

When running with `sam local start-api`, a lambda authorizer returning a policy for `arn:aws:execute-api:us-east-1:123456789012:1234567890/Prod/*/*` does not permit requests to the root URL, which has a methodArn of `arn:aws:execute-api:us-east-1:123456789012:1234567890/Prod/GET/`. This does not reflect the deployed behaviour.

### Steps to reproduce:

- Create an application with a lambda authorizer which returns a policy with statement 
```json
{
    "Action": "execute-api:Invoke",
    "Effect": "Allow",
    "Resource": ["arn:aws:execute-api:us-east-1:123456789012:1234567890/Prod/*/*"],
}
```
- Call a lambda mounted at `/`

### Observed result:

`{"message":"User is not authorized to access this resource"}`

### Expected result:

Request should be permitted, as in the deployed API gateway

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```json
{
  "version": "1.95.0",
  "system": {
    "python": "3.8.13",
    "os": "macOS-13.5-arm64-arm-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.2",
    "aws_cdk": "Not available",
    "terraform": "1.5.5"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}
```

lucashuy:
Thanks for raising this issue and the PR to fix this! I can reproduce this issue with a template that has a method wildcard, and path wildcard, and can see that with a root resource ("`/`"), it matches correctly when deployed vs when using `local start-api`.

PR #5778, opened 2023-08-16
fix: Local lambda authorizer ARN matching

ethanmills:
Update the matching logic to permit a wildcard to match an empty string. This allows a wildcard resource statement to match the root URL, which mimics the real behaviour of API Gateway.

#5777 

Details on the issue

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/local/apigw/authorizers/lambda_authorizer.py
+++ b/samcli/local/apigw/authorizers/lambda_authorizer.py
@@ -355,7 +355,7 @@ class LambdaAuthorizer(Authorizer):
 
     def _is_resource_authorized(self, response: dict, method_arn: str) -> bool:
         """
-        Validate the if the current method ARN is actually authorized
+        Validate if the current method ARN is actually authorized
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


lucashuy:
Hi, thanks for opening this PR to fix the matching! The unit test that you've added looks good, and should cover what we are looking for in a unit test.

As for any additional tests, it might be worth it to also add an integration test case that tests this behaviour. To do so, we can add a path for `"/"` into one of our testing template files, similar to this: https://github.com/aws/aws-sam-cli/blob/4395ecba44278e44723c7e29fb777fb813f0c3e0/tests/integration/testdata/start_api/lambda_authorizers/swagger-api.yaml#L49-L56

Then to get the test suite to actually run this case, we can add an entry that is pretty much the same as https://github.com/aws/aws-sam-cli/blob/4395ecba44278e44723c7e29fb777fb813f0c3e0/tests/integration/local/start_api/lambda_authorizers/test_swagger_authorizer_definitions.py#L19, but with an empty path (representing the root path).

Feel free to drop a comment if anything needs some clarification, or if something doesn't end up working out.

ethanmills:
@lucashuy Thanks for that. I've added an integration test as directed which fails before the change and passes afterwards. Ready for re-review 🙂

lucashuy (approved):
Thanks for addressing the changes and adding tests, they look good to me

moelasmar (approved):

Issue #3609, opened 2022-01-31
Setting --docker-network option on sam build command has no effect


bcornies:
### Description:
When running `sam build --docker-network host` I am expecting the Docker container that is under build to use the network specified by the `docker-network` flag.

PackageType: Image

### Steps to reproduce:
1. Create a simple SAM project w/ PackageType=Image
2. Add a `RUN SLEEP 100` to the Dockerfile to force the build process to pause so it can be inspected
3. Run `sam build --docker-network host`
4. While the Docker image is building, run `docker inspect --format='{{range $k, $v := .NetworkSettings.Networks}}{{$k}}{{end}}' [container]`

### Observed result:
Regardless of what is passed to the `--docker-network` flag, the under-build Docker container always uses the `bridge` network.

### Expected result:
I would expect the under-build Docker container to use the network specified by the `--docker-network` option.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

SAM CLI version: 1.36.0
OS: CentOS 7
AWS region: us-east-2

```
$ sam build --docker-network host --debug
2022-01-31 14:01:52,006 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2022-01-31 14:01:52,006 | Using config file: samconfig.toml, config environment: default
2022-01-31 14:01:52,006 | Expand command line arguments to:
2022-01-31 14:01:52,006 | --template_file=/home/bryan.cornies/repos/git/bids/teamcity-agent-authorizer/template.yaml --docker_network=host --build_dir=.aws-sam/build --cache_dir=.aws-sam/cache 
2022-01-31 14:01:52,173 | 'build' command is called
2022-01-31 14:01:52,180 | No Parameters detected in the template
2022-01-31 14:01:52,204 | 2 stacks found in the template
2022-01-31 14:01:52,204 | No Parameters detected in the template
2022-01-31 14:01:52,251 | 2 resources found in the stack 
2022-01-31 14:01:52,251 | No Parameters detected in the template
2022-01-31 14:01:52,286 | Found Serverless function with name='TeamCityAgentAuthorizer' and ImageUri='None'
2022-01-31 14:01:52,286 | --base-dir is not presented, adjusting uri ./TeamCityAgentAuthorizer relative to /home/bryan.cornies/repos/git/bids/teamcity-agent-authorizer/template.yaml
[...]

PR #5785, opened 2023-08-17
fix: Pass network mode to image build method

lucashuy:

#3609

Previously, `--network-mode` was only being passed to functions and layers that are Zip based when container builds were enabled. When an Image based function was being built, the network mode was not being passed down to the Docker API call.

Checks if there is a container manager and if the Docker network was set. If it was, include it in the build arguments when building images. 

None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -412,8 +412,12 @@ class ApplicationBuilder:
             "platform": get_docker_platform(architecture),
             "rm": True,
         }
+
         if docker_build_target:
             build_args["target"] = cast(str, docker_build_target)
+
+        if self._container_manager and self._container_manager.docker_network_id:
+            build_args["network_mode"] = cast(str, self._container_manager.docker_network_id)
 
         try:
             (build_image, build_logs) = self._docker_client.images.build(**build_args)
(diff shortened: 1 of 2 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -412,9 +412,13 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
             "platform": get_docker_platform(architecture),
             "rm": True,
         }
+
         if docker_build_target:
             build_args["target"] = cast(str, docker_build_target)
 
+        if self._container_manager and self._container_manager.docker_network_id:
+            build_args["network_mode"] = cast(str, self._container_manager.docker_network_id)
```
moelasmar, line 420: I think the provided `docker-network` parameter is related to the build container, and not the docker image we create for the lambda function. I believe the problem is the customer is using `host` as docker network, and for some reason that I do not remember, we do not update the created docker network if the provided value is host. check this code snippet:
https://github.com/aws/aws-sam-cli/blob/8bfda3287c617d3fadd0bf670d75bb16e8d07dd6/samcli/local/docker/container.py#L217-L224

lucashuy: During offline discussion, it was discussed that building images is likely fine with this change, however building resources under the `-u/--use-container` flag probably has the same issue where the `--network-mode host` does not do anything since the linked logic also binds ports, which is mutually exclusive.

lucashuy:
Closing due to the need to investigate the impact of `--network-mode` since passing `host` is only available on Linux systems. This is also conflicting with the other container configuration SAM CLI does (since the container init logic is the same as `local`).

PR #5787, opened 2023-08-17
chore: Update description for container env var option to mention debug mode

lucashuy:

None.

Currently, the option to pass in environment variables to local related containers are scoped to only happen when a debugging option (eg. `--debug-port`) is passed in. The description does not mention this, potentially causing confusion as to why the option is not working.

Mentions that the option sends environment variables only in a debugging session.

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/cli_common/options.py
```diff
@@ -166,7 +166,8 @@ def invoke_common_options(f):
             click.option(
                 "--container-env-vars",
                 type=click.Path(exists=True),
-                help="JSON file containing environment variables to be set within the container environment",
+                help="JSON file containing environment variables to be set within the container "
+                "environment when used in a debugging session locally.",
```
moelasmar, line 170: do we need to mention that these are extra environment variables and not a replacement for `--env-vars` parameter?

lucashuy: Maybe something like this can help make this a little more clear:

`JSON file containing additional environment variables to be set within the container when used in a debugging session locally.`

Adding the word "additional" and removed "environment"

moelasmar (approved):

mildaniel (approved):

PR #5808, opened 2023-08-21
feat: Add `--save-params` option

Leo10Gama:
With the addition of this parameter, it will be possible for users to save parameters to the config file directly from the terminal as they are passed in. This will simplify the process of setting up configurations, since the file doesn't have to be open.

The `--save-params` option has been added to most of the existing commands. With it, when run, any command passed in via the terminal will be saved to the provided config file.

Most commands now have an additional parameter, and an additional decorator on their `cli()` methods.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/cli/cli_config_file.py
```diff
@@ -230,6 +230,81 @@ def get_ctx_defaults(
...
+
+    for param_name, param_source in ctx._parameter_source.items():
+        if param_name in params_to_exclude:
+            continue  # don't save certain params
+        if param_source != ParameterSource.COMMANDLINE:
+            continue  # only parse params retrieved through CLI
+
+        # if param was passed via command line, save to file
+        param_value = ctx.params.get(param_name, None)
+        if param_value is None:
```
lucashuy, line 268: nit: Could we add a debug line here that a parameter was skipped?

samcli/cli/cli_config_file.py
```diff
@@ -230,6 +230,81 @@ def get_ctx_defaults(
...
+            continue  # no value given, ignore
+        config_file.put(cmd_names, "parameters", param_name, param_value, config_env_name)
```
lucashuy, line 270: Conversely, can we add a log line somewhere noting which parameters were saved? We can probably just have it as a single line, instead of logging each parameter we saved in a new line in the terminal.

tests/unit/cli/test_cli_config_file.py
```diff
@@ -236,3 +243,58 @@ def test_get_ctx_defaults_nested(self):
         get_ctx_defaults("intent-answer", provider, mock_context4, "default")
 
         provider.assert_called_with(None, "default", ["local", "generate-event", "alexa-skills-kit", "intent-answer"])
+
+    def test_save_command_line_args(self):
```
lucashuy, line 247: Can we split this test up into potentially 4 tests:
- Test missing `save-param` option
- Test that things get saved correctly
- Test non-CLI options aren't saved
- Test that the excluded parameters (config and save params) aren't saved.

That way should anything change, it'll be easier to update the correct test(s)

Leo10Gama: We could, though I feel like keeping everything in one test means we can check for multiple things without duplicating as much code, since things like setup and checking will look similar for each run.
Though I would be in favour of splitting off the case of `save_params = False` as its own test, since that's different behaviour entirely.

samcli/cli/cli_config_file.py
```diff
@@ -266,8 +267,15 @@ def save_command_line_args_to_config(
         # if param was passed via command line, save to file
         param_value = ctx.params.get(param_name, None)
         if param_value is None:
+            LOG.debug(f"Parameter {param_name} was not saved, as its value is None.")
             continue  # no value given, ignore
         config_file.put(cmd_names, "parameters", param_name, param_value, config_env_name)
+        saved_params.update({param_name: param_value})
+
+    LOG.debug(
```
lucashuy, line 275: Should this be an info log instead of a debug? Mainly asking since this is something that the customer is actively doing (providing `--save-params`), so we might want to report that something happened from the SAM CLI side of things that doesn't require a `--debug` to find out.

tests/unit/cli/test_cli_config_file.py
```diff
@@ -299,6 +303,63 @@ def test_save_command_line_args(self):
...
+        self.assertNotIn(
+            "param_not_from_commandline", params.keys(), "Param not passed in via COMMANDLINE should not be saved"
+        )
+
+    def test_dont_save_command_line_args_if_value_is_none(self):
+        mock_samconfig, mock_config_file = self._setup_mock_samconfig()
+        self.ctx = self._setup_context(
+            params={
+                "save_params": True,
+                "some_param": "value",
```
lucashuy, line 343: nit: We could remove this since this case is already being tested and isn't checked here

lucashuy (commented):
Thanks for making the changes! Just two minor comments, but the other changes look good to me

tests/unit/cli/test_cli_config_file.py
Leo10Gama: The case is there to make sure that the parameter itself actually gets written. If the only parameters being saved are `save_params` and `none_param` which don't get written to file, then the check for `default.command.parameters` fails, so the behaviour is indistinguishable from if the decorator just straight up doesn't run.

samcli/cli/cli_config_file.py
Leo10Gama: Yup, giving customers that feedback would definitely help, will make that change 👍

lucashuy (approved):

samcli/cli/cli_config_file.py
```diff
@@ -230,6 +230,89 @@ def get_ctx_defaults(
...
+    """Composite decorator to add --save-params flag to a command.
+
+    When used, this command should be placed as the LAST of the click option/argument decorators to preserve the flow
+    of execution. The decorator itself will add the --save-params option, and, if provided, save the provided commands
+    from the terminal to the config file.
+    """
+    return click.option(
+        "--save-params",
+        is_flag=True,
+        help="Save the terminal-provided parameters to the configuration file.",
```
hnnasit, line 312: nit/suggestion: This could also be changed to CLI-provided or CommandLine provided parameters

samcli/cli/cli_config_file.py
```diff
@@ -230,6 +230,89 @@ def get_ctx_defaults(
...
+
+        # if param was passed via command line, save to file
+        param_value = ctx.params.get(param_name, None)
+        if param_value is None:
+            LOG.debug(f"Parameter {param_name} was not saved, as its value is None.")
+            continue  # no value given, ignore
+        config_file.put(cmd_names, "parameters", param_name, param_value, config_env_name)
+        saved_params.update({param_name: param_value})
+
+    LOG.info(
```
hnnasit, line 275: Should this log be after the config file flushes and saves the parameters?

hnnasit (approved):
Left small comments but overall looks good to me!

samcli/cli/cli_config_file.py
Leo10Gama: There's likely a way for that to work, but keeping it as its own decorator ensures that it'll always work. Since it has to be the last parameter decorator in the function, grouping it with another function that returns composite decorators like this could cause issues, so keeping it separate prevents that from potentially happening.

samcli/cli/cli_config_file.py
Leo10Gama: Good point, would probably be best to have the message after the file is written 👍  Will update that

PR #5809, opened 2023-08-21
feat: Added environment variable validation for Terraform arguments

lucashuy:

None.

Validates that the environment variables that are passed in when running commands do not override anything we define or set.

Checks if the environment variables contain blocked arguments. If they do, exit and display a message letting the user know to omit them.

This check happens in both the Makefile script and the prepare hook logic.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
+++ b/samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
@@ -56,6 +56,16 @@ from zip import unzip  # type: ignore
 LOG = logging.getLogger(__name__)
 
 TF_BACKEND_OVERRIDE_FILENAME = "z_samcli_backend_override"
+TF_BLOCKED_ARGUMENTS = [
+    "-target",
+    "-destroy",
+]
+TF_ENVIRONMENT_VARIABLE_DELIM = "="
+TF_ENVIRONMENT_VARIABLES = [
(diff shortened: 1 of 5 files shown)


samcli/hook_packages/terraform/copy_terraform_built_artifacts.py
```diff
@@ -276,6 +286,31 @@ def create_backend_override():
         cli_exit()
 
 
+def validate_environment_variables():
```
mildaniel, line 289: What is the scenario where these environment variables aren't set when checking at the hook level but are when running this script?

lucashuy: Normally, this shouldn't happen when running `sam build` since the customer shouldn't have to manually run `sam build -t <aws sam iacs metadata folder>/template.json`, but the customer is still able to run this manually. So having it in two locations solves:
1. `sam build --hook-name` 
    - hook check covers this
3. `sam build --hook-name`, then for some reason `sam build -t <iacs metadata template>`
   - hook check covers first build run
   - script covers the second run

mildaniel (approved):

moelasmar (approved):

PR #5842, opened 2023-08-24
feat: Add integration tests to `--save-params`

Leo10Gama:
Due to the previous changes adding the option `--save-params`, we must ensure behaviour is consistent by using an integration test.

An integration test is added to verify the behaviour of the `--save-params` option, ensuring it works as intended.

One additional `build` integration test is added.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/buildcmd/test_build_samconfig.py
```diff
@@ -81,6 +82,44 @@ def test_samconfig_parameters_are_overridden(self, extension):
...
+            self.assertNotEqual(params, {}, "samconfig.toml was not parsed correctly")
+            self.assertIn("parallel", params.keys(), "New key-value pair should be written to config file")
+            self.assertTrue(
+                params.get("build_dir", None) == "new_dir",
+                "New value for existing key build_dir should overwrite old value",
+            )
+        finally:
+            # remove temp dir
+            config_path = Path(self.working_dir, "samconfig.toml")
+            os.remove(config_path)
```
lucashuy, line 121: This should already happen in the `tearDown` method in the parent class (`BuildIntegBase`), did we need to do this here or could we just rely on the tear down in the parent?

tests/integration/buildcmd/test_build_samconfig.py
```diff
@@ -81,6 +82,44 @@ def test_samconfig_parameters_are_overridden(self, extension):
         self.assertIn(overrides["Runtime"], stderr, "parameter_overrides should use overridden runtime")
         self.assertNotIn("SomeURI", stderr, f"parameter_overrides should not read ANY values from samconfig{extension}")
 
+    def test_save_params_saves_params(self):
+        # setup temp dir
+        new_template_location = Path(self.working_dir, "template.yaml")
+        new_template_location.write_text(Path(self.template_path).read_text())
+        config_contents = Path(self.test_data_path, configs[".toml"]).read_text()
+        new_path = Path(self.working_dir, "samconfig.toml")
```
lucashuy, line 90: nit: `new_path` -> `new_config_path`

tests/integration/buildcmd/test_build_samconfig.py
Leo10Gama: Good call. Just tested, and it seems we can rely on the parent behaviour. Will remove the unnecessary logic 👍

lucashuy (approved):

hnnasit (approved):

PR #5852, opened 2023-08-28
Chore: Use the latest stack event timestamp as a marker to get the new deployment events

moelasmar:
add some logs to deploy command to investigate issue # https://github.com/aws/aws-sam-cli/issues/5109

The problem in that issue is the WSL environment clock is not in sync with the actual time, it was 3 minutes ahead, and so the loop we use to describe the deployment events never ends, as it is not able to find any new events.

I updated the wait_for_execute method that wait for a change set to be executed, to take a time_stamp_marker as an extra parameter to use it to determine the events to be printed. The value of this marker will be the latest stack event if the stack exists, or 0 in case if the stack does not exist. 

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/deploy/deployer.py
+++ b/samcli/lib/deploy/deployer.py
@@ -647,7 +647,6 @@ class Deployer:
                 kwargs["DisableRollback"] = disable_rollback  # type: ignore
                 # get the latest stack event, and use 0 in case if the stack does not exist
                 marker_time = self.get_last_event_time(stack_name, 0)
-                print(f">>>>>>> {marker_time}")
                 result = self.update_stack(**kwargs)
                 self.wait_for_execute(
                     stack_name, "UPDATE", disable_rollback, on_failure=on_failure, time_stamp_marker=marker_time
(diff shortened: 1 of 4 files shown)


samcli/lib/deploy/deployer.py
```diff
@@ -384,6 +385,7 @@ def describe_stack_events(
         while stack_change_in_progress and retry_attempts <= self.max_attempts:
             try:
                 # Only sleep if there have been no retry_attempts
+                LOG.debug("Trail # %d to get the stack %s create events", retry_attempts, stack_name)
```
sriram-mv, line 388: nit: typo

sriram-mv (commented):
Overall looks good.

samcli/lib/deploy/deployer.py
moelasmar: fixed

lucashuy (approved):

mndeveci (approved):

PR #5870, opened 2023-08-30
feat: Resolve symlinks at root level and mount to container

lucashuy:

None.

Previously, any symlinked top level files in the source code folder for Lambda functions were not being mounted/linked correctly when running in a container.

Goes through the top level files and directories and resolves the symlinks if there are any. Then we mount the resolved files/directories to that location in the container.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/testdata/invoke/symlinked/src/app.py
```diff
@@ -0,0 +1,4 @@
+def handler(e, c):
+    with open("./linked-file.txt", "r") as file:
```
mildaniel, line 2: This behaviour seems a little odd to me. This file is a directory up on the host machine, but in the container it's mounted in the same dir. I feel like this might make for confusing UX, or maybe I'm missing the use-case here/how it relates to build-in-source.

tests/integration/testdata/invoke/symlinked/source-file.txt
```diff
@@ -0,0 +1 @@
+cats
```
mildaniel, line 1: I prefer dogs 🐶

tests/integration/testdata/invoke/symlinked/src/app.py
lucashuy: The linked file here is actually one that lives up a folder level, testing if the link resolution behaviour is able to mount the file correctly
```
$ tree .
.
├── source-file.txt
├── src
│   ├── app.py
│   └── linked-file.txt -> ../source-file.txt
└── template.yaml
```
I can update this test to have a real world example of building + invoking a symlinked dependency folder since this that is more relevant to build in source.

mildaniel (approved):

samcli/local/docker/container.py
```diff
@@ -225,6 +226,45 @@ def create(self):
...
+                if not file.is_symlink():
+                    continue
+
+                host_resolved_path = os.path.realpath(file.path)
+                container_full_path = pathlib.Path(self._working_dir, file.name).as_posix()
+
+                additional_volumes[host_resolved_path] = {
+                    "bind": container_full_path,
+                    "mode": mount_mode,
+                }
```
mndeveci, line 259: As we discussed offline, we need to check with security about this part before releasing it.

mndeveci (approved):

Issue #5901, opened 2023-09-07
Bug: Using YAML anchors for Resource Metadata makes resource disappear due to template mutation


davidjb:
### Description:

I have a large number of Go Lambda functions defined in my `template.yaml` and I was looking to simply my SAM template and make it more DRY by using anchors on their `Metadata` property like so:

```yaml
Resources:
  FirstFunction:
    Type: AWS::Serverless::Function
    Metadata: &GoFunctionMetadata
      BuildMethod: go1.x
      BuildProperties:
        TrimGoPath: true
    [...]
  SecondFunction:
    Metadata: *GoFunctionMetadata
  [...]
```

However, when doing this, the given function becomes no longer available to SAM (see below).

The reasoning because SAM adds the `SamResourceId` key within a resource's metadata (https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/samlib/resource_metadata_normalizer.py#L95-L99) and this gets used to create a list of available resources to build (https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/build/build_context.py#L565-L568). The problem occurs because the underlying `OrderedDict` structure is shared between functions, then the shared object gets mutated by the the normalizer when setting its properties.

This same issue can/will occur anywhere else the template structure where the template is being mutated so prior making changes, the original data structure should be copied.

### Steps to reproduce:

1. Create `template.yaml` like so:

```yaml
AWSTemplateFormatVersion: "2010-09-09"
Transform: AWS::Serverless-2016-10-31

Resources:
  FirstFunction:
    Type: AWS::Serverless::Function
    Metadata: &GoFunctionMetadata
      BuildMethod: go1.x
      BuildProperties:
        TrimGoPath: true
    Properties:
      CodeUri: dummy
      Handler: bootstrap
      Runtime: provided.al2
  SecondFunction:
    Type: AWS::Serverless::Function
    Metadata: *GoFunctionMetadata
    Properties:
      CodeUri: dummy
      Handler: bootstrap
      Runtime: provided.al2
  ThirdFunction:
    Type: AWS::Serverless::Function
    Metadata: *GoFunctionMetadata
    Properties:
      CodeUri: dummy
      Handler: bootstrap
      Runtime: provided.al2
```

2. Run `sam build SecondFunction --debug` 

### Observed result:

```sh
$ sam build SecondFunction --debug
2023-09-07 17:39:26,871 | No config file found in this directory.
[...]

PR #5904, opened 2023-09-07
fix: copy resource metadata before normalizing the information in it

mndeveci:

#5901

When using value references in yaml file, the objects that is been created are the same instance. For that reason, changing one value from one instance will also affect the other instances.

The behaviour above causes issues when updating/normalizing resource metadata information. With that, we are adding `SamResourceId` information, but if `Metadata` information is been shared between resources, then it updates for all of them, which creates issues down the road when we are extracting functions.

This change copies metadata before starting changes, and assigns it back once it is finalized, so that it won't affect all of the references.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/lib/samlib/test_resource_metadata_normalizer.py
+++ b/tests/unit/lib/samlib/test_resource_metadata_normalizer.py
@@ -457,7 +457,7 @@ class TestResourceMetadataNormalizer(TestCase):
         self.assertEqual("new path", template_data["Resources"]["Function1"]["Properties"]["Code"])
         self.assertEqual("Function1", template_data["Resources"]["Function1"]["Metadata"]["SamResourceId"])
 
-    def test_referenced_data(self):
+    def test_with_referenced_metadata(self):
         input_template = """
         Resources:
           FirstFunction:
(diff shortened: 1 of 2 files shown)


tests/unit/lib/samlib/test_resource_metadata_normalizer.py
```diff
@@ -456,6 +457,53 @@ def test_skip_normalizing_already_normalized_resource(self):
         self.assertEqual("new path", template_data["Resources"]["Function1"]["Properties"]["Code"])
         self.assertEqual("Function1", template_data["Resources"]["Function1"]["Metadata"]["SamResourceId"])
 
+    def test_referenced_data(self):
```
sriram-mv, line 460: Where are we asserting based on the new changes?

sriram-mv (changes requested):

tests/unit/lib/samlib/test_resource_metadata_normalizer.py
mndeveci: Before this change, `SamResourceId` is assigned as `FirstFunction` for all resources due to the reason I explained in the PR desc.

After this change, all resources must have their logical id as `SamResourceId`, we are asserting them at the end of this changed block.

There is a redundant function that I forgot to remove, please ignore that, will be removing it now.

sriram-mv (approved):

lucashuy (approved):

PR #5909, opened 2023-09-08
fix the hanging problem in windows if the terraform plan command failed

moelasmar:
In windows, if Terraform plan command failed, the std out stream will not be closed till we read all the logs in the std err stream. The solution is to redirect the Terraform plan subprocess stderr stream to write in stdout, so we can read all the plan command logs from stdout, and we do not need to switch to the stderr in case if the command failed. This change only applied for Windows. I also added a check to differentiate between the plan command normal output, and errors.

I added a new integration test case to make sure that SAM CLI will not hang in case if terraform plan command failed.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/subprocess_utils.py
```diff
@@ -3,15 +3,19 @@
...
 from concurrent.futures.thread import ThreadPoolExecutor
-from subprocess import PIPE, Popen
+from subprocess import PIPE, STDOUT, Popen
 from time import sleep
 from typing import Any, AnyStr, Callable, Dict, Optional, Union
 
 from samcli.commands.exceptions import UserException
 from samcli.lib.utils.stream_writer import StreamWriter
 
+TERRAFORM_ERROR_PREFIX = [27, 91, 51, 49]
```
mildaniel, line 16: Are these documented somewhere?

samcli/lib/utils/subprocess_utils.py
```diff
@@ -92,10 +96,18 @@ def _print_loading_pattern():
                     # we read from subprocess stdout to avoid the deadlock process.wait function
                     # for more detail check this python bug https://bugs.python.org/issue1256
                     for line in process.stdout:
-                        decoded_line = _check_and_process_bytes(line)
+                        is_error = (
+                            not IS_NOT_WINDOWS
+                            and len(line) >= len(TERRAFORM_ERROR_PREFIX)
+                            and [c for c in line[0:4]] == TERRAFORM_ERROR_PREFIX
```
mildaniel, line 102: Is this the same as 
```suggestion
                            and line[0:4] == TERRAFORM_ERROR_PREFIX
```

samcli/lib/utils/subprocess_utils.py
```diff
@@ -92,10 +96,18 @@ def _print_loading_pattern():
                     # we read from subprocess stdout to avoid the deadlock process.wait function
                     # for more detail check this python bug https://bugs.python.org/issue1256
                     for line in process.stdout:
-                        decoded_line = _check_and_process_bytes(line)
+                        is_error = (
+                            not IS_NOT_WINDOWS
```
mildaniel, line 100: nit: Can we maybe flip the constant to clean up this condition? I think it would make it easier to read since this double negative reads a little awkward. 

`IS_WINDOWS` instead of `IS_NOT_WINDOWS`

mildaniel (approved):
Thanks Mohamed! Left a few minor comments but LGTM

samcli/lib/utils/subprocess_utils.py
moelasmar: No, I depended on what returned from terraform while testing

samcli/lib/utils/subprocess_utils.py
moelasmar: No, line[0:4] will return bytes string and I wasn't able to build the same bytes string to compare

mildaniel (approved):

samcli/lib/utils/subprocess_utils.py
mndeveci: Can we add some documentation about what are their meaning and where we can find more information?

samcli/lib/utils/subprocess_utils.py
```diff
@@ -92,10 +96,18 @@ def _print_loading_pattern():
                     # we read from subprocess stdout to avoid the deadlock process.wait function
                     # for more detail check this python bug https://bugs.python.org/issue1256
                     for line in process.stdout:
-                        decoded_line = _check_and_process_bytes(line)
+                        is_error = (
+                            IS_WINDOWS
+                            and len(line) >= len(TERRAFORM_ERROR_PREFIX)
+                            and line[0:4] == bytes(TERRAFORM_ERROR_PREFIX)
+                        )
```
mndeveci, line 103: This utility module and function seems very generic, can we put terraform specific stuff behind a flag so that this will be only applied to the places where it is been called for terraform stuff?

mndeveci (commented):
Thanks for fixing this issue @moelasmar!

One other thing I was wondering is setting some timeout to the process calls that we are making. Can we add some timeouts for them so that we won't be waiting indefinitely?

samcli/lib/utils/subprocess_utils.py
moelasmar: done

mndeveci (approved):

PR #5933, opened 2023-09-13
feat: Building projects in source

lucashuy:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


appveyor-linux-binary.yml
```diff
@@ -19,7 +19,7 @@ environment:
   PYTHON_HOME: "$HOME/venv3.11/bin"
   PYTHON_VERSION: '3.11'
   AWS_DEFAULT_REGION: us-east-1
-  NODE_VERSION: "14.17.6"
+  NODE_VERSION: "18.18.2"
```
mndeveci, line 22: Does this affect other tests? And do we need to change settings for windows as well?

samcli/cli/types.py
```diff
@@ -474,3 +475,61 @@ def convert(self, value, param, ctx):
         LOG.debug("Converting provided %s option value to Enum", param.opts[0])
         value = super().convert(value, param, ctx)
         return self.enum(value)
+
+
+class SyncWatchExcludeType(click.ParamType):
```
mndeveci, line 480: Apologies that I missed in the original review but what happens if user provides multiple exclude folders for same function? Is it going to work or one will replace the other?

mndeveci (commented):
🥳 LGTM! Left 2 small-ish comments.

samcli/cli/types.py
lucashuy: If this is provided multiple times, they'll all be excluded. If a user passes in:
- `--watch-exclude Function=file1`
- `--watch-exclude Function=file2`
- `--watch-exclude Function=file3`
all three files will be excluded from being watched.

Same thing with folders, in the case that the user passes in
- `--watch-exclude Function=folder/child1`
- `--watch-exclude Function=folder`
they both will still be passed in, however this is just the same as passing in `--watch-exclude Function=folder`

appveyor-linux-binary.yml
lucashuy: The tests are passing on Windows (ref: https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-canary-windows-dev/builds/48658868/job/ql8agf8u9fgncwtj?fullLog=true), though it might be worth it to pin `npm` on Windows too.

mildaniel (approved):

PR #5972, opened 2023-09-21
fix: Invoke stderr stream missing stack trace

mildaniel:

N/A

After updating `sam local` commands to use the RIE in version 1.13.1, when a stack trace is streamed from the emulation container it contains carriage return characters that then don't get displayed by the stream writer.

Replaces the carriage return characters with the system's linesep character.

Will display addition logs when using local invoke commands that contain a stack trace.

Before:
```
Mounting C:\Users\Administrator\ws\sam-app\.aws-sam\build\HelloWorldFunction as /var/task:ro,delegated, inside runtime container
START RequestId: 06dd2e5e-af4c-47b0-b893-a9979544e244 Version: $LATEST
    "body": son.dumps({y", line 38, in lambda_handler
END RequestId: 06dd2e5e-af4c-47b0-b893-a9979544e244
REPORT RequestId: 06dd2e5e-af4c-47b0-b893-a9979544e244  Init Duration: 0.15 ms  Duration: 155.80 ms     Billed Duration: 156 ms Memory Size: 128 MB     Max Memory Used: 128 MB
```

After:
```
Mounting C:\Users\Administrator\ws\sam-app\.aws-sam\build\HelloWorldFunction as /var/task:ro,delegated, inside runtime container
START RequestId: 5ff38d1a-03ed-426a-8a4e-aef0b37bf0de Version: $LATEST
[ERROR] NameError: name 'son' is not defined
  File "/var/task/app.py", line 38, in lambda_handler
    "body": son.dumps({
END RequestId: 5ff38d1a-03ed-426a-8a4e-aef0b37bf0de
REPORT RequestId: 5ff38d1a-03ed-426a-8a4e-aef0b37bf0de  Init Duration: 0.32 ms  Duration: 110.39 ms     Billed Duration: 111 ms Memory Size: 128 MB     Max Memory Used: 128 MB
```

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -469,6 +469,7 @@ class Container:
 
     @staticmethod
     def _handle_data_writing(output_stream: Union[StreamWriter, io.BytesIO, io.TextIOWrapper], output_data: bytes):
+        output_data = output_data.decode("utf-8").replace("\r", os.linesep).encode("utf-8")
         if isinstance(output_stream, StreamWriter):
             output_stream.write_bytes(output_data)
             output_stream.flush()
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -469,6 +469,10 @@ class Container:
 
     @staticmethod
     def _handle_data_writing(output_stream: Union[StreamWriter, io.BytesIO, io.TextIOWrapper], output_data: bytes):
+        # Decode the output and strip the string of carriage return characters. Stack traces are returned
+        # with carriage returns from the RIE. If these are left in the string then only the last line after
+        # the carriage return will be printed instead of the entire stack trace. Encode the string after cleaning
+        # to be printed by the correct output stream
         output_data = output_data.decode("utf-8").replace("\r", os.linesep).encode("utf-8")
         if isinstance(output_stream, StreamWriter):
             output_stream.write_bytes(output_data)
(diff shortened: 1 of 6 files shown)


samcli/local/docker/container.py
```diff
@@ -469,6 +469,7 @@ def _write_container_output(
 
     @staticmethod
     def _handle_data_writing(output_stream: Union[StreamWriter, io.BytesIO, io.TextIOWrapper], output_data: bytes):
+        output_data = output_data.decode("utf-8").replace("\r", os.linesep).encode("utf-8")
```
mndeveci, line 472: Might add a comment here which points to this PR for details. Also, did we confirm this behavior on Windows?

mndeveci (commented):
Thanks for fixing this issue! LGTM in general, left a small comment

samcli/local/docker/container.py
mildaniel: Yes, I validated on Windows. I can add a comment.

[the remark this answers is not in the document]
mildaniel: The method writes to other stream types such as `io.BytesIO` and `io.TextIOWrapper` which take bytes so I figured re-encoding would require less overall change.

sriram-mv (approved):
Overall looks good to me.

Please add a comment in the code which draws attention to this change with something like "Important! DO NOT CHANGE", and an explanation of why. Because the decode, encode can seem like weird behavior.

mndeveci (approved):

PR #6013, opened 2023-09-29
test: Refactor Terraform Build Tests

mildaniel:

N/A

Currently, the Terraform-related build tests continuously re-build the same function code using the same build method for 30+ functions in several different test classes. This is redundant and takes a long time to execute without providing much additional validation. Currently, 5 workers are required to run these tests on AppVeyor with the longest test class taking over an hour.

This PR reworks the way these tests are executed in a way that requires a single runner with all tests being completed in < 1 hour.

Instead the new testing flow is:
1. Build a single function. This will produce the metadata file for all functions but will only run the build method once per test class.
2. Validate that the resources produced in the metadata file are correct by comparing against an expected `template.yaml` file.
3. Test local invoke against the single built function instead of all functions (the function code is the same for them all anyways, no point in invoking them all). 

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/buildcmd/test_build_terraform_applications.py
```diff
@@ -50,20 +60,83 @@ def setUpClass(cls):
...
-        process = Popen(command_list, stdout=PIPE, stderr=PIPE, stdin=PIPE, env=env, cwd=self.working_dir)
+        process = Popen(
+            command_list, stdout=PIPE, stderr=PIPE, stdin=PIPE, env=env, cwd=self.terraform_application_execution_path
+        )
         try:
             (stdout, stderr) = process.communicate(input=input, timeout=timeout)
             return stdout, stderr, process.returncode
         except TimeoutExpired:
             process.kill()
             raise
```
mndeveci, line 72: nit: Why not using the common `runCommand` method that we have? Are there any differences for Terraform tests?

tests/integration/buildcmd/test_build_terraform_applications.py
```diff
@@ -50,20 +60,83 @@ def setUpClass(cls):
...
+            return
+
+        command_list_parameters = {"function_identifier": self.function_identifier, "hook_name": "terraform"}
+        if self.build_in_container:
+            command_list_parameters["use_container"] = True
+            command_list_parameters["build_image"] = self.docker_tag
+            if self.override:
+                command_list_parameters[
+                    "container_env_var"
+                ] = "TF_VAR_HELLO_FUNCTION_SRC_CODE=./artifacts/HelloWorldFunction2"
```
mndeveci, line 87: Will `/` be a problem on Windows?

mildaniel: This shouldn't be an issue. This part of the code was just moved and isn't anything new.

tests/integration/buildcmd/test_build_terraform_applications.py
mildaniel: I think there was a reason initially but I can't remember what that is. I don't think we need it anymore but we've already coupled many of our test cases to this, so I'm just going to use this method as a wrapper to call our existing `run_command()` (since the cases expect a slightly different response).

lucashuy (approved):

mndeveci (approved):

Issue #6033, opened 2023-10-04
Bug: `sam local start-lambda` encounters an exception after running the lambda's app code


sbchisholm:
### Description:
`sam local start-lambda` encounters the following exception after running lambda app code:
```
AttributeError: '_io.StringIO' object has no attribute 'buffer'
```
The application code runs without issue but then `sam` is hitting to the following exception (see traceback below).

### Steps to reproduce:
1. run sam using:  `sam local start-lambda`
1.  invoke the lambda on the local sam instance (in my case I'm using the `RequestResponse` invocation type)

### Observed result:
```
Exception on /2015-03-31/functions/FunctionName/invocations [POST]
Traceback (most recent call last):
  File "/home/steve/.local/lib/python3.8/site-packages/flask/app.py", line 2529, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/steve/.local/lib/python3.8/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/steve/.local/lib/python3.8/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/steve/.local/lib/python3.8/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/local/lambda_service/local_lambda_invoke_service.py", line 169, in _invoke_request_handler
    self.lambda_runner.invoke(function_name, request_data, stdout=stdout_stream_writer, stderr=self.stderr)
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/commands/local/lib/local_lambda.py", line 144, in invoke
    self.local_runtime.invoke(
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/lib/telemetry/metric.py", line 324, in wrapped_func
    return_value = func(*args, **kwargs)
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/local/lambdafn/runtime.py", line 201, in invoke
    container.wait_for_result(
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/local/docker/container.py", line 368, in wait_for_result
    stdout.write_bytes(response)
  File "/home/steve/.local/lib/python3.8/site-packages/samcli/lib/utils/stream_writer.py", line 35, in write_bytes
[...]

PR #6034, opened 2023-10-04
fix: Writing to bytes TextIO stream in StreamWriter causing AttributeError on buffer member

sbchisholm:
#6033

Fixes the bug raise in #6033 where the instance of the stream member in StreamWriter does not have the buffer attribute.

Rather than using the buffer attribute to write bytes, the bytes are converted to a string first with decode and then written with the write method (which accepts strings)

Unsure.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


wolverian:
Looking at the users of this function, for example https://github.com/aws/aws-sam-cli/blob/8126aaac29ff8706b14f29a126ab7bb829d81b24/samcli/local/docker/container.py#L476-L478 and https://github.com/aws/aws-sam-cli/blob/8126aaac29ff8706b14f29a126ab7bb829d81b24/samcli/commands/remote/remote_invoke_context.py#L276 would it make more sense to remove this function and change the calls to use `write_str` instead?

sbchisholm:
> Looking at the users of this function, for example
> 
> https://github.com/aws/aws-sam-cli/blob/8126aaac29ff8706b14f29a126ab7bb829d81b24/samcli/local/docker/container.py#L476-L478
> 
> and
> https://github.com/aws/aws-sam-cli/blob/8126aaac29ff8706b14f29a126ab7bb829d81b24/samcli/commands/remote/remote_invoke_context.py#L276
> 
> would it make sense to remove this function and change the calls to use `write_str` instead?

If that's true, then that would make sense but it looks to also be used in here:
https://github.com/aws/aws-sam-cli/blob/8126aaac29ff8706b14f29a126ab7bb829d81b24/samcli/local/docker/container.py#L367-L368

wolverian:
Sure, but does that make sense? In that case, if this PR is merged, the bytes are decoded to a string anyway, it's just hidden from the caller and could surprise them if they assumed the bytes wouldn't be interpreted in any specific encoding. 

Rather, either this abstraction should offer a real way to write raw bytes with no interpretation (decoding) _or_ it should only offer a way to write strings. Hiding these string to bytes and back conversions will shoot the user in the foot eventually.

sbchisholm:
> Sure, but does that make sense? In that case, if this PR is merged, the bytes are decoded to a string anyway, it's just hidden from the caller and could surprise them if they assumed the bytes wouldn't be interpreted in any specific encoding.
> 
> Rather, in my opinion, either this abstraction should offer a way to write raw bytes with no interpretation (decoding) _or_ it should _only_ offer a way to write strings.
> 
> Hiding these bytes to string conversions will shoot the user in the foot eventually.

That's a great point, I'll update the PR.

sriram-mv:
I'm adding both @jfuss and @mildaniel as reviewers given their previous experience in bytes/str operation.

My take is that we are explicitly changing behavior from bytes to str and there could be side-effects. There was an explicit PR to decode - replace carriage returns - reencode. That behavior seems to be removed at this point.

wolverian:
> My take is that we are explicitly changing behavior from bytes to str and there could be side-effects. There was an explicit PR to decode - replace carriage returns - reencode. That behavior seems to be removed at this point.

It might be a good idea to triage which commit triggered this bug. That might help with figuring out what the right thing to do here is. FWIW,  I think this PR looks logical, but I'm unfamiliar with this codebase and its history.

This is fairly high in priority in my opinion as I simply can't use the latest version SAM CLI right now due to this bug.

mildaniel:
I did some digging here and it seems as though the issue is that the `StreamWriter` class expects the stream to be of type `TextIO`, but there are times where an `io.StringIO` is passed in https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/lambda_service/local_lambda_invoke_service.py#L165. 

`io.StringIO` is not of type `TextIO` and does not have a buffer property that `StreamWriter` expects of a `TextIO` object. I'm not sure why this wasn't picked up by any type checking, but in any case, causes this exception whenever it's called with an `io.StringIO` stream. Changing the instantiators of the `StreamWriter` class to all use `TextIO` would require pretty significant refactoring since those instances also use methods and properties unique to `io.StringIO`. 

Although this PR is more of a solution to the symptom rather than root cause, I'm inclined to go with it so I don't really see any negatives here and it allows us to patch an important issue. Another solution is to update the type in the `StreamWriter` class to be a union of `StreamWriter`, `io.TextIO` and `io.TextIOWrapper` and handle each one in the implementation of the `write` method.

@sriram-mv what are your thoughts here?

mildaniel (approved):

moelasmar (approved):

Issue #6023, opened 2023-10-03
Bug: sam build - DockerImagePullFailedException


chostrander:
### Description:

using aws samples:  https://github.com/aws-samples/aws-sam-cli-java-examples/tree/main/03-maven-graalvm-custom

I execute this command per the instructions: sam build --use-container --build-image sam/custom-graal-image

I get the following:
`Starting Build inside a container                                                                                                                                                                         
Building codeuri: /software/projects/aws-sam-cli-java-examples/03-maven-graalvm-custom/HelloWorldFunction runtime: provided.al2 metadata: {'BuildMethod': 'makefile'} architecture: x86_64 functions:     
HelloWorldFunctionGraalVM                                                                                                                                                                                 

Error: Could not find sam/custom-graal-image image locally and failed to pull it from docker.
Traceback:
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1688, in invoke
  File "click/core.py", line 1434, in invoke
  File "click/core.py", line 783, in invoke
  File "click/decorators.py", line 92, in new_func
  File "click/core.py", line 783, in invoke
  File "samcli/lib/telemetry/metric.py", line 184, in wrapped
  File "samcli/lib/telemetry/metric.py", line 149, in wrapped
  File "samcli/lib/utils/version_checker.py", line 42, in wrapped
  File "samcli/cli/main.py", line 95, in wrapper
  File "samcli/commands/build/command.py", line 170, in cli
  File "samcli/commands/build/command.py", line 266, in do_cli
  File "samcli/commands/build/build_context.py", line 281, in run
  File "samcli/lib/build/app_builder.py", line 217, in build
  File "samcli/lib/build/build_strategy.py", line 81, in build
  File "samcli/lib/build/build_strategy.py", line 91, in _build_functions
  File "samcli/lib/build/build_strategy.py", line 164, in build_single_function_definition
  File "samcli/lib/build/app_builder.py", line 697, in _build_function
  File "samcli/lib/build/app_builder.py", line 939, in _build_function_on_container
  File "samcli/local/docker/manager.py", line 115, in run
  File "samcli/local/docker/manager.py", line 88, in create

An unexpected error was encountered while executing "sam build".
Search for an existing issue:                                                                                                                                                                             
https://github.com/aws/aws-sam-cli/issues?q=is%3Aissue+is%3Aopen+Bug%3A%20sam%20build%20-%20DockerImagePullFailedException                                                                                
[...]

PR #6039, opened 2023-10-04
fix: gracefully exit on not finding a local image

sriram-mv:
- For `build` with `--use-container` and `--build-image`

https://github.com/aws/aws-sam-cli/issues/6023

Gracefully fail and not throw a traceback

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -40,9 +40,10 @@ from samcli.lib.utils.colors import Colored, Colors
 from samcli.lib.utils import osutils
 from samcli.lib.utils.packagetype import IMAGE, ZIP
 from samcli.lib.utils.stream_writer import StreamWriter
+from samcli.local.docker.exceptions import ContainerNotStartableException
 from samcli.local.docker.lambda_build_container import LambdaBuildContainer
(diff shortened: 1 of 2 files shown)


samcli/lib/build/app_builder.py
```diff
@@ -960,6 +961,10 @@ def _build_function_on_container(
             # "/." is a Docker thing that instructions the copy command to download contents of the folder only
             result_dir_in_container = response["result"]["artifacts_dir"] + "/."
             container.copy(result_dir_in_container, artifacts_dir)
+
+        except (DockerImagePullFailedException) as ex:
+            raise BuildInsideContainerError(ex)
```
mndeveci, line 966: nit: Should we adjust any wording inside the exception or relaying is OK (not sure what will be the output here)?

mndeveci (approved):

hawflau (approved):

samcli/lib/build/app_builder.py
sriram-mv: Relay should be fine IMO.

```
Error: Could not find sam/custom-graal-iage image locally and failed to pull it from docker.
```

its pretty clear.

Issue #6053, opened 2023-10-09
Bug: sam local start-api with debug port runs into TypeError


darhsu:
<!-- Make sure we don't have an existing Issue that reports the bug you are seeing (both open and closed). 
If you do find an existing Issue, re-open or add a comment to that Issue instead of creating a new one. -->

### Description:
<!-- Briefly describe the bug you are facing.-->
When running `sam local start-api` with `debug_port = 5858` in the `samconfig.toml` file, I get a `TypeError: 'Integer' object is not iterable`.

However, there's no error when running `sam local start-api --debug-port 5858` when adding the --debug-port flag directly in the command. 


### Steps to reproduce:
<!-- Provide detailed steps to replicate the bug, including steps from third party tools (CDK, etc.) --> 
1. `sam init` # create hello world example with Python and zip
2. Add `debug_port = 5858` in `samconfig.toml`:
``` toml
[default.local_start_api.parameters]
warm_containers = "EAGER"
debug_port = 5858
```
3. `sam build`
4. `sam local start-api`



### Observed result:
<!-- Please provide command output with `--debug` flag set. -->
``` bash
/workspaces/sam-app $ sam local start-api --debug
2023-10-09 05:11:18,765 | Config file location: /workspaces/sam-app/samconfig.toml                                                                                                                                                
2023-10-09 05:11:18,769 | Loading configuration values from [default.['local', 'start-api'].parameters] (env.command_name.section) in config file at '/workspaces/sam-app/samconfig.toml'...                                      
2023-10-09 05:11:18,770 | Configuration values successfully loaded.                                                                                                                                                               
2023-10-09 05:11:18,771 | Configuration values are: {'stack_name': 'sam-app', 'warm_containers': 'EAGER', 'debug_port': 5858}                                                                                                     
2023-10-09 05:11:18,775 | Using SAM Template at /workspaces/sam-app/.aws-sam/build/template.yaml                                                                                                                                  
Traceback (most recent call last):
  File "samcli/__main__.py", line 12, in <module>
  File "click/core.py", line 1157, in __call__
  File "click/core.py", line 1078, in main
  File "click/core.py", line 1688, in invoke
  File "click/core.py", line 1686, in invoke
  File "click/core.py", line 943, in make_context
  File "click/core.py", line 1408, in parse_args
  File "click/core.py", line 2396, in handle_parse_result
  File "click/core.py", line 2950, in consume_value
TypeError: 'Integer' object is not iterable
[...]

PR #6108, opened 2023-10-18
fix: Handle input parameters classified as multiple value click options

mildaniel:

https://github.com/aws/aws-sam-cli/issues/6053

Click will handle single values and convert them to a list before being processed if the option is set to `multiple=True`. When using a `samconfig` file, the values are injected into click after this conversion resulting in click crashing when these values are processed. This leads to a discrepancy in behaviour between command line options and configuration file options.

Check the configured resources in the `samconfig` file and ensure the same type of option processing is done to convert single value types to a list before giving them to click.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -147,6 +147,10 @@ def handle_parse_options(resolved_config: dict) -> None:
                 allow_multiple = options_map[config_name].multiple
                 if allow_multiple and not isinstance(config_value, list):
                     resolved_config[config_name] = [config_value]
+                    LOG.debug(
+                        f"Adjusting value of {config_name} to be a list "
+                        f"since this option is defined with 'multiple=True'"
+                    )
             except (AttributeError, KeyError):
                 LOG.debug(f"Unable to parse option: {config_name}. Leaving option as inputted")
 
--- a/samcli/cli/cli_config_file.py
+++ b/samcli/cli/cli_config_file.py
@@ -103,7 +103,7 @@ class ConfigProvider:
             # so that click defaults work out of the box.
             resolved_config = dict(samconfig.get_all(self.cmd_names, self.section, env=config_env).items())
             handle_parse_options(resolved_config)
-            LOG.debug("Configuration values successvfully loaded.")
+            LOG.debug("Configuration values successfully loaded.")
             LOG.debug("Configuration values are: %s", resolved_config)
 
         except KeyError as ex:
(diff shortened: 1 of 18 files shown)


samcli/cli/cli_config_file.py
```diff
@@ -123,6 +124,51 @@ def __call__(self, config_path: Path, config_env: str, cmd_names: List[str]) ->
...
+
+
+def get_options_map() -> dict:
+    """
+    Attempt to get all of the options that exist for a command.
+    Return a mapping from each option name to that options' properties.
+
+    Returns
+    -------
+    Optional[dict]
```
mndeveci, line 161: nit this is not optional though since we are returning an empty array in case of an error

samcli/cli/cli_config_file.py
```diff
@@ -123,6 +124,51 @@ def __call__(self, config_path: Path, config_env: str, cmd_names: List[str]) ->
...
+    Parameters
+    ----------
+    resolved_config: dict
+        Configuration options extracted from the configuration file
+    """
+    options_map = get_options_map()
+    for config_name, config_value in resolved_config.items():
+        if config_name in options_map:
+            try:
+                allow_multiple = options_map[config_name].multiple
```
mndeveci, line 147: Is there a chance that `.multiple` won't be available in the object which is returned in the dictionary? We might want to be more defensive approach here.

mildaniel: There is a chance, which is why the except block below catches an `AttributeError`. There is a unit test case that covers this scenario.

samcli/cli/cli_config_file.py
mildaniel: Oops, changed the code but forgot to change the docstring.

samcli/cli/cli_config_file.py
```diff
@@ -123,6 +124,51 @@ def __call__(self, config_path: Path, config_env: str, cmd_names: List[str]) ->
...
+                if allow_multiple and not isinstance(config_value, list):
+                    resolved_config[config_name] = [config_value]
```
mndeveci, line 149: Can we add a debug log here so that we will know that we altered the original value?

mndeveci (approved):

samcli/cli/cli_config_file.py
mildaniel: Done

hnnasit (approved):

PR #6132, opened 2023-10-23
chore(tests): Node build in source test

lucashuy:
#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


tests/integration/buildcmd/test_build_in_source.py
```diff
@@ -107,3 +112,71 @@ def test_builds_successfully_with_local_dependency(self):
...
+        [
+            (True, True),  # build in source
+            (False, False),  # don't build in source
+            (None, False),  # use default for workflow (don't build in source)
+        ]
+    )
+    @pytest.mark.flaky(reruns=3)
+    def test_builds_successfully_without_local_dependencies(self, build_in_source, dependencies_expected_in_source):
+        self.codeuri_path = Path(self.test_data_path, "Node")
+        self.source_directories = [str(self.codeuri_path)]
```
mildaniel, line 152: I think we should move the "source" to still be a temp directory. If we run this in parallel, this directory becomes a shared resource that could cause issues. It also means we won't pollute the test data directory in case we forget to clean something up.

tests/integration/buildcmd/test_build_in_source.py
```diff
@@ -107,3 +112,71 @@ def test_builds_successfully_with_local_dependency(self):
...
+
+        overrides = self.get_override(
+            runtime="nodejs16.x", code_uri=self.CODE_URI, architecture="x86_64", handler="main.lambdaHandler"
```
mildaniel, line 155: nit: Let's use the latest supported nodejs version

mildaniel (approved):

mildaniel (changes requested):
Left a few minor comments

tests/integration/buildcmd/test_build_in_source.py
```diff
@@ -107,3 +115,63 @@ def test_builds_successfully_with_local_dependency(self):
...
+        run_command(command_list, self.working_dir)
+        self.validate_node_modules(expected_built_in_source)
+
+    @pytest.mark.flaky(reruns=3)
+    def test_builds_successfully_with_local_dependency(self):
+        codeuri_folder = Path("Esbuild", "NodeWithLocalDependency")
+
+        shutil.copytree(Path(self.test_data_path, "Esbuild"), self.working_dir, dirs_exist_ok=True)
+        self.codeuri_path = Path(self.test_data_path, codeuri_folder)
+        self.CODE_URI = str(codeuri_folder)
```
mildaniel, line 169: Is this property used anywhere else by a parent class?

tests/integration/buildcmd/test_build_in_source.py
```diff
@@ -107,3 +115,63 @@ def test_builds_successfully_with_local_dependency(self):
...
+            runtime="nodejs18.x", code_uri=self.codeuri_path, architecture="x86_64", handler="main.lambdaHandler"
+        )
+        command_list = self.get_command_list(build_in_source=build_in_source, parameter_overrides=overrides, debug=True)
+
+        run_command(command_list, self.working_dir)
+        self.validate_node_modules(expected_built_in_source)
+
+    @pytest.mark.flaky(reruns=3)
+    def test_builds_successfully_with_local_dependency(self):
+        codeuri_folder = Path("Esbuild", "NodeWithLocalDependency")
```
mildaniel, line 165: Is it possible to restructure the testdata in a way where we can move the copytree logic to the setup methods? The copying logic isn't part of the test case itself but sets up the environment for the tests to run in, I feel like that logic should be abstracted away like in other test classes.

tests/integration/buildcmd/test_build_in_source.py
lucashuy: Not anymore, this was part of a previous validation method that was never committed. Will remove to not conflict with anything in the future.

tests/integration/buildcmd/test_build_in_source.py
lucashuy: Thats a good point. As it is right now, the copy here copies this specific test data folder and not the entire `buildcmd` test data folder to save time. We could move this to the setup and make it more broad (copy the entire folder), or we could just move those two tests into the scratch folder from within setup. I think that restructuring is also possible, but might also need to be updated else where/copied again for this test.

In any case, I can look to improve this.

lucashuy: Updated test class setup to copy the `"Esbuild"` folder. We can use this instead for both normal and local dep testing since it contains all the folders for both tests.

mildaniel (approved):
Thanks for addressing the comments. LGTM!

tests/integration/buildcmd/test_build_in_source.py
```diff
@@ -17,15 +22,22 @@ class TestBuildCommand_BuildInSource_Makefile(BuildIntegProvidedBase):
...
+
+        self.code_uri = "provided_create_new_file"
+        test_data_code_uri = Path(self.test_data_path, self.code_uri)
+        self.file_created_from_make_command = "file-created-from-make-command.txt"
+
+        scratch_code_uri_path = Path(self.working_dir, self.code_uri)
+        self.code_uri_path = str(scratch_code_uri_path)
+
+        # copy source code into temporary directory and update code uri to that scratch dir
+        osutils.copytree(test_data_code_uri, scratch_code_uri_path)
```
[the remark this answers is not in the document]
lucashuy: The base class (`BuildIntegBase`) doesn't appear to be copying any test files, it looks to just create the scratch directories.

mndeveci (approved):
LGTM, let's make sure they are not clashing with each other or with any other existing tests since we are running them in parallel.

PR #6236, opened 2023-11-09
feat: Resolve root level symlinked folders in Layers

lucashuy:

None.

As a result of the new build in source behaviour, Nodejs Layers will contain a `node_modules` folder that is a symbolic link to the source `node_modules`. Mounting this will not work in Docker.

Resolves this link using the same logic as Functions before the container is created.

All Layers that contain symlinked resources at the root level will be resolved.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_container.py
+++ b/samcli/local/docker/lambda_container.py
@@ -2,8 +2,11 @@
 Represents Lambda runtime containers.
 """
 import logging
-from typing import List
+from pathlib import Path
+from typing import Dict, List
 
+from samcli.lib.build.workflow_config import LAYER_SUBFOLDERS
+from samcli.lib.providers.provider import LayerVersion
 from samcli.lib.utils.packagetype import IMAGE
 from samcli.local.docker.lambda_debug_settings import LambdaDebugSettings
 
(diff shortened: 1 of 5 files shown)


samcli/local/docker/lambda_container.py
```diff
@@ -133,11 +137,60 @@ def __init__(
...
+                    LOG.debug(
+                        "Skipping symlink check for layer %s, layer does not have locally defined resources",
+                        layer.layer_id,
+                    )
+                    continue
+
+                # eg. `.aws-sam/build/MyLayer` + `nodejs`
+                artifact_layer_path = Path(layer.codeuri, layer_folder)
+                # eg. `/opt` + `nodejs`
+                container_bind_path = Path(LambdaImage._LAYERS_DIR, layer_folder)
```
mndeveci, line 185: Curious, what happens if there are multiple layers used by the function? Are we going to accept that last one will replace the earlier mappings?

lucashuy: Updated, turns out that the tar library includes an option that lets it resolve symlinks for us. No need to write custom logic to do this.

mildaniel (approved):

mndeveci (approved):

PR #6258, opened 2023-11-13
Return the deployment result in exit code for all failures

demolitionmode:

#5103

When a stack fails to deploy, the exit code should reflect this regardless of it being rolled back (through use of `--on-failure DELETE`)

Setting `--on-failure DELETE` no longer skips the `raise` line; rather, the condition will only determine whether `deployer.rollback_delete_stack()` is called or not.

Users who rely on the previous behaviour would need to ignore this fix, with something like `sam deploy || true`

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel:
Thanks @demolitionmode for raising this PR. As you mentioned, this is technically a breaking change that can impact existing customer. We might need a different way to enable this behaviour that is opt-in instead of by default, potentially a new command line flag. 

I will bring this to the team to discuss the best approach.

mildaniel:
After discussing with the team, we think it's okay to proceed with this PR. When ready, we will create an announcement explain this change in behaviour to existing customers.

mildaniel (approved):

jysheng123 (approved):

Issue #6253, opened 2023-11-13
Bug: crashes if Function is missing `Handler` property (`Error: Invalid type <class 'NoneType'>`)


davidjb:
### Description:

When a `AWS::Serverless::Function` is defined without a `Handler` property, attempting to `sam build` crashes out with a traceback.

I'm aware that `Handler` is a required property when `PackageType` is `Zip` (the default), but this crash should be avoided and be more descriptive of what the problem is.

### Steps to reproduce:

1. Create a `template.yml` like so:
   ```yaml
    AWSTemplateFormatVersion: "2010-09-09"
    Transform: AWS::Serverless-2016-10-31
    Resources:
      MyFunction:
        Type: AWS::Serverless::Function
        Metadata:
          BuildMethod: makefile
        Properties:
          CodeUri: foo
          #Handler: bootstrap
          Runtime: provided.al2
          Architectures: [ arm64 ]
   ```
2. Run `sam build`.
3. Obverse crash and traceback as below. Uncomment the `#Handler` line for initialisation to succeed.

### Observed result:

```sh
# sam build --debug
2023-11-13 05:42:24,461 | No config file found in this directory.
2023-11-13 05:42:24,463 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/samconfig.toml'
2023-11-13 05:42:24,464 | Config file location: /samconfig.toml
2023-11-13 05:42:24,465 | Config file '/samconfig.toml' does not exist
2023-11-13 05:42:24,466 | Using SAM Template at /template.yml
2023-11-13 05:42:24,476 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/samconfig.toml'
2023-11-13 05:42:24,476 | Using config file: samconfig.toml, config environment: default
2023-11-13 05:42:24,477 | Expand command line arguments to:
2023-11-13 05:42:24,477 | --template_file=/template.yml --mount_with=READ --build_dir=.aws-sam/build --cache_dir=.aws-sam/cache
2023-11-13 05:42:24,506 | 'build' command is called
2023-11-13 05:42:24,507 | No Parameters detected in the template
2023-11-13 05:42:24,518 | There is no customer defined id or cdk path defined for resource MyFunction, so we will use the resource logical id as
the resource id
2023-11-13 05:42:24,519 | 0 stacks found in the template
2023-11-13 05:42:24,519 | No Parameters detected in the template
[...]

PR #6288, opened 2023-11-17
fix: crashing if handler is missing

sidhujus:

#6253 

There is currently no validation on if the handler is defined within a template

Checks if the handler is defined before trying to access and assign it

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/sam_function_provider.py
+++ b/samcli/lib/providers/sam_function_provider.py
@@ -446,9 +446,7 @@ class SamFunctionProvider(SamBaseProvider):
             codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
 
         package_type = resource_properties.get("PackageType", ZIP)
-        if package_type == ZIP and (
-            resource_properties.get("Handler") is None or resource_properties.get("Handler") == ""
-        ):
+        if package_type == ZIP and not resource_properties.get("Handler"):
             raise MissingFunctionHandlerException(f"Could not find handler for function: {name}")
 
         function_build_info = get_function_build_info(
--- a/samcli/lib/providers/sam_function_provider.py
+++ b/samcli/lib/providers/sam_function_provider.py
@@ -446,7 +446,9 @@ class SamFunctionProvider(SamBaseProvider):
             codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
 
         package_type = resource_properties.get("PackageType", ZIP)
-        if package_type == ZIP and resource_properties.get("Handler") is None:
+        if package_type == ZIP and (
+            resource_properties.get("Handler") is None or resource_properties.get("Handler") == ""
+        ):
             raise MissingFunctionHandlerException(f"Could not find handler for function: {name}")
 
         function_build_info = get_function_build_info(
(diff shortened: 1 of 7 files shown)


jysheng123 (commented):
I would add a unit test raising the specific error thrown without the handler present other than that LGTM

samcli/lib/providers/sam_function_provider.py
```diff
@@ -445,6 +446,9 @@ def _build_function_configuration(
             codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
 
         package_type = resource_properties.get("PackageType", ZIP)
+        if package_type == ZIP and resource_properties.get("Handler") is None:
```
lucashuy, line 449: Should we also validate for an empty handler (ie. `Handler: ""`)? If that's also an invalid case then we can include that in this check.

sidhujus: Yea thats a good call, ill add that in there as well

tests/unit/commands/local/lib/test_sam_function_provider.py
```diff
@@ -1780,6 +1782,23 @@ def test_must_skip_non_existent_properties(self):
         self.assertEqual(expected, result)
 
 
+class TestSamFunctionProvider_build_function_configuration(TestCase):
+    def test_must_convert(self):
```
jysheng123, line 1786: Think this test name should be descriptive like test_handler_missing_exception

jysheng123: Could also consider making another test for the emtpy handler string for maximum coverage

jysheng123 (approved):
Once the above comments are addressed it LGTM!

samcli/lib/providers/sam_function_provider.py
```diff
@@ -446,7 +446,9 @@ def _build_function_configuration(
             codeuri = SamLocalStackProvider.normalize_resource_path(stack.location, codeuri)
 
         package_type = resource_properties.get("PackageType", ZIP)
-        if package_type == ZIP and resource_properties.get("Handler") is None:
+        if package_type == ZIP and (
+            resource_properties.get("Handler") is None or resource_properties.get("Handler") == ""
+        ):
```
lucashuy, line 451: nit: this could be changed to a false-y check (not sure if we also wanted to validate type)

tests/unit/commands/local/lib/test_sam_function_provider.py
```diff
@@ -1781,6 +1782,23 @@ def test_must_skip_non_existent_properties(self):
         self.assertEqual(expected, result)
 
 
+class TestSamFunctionProvider_build_function_configuration(TestCase):
```
lucashuy, line 1785: Could we add/parameterize this test to also test for an empty string Handler?

lucashuy (commented):
Thanks for addressing the blank string Handler, left a comment about testing

tests/unit/commands/local/lib/test_sam_function_provider.py
sidhujus: ah yea i forgot to change the name from the test that i copied

lucashuy (approved):
Thanks for addressing the changes

PR #6297, opened 2023-11-17
fix: Update async utils to remove upcoming deprecated call

lucashuy:

None.

`asyncio.get_event_loop()` is being deprecated in the future. Deprecation warns are starting to appear, which will cause tests to fail since warnings are treated as errors.

Removes the call to `asyncio.get_event_loop()` as a default value for the tasks.

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/async_utils.py
+++ b/samcli/lib/utils/async_utils.py
@@ -1,15 +1,18 @@
 """
 Contains asyncio related methods and helpers
 """
-import asyncio
 import logging
+from asyncio import AbstractEventLoop, gather, new_event_loop
 from concurrent.futures.thread import ThreadPoolExecutor
 from functools import partial
+from typing import Callable, List, Optional
 
 LOG = logging.getLogger(__name__)
 
 
-async def _run_given_tasks_async(tasks, event_loop=asyncio.get_event_loop(), executor=None):
+async def _run_given_tasks_async(
+    tasks: List[Callable], event_loop: AbstractEventLoop, executor: Optional[ThreadPoolExecutor] = None
+) -> list:
     """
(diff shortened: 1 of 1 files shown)


samcli/lib/utils/async_utils.py
```diff
@@ -1,15 +1,18 @@
...
 from concurrent.futures.thread import ThreadPoolExecutor
 from functools import partial
+from typing import Callable, List, Optional
 
 LOG = logging.getLogger(__name__)
 
 
-async def _run_given_tasks_async(tasks, event_loop=asyncio.get_event_loop(), executor=None):
+async def _run_given_tasks_async(
+    tasks: List[Callable], event_loop: AbstractEventLoop, executor: Optional[ThreadPoolExecutor] = None
```
hawflau, line 14: Did we use the default value at all?

lucashuy: No, since the creation of this util the event loop was always specified: https://github.com/aws/aws-sam-cli/blob/bbaefffe3dd3e15741db5fbf422255df0ad4bd00/samcli/lib/utils/async_utils.py#L127-L131

mndeveci (approved):

samcli/lib/utils/async_utils.py
```diff
@@ -111,7 +116,7 @@ def add_async_task(self, function, *args):
         """
         self._async_tasks.append(partial(function, *args))
 
-    def run_async(self, default_executor=True):
+    def run_async(self, default_executor: bool = True) -> list:
```
mndeveci, line 119: Thanks for adding type definitions!

hawflau (approved):

PR #6299, opened 2023-11-17
feat: Add additional default excludes and exclude option for sync

lucashuy:

None.

With the new build in source behaviour, some files (a `package-lock.json` file for example) are being changed every time a sync flow runs. This causes another sync flow to be queued up, causing an infinite loop.

Adds some default excludes (`node_modules/`, and `package-lock.json`) from being watched in `sam sync --watch --build-in-source` and introduces a new command line option to pass in custom matches for files to exclude.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/sync/watch_manager.py
```diff
@@ -132,7 +135,10 @@ def _add_code_triggers(self) -> None:
         resource_ids = get_all_resource_ids(self._stacks)
         for resource_id in resource_ids:
             try:
-                trigger = self._trigger_factory.create_trigger(resource_id, self._on_code_change_wrapper(resource_id))
+                additional_excludes = self._watch_exclude.get(resource_id.resource_iac_id, [])
```
mndeveci, line 138: Is `resource_iac_id` equivalent of full-path in Function resource dataclass?

mildaniel (approved):

samcli/lib/utils/resource_trigger.py
```diff
@@ -17,7 +17,7 @@
 from samcli.lib.utils.resources import RESOURCES_WITH_LOCAL_PATHS
 from samcli.local.lambdafn.exceptions import FunctionNotFound, ResourceNotFound
 
-AWS_SAM_FOLDER_REGEX = "^.*\\.aws-sam.*$"
+DEFAULT_WATCH_IGNORED_RESOURCES = ["^.*\\.aws-sam.*$", "^.*node_modules.*$", "^.*package-lock\\.json$"]
```
mndeveci, line 20: Should we exclude the `package-lock.json` which is in the root folder of `CodeUri`? I thought we should only exclude the one which is inside `node_modules` folder.

samcli/lib/utils/resource_trigger.py
```diff
@@ -182,6 +183,10 @@ def __init__(
         self._on_code_change = on_code_change
         self.base_dir = base_dir
 
+        self._watch_exclude = [*DEFAULT_WATCH_IGNORED_RESOURCES]
+        for exclude in watch_exclude or []:
+            self._watch_exclude.append(f"^.*{exclude}.*$")
```
mndeveci, line 188: Do we need to wrap it with regex, can't we just let customers to define their own input?

samcli/lib/utils/resource_trigger.py
lucashuy: That's a good call out, I'll double check that this is not actually needed and will remove it.

samcli/lib/utils/resource_trigger.py
lucashuy: Did you mean letting customers define their own regex? We can do that, but it'll mean that the customer has to make sure they match the parent path correctly so that the resource is ignored

mndeveci (approved):
LGTM! Can we also add some tests to verify `samconfig.toml` behavior?

samcli/lib/utils/resource_trigger.py
mndeveci: Let's get this out like this, if customers want to add regex from their side, we can add this prefix & suffix conditionally.

Issue #6243, opened 2023-11-10
SAM deploy requires CAPABILITY_AUTO_EXPAND for the aws-sam-cli-managed-default stack in eu-south-1 region


cmogeni:
### Description:
When running the SAM deploy command in the eu-south-1 region, it requires CAPABILITY_AUTO_EXPAND for the "aws-sam-cli-managed-default". Currently the SAM CLI creates this stack with the "CAPABILITY_IAM" [capability](https://github.com/aws/aws-sam-cli/blob/3102565c30fb06f2165b1a53a0438773bd2dcf2a/samcli/lib/utils/managed_cloudformation_stack.py#L241) but in this case it requires CAPABILITY_AUTO_EXPAND

### Steps to reproduce:
The only difference with the normal flow is that the stack was "aws-sam-cli-managed-default" stack has been deleted, but it should still be possible to run "sam deploy" because that should auto-detect and recreate the stack, which it does but it runs into the capabilities issue described above

### Observed result:
sam deploy --stack-name mystack --region eu-south-1 --resolve-s3 --s3-prefix redirect-lambda --parameter-overrides MyParam=param Myparam2=param2 --debug

```
2023-11-10 11:51:03,053 | Config file location: /myproject/samconfig.toml
2023-11-10 11:51:03,057 | Loading configuration values from [default.['deploy'].parameters] (env.command_name.section) inconfig file at'/myproject/samconfig.toml'...
2023-11-10 11:51:03,059 | Configuration values successfully loaded.
2023-11-10 11:51:03,060 | Configuration values are: {'stack_name': 'mystack', 'capabilities':'CAPABILITY_IAM CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND', 'confirm_changeset': False, 'resolve_s3': False}
2023-11-10 11:51:03,074 | Using SAM Template at /myproject/.aws-sam/build/template.yaml
2023-11-10 11:51:03,144 | Using config file: samconfig.toml, config environment: default
2023-11-10 11:51:03,146 | Expand command line arguments to:
2023-11-10 11:51:03,148 | --template_file=/myproject/.aws-sam/build/template.yaml --stack_name=mystack --resolve_s3 --s3_prefix=redirect-lambda --fail_on_empty_changeset --on_failure=ROLLBACK --capabilities=['CAPABILITY_IAM', 'CAPABILITY_NAMED_IAM', 'CAPABILITY_AUTO_EXPAND']
[...]

PR #6309, opened 2023-11-21
Add capabilities for Managed Stack

jysheng123:
#6243

Some regions (need to investigate why this is the case) have some flaky runs where sometimes the managed stack need special permissions to create that it does not need in normal regions. Note that I say flaky here because I was not able to recreate the error but a fellow developer was able to so there is some mismatch there.

Add the capabilities needed so that it does not matter what region is the stack is being deployed in

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


lucashuy (approved):

jysheng123:
Pull Request is not needed as this is a CFN bug that will be fixed next release on their side.

PR #6331, opened 2023-11-23
fix(test): Hanging init tests with new structured logging input

mildaniel:

N/A

Addresses the issue of the canaries hanging for 6 hours. The new structured logging input added an additional prompt to the guided init flow, causing the test to halt while waiting for an input that was never provided.

1. Provides required input for interactive prompts that are halted and waiting for structured logging input
2. Adds a timeout to the interactive init tests so that in the future we don't see this kind of hanging

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -43,28 +43,6 @@ def setUpClass(cls):
...
-1
-N
-5
-1
-2
-eb-app-maven
-    """
-        with tempfile.TemporaryDirectory() as temp:
-            runner = CliRunner()
-            runner.invoke(init_cmd, ["--output-dir", temp], input=user_input)
```
mndeveci, line 67: Why we removed this block, is this replaced with something else?

hnnasit (approved):

tests/integration/init/schemas/schemas_test_data_setup.py
```diff
@@ -43,28 +43,6 @@ def setUpClass(cls):
         setup_partner_schema_data("partner-registry", schemas_client)
         setup_schema_data_for_pagination("test-pagination", schemas_client)
         setup_non_partner_schema_data("other-schema", schemas_client)
-        # WHEN the user follows interactive init prompts
```
super132, line 46: Why removing the inputs below?

tests/integration/init/schemas/schemas_test_data_setup.py
mildaniel: Not sure what the use of running an init before the test class was or why this was introduced at all. This init flow is also very out-of-date and the tests pass without it.

tests/integration/init/schemas/schemas_test_data_setup.py
mildaniel: Answered here https://github.com/aws/aws-sam-cli/pull/6331#discussion_r1403662701

tests/integration/init/schemas/schemas_test_data_setup.py
hnnasit: We also don't run any assertions on the output of this command and the inputs were wrong for a while.

mndeveci: Yeap this looks weird to me as well. All setup stuff are done in the functions above, so this part shouldn't be required. Let's wait and see the canary runs

tests/integration/init/schemas/test_init_with_schemas_command.py
```diff
@@ -270,14 +294,17 @@ def test_init_interactive_with_event_bridge_app_aws_schemas_go(self):
             self.assertTrue(expected_output_folder.is_dir())
             self.assertTrue(Path(expected_output_folder, "HelloWorld", "schema").is_dir())
 
+    @pytest.mark.timeout(300)
```
mndeveci, line 297: Can we set a default in `pytest.ini` file so that we don't need to annotate every test method?

mndeveci (approved):
🥳

tests/integration/init/schemas/schemas_test_data_setup.py
mildaniel: These have all passed on Windows and Linux now

tests/integration/init/schemas/test_init_with_schemas_command.py
mildaniel: Added a global timeout

PR #6322, opened 2023-11-23
Add architecture validations for SAM build LambdaLayer

bentvelj:
- Add a validation check for LambdaLayer `BuildArchitecture` to ensure that it is a valid architecture and one of the `CompatibleArchitectures` listed.  If the `BuildArchitecture` is invalid, only a **warning** is displayed to inform the customer. Hard failing is not done for Makefile `BuildMethod` to preserve customer freedom in building with their own Makefile.

- Adds architecture validation for when `--use-container` is specified for building a LambdaLayer. This now gives the customer a much clearer error message regarding the architecture instead of a "failed to pull image" error. (@mndeveci's changes)

- Adds integration test to verify that a clear error message is displayed in the case of a customer attempting to build a function with an invalid architecture.

#### Which issue(s) does this change fix?
#6110 

#### Why is this change necessary?
- Customers may be confused why LambdaLayers with invalid `BuildArchitecture`s still build just fine. A warning message will help customers catch possible errors in their template.

- Regarding the `--use-container` option for building a layer, reporting an invalid architecture error addresses the actual root of the problem instead of showing a consequence of having an invalid architecture ("failed to pull image" error). This makes it easier for customers to resolve.

#### How does it address the issue?
- Add warning message if the LambdaLayer being built has a `BuildMethod` of `Makefile`. Architecture validation for other `BuildMethod`s are already done in LambdaBuilders, but nothing is done for `Makefile`.

#### What side effects does this change have?

N/A

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [x] Write/update integration tests
- [ ] Write/update functional tests if needed
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [x] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

Issue #6110, opened 2023-10-19
Bug: Builds are permitted with invalid Function `Architectures`, Layer `CompatibleArchitectures` and Parameter `AllowedValues`


davidjb:
### Description:

`sam build` doesn't attempt to validate values set for either Function `Architectures, LayerVersion `CompatibleArchitectures`, or Parameter `AllowedValues`. Despite the documentation stating that these properties have a set of valid values, the values aren't validated, leading to either successful builds, attempts to download invalid containers, or failure later in the build process. In the case of `AllowedValues`, the affect of this can lead to any sort of breakage, since a `Parameter` can be used anywhere.

This is related to https://github.com/aws/aws-sam-cli/issues/3244 and I wrote in an update to my comment on that issue.

### Steps to reproduce:

1. Create `template.yml`:

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Parameters:
  MyParam:
    Default: C
  AllowedValues: [A, B, C, D]

Resources:
  MyLayer:
    Type: AWS::Serverless::LayerVersion
    Properties:
      ContentUri: mylayer
      CompatibleRuntimes:
        - provided.al2
      CompatibleArchitectures:
        - fake
        - ${MyParam}
    Metadata:
      BuildMethod: makefile
      BuildArchitecture: fake

  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: myfunction
      Handler: app.lambdaHandler
      Runtime: python3.11
      Architectures:
        - fake
```

1. Shim out `mylayer/Makefile` with:
```make
build-MyLayer:
	echo 'Done'
```

1. Run builds and observe results, as per next section below.

### Observed result:

1. Run `sam build MyLayer` and observe success (e.g. no validation)
2. Run `sam build --use-container MyLayer` and observe attempt to download a container image from ECR with the invalid arch:
```
2023-10-19 17:13:09,717 | No config file found in this directory.
2023-10-19 17:13:09,718 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/demo/samconfig.toml'
2023-10-19 17:13:09,719 | Config file location: /demo/samconfig.toml
2023-10-19 17:13:09,719 | Config file '/demo/samconfig.toml' does not exist
2023-10-19 17:13:09,721 | Using SAM Template at /demo/template.yml
[...]

Issue #3747, opened 2022-03-16
PythonPipBuilder:ResolveDependencies errors on arm64, Incorrect Layer build behaviors


brysontyrrell:
Note: this is another open issue on this build error #1840 but there have been no updates since December and there are other behaviors I'm noticing that are not a part of that issue.

### Description:

I have a SAM template where I have defined the architecture at `arm64` for my functions and for my shared Lambda Layer.

I am working on an Apple Silicon MacBook Pro (arm64 host).

I am installing a private package from CodeArtifact that has a dependency of `jsonschema` that has a dependency of `pyrsistent`.

The Lambda Layer is configured to use the builder:

```yaml
  SharedLayer:
    Type: AWS::Serverless::LayerVersion
    Properties:
      LayerName: !Ref AWS::StackName
      ContentUri: ./src/layers/shared
      CompatibleArchitectures:
        - arm64
      CompatibleRuntimes:
        - python3.9
    Metadata:
      BuildMethod: python3.9
```

If I run `sam build` without a container and view the build files in `.aws-sam/build` I can see that it compiled for x86 by some of the binaries that are present: `pvectorc.cpython-39-x86_64-linux-gnu.so`. The layer's compatibility is clearly set for `arm64`.

I then tried to run the build in a container. The Layer pulled the `x86_64` image!

I added the package to a requirements file for the Lambda function to compare. It did pull the `arm64`, but then it hit the following error:

`Error: PythonPipBuilder:ResolveDependencies - {pyrsistent==0.18.1(wheel)}`

Running `sam build` without using a container yielded the same error.

Switching the Lambda Function to use `x86_64` results in a successful build.

I launched a container to test installing the package ( `docker run --rm -it build-python3.9:latest-arm64` ) and it did so without an issue. 

Here's a summary of the `sam build` issues:

* Lambda Layer - On an arm64 host builds are locked to x86_64.
* Lambda Layer - Builds are locked to the x86_64 build image even if compatibility is only arm64.
* Lambda Function - The build fails at the `pyrsistent` dependency (error above) using the arm64 image or building on an arm64 host. This is _NOT_ reproducible inside the arm64 image or on an arm64 host using `pip` directly.

_These issues have halted all attempts to migrate to Graviton for Lambda Functions with my group._

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS 12.3
2. `sam --version`: 1.40.1
3. AWS region: N/A

[...]

PR #6355, opened 2023-11-28
Warn for missing BuildArchitecture during Layer build

bentvelj:
**NOTE**: I've moved the check for `BuildArchitecture` ∈ `CompatibleArchitectures` from PR https://github.com/aws/aws-sam-cli/pull/6322 (related issue: #6110) into the Layer build, so that it's not exclusive to Makefile Layer builds.

To summarize:

- All Layer builds will warn if `BuildArchitecture` is missing
- All Layer builds will warn if `BuildArchitecture` is not in `CompatibleArchitectures`
- Only Makefile Layer builds will check and warn on **invalid** (not `x86_64` or `arm64`) `BuildArchitecture`. Other BuildMethods will verify this in LambdaBuilders already, and fail on bad architecture.

#3747 

Customer is confused as to why the build does not use their `CompatibleArchitectures` when they omit `BuildArchitecture`. It has already been decided to default to the case of `x86_64` in the case of missing `BuildArchitecture`, so we can warn the customer to add it if it is missing.

Adds a warning to Layer builds when `BuildArchitecture` is missing, and that `x86_64` is being used as a default.
Also warns if the `BuildArchitecture` specified is not in `CompatibleArchitectures`.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


mildaniel (approved):

sidhujus (approved):
LGTM!

PR #6382, opened 2023-12-04
chore: Update maintainers list

hnnasit:

N/A

Updates the maintainers list for tagging internal/external labels

^ same as above

No side effects

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


.github/workflows/pr-labeler.yml
```diff
@@ -25,9 +25,9 @@ jobs:
           script: |
             const maintainers = [
               'aws-sam-cli-bot',
-              'jfuss', 'hoffa', 'awood45', 'aahung', 'hawflau', 'mndeveci', 'ssenchenko',
-              'qingchm', 'moelasmar', 'xazhao', 'mildaniel', 'marekaiv', 'torresxb1',
```
mildaniel, line 29: Should we keep people that are still internal at Amazon? I wonder if we remove some of these people if we'll have large spikes in our responses dashboard if they were the last ones to respond to an issue

hnnasit: I suppose we could still keep the internal people, but this GH action would be run on new issues and PR's right? This would not impact any previous issues or issues.

mildaniel: That's right, these only run on new issues, good call.

Let's discuss who should stay on this list offline.

mildaniel (approved):

mndeveci (approved):
LGTM! I think we need to update following ones as well;
- https://github.com/aws/aws-lambda-builders/blob/develop/.github/workflows/pr-labeler.yml
- https://github.com/aws/aws-sam-cli-app-templates/blob/master/.github/workflows/pr-labeler.yml
- https://github.com/aws/aws-sam-build-images/blob/develop/.github/workflows/pr-labeler.yml

PR #6402, opened 2023-12-05
fix: Recursively call method on directory children

lucashuy:

None.

A previous fix did not check the contents of directories since the path list only contained the parent folders, not the individual files.

Recursively checks the children if the path is a folder.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/utils/tar.py
+++ b/samcli/lib/utils/tar.py
@@ -62,7 +62,7 @@ def create_tarball(
         tarballfile.close()
 
 
-def _validate_destinations_exists(tar_paths: List[Union[str, Path]]) -> bool:
+def _validate_destinations_exists(tar_paths: Union[List[Union[str, Path]], List[Path]]) -> bool:
     """
     Validates whether the destination of a symlink exists by resolving the link
     and checking the resolved path.
--- a/samcli/lib/utils/tar.py
+++ b/samcli/lib/utils/tar.py
@@ -79,7 +79,14 @@ def _validate_destinations_exists(tar_paths: Union[List[Union[str, Path]], List[
     """
     for file in tar_paths:
         file_path_obj = Path(file)
-        resolved_path = file_path_obj.resolve()
+
+        try:
+            resolved_path = file_path_obj.resolve()
+        except OSError:
+            # this exception will occur on Windows and will return
+            # a WinError 123
+            LOG.warning("Failed to determine the file on the system")
+            return False
 
         if file_path_obj.is_dir():
             # recursively call this method to validate the children are not symlinks to empty locations
--- a/samcli/lib/utils/tar.py
+++ b/samcli/lib/utils/tar.py
@@ -85,7 +85,7 @@ def _validate_destinations_exists(tar_paths: Union[List[Union[str, Path]], List[
         except OSError:
             # this exception will occur on Windows and will return
             # a WinError 123
-            LOG.warning("Failed to determine the file on the system")
+            LOG.warning(f"Failed to resolve file {file_path_obj} on the host machine")
             return False
 
         if file_path_obj.is_dir():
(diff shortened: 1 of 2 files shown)


samcli/lib/utils/tar.py
```diff
@@ -79,9 +79,22 @@ def _validate_destinations_exists(tar_paths: List[Union[str, Path]]) -> bool:
...
         file_path_obj = Path(file)
-        resolved_path = file_path_obj.resolve()
 
-        if file_path_obj.is_symlink() and not resolved_path.exists():
+        try:
+            resolved_path = file_path_obj.resolve()
+        except OSError:
+            # this exception will occur on Windows and will return
+            # a WinError 123
+            LOG.warning("Failed to determine the file on the system")
```
bentvelj, line 88: Would it be more helpful to Log the path that it failed to resolve, since this single function will validate a list of `tar_paths`.

```suggestion
            LOG.warning("Failed to resolve the path %s on the system", file_path_obj.abspath())
```

samcli/lib/utils/tar.py
```diff
@@ -79,9 +79,22 @@ def _validate_destinations_exists(tar_paths: List[Union[str, Path]]) -> bool:
...
+
+        if file_path_obj.is_dir():
+            # recursively call this method to validate the children are not symlinks to empty locations
+            children = list(file_path_obj.iterdir())
+            if not _validate_destinations_exists(children):
+                # exits early
+                return False
+        elif file_path_obj.is_symlink() and not resolved_path.exists():
             LOG.warning(f"Symlinked file {file_path_obj} -> {resolved_path} does not exist!")
             return False
```
bentvelj, line 99: There is currently no logic after this `if not _validate_destinations_exists(children)` check . Is there reason to believe there would be more logic in the future? If not we can reduce some nesting here, and get rid of the`else`
```suggestion
        if file_path_obj.is_dir():
            # recursively call this method to validate the children are not symlinks to empty locations
            children = list(file_path_obj.iterdir())
           return _validate_destinations_exists(children):

        if file_path_obj.is_symlink() and not resolved_path.exists():
            LOG.warning(f"Symlinked file {file_path_obj} -> {resolved_path} does not exist!")
            return False
```

tests/unit/lib/utils/test_tar.py
```diff
@@ -196,9 +196,32 @@ def test_validating_symlinked_tar_path(self, does_resolved_exist, path_mock):
...
+        mock_dir_object.is_symlink.return_value = False
+        mock_dir_object.is_dir.return_value = True
+        mock_dir_object.iterdir.return_value = ["mock_child"]
+        mock_dir_object.resolve = Mock()
+
+        path_mock.side_effect = [mock_dir_object, mock_child]
+
+        result = _validate_destinations_exists(["mock_folder"])
+
+        self.assertEqual(result, False)
```
bentvelj, line 227: Nit: Could use `self.assertFalse` here.

Also, shouldn't we have a positive case to test as well?

bentvelj (changes requested):

tests/unit/lib/utils/test_tar.py
lucashuy: Good point, will parameterize this test to test for a valid child file

samcli/lib/utils/tar.py
lucashuy: The reason for the nested `if` statement is to return early only if we found an invalid symlink. Returning true here will stop the loop and not evaluate the other tar paths.

hnnasit (approved):

samcli/lib/utils/tar.py
bentvelj: Ah yes the loop, my mistake. That makes sense

Issue #6173, opened 2023-10-31
Bug: Sam Local invoke has deliberate delay of 1s


andyfase:
### Description:
I typically use SAM Local within a mocked integration test framework passing in a number of events using pytest. I mock dependencies using DynamoDB local or Moto so that a full test suite can be run - typically this involves 40-50 test invokes of different events. The entire environment is run within a number of docker containers within a docker network - with SAM, DDB Local and Moto all sitting in containers connected over a docker network.

Running this on a linux environment (or seemingly any x86 environment) this will typically take ~2 seconds to run, with each invoke being measured in < 50ms. I can see this when run within Codebuild for example.

However when run locally on my M1 Mac, the tests takes over 1 second each to run, making it non optimal for local development.

SAM version
```
~/.../arch/src/web-api> sam --version
SAM CLI, version 1.100.0
```

Run using 
```
sam local start-lambda --docker-network lambda-local -t build/template.yaml --warm-containers EAGER --debug
```

Debug logs show

```
2023-10-31 08:43:37,119 | Found one Lambda function with name 'app'
2023-10-31 08:43:37,122 | Invoking index.handler (python3.10)
2023-10-31 08:43:37,124 | Resolving code path. Cwd=/path/to/arch/src/web-api/build,
CodeUri=/path/to/arch/src/web-api/build/app
2023-10-31 08:43:37,125 | Resolved absolute path to code is /path/to/arch/src/web-api/build/app
2023-10-31 08:43:37,147 | Reuse the created warm container for Lambda function 'app'
2023-10-31 08:43:37,164 | Lambda function 'app' is already running
2023-10-31 08:43:37,167 | Starting a timer for 90 seconds for function 'app'
START RequestId: a30c1bb5-375c-45a7-b4e7-d5a4bc0347e4 Version: $LATEST
END RequestId: a30c1bb5-375c-45a7-b4e7-d5a4bc0347e4
REPORT RequestId: a30c1bb5-375c-45a7-b4e7-d5a4bc0347e4	Duration: 24.22 ms	Billed Duration: 25 ms	Memory Size: 128 MB	Max Memory Used: 128 MB

2023-10-31 08:43:38 127.0.0.1 - - [31/Oct/2023 08:43:38] "POST /2015-03-31/functions/app/invocations HTTP/1.1" 200 -
```

[...]

PR #6418, opened 2023-12-07
fix: remove 1 sec delay on invoke

sidhujus:
#6173 

Currently there is an explicit 1 second sleep in the code to work around a race condition and ensure that everything that is printed to the cli is done in the correct order.

The original race condition occurs when the main thread finishes its execution and prints the flask response to stdout, while there are still some things from the docker container that need to be printed.

This can be reproduced currently by the following steps (after removing the sleep in dev version of sam)
```
1. sam init a hello world template (I did any version of node.js)
2. change the lambda code to just print something
3. samdev local start-lambda -t template.yaml
4. aws lambda invoke --endpoint http://localhost:3001 --function-name HelloWorldFunction out.txt (in a separate console)
5. repeat step 4 until the ordering of the output is incorrect (The final thing printed is not the following)
6. 2023-12-07 14:41:35 127.0.0.1 - - [07/Dec/2023 14:41:35] "POST /2015-03-31/functions/HelloWorldFunction/invocations HTTP/1.1" 200

```

using a threading.Event, that gets set only when the regex matches the pattern, to wait until it is safe to end the flask request

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -533,7 +533,7 @@ class Container:
 
         if isinstance(output_stream, io.TextIOWrapper):
             output_stream.buffer.write(output_str.encode("utf-8"))
-        if re.match(pattern, output_str) is not None and event is not None:
+        if re.match(pattern, output_str) is not None and event:
             event.set()
 
     # This method exists because otherwise when writing tests patching/mocking threading.Event breaks everything
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -400,7 +400,7 @@ class Container:
 
         # NOTE(jfuss): Adding a sleep after we get a response from the contianer but before we
         # we write the response to ensure the last thing written to stdout is the container response
-        #time.sleep(1)
+        # time.sleep(1)
         if isinstance(response, str):
             stdout.write_str(response)
         elif isinstance(response, bytes):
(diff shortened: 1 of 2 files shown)


samcli/local/docker/container.py
```diff
@@ -421,11 +427,11 @@ def wait_for_logs(
         if not self.is_created():
             raise RuntimeError("Container does not exist. Cannot get logs for this container")
 
-        real_container = self.docker_client.containers.get(self.id)
+        real_container = self.docker_client.containers.get(self.id)  #
```
hawflau, line 430: nit: extra #?

samcli/local/docker/container.py
```diff
@@ -398,21 +404,21 @@ def wait_for_result(self, full_path, event, stdout, stderr, start_timer=None):
         if timer:
             timer.cancel()
 
-        # NOTE(jfuss): Adding a sleep after we get a response from the contianer but before we
-        # we write the response to ensure the last thing written to stdout is the container response
-        time.sleep(1)
+        self._logs_thread_event.wait()
```
mndeveci, line 407: Does `.wait` has timeout option? If so we might want to set it in case it keeps waiting indefinetely when the message below never gets printed (I guess only possible if RIE/rapid crashes).

mndeveci (approved):

hawflau (approved):

Issue #6172, opened 2023-10-31
Bug: Changing parameters-override when running `sam sync` does not force update


GeeWee:
### Description:

When I run e.g. `sam sync --watch --parameter-overrides "MYVAR=foo1"` the environment variables do not update in my lambda.

However if I edit `template.yml`, the variables are updated

### Steps to reproduce:

Run
`sam sync --watch --parameter-overrides "MYVAR=foo1"`
Have the parameter passed to an environment variable.
See the env variable is `foo1`

Without touching anything, run:
`sam sync --watch --parameter-overrides "MYVAR=foo2"`
Run the lambda, see the value is still `foo1`

Now touch the file, and either re-run the command or leave it running, re-execute the lambda, and see the value is now properly changed.

Cannot provide debug logs as I hit the github comment length limit.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
{
  "version": "1.99.0",
  "system": {
    "python": "3.8.13",
    "os": "macOS-14.0-arm64-arm-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "Not available",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}
```

moelasmar:
Thanks @GeeWee for raising this issue. I will investigate it and update you.

moelasmar:
I verified this issue, the root cause of this problem is in SAM CLI logic, we compared the local template file, and the deployed stack template to decide if we should update the infra or not (check [this snippet](https://github.com/aws/aws-sam-cli/blob/111952592ae9d236c1abf28a8b75725aa1e5bcbd/samcli/lib/sync/infra_sync_executor.py#L219C2-L335)), and we did not include the stack parameters in the comparison, I need to check if this was decided or we just missed it.

For now to resolve this problem you can add `--no-skip-deploy-sync` the flag to the `sam sync` command to force updating the infrastructure.

PR #6426, opened 2023-12-08
fix: consider parameter overrides during sam sync

bentvelj:
#6172

`sam sync` did not consider parameter overrides when detecting changes. This leads to unexpected behaviour when the customer will set new parameter overrides when starting a sync, but their stack will behave with the old parameters.

When deciding whether or not to skip infra sync, we now ensure that `parameter_overrides` ⊆ `stack parameters`. Otherwise, sync infrastructure.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


sidhujus (approved):

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -502,6 +528,25 @@ def get_template(self, template_path: str) -> Optional[Dict]:
...
+            The name of the stack
+
+        Returns
+        -------
+            List of Dicts in the form { 'ParameterKey': Foo, 'ParameterValue': Bar }
+
+        """
+        return cast(
+            List[Dict[str, str]],
+            self._cfn_client.describe_stacks(StackName=stack_name).get("Stacks")[0].get("Parameters", []),
```
mndeveci, line 547: This might fail if stack haven't been deployed yet. For that reason I would rather check the size of the collection first instead of accessing the `[0]` element there.

tests/unit/lib/sync/test_infra_sync_executor.py
```diff
@@ -518,6 +518,61 @@ def test_auto_skip_infra_sync_exception(self, session_mock, get_template_mock, l
...
+        [
+            ({"Foo": "Bar"}, True),  # Subset
+            ({"Foo": "Bar", "Apples": "Oranges"}, True),  # Equal
+            ({"Foo": "Bar", "Apples": "Grapes"}, False),  # One pair matches the other does not
+            (
+                {"Foo": "Bar", "Apples": "Oranges", "Red": "Blue"},
+                False,
+            ),  # Overrides is a superset of current parameters
+        ]
+    )
```
mndeveci, line 532: Can we have a test case where parameters will not be in the same order?

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -272,6 +274,27 @@ def _auto_skip_infra_sync(
...
+
+        flat_current_stack_parameters = {}
+        for param in current_stack_parameters:
+            flat_current_stack_parameters[param["ParameterKey"]] = param["ParameterValue"]
+
+        # Check for parameter overrides being a subset of the current stack parameters
+        if not (parameter_overrides.items() <= flat_current_stack_parameters.items()):
+            LOG.debug("Detected changes between Parameter overrides and the current stack parameters.")
+            return False
+
```
mndeveci, line 297: This function (`_auto_skip_infra_sync`) seems a little bit bloated now, do you think moving this logic into its own function would be better?

mndeveci (changes requested):

samcli/lib/sync/infra_sync_executor.py
bentvelj: It's possible for the `describe_stacks` call to succeed for a given stack name but not have any elements in the `Stacks`? 

interesting, will do

samcli/lib/sync/infra_sync_executor.py
bentvelj: Will do

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -272,6 +274,16 @@ def _auto_skip_infra_sync(
             LOG.debug("The current template is different from the last deployed version, we will not skip infra sync")
             return False
 
+        try:
+            current_stack_params = self._get_stack_parameters(stack_name)
+        except ClientError as ex:
+            LOG.debug("Unable to fetch stack Parameters from stack with name %s", stack_name, exc_info=ex)
+            return False
```
mndeveci, line 281: I would move this into new function as well since this is purely related to that functionality

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -502,6 +517,61 @@ def get_template(self, template_path: str) -> Optional[Dict]:
...
+
+        Returns
+        -------
+            List of Dicts in the form { 'ParameterKey': Foo, 'ParameterValue': Bar }
+
+        """
+        stacks = self._cfn_client.describe_stacks(StackName=stack_name).get("Stacks")
+
+        if len(stacks) < 1:
+            LOG.error(
```
mndeveci, line 565:
```suggestion
            LOG.info(
```

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -272,6 +274,14 @@ def _auto_skip_infra_sync(
             LOG.debug("The current template is different from the last deployed version, we will not skip infra sync")
             return False
 
+        try:
+            if not self._param_overrides_subset_of_stack_params(stack_name, parameter_overrides):
+                LOG.debug("Detected changes between Parameter overrides and the current stack parameters.")
+                return False
+        except ClientError as ex:
+            LOG.debug("Unable to fetch stack Parameters from stack with name %s", stack_name, exc_info=ex)
+            return False
```
mndeveci, line 283: This `ClientError` is part of `_param_overrides_subset_of_stack_params` function. No need to bubble it up here, please handle it inside the function and then return the appropriate response back to here.

samcli/lib/sync/infra_sync_executor.py
```diff
@@ -221,6 +222,7 @@ def _auto_skip_infra_sync(
         packaged_template_path: str,
         built_template_path: str,
         stack_name: str,
+        parameter_overrides: dict = {},
```
mndeveci, line 225: Apologies that I didn't see this earlier but python has weird behavior with default values of collection type parameters. (see: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)

Either remove the default value here, or set the default to `None` and instantiate inside the function with `parameter_overrides = parameter_overrides or {}`.

And please add this parameter to the function documentation below.

mndeveci (changes requested):

samcli/lib/sync/infra_sync_executor.py
bentvelj: Good point, forgot to add it to the docstring. Also was not aware of that behaviour, good to know

samcli/lib/sync/infra_sync_executor.py
bentvelj: Fair, I'll move all error handling and logging into the parameter function

mndeveci (approved):

PR #6435, opened 2023-12-11
Revert "chore: Upgrade Mac installer to python3.11"

mildaniel:
Reverts aws/aws-sam-cli#6424

Revert due to issues discovered generating artifacts on arm64


hnnasit (approved):

.github/workflows/validate_pyinstaller.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: "3.7"
```
hnnasit, line 54: I think we can bump this to 3.8 right?

mildaniel: We can, not sure it matters too much here. I'll raise a quick follow-up PR to update that.

mndeveci (approved):

PR #6441, opened 2023-12-12
fix: Run schema generation script as module instead

lucashuy:

None.

The schema validation job was not working for a while for two reasons:
1. Running the script with `python module/script.py` did not work properly
2. The Python command was nested in a `diff` command, causing the failure to not be propagated upwards

1. Change the command to use the `-m` flag instead
2. Changed the nested command to execute the `cat` command iff the Python command runs successfully, causing the `diff` check to fail

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mndeveci (approved):
I wonder if we can comment the warning message in the PR which says "please run `make schema`" so that it will be visible to the contributor. 🤔

hnnasit (approved):

lucashuy:
> I wonder if we can comment the warning message in the PR which says "please run make schema" so that it will be visible to the contributor. 🤔

Thats something we could do since it should be straightforward enough, but we might also want to consider expanding this to also respond if `make pr` fails, or any of the other jobs?

Going to merge this in the mean time, but something we can consider adding later.

Issue #6310, opened 2023-11-21
Bug: sam deploy - ValueError


kaihendry:
### Description:
Unable to deploy https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/examples/sam

### Steps to reproduce:
Checkout https://github.com/aws-powertools/powertools-lambda-typescript/tree/main/examples/sam

    sam build && sam deploy --guided

### Observed result:

```
❯ sam deploy --guided --debug
2023-11-21 18:05:55,799 | No config file found in this directory.
2023-11-21 18:05:55,801 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/Users/hendry/sam/powertools-example/samconfig.toml'
2023-11-21 18:05:55,801 | Config file location: /Users/hendry/sam/powertools-example/samconfig.toml
2023-11-21 18:05:55,802 | Config file '/Users/hendry/sam/powertools-example/samconfig.toml' does not exist
2023-11-21 18:05:55,812 | Using SAM Template at /Users/hendry/sam/powertools-example/.aws-sam/build/template.yaml
2023-11-21 18:05:55,838 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/Users/hendry/sam/powertools-example/samconfig.toml'
2023-11-21 18:05:55,848 | Using config file: samconfig.toml, config environment: default
2023-11-21 18:05:55,849 | Expand command line arguments to:
2023-11-21 18:05:55,849 | --guided --template_file=/Users/hendry/sam/powertools-example/.aws-sam/build/template.yaml --fail_on_empty_changeset --on_failure=ROLLBACK --stack_name=sam-app
2023-11-21 18:05:55,972 | Sam customer defined id is more priority than other IDs. Customer defined id for resource getAllItemsFunction is getAllItemsFunction
2023-11-21 18:05:55,973 | Sam customer defined id is more priority than other IDs. Customer defined id for resource getByIdFunction is getByIdFunction
2023-11-21 18:05:55,973 | Sam customer defined id is more priority than other IDs. Customer defined id for resource getUuidFunction is getUuidFunction
2023-11-21 18:05:55,974 | There is no customer defined id or cdk path defined for resource uuidApiUrlParameter, so we will use the resource logical id as the resource id
2023-11-21 18:05:55,974 | Sam customer defined id is more priority than other IDs. Customer defined id for resource putItemFunction is putItemFunction
[...]

PR #6471, opened 2023-12-19
fix: skip checks on non serverless api resources

sidhujus:
#6310 

if an api event has a ref to a non serverless api resource we still perform checks as if its a serverless api resource.

skips checking for auth on `Aws::ApiGateway::RestApi` resources

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/deploy/auth_utils.py
```diff
@@ -101,6 +101,11 @@ def _auth_id(resources_dict, event_properties, identifier):
     """
     resource_name = event_properties.get(identifier, "")
     api_resource = resources_dict.get(resource_name, {})
+
+    # Auth does not apply to ApiGateway::RestApi resources so return true and continue
+    if api_resource and api_resource["Type"] == "AWS::ApiGateway::RestApi":
```
hawflau, line 106: Does this apply to [`AWS::ApiGatewayV2::Api`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html) as well?

sidhujus: Ah yea good catch. I think it should since if the event refers to an httpApi then it would have to be this

samcli/commands/deploy/auth_utils.py
```diff
@@ -101,6 +101,13 @@ def _auth_id(resources_dict, event_properties, identifier):
     """
     resource_name = event_properties.get(identifier, "")
     api_resource = resources_dict.get(resource_name, {})
+
+    # Auth does not apply to ApiGateway::RestApi resources so return true and continue
+    if api_resource and (
+        api_resource["Type"] == "AWS::ApiGateway::RestApi" or api_resource["Type"] == "AWS::ApiGatewayV2::Api"
```
mildaniel, line 107: Let's use `.get()` instead to avoid accidental `AttributeErrors` in the case of a malformed template.
```suggestion
        api_resource.get("Type", "") == "AWS::ApiGateway::RestApi" or api_resource.get("Type", "") == "AWS::ApiGatewayV2::Api"
```

samcli/commands/deploy/auth_utils.py
```diff
@@ -101,6 +101,13 @@ def _auth_id(resources_dict, event_properties, identifier):
     """
     resource_name = event_properties.get(identifier, "")
     api_resource = resources_dict.get(resource_name, {})
+
+    # Auth does not apply to ApiGateway::RestApi resources so return true and continue
+    if api_resource and (
```
mildaniel, line 106: We have some constants here that you may be able to re-use https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/resources.py

mildaniel (commented):
Couple minor comments, looks good overall!

tests/unit/commands/deploy/test_auth_utils.py
```diff
@@ -88,6 +88,20 @@ def test_auth_per_resource_defined_on_api_resource(self):
         _auth_per_resource = auth_per_resource([Stack("", "", "", {}, self.template_dict)])
         self.assertEqual(_auth_per_resource, [("HelloWorldFunction", True)])
 
+    def test_auth_per_resource_on_non_serverless_api(self):
+        self.template_dict["Resources"]["HelloWorldApi"] = OrderedDict(
+            [
+                ("Type", "AWS::ApiGateway::RestApi"),
```
hawflau, line 94: Can we also add the test case for `AWS::ApiGatewayV2::Api`? (to make sure the branch is covered)

mildaniel (approved):

hawflau (approved):
LGTM!

Issue #6260, opened 2023-11-14
Bug: out-of-memory local `Function` invocations are killed silently by Docker


davidjb:
### Description:

Invoking a function with `sam local invoke` will silently fail when memory allocation inside the function exceeds its allocated `MemorySize`. Behind the scenes, the Docker OOM Killer has terminated the processes and container, but SAM doesn't report this occurred, even with debugging enabled.

The logging in the output below produces an accounting for the `Max Memory Used` by a function invocation, but the number isn't indicative of what the function actually consumed, appearing to always match the `MemorySize` set in the SAM template. In other words, when changing this to 10240 MB, the `Memory Size` and `Max Memory Used` values both increase to 10240 MB, even with the same code running, or - for good measure - changing the example `main.go` below to just creating a slice of 1 element.  

SAM knows the process has returned an invalid response (`Failed to deserialize response from RIE, returning the raw response as is`) but that's the extent of what gets reported.

### Steps to reproduce:

1. Create a `template.yml` like so:

```yaml
AWSTemplateFormatVersion: "2010-09-09"
Transform: AWS::Serverless-2016-10-31
Resources:
  OutOfMemoryFunction:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: go1.x
    Properties:
      CodeUri: functions/outOfMemory
      Handler: bootstrap
      Runtime: provided.al2
      Architectures: [arm64]
      MemorySize: 128 # Switch to 1024 for success
      Timeout: 120
```
2. Create `functions/outOfMemory/main.go`:
```go
package main

import (
	"fmt"
	"runtime"

	"github.com/aws/aws-lambda-go/lambda"
)

func HandleRequest() (*string, error) {
    fmt.Println("Starting")
	m := runtime.MemStats{}
	data := make([]int, 10000000000)
	fmt.Println(len(data), "ints")
	runtime.ReadMemStats(&m)
	fmt.Println(m.TotalAlloc/1024/1024, "MB memory used")
	result := "success"
	return &result, nil
}

func main() {
	lambda.Start(HandleRequest)
}
```
3. Create `functions/outOfMemory/go.mod`:
```go
module outOfMemory
go 1.21.3
require github.com/aws/aws-lambda-go v1.41.0
```

### Observed result:

```sh
$ sam local invoke OutOfMemoryFunction --debug
2023-11-14 14:11:25,500 | No config file found in this directory.
[...]

PR #6476, opened 2023-12-20
fix: Explicitly exit when container is out-of-memory

mildaniel:

https://github.com/aws/aws-sam-cli/issues/6260

When an invoke container exits because it's out of memory, SAM CLI doesn't communicate this to customers and it's unclear why the invocation failed.

Explicitly note the exit reason is out-of-memory and exit with return code 1 instead of 0.

Existing invocations that fail due to out-of-memory will return exit code 1.

#### Mandatory Checklist

- [] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container_analyzer.py
+++ b/samcli/local/docker/container_analyzer.py
@@ -1,0 +1,48 @@
+"""
+Class for handling the analysis and inspection of Docker containers
+"""
+import logging
+from dataclasses import dataclass
+
+from samcli.local.docker.container import Container
+from samcli.local.docker.manager import ContainerManager
+
+LOG = logging.getLogger(__name__)
+
+DEFAULT_OUT_OF_MEMORY = False
+
+
+@dataclass
+class ContainerState:
+    out_of_memory: bool
+
+
+class ContainerAnalyzer:
+    def __init__(self, container_manager: ContainerManager, container: Container):
+        self.container_manager = container_manager
+        self.container = container
+
+    def inspect(self) -> ContainerState:
+        """
+        Inspect the state of a container by calling the "inspect()" API that Docker provides.
+        Extract relevant information into a ContainerState object.
+
+        Returns
+        -------
+        ContainerState:
+            Returns a ContainerState object with relevant container data
+        """
+        if not self.container.id:
+            LOG.debug("Container ID not defined, unable to fetch container state")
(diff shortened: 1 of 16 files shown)


samcli/local/docker/container_analyzer.py
```diff
@@ -0,0 +1,48 @@
...
+        ContainerState:
+            Returns a ContainerState object with relevant container data
+        """
+        if not self.container.id:
+            LOG.debug("Container ID not defined, unable to fetch container state")
+            return ContainerState(DEFAULT_OUT_OF_MEMORY)
+
+        state = self.container_manager.inspect(self.container.id)
+
+        if isinstance(state, bool):
```
mildaniel, line 41: This check is admittedly a little weird, but mypy won't like it if we don't do this, and the reason we return a bool here is to follow the legacy pattern in `container_manager`.

samcli/local/docker/manager.py
```diff
@@ -193,6 +194,25 @@ def has_image(self, image_name):
...
+
+        Returns
+        -------
+        Union[bool, dict]
+            Container inspection state if successful, False otherwise
+        """
+        try:
+            return cast(dict, self.docker_client.api.inspect_container(container))
+        except (docker.errors.APIError, docker.errors.NullResource):
+            return False
```
mndeveci, line 214: Let's add some debug logs if we need to investigate an issue later on

samcli/local/lambdafn/runtime.py
```diff
@@ -223,9 +226,30 @@ def _on_invoke_done(self, container):
...
+        Raises
+        -------
+        ContainerFailureError
+            If the exit reason is due to out-of-memory, return exit code 1
+
+        """
+        container_analyzer = ContainerAnalyzer(self._container_manager, container)
+        exit_state = container_analyzer.inspect()
+        if exit_state.out_of_memory:
+            raise ContainerFailureError("Container invocation failed due to maximum memory usage")
```
mndeveci, line 251: Just thinking out loud, do we need to flush any outputs before ending the execution here?

mildaniel: We should be okay here, the RIE should be done executing and all logs streamed at this point.

jysheng123 (approved):

mndeveci (approved):

Issue #6444, opened 2023-12-13
Bug: sam build is warning about missing BuildArchitecture in a layer when I don't have one.


mew1033:
### Description:
I have a template that defines a `AWS::Serverless::Function` resource. It has a `Layers` section:
```yaml
      Layers:
        - !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:017000801446:layer:AWSLambdaPowertoolsPythonV2-Arm64:45
```
When I `sam build` the template, it gives me the following warning 3 times:

`WARNING: No BuildArchitecture specifed in Layer `None` Metadata. Defaulting to x86_64.`

Removing the layers section makes the warning go away.

I don't think the warning causes any issues, but it doesn't seem like it should be there.

### Observed result:
2023-12-13 15:26:27,691 | WARNING: No BuildArchitecture specifed in Layer `None` Metadata. Defaulting to x86_64.                                                                                                                                                                                  

### Expected result:
No warning

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Ubuntu 22.04 (in WSL)
2. `sam --version`: 1.105.0
3. AWS region: 

```
{
  "version": "1.105.0",
  "system": {
    "python": "3.11.3",
    "os": "Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.7",
    "aws_cdk": "Not available",
    "terraform": "1.6.5"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}
```

hnnasit:
Hi @mew1033, thanks for reporting the issue, marking this as a UX bug.

jojo786:
Also getting this warning when using LambdaAdapterLayer

jim-meyer:
[...]

PR #6508, opened 2024-01-03
Fix Erroneous "No BuildArchitecture specified in Layer" Warning

bentvelj:
#6444 

Customers were reporting the warning of `No BuildArchitecture specified in Layer` when building Lambda Functions which referenced existing Lambda Layers. The error message also incorrectly appears while building Lambda Layers without a BuildMethod at all.

Moved the logic which checks for missing BuildArchitecture from the LayerVersion constructor to the app builder `build_layer` function, so that warnings are not printed for layers that are not built.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


sidhujus (approved):
LGTM!

samcli/lib/build/app_builder.py
```diff
@@ -502,6 +502,13 @@ def _build_layer(
         str
             Path to the location where built artifacts are available
         """
+
+        if layer_metadata and "BuildArchitecture" not in layer_metadata:
+            LOG.warning(
+                "WARNING: No BuildArchitecture specified in Layer `%s`" + " Metadata. Defaulting to x86_64.",
+                layer_name,
+            )
```
mndeveci, line 510: I think we want to print this warning message if layer has `arm64` architecture as [`CompatibleArchitectures`](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-layerversion.html#sam-layerversion-compatiblearchitectures) but it doesn't have `BuildArchitecture` property

bentvelj: There is [another warning ](https://github.com/aws/aws-sam-cli/blob/be0eb703c844db2498fbb6919b0c8c20997a4943/samcli/lib/providers/provider.py#L256-L261)in the LayerVersion constructor that should take care of this.

It would output
```
WARNING: Layer `Layer` has BuildArchitecture `x86_64`, which is not listed in CompatibleArchitectures. 
```

bentvelj: This _may_ be slightly confusing if the customer doesn't know that x86 is the default? Maybe we can adjust the messaging here

mndeveci: My concern with this one is; if I have a layer which is supposed to be built with `x86` I will keep getting these warning messages all the time, which can be annoying.

bentvelj: Right. As discussed, going to restructure according to these rules:

There are two cases where we'd like to warn the customer
1. Compatible Architectures is only x86 (or not present) but Build Architecture is arm64
2. Build Architecture is x86 (or not present) but Compatible Architectures is only arm64

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1736,18 +1791,14 @@ def test_build_fails_with_missing_metadata(self, runtime, use_container, layer_i
...
         self.assertFalse(self.default_build_dir.joinpath(layer_identifier).exists())
 
-    @parameterized.expand([("python3.7", False, "LayerOne"), ("python3.7", "use_container", "LayerOne")])
-    def test_build_with_missing_buildarchitecture(self, runtime, use_container, layer_identifier):
-        if use_container and (SKIP_DOCKER_TESTS or SKIP_DOCKER_BUILD):
-            self.skipTest(SKIP_DOCKER_MESSAGE)
-
-        overrides = {"LayerBuildMethod": runtime, "LayerContentUri": "PyLayer"}
+    def test_function_build_succeeds_with_referenced_layer(self):
+        overrides = {"Runtime": "python3.8"}
```
mndeveci, line 1795: Why did we replace the parameterized test (which test with and without container build) with constant one (which only test without container)?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1691,7 +1694,59 @@ def test_build_layer_with_architecture_not_compatible(self, build_method, use_co
...
         self.assertIn(
-            f"Layer `{layer_identifier}` has BuildArchitecture `x86_64`, which is not listed in CompatibleArchitectures.",
+            f"Layer '{layer_identifier}' has BuildArchitecture x86_64, which is not listed in CompatibleArchitectures",
+            str(command_result.stderr.decode("utf-8")),
+        )
+        # Build should still succeed
+        self.assertEqual(command_result.process.returncode, 0)
+
+    @parameterized.expand([("python3.8", False), ("python3.8", "use_container")])
+    def test_build_arch_no_compatible_arch(self, runtime, use_container):
```
mndeveci, line 1704: nit: AFAIU, these tests are not validating the case where there is no CompatibleArchitectures or no BuildArchitecture present, can we add some tests for them (could be unit tests)?

tests/integration/buildcmd/test_build_cmd.py
bentvelj: Ah, I didn't leave the parameterization out on purpose, I just forgot to re-add it after writing the new test. Will fix.

tests/integration/buildcmd/test_build_cmd.py
bentvelj: I believe the[ LayerOne build tests ](https://github.com/aws/aws-sam-cli/blob/1029dcb09dfb5b2a2c783343efaf31a695bf648f/tests/integration/buildcmd/test_build_cmd.py#L1606)should cover that?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1736,18 +1791,19 @@ def test_build_fails_with_missing_metadata(self, runtime, use_container, layer_i
         self.assertEqual(command_result.process.returncode, 1)
         self.assertFalse(self.default_build_dir.joinpath(layer_identifier).exists())
 
-    @parameterized.expand([("python3.7", False, "LayerOne"), ("python3.7", "use_container", "LayerOne")])
-    def test_build_with_missing_buildarchitecture(self, runtime, use_container, layer_identifier):
+    @parameterized.expand([False, "use_container"])
+    def test_function_build_succeeds_with_referenced_layer(self, use_container):
```
mndeveci, line 1795: Looks like this test is changed to validate something else (it was building the layer before but now it is building the function)?

bentvelj: This is because we elected not to warn on JUST a missing BuildArchitecture, which is what this test was doing before. I think the diff is misleading here because I wasn't changing a test, but rather replacing it entirely.

mndeveci (approved):

Issue #6369, opened 2023-11-30
Bug: sam invoke local throw error 'utf-8' codec can't decode byte 0xff


yun-sangho:
### Description:

I have the lambda function that returns raw binary image not encoded using utf-8.
I used to test it using sam well, but after I update sam when I invoked the lambda using `sam invoke local`, I got the error that not `'utf-8' codec can't decode byte 0xff`.

Then I checked the pr #6034, because of it the binary result of a lambda always decoded by utf-8 when invoke local and probably it cause the above decode error.

Do I have to return only  utf-8 encoded binary from a lambda? or is it a bug?

### Observed result:

```
Error: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

### Expected result:

sam invoke local work without the error

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: mac 14.1.1 (23B81)
2. `sam --version`: 1.103.0
3. AWS region: ap-northeast-2

```
{
  "version": "1.103.0",
  "system": {
    "python": "3.12.0",
    "os": "macOS-14.1.1-arm64-arm-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.7",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}

```

```
Error: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
Traceback:
  File "/opt/homebrew/opt/python-click/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/opt/python-click/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/opt/python-click/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #6509, opened 2024-01-04
Fix: sam invoke local throw error 'utf-8' codec can't decode byte 0xff

sidhujus:
#6369

The code was previously refactored to use Strings to handle the IO related to sam local invoke and sam local start-lambda this fixed some issues where certain strings werent displayed correctly, but broke the functionality to support lambdas returning images

adds an extra BytesIO object to the stream writer and checks if it was written to in order to determine if it should return strings or the bytes

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/utils/stream_writer.py
```diff
@@ -17,12 +18,35 @@ def __init__(self, stream: TextIO, auto_flush: bool = False):
...
+        if self._stream_bytes:
+            if isinstance(self._stream_bytes, TextIOWrapper):
+                self._stream_bytes.buffer.write(output)
+                if self._auto_flush:
+                    self._stream_bytes.flush()
+
+            elif isinstance(self._stream_bytes, BytesIO):
+                self._stream_bytes.write(output)
+                if self._auto_flush:
+                    self._stream_bytes.flush()
```
mndeveci, line 48: Can we use early return at least from the first `if` statement?
```suggestion
        if not self._stream_bytes:
            return
        if isinstance(self._stream_bytes, TextIOWrapper):
            self._stream_bytes.buffer.write(output)
            if self._auto_flush:
                self._stream_bytes.flush()

        elif isinstance(self._stream_bytes, BytesIO):
            self._stream_bytes.write(output)
            if self._auto_flush:
                self._stream_bytes.flush()
```

samcli/local/docker/container.py
```diff
@@ -400,13 +403,18 @@ def wait_for_result(self, full_path, event, stdout, stderr, start_timer=None):
...
         if timer:
             timer.cancel()
 
         self._logs_thread_event.wait(timeout=1)
         if isinstance(response, str):
             stdout.write_str(response)
+        elif isinstance(response, bytes) and is_image:
+            # b64_image = base64.b64encode(response)
+            # # Have to decode into utf-8 so we can write
+            # stdout.write_str(b64_image.decode("utf-8"))
```
mndeveci, line 416: Is line 414 and 416 extra (left from debugging maybe)?

samcli/local/services/base_local_service.py
```diff
@@ -103,7 +105,9 @@ def get_lambda_output(stdout_stream: io.StringIO) -> Tuple[str, bool]:
         bool
             If the response is an error/exception from the container
         """
-        lambda_response = stdout_stream.getvalue()
+        lambda_response: Union[str, bytes] = stdout_stream_str.getvalue()
+        if stdout_stream_bytes and not lambda_response:
```
mndeveci, line 109: Is `stdout_stream_bytes` the raw value of `lambda_response` (when `lambda_response` is available)?

sidhujus: Yea. Currently we write the response of lambdas to `stdout_stream_str` then it gets read and returned from here, and if the response isn't json deserializeable then we just decode the raw response to utf-8 put it in the `stdout_stream_str` instead. I had to add a second stream (`stdout_stream_bytes`) since the StringIO object can't accept bytes and images can't be decoded to utf-8 without first changing them to base64, which isn't what lambda does (and isn't how SAM previously handled it).

samcli/lib/utils/stream_writer.py
```diff
@@ -1,28 +1,55 @@
...
+        stream_bytes io.TextIO | io.BytesIO
+            Stream to wrap if bytes are being written
         auto_flush bool
             Whether to autoflush the stream upon writing
         """
         self._stream = stream
+        self._stream_bytes = stream_bytes
         self._auto_flush = auto_flush
+        if isinstance(stream, TextIOWrapper):
+            self._stream_bytes = stream
```
bentvelj, line 26: Nit:
```suggestion
        self._stream_bytes = stream if isinstance(stream, TextIOWrapper) else stream_bytes
        self._auto_flush = auto_flush
```
Maybe a one liner is a bit clearer here rather than assigning twice?

samcli/lib/utils/stream_writer.py
```diff
@@ -1,28 +1,55 @@
...
+            return
+        if isinstance(self._stream_bytes, TextIOWrapper):
+            self._stream_bytes.buffer.write(output)
+            if self._auto_flush:
+                self._stream_bytes.flush()
+
+        elif isinstance(self._stream_bytes, BytesIO):
+            self._stream_bytes.write(output)
+            if self._auto_flush:
+                self._stream_bytes.flush()
```
bentvelj, line 51:
```suggestion
        if isinstance(self._stream_bytes, TextIOWrapper):
            self._stream_bytes.buffer.write(output)
        elif isinstance(self._stream_bytes, BytesIO):
            self._stream_bytes.write(output)
            
        if self._auto_flush:
            self._stream_bytes.flush()
```
Pull out the flush check so we only have to write it once?

bentvelj: If I understand correctly, since we know `stream_bytes` will be either `TextIO` or `BytesIO`, it should always fall under one of the cases, and thus that auto_flush check will always run?

sidhujus: yea i though that also but the linter doesn't seem to recognize that since it gives me this error `error: "object" has no attribute "flush"  [attr-defined]` which I guess is happening because it just assumes its a generic object because of the Union type

mndeveci:
Can we add some integration tests to validate the logic?

samcli/lib/utils/stream_writer.py
bentvelj: Interesting, all good then

bentvelj (approved):

tests/integration/testdata/invoke/main.py
```diff
@@ -60,3 +60,10 @@ def execute_git(event, context):
 
 def no_response(event, context):
     print("lambda called")
+
+
+def image_handler(event, context):
+    f = open("image-for-lambda.png", "rb")
+    image_bytes = f.read()
+    f.close()
```
mndeveci, line 68: you can open a file with context which should close it automatically
```suggestion
    with open("image-for-lambda.png", "rb") as f:
        image_bytes = f.read()
```

tests/integration/testdata/invoke/template-return-image.yaml
```diff
@@ -0,0 +1,12 @@
+AWSTemplateFormatVersion : '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: A hello world application.
+
+Resources:
+  GetImageFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: main.image_handler
+      Runtime: python3.9
```
mndeveci, line 10: let's use a newer python version
```suggestion
      Runtime: python3.12
```

mndeveci (approved):
Thanks for addressing previous comments, left some small comments with suggestions, rest looks good to me.

emanuelverardi:
Hello! 👋 I've just tested the latest SAM CLI tag: SAM CLI, version 1.107.0.dev202401180901, and unfortunately, I'm still encountering an issue when attempting to upload images to my Lambda function locally. Strangely, this process works seamlessly using API Gateway and Lambda on AWS; the problem appears to be specific to the local SAM environment.

The error I'm encountering is a UnicodeDecodeError during HTTP request processing: 'utf-8' codec can't decode byte 0xff at position 263, signaling an invalid start byte.

I've confirmed the issue persists in the current branch. Here are the details:

sam --version
SAM CLI, version 1.107.0.dev202401180901

sidhujus:
Hi @emanuelverardi thanks for letting us know about the issue. Could you let us know which command you ran and the steps to reproduce the error?

sidhujus:
Hi @emanuelverardi Thanks for letting us know about your issue. Could you provide some more information on what commands you ran and any steps we can take to reproduce your issue? Thanks!

emanuelverardi:
> Hi @emanuelverardi Thanks for letting us know about your issue. Could you provide some more information on what commands you ran and any steps we can take to reproduce your issue? Thanks!

Hello @sidhujus 

I'm currently working on implementing file uploads in my production environment using AWS API Gateway and AWS Lambda. In my development environment, I'm using SAM CLI, version 1.107.0.dev202401180901.

I've integrated a JavaScript library called BlueImp File Upload (https://github.com/blueimp/jQuery-File-Upload) to handle the data posting to my Lambda function.

Although the Lambda function is being invoked successfully, I'm encountering an issue specifically when uploading images. The error log is as follows:

```
2024-01-18 17:15:45 127.0.0.1 - - [18/Jan/2024 17:15:45] "OPTIONS /upload HTTP/1.1" 200 -
UnicodeDecodeError while processing HTTP request: 'utf-8' codec can't decode byte 0xff in position 263: invalid start byte
2024-01-18 17:15:45 127.0.0.1 - - [18/Jan/2024 17:15:45] "POST /upload HTTP/1.1" 502 -
```

I'd like to highlight that the function works seamlessly on AWS, but when running it with SAM, I encounter this byte 0xff error.

Any insights or assistance on resolving this issue would be greatly appreciated. Thanks!
Thanks.

Issue #6442, opened 2023-12-12
Bug: `sam local` includes extraneous `version` attribute in `ApiGatewayLambdaEvent` (incompatible with aws-lambda-rust-runtime)


ramosbugs:
### Description:

#1942 added support for API Gateway HTTP (aka v2) APIs to `sam local`. This change also modified the API Gateway REST API (aka v1) request payload to include a `"version": "1.0"` attribute that is **not** part of the payload that API Gateway includes in production.

This extraneous `version` attribute breaks compatibility with runtimes like the [aws-lambda-rust-runtime](https://github.com/awslabs/aws-lambda-rust-runtime) that [reject payloads that include unexpected fields](https://github.com/awslabs/aws-lambda-rust-runtime/blob/d513b13b4c48122602c0690f55147607f3bcc0da/lambda-events/src/event/apigw/mod.rs#L16).

As a consequence, the current versions of aws-sam-cli and aws-lambda-rust-runtime are mutually incompatible with one another for local testing. This is quite frustrating.

Since `sam local` is supposed to behave identically (as much as possible) to the corresponding AWS services, I feel the `version` attribute should be removed from `ApiGatewayLambdaEvent`:
https://github.com/aws/aws-sam-cli/blob/aa132f863a6580ed2e279fabfc44f7fc6af7f2f1/samcli/local/events/api_event.py#L208

### Steps to reproduce:

1. Define an `AWS::Serverless::Api` resource and corresponding Lambda proxy integration
2. Run `sam local`

### Observed result:

Lambda receives a request payload that includes `"version": "1.0"` (not present in API Gateway payloads).

### Expected result:

Lambda should receive a request payload without `version` that mirrors the payload API Gateway sends in production.

hnnasit:
Hi @ramosbugs, thanks for reporting the issue. I see your PR in `aws-lambda-rust-runtime` got merged. Does that unblock you? As for the `version` field, I do see it in the [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) for both the versions of API Gateway.

ramosbugs:
> Hi @ramosbugs, thanks for reporting the issue. I see your PR in `aws-lambda-rust-runtime` got merged. Does that unblock you?

Yup, but I think this is still a bug in aws-sam-cli worth fixing.

[...]

PR #6532, opened 2024-01-05
fix: remove version from rest api payload

sidhujus:
#6442

Currently the v1 payload is used for both http API's that specify v1 and for REST api's, which is technically wrong since the REST payload does not contain the `version` field (but is identical in every other aspect that I checked)

adds an extra parameter to the init method that specifies which type of api is being used.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


sidhujus:
https://github.com/aws/aws-sam-cli/pull/6532/files#diff-1cea545ee08253b7df8b2edd1e1809613a5e05666c2e7609e89aa036920651f6R372

I couldn't find anything related to where this is used or whether by v1 we are referring to REST apis or v1 HTTP api payloads

samcli/local/events/api_event.py
```diff
@@ -205,7 +208,9 @@ def __init__(
         if not isinstance(stage_variables, dict) and stage_variables is not None:
             raise TypeError("'stage_variables' must be of type dict or None")
 
-        self.version = "1.0"
+        # v1 payloads and rest api payloads are identical save for the version field
+        if api_type == Route.HTTP:
+            self.version = "1.0"
```
mndeveci, line 213: It looks like storing `self.version` is redundant as it can only have constant value which is not used by the change below as well. Can you check if it has other usages from other places, if not we may remove it completely?

samcli/local/events/api_event.py
```diff
@@ -169,6 +171,7 @@ def __init__(
         stage_variables=None,
         path=None,
         is_base_64_encoded=False,
+        api_type=Route.API,
```
mndeveci, line 174: Can you add this new option into the method documentation below?

mndeveci (approved):

Issue #6428, opened 2023-12-11
Bug: `sam local` improperly rejects OpenAPI specs with empty `security` attributes


ramosbugs:
### Description:

The OpenAPI spec [explicitly](https://swagger.io/specification/) defines `security: [{}]` to mean that an endpoint has security optional (i.e., is unauthenticated):
> To make security optional, an empty security requirement (`{}`) can be included in the array.

Amazon API Gateway follows the spec and accepts this syntax, at least for REST APIs. However, #5005 changed the behavior of `sam local` to reject OpenAPI specs that have `{}` entries under `security`:
https://github.com/aws/aws-sam-cli/blob/ec451f2be7e4f9a8290f32f4c7a51ff19f757fce/samcli/commands/local/lib/swagger/parser.py#L382-L386

### Steps to reproduce:

Add `security: [{}]` to any operation in an OpenAPI spec.

### Observed result:

`sam local` fails with the following error:
```
Error: Invalid security definition found, authorizers for path='/foo/bar' method='post' must contain an authorizer
```

### Expected result:

`sam local` should treat the operation as if no `security` attribute were defined (i.e., no authorizer).

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: MacOS
2. `sam --version`: `SAM CLI, version 1.104.0`
3. AWS region: N/A

```
{
  "version": "1.104.0",
  "system": {
    "python": "3.8.13",
    "os": "macOS-14.1-arm64-arm-64bit"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.6",
    "aws_cdk": "Not available",
    "terraform": "1.6.2"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}
```

cc: @lucashuy

jysheng123:
Hi, thanks for bringing up the issue. This seems like a bug on our end. I have marked it and will let the team know about it thanks.

PR #6537, opened 2024-01-08
Allow Security Field for Schema to be optional 

jysheng123:
#6428

As per the open API document here https://swagger.io/specification/, users are allowed to specify if they want their endpoint to be security optional, this is done by passing security in the spec as `security: [{}]`, however previously, this was treated as an error and raises an exception when it should be treated as no authorizer being passed in.
Checks the edge case and does not raise error when it occurs.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


sidhujus (approved):

ramosbugs:
thank you for fixing this!

hawflau (approved):

Issue #6449, opened 2023-12-14
Bug: sam delete works with global parameters defined in toml, but not yaml


rhbecker:
### Description:

The `sam delete` command cannot seem to find parameters defined in a `global` section of `samconfig.yaml` (but the same parameters specified in `samconfig.toml` work fine).

### Steps to reproduce:

These commands work fine with both `toml` and `yaml` editions of `samconfig` ...

```shell
sam build --config-env test
```
```shell
sam deploy --config-env test
```

This command works fine with the `toml` edition, but not the `yaml` edition of `samconfig` ...

```shell
sam delete --config-env test
```

Example toml file ...

```toml
version = 0.1

[test.global.parameters]
confirm_changeset = true
image_repositories = [ ]
parameter_overrides = [ ]
region = "us-west-2"
s3_bucket = "build-artifacts-REDACTED-ID"
s3_prefix = "cloudformation/sam-cli/test"
stack_name = "test"
template = ".dev/tests/test.yaml"
```

Example yaml file ...

```yaml
version: 0.1
test:
  global:
    parameters:
      confirm_changeset: true
      image_repositories: []
      parameter_overrides: []
      region: us-west-2
      s3_bucket: build-artifacts-REDACTED-ID
      s3_prefix: cloudformation/sam-cli/test
      stack_name: test
      template: .dev/tests/test.yaml
```

### Observed result:

Executing `sam delete --config-env test` with the `yaml` edition of `samconfig` results in a prompt ...

> Enter stack name you want to delete

... but `stack_name` _is_ specified.

If I play along, and execute ...

```shell
sam delete --config-env test --stack-name test
```

... it deletes the stack, but then can't find the parameters it needs to use to find the template in S3 ...

> Warning: Cannot resolve s3 bucket information from command options , local config file or cloudformation template. Please use --s3-bucket next time and delete s3 files manually if required.

### Expected result:

Configuration specified via `yaml` should function in an equivalent manner to configuration specified via `toml`.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Ubuntu via WSL on Windows 10
2. `sam --version`: 1.105.0
3. AWS region: us-west-2

```
# Paste the output of `sam --info` here
{
  "version": "1.105.0",
  "system": {
    "python": "3.11.3",
    "os": "Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.31"
  },
  "additional_dependencies": {
[...]

PR #6541, opened 2024-01-09
fix: sam delete not picking up global parameters from samconfig.yaml

sidhujus:
#6449 

the delete command was doing its own parsing of the config file

switch to use the decorator that reads values from config files and passes those to the function

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jysheng123 (approved):

samcli/commands/delete/delete_context.py
```diff
@@ -77,32 +71,6 @@ def __enter__(self):
...
-            config_path=self.config_file, config_env=self.config_env, cmd_names=[CONFIG_COMMAND]
-        )
-        if not config_options:
-            return
-
-        if not self.stack_name:
-            self.stack_name = config_options.get("stack_name", None)
-        # If the stack_name is same as the one present in samconfig file,
-        # get the information about parameters if not specified by user.
-        if self.stack_name and self.stack_name == config_options.get("stack_name", None):
```
lucashuy, line 95: Do we still need to have this logic in the delete code somewhere? If I remember correctly, the config decorator will load all parameters into the click context.

So if the options in the config were all related to stack `xyz`, and I instead did `sam delete --stack-name abc`, will it use the parameters related to `xyz`, for stack `abc`?

The previous behaviour looks like it won't, and instead will just not set the other options listed below, but feel free to call this out if's a non issue

tests/unit/commands/delete/test_delete_context.py
```diff
@@ -149,14 +121,12 @@ def test_delete_no_user_input(
     def test_delete_context_valid_execute_run(self, get_boto_client_provider_mock, patched_click_get_current_context):
         patched_click_get_current_context = MagicMock()
         with DeleteContext(
-            stack_name=None,
+            stack_name="test",
```
lucashuy, line 124: Did something else get read in for these values to `test`?

lucashuy (dismissed):

samcli/commands/delete/delete_context.py
mndeveci: This is a good catch. We may run integration tests manually to see if there are some tests which validates this behavior. 

It is also weird that we only set these parameters if stack name matches, if not we are just leaving that as it is, which might fail the command there 🤔

tests/unit/commands/delete/test_delete_context.py
```diff
@@ -73,26 +67,6 @@ def test_delete_context_enter(self, get_boto_client_provider_mock):
             )
         ),
     )
-    @patch("samcli.commands.delete.delete_context.click.get_current_context")
```
mndeveci, line 76: Why did we remove this test? This should resolve as it was written since those are the provided values right?

sidhujus: Since we're using the decorator to read the config and its values all the parsing of the config file happens there instead, and the DeleteContext just has the values passed in now instead of doing any parsing.

tests/unit/commands/delete/test_delete_context.py
sidhujus: Yea it was reading in some values from a config file somewhere, but since all that happens outside of the context now i had to set the values here

samcli/commands/delete/delete_context.py
sidhujus: I tried it out with my changes, and the behaviour now is that it will prioritize parameters passed in via the cli options over config values, but theres no logic to skip setting the other variables if the stack name differs between the config and cli parameter

hnnasit (approved):
I agree, `sam delete` configuration should be aligned with `sam deploy`. Changes look good to me overall.

PR #6551, opened 2024-01-12
fix: Added missing add_host parameter to start-api entry methods

lucashuy:

N/A

The recent `--add-host` PR that was merged in missed adding the new parameter to the methods that call the invoke context. This somehow passed CI jobs, but was caught by canaries.

Adds the missing parameter to the cli methods for tests and business logic.

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mndeveci (approved):
Thanks for patching this. Any other test that we can add to prevent this not happening again in the future?

hawflau (approved):

PR #6559, opened 2024-01-15
Fix: sam invoke local throw error 'utf-8' codec can't decode byte 0xff

sidhujus:
#6369

The code was previously refactored to use Strings to handle the IO related to sam local invoke and sam local start-lambda this fixed some issues where certain strings werent displayed correctly, but broke the functionality to support lambdas returning images

adds an extra BytesIO object to the stream writer and checks if it was written to in order to determine if it should return strings or the bytes

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


tests/integration/testdata/invoke/template-return-image.yaml
```diff
@@ -7,6 +7,6 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       Handler: main.image_handler
-      Runtime: python3.12
+      Runtime: python3.11
```
lucashuy, line 10: Did we ever find a root cause for this?

lucashuy (approved):

tests/integration/testdata/invoke/template-return-image.yaml
sidhujus: When we make the request to lambda it only ever returns a 502, provided the host machine is windows and the lambda is returning raw bytes. I assume that it's because a bug between docker and the python 3.12 runtime but I wasn't able to narrow it down further than that

mndeveci (approved):

tests/integration/testdata/invoke/template-return-image.yaml
mndeveci: Could that be docker version limitation for these new (al2023 based) runtimes? Were you able to run the tests on a Windows machine with newer Docker version to see if these tests are passing or not?

PR #6622, opened 2024-01-31
fix: lock invocation per function for not having concurrent requests

mndeveci:
#6157

When running `sam local start-api` SAM CLI spins up one container per lambda function. Any concurrent request might fail (or have side-effects) due to this limitation.

Create lock per function endpoint and only allow single execution per function.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/docker/container.py
```diff
@@ -378,11 +383,22 @@ def wait_for_http_response(self, name, event, stdout) -> Tuple[Union[str, bytes]
...
 
-        resp = requests.post(
-            self.URL.format(host=self._container_host, port=self.rapid_port_host, function_name="function"),
-            data=event.encode("utf-8"),
-            timeout=(self.RAPID_CONNECTION_TIMEOUT, None),
-        )
+        # generate a lock key with host-port combination which is unique per function
+        lock_key = f"{self._container_host}-{self.rapid_port_host}"
+        LOG.debug("Getting lock for the key %s", lock_key)
+        with CONCURRENT_CALL_MANAGER_LOCK:
```
mildaniel, line 389: Why do we need this lock and not just the `CONCURRENT_CALL_MANAGER`?

tests/integration/local/start_api/test_start_api.py
```diff
@@ -164,7 +165,7 @@ def setUp(self):
         self.url = "http://127.0.0.1:{}".format(self.port)
         HTTPConnection._http_vsn_str = "HTTP/1.1"
 
-    @pytest.mark.flaky(reruns=3)
+    # @pytest.mark.flaky(reruns=3)
```
mildaniel, line 168: Testing comment?

samcli/local/docker/container.py
mndeveci: We need locks for each lambda container, and in order to create/access that lock we need another one. So `CONCURRENT_CALL_MANAGER_LOCK` locks the first call to either create/access the lock for certain lambda function. And that accessed/created lock will actually lock the RIE call below.

tests/integration/local/start_api/test_start_api.py
mndeveci:
```suggestion
    @pytest.mark.flaky(reruns=3)
```

mildaniel (approved):

hnnasit (approved):

Issue #6510, opened 2024-01-04
Bug: Functions are permitted with invalid `MemorySize`


davidjb:
### Description:

As described in https://github.com/aws/aws-sam-cli/issues/6110, SAM (`build`, `local invoke`, etc) doesn't attempt to validate values for function `MemorySize`.

This value should be with 128 and 10240 MB according to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize but values such as `1`, `-1`, etc are allowed. For a Docker-based local invocation, a value < 6 MB will cause Docker to refuse to start the container.

### Steps to reproduce:

1. Create `template.yml`:

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Resources:
  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: myfunction
      Handler: app.lambdaHandler
      Runtime: python3.11
      MemorySize: 1
```

2. Run builds and observe results, as per next section below.

### Observed result:

`sam build MyFunction` succeeds, despite `MemorySize: 1` being an invalid value. Trying to invoke `sam local invoke MyFunction` fails because Docker requires a minimum of 6MB of memory allocated:

    ```sh
    Error: 400 Client Error for http+docker://localhost/v1.35/containers/create: Bad Request ("Minimum memory limit allowed is 6MB")
    Traceback:
    ...
    ```

### Expected result:

Validation errors would be shown when the SAM config is loaded, before the values are used for building and builds would not succeed when invalid values are used.

As per https://github.com/aws/aws-sam-cli/issues/3244, this value will be validated but only at deploy time, assuming you're able to create a configuration that is able to be built but not deploy.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
{
  "version": "1.105.0",
  "system": {
    "python": "3.11.6",
    "os": "Linux-6.4.16-linuxkit-aarch64-with-glibc2.35"
  },
  "additional_dependencies": {
    "docker_engine": "24.0.6",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
[...]

PR #6679, opened 2024-02-13
Add `MemorySize` validation for function build

bentvelj:
Validate function `MemorySize` to be an integer within range `[128, 10240]` as per [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize).

Also had to fix a bunch of unit tests that previously had fake MemorySize.

#6510 

Customers were permitted to build Lambda functions with invalid MemorySize, causing late/unexpected failure

Adds a validation step to tell the customer the MemorySize is invalid (not an integer, or outside the range `[128,10240]`)

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/local/start_api/test_start_api.py
+++ b/tests/integration/local/start_api/test_start_api.py
@@ -19,6 +19,13 @@ from .start_api_integ_base import StartApiIntegBaseClass, WritableStartApiIntegB
 from ..invoke.layer_utils import LayerUtils
 
 
+# class TestInvalidTempate(StartApiIntegBaseClass):
+#     template_path = "/testdata/start_api/invalid-template.yaml"
+
+#     def test(self):
+#         self.assertIsNotNone(self.start_api_process.poll())
+
+
 @parameterized_class(
     ("template_path",),
     [
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -1,6 +1,7 @@
 """
 Reads CLI arguments and performs necessary preparation to be able to run the function
 """
+
 import errno
 import json
 import logging
@@ -14,6 +15,7 @@ from samcli.commands.exceptions import ContainersInitializationException
 from samcli.commands.local.cli_common.user_exceptions import DebugContextException, InvokeContextException
 from samcli.commands.local.lib.debug_context import DebugContext
 from samcli.commands.local.lib.local_lambda import LocalLambdaRunner
+from samcli.commands.validate.validate import do_cli as run_sam_validate
 from samcli.lib.providers.provider import Function, Stack
 from samcli.lib.providers.sam_function_provider import RefreshableSamFunctionProvider, SamFunctionProvider
 from samcli.lib.providers.sam_stack_provider import SamLocalStackProvider
@@ -99,6 +101,7 @@ class InvokeContext:
         container_host_interface: Optional[str] = None,
         add_host: Optional[dict] = None,
         invoke_images: Optional[str] = None,
+        ctx: Optional[dict] = None,
     ) -> None:
         """
         Initialize the context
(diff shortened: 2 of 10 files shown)


tests/integration/testdata/buildcmd/template.yaml
```diff
@@ -8,15 +8,19 @@ Parameters:
     Type: String
   Handler:
     Type: String
+  MemorySize:
+    Type: Number
+    Default: "512"
 
 Resources:
-
+
```
bentvelj, line 16:
```suggestion
```

hawflau (changes requested):
Adding validation in build doesn't seem to me that it will resolve the issue customer described. `sam build` is not a prerequisite of `sam local invoke` (e.g. when the Lambda code has no external dependency at all or when customer has their own external build steps). `sam local *` seems to be the place where an incorrect config would fail invocation and thus needs validation.

I also wonder if we want to add our own validation logic. It's a duplication of knowledge as we already have `sam validate --lint` which would catch config errors. We can update `sam local *` to run `sam validate --lint` before starting to do the `local` stuff.

samcli/lib/build/app_builder.py
```diff
@@ -269,6 +273,39 @@ def _get_build_graph(
...
+                function_memory = int(function.memory)
+            except ValueError as ex:
+                raise InvalidFunctionDefinition(
+                    f'Function {function.function_id} has invalid MemorySize "{function.memory}". Must be an integer between 128 and 10240.'
+                ) from ex
+
+            if function_memory < MIN_MEMORY_SIZE or function_memory > MAX_MEMORY_SIZE:
+                raise InvalidFunctionDefinition(
+                    f"Function {function.function_id} has MemorySize {function.memory}, but it must be an integer between 128 and 10240."
+                )
```
mndeveci, line 307: Can we give a warning message rather than failing with the exception here? My concern is if this is changed and we missed it, then customers will be blocked building their lambda functions.

mndeveci:
> Adding validation in build doesn't seem to me that it will resolve the issue customer described. `sam build` is not a prerequisite of `sam local invoke` (e.g. when the Lambda code has no external dependency at all or when customer has their own external build steps). `sam local *` seems to be the place where an incorrect config would fail invocation and thus needs validation.

In align with @hawflau's feedback here, I think this validation should be done at `sam local` command so that we won't spin up docker container with invalid memory configuration. Or we could catch this error and report it back to customer with a detailed message.

I wouldn't add a validation logic to `sam build` and let `sam validate --lint` to verify template is valid and deployable.

samcli/lib/build/app_builder.py
bentvelj: Good point, however with Wing's comment we'll only be relying on `cfnlint` which we can expect to stay in sync with the CFN rules?

bentvelj:
I've refactored to run the `sam validate` command in the `__entry__` of `InvokeContext`, which is run before `sam local invoke, start-api, start-lambda`. The caveat to this is that we need command context passed into `InvokeContext`.  This is an optional parameter, to avoid breaking all the isolated `InvokeContext` unit tests, and if the ctx is not provided, the `sam validate` step is skipped.

Integ tests to follow.

tests/integration/local/start_api/test_start_api.py
```diff
@@ -19,6 +19,13 @@
 from ..invoke.layer_utils import LayerUtils
 
 
+# class TestInvalidTempate(StartApiIntegBaseClass):
```
hawflau, line 22: nit: remove the lines if they are not needed

bentvelj: Oops, didn't mean to commit this

bentvelj:
Closing in favour of new [PR](https://github.com/aws/aws-sam-cli/pull/6719) on separate branch

Issue #6663, opened 2024-02-07
Concurrent requests to `sam local start-api` cause crash in _clean_decompressed_paths function and 502 in browser


jg210:
### Description:

Concurrent requests from browser when using `sam local start-api` cause crash in sam's _clean_decompressed_paths function.

### Steps to reproduce:

https://github.com/jg210/spring-experiments/pull/35/commits/d09a8fe648d5b61205a3e763d117097a72497036, but don't use --warm-containers EAGER.

Likely happens for any page with two files that are similar size.

### Observed result:

An index.html makes browser load js bundle and css file at same time. Only one succeeds (not always same one) since the concurrent cleanups here collide with each other. Different file is mentioned in the crash each time reproduce it.

```
Exception on /assets/index-tz9LntmO.js [GET]                                                                                                                                                                                                                                              
Traceback (most recent call last):                                                                                                                                                                                                                                                        
  File "/home/jg210/repo/jg210/spring-experiments/.venv/lib/python3.12/site-packages/flask/app.py", line 1463, in wsgi_app                                                                                                                                                                
    response = self.full_dispatch_request()                                                                                                                                                                                                                                               
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                               
  File "/home/jg210/repo/jg210/spring-experiments/.venv/lib/python3.12/site-packages/flask/app.py", line 872, in full_dispatch_request                                                                                                                                                    
    rv = self.handle_user_exception(e)                                                                                                                                                                                                                                                    
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                    
  File "/home/jg210/repo/jg210/spring-experiments/.venv/lib/python3.12/site-packages/flask/app.py", line 870, in full_dispatch_request                                                                                                                                                    
    rv = self.dispatch_request()                                                                                                                                                                                                                                                          
         ^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                          
  File "/home/jg210/repo/jg210/spring-experiments/.venv/lib/python3.12/site-packages/flask/app.py", line 855, in dispatch_request                                                                                                                                                         
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]                                                                                                                                                                               
[...]

PR #6695, opened 2024-02-14
Add a Lock for removing decompressed paths

jysheng123:
#6663 

When a temporary directories are made and parallel requests are made, it concurrently tries to delete the same temp directory, resulting in errors due to race conditions.

By adding a lock on each instance of the runtime, we are removing the temp directory one at a time, then the next run of the same instance sees that it is already deleted and does not try to remove an nonexistent path

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/local/lambdafn/runtime.py
```diff
@@ -353,9 +353,10 @@ def _clean_decompressed_paths(self):
         Clean the temporary decompressed code dirs
         """
         LOG.debug("Cleaning all decompressed code dirs")
-        for decompressed_dir in self._temp_uncompressed_paths_to_be_cleaned:
-            shutil.rmtree(decompressed_dir)
-        self._temp_uncompressed_paths_to_be_cleaned = []
+        with self._container_manager._lock:
```
lucashuy, line 356: Could we make a different lock instead of using the one that the manager uses for image pulling? That way it doesn't mess with the image lock managing logic and the temp directory cleaning logic can happen separately.

tests/integration/local/start_api/test_start_api.py
```diff
@@ -148,6 +148,39 @@ def test_large_input_request_http10(self):
...
+        HTTPConnection._http_vsn_str = "HTTP/1.1"
+
+    @pytest.mark.flaky(reruns=3)
+    @pytest.mark.timeout(timeout=600, method="thread")
+    def test_same_endpoint_zipped(self):
+        """
+        Send two requests to the same path at the same time. This is to ensure we can handle
+        multiple requests at once and do not block/queue up requests
+        """
+        number_of_requests = 10
```
lucashuy, line 165: The comment says two request, but the variable is set to 10, is it suppose to be 10 requests?

tests/integration/local/start_api/test_start_api.py
```diff
@@ -148,6 +148,39 @@ def test_large_input_request_http10(self):
         self.assertEqual(response.raw.version, 11)
 
 
+class TestParallelRequestsZipped(StartApiIntegBaseClass):
```
lucashuy, line 151: If this is the same test as the one below (`TestParallelRequests`), we could parallelize the endpoint instead.

tests/integration/local/start_api/test_start_api.py
jysheng123: Yeah, Ill change the comment

sidhujus (approved):

lucashuy (approved):
Thanks for making the changes

PR #6719, opened 2024-02-20
Handle Container Creation Failure in `local invoke`, `local start-api` and `local start-lambda`

bentvelj:
Re-raising this [PR](https://github.com/aws/aws-sam-cli/pull/6679) to handoff to another developer, since the previous PR was from a personal fork.

Validate function `MemorySize` to be an integer within range `[128, 10240]` as per [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-memorysize).

Also had to fix a bunch of unit tests that previously had fake MemorySize.

#6510 

Customers were permitted to build Lambda functions with invalid MemorySize, causing late/unexpected failure

- during `__init__` of `InvokeContext`, run lint and display any lint errors as warning
- Catch any Docker API error while creating container and reraise as `DockerContainerCreationFailedException`. Each `local` command can handle the error differently:
  - `local invoke` will surface this error and exit
  - `local start-api` and `local start-lambda` will display a warning but would not exit the process

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -219,6 +225,10 @@ def __enter__(self) -> "InvokeContext":
 
         self._stacks = self._get_stacks()
 
+        if self._ctx:
+            LOG.info("Validating template %s\n", self._template_file)
+            run_sam_validate(ctx=self._ctx, template=self._template_file, lint=True)
```
mndeveci, line 230: Is this going to fail the command if validation fails?

hawflau: seems like it could raise User Exception. Does it make sense to not be able to run invoke if a template is invalid?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -154,6 +157,8 @@ def __init__(
             Optional. Docker extra hosts support from --add-host parameters
         invoke_images dict
             Optional. A dictionary that defines the custom invoke image URI of each function
+        ctx dict
+            Optional. Command invoke context used to invoke commands before invoking `sam local` logic
```
mndeveci, line 161: If we are only using to grab the `debug` flag, maybe we can pass that field instead of the whole `Context` object?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -220,6 +226,11 @@ def __enter__(self) -> "InvokeContext":
 
         self._stacks = self._get_stacks()
 
+        if self._ctx:
+            _, matches_output = get_lint_matches(self._template_file, self._ctx.debug, self._aws_region)
```
mndeveci, line 230: Should we wrap this call with `try/except` since there might be some exceptions which is thrown by linter logic? (I saw we are checking valid region configuration for instance).

mildaniel (approved):

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -220,6 +225,13 @@ def __enter__(self) -> "InvokeContext":
 
         self._stacks = self._get_stacks()
 
+        try:
+            _, matches_output = get_lint_matches(self._template_file, self._verbose, self._aws_region)
+            if matches_output:
+                LOG.warning("Lint Error found, containter creation might fail: %s", matches_output)
+        except UserException as e:
```
mndeveci, line 232: Can we make sure everything is wrapped as `UserException` so there won't be a new exception type which might be thrown from `get_lint_matches`?

mndeveci (approved):

samcli/commands/local/cli_common/invoke_context.py
hawflau: errors within cfn-lint seem to be caught and wrapped as `CfnLintExitException` which is handled in this method. I think it should be fine.

PR #6721, opened 2024-02-20
Centralize Configurations for Runtime

hawflau:

N/A

- To reduce number of code changes due to future new/deprecated Lambda Runtimes. 

- The idea is to put all runtime configurations in one place, and use that to generate all the mappings and data
- This PR contains the centralized config and a small amount of refactor. If this PR looks good, it will be merged and we can move on the whole refactor.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    ARM64 = "arm64"
+
+
+@dataclass
+class FamilyDataMixin:
+    key: str
+    layer_subfolder: Optional[str]
+    dep_manager: List[Tuple[str, Path]] = field(default_factory=list)  # (package manager, path to local init template)
+    build: bool = True
+    eb_code_binding: Optional[str] = None  # possible values are "Java8", "Python36", "Go1", "TypeScript3"
```
mildaniel, line 27: What is this field for?

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    is_lambda_enum: bool = True
+    # build_workflow: WorkflowConfig
+    # workflow_selector: ???
+
+    # debug settings - samcli/local/docker/lambda_debug_settings.py
+    # debug_entry_setting: callable
+
+    # runtime enum - samcli/local/docker/lambda_image.py
+
+    # DEPRECATED_RUNTIMES - samcli/lib/build/constants.py
```
mildaniel, line 86: What are these comments for?

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    is_deprecated: bool = False
+
+    # local manifest - samcli/lib/init/local_manifest.json
+    # is local manifest useful at all?
```
mildaniel, line 90: I believe we fall back to the local manifest if there's an error fetching from GitHub

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    def archs_as_list_of_str(self) -> List[str]:
+        return [a.value for a in self.archs]
+
+    @property
+    def is_provided(self):
+        return self.family == Family.PROVIDED
+
+    def to_local_init_manifest(self) -> List[Dict]:
```
mildaniel, line 98: Is there a plan to implement this? Should we include a SIM in a comment?

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+                "useCaseName": "Hello World Example"
+            }
+        ]
+        """
+        raise NotImplementedError()
+
+
+class RuntimeEnumBase(RuntimeDataMixin, Enum):
+    @classmethod
+    def from_str(cls, runtime: Optional[str]) -> "RuntimeEnumBase":
```
mildaniel, line 125: The return type has to be in quotes because python doesn't like you returning an instance from a class method?

mildaniel (commented):
LGTM, just left some questions to better my understanding

samcli/lib/runtimes/base.py
hawflau: This is for generating the EventBridge event code-binding

samcli/lib/runtimes/base.py
hawflau: These are basically TODOs that I need to implement progressively. 

Each item will be cleaned up once I address it. But I believe for DEPRECATED_RUNTIME has been addressed in PR already. I'll remove it

samcli/lib/runtimes/base.py
hawflau: yes, internally I have a Story SIM to track all the subtasks

samcli/lib/runtimes/base.py
hawflau: What I wanted to do is to return the self type.
The type checker didn't like it somehow

Seems like there's a `Self` type but it's available in 3.11+ https://peps.python.org/pep-0673/

samcli/lib/runtimes/base.py
hawflau: yeah, I think my conclusion was that we can generate it during runtime as a dictionary instead of reading from a json file. That's why I have the `to_local_init_manifest` method to be implemented below

samcli/lib/runtimes/base.py
mildaniel: I'm just not familiar with this? Why can it only be those values?

samcli/lib/runtimes/base.py
mildaniel: Looks like you might be able to do something using generic bounds like
```
Self = TypeVar("Self", bound="Shape")

class Shape:
    @classmethod
    def from_config(
        cls: type[Self], config: dict[str, float]
    ) -> Self:
        return cls(config["scale"])
```

though I don't particularly think that's any more intuitive than what you have now

samcli/lib/runtimes/base.py
mildaniel: Do you have a strong reason for generating it instead of reading from a file?

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    ruby32 = (
+        "ruby3.2",
+        Family.RUBY,
+        "amazon/ruby3.2-base",
+        [Architecture.X86_64, Architecture.ARM64],
+    )
+
+
+@unique
+class DeprecatedRuntime(RuntimeEnumBase):
```
mndeveci, line 283: When a runtime deprecates, we move it into this list by just preserving the first two fields?

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+    nodejs12x = ("nodejs12.x", Family.NODEJS)
+    nodejs14x = ("nodejs14.x", Family.NODEJS)
+    python27 = ("python2.7", Family.PYTHON)
+    python36 = ("python3.6", Family.PYTHON)
+    python37 = ("python3.7", Family.PYTHON)
+    ruby25 = ("ruby2.5", Family.RUBY)
+    ruby27 = ("ruby2.7", Family.RUBY)
+
+
+def runtime_dep_template_mapping(runtimes: List[RuntimeDataMixin]) -> dict:
```
mndeveci, line 300: nit: can we use full names in the method, I am not sure if `dep` is for dependency or something else?

samcli/lib/utils/architecture.py
```diff
@@ -55,16 +35,26 @@ def validate_architecture_runtime(function: "Function") -> None:
     if function.packagetype == IMAGE:
         return
 
-    runtime_architectures = SUPPORTED_RUNTIMES.get(cast(str, function.runtime), [])
+    runtime_architectures = []
+    try:
+        runtime = Runtime.from_str(function.runtime)
+        runtime_architectures.extend(runtime.archs_as_list_of_str())
+    except ValueError:
+        LOG.debug("Unrecognized runtime %s", function.runtime)
```
mndeveci, line 43: Should this be `debug` or `warning`/`error`?

samcli/lib/runtimes/base.py
hawflau: https://docs.aws.amazon.com/eventbridge/latest/schema-reference/v1-registries-name-registryname-schemas-name-schemaname-language-language.html#PutCodeBinding

These are the enums / languages that EB Schema Code Binding supports.

samcli/lib/runtimes/base.py
hawflau: Having a file means we need to have another location to maintain the list of runtimes. The goal for runtime improvement is to minimize the number of changes needed for adding/deprecating a runtime

samcli/lib/runtimes/base.py
hawflau: yep, that's right

mildaniel (approved):

samcli/lib/utils/architecture.py
hawflau: I updated it to `warning`. But I don't think it matters too much. At the end, if the runtime is unrecognized, this means `runtime_architectures` will be empty. And therefore L45 will always true and will raise `UnsupportedRuntimeArchitectureError`. 

And if no `Architecture` is provided, it will still fail with the unsupported runtime error like below:
```
Invoking bootstrap (providedi.al2023)                                                                                                                                                                                                                                                                                                                                                                                                                        
Error: Unsupported Lambda runtime: providedi.al2023. For a list of supported runtimes, please visit https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
```

samcli/lib/runtimes/base.py
```diff
@@ -0,0 +1,342 @@
...
+from dataclasses import dataclass, field
+from enum import Enum, unique
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple, cast
+
+LOG = logging.getLogger(__name__)
+
+_init_path = Path(os.path.dirname(__file__)).parent.parent
+_templates = _init_path / "lib" / "init" / "templates"
+_lambda_images_templates = _init_path / "lib" / "init" / "image_templates"
```
mndeveci, line 12: Sorry to miss this in previous review but constants are usually defined in all-uppercase

mndeveci (approved):
LGTM! Left a non-blocking comment

samcli/lib/runtimes/base.py
hawflau: yep I can change them in subsequent PRs

PR #6753, opened 2024-02-28
fix: don't error out if no content-type header

sidhujus:
#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write/update unit tests
- [ ] Write/update integration tests
- [ ] Write/update functional tests if needed
- [ ] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #6757, opened 2024-02-29
fix: Use .get to avoid key error when content-type is not in response

BradleyJinks:
When using this with the  `Amazon.Lambda.RuntimeSupport` dotnet package it looks like a Content-Type is not returned and a key error is thrown. I have verified that this fixes the issue with the package.

The .get method returns None if the value does not exist unlike direct access which throws a key error

None that I can think of.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


BradleyJinks:
Closing as this is addressed in #6753

PR #6821, opened 2024-03-13
Revert "Revert "Fix bug in PR#6719""

hawflau:
Reverts aws/aws-sam-cli#6800


hawflau:
Closing after discussed with team. We might need to rework the solution to avoid having cfn-lint modified any env var

PR #6831, opened 2024-03-14
fix: use parameter overrides in package context

sidhujus:
When running sam deploy with parameter overrides the overridden values are not being provided to the package context which causes intrinsic !Ref functions to not get resolved, which leads to a deployment failing when it shouldn't. This only happens when trying to set the `SemanticVersion` of a `serverless::application` through the via a !Ref to a property in the template that does not have a valid default value provided

passes the values into the package context

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jysheng123:
Can you add an integration test case for this situation, other than that it looks fine to me

jysheng123 (approved):

samcli/commands/deploy/command.py
```diff
@@ -326,6 +326,11 @@ def do_cli(
             region=guided_context.guided_region if guided else region,
             profile=profile,
             signing_profiles=guided_context.signing_profiles if guided else signing_profiles,
+            parameter_overrides=(
+                sanitize_parameter_overrides(guided_context.guided_parameter_overrides)
+                if guided
+                else parameter_overrides
+            ),
```
lucashuy, line 333: Could this be extracted out as a variable, since the `DeployContext` uses this param overrides too? We can also unit test to make sure that the package context and deploy context actually receive the param overrides.

lucashuy (approved):

PR #6833, opened 2024-03-14
chore: Deprecate java8 for provided

hnnasit:
Note: `java8` is being deprecated for only provided OS. Since some integ tetsts for `java8.al2` still use the `java8` templates, these are not removed.

Java8 lambda runtime for provided is deprecated based on https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html and needs to be removed from SAM CLI

Removes integ tests and init templates for java8

No side effects

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel (approved):
LGTM, just make sure to update this list https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/constants.py#L7

sidhujus (approved):

Issue #3244, opened 2021-09-06
Support immediate validation of parameter values on entry


specious:
It would be valuable to have immediate parameter value validation upon entry.

For example, a parameter may be defined in `template.yml` (e.g. as per the [official documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-masteruserpassword) for RDS password constraints): 

```yaml
Parameters:
  DBAdminPassword:
    Description: Database admin password
    Type: String
    MinLength: 8
    MaxLength: 128
    AllowedPattern: "[^\\s@\"\\/]*"
    ConstraintDescription: 'The password can include any printable ASCII character except "/", """, or "@".'
```

It would then be useful to have the value validated against both the implicit constraints and explicitly stated constraints upon entry:

```
$ sam deploy -g

Configuring SAM deploy
======================

        Looking for config file [samconfig.toml] :  Found
        Reading default arguments  :  Success

        Setting default arguments for 'sam deploy'
        =========================================
        Stack Name [myapp-backend]:
        AWS Region [us-east-1]:
        Parameter DBInstanceClass [db.t2.micro]:
        Parameter DBPostgresVersion [12.7]:
        Parameter DBAllocatedStorage [5]:
        Parameter DBName [mydb]:
        Parameter DBAdminUser [admin]:
        Parameter DBAdminPassword [] (The password can include any printable ASCII character except "/", """, or "@".): ZCd)6Qyt~T^<"/\!3i,2L:&,&=*NzU#[

        Invalid parameter value.

        Parameter DBAdminPassword [] (The password can include any printable ASCII character except "/", """, or "@".): 
```

The current behavior is to let the value pass until the deployment initiation phase, at which point the deployment attempt doesn't pass validation:

```
$ sam deploy -g

Configuring SAM deploy
======================

        Looking for config file [samconfig.toml] :  Found
        Reading default arguments  :  Success

        Setting default arguments for 'sam deploy'
        =========================================
        Stack Name [myapp-backend]:
        AWS Region [us-east-1]:
        Parameter DBInstanceClass [db.t2.micro]:
        Parameter DBPostgresVersion [12.7]:
        Parameter DBAllocatedStorage [5]:
        Parameter DBName [mydb]:
        Parameter DBAdminUser [admin]:
[...]

PR #6922, opened 2024-04-09
Handle docker container creation error

hawflau:

#6510

To catch and handle docker container creation errors when running `sam local invoke`, `sam local start-api` and `sam local start-lambda`

Catch the error and propagate it to CLI

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/invoke/cli.py
+++ b/samcli/commands/local/invoke/cli.py
@@ -16,7 +16,11 @@ from samcli.commands.local.invoke.core.command import InvokeCommand
 from samcli.commands.local.lib.exceptions import InvalidIntermediateImageError
 from samcli.lib.telemetry.metric import track_command
 from samcli.lib.utils.version_checker import check_newer_version
-from samcli.local.docker.exceptions import ContainerNotStartableException, PortAlreadyInUse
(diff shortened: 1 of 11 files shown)


mndeveci (approved):
Changes LGTM! Can we add any integration tests that we can validate the changes?

jysheng123 (commented):
I think the code looks fine, but im not too sure if this would fix the customer's issue specifically. I think the customer is asking for validation before trying to create the container that will say the memory size is invalid, the code changes here looks like it will just do validation during the invoke, but it should be validating during the SAM CLI build step.

jysheng123 (approved):

hawflau:
To summarize before merging

The change in this PR is to catch and handle Docker Container Creation Error and show a message about the error and telling customer to check the template. SAM CLI provides the `sam validate --lint` command to catch template issues.
 
The expectation on the issue #6510 was "Validation errors would be shown when the SAM config is loaded, before the values are used for building and builds would not succeed when invalid values are used", which was captured in #3244 as a feature request.

Issue #6909, opened 2024-04-05
Feature request: AWS::Serverless::Function.ImageUri Should Also Accept a Local File Path


chrisoverzero:
### Describe your idea/feature/enhancement

I wish SAM CLI would accept a local path to an image archive (the kind of thing which can be loaded by `docker image load`) for the value of `ImageUri`.

### Proposal

For an`ImageUri` containing a local path to an image archive, the command `sam build` would perform the equivalent of `docker image load` (in the same way that it will currently perform the equivalent of `docker image build`) and write the image name and tag into the built template as the new value of `ImageUri`.

The command `sam package` would do the same thing, but additionally perform the equivalent of the `docker image push` and write the address of the image in the remote repository into the output template as the new value of `ImageUri`.

In general, it would make `ImageUri` work a little more like `CodeUri`.

Things to consider:
1. _Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)._ Yes, the `ImageUri` property would have to be documented to accept this kind of value.
2. _What if the archive is associated with multiple tags?_ If we have the taste for it, that could be an error. Alternatively, an action equivalent to `docker image import` could be performed instead, assigning a single canonical, generated name to the image in the same way as `sam build` does.
3. _Why would anyone want this?_ W-Well, I want this because my build system is capable of creating Docker images as file archives without the `docker` command-line tool or the docker service available. I know of other teams that produce file archives of Docker images for scanning purposes. Lacking either end of Docker simplifies those steps in the CI process, so we only need to associate the docker-in-docker service with the job in which we perform `sam` commands for the final deployment. Also, it's much easier in many ways to manage build artifacts which are normal files.
4. _Is this a thing that SAM CLI can possibly do?_ I admit I don't know for certain. `sam build` does its Docker work without the `docker` CLI tool available by using `docker+http://` URLs somehow. I _presume_ it's issuing commands to the Docker service over a more direct interface, and I see this capability in the Docker API, so it feels possible.

### Additional Details

I would be willing to take on this work if this request is accepted. I've contributed to serverless-application-model and aws-sam-build-images before, so I probably wouldn't be _completely_ in over my head.

lucashuy:
Thanks for raising this issue. We open to accepting this request, and can chat here about implementation details, that way other team members can find this issue and reply in case they had more context or ideas.

[...]

PR #6930, opened 2024-04-10
feat: sam Commands Understand Local File Paths for `AWS::Serverless::Function.ImageUri`

chrisoverzero:
As a summary, sam has learned to load an an image from a local archive before proceeding with the `build`, `package`, and `deploy` commands.

When running `sam build` with an `ImageUri` pointing to a local file, sam will load that archive into an image, then write the ID of the image to the `ImageUri` property of the built template. ID works the same as a tag for the Docker API, so business continues as usual from here. The reason behind writing ID is that a loaded image could be associated with multiple tags, and selecting one arbtrarily leads to difficulties in the deploy command.

The package and deploy commands have three kinds of value for `ImageUri` to consider. First, a value of the form `{repo}:{tag}`. This functions as it always has. Second, an image ID (in the form `sha256:{digest}`) which is probably the output of `sam build`. In this case, the tag translation uses the name of the as its input. Otherwise, they'd all end up with names starting with "sha256". Last, a local file. In this case, it proceeds as it does in the build command: Load the archive into an image first, then pass the resource name into tag translation.

See: #6909

#6909

From the initial ticket:

> W-Well, I want this because my build system is capable of creating Docker images as file archives without the docker command-line tool or the docker service available. I know of other development teams who produce file archives of Docker images for scanning purposes. Lacking either end of Docker simplifies those steps in the CI process, so we only need to associate the docker-in-docker service with the job in which we perform sam commands for the final deployment. Also, it's much easier in many ways to manage CI artifacts which are normal files.

sam has learned to load an an image from a local archive before proceeding with the `build`, `package`, and `deploy` commands when such a local archive is the value of `ImageUri` for an `AWS::Serverless::Function`

The kinds of values understood for `ImageUri` has increased:  A docker tag, a docker ID, and a local file archive. Though I think that a docker ID would always have _worked_, but unpleasantly. Either way, it wasn't documented and presumably not supported. Now it would be.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/build/build_graph.py
```diff
@@ -107,6 +107,7 @@ def _toml_table_to_function_build_definition(uuid: str, toml_table: tomlkit.api.
     function_build_definition = FunctionBuildDefinition(
         toml_table.get(RUNTIME_FIELD),
         toml_table.get(CODE_URI_FIELD),
+        None,
```
chrisoverzero, line 110: Anything related to loading from TOML was skipped over because I don't know whether that's a feature we/you want to support. It can always be added later if there's a requirement for feature parity.

samcli/lib/build/build_strategy.py
```diff
@@ -166,6 +168,7 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
         result = self._build_function(
             build_definition.get_function_name(),
             build_definition.codeuri,  # type: ignore
+            build_definition.imageuri,
```
jysheng123, line 171: Nit: Similar to the other optional parameters, should this also have a `# type: ignore?`

samcli/lib/package/ecr_uploader.py
```diff
@@ -76,10 +77,26 @@ def upload(self, image, resource_name):
         if not self.login_session_active:
             self.login()
             self.login_session_active = True
+
+        # cosborn: We sometimes use the resource_name as the `image` parameter to `tag_translation`.
```
jysheng123, line 81: What does cosborn mean here?

samcli/lib/build/build_strategy.py
chrisoverzero: They appear to have that annotation because they're _secretly_ optional. The parameter is typed as `str`, but the argument is passed as `Optional[str]`, and this quiets the linter. To be consistent, I'd type `imageuri` the same – but since it _is_ optional, being truthful about it seemed better.

samcli/lib/package/ecr_uploader.py
chrisoverzero: That's me. Around the code base, larger comments are signed, so I tried to follow suit. I'm happy to remove it if there's a convention.

jysheng123: I think for the final merge, we don't need to the signature anymore, otherwise the documentation is very clear, maybe would prefer if it less personal instead of using terms like We but thats just me nitpicking

chrisoverzero: Sure thing. I'll make that change now.

tests/unit/lib/build_module/test_build_strategy.py
```diff
@@ -833,7 +839,7 @@ def test_wrapper_with_or_without_container(self, mocked_read, mocked_write, runt
             use_container,
         )
 
-        build_definition = FunctionBuildDefinition(runtime, "codeuri", "packate_type", X86_64, {}, "handler")
+        build_definition = FunctionBuildDefinition(runtime, "codeuri", None, "packate_type", X86_64, {}, "handler")
```
jysheng123, line 842: I think this is an old typo, but can we change it to package_type from packate_type for the occurrences?

chrisoverzero:
> I do think however this does need some testing both unit/integration test wise.

I agree and I'm happy to try. Much of the test code feels a bit beyond my current Python familiarity, so it could take me some time to figure it out.

jysheng123:
No worries take your time, I would be more than happy to take a look at your tests/ if you have any questions.

chrisoverzero:
@jysheng123 When I go to run integration tests, will that perform deployments, or would only end-to-end tests do that? I need to know if I should have a scratch account ready to go so I don't get the _hey what is happening cloudtrail is going crazy with deployments to us-east-1 or something_ kind of Slack messages.

jysheng123:
Some of our integration tests will perform deployments, they should be deleted after the tests are done but if you want to be safe and not touch your actual account, I would recommend using another account

chrisoverzero:
Great, thanks so much. I will work on getting that set up.

tests/integration/package/test_package_command_image.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
jysheng123, line 1: Is this logging needed? I dont see it used in the codebase below

jysheng123 (approved):
Looks good to me? Ill ask another developer to take a look at this, thanks for bringing up the PR

tests/integration/package/test_package_command_image.py
chrisoverzero: Oops, I’m sure that’s left over from testing. Would you like me to remove it, or not touch this.PR after your approval?

sidhujus:
Hi @chrisoverzero thanks for this PR and the thorough testing! The integration tests you added seem to be failing because the template is in the incorrect location. For these tests the test data lives under the `package` folder https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/package/package_integ_base.py#L62 (the `DeployIntegBase` class inherits from the `PackageIntegBase` class. Could you move the required files into the correct location and then I can verify the new integration tests pass. Thanks!

tests/integration/package/test_package_command_image.py
sidhujus: we should remove any unused imports

sidhujus:
other than fixing the failing integration test and removing that one unused import the changes look good to me overall as well. Once thats fixed and we verify that the tests are all passing we can merge this in, thanks once again!

sidhujus (approved):

PR #6989, opened 2024-04-29
fix: mac validate pyinstaller workflow to not use latest mac image

sidhujus:
Using MacOS14 leads to issues with python/pip not identifying the installed version of OpenSSL which causes errors when trying to install dependencies. This does not happen with MacOS13 so pinning the version to 13 to let the PR workflow pass 

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


hnnasit (approved):
Could you also update the description about why we are pinning it to macos-13?

hawflau (approved):

PR #6996, opened 2024-04-29
fix: use correct openssl when building python

sidhujus:
our mac pyinstaller build does not specify the openssl directory so we are unable to install packages through pip

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/validate_pyinstaller.yml
+++ b/.github/workflows/validate_pyinstaller.yml
@@ -39,7 +39,7 @@ jobs:
 
   build-for-mac:
     name: build-pyinstaller-macos
-    runs-on: macos-13
+    runs-on: macos-latest
     if: github.repository_owner == 'aws'
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/validate_pyinstaller.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
   build-for-mac:
     name: build-pyinstaller-macos
-    runs-on: macos-13
```
mildaniel, line 42: I think this was pinned to 13 before since that's what our installer system uses. Can we revert this back?

hnnasit (approved):

.github/workflows/validate_pyinstaller.yml
mildaniel: Why would we not want to keep it pinned to the version that most closely matches our internal systems?

mildaniel (approved):

.github/workflows/validate_pyinstaller.yml
mildaniel: After giving it some more thought, I don't think it matters much which way we go here. Let's get these merged in

PR #6568, opened 2024-01-17
Added the ability to specify --accountid parameter on the command line

defenderkev:
#### Which issue(s) does this change fix?
#2325

#### Why is this change necessary?
Because currently if you have !Sub xxx`${AWS::AccountId}`xxx in, for example, a layer definition, SAM uses a default substitution which doesn't reference the correct Account ID

#### How does it address the issue?
It adds the ability to specify a command line option( --accountid ). This parameter is then used during the substitution of parameters in the CloudFormation template to replace ${AWS::AccountId} rather than using a hardcoded default (123456789012) 

#### What side effects does this change have?
None that I can see

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [n/a] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [n/a ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [n/a] Write/update integration tests
- [n/a] Write/update functional tests if needed
- [x] `make pr` passes
- [n/a] `make update-reproducible-reqs` if dependencies were changed
- [x] Write documentation
      -  parameter description added to 'AWS Credential Options' section for --help

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

Issue #2325, opened 2020-10-25
Enable Overriding of AWS::AccountId psuedo-parameter for local lambda invocation


Bradleywboggs:
### Describe your idea/feature/enhancement

I'd like to be able to override the `AWS::AccountId` pseudo parameter in `sam local` commands , similar to how the `--region`  option overrides the `AWS::Region` pseudo-parameter.
This most common use-case where this would be needed is if you have  paramaterized ARNs in the `Layers` node of the template. However, I have a use-case where even in local development, I may want to do end-to-end testing which interacts with already deployed resources in the env, whose ARN's also have the `AWS::AccountId` pseudo-parameter in my template.
 
At this point, if you depend on this parameter to access deployed resources, you receive a `403`, because you don't have access to resources from account `1234567890`
### Proposal

1.  add `--account-id abcde12345` option in `sam local` commands, possibly like the the `aws_creds_options` (doing it this way would mean adding an `account_id` prop to the Context obj-- not sure if it'd just be preferable to have it be a standalone value accessible through `account_id` in the `cli` method, so as not to modify the existing Context object),
```
def account_id_option(f):
    """
    Configures --account-id option for CLI

    :param f: Callback Function to be passed to Click
    """
    def callback(ctx, param, value):
        state = ctx.ensure_object(Context)
        state.account_id = value
        return value

    return click.option(
        "--account-id",
        expose_value=False,
        help="Select a specific account to associate with your invocation.",
        callback=callback,
    )(f)
```
2.) Add `_aws_account_id` property to the `InvokeContext` class
3. ) Change `IntrinsicSymbolTable.handle_pseudo_account_id`
from: 
```
@staticmethod
def handle_pseudo_account_id():
"""
This gets a default account id from SamBaseProvider.
Return
-------
A pseudo account id
"""
return IntrinsicsSymbolTable.DEFAULT_PSE
```
to (matching the impl of `handle_psuedo_region)`:
```
def handle_pseudo_account_id(self):
"""Gets the accountId from the environment and defaults to a the default accountID from the global variables.

This is only run if it is not specified by the logical_id_translator as a default.

Return
-------
An account id
"""
return (
self.logical_id_translator.get(IntrinsicsSymbolTable.AWS_REGION)
or os.getenv("AWS_ACCOUNT_ID")
[...]

PR #7013, opened 2024-05-02
Allow `sam local invoke` to retrieve account id from current logged in session

defenderkev:
https://github.com/aws/aws-sam-cli/issues/2325
Replaces Pull Request #6568 after feedback from team

Why is this change necessary?
Because currently if you have !Sub xxx${AWS::AccountId}xxx in, for example, a layer definition, SAM uses a default substitution which doesn't reference the correct Account ID

During instantiation of the `InvokeContext` object an attempt is made to retrieve the caller identity from STS. Assuming this succeeds, the account id is put into `_global_parameter_overrides`. If a token retrieval error occurs, a warning message is printed and the existing default value is used.

None that I can see

Mandatory Checklist

[n/a] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
[n/a ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
 Write/update unit tests
[n/a] Write/update integration tests
[n/a] Write/update functional tests if needed
 make pr passes
[n/a] make update-reproducible-reqs if dependencies were changed
 Write documentation
- Not sure where this should be documented. Happy to do so if pointed in the right direction


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -172,6 +174,17 @@ def __init__(
...
+        # Attempt to get the Account ID from the current session (if any)
+        sts = boto3.client("sts")
+        try:
+            account_id = sts.get_caller_identity()["Account"]            
+            if account_id:
+                if self._global_parameter_overrides is None:
+                    self._global_parameter_overrides = {}
+                self._global_parameter_overrides["AWS::AccountId"] = account_id
+        except TokenRetrievalError:
+            LOG.warning( "No current session found, using default AWS::AccountId" )
```
mildaniel, line 187: nit: Can we move this logic into its own function? It should also make it easier to test.

mildaniel (commented):
Thanks for the contribution. This looks good to me overall. I left one small nit-picky comment. Can you also add some unit testing?

samcli/commands/local/cli_common/invoke_context.py
defenderkev: I've moved the code to its own function. Happy to write unit tests but I would appreciate some pointers on where to start, as the functionality relies on whether the user running the code is/isn't logged into an AWS account. I'm not sure how to mock the response from `sts.get_caller_identity`

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +348,19 @@ def _clean_running_containers_and_related_resources(self) -> None:
         cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
         cast(RefreshableSamFunctionProvider, self._function_provider).stop_observer()
 
+    def _add_account_id_to_global(self, parameters) -> Dict|None:
+        # Attempt to get the Account ID from the current session (if any)
+        sts = boto3.client("sts")
+        try:
+            account_id = sts.get_caller_identity()["Account"]
```
mildaniel, line 355: Let's do
```suggestion
            account_id = sts.get_caller_identity().get("Account")         
```
instead to avoid throwing a `KeyError` here in the event that we get a bad response.

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +348,19 @@ def _clean_running_containers_and_related_resources(self) -> None:
         cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
         cast(RefreshableSamFunctionProvider, self._function_provider).stop_observer()
 
+    def _add_account_id_to_global(self, parameters) -> Dict|None:
```
mildaniel, line 351:
```suggestion
    def _add_account_id_to_global(self, parameters) -> Optional[Dict]:
```

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +348,19 @@ def _clean_running_containers_and_related_resources(self) -> None:
...
+        # Attempt to get the Account ID from the current session (if any)
+        sts = boto3.client("sts")
+        try:
+            account_id = sts.get_caller_identity()["Account"]            
+            if account_id:
+                if parameters is None:
+                    parameters = {}
+                parameters["AWS::AccountId"] = account_id
+        except TokenRetrievalError:
+            LOG.warning( "No current session found, using default AWS::AccountId" )
```
mildaniel, line 361: Make sure to run `make format` otherwise the linter might complain about some of the spacing here.

samcli/commands/local/cli_common/invoke_context.py
mildaniel, line 351: We probably don't need to pass in the parameters since we have access to them through `self`.

mildaniel (changes requested):

samcli/commands/local/cli_common/invoke_context.py
mildaniel: You'll want to write the tests in this file https://github.com/aws/aws-sam-cli/blob/develop/tests/unit/commands/local/cli_common/test_invoke_context.py. There should be a few examples in our unit tests that demonstrate mocking external calls. Here's on you can use as reference: https://github.com/aws/aws-sam-cli/blob/develop/tests/unit/commands/local/cli_common/test_invoke_context.py#L1355

defenderkev: I've made the suggested changes (as you probably noticed Python is not my first language!) and added unit tests to make sure the new function works both with and without a called id being returned by STS

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1360,3 +1360,20 @@ def test_must_pass_custom_region(self, get_stacks_mock):
         get_stacks_mock.assert_called_with(
             "template_file", parameter_overrides=None, global_parameter_overrides={"AWS::Region": "my-custom-region"}
         )
+
+
+class TestInvokeContext_add_account_id_to_global(TestCase):
+    def test_must_work_with_no_token(self):
+        invoke_context = InvokeContext("template_file")
+        invoke_context._add_account_id_to_global()
+        assert invoke_context._global_parameter_overrides is None
```
mildaniel, line 1369: Let's use the assertions provided through `pytest`.
```suggestion
        self.assertIsNone(invoke_context._global_parameter_overrides)
```

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1360,3 +1360,20 @@ def test_must_pass_custom_region(self, get_stacks_mock):
...
+
+    @patch("boto3.client")
+    def test_must_work_with_token(self, get_caller_identity_mock):
+        get_caller_identity_mock.return_value.get_caller_identity.return_value.get.return_value = "210987654321"
+        invoke_context = InvokeContext("template_file")
+        invoke_context._add_account_id_to_global()
+        print("\n")
+        print(invoke_context._global_parameter_overrides)
+        print("\n")
+        assert invoke_context._global_parameter_overrides.get("AWS::AccountId") is "210987654321"
```
mildaniel, line 1379: Same comment as above about the assertions

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +349,18 @@ def _clean_running_containers_and_related_resources(self) -> None:
         cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
         cast(RefreshableSamFunctionProvider, self._function_provider).stop_observer()
 
+    def _add_account_id_to_global(self):
+        # Attempt to get the Account ID from the current session (if any)
```
mildaniel, line 353: Let's replace this comment with a doctstring

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1360,3 +1360,20 @@ def test_must_pass_custom_region(self, get_stacks_mock):
...
+        invoke_context._add_account_id_to_global()
+        assert invoke_context._global_parameter_overrides is None
+
+    @patch("boto3.client")
+    def test_must_work_with_token(self, get_caller_identity_mock):
+        get_caller_identity_mock.return_value.get_caller_identity.return_value.get.return_value = "210987654321"
+        invoke_context = InvokeContext("template_file")
+        invoke_context._add_account_id_to_global()
+        print("\n")
+        print(invoke_context._global_parameter_overrides)
```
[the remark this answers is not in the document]
defenderkev: Yeah, I was trying to figure out how to mock out `get_caller_identity` and forgot to take the `print`s out. I'll fix it

tests/unit/commands/local/cli_common/test_invoke_context.py
defenderkev: done

samcli/commands/local/cli_common/invoke_context.py
defenderkev: done

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +349,22 @@ def _clean_running_containers_and_related_resources(self) -> None:
         cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
         cast(RefreshableSamFunctionProvider, self._function_provider).stop_observer()
 
+    def _add_account_id_to_global(self):
+        """
+        Attempts to get the Account ID from the current session
+        If there is no current session, the standard parameter override for
+        AWS::AccountId is used
+        """
+        sts = boto3.client("sts")
```
hawflau, line 358: please update to create a client using [`get_boto_client_provider_with_config`](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/boto_utils.py#L65)? This will ensure the client is created with provided aws profile (which is stored at `self._aws_profile`.

You can see an example of how it's used at https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/delete/delete_context.py#L79-L96

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +349,22 @@ def _clean_running_containers_and_related_resources(self) -> None:
...
+        try:
+            account_id = sts.get_caller_identity().get("Account")
+            if account_id:
+                if self._global_parameter_overrides is None:
+                    self._global_parameter_overrides = {}
+                self._global_parameter_overrides["AWS::AccountId"] = account_id
+        except TokenRetrievalError:
```
hawflau, line 365: I think the right exception type to catch is `botocore.exceptions.ClientError`, as explained [here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#aws-service-exceptions)

hawflau (changes requested):
Thanks for your contributions!
Please check my feedback

samcli/commands/local/cli_common/invoke_context.py
defenderkev: Updated

samcli/commands/local/cli_common/invoke_context.py
defenderkev: I tried using ClientError but that made all the tests fail. I think TokenRetrievalError is correct. Maybe ClientError is for failures instantiating/setting up the client.

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +349,25 @@ def _clean_running_containers_and_related_resources(self) -> None:
         cast(WarmLambdaRuntime, self.lambda_runtime).clean_running_containers_and_related_resources()
         cast(RefreshableSamFunctionProvider, self._function_provider).stop_observer()
 
+    def _add_account_id_to_global(self):
+        """
+        Attempts to get the Account ID from the current session
+        If there is no current session, the standard parameter override for
+        AWS::AccountId is used
+        """
+        client_provider = samcli.lib.utils.boto_utils.get_boto_client_provider_with_config()
```
lucashuy, line 358: Could we update the import at the top of this file to import that method instead of qualifying it here?

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -472,6 +495,7 @@ def _is_debugging(self) -> bool:
 
     def _get_stacks(self) -> List[Stack]:
         try:
+            print(SamLocalStackProvider.get_stacks)
```
lucashuy, line 498: Left over debugging extra `print` statement here, lets remove it

lucashuy (commented):
Thanks for maintaining this PR thus far, left some pretty minor comments

samcli/commands/local/cli_common/invoke_context.py
defenderkev: I couldn't get the test to patch correctly if I used `from samcli.lib.utils.boto_utils  import get_boto_client_provider_with_config`. Happy to update if someone can assist here; Python is not my first language..

samcli/commands/local/cli_common/invoke_context.py
defenderkev: Done.

samcli/commands/local/cli_common/invoke_context.py
lucashuy: Was the issue that the patch wasn't working when moving the import to the top of the file? You might need to patch it using from the `invoke_context` file instead. That is, instead of
```
@patch("samcli.lib.utils.boto_utils.get_boto_client_provider_with_config")
```
The following patch is used instead
```
@patch("samcli.commands.local.cli_common.invoke_context.get_boto_client_provider_with_config")
```

lucashuy:
It looks like the previous CI check failed for linting reasons, can you run `make format` to automatically fix them?

samcli/commands/local/cli_common/invoke_context.py
defenderkev: updated the import and the patch as requested.

defenderkev: Thanks for the help - I've done the update

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -345,6 +349,25 @@ def _clean_running_containers_and_related_resources(self) -> None:
...
+
+        try:
+            account_id = sts.get_caller_identity().get("Account")
+            if account_id:
+                if self._global_parameter_overrides is None:
+                    self._global_parameter_overrides = {}
+                self._global_parameter_overrides["AWS::AccountId"] = account_id
+        except (NoCredentialsError, TokenRetrievalError):
+            LOG.warning("No current session found, using default AWS::AccountId")
+
```
jysheng123, line 370: If this function is ok to error, can you also add a generic exception handling right now? If theres no exception handling here, if an error occurs outside of the 2 defined errors, there is potential for this function to panic the execution

jysheng123 (changes requested):
Left one comment about error handling, but otherwise this PR looks good to me

samcli/commands/local/cli_common/invoke_context.py
defenderkev: I caatch these two exceptions because they can be the consequences of
- Not having a `~/.aws/credentials` or `~/.aws/config` file
- Valid configuration but token is expired
If anything else causes an error, we don't know what's caused it and in my opinion the execution should panic because we don't know how to handle it.
Thoughts? I mean, I'm happy to put in a generic `except` and continue but I'm not sure if that's what we should do here

lucashuy: We might want to consider catching the `ClientError` exception here too. Using expired credentials, botocore returns:

```
botocore.exceptions.ClientError: An error occurred (ExpiredToken) when calling the GetCallerIdentity operation: The security token included in the request is expired
```

You had mentioned that other tests start failing once the `ClientError` was caught, maybe we could look at fixing those?

kevin-james-sp: What I've done is add catching `ClientError` to the new method, and patching the new method for the existing tests so it's just a stub; they don't need a live connection so inserting a live account ID is irrelevant. All tests are passing now

kevin-james-sp: I also put back in some definitions of `aws_profile` I had taken out of some of those tests. Now the only differences to upstream are the patching of `_add_account_id_to_global` and the extra tests for the new code.

defenderkev: @lucashuy What I've done is add catching `ClientError` to the new method, and patching the new method for the existing tests so it's just a stub; they don't need a live connection so inserting a live account ID is irrelevant. All tests are passing now
I also put back in some definitions of `aws_profile` I had taken out of some of those tests. Now the only differences to upstream are the patching of `_add_account_id_to_global` and the extra tests for the new code.

hnnasit (approved):
LGTM! Thanks for your contributions.

mildaniel (approved):

hawflau (approved):

jysheng123 (approved):

Issue #6427, opened 2023-12-09
Bug: sam sync does not have --container-env-var-file parameter


zyrorl:
### Description:
It appears that sam sync doesn't support passing in of `--container-env-var-file` when using the `--use-container` flag.

This means that when building the code it's missing any environment variables needed to compile the project.

### Steps to reproduce:
Run `sam sync --use-container --container-env-var-file container.env.json`. Variable files are not respected as per `sam build` or other commands.

### Observed result:
Container builds that require env vars have those vars unset.

### Expected result:
Environment variables are passed into container as per `sam build`

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
{
  "version": "1.104.0",
  "system": {
    "python": "3.11.3",
    "os": "Linux-5.10.201-191.748.amzn2.x86_64-x86_64-with-glibc2.26"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.25",
    "aws_cdk": "2.114.1 (build 02bbb1d)",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}```

jysheng123:
Hi, thanks for bringing up the issue. It looks like we only support --container-env-file option on `sam build` currently, but this looks like a quick feature request to implement. I will let the team know about it to see about putting it in the future.

zyrorl:
any updates on this?

nitintecg:
@jysheng123 i would like to worked on this. But, I'm facing one issue while writing unit tests for same. 
Can you share me specific doc/contributing reference guide ?

hnnasit:
Hi @nitintecg, if you are interested to contribute feel free to open a PR and our team would be happy to review and assist with unit tests.  I don't recall if we have specific guidelines on writing unit tests, but you can follow the existing unit tests to get some idea.

hnnasit:
Closing as the the PR has been merged and released.

PR #7019, opened 2024-05-04
fix: allow --container-env-var-file in sync command

nitintecg:
#6427 

It fixes this issue by allowing --container-env-var-file in sync command similar to build command

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


mildaniel (approved):
Thanks for the contribution, looks good to me!

samcli/commands/sync/command.py
```diff
@@ -158,6 +158,14 @@
...
     " by comparing the local template with the template deployed in cloud.",
 )
+@click.option(
+    "--container-env-var-file",
+    "-ef",
+    default=None,
+    type=click.Path(),  # Must be a json file
+    help="Environment variables json file (e.g., env_vars.json) to be passed to containers.",
+    cls=ContainerOptions,
+)
```
mndeveci, line 168: We usually keep common options in same place to reduce duplication. One example can be found here; https://github.com/aws/aws-sam-cli/blob/232f88acc3a1ccc2056eab4d0490ff710f4a9674/samcli/commands/_utils/options.py#L798-L813

Would you mind moving this option to there as well and use it in both places?

mndeveci (commented):
Thanks for addressing this issue! LGTM in general, left a small comment there. Let me know if you have any questions!

samcli/commands/sync/command.py
nitintecg: yes sure

samcli/commands/build/command.py
```diff
@@ -92,14 +93,7 @@
...
 )
-@click.option(
-    "--container-env-var-file",
-    "-ef",
-    default=None,
-    type=click.Path(),  # Must be a json file
-    help="Environment variables json file (e.g., env_vars.json) to be passed to build containers.",
-    cls=ContainerOptions,
-)
+@container_env_var_file_option(cls=ContainerOptions)
```
mndeveci, line 96:
```suggestion
@container_env_var_file_option
```

samcli/commands/_utils/options.py
```diff
@@ -1009,3 +1009,22 @@ def watch_exclude_click_option():
...
+        default=None,
+        type=click.Path(),  # Must be a json file
+        help="Environment variables json file (e.g., env_vars.json) to be passed to containers.",
+        cls=cls,
+    )
+
+
+@parameterized_option
+def container_env_var_file_option(f, cls):
+    return container_env_var_file_click_option(cls)(f)
```
mndeveci, line 1030: Sorry to bug you again, but we don't need to keep `ContainerOptions` as a parameter since it will be the case all the time. Can we directly use the type in the newly created function above?

I am adding my suggestion in the comments, let me know if it makes sense to you so that we can apply it quickly (we might need to add the import the `ContainerOptions` class in this file though).
```suggestion
def container_env_var_file_click_option():
    """
    Click option to --container-env-var-file option
    """
    return click.option(
        "--container-env-var-file",
        "-ef",
        default=None,
        type=click.Path(),  # Must be a json file
        help="Environment variables json file (e.g., env_vars.json) to be passed to containers.",
        cls=ContainerOptions,
    )


@parameterized_option
def container_env_var_file_option(f):
    return container_env_var_file_click_option()(f)
```

samcli/commands/sync/command.py
```diff
@@ -158,6 +159,7 @@
     help="This option will skip the initial infrastructure deployment if it is not required"
     " by comparing the local template with the template deployed in cloud.",
 )
+@container_env_var_file_option(cls=ContainerOptions)
```
mndeveci, line 162:
```suggestion
@container_env_var_file_option
```

mndeveci (changes requested):

samcli/commands/_utils/options.py
nitintecg: @mndeveci makes sense. So i had one doubt should i move `ContainerOptions` class to some common place like inside `_utils` module ?

mndeveci: @nitintecg I've checked other places and it looks like this is a pattern that we use before. So I will keep as it is now since this will touch other places.

mndeveci (approved):

PR #7064, opened 2024-05-15
chore(tests): Refactor tests to support easily invoking AL2023 based runtimes

lucashuy:

None

Our Windows testing environment does not have the latest version of Docker installed inside of it, causing issues when invoking AL2023 based runtime images. This results in skipping the AL2023 based runtime tests specifically for Docker.

Splits up the test cases that use Docker and test the new AL2023 based runtime images into two. AL2023 based tests are then marked with a `pytest` marker to make them easily toggle-able within the `pytest` command.

The two tests will be running the same testing parameters and assert conditions.

The goal at the end of this (follow up PR to address the other tests), is to remove the `runtime_supported_by_docker()` method, and to instead have two separate test methods marked by `pytest` decorators. 

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


appveyor-windows-binary.yml
```diff
@@ -49,10 +49,6 @@ init:
   - ps: New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
   - ps: git config --system core.longpaths true
 
-cache:
-  - C:\ProgramData\chocolatey\bin -> appveyor.yml
-  - C:\ProgramData\chocolatey\lib -> appveyor.yml
-
```
mndeveci, line 55: out of curiosity, why did we removed the caching part here?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -523,6 +509,11 @@ class TestBuildCommand_PythonFunctions_WithoutDocker(BuildIntegPythonBase):
     use_container = False
 
     def test_with_default_requirements(self):
+        # lucashuy: temporarily set docker check here
+        # until we refactor this set of tests
+        if IS_WINDOWS and not SKIP_DOCKER_TESTS and not runtime_supported_by_docker(self.runtime):
```
mndeveci, line 514: I thought we are getting rid of `runtime_supported_by_docker` and using `pytest.mark` instead? Do we still need this function check here?

lucashuy: This part of the code hasn't been touched yet , saving that for the next PR so that this one doesn't become too bloated in size. We'll need to keep this here for now.

appveyor-windows-binary.yml
lucashuy: This doesn't work for a couple of reasons:
1. The cache works by specifying a cached directory, and file to watch to invalidate the cache. We don't have a file named `appveyor.yml`, so the cache will never be invalidated correctly.
2. Its not enough to just watch the chocolatey directories as some tools we install via chocolatey dump their files in other locations. For example, we install JDK 21 via chocolatey, but the runtime is saved into `C:\Program Files\...`, while the binaries and chocolatey metadata is saved in the chocolatey folders. This causes a "desync" and will cause the job to fail.

We're going to disable it for now since this needs some more investigation/work before turning it back on.

tests/integration/buildcmd/test_build_cmd.py
lucashuy: Ah, this particular check might be unecessary, let me remove it since this integ test isn't suppose to require an invoke anyways.

mndeveci (approved):

jysheng123:
Have new branch: `al2023-test-changes-1` with newer changes and rebased to track this issue, so we can close this draft PR

Issue #6899, opened 2024-04-02
Add option to override resolve_s3 from samconfig


ExtremoPenguin:
### Describe your idea/feature/enhancement

If samconfig sets `resolve_s3` there is no way to override this setting using command line arguments. There should be the option to clear this flag using the command line arguments so that `--s3-bucket` can be used instead when needed.

### Proposal

Add the option to pass `--resolve-s3 false` on the CLI which would be the equivalent of not setting  `--resolve-s3` at all.

### Additional Details

We try to setup our projects so that developers can stand up our applications in their own sandbox accounts with same settings that our CI builds use when deploying to production environments. samconfig has been very useful for that as we can move most of the CLI arguments to configuration that is checked into git. Developers deploying to their test environments can then simply run `sam deploy` without needing to copy and paste a bunch of CLI arguments.

The issue we are running into is that for developer environments, using `resolve_s3` in samconfig works great and automatically detects the SourceBucket from the aws-sam-cli-managed-default stack. But in our CI pipelines, we are using the user and role generated from `sam pipeline bootstrap` which has its own artifact bucket and does not have permissions to the bucket in aws-sam-cli-managed-default. This means we need to specify `--s3-bucket` in the CI environments. But because `--resolve-s3` is a flag, it can not be unset and using `--s3-bucket` throws an error since resolve-s3 and s3-bucket are mutually exclusive.

The work around is fairly minor. Just don't include resolve_s3 in samconfig and require developers include it as a command line argument when running sam. But it would be nice if everything set in samconfig could be overridden by the CLI arguments when needed.

lucashuy:
Thanks for opening this feature request. It looks like the inverse of the flag needs to be configured (ie. `--no-resolve-s3`) so that your desired behaviour is achieved. Let me raise this with the team for prioritization and to see if this aligns with what we wanted to do with this option.

In the mean time, I'm going to mark this as a good first issue for folks that wanted to contribute to this project.

nitintecg:
@lucashuy I would like to contribute this issue. I'm having doubt that new option`--no-resolve-s3` will ignores `resolve-s3` option if its present. for eg like this in deploy `command.py`  file 
```
if not no_resolve_s3 and resolve_s3: # here means resolve_s3 will be supported when no_resolve_s3 is present
            if bool(s3_bucket):
                raise DeployResolveS3AndS3SetError()
 ```
am i right here ?

lucashuy:
[...]

PR #7107, opened 2024-05-27
fix: modified option resolve_s3 to allow override from samconfig

nitintecg:
#6899 

Modified resolve_s3 option to allow override from samconfig 

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation

### DIFF
--- a/tests/integration/deploy/test_deploy_command.py
+++ b/tests/integration/deploy/test_deploy_command.py
@@ -423,6 +423,39 @@ to create a managed default bucket, or run sam deploy --guided",
             deploy_process_execute.stderr,
         )
 
+    def test_deploy_with_no_resolve_s3_option(self):
+        template_path = self.test_data_path.joinpath("aws-serverless-function.yaml")
+
+        stack_name = self._method_to_stack_name(self.id())
+
+        deploy_command_list = self.get_deploy_command_list(
+            template_file=template_path,
+            stack_name=stack_name,
+            capabilities="CAPABILITY_IAM",
+            s3_prefix=self.s3_prefix,
+            force_upload=True,
+            notification_arns=self.sns_arn,
+            parameter_overrides="Parameter=Clarity",
+            kms_key_id=self.kms_key,
+            no_execute_changeset=False,
+            tags="integ=true clarity=yes foo_bar=baz",
+            confirm_changeset=False,
+        )
+
+        deploy_command_list.append("--no-resolve-s3")
+
+        deploy_process_execute = self.run_command(deploy_command_list)
+        # Error asking for s3 bucket
+        self.assertEqual(deploy_process_execute.process.returncode, 1)
+        self.assertIn(
+            bytes(
+                f"S3 Bucket not specified, use --s3-bucket to specify a bucket name, or use --resolve-s3 \
+to create a managed default bucket, or run sam deploy --guided",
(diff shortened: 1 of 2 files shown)


nitintecg:
@lucashuy  can you help me in writing/reviewing and running  the tests files ?

lucashuy:
Hi, thanks for opening the PR. For testing, we have two kinds of test: unit tests and integration tests.

You can run unit tests on your own without any AWS deployments by running `make test`. This will take a moment to run. We likely don't need to add anything here, but rather making sure that anything that breaks is fixed.

For integration tests, we can add a test similar to this: https://github.com/aws/aws-sam-cli/blob/e0bc44db64c9e81ea25521ed5cbd8fc8d2245c9c/tests/integration/deploy/test_deploy_command.py#L309-L325

So instead of adding a new check in the base class for `--no-resolve-s3`, we can just directly append it to the end of the generated command.

tests/integration/deploy/test_deploy_command.py
```diff
@@ -423,6 +423,30 @@ def test_deploy_without_s3_bucket(self, template_file):
...
+            notification_arns=self.sns_arn,
+            parameter_overrides="Parameter=Clarity",
+            kms_key_id=self.kms_key,
+            no_execute_changeset=False,
+            tags="integ=true clarity=yes foo_bar=baz",
+            confirm_changeset=False,
+        )
+
+        deploy_process_execute = self.run_command(deploy_command_list)
+        self.assertEqual(deploy_process_execute.process.returncode, 0)
```
lucashuy, line 448: building from my top level comment, we can check for the error message about missing bucket instead, like this test: https://github.com/aws/aws-sam-cli/blob/e0bc44db64c9e81ea25521ed5cbd8fc8d2245c9c/tests/integration/deploy/test_deploy_command.py#L393-L424

tests/integration/deploy/test_deploy_command.py
```diff
@@ -423,6 +423,40 @@ def test_deploy_without_s3_bucket(self, template_file):
             deploy_process_execute.stderr,
         )
 
+    @parameterized.expand(["aws-serverless-function.yaml", "cdk_v1_synthesized_template_zip_functions.json"])
```
lucashuy, line 426: We can probably just test against a single template since its an expected failure with a CLI option. The action here is to simply remove the `@parameterized.expand`, removing the `template_file` in the arguments, and changing `template_path = self.test_data_path.joinpath(template_file)` to just use `aws-serverless-function.yaml` instead.

This way, we only end up using that single template to validate the new command line argument.

lucashuy (commented):
Thanks for making the changes! Left a minor comment, I'll also start the Github CI runners since those run `make pr` on a couple of different platforms and Python version combinations

mildaniel (commented):
LGTM, will approve when @lucashuy's comments are addressed.

tests/integration/deploy/test_deploy_command.py
nitintecg: sure

lucashuy (approved):
Thanks for following up with your changes! I've run the integration test and validated that it is running correctly. LGTM

mildaniel (approved):

PR #7108, opened 2024-05-28
fix(tests): Keep original test data path in its own variable to avoid overwriting

lucashuy:

None.

For some of our integration tests, it looks we overwrite the testing data path with a new testing path. This won't work if the test failed and reran, since it'll replace the old temp folder, with a new empty temp folder, and attempt to copy the contents of the old temp folder into the new one.

The proper behaviour should be to create a new temp folder, and to copy the originally determined test data folder into this new temp folder.

Save the path of the original test data folder in it's own class variable.

None

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/deploy/deploy_integ_base.py
+++ b/tests/integration/deploy/deploy_integ_base.py
@@ -2,7 +2,7 @@ import shutil
 import tempfile
 from pathlib import Path
 from enum import Enum, auto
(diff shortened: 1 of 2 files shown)


lucashuy:
Superseded by #7109

Issue #6910, opened 2024-04-05
Token Authorizer doesn't get linked if FunctionPayloadType missing in template.yml 


Tolsto:
There is an issue where Token Authorizers are not being linked correctly if the FunctionPayloadType is missing in the template.yml file. This is due to the default value for `authorizer_type` being initialized as `token`, but then compared against `TOKEN` in a case-sensitive manner.

https://github.com/aws/aws-sam-cli/blob/c3e154763d758e17ffa86eb7d05ddcc9b27a095e/samcli/lib/providers/sam_api_provider.py#L321-L324

lucashuy:
Thanks for reporting this, we'll prioritize a fix for this within the team's capacity.

lucashuy:
This fix has been released as part of the [SAM CLI 1.119.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.119.0) release. Closing issue.

PR #7148, opened 2024-06-07
fix: Updated default value of authorizer type to be upper case TOKEN

hnnasit:

https://github.com/aws/aws-sam-cli/issues/6910

Token authorizers are not being linked if `FunctionPayloadType` value is not being set. The default value is set to `token` but the check to link is checking against `TOKEN`.

Allows linking authorizers when `FunctionPayloadType` is not set.

No side effects.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/sam_api_provider.py
```diff
@@ -299,7 +299,7 @@ def _extract_authorizers_from_props(logical_id: str, auth: dict, collector: ApiC
         authorizers: Dict[str, Authorizer] = {}
 
         for auth_name, auth_props in auth.get(SamApiProvider._AUTHORIZERS, {}).items():
-            authorizer_type = auth_props.get(SamApiProvider._FUNCTION_TYPE, LambdaAuthorizer.TOKEN)
+            authorizer_type = auth_props.get(SamApiProvider._FUNCTION_TYPE, LambdaAuthorizer.TOKEN.upper())
```
mildaniel, line 302: Is this constant used in other places where it needs to be lowercase?

jysheng123: same comment as daniel other than that LGTM

jysheng123 (approved):

samcli/lib/providers/sam_api_provider.py
hnnasit: Nope, `authorizer_type` is only used to check if the authorizer can be extracted. There isn't any other occurrence in that function.

PR #7160, opened 2024-06-12
chore: bump version to 1.119.0

lucashuy:

### DIFF
--- a/samcli/__init__.py
+++ b/samcli/__init__.py
@@ -2,4 +2,4 @@
 SAM CLI version
 """
 
-__version__ = "1.119"
+__version__ = "1.119.0"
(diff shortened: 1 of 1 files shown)


samcli/__init__.py
```diff
@@ -2,4 +2,4 @@
 SAM CLI version
 """
 
-__version__ = "1.118.0"
+__version__ = "1.119"
```
lucashuy, line 5:
```suggestion
__version__ = "1.119.0"
```

jysheng123: do we not need the `119.0`?

jysheng123 (approved):

samcli/__init__.py
lucashuy: Updated

PR #7161, opened 2024-06-12
fix(tests): Fix local test failing for account ID checks

lucashuy:

None

A recent change made to the local command behaviour causes some of the integration tests to fail. This is due to the local integration tests never expecting there to be an actual valid account ID, however this is no longer the case.

Checks if the account ID looks and smells like an account ID, instead of checking for an exact value.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/local/invoke/invoke_integ_base.py
+++ b/tests/integration/local/invoke/invoke_integ_base.py
@@ -131,7 +131,7 @@ class InvokeIntegBase(TestCase):
             raise
 
 
-class IntegrationIntegBase(InvokeIntegBase):
+class IntegrationCliIntegBase(InvokeIntegBase):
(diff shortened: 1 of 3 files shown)


hnnasit (approved):

tests/integration/local/invoke/invoke_integ_base.py
```diff
@@ -129,3 +129,14 @@ def run_command(self, command_list, env=None, cwd=None):
         except TimeoutExpired:
             process.kill()
             raise
+
+
+class IntegrationIntegBase(InvokeIntegBase):
```
hnnasit, line 134: nit: Can we update the naming to add CLI as well?

hawflau (approved):

PR #7162, opened 2024-06-13
feat: Add wait timeout for deploy command

### LINKED ISSUE #4478: How do I increase the time limit for sam deploy?
I believe that the time limit for `sam deploy` is 60 minutes.

I don't see it documented, but I see it in the code of the `wait_for_execute` function.

https://github.com/aws/aws-sam-cli/blob/fc9653313ac0d35e995b80ed8215f5587e20d16a/samcli/lib/deploy/deployer.py#L459-L510

The `SAM_CLI_POLL_DELAY` is documented, but it doesn't seem to be referenced by this function.

I have a stack that often takes more than 1 hour to complete an update (it contains multiple stack sets across many accounts and regions). How can I have sam wait for it to complete?

cullenwatson:
* closes #4478


jysheng123 (commented):
Would you be able to make a unit test where the max_wait_duration is really low like 5 seconds and test that it goes through?

cullenwatson:
> Would you be able to make a unit test where the max_wait_duration is really low like 5 seconds and test that it goes through?

It is in minutes. Is there a way we could test one minute timeout?

jysheng123:
Hmm, yeah on second thought, this seems fine to me to test since the previous hard coded value was 120 seconds anyways,  maybe a NIT that im thinking of is the possibility of the sam deploy time outing for an in definite amount of time, so maybe something like setting a reasonable upper limit like 10 minutes to me.

cullenwatson:
ok i just put min of 60 minutes and max 480 minutes as 480 was what the op wanted

hawflau:
Hi @cullenwatson could you please check the failed checks and fix accordingly? Seems like there's some linting issues, and you also need to run `make schema` to update the SAM CLI command schema as you added a new command option

lucashuy:
Hi @cullenwatson, just wanted to check in to see if you had any blockers or help wanted to continue contributing.

cullenwatson:
ok the tests are passing now besides the 2 that are failing also in main

jysheng123 (approved):

mildaniel (approved):

iainelder:
@cullenwatson , thanks for this improvement!

I'm no longer working on the project that needed SAM to run longer than 1 hour, so I'm no longer in a position to test this.

I've passed on the good news anyway.

PR #7251, opened 2024-07-16
chore(tests): Refactor tests to support easily invoking AL2023 based runtimes

jysheng123:

None

Our Windows testing environment does not have the latest version of Docker installed inside of it, causing issues when invoking AL2023 based runtime images. This results in skipping the AL2023 based runtime tests specifically for Docker.

Splits up the test cases that use Docker and test the new AL2023 based runtime images into two. AL2023 based tests are then marked with a `pytest` marker to make them easily toggle-able within the `pytest` command.

The two tests will be running the same testing parameters and assert conditions.

The goal is to remove the `runtime_supported_by_docker()` method, and to instead have two separate test methods marked by `pytest` decorators. 

None.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


jysheng123:
Note: Lots of lines are changed but most of them are separating test cases into files separated by runtimes, so a lot of lines are reused

appveyor-windows-al2023.yml
```diff
@@ -0,0 +1,181 @@
...
+  # Switch to Docker Linux containers
+  - ps: '& $Env:ProgramFiles\Docker\Docker\DockerCli.exe -SwitchLinuxEngine'
+  - "docker info"
+  - "docker version"
+
+  # Upgrade setuptools, wheel and virtualenv
+  - "python -m pip install --upgrade setuptools wheel virtualenv"
+  # Install pip for the python versions which is used by the tests
+  - "C:\\Python39-x64\\python.exe -m pip install --upgrade pip"
+  - "C:\\Python310-x64\\python.exe -m pip install --upgrade pip"
```
lucashuy, line 71: Did we still need these two lines if we are just running 3.12 tests?

tests/integration/buildcmd/test_build_cmd_arm64.py
```diff
@@ -36,22 +36,28 @@ class TestBuildCommand_PythonFunctions_With_Specified_Architecture_arm64(BuildIn
             ("python3.9", "Python", False),
             ("python3.10", "Python", False),
             ("python3.11", "Python", False),
-            ("python3.12", "Python", False),
```
lucashuy, line 39: Does the test also perform an invoke? If not, then we can leave the ones with `use_container=False` in the old AppVeyor jobs. Same comment for the other AL2023 based runtimes that had that container `use_container` check

tests/integration/buildcmd/test_build_cmd_node.py
```diff
@@ -0,0 +1,170 @@
...
+        ("esbuild_templates/template_with_metadata_node_options.yaml",),
+        ("esbuild_templates/template_with_metadata_global_node_options.yaml",),
+    ],
+)
+class TestBuildCommand_EsbuildFunctionProperties(BuildIntegEsbuildBase):
+    @parameterized.expand(
+        [
+            ("nodejs16.x", "../Esbuild/TypeScript", "app.lambdaHandler", "x86_64"),
+            ("nodejs18.x", "../Esbuild/TypeScript", "app.lambdaHandler", "x86_64"),
+            ("nodejs20.x", "../Esbuild/TypeScript", "app.lambdaHandler", "x86_64"),
```
lucashuy, line 133: Do we need to split the node20 esbuild tests out here too?

tests/integration/buildcmd/test_build_cmd_python.py
```diff
@@ -0,0 +1,485 @@
...
+        "template",
+        "FUNCTION_LOGICAL_ID",
+        "overrides",
+        "runtime",
+        "codeuri",
+        "check_function_only",
+        "prop",
+    ),
+    [
+        ("template.yaml", "Function", True, "python3.8", "Python", False, "CodeUri"),
```
lucashuy, line 269: nit: we can clean this up, some of the parameterized arguments are the same in all test cases

lucashuy (commented):
Thanks for making the changes. I made a first pass and left some comments

mndeveci:
Why build tests are failing in Github Actions?

appveyor-windows-binary.yml
```diff
@@ -49,10 +49,6 @@ init:
   - ps: New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
   - ps: git config --system core.longpaths true
 
-cache:
-  - C:\ProgramData\chocolatey\bin -> appveyor.yml
-  - C:\ProgramData\chocolatey\lib -> appveyor.yml
-
```
mndeveci, line 55: Curious, why did we remove these caches? Are they causing any issues?

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -492,389 +263,36 @@ def _validate_skipped_built_function(
...
-class TestBuildCommand_RubyFunctions(BuildIntegRubyBase):
-    @parameterized.expand(["ruby3.2", "ruby3.3"])
+    @parameterized.expand([("ruby3.3",)])
     @skipIf(SKIP_DOCKER_TESTS or SKIP_DOCKER_BUILD, SKIP_DOCKER_MESSAGE)
-    def test_building_ruby_in_container(self, runtime):
-        if IS_WINDOWS and not runtime_supported_by_docker(runtime):
-            self.skipTest(RUNTIME_NOT_SUPPORTED_BY_DOCKER_MSG)
+    @pytest.mark.al2023
+    def test_building_ruby_al2023(self, runtime):
         self._test_with_default_gemfile(runtime, "use_container", "Ruby", self.test_data_path)
```
mndeveci, line 279: Aren't we removing `ruby3.2` from this tests?

jysheng123: No, we since we only have 1 runtime in the old non-al2023 tests, the test_building_ruby_3_2 just directly calls ruby3.2 in the runtime so it is still the same test, just not parametrized

appveyor-windows-binary.yml
jysheng123: Im copying @lucashuy words in his old PR since this a change from back then: 
This doesn't work for a couple of reasons:

1. The cache works by specifying a cached directory, and file to watch to invalidate the cache. We don't have a file named appveyor.yml, so the cache will never be invalidated correctly.
2. Its not enough to just watch the chocolatey directories as some tools we install via chocolatey dump their files in other locations. For example, we install JDK 21 via chocolatey, but the runtime is saved into C:\Program Files\..., while the binaries and chocolatey metadata is saved in the chocolatey folders. This causes a "desync" and will cause the job to fail.
We're going to disable it for now since this needs some more investigation/work before turning it back on.

(https://github.com/aws/aws-sam-cli/pull/7064#discussion_r1604001078)

mndeveci (approved):

lucashuy (approved):

PR #7284, opened 2024-07-26
fix: change openssl url in build linux 

sidhujus:

the url now redirects to github which redirects to the release artifact, this changes the url we use to be the one before the final redirect

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


mildaniel (approved):

installer/pyinstaller/build-linux.sh
```diff
@@ -49,7 +49,7 @@ mkdir -p .build/output/openssl
 cd .build/output/openssl
 
 echo "Building OpenSSL"
-curl "https://www.openssl.org/source/openssl-${openssl_version}.tar.gz" --output openssl.tar.gz
+curl -L "https://github.com:443/openssl/openssl/releases/download/OpenSSL_${openssl_version//./_}/openssl-${openssl_version}.tar.gz" --output openssl.tar.gz
```
lucashuy, line 52: We can probably remove the port here. Not that I'd imagine Github would open their webserver on anything other than port 443, but its not required.

hawflau (approved):

lucashuy (approved):

mndeveci:
Btw we could have added this flag for other places as well. Like this one; https://github.com/aws/aws-sam-cli/blob/develop/installer/pyinstaller/build-linux.sh#L52

PR #7306, opened 2024-08-02
fix: Remove str type constrain in Fn:FindInMap to avoid the issue of …

dkphm:
Remove the restriction on `str` type when `Fn:FindInMap` is called.

#6975 

To allow customer to start lambda locally.

Customer cannot start lambda locally if the yaml template include a function `Fn:FindInMap` inside a `timeout` property.

N/A

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
+++ b/samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
@@ -492,7 +492,7 @@ class IntrinsicResolver:
         )
 
         second_level_value = top_level_value.get(second_level_key)
-        verify_intrinsic_type_str(
+        verify_non_null(
             second_level_value,
             IntrinsicResolver.FN_FIND_IN_MAP,
             message="The SecondLevelKey is missing in the Mappings dictionary in Fn::FindInMap  "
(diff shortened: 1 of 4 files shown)


samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
```diff
@@ -492,7 +492,7 @@ def handle_find_in_map(self, intrinsic_value, ignore_errors):
         )
 
         second_level_value = top_level_value.get(second_level_key)
-        verify_intrinsic_type_str(
+        verify_non_null(
```
mndeveci, line 495: This is changing the general intrinsic resolver functionality, which might have other side affects. Can you explain how you tested this change? Also, I couldn't find any tests to validate this change.

mndeveci (changes requested):

samcli/lib/intrinsic_resolver/intrinsic_property_resolver.py
dkphm: Sorry I forgot to add the unit test, will do it now. Thanks!

mndeveci: Thanks for adding those tests!

One final comment (at least for now) what happens if the key doesn't exist in the map? What would be the difference between previous and new behavior there?

dkphm: Thanks for the feedback!

There is no change with the behavior of `Fn:FindInMap` when the key doesn't exist in the map. The difference is when the value is not in String format, it won't return empty value anymore but instead will resolve into a value.

hawflau (approved):

mndeveci (approved):

Issue #7236, opened 2024-07-09
Getting `ValueError: Function name is required` when running API Gateway from Terraform


sevetseh28:
Hi everyone. I'm running my API Gateway locally with:

`sam local start-api --hook-name terraform --disable-authorizer`

The API Gateway starts without issues ,but when hitting any endpoint I get the following:

```python
Traceback (most recent call last):                                                                                                     
  File "flask/app.py", line 1473, in wsgi_app                                                                                          
  File "flask/app.py", line 882, in full_dispatch_request                                                                              
  File "flask/app.py", line 880, in full_dispatch_request                                                                              
  File "flask/app.py", line 865, in dispatch_request                                                                                   
  File "samcli/local/apigw/local_apigw_service.py", line 725, in _request_handler                                                      
  File "samcli/local/apigw/local_apigw_service.py", line 618, in _invoke_lambda_function                                               
  File "samcli/commands/local/lib/local_lambda.py", line 118, in invoke                                                                
  File "samcli/lib/providers/sam_function_provider.py", line 122, in get                                                               
ValueError: Function name is required
```

I ran the above with and without the authoriser, but I've also tried running the authoriser individually with:
`sam local invoke ModuleIdeApiAwsLambdaFunctionAuthorizerFunction836CD508 --hook-name terraform --skip-prepare-infra --event authoriser-event.json`

and it works well when running the Lambda function alone. Any ideas what could be wrong? 

Thank you!

lucashuy:
Can you provide a sample project for us to reproduce this issue? The bounds of what we support with `sam local` is limited, and we may not support your use case (ie. APIGW integration with a state machine), but we'll be able to help further with a project.

sevetseh28:
hey @lucashuy thanks for your reply. I ended up changing my integration type to Lambda proxy integration for more flexibility and endpoint customisation. 
I was initially using Step Functions integration but sam doesn't seem to support it, I read the source code in `sam_function_provider.py` and it explicitly mentions this in a doctoring.

lucashuy:
Our support for Lambda invocations locally is limited to directly invoking the Lambda (through `sam local invoke` and `sam local start-lambda`), and with the API Gateway Lambda integrations.

Unfortunately we don't support the local invocations of Lambda through other integration means with API Gateway.

[...]

PR #7392, opened 2024-08-21
fix: Avoid crashing on invalid function name

mildaniel:

https://github.com/aws/aws-sam-cli/issues/7236

Avoid SAM CLI hard-crashing when it is unable to invoke a Lambda function due to an invalid function name.

Uses a custom-exception and handle it during function invocation.

N/A

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/providers/exceptions.py
+++ b/samcli/lib/providers/exceptions.py
@@ -83,3 +83,12 @@ class MissingLocalDefinition(Exception):
     @property
     def property_name(self) -> str:
         return self._property_name
+
+
+class MissingFunctionNameException(Exception):
+    """
+    Exception when a resource does not have function name specified
+    """
+
+    def __init__(self) -> None:
+        super().__init__("Unable to get Lambda function because the function identifier is not defined.")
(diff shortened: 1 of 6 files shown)


hnnasit (approved):

samcli/lib/providers/exceptions.py
```diff
@@ -83,3 +83,12 @@ def resource_identifier(self) -> "ResourceIdentifier":
     @property
     def property_name(self) -> str:
         return self._property_name
+
+
+class MissingFunctionNameException(Exception):
```
lucashuy, line 88: Would this be a `UserException`, instead of a general `Exception`? This way it is returned as exit code `1` instead of `255`

mildaniel: `UserException` is part of the interface and shouldn't be used in the `lib/`. This exception isn't propagated to the user, it's handled by the caller of invoke and a service response is chosen accordingly.

lucashuy (approved):

Issue #7398, opened 2024-08-23
Feature request: Prune Old Versions


driverpt:
### Describe your idea/feature/enhancement
We should be able to tell SAM CLI to delete AWS Lambda Older Versions (than the ones being deleted) to prevent Resource Excessive Usage.

### Proposal
Add a property like:
```
DeleteOlderVersions: true
```

Which will set the `RetainPolicy` of the Version to `Delete`

We use SnapStart, so we are somehow "protected" against bad versions.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)

driverpt:
Do you have any suggestions on where to begin?

vicheey:
The [PR: Adds support to delete old versions when successful deployment #7414](https://github.com/aws/aws-sam-cli/pull/7414) enables SAM CLI to automatically clean up previous Lambda function versions after publishing a new version and updating the alias. This cleanup functionality only takes effect when developers use the sam sync command during development to update Lambda function resources.

_**Would it be valuable to add a flag in SAM templates that modifies the DeletionPolicy for function versions?**_ 
This would enable CloudFormation deployments to automatically clean up old versions, creating consistent 
behavior between sam sync and regular CloudFormation deployments."

driverpt:
Hey, the purpose of my PR was to actually cleanup old versions on `sam deploy`. The goal is to take advantage of SnapStart and have some sort of Blue-Green Deployment where I don't need to have a separate function to cleanup older versions.

PR #7414, opened 2024-08-27
Adds support to delete old versions when successful deployment

driverpt:
#7398

Because there's a limit on how many versions a Lambda has.

Automatically deletes old version. Very useful when on SnapStart

None, it's disabled by default

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


jysheng123:
Hi, Thanks for bringing up this PR. The code looks good to me so Ill run the approval workflow to get the testing process started, but can you also create an integration tests testing the happy path from beginning to end?

jysheng123:
Can you also run `make black` to fix the formatting changes?

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -90,13 +93,18 @@ def compare_remote(self) -> bool:
...
         function_physical_id = self.get_physical_id(self._function_identifier)
+        current_alias_version = self._get_version_alias_if_exists()
         version = self._lambda_client.publish_version(FunctionName=function_physical_id).get("Version")
         self._local_sha = str_checksum(str(version), hashlib.sha256())
         LOG.debug("%sCreated new function version: %s", self.log_prefix, version)
         if version:
             self._lambda_client.update_alias(
                 FunctionName=function_physical_id, Name=self._alias_name, FunctionVersion=version
             )
+            if self._delete_old_alias and current_alias_version:
```
lucashuy, line 104: Can we add a log message here? This is deleting something, so we'll want to let people know that an old version is now being deleted.

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -90,13 +93,18 @@ def compare_remote(self) -> bool:
...
+                function_name_w_version = "{}:{}".format(function_physical_id, current_alias_version)
```
lucashuy, line 105: nit: changing `_w_` to `_with_`, or renaming to something like `formatted_function_name_version` for future contributors that are not familiar with the "with" shorthand

lucashuy (commented):
Thanks for opening this PR, left some comments. Additionally, we might want to consider adding this to the `sam deploy` command. This might need some more thought though if this makes sense to you and the team.

lucashuy:
Hi @driverpt, just wanted to check in about the state of this PR and if you had the capacity to work on it or if there was anything you needed to move forward

lucashuy:
I've chatted with some team members, and the `Metadata` section is likely a better location, than inside of `Properties`. This won't require a spec change, and will only impact SAM CLI as the `Metadata` doesn't impact SAM transform. Having a CLI option was also brought up, but using an option here might make the UX a little messy in terms of deciding how we want to pass in function names into the option.

mndeveci:
Hey @driverpt looking at the discussion here, I saw that recommendation is to use `Metadata` of the resources. Please let us know when you can follow-up with those changes, so that we can continue the review.

Thanks!

hnnasit:
Hi @driverpt, following up to check if you were planning to make the above requested changes and had any questions the team could help with.

driverpt:
Hello. Haven't yet had the time to implement this. I'll probably take a look at it this weekend

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -90,13 +93,18 @@ def compare_remote(self) -> bool:
...
+                self._lambda_client.delete_function(FunctionName=function_name_w_version)
```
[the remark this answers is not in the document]
driverpt: No, because there is no current alias. It only deletes old alias if there is an existing alias

samcli/lib/sync/flows/alias_version_sync_flow.py
```diff
@@ -107,3 +115,13 @@ def _get_resource_api_calls(self) -> List[ResourceAPICall]:
...
         """Combination of function identifier and alias name can used to identify each unique SyncFlow"""
         return self._function_identifier, self._alias_name
+
+    def _get_version_alias_if_exists(self) -> Optional[str]:
+        try:
+            return str(
+                self._lambda_client.get_alias(
+                    FunctionName=self.get_physical_id(self._function_identifier), Name=self._alias_name
+                ).get("FunctionVersion")
+            )
```
mndeveci, line 125: Wrapping `.get` method response with `str` will cause `None` value to be returned as `"None"` (string), so the IF statement above will not work as intended.
```suggestion
            return self._lambda_client.get_alias(
                FunctionName=self.get_physical_id(self._function_identifier), Name=self._alias_name
            ).get("FunctionVersion")
```

mndeveci (commented):
Thanks for your contribution, left a comment about one method response where you wrapped it up with `str` method.

samcli/lib/sync/flows/alias_version_sync_flow.py
vicheey: Hi @driverpt, thank you for your contribution. 
We would like to circle back to the suggestion. Feel free to mention any of us here after you get a change to implement the change.

driverpt: Hello,

What do you suggest to do here? I'm not expert in Python

roger-zhangg:
Seems some lint issue, please help to run `make pr` locally and resolve outstanding issue. Thanks!

mbfreder:
The PR is still failing some checks. Looks like a formatting issue. I think you should be able to solve it by running ```make format```. @driverpt could you please help to that? Thanks

driverpt:
Done, can you please re-review ?

vicheey:
Thank you. I am taking a look.

vicheey:
Everything looks good. The only lingering point is the the naming that I'll discuss with the team members. I'll get back to you soon. Once again, thank you for your amazing contribution.

PR #7422, opened 2024-08-29
fix: Race conditions with multiple downloads on the same layer

dkphm:
Customer reports that when the SAM template contains multiple functions, some shared the same lambda layer, `local start-api` throw the exception

```
2024-08-27 09:43:43,083 | Lambda functions containers initialization failed because of [Errno 2] No such file or directory: '/Users/aguro/.aws-sam/layers-pkg/NFTUserLayer-3-ff9c237555.zip'
```

This is because there was a race condition when multiple layer_downloaders were running to build multiple functions concurrently. Some could finish downloading the file first and remove the .zip file, causing the other one failed to unzip.

The fix is to ask each downloader to download into separate zip file (by adding the thread id to the file name), but the extracted folder remains the same.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/layers/layer_downloader.py
+++ b/samcli/local/layers/layer_downloader.py
@@ -3,7 +3,7 @@ Downloads Layers locally
 """
 
 import logging
-import threading
+import uuid
 from pathlib import Path
 from typing import List
 
@@ -107,8 +107,7 @@ class LayerDownloader:
             LOG.info("%s is already cached. Skipping download", layer.arn)
             return layer
 
-        current_thread_id = str(threading.get_ident())
-        layer_zip_path = f"{layer.codeuri}_{current_thread_id}.zip"
+        layer_zip_path = f"{layer.codeuri}_{uuid.uuid4().hex}.zip"
         layer_zip_uri = self._fetch_layer_uri(layer)
         unzip_from_uri(
             layer_zip_uri,
--- a/samcli/local/layers/layer_downloader.py
+++ b/samcli/local/layers/layer_downloader.py
@@ -108,7 +108,7 @@ class LayerDownloader:
             return layer
 
         current_thread_id = str(threading.get_ident())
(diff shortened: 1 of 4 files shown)


samcli/local/layers/layer_downloader.py
```diff
@@ -106,7 +107,8 @@ def download(self, layer: LayerVersion, force=False) -> LayerVersion:
             LOG.info("%s is already cached. Skipping download", layer.arn)
             return layer
 
-        layer_zip_path = layer.codeuri + ".zip"
+        current_thread_id = str(threading.get_ident())
+        layer_zip_path = f"{layer.codeuri}_{current_thread_id}.zip"
```
mndeveci, line 111: If we are looking for getting a unique value, why we are not using `uuid`?

dkphm: Oh I was thinking about one file per thread, but yeah using `uuid` is also good for this case.

mndeveci (approved):

samcli/local/layers/layer_downloader.py
```diff
@@ -106,7 +107,7 @@ def download(self, layer: LayerVersion, force=False) -> LayerVersion:
             LOG.info("%s is already cached. Skipping download", layer.arn)
             return layer
 
-        layer_zip_path = layer.codeuri + ".zip"
+        layer_zip_path = f"{layer.codeuri}_{uuid.uuid4().hex}.zip"
         layer_zip_uri = self._fetch_layer_uri(layer)
         unzip_from_uri(
```
lucashuy, line 112: quick question, but will we come across any issues when the Layer is unpacked into the same directory across different threads?

dkphm: Oh it should be fine. This is what we want :)

lucashuy (approved):

PR #7457, opened 2024-09-09
chore: bump python pysintaller versions

jysheng123:
Because of CVE from https://www.cve.org/CVERecord?id=CVE-2024-45490, we need to bump python versions of pyinstaller to latest version

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/installer/pyinstaller/build-linux.sh
+++ b/installer/pyinstaller/build-linux.sh
@@ -12,7 +12,7 @@ if [ "$python_library_zip_filename" = "" ]; then
 fi
 
 if [ "$python_version" = "" ]; then
-    python_version="3.11.8";
+    python_version="3.11.10";
 fi
 
 if [ "$openssl_version" = "" ]; then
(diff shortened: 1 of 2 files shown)


installer/pyinstaller/build-linux.sh
```diff
@@ -12,7 +12,7 @@ if [ "$python_library_zip_filename" = "" ]; then
 fi
 
 if [ "$python_version" = "" ]; then
-    python_version="3.11.8";
+    python_version="3.11.10";
```
hnnasit, line 15: Should we also bump the python versions in GH actions?

hnnasit: Looks like we don't pin [them](https://github.com/aws/aws-sam-cli/blob/develop/.github/workflows/validate_pyinstaller.yml), so should be alright.

hnnasit (approved):

installer/pyinstaller/build-linux.sh
jysheng123: I thought for GH actions, you can't specify the minor version, just the major version, hopefully it will just use the latest

lucashuy (approved):

mndeveci (approved):

Issue #7408, opened 2024-08-26
Bug: `sam init` error: "could not clone repo" (non-zero exit status 128)


mcalligator:
### Description:
When I run `sam init`, it fails to clone the `aws/aws-sam-cli-app-templates` repo, and thus does not properly initialise the project directory.  This happens regardless of which Quick Start Application Template I select.

### Steps to reproduce:
1. Install latest version of SAM CLI (1.123.0).
2. Run `sam init`:
```
Which template source would you like to use?
        1 - AWS Quick Start Templates
        2 - Custom Template Location
Choice: 1

Choose an AWS Quick Start application template
        1 - Hello World Example
        2 - Data processing
...
        15 - DynamoDB Example
        16 - Machine Learning
Template: 1

Use the most popular runtime and package type? (Python and zip) [y/N]: y

Would you like to enable X-Ray tracing on the function(s) in your application?  [y/N]: n

Would you like to enable monitoring using CloudWatch Application Insights? [y/N]: n

Would you like to set Structured Logging in JSON format on your Lambda functions?  [y/N]: n

Project name [sam-app]: sam-app
```
I did a thorough search of existing issues (closed and open), and found none closely matching this one.  I might be doing something wrong, but for now, I've designated it as a bug.

### Observed result:
```
2024-08-26 21:05:54,743 | Cache directory does not yet exist, creating one.
2024-08-26 21:05:54,749 |
Cloning from https://github.com/aws/aws-sam-cli-app-templates (process may take a moment)
2024-08-26 21:05:54,760 | WARN: Could not clone repo https://github.com/aws/aws-sam-cli-app-templates
Traceback (most recent call last):
  File "samcli/lib/utils/git_repo.py", line 149, in clone
  File "subprocess.py", line 466, in check_output
  File "subprocess.py", line 571, in run
subprocess.CalledProcessError: Command '['git', 'clone', 'https://github.com/aws/aws-sam-cli-app-templates',
'aws-sam-cli-app-templates']' returned non-zero exit status 128.
2024-08-26 21:05:54,765 | Clone error, attempting to use an old clone from a previous run

    -----------------------
    Generating application:
    -----------------------
    Name: sam-son
    Runtime: python3.9
    Architectures: x86_64
    Dependency Manager: pip
    Application Template: hello-world
    Output Directory: .
    Configuration file: sam-son/samconfig.toml

    Next steps can be found in the README file at sam-son/README.md

Commands you can use next
=========================
[...]

PR #7566, opened 2024-10-14
fix: Update to OpenSSL 3.3.1 to avoid the issue with subprocess.check…

dkphm:
…_output

#7408

To address the issue of `git clone` return status 128 with `sam init` running on some Linux distributions.

The issue happened because older version of OpenSSL 3 interfered with Python's subprocess PIPE output, causing `git clone` exited with an error.

Newer version of OpenSSL 3 might cause other issues.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/installer/pyinstaller/build-linux.sh
+++ b/installer/pyinstaller/build-linux.sh
@@ -16,7 +16,7 @@ if [ "$python_version" = "" ]; then
 fi
 
 if [ "$openssl_version" = "" ]; then
-    openssl_version="3.0.15";
+    openssl_version="3.3.1";
 fi
 
 if [ "$zlib_version" = "" ]; then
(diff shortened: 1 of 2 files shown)


mndeveci (approved):

installer/pyinstaller/build-linux.sh
```diff
@@ -16,7 +16,7 @@ if [ "$python_version" = "" ]; then
 fi
 
 if [ "$openssl_version" = "" ]; then
-    openssl_version="3.0.15";
+    openssl_version="3.3.1";
```
hnnasit, line 19: Should we also make sure there's no CVE's detected in the new version being used? We can also monitor that once this PR is merged.

mndeveci: Good point, it seems like there is no report about this version yet (only some fixes). https://openssl-library.org/news/vulnerabilities/index.html

I agree that we should monitor pipeline afterwards

PR #7593, opened 2024-10-21
feat: Dynamically determine latest init runtime

lucashuy:

None

The current "most popular" runtime is hard coded to Python 3.9.

Dynamically checks the (Python) runtimes we support, and returns the latest one.

```
Choose an AWS Quick Start application template
        1 - Hello World Example
        2 - Data processing
.
.
<truncated>
.
.
Template: 1

Use the most popular runtime and package type? (python3.12 and zip) [y/N]: y

Would you like to enable X-Ray tracing on the function(s) in your application?  [y/N]: 

Would you like to enable monitoring using CloudWatch Application Insights?
For more info, please view https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-application-insights.html [y/N]: 

Would you like to set Structured Logging in JSON format on your Lambda functions?  [y/N]: 

Project name [sam-app]: sam2

    -----------------------
    Generating application:
    -----------------------
    Name: sam2
    Runtime: python3.12
    Architectures: x86_64
    Dependency Manager: pip
    Application Template: hello-world
    Output Directory: .
    Configuration file: sam2/samconfig.toml
    
    Next steps can be found in the README file at sam2/README.md

$ cat sam2/template.yaml | grep Runtime
      Runtime: python3.12
```

None.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -12,7 +12,7 @@ import click
 from botocore.exceptions import ClientError, WaiterError
 
 from samcli.commands._utils.options import generate_next_command_recommendation
-from samcli.commands.exceptions import InvalidInitOptionException, SchemasApiException
+from samcli.commands.exceptions import InvalidInitOptionException, PopularRuntimeNotFoundException, SchemasApiException
 from samcli.commands.init.init_flow_helpers import (
     _get_image_from_runtime,
     _get_runtime_from_image,
@@ -335,10 +335,8 @@ def _get_latest_python_runtime() -> str:
     str:
         The name of the latest Python runtime (ex. "python3.12")
     """
-
-    # set python3.9 as fallback
-    latest_major = 3
-    latest_minor = 9
+    latest_major = 0
+    latest_minor = 0
 
     compiled_regex = re.compile(r"python(.*?)\.(.*)")
 
@@ -367,6 +365,12 @@ def _get_latest_python_runtime() -> str:
             latest_minor = version_minor
         elif version_major == latest_major:
             latest_minor = version_minor if version_minor > latest_minor else latest_minor
+
+    if not latest_major:
+        # major version is still 0, assume that something went wrong
+        # this in theory should not happen as long as Python is
+        # listed in the SUPPORTED_RUNTIMES constant
+        raise PopularRuntimeNotFoundException("Was unable to search for the latest supported runtime")
 
     selected_version = f"python{latest_major}.{latest_minor}"
 
--- a/samcli/commands/init/interactive_init_flow.py
+++ b/samcli/commands/init/interactive_init_flow.py
@@ -4,6 +4,7 @@ Isolates interactive init prompt flow. Expected to call generator logic at end o
 
 import logging
 import pathlib
+import re
 import tempfile
 from typing import Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


samcli/commands/init/interactive_init_flow.py
```diff
@@ -323,6 +325,49 @@ def _generate_from_use_case(
...
+
+        try:
+            version_major = int(matched_groups[0])
+            version_minor = int(matched_groups[1])
+        except (ValueError, IndexError):
+            LOG.debug(f"Failed to parse version while checking {runtime}")
+            continue
+
+        latest_major = version_major if version_major > latest_major else latest_major
+        latest_minor = version_minor if version_minor > latest_minor else latest_minor
```
lucashuy, line 366: Looking at this now, this isn't going to work in the event the major number is updated, but then the minor was not. I'll change this and add a unit test

samcli/commands/init/interactive_init_flow.py
```diff
@@ -323,6 +325,49 @@ def _generate_from_use_case(
...
+
+    Returns
+    -------
+    str:
+        The name of the latest Python runtime (ex. "python3.12")
+    """
+
+    # set python3.9 as fallback
+    latest_major = 3
+    latest_minor = 9
```
mndeveci, line 341: Should we fail if we can't find a version? My only concern is, we now need to keep this up once a python runtime is getting deprecated. 🤔

samcli/commands/init/interactive_init_flow.py
```diff
@@ -323,6 +325,49 @@ def _generate_from_use_case(
...
+            version_major = int(matched_groups[0])
+            version_minor = int(matched_groups[1])
+        except (ValueError, IndexError):
+            LOG.debug(f"Failed to parse version while checking {runtime}")
+            continue
+
+        latest_major = version_major if version_major > latest_major else latest_major
+        latest_minor = version_minor if version_minor > latest_minor else latest_minor
+
+    return f"python{latest_major}.{latest_minor}"
```
mndeveci, line 368: Should we log the version before we return?

tests/unit/commands/init/test_cli.py
```diff
@@ -3193,3 +3196,30 @@ def test_init_cli_generate_app_template_provide_via_application_insights_options
             True,
             False,
         )
+
+    def test_latest_python_fetcher_returns_latest(self):
+        latest_python = "python3.100000"
```
mndeveci, line 3201: I am not sure who from our team can see this version getting released 😂

samcli/commands/init/interactive_init_flow.py
lucashuy: We can, I can log it as a debug message since info would immediately log this in the middle of the interactive flow. The customer already sees the latest determined version too (made a change below to now include this in the click prompt).

samcli/commands/init/interactive_init_flow.py
lucashuy: That's a good call out, the entire point of doing this dynamically was to avoid having to update this manually 😅 

We can fail instead of returning some kind of default.

mndeveci (approved):

hnnasit (approved):

Issue #7585, opened 2024-10-19
Bug: sam local invoke - KeyError


schambers-slate:
### Description:
Running any sam local commands results in this error. Randomly started happening yesterday and I'm unable to run `sam local`

This started happening randomly and I'm unable to run any sam commands locally included sam local start-api, sam local invoke, etc. All cli commands with sam local result in same output below.

### Steps to reproduce:
Create a new project via sam init
run sam local start-api

### Observed result:
~/projects/redacted » sam local invoke sampleLambda --debug                                                                                                                                                                        redacted@Slate-MBP
2024-10-19 09:50:10,661 | Config file location: /Users/redacted/projects/redacted/samconfig.toml
2024-10-19 09:50:10,662 | Loading configuration values from [default.['local', 'invoke'].parameters] (env.command_name.section) in config file at '/Users/redacted/projects/redacted/samconfig.toml'...
2024-10-19 09:50:10,663 | Configuration values successfully loaded.
2024-10-19 09:50:10,663 | Configuration values are: {'stack_name': 'redacted'}
2024-10-19 09:50:10,666 | Using SAM Template at /Users/redacted/projects/redacted/template.yaml
2024-10-19 09:50:10,681 | Using config file: samconfig.toml, config environment: default
2024-10-19 09:50:10,682 | Expand command line arguments to:
2024-10-19 09:50:10,682 | --template_file=/Users/redacted/projects/redacted/template.yaml --function_logical_id=sampleLambda --no_event --layer_cache_basedir=/Users/redacted/.aws-sam/layers-pkg --container_host=localhost
--container_host_interface=127.0.0.1
2024-10-19 09:50:10,683 | local invoke command is called
2024-10-19 09:50:10,726 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2024-10-19 09:50:10,754 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2024-10-19 09:50:10,755 | Unable to find Click Context for getting session_id.
2024-10-19 09:50:10,755 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': 'e9ebed45-0a06-406e-9de7-81fe11c607e7', 'installationId': '45bb19e2-1959-43a3-ac1b-6b7fdfb3b4b3', 'sessionId': '382e9533-3a04-4cf1-9501-cb19a0412246',
[...]

PR #7600, opened 2024-10-22
fix: handle generic error

sidhujus:

https://github.com/aws/aws-sam-cli/issues/7585 there was an instance of boto3 returning an unhandled exception

This part of local is not critical to its execution so we should be fine to just blanket catch all exceptions here.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/local/cli_common/invoke_context.py
+++ b/samcli/commands/local/cli_common/invoke_context.py
@@ -365,7 +365,7 @@ class InvokeContext:
                 if self._global_parameter_overrides is None:
                     self._global_parameter_overrides = {}
                 self._global_parameter_overrides["AWS::AccountId"] = account_id
-        except (NoCredentialsError, TokenRetrievalError, ClientError):
+        except (Exception):
             LOG.warning("No current session found, using default AWS::AccountId")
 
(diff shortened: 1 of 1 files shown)


samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -365,7 +365,7 @@ def _add_account_id_to_global(self) -> None:
                 if self._global_parameter_overrides is None:
                     self._global_parameter_overrides = {}
                 self._global_parameter_overrides["AWS::AccountId"] = account_id
-        except (NoCredentialsError, TokenRetrievalError, ClientError):
+        except (Exception):
             LOG.warning("No current session found, using default AWS::AccountId")
```
lucashuy, line 369: Not blocking anything, but I wonder if its worth logging what exception we voided as a debug log. Since this method is a "best effort" attempt at getting an account ID, its not critical, but could help with visibility into what happened.

lucashuy (approved):

hawflau (approved):

Issue #6657, opened 2024-02-06
Deploy command freezes when tag containing space is provided in samconfig file


jbutz:
### Description:
There appears to be a situation where deployment either freezes and does not proceed, or takes an exceptionally long time to proceed, based on the tag values provided. For me it appears to be based on the presence of a space in a tag value being passed through the `samconfig.toml` file. The same issue appeared to occur when I tried to use a YAML samconfig file as well.

### Steps to reproduce:
1. Initialize a new SAM App, I used Node v20 Hello World Example
2. Deploy the app without any changes
3. Add the following under `[default.deploy.parameters]` in the samconfig.toml file
    ```toml
    tags = [
        "ProjectName=sample-aws-sam-debugging-app",
        "SystemCustodian=john.doe@example.com",
        "SystemOwner=alice.public@example.com",
        "PrimaryItContact=john.doe@example.com",
        "CostCenter=1000000",
        "CostCenterApprover=jane.doe@example.com",
        'Level1BusinessArea="Digital Core"',
        "DataClassification=125796",
        "Hipaa=No",
    ]
    ```
4. Deploy the application again, it appears to freeze. ctrl + c to abort the deploy
5. Remove the space in `Digital Core`
6. Deploy the application again, the deploy will proceed at the expected rate

### Observed result:
```
> sam deploy --region us-east-2 --debug
2024-02-06 12:52:10,544 | Config file location: /Users/jbutz/workspace/aws-sam-debug/sam-app/samconfig.toml                                                                                                                   
2024-02-06 12:52:10,546 | Loading configuration values from [default.['deploy'].parameters] (env.command_name.section) in config file at '/Users/jbutz/workspace/aws-sam-debug/sam-app/samconfig.toml'...                     
2024-02-06 12:52:10,547 | Configuration values successfully loaded.                                                                                                                                                           
2024-02-06 12:52:10,547 | Configuration values are: {'stack_name': 'sam-app', 'capabilities': 'CAPABILITY_IAM', 'confirm_changeset': True, 'resolve_s3': True, 'tags': ['ProjectName=sample-aws-sam-debugging-app',           
'SystemCustodian=john.doe@example.com', 'SystemOwner=alice.public@example.com', 'PrimaryItContact=john.doe@example.com', 'CostCenter=1000000', 'CostCenterApprover=jane.doe@example.com', 'Level1BusinessArea="Digital Core"',
'DataClassification=125796', 'Hipaa=No']}                                                                                                                                                                                     
[...]

PR #7625, opened 2024-10-29
fix: Another approach to parse tag string with spaces to avoid long running time  on regex matching.

dkphm:

#6657 

- Previous fix (https://github.com/aws/aws-sam-cli/pull/7049) only handle the case when `tags` are provided through `samconfig.toml`, which resulted into a list to `CfnTags`, but it doesn't fix the case when customer specify `--tags` parameter. In the later case, the provided value will be in a string, and we will face the issue with `re.findall` again, mainly due to the complex regex provided to parse `{tag}={tag}` pattern in `tags`.

- The solution is instead of parsing a full complex regex with `{tag}={tag}`, we will look for any quote strings with space in the value (e.g. "Test App"), replace the space with some replacement, and restart the parsing process.

Run `sam deploy` with `--tags tag1="value 1" tag2="value 2"` 

Customer might not be able to set tags properly during `sam deploy`.

#### Mandatory Checklist

- [X] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -16,7 +16,8 @@ PARAM_AND_METADATA_KEY_REGEX = """([A-Za-z0-9\\"\']+)"""
 
 LOG = logging.getLogger(__name__)
 
-def _generate_match_regex(match_pattern, delim = None):
+
+def _generate_match_regex(match_pattern, delim=None):
     """
     Creates a regex string based on a match pattern (also a regex) that is to be
     run on a string (which may contain escaped quotes) that is separated by delimiters.
@@ -31,8 +32,7 @@ def _generate_match_regex(match_pattern, delim = None):
     str: regex expression
 
     """
-    result = (f"""(\\"(?:\\\\{match_pattern}|[^\\"\\\\]+)*\\"|"""
-        + f"""\'(?:\\\\{match_pattern}|[^\'\\\\]+)*\'""")
+    result = f"""(\\"(?:\\\\{match_pattern}|[^\\"\\\\]+)*\\"|""" + f"""\'(?:\\\\{match_pattern}|[^\'\\\\]+)*\'"""
 
     if delim is not None:
         # Non capturing groups reduces duplicates in groups, but does not reduce matches.
@@ -356,6 +356,7 @@ class CfnTags(click.ParamType):
                 return None
             tags_dict.update(parsed_tag)
         return tags_dict
+
 
 class SigningProfilesOptionType(click.ParamType):
     """
--- a/samcli/cli/types.py
+++ b/samcli/cli/types.py
@@ -16,8 +16,7 @@ PARAM_AND_METADATA_KEY_REGEX = """([A-Za-z0-9\\"\']+)"""
 
 LOG = logging.getLogger(__name__)
 
-
-def _generate_match_regex(match_pattern, delim):
+def _generate_match_regex(match_pattern, delim = None):
     """
     Creates a regex string based on a match pattern (also a regex) that is to be
     run on a string (which may contain escaped quotes) that is separated by delimiters.
@@ -32,13 +31,14 @@ def _generate_match_regex(match_pattern, delim):
     str: regex expression
 
     """
+    result = (f"""(\\"(?:\\\\{match_pattern}|[^\\"\\\\]+)*\\"|"""
+        + f"""\'(?:\\\\{match_pattern}|[^\'\\\\]+)*\'""")
 
-    # Non capturing groups reduces duplicates in groups, but does not reduce matches.
-    return (
-        f"""(\\"(?:\\\\{match_pattern}|[^\\"\\\\]+)*\\"|"""
-        + f"""\'(?:\\\\{match_pattern}|[^\'\\\\]+)*\'|"""
-        + f"""(?:\\\\{match_pattern}|[^{delim}\\"\\\\]+)+)"""
-    )
+    if delim is not None:
+        # Non capturing groups reduces duplicates in groups, but does not reduce matches.
+        return result + f"""|(?:\\\\{match_pattern}|[^{delim}\\"\\\\]+)+)"""
+    else:
+        return result + ")"
 
 
 def _unquote_wrapped_quotes(value):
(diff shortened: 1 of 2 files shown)


samcli/cli/types.py
```diff
@@ -286,6 +335,48 @@ def _space_separated_key_value_parser(tag_value):
...
+        Parameters
+        ----------
+        tag_value
+        """
+        tags_dict = {}
+        for value in tag_value.split():
+            parsed, parsed_tag = CfnTags._standard_key_value_parser(value)
+            if not parsed:
+                return False, None
+            tags_dict = {**tags_dict, **parsed_tag}
```
lucashuy, line 351: nit (if `parsed_tag` is of type `dict`)
```suggestion
            tags_dict.update(parsed_tag)
```

samcli/cli/types.py
```diff
@@ -286,6 +335,48 @@ def _space_separated_key_value_parser(tag_value):
...
+        Method to parse space separated `Key1=Value1 Key2=Value2` type tags without using regex.
+        Parameters
+        ----------
+        tag_value
+        """
+        tags_dict = {}
+        for value in tag_value.split():
+            parsed, parsed_tag = CfnTags._standard_key_value_parser(value)
+            if not parsed:
+                return False, None
```
lucashuy, line 350: How come we're also returning a boolean here? Can we just return `None`, or an empty `dict` instead of returning a `tuple`?

samcli/cli/types.py
```diff
@@ -222,13 +248,36 @@ def convert(self, value, param, ctx):
...
-                groups = re.findall(self._pattern, val)
-
-                if not groups:
-                    fail = True
-                for group in groups:
-                    key, v = group
-                    self._add_value(result, _unquote_wrapped_quotes(key), _unquote_wrapped_quotes(v))
+                # Instead of parsing a full {tag}={tag} pattern, we will try to look for quoted string with spaces
+                # first, remove all the spaces and start over again.
+
```
lucashuy, line 253: Is there missing code here, or maybe duplicate comment? This comment and the one under say "first"

samcli/cli/types.py
```diff
@@ -222,13 +248,36 @@ def convert(self, value, param, ctx):
...
+                # First, we need to unquote a full string
+                modified_val = _unquote_wrapped_quotes(val)
+
+                # Next, looking for a quote strings that contain spaces and proceed to replace them
+                quoted_strings_with_spaces = re.findall(self._quoted_pattern, modified_val)
+                quoted_strings_without_spaces = [self._replace_spaces(s) for s in quoted_strings_with_spaces]
```
lucashuy, line 259: nit: no single character variable names

samcli/cli/types.py
```diff
@@ -222,13 +248,36 @@ def convert(self, value, param, ctx):
...
                     self._add_value(result, _unquote_wrapped_quotes(k), _unquote_wrapped_quotes(tags[k]))
             else:
-                groups = re.findall(self._pattern, val)
-
-                if not groups:
-                    fail = True
-                for group in groups:
-                    key, v = group
-                    self._add_value(result, _unquote_wrapped_quotes(key), _unquote_wrapped_quotes(v))
+                # Instead of parsing a full {tag}={tag} pattern, we will try to look for quoted string with spaces
```
lucashuy, line 251: Can we refactor this new logic out to a separate method to unit test? There's a lot of moving parts here

samcli/cli/types.py
```diff
@@ -194,6 +219,7 @@ def __init__(self, multiple_values_per_key=False):
     TAG_REGEX = '[A-Za-z0-9\\"_:\\.\\/\\+-\\@=]'
 
     _pattern = r"{tag}={tag}".format(tag=_generate_match_regex(match_pattern=TAG_REGEX, delim=" "))
+    _quoted_pattern = _generate_quoted_match_regex(match_pattern=TAG_REGEX)
```
lucashuy, line 222: Can we reuse or refactor the old `_generate_match_regex`? They share similar return values.

samcli/cli/types.py
```diff
@@ -222,13 +248,36 @@ def convert(self, value, param, ctx):
...
+                # Next, looking for a quote strings that contain spaces and proceed to replace them
+                quoted_strings_with_spaces = re.findall(self._quoted_pattern, modified_val)
+                quoted_strings_without_spaces = [self._replace_spaces(s) for s in quoted_strings_with_spaces]
+                for s, replacement in zip(quoted_strings_with_spaces, quoted_strings_without_spaces):
+                    modified_val = modified_val.replace(s, replacement[0])
+
+                # Finally, restart the parsing with key=value separated by (multiple) spaces.
+                parsed, tags = self._multiple_space_separated_key_value_parser(modified_val)
+                if parsed:
+                    for k in tags:
```
lucashuy, line 266: nit:
```suggestion
                    for key, value in tags.items():
```

samcli/cli/types.py
```diff
@@ -286,6 +335,48 @@ def _space_separated_key_value_parser(tag_value):
...
+    def _replace_spaces(text, replacement="_"):
+        """
+        Replace spaces in a text with a replacement together with its original locations.
+        Input: "test 1"
+        Output: "test_1" [4]
+        """
+        space_positions = [i for i, char in enumerate(text) if char == " "]
+        modified = text.replace(" ", replacement)
+
+        return modified, space_positions
```
lucashuy, line 364: Thoughts on using dataclasses for return objects, instead of returning tuples and performing array indexes to get the modified strings/replacement index array?

I imagine that changing things to dataclasses will make it easier for future readers to understand what is happening quickly.

lucashuy:
> **What side effects does this change have?**
>
> Customer might not be able to set tags properly during `sam deploy`.

What does this mean? Does this change how tags works for customers?

samcli/cli/types.py
dkphm: This is tricky because the block below

```
groups = re.findall(self._pattern, val)
```

take longer time to return

samcli/cli/types.py
dkphm: They have different regex, hence creating a new function.

lucashuy: The difference between the two is the extra regex string for the deliminator right? Can we make the deliminator argument optional, then just append the extra deliminator regex it was passed in? Another option is to make constants

hawflau (approved):

lucashuy (approved):

PR #7649, opened 2024-11-05
chore: remove go1.x & provided from validate tests

mndeveci:
Removing `go1.x` and `provided` from list of runtimes we test for `sam validate`

### DIFF
--- a/tests/integration/validate/test_validate_command.py
+++ b/tests/integration/validate/test_validate_command.py
@@ -170,7 +170,6 @@ class TestValidate(TestCase):
         supported_runtimes = [
             "dotnet8",
             "dotnet6",
-            "go1.x",
             "java21",
             "java17",
             "java11",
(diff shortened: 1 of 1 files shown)


tests/integration/validate/test_validate_command.py
```diff
@@ -170,7 +170,6 @@ def test_lint_supported_runtimes(self):
         supported_runtimes = [
             "dotnet8",
             "dotnet6",
-            "go1.x",
```
hnnasit, line 173: I was just looking at the canaries, I think provided also needs to be removed right?

mndeveci: Right it is confusing since all printed in the same line. Removed it as well.

dkphm (approved):

hawflau (approved):

Issue #7671, opened 2024-11-12
Feature request: Docker build logs on failure


palfrey:
### Describe your idea/feature/enhancement

When you run `sam build --use-container` and your Docker build fails, you get something like the following

```
Building image for EmploymentTimeEmployeeCachingFunction function
Setting DockerBuildArgs for EmploymentTimeEmployeeCachingFunction function
Failed building function EmploymentTimeEmployeeCachingFunction

Build Failed
Error: The command '/bin/sh -c exit 1' returned a non-zero code: 1
```

This isn't very useful, as I've got no idea what the build logs actually look like that caused this.

### Proposal

Dump the docker build logs on failure. So, for the above example, instead I would like something like
```
Building image for EmploymentTimeEmployeeCachingFunction function
Setting DockerBuildArgs for EmploymentTimeEmployeeCachingFunction function
Failed building function EmploymentTimeEmployeeCachingFunction
Step 1/19 : FROM public.ecr.aws/lambda/python:3.11 AS base
 ---> 37318d609dfb
Step 2/19 : RUN exit 1
 ---> [Warning] The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
 ---> Running in 6420d9378807
The command '/bin/sh -c exit 1' returned a non-zero code: 1
Build Failed
Error: The command '/bin/sh -c exit 1' returned a non-zero code: 1
```
### Additional Details

The demo log was generated with the following extra code in the `BuildError` handler in app_builder.py
```python
            for log in ex.build_log:
                if "stream" in log:
                    self._stream_writer.write_str(log["stream"])
                elif "error" in log:
                    self._stream_writer.write_str(log["error"])
```

mndeveci:
Thanks for creating this issue.

This makes sense from our side, and it should improve the user experience in general. Since you already find the point which needs the update, would like to raise a pull request with the change?

palfrey:
See https://github.com/aws/aws-sam-cli/pull/7675

palfrey:
Released as part of [1.133.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.133.0) and confirmed working

PR #7675, opened 2024-11-13
Add logs on docker build failure

palfrey:

#7671

So we can actually debug Docker build issues

By adding logs :)

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/build/app_builder.py
```diff
@@ -430,6 +430,11 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
             LOG.debug("%s image is built for %s function", build_image, function_name)
         except docker.errors.BuildError as ex:
             LOG.error("Failed building function %s", function_name)
+            for log in ex.build_log:
```
jonife, line 433: Curios Questions: If ex.build_log can be large, iterating over it and writing output in a potentially synchronous manner might be slow. Should we consider case whether `_stream_writer.write_str()` is buffered or should be optimized for larger outputs. I checked here https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/utils/stream_writer.py#L52

samcli/lib/build/app_builder.py
```diff
@@ -430,6 +430,11 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
...
+                if "stream" in log:
+                    self._stream_writer.write_str(log["stream"])
```
jonife, line 435: When writing to `_stream_writer`, I think checking if it is clear what type of output will be shown. I think we should include context to indicate whether it is part of a standard log or an error message, such as prepending "[Error]: " when writing log["error"].


```
except docker.errors.BuildError as ex:
    LOG.error("Failed building function %s", function_name)
    for log in ex.build_log:
        if "stream" in log:
            self._stream_writer.write_str(f"Log: {log['stream']}")
        elif "error" in log:
            self._stream_writer.write_str(f"[Error]: {log['error']}")
        else:
            LOG.warning("Unexpected log entry format: %s", log)

```

jonife: We should consider a way to differentiate between different types of docker.errors.BuildError logs (e.g., specific errors vs. general stream logs), more granular handling might be helpful. This could mean adding conditionals or further parsing within the loop. I was thinking

samcli/lib/build/app_builder.py
```diff
@@ -430,6 +430,11 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
...
+                elif "error" in log:
+                    self._stream_writer.write_str(log["error"])
```
roger-zhangg, line 437: Seems we have a very similar function here 
https://github.com/aws/aws-sam-cli/blob/a18782ac09006914664a064bcd5c8964c2c97ca7/samcli/lib/build/app_builder.py#L443

I wonder if it's possible to reuse this function in this place
```suggestion
            self._stream_lambda_image_build_logs(ex.build_log, function_name)
```

palfrey: Looked at that initially, but rejected because the `_stream_lambda_image_build_logs` uses `LogStreamer._stream_write` which will throw a `LogStreamError` if it sees an error rather than logging it. Provided the logs are always of the form `<several 'stream' errors>,<one 'error' log>` this will work, otherwise not.

This is incompatible with @jonife's suggestion from above BTW. Which one should I build?

roger-zhangg: Thanks for the clarification. Yes you are right. This function won't work on logs that has multiple error. Maybe we can consider adding a param in `LogStreamer ` to suppress the error. In this way it will also address @jonife concerns

palfrey: Adding the param would work, I'll have a look at that. It won't address @jonife's concerns though, as `LogStreamer` doesn't write anything about what sort of log it is. My experiments with it so far indicate that for the usual Docker log patterns it literally does what my patch did (minus the error problems).

roger-zhangg:
@palfrey Thanks for implementing the change! FYI here I believe this is a vaild test failure in CI, please check
```
FAILED tests/unit/lib/build_module/test_app_builder.py::TestApplicationBuilder_build_lambda_image_function::test_can_raise_build_error - AssertionError: [call('Some earlier log'), call(''), call('Build failed'), call('\r\n')] != [call('Some earlier log'), call(''), call('Build failed'), call('\n')]
```

palfrey:
> @palfrey Thanks for implementing the change! FYI here I believe this is a vaild test failure in CI, please check
> 
> ```
> FAILED tests/unit/lib/build_module/test_app_builder.py::TestApplicationBuilder_build_lambda_image_function::test_can_raise_build_error - AssertionError: [call('Some earlier log'), call(''), call('Build failed'), call('\r\n')] != [call('Some earlier log'), call(''), call('Build failed'), call('\n')]
> ```

Looks like a Windows-specific line-ending failure. Got a probable fix, but no local Windows machine so will have to rely on CI to test this one.

samcli/lib/docker/log_streamer.py
```diff
@@ -73,7 +74,12 @@ def _stream_write(self, _id: str, status: str, stream: str, progress: str, error
         :param error: docker log error
         """
         if error:
-            raise LogStreamError(msg=error)
+            if self._throw_on_error:
+                raise LogStreamError(msg=error)
+            else:
+                self._stream.write_str(error)
+                return
```
mndeveci, line 81: When we write this to the error stream, shouldn't we also throw it again? Changing this variable into opposite boolean to indicate whether an error should be written into stream or not could be better 🤔

palfrey: No, because you might have multiple error lines. Also the throw in this features case is with the whole Docker build exception as the content vs just an individual error line.

roger-zhangg (approved):
LGTM ❤️

jonife (approved):
Thank you for your contribution

palfrey:
So how does this get final approval? Anyone have any sense of the timelines on this one?

hnnasit (approved):

palfrey:
This went out as part of [v1.133.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.133.0)

PR #7745, opened 2024-11-26
Add support for --no-use-container option

jonife:
`use-container` currently doesn't have an opposite option called `no-use-container`, user wants to use the IDE toolkit which depends on samcli to set default behaviors but still wants the flexibility of using the opposite option to carry out different operation.

Add support for `no-use-container` allow the cli to pass `use-container` = false in the samConfig file 

`no-use-container` can result to` use_container = false` in samConfig
`sam sync` 

```
Container Options:                        
                                    
    -u, --use-container / --no-use-container
                                    Build functions within an AWS Lambda-like container.
                                                                      
```

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/commands/_utils/options.py
+++ b/samcli/commands/_utils/options.py
@@ -849,8 +849,10 @@ def resolve_image_repos_option(f):
 
 def use_container_build_click_option():
     return click.option(
-        "--use-container",
+        "--use-container/--no-use-container",
         "-u",
+        required=False,
+        default=False,
         is_flag=True,
         help="Build functions within an AWS Lambda-like container.",
     )
--- a/tests/unit/commands/samconfig/test_samconfig.py
+++ b/tests/unit/commands/samconfig/test_samconfig.py
@@ -163,6 +163,65 @@ class TestSamConfigForAllCommands(TestCase):
             )
 
     @patch("samcli.commands.build.command.do_cli")
+    def test_build_with_no_use_container(self, do_cli_mock):
+        config_values = {
+            "resource_logical_id": "foo",
+            "template_file": "mytemplate.yaml",
+            "base_dir": "basedir",
+            "build_dir": "builddir",
+            "cache_dir": "cachedir",
+            "cache": True,
+            "use_container": False,
+            "manifest": "requirements.txt",
+            "docker_network": "mynetwork",
+            "skip_pull_image": True,
+            "parameter_overrides": "ParameterKey=Key,ParameterValue=Value ParameterKey=Key2,ParameterValue=Value2",
+            "container_env_var": [("")],
+            "container_env_var_file": "file",
+            "build_image": [("")],
+            "exclude": [("")],
+            "mount_with": "read",
+        }
+
+        with samconfig_parameters(["build"], self.scratch_dir, **config_values) as config_path:
+            from samcli.commands.build.command import cli
+
+            LOG.debug(Path(config_path).read_text())
+            runner = CliRunner()
+            result = runner.invoke(cli, [])
+
+            LOG.info(result.output)
+            LOG.info(result.exception)
+            if result.exception:
+                LOG.exception("Command failed", exc_info=result.exc_info)
+            self.assertIsNone(result.exception)
+
+            do_cli_mock.assert_called_with(
+                ANY,
+                "foo",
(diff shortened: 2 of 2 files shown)


tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -162,6 +162,65 @@ def test_build(self, do_cli_mock):
...
+            "exclude": [("")],
+            "mount_with": "read",
+        }
+
+        with samconfig_parameters(["build"], self.scratch_dir, **config_values) as config_path:
+            from samcli.commands.build.command import cli
+
+            LOG.debug(Path(config_path).read_text())
+            runner = CliRunner()
+            result = runner.invoke(cli, [])
```
valerena, line 191: What you probably want to check here is that the parameter in the **command line** gets picked up (`--no-use-container`, that you're adding). To test that, you need to pass it in this part. Like [this line](https://github.com/aws/aws-sam-cli/blob/36e033b8d01e4ea13844023b9e3070f108389b67/tests/unit/commands/samconfig/test_samconfig.py#L251). 

Otherwise, you're just testing that the samconfig works fine, not the command line argument. (Probably the same for sync below)

samcli/commands/_utils/options.py
```diff
@@ -849,8 +849,10 @@ def resolve_image_repos_option(f):
 
 def use_container_build_click_option():
     return click.option(
-        "--use-container",
+        "--use-container/--no-use-container",
```
roger-zhangg, line 852: If customer doesn't specify `--use-container` or `--no-use-container`, will sam cli still use the value from samconfig.toml after this change?

roger-zhangg (commented):
if customer has `use container` in sam config.toml, Will `sam build --no-use-container --save-params` replace `--use-container` in samconfig.toml with `--no-use-container` ?

tests/unit/commands/samconfig/test_samconfig.py
jonife: Both tests are different. The use cases in the link you provided focus on testing overrides. For example, if my configuration includes `"use_container": True`, I can pass `--no-use-container` to override it, effectively setting `"use_container": False`.

This is a draft PR to validate whether make schema works. Please note that the PR is not complete as you review it.

samcli/commands/_utils/options.py
jonife: This documentation might help answer your question: [SAM CLI Configuration File Precedence](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-config.html#serverless-sam-cli-config-precedence).

jonife:
> if customer has `use_container=true` in sam config.toml, Will `sam build --no-use-container --save-params` update `use_container=true` in samconfig.toml to `use_container=false` ?

yes

samcli/commands/_utils/options.py
```diff
@@ -849,8 +849,10 @@ def resolve_image_repos_option(f):
...
         "-u",
+        required=False,
+        default=False,
         is_flag=True,
```
roger-zhangg, line 856: Non-blocking, `is_flag` could be removed after this change.

tests/unit/commands/samconfig/test_samconfig.py
hnnasit: I suppose we could have 1 unit test for testing the default config value as here and another one like @valerena mentioned about with `use-container=True` and override it with `--no-use-container` to assert it gets set to `False`. The `sam sync` test below could be updated with the latter case since both cases stem from the same click option anyways (we might not need duplicate tests for each command).

samcli/commands/_utils/options.py
vicheey: If the current samconfig.toml file has `use-container = True` and customer run sam command with `--no-use-container --save-parameters`, will we remove the `use-container = True` from samconfig.toml file before saving `--no-use-container = True`?

jonife:
closing this pr in favor of https://github.com/aws/aws-sam-cli/pull/7769 due to inability to rebase locally.

jonife:
closing this pr in favor of https://github.com/aws/aws-sam-cli/pull/7769 due to inability to rebase locally.

Issue #7298, opened 2024-08-01
Bug: sam local start-api fails to initialise lambda function containers within podman/toolbox container


AidanOB:
### Description:
When attempting to run `sam local start-api` from within podman/toolbox on Fedora 40 Kinoite, the lambda function containers fail to initialise. Command hangs for over an hour (at least) without initialising containers. On one occassion, when left overnight it initialised the container and was ready to be called. 

To test, utilised AWS Lambda powertools tutorial template found here: https://docs.powertools.aws.dev/lambda/python/latest/tutorial/#code-example

Output `Initializing the lambda functions containers.`

### Steps to reproduce:
Fedora 40 Kinoite -> Toolbox
```
toolbox create sam-test
toolbox enter sam-test
cd temp
sudo ./sam-installation/install
export DOCKER_HOST=unix:///run/user/1000/podman/podman.sock
sam build && sam local start-api --debug
```

### Observed result:
```
[user@toolbox hello-world-tutorial]$ sam build && sam local start-api --debug
Starting Build use cache                                                                                                              
Manifest is not changed for (HelloWorldFunction), running incremental build                                                           
Building codeuri: /var/home/user/hello-world-tutorial/hello_world runtime: python3.12 metadata: {} architecture: x86_64 
functions: HelloWorldFunction                                                                                                         
 Running PythonPipBuilder:CopySource                                                                                                  
 Running PythonPipBuilder:CopySource                                                                                                  

Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Validate SAM template: sam validate
[*] Invoke Function: sam local invoke
[*] Test Function in the Cloud: sam sync --stack-name {{stack-name}} --watch
[*] Deploy: sam deploy --guided
2024-08-01 13:57:06,971 | Config file location: /var/home/user/hello-world-tutorial/samconfig.toml                      
2024-08-01 13:57:06,973 | Loading configuration values from [default.['local', 'start-api'].parameters] (env.command_name.section) in 
config file at '/var/home/user/hello-world-tutorial/samconfig.toml'...                                                  
2024-08-01 13:57:06,974 | Configuration values successfully loaded.                                                                   
2024-08-01 13:57:06,975 | Configuration values are: {'stack_name': 'hello-world-tutorial', 'warm_containers': 'EAGER'}                
[...]

PR #7760, opened 2024-12-01
Fix blocking image observer

chiting:
I'm not very good with python but I figured this should fix some issues happening when running locally with Podman.

I'm also not sure how it behaves with Docker or other alternatives (like Colima).

https://github.com/aws/aws-sam-cli/issues/7298

Calling the events function from the main thread will block until it receives data as docker-py uses urllib3.openurl under the hood which is blocking (read).

It can be replicated using `curl`:
```
# my docker.sock is a symlink to podman API socket
% curl -vv --unix-socket /var/run/docker.sock 'http://localhost/v1.35/events?filters=%7B%22type%22%3A+%5B%22image%22%5D%7D'
```
The above will block until it receives data.

We don't want to block the main thread, so moving the self.events assignment in a thread (the function that actually watches for changes) seems to be the solution.

I can't see any side effect this would introduce.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


chiting:
Would it be better to add `hasattr` check on https://github.com/aws/aws-sam-cli/blob/3db063b800bf29dddd3f4fb44692caa46e546fc3/samcli/lib/utils/file_observer.py#L371?

samcli/lib/utils/file_observer.py
chiting, line 1: Would adding a `hasattr` check on https://github.com/aws/aws-sam-cli/pull/7760/files#diff-6fd756ada91f34759a452371f0fe2a7e02622333f4ed26f48b94279f014b2d96R371 be desirable?

PR #7769, opened 2024-12-03
Add support for --no-use-container option 

jonife:
`use-container` currently doesn't have an opposite option called `no-use-container`, user wants to use the IDE toolkit which depends on samcli to set default behaviors but still wants the flexibility of using the opposite option to carry out different operation.

Add support for `no-use-container` allow the cli to pass `use-container` = false in the samConfig file 

`no-use-container` can result to` use_container = false` in samConfig
`sam sync` 

```
Container Options:                        
                                    
    -u, --use-container / --no-use-container
                                    Build functions within an AWS Lambda-like container.
                                                                      
```

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/tests/unit/commands/samconfig/test_samconfig.py
+++ b/tests/unit/commands/samconfig/test_samconfig.py
@@ -163,6 +163,182 @@ class TestSamConfigForAllCommands(TestCase):
             )
 
     @patch("samcli.commands.build.command.do_cli")
+    def test_build_with_no_use_container(self, do_cli_mock):
+        config_values = {
+            "resource_logical_id": "foo",
+            "template_file": "mytemplate.yaml",
+            "base_dir": "basedir",
+            "build_dir": "builddir",
+            "cache_dir": "cachedir",
+            "cache": True,
+            "use_container": False,
+            "manifest": "requirements.txt",
+            "docker_network": "mynetwork",
+            "skip_pull_image": True,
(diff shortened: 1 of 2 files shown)


hnnasit (approved):

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -1027,6 +1203,97 @@ def test_sync(
                 {"HelloWorld": ["file.txt", "other.txt"], "HelloMars": ["single.file"]},
             )
 
+    @patch("samcli.commands._utils.experimental.is_experimental_enabled")
+    @patch("samcli.lib.cli_validation.image_repository_validation._is_all_image_funcs_provided")
+    @patch("samcli.lib.cli_validation.image_repository_validation.get_template_artifacts_format")
+    @patch("samcli.commands._utils.template.get_template_artifacts_format")
+    @patch("samcli.commands._utils.options.get_template_artifacts_format")
+    @patch("samcli.commands.sync.command.do_cli")
+    def test_sync_with_no_use_container(
```
lucashuy, line 1212: qq: is this test suppose to set the `--no-use-container` option? Otherwise at first glance, it looks like the test is checking the same behaviour as before (ie. this uses the config value and we didn't specify any CLI options)

lucashuy (approved):
Thanks for making changes!

Issue #7839, opened 2025-01-17
Bug: local invoke use expired Ruby 3.3 image which contains bug not fixed


elct9620:
### Description:

The latest `public.ecr.aws/lambda/ruby:3.3` is upgraded to `3.3.5` but `sam local invoke` still uses `public.ecr.aws/lambda/ruby:3.3-preview` and the version is `3.3.0` this version has some bug cause `sam local invoke` cannot work correctly.

https://bugs.ruby-lang.org/issues/20090

### Steps to reproduce:

Add a ruby gem `dry-system` with version `1.2.1` with this content

```ruby
require 'dry/system'

class Application < Dry::System::Container
end
```

When invoking the local function the error is raised.

```
{
  "errorMessage": "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/plugins/monitoring.rb:35: anonymous block parameter is also used within block",
  "errorType": "Init<SyntaxError>",
  "stackTrace": [
    "/var/lang/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'",
    "/var/lang/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'",
    "/opt/ruby/3.3.0/gems/zeitwerk-2.7.1/lib/zeitwerk/core_ext/kernel.rb:26:in `require'",
    "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/plugins.rb:65:in `<module:Plugins>'",
    "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/plugins.rb:5:in `<module:System>'",
    "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/plugins.rb:4:in `<module:Dry>'",
    "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/plugins.rb:3:in `<top (required)>'",
    "/var/lang/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'",
    "/var/lang/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'",
    "/opt/ruby/3.3.0/gems/zeitwerk-2.7.1/lib/zeitwerk/core_ext/kernel.rb:26:in `require'",
    "/opt/ruby/3.3.0/gems/dry-system-1.2.1/lib/dry/system/container.rb:56:in `<class:Container>'",
[...]

PR #7849, opened 2025-01-25
chore(runtime): move ruby3.3 out of test runtime

roger-zhangg:
ruby 3.3 is officially supported

#7839

The latest public.ecr.aws/lambda/ruby:3.3 is upgraded to 3.3.5 but sam local invoke still uses public.ecr.aws/lambda/ruby:3.3-preview and the version is 3.3.0 this version has some bug cause sam local invoke cannot work correctly.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/lambda_image.py
+++ b/samcli/local/docker/lambda_image.py
@@ -31,7 +31,7 @@ LOG = logging.getLogger(__name__)
 
 RAPID_IMAGE_TAG_PREFIX = "rapid"
 
-TEST_RUNTIMES = []
+TEST_RUNTIMES = ["N/A"]
 
 
 class Runtime(Enum):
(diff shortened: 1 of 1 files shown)


samcli/local/docker/lambda_image.py
```diff
@@ -31,7 +31,7 @@
 
 RAPID_IMAGE_TAG_PREFIX = "rapid"
 
-TEST_RUNTIMES = ["ruby3.3"]
+TEST_RUNTIMES = ["N/A"]
```
valerena, line 34: I see that you tried to do `TEST_RUNTIMES = []` first (which seems like the most logical option) but it failed to build. 

However, the failure [in the build](https://github.com/aws/aws-sam-cli/actions/runs/12960487479/job/36154409106) is 
```
samcli/local/docker/lambda_image.py:34: error: Need type annotation for "TEST_RUNTIMES" (hint: "TEST_RUNTIMES: list[<type>] = ...")  [var-annotated]
```

So I feel it would be better to actually leave it empty and with that type annotation, which I think it should be something like this:

```
TEST_RUNTIMES: list[str] = []
```

Vandita2020 (approved):

valerena (approved):

PR #7865, opened 2025-02-05
fix: Resolve symlinks on local invoke only

valerena:
SAM CLI shouldn't be mounting symlinks by default when building in a container

Symbolic links are used in SAM CLI to achieve faster build times with `build-in-source`. The build process creates symlinks for dependencies, that then can be used when running `sam local invoke`. These symlinks don't need to be mounted during `sam build`.

We add a new context property for the container to make a different decision depending on the context.

If SAM CLI users were relying on this behavior during builds, their workflow might get broken. Please create an issue if you're affected by this change, and we'll prioritize to add a new property to allow mounting symlinks in any case.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -13,6 +13,7 @@ import socket
 import tempfile
 import threading
 import time
+from enum import Enum
 from typing import Dict, Iterator, Optional, Tuple, Union
 
 import docker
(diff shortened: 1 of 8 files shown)


seshubaws (approved):

samcli/local/docker/container.py
```diff
@@ -648,3 +658,18 @@ def is_running(self):
             return real_container.status == "running"
         except docker.errors.NotFound:
             return False
+
+    def _resolve_symlinks(self, context) -> bool:
+        """_summary_
+
+        Parameters
+        ----------
+        context : sacli.local.docker.container.ContainerContext
```
seshubaws, line 667: nit: samcli but not a blocker

roger-zhangg (approved):
LGTM

PR #7883, opened 2025-02-11
test: Increase start-api tests timeouts

valerena:
The start-api tests take longer on Windows, so we increase the timeout so they don't fail.

Failure on integration tests. 

Tests were failing just because of a timeout and not because of real problems in the tests. This were all "invalid template" tests, that had a very small timeout, but it depends on the machine where these tests are run.

Increasing the timeout

If the test were to get stuck forever, it will take longer for the overall test infrastructure to stop this particular test.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


seshubaws (approved):

vicheey (approved):
Thank you for the fix.

PR #7891, opened 2025-02-14
feat: New parameter to resolve and mount symlinks in container

valerena:
`--mount-symlinks` will make SAM CLI to always mount symlinks present among the files to build or invoke (top level only). We're making the default for symlinks to not be mounted, except ones needed to use "build-in-source" (`node_modules` for NodeJS)

By default SAM CLI shouldn't be resolving top level symlinks, but we do add a parameter so the user can force that behavior when necessary. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


vicheey (approved):
Thank you.

seshubaws (approved):

PR #7892, opened 2025-02-15
feat: New param to remove memory limit in invoke

valerena:
`--no-memory-limit` removes the memory limitation in the container during local invoke, even when there's a memory configured in the SAM template.

In certain conditions, Docker fails to run when a memory limit is passed: https://stackoverflow.com/questions/70872355/how-to-resolve-cgroup-error-when-running-docker-container-inside-a-docker-contai

Allow customers to avoid the Docker issue mentioned above

Add a new parameter so the memory limit is not automatically passed from the template.

None

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


samcli/commands/local/invoke/cli.py
```diff
@@ -66,6 +66,12 @@
 )
 @click.option("--no-event", is_flag=True, default=True, help="DEPRECATED: By default no event is assumed.", hidden=True)
 @mount_symlinks_option
+@click.option(
+    "--no-memory-limit",
+    default=False,
+    is_flag=True,
+    help="Remove the Memory limit during local invoke. This runs the container without the --memory parameter",
```
vicheey, line 73: Are we doing this only for `sam local invoke` only? I think the issue happens for  `sam local start-api` and `sam local start-lambda` as well. 

The message here is specific to `local invoke`. Suggesting more generic message: 
`Remove the memory limit when running Docker container.`

vicheey (changes requested):

roger-zhangg (approved):
lgtm

samcli/commands/local/invoke/cli.py
valerena: Good point. we might need the same for the other commands. I'll update that, also with a generic message.

vicheey (approved):

Issue #4053, opened 2022-07-13
`sam deploy` for deduped build uploads identical build artifact many times


yskkin:
### Describe your idea/feature/enhancement

I `sam deploy`-ed 17 lambdas which shares `CodeUri` and `Runtime` (thus deduped build is enabled). 
Got following output:

```
Uploading to my-stack-name/eb7b8e83[...snip...]  17287115 / 17287115  (100.00%)
File with same data already exists at my-stack-name/eb7b8e83[...snip...], skipping upload
[... another identical 15 lines ...]
```

This output implies `sam deploy` unnecessarily uploads identical artifact.

### Proposal

Ignore identical upload.
It takes few seconds for each skipped upload.
Ignoring identical upload can improve performance of `sam deploy`.

jfuss:
@yskkin Not sure I follow what you are proposing here. The output states "skipping upload" so not sure what is indicating unnecessary uploads.

konarskis:
Thanks for looking into this guys.
This is a real blocker for us, as it's taking nearly 50 minutes to deploy the stack at times...
Any way this can be prioritized?

We'll have to switch over to CDK otherwise I'm afraid, as the release times are just too long.

Thanks again

jfuss:
@konarskis If this is blocking you, you are free to submit a PR to patch.

mbklein:
I started looking into this issue because I was experiencing the same thing @konarskis was. It turns out it's not the `head_object` check that's taking so long – it's the fact that `sam build` puts each function into its own directory within `.aws-sam/build`, and then `sam deploy` zips up each of those directories in turn and then checks to see if the resulting file already exists in s3.

So for two functions, `functionA` and `functionB`, with the same `CodeUri` and `Runtime`, it has to:

1. Zip up `.aws-sam/build/functionA` and get the md5 of the result
2. Check to see if a file with that md5 has already been uploaded
3. It hasn't, so upload it
4. Zip up `.aws-sam/build/functionA` and get the md5 of the result
5. Check to see if a file with that md5 has already been uploaded
6. It has, so skip it

It's Step 4 that's the time consuming one for each function, not Step 5.

[...]

PR #7923, opened 2025-03-06
Experimental flag to boost 'package' command performance

eyalroth:
#4053

`sam package` (and `deploy`) can take an extremely long time.

This happens because the command zips every lambda resource in a template, even if all the lambdas point to the same local path. For instance, with NodeJS lambdas, there are many files involved due to the nature of npm/node, and that puts a serious dent on the disk (especially so in CI environments where SSD is not guarenteed).

Normally, `sam build` will create a separate directory for each lambda resource in the template. However, there is an experimental flag -- `SAM_CLI_BETA_BUILD_PERFORMANCE` -- which creates only one shared directory for all the lambdas. This boosts `sam build` performance tremendously, and it works because in most cases, all lambdas in a project share the same code (just a different handler endpoint).

This PR adds a new flag -- `SAM_CLI_BETA_PACKAGE_PERFORMANCE` -- that is aimed to work in conjunction with the build performance flag.

All that it does is add a cache that is currently only used by `ResourceZip` + `S3Uploader`. The uploader checks if the local path was previously uploaded, and if so, it simply returns the cached upload url.

Should have no effect as the changes are only enabled when using the new flag.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/_utils/experimental.py
+++ b/samcli/commands/_utils/experimental.py
@@ -46,6 +46,9 @@ class ExperimentalFlag:
     BuildPerformance = ExperimentalEntry(
         "experimentalBuildPerformance", EXPERIMENTAL_ENV_VAR_PREFIX + "BUILD_PERFORMANCE"
     )
+    PackagePerformance = ExperimentalEntry(
+        "experimentalPackagePerformance", EXPERIMENTAL_ENV_VAR_PREFIX + "PACKAGE_PERFORMANCE"
+    )
     IaCsSupport = {
         "terraform": ExperimentalEntry(
             "experimentalTerraformSupport", EXPERIMENTAL_ENV_VAR_PREFIX + "TERRAFORM_SUPPORT"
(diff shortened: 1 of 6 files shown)


roger-zhangg:
Hi @eyalroth thanks for the PR, I see currently there are few errors in CI, do you mind helping to fix them?

eyalroth:
Thank you @roger-zhangg! 

Turns out that when I initially ran `make pr`, it actually failed but I didn't notice.

It failed regardless of my changes, but because I was running with Python 3.12 (default installed on my MacOS). It took me some time to realize I need to install Python 3.9 and setup the virtual env with it. Only then the tests actually worked. It might be worth updating the development guide with these requirements in mind.

I fixed/added relevant tests, amended the original commit, rebased on latest develop and force-pushed.

`make pr` passes except for `test_import_module_proxy.py` test, which I think is not related to my changes and would fail on my machine regardless.

roger-zhangg (approved):
LGTM, thank you for your contribution!

valerena (approved):

PR #7938, opened 2025-03-17
Feat allow dict values in generate event cmd

lvthillo:

I want to add the ability to override not only string values but also dictionary values in sam local generate-event. Currently, after running the CLI command, you have to manually adjust the file values.

An example is queryStringParameters in API Gateway. This is not supported via the CLI command. When making basic modifications to event-mapping.json and apigateway/AwsProxy.json to support the parameter, a string is returned, whereas queryStringParameters expects a dictionary.

That's why I'm making changes to `generate_event` to check if a value is of type dict. If so, the string value will be replaced with a dictionary.

I'm creating a draft PR to gather feedback and input on this feature and to see if it's considered a viable option (so I just added this for queryStringParams for API GW aws- proxy.

Test:

```sh
samdev local generate-event apigateway aws-proxy --method GET --path document --body "" --querystringparameters '{"documentId": "1044", "versionId": "v_1"}' --account-id '12345'
```

Output:

```json
{
  "body": "",
  "resource": "/{proxy+}",
  "path": "/document",
  "httpMethod": "GET",
  "isBase64Encoded": true,
  "queryStringParameters": {
    "documentId": "1044",
    "versionId": "v_1"
  },
  "multiValueQueryStringParameters": {
    "foo": [
      "bar"
    ]
  },
  "pathParameters": {
    "proxy": "/document"
  },
  "stageVariables": {
    "baz": "qux"
  },
....
}
```

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/generated_sample_events/events.py
+++ b/samcli/lib/generated_sample_events/events.py
@@ -95,6 +95,9 @@ class Events:
         if hashing is not None:
             transformed = self.hash(hashing, transformed)
 
+        if properties.get("type") == "dict":
+            transformed = json.loads(transformed)
+
         return transformed
 
     @staticmethod
@@ -180,4 +183,19 @@ class Events:
         data = json.dumps(data, indent=2)
 
         # return the substituted file (A string containing the rendered template.)
-        return renderer.render(data, values_to_sub)
+        rendered = renderer.render(data, values_to_sub)
+
+        # Parse the rendered result
+        rendered_json = json.loads(rendered)
+
+        key_map = {k.lower(): k for k in rendered_json.keys()}
+
+        # Update dictionary values in rendered result
+        for key, value in values_to_sub.items():
+            if isinstance(value, dict):
+                original_key = key_map.get(key.lower())
+                if original_key and isinstance(rendered_json[original_key], str):
+                    rendered_json[original_key] = value
+
+        # Return the final JSON with proper dictionary values
+        return json.dumps(rendered_json, indent=2)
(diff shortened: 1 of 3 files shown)


vicheey:
Thank you for your contribution. Let me take look and do some test.

tests/unit/commands/local/generate_event/test_event_generation.py
```diff
@@ -73,6 +73,61 @@ def test_transform(self):
...
+            "apigateway",
+            "aws-proxy",
+            {
+                "method": "GET",
+                "path": "test",
+                "body": "",
+                "querystringparameters": '{"documentId": "1044", "versionId": "v_1"}',
+            },
+        )
+        import json
```
vicheey: nit: please bring this to the top section.

seshubaws (approved):

samcli/lib/generated_sample_events/events.py
```diff
@@ -180,4 +196,24 @@ def generate_event(self, service_name: str, event_type: str, values_to_sub: Dict
...
 
         # return the substituted file (A string containing the rendered template.)
-        return renderer.render(data, values_to_sub)
+        rendered = renderer.render(data, values_to_sub)
+
+        # Parse the rendered result
+        rendered_json = json.loads(rendered)
+
+        # Replace placeholders with actual dict values
+        key_map = {k.lower(): k for k in rendered_json.keys()}
```
seshubaws, line 205: Thanks for all your work @lvthillo! Not sure if we even want to support this case but will this work with nested query string parameters? Otherwise LGTM

lvthillo: I’ve added a test that proves this should work

PR #7950, opened 2025-03-24
feat: add --serverless-rules to sam validate --lint

shashax42:

This PR adds a new feature and is not associated with a specific issue number.

The AWS SAM CLI's `validate --lint` command is useful for validating CloudFormation templates, but currently lacks the ability to apply rules specific to Serverless applications. Serverless applications have different characteristics compared to general CloudFormation templates, requiring additional validation rules tailored to their specific needs.

This PR adds a `--serverless-rules` option to the `sam validate --lint` command, allowing users to leverage rules from the cfn-lint-serverless package. This enables users to perform additional validations specific to Serverless applications. When enabled, this option validates Serverless application best practices such as Lambda function memory size, timeout, log retention, API Gateway stage logging, and throttling settings.

- This change does not affect existing functionality. The `--serverless-rules` option is only activated when explicitly specified.
- If the cfn-lint-serverless package is not installed, users receive a clear error message prompting them to install the package.
- When running in debug mode, additional information about Serverless Rules application and related settings is displayed.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/commands/validate/validate.py
```diff
@@ -57,20 +57,27 @@ class SamTemplate:
     "Create a cfnlintrc config file to specify additional parameters. "
     "For more information, see: https://github.com/aws-cloudformation/cfn-lint",
 )
+@click.option(
+    "--serverless-rules",
+    is_flag=True,
```
vicheey, line 62: Any particular region to enable this flag by default?

shashax42: @vicheey  No, there's no particular region requirement. You can enable this flag by default in all regions.

samcli/commands/validate/validate.py
```diff
@@ -170,20 +199,97 @@ def _lint(ctx: Context, template: str, template_path: str) -> None:
         cfn_lint_logger.propagate = True
         cfn_lint_logger.setLevel(logging.DEBUG)
 
+    print(f"Debug info: initial linter_config = {linter_config}")
```
valerena, line 202: Don't use `print` for debugging strings, use the `logging` module instead.

```
import logging

...

LOG.debug("Debug info: initial linter_config %s", linter_config)
```

Basic usage on this other file: https://github.com/aws/aws-sam-cli/blob/bdf9a7f19aff9064b6139c59c036cd12af01c539/samcli/lib/utils/codeuri.py#L5-L8 

https://github.com/aws/aws-sam-cli/blob/bdf9a7f19aff9064b6139c59c036cd12af01c539/samcli/lib/utils/codeuri.py#L30

Also, prefer using a message with the string format only, and the values as parameters of `debug` (like `"Debug info: initial linter_config %s", linter_config`) instead of formatting in the string directly (`f"Debug info: initial linter_config = {linter_config}"`). When `f" {parameter}"` is used for logging, the values are formatted even when logging is off, just adding extra processing that's not needed. (full details about how to use it in here: https://docs.python.org/3/library/logging.html#logging.Logger.debug)

samcli/commands/validate/validate.py
```diff
@@ -170,20 +199,97 @@ def _lint(ctx: Context, template: str, template_path: str) -> None:
...
+
+    # Initialize variable to handle both options together
+    rules_to_append = []
+    
+    # Support for previous serverless_rules option (deprecated)
+    if serverless_rules:
```
valerena, line 208: Just to emphasize what I mentioned before, don't add any of the `serverless_rules` code.

samcli/commands/validate/validate.py
```diff
@@ -57,20 +57,43 @@ class SamTemplate:
...
+    "Requires the cfn-lint-serverless package to be installed. "
+    "Use --extra-lint-rules=\"cfn_lint_serverless.rules\" instead. "
+    "For more information, see: https://github.com/awslabs/serverless-rules",
+)
+@click.option(
+    "--extra-lint-rules",
+    help="Specify additional lint rules to be used with cfn-lint. "
+         "Format: module.path (e.g. 'cfn_lint_serverless.rules')",
+    default=None
+)
```
valerena, line 73: I think you need to pass `multiple=True` to allow the parameter to be passed multiple times (https://click.palletsprojects.com/en/stable/options/#multiple-options and it will be stored as a list).

You mentioned in the documentation that that is actually supported, did you test passing it multiple times? (I imagine the last one will just replace any other one if we don't add the `multiple=True`)

samcli/commands/validate/validate.py
```diff
@@ -170,20 +199,97 @@ def _lint(ctx: Context, template: str, template_path: str) -> None:
...
         raise UserException(
             f"AWS Region was not found. Please configure your region through the --region option.\n{ex}",
             wrapped_from=ex.__class__.__name__,
         ) from ex
+    except Exception as e:
+        print(f"Debug info: exception occurred = {e}")
+        raise
 
     if not matches:
+        print(f"Debug info: template validation successful")
```
valerena, line 288: I don't think we need all these debugging messages though. Some make sense, but this part is already printing something else already.

tests/unit/commands/validate/test_cli.py
```diff
@@ -1,19 +1,24 @@
...
-from samcli.commands.validate.validate import do_cli, _read_sam_file, _lint, SamTemplate
+from samcli.commands.validate.validate import do_cli, _read_sam_file, _lint, SamTemplate, cli
 
 ctx_mock = namedtuple("ctx_mock", ["profile", "region"])
 ctx_lint_mock = namedtuple("ctx_lint_mock", ["debug", "region"])
 
 
 class TestValidateCli(TestCase):
+    def setUp(self):
+        # datetime.utcnow() 사용에 대한 경고 무시
```
valerena, line 19: Please include comments in English, thanks.

valerena (changes requested):

shashax42:
@valerena 
## Issue: Enhance --extra-lint-rules Feature in SAM CLI (Addressing Review Feedback)

### Description
This PR enhances the `--extra-lint-rules` feature in SAM CLI, addressing all feedback from valerena.

### Addressed Feedback
1. Added `multiple=True` flag to `--extra-lint-rules` option to allow specifying multiple rule modules
2. Completely removed all code and tests related to the never-officially-released `--serverless-rules` option
3. Replaced debug `print` statements with proper `logging` module usage
4. Changed all comments to English for consistency
5. Thoroughly updated test cases to reflect the new implementation

### Test Code Improvements
- Removed obsolete `test_serverless_rules_deprecated_with_extra_lint_rules` test
- Updated all test cases to handle `extra_lint_rules` as a tuple (supporting `multiple=True`)
- Removed all references to `serverless_rules` from test parameters
- Fixed assertions to match the new parameter structure
- Changed Korean test comments to English

### Implementation Details
- Users can now use the `--extra-lint-rules` option multiple times or specify multiple modules separated by commas
- Code is cleaner and more maintainable
- Test code now correctly validates the new interface

### Related PR
#7950

shashax42:
@valerena 
Will this PR be merged?

PR #7962, opened 2025-04-02
Address ruff PLW1508

mbfreder:

Address [PLW1508](https://docs.astral.sh/ruff/rules/invalid-envvar-default/). This is needed to merge https://github.com/aws/aws-sam-cli/pull/7947. Currently, some checks are failing on the PR.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container.py
+++ b/samcli/local/docker/container.py
@@ -39,7 +39,7 @@ from samcli.local.docker.utils import NoFreePortsError, find_free_port, to_posix
 
 LOG = logging.getLogger(__name__)
 
-CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT", 20))
+CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT", "20"))
 DEFAULT_CONTAINER_HOST_INTERFACE = "127.0.0.1"
 
 # Keep a lock instance to access the locks for individual containers (see dict below)
--- a/samcli/local/lambdafn/remote_files.py
+++ b/samcli/local/lambdafn/remote_files.py
@@ -30,7 +30,7 @@ def unzip_from_uri(uri, layer_zip_path, unzip_output_dir, progressbar_label):
         Label to use in the Progressbar
     """
     try:
-        get_request = requests.get(uri, stream=True, verify=os.environ.get("AWS_CA_BUNDLE", True))
+        get_request = requests.get(uri, stream=True, verify=os.environ.get("AWS_CA_BUNDLE", "True"))
 
         with open(layer_zip_path, "wb") as local_layer_file:
             file_length = int(get_request.headers["Content-length"])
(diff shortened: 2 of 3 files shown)


samcli/local/lambdafn/remote_files.py
```diff
@@ -30,7 +30,7 @@ def unzip_from_uri(uri, layer_zip_path, unzip_output_dir, progressbar_label):
         Label to use in the Progressbar
     """
     try:
-        get_request = requests.get(uri, stream=True, verify=os.environ.get("AWS_CA_BUNDLE", True))
+        get_request = requests.get(uri, stream=True, verify=os.environ.get("AWS_CA_BUNDLE", "True"))
```
valerena, line 33: This is actually wrong. 

`requests.get` (which behind the scenes just calls `requests.request`) has this `verify` parameter that can be either a path to a CA bundle to use **or a boolean** (https://requests.readthedocs.io/en/latest/api/#requests.request)

So it doesn't really accept a `"True"` string. 

If `ruff` is complaining about this (which kind of make sense because we're returning two different types from one `os.environ.get`), we can do maybe

```py
, verify=os.environ.get("AWS_CA_BUNDLE") or True)
```

Or just add a comment to ignore the rule for this line.

tests/unit/local/lambdafn/test_remote_files.py
```diff
@@ -36,7 +36,7 @@ def test_successfully_unzip_from_uri(
         path_patch.assert_called_with("layer_zip_path")
         path_mock.unlink.assert_called()
         unzip_patch.assert_called_with("layer_zip_path", "output_zip_dir", permission=0o700)
-        os_patch.environ.get.assert_called_with("AWS_CA_BUNDLE", True)
+        os_patch.environ.get.assert_called_with("AWS_CA_BUNDLE", "True")
```
valerena, line 39: I guess you'll have to change this test if you change the behavior of `os.environ.get` in the other file, but this is probably not it.

samcli/local/docker/container.py
```diff
@@ -39,7 +39,7 @@
 
 LOG = logging.getLogger(__name__)
 
-CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT", 20))
+CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT", "20"))
```
valerena, line 42: (Just as a small rant incoming, but this code is okay how you did it)

Conceptually this is weird, but I imagine this is the easiest way to do it.

Because we're passing a `"20"` (string) that will be converted to a number with `float(..)`.. But in an ideal world we just want to pass the number right away instead of creating the intermediate string. 

What we would want is to do something like

```
CONTAINER_CONNECTION_TIMEOUT = float(os.environ.get("SAM_CLI_CONTAINER_CONNECTION_TIMEOUT")) if "SAM_CLI_CONTAINER_CONNECTION_TIMEOUT" in os.environ else 20
```
(check if the envvar exists, and only in that case convert to float, otherwise pass the number directly)

but that's terrible code too, so we can just keep it with what you did here.

vicheey:
Address in this https://github.com/aws/aws-sam-cli/pull/7947

PR #7986, opened 2025-04-14
fix: make pr failure for python 3.12+

vicheey:
Make failure with following error when using python 3.12+
```
tests/end_to_end/test_runtimes_e2e.py:1: error: Cannot find implementation or library stub for module named "distutils.dir_util"  [import-not-found]
tests/end_to_end/test_runtimes_e2e.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

Make pr failure.

The distutils module was deprecated in Python 3.10 and removed in Python 3.12. Therefore, distutils.dir_util is no longer available in Python 3.12 or later versions.
The recommended replacement for distutils.dir_util is the shutil module, which provides similar functionality and is still actively maintained in Python.

 The error was in the `tests/end_to_end/test_runtimes_e2e.py` file. The `mypy` type checker was failing because it couldn't find the implementation or library stub for the `distutils.dir_util` module.

Changes made:
   • Removed the import of `copy_tree` from `distutils.dir_util`
   • Added an import for `shutil` instead
   • Modified the function call to use `shutil.copytree` with the `dirs_exist_ok=True` parameter to ensure it works correctly

No. Update the test file. 

Verified success with python 3.11.11 and 3.13.0

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/end_to_end/test_runtimes_e2e.py
+++ b/tests/end_to_end/test_runtimes_e2e.py
@@ -1,4 +1,3 @@
-# This avoids the type incompatibility issue
 import shutil
 import os
 from unittest import skipIf
--- a/tests/end_to_end/test_runtimes_e2e.py
+++ b/tests/end_to_end/test_runtimes_e2e.py
@@ -1,10 +1,10 @@
-from distutils.dir_util import copy_tree
+# Use shutil.copytree directly instead of distutils.dir_util.copy_tree
+# This avoids the type incompatibility issue
+import shutil
+import os
 from unittest import skipIf
-
 import json
 from pathlib import Path
-
-import os
 from parameterized import parameterized_class
 
 from tests.end_to_end.end_to_end_base import EndToEndBase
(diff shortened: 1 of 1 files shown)


Vandita2020:
More explanation: 
The distutils module was deprecated in Python 3.10 and removed in Python 3.12. Therefore, `distutils.dir_util` is no longer available in Python 3.12 or later versions.
The recommended replacement for `distutils.dir_util` is the `shutil` module, which provides similar functionality and is still actively maintained in Python.

Vandita2020 (approved):

tests/end_to_end/test_runtimes_e2e.py
```diff
@@ -1,10 +1,10 @@
-from distutils.dir_util import copy_tree
+# Use shutil.copytree directly instead of distutils.dir_util.copy_tree
+# This avoids the type incompatibility issue
```
valerena, line 2: This comment probably doesn't need to be here (you can add it to the commit message if you want). It talks about the change you're making, not about the existing code. It doesn't make sense to have a comment about `distutils`, when it's not in the code anymore.

Also, we already use `shutil.copytree` in other parts of the code.

tests/end_to_end/test_runtimes_e2e.py
```diff
@@ -1,10 +1,10 @@
-from distutils.dir_util import copy_tree
+# Use shutil.copytree directly instead of distutils.dir_util.copy_tree
```
vicheey, line 1:
```suggestion
```

tests/end_to_end/test_runtimes_e2e.py
vicheey, line 2:
```suggestion
```

valerena (approved):

PR #8007, opened 2025-05-02
feat(metrics): collect the DOCKER_HOST environment variable path

pendo324:
It allows SAM CLI to have a proxy measurement on which tools users are using as its backend. This allows the SAM CLI team to prioritize testing and features from various tools which users may be using to provide their `DOCKER_HOST`.

This change collects the last part of URIs/paths, in hope that these last parts shed light on which tool the user is using with SAM CLI. For example, if a user is using a local version of docker, this new metric would collect `docker.sock`. This is only checking the last part of paths in an effort to not needlessly collect what may be private URLs to remote docker hosts which user's might be using.

None?

#### Mandatory Checklist

- [N/A] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [N/A] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [N/A] Write/update unit tests
- [N.A] Write/update functional tests if needed
- [x] `make pr` passes
- [N/A] Write documentation


valerena:
Thanks for the contribution! There are a few errors on `make pr` (and in other tests), if you can take a look.

At least in `make pr`
```
samcli/lib/telemetry/metric.py:495: error: Missing positional argument "path" in call to "exists"  [call-arg]
samcli/lib/telemetry/metric.py:497: error: Returning Any from function declared to return "str"  [no-any-return]
samcli/lib/telemetry/metric.py:500: error: Returning Any from function declared to return "str"  [no-any-return]
```

The first one is because of `if os.path.exists():` (I imagine there's a variable missing there, probably `parsed.path`)

The other ones might be just errors on `typing` with `os.path.basepath` 🤔. You might be able to just do `str(..)` to fix those.

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,28 @@ def _get_execution_environment(self) -> str:
...
+        URLs and local file system paths. If the DOCKER_HOST is not set, returns an empty string.
+
+        Examples:
+            - unix:///var/run/docker.sock -> docker.sock
+            - tcp://localhost:1234 -> localhost:1234
+            - /var/run/docker.sock -> docker.sock
+        """
+
+        parsed = urlparse(self._gc.docker_host)
+        if not parsed.scheme == "":
```
valerena, line 494: Did you mean the opposite here? I imagine that **if** the schema is empty, then I can check if it's a file that exists, but here you're doing that if it's **not** empty. If it happens that it's not empty, then we could be calling `os.path.exists("unix:///var/run/docker.sock")` which returns False.

valerena (changes requested):
There are some more [tests failing](https://github.com/aws/aws-sam-cli/actions/runs/15144876368/job/42578032835?pr=8007) on [test_installed_metrics.py](https://github.com/aws/aws-sam-cli/blob/8473d598a88ee5e309cb199fa98f7a09560d9491/tests/integration/telemetry/test_installed_metric.py#L43-L53) 

I added another comment to confirm the logic. 

You can run `pytest -vv tests/integration/telemetry` to check those. They don't run with the standard `make pr` because they're integration tests, but they should run quickly (as opposed to other integration tests which could take hours)

samcli/lib/telemetry/metric.py
pendo324: You're correct, but honestly I think this logic was a little too overcomplicated anyway. I've simplified and removed the conditional in the latest revision.

The integ tests are also passing on my local machine, just waiting for GitHub CI to trigger

pendo324:
@valerena, is the failing Windows smoke test related to this change? If so, I'm not sure how and could use some guidance on how to fix it

mbfreder:
There are some failing tests (related to logs cli command), although they don't seem to be related to the changes on this PR.

vicheey:
This PR introduce new metric dimension to data. We need to update the data schema first before merging this PR.

vicheey:
Ran `pytest -vv tests/integration/telemetry` and confirm success: 
```
 pytest -vv tests/integration/telemetry --tb=short
 ⋮ 
 ↳ Purpose: Run all integration telemetry tests

============================= test session starts ==============================
platform darwin -- Python 3.11.12, pytest-8.4.1, pluggy-1.5.0 -- /Users/vichym/.pyenv/versions/3.11.12/envs/aws-aws-sam-cli-env-3.11/bin/python
cachedir: .pytest_cache
hypothesis profile 'default'
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
...

tests/integration/telemetry/test_telemetry_contract.py::TestTelemetryContract::test_must_not_send_metrics_if_disabled_using_envvar 
-------------------------------- live log call ---------------------------------
INFO     werkzeug:_internal.py:97 127.0.0.1 - - [27/Aug/2025 23:58:52] "POST /metrics HTTP/1.1" 200 -
INFO     werkzeug:_internal.py:97 127.0.0.1 - - [27/Aug/2025 23:58:52] "POST /metrics HTTP/1.1" 200 -
PASSED                                                                   [ 92%]
tests/integration/telemetry/test_telemetry_contract.py::TestTelemetryContract::test_must_send_metrics_if_enabled_via_envvar 
-------------------------------- live log call ---------------------------------
INFO     werkzeug:_internal.py:97 127.0.0.1 - - [27/Aug/2025 23:58:56] "POST /metrics HTTP/1.1" 200 -
INFO     werkzeug:_internal.py:97 127.0.0.1 - - [27/Aug/2025 23:58:56] "POST /metrics HTTP/1.1" 200 -
PASSED                                                                   [100%]

======================== 12 passed, 2 skipped in 36.17s ========================
```

pendo324:
Thanks for the fixes, @vicheey. LGTM

samcli/lib/telemetry/metric.py
```diff
@@ -237,6 +237,9 @@ def _send_command_run_metrics(ctx: Context, duration: int, exit_reason: str, exi
         metric_specific_attributes["projectName"] = get_project_name()
         metric_specific_attributes["initialCommit"] = get_initial_commit_hash()
 
+    # Add container host information for all command runs
+    metric_specific_attributes["containerHost"] = Metric("")._get_container_host()
```
valerena: why creating a new `Metric("")` object instead of using the existing `metric`? 

```suggestion
    metric_specific_attributes["containerHost"] = metric._get_container_host()
```

samcli/lib/telemetry/metric.py
```diff
@@ -436,6 +439,7 @@ def _add_common_metric_attributes(self):
         self._data["installationId"] = self._gc.installation_id
         self._data["sessionId"] = self._session_id
         self._data["executionEnvironment"] = self._get_execution_environment()
+
```
valerena: minor: remove this line, just because it doesn't make sense to have a random empty line without a reason

samcli/lib/telemetry/event.py
```diff
@@ -283,6 +283,9 @@ def _send_events_in_thread():
             msa["events"] = [e.to_json() for e in EventTracker._events]
             EventTracker._events = []  # Manual clear_trackers() since we're within the lock
 
+        # Add container host information for all event runs
+        msa["containerHost"] = Metric("")._get_container_host()
```
valerena: Same comment as in `telemetry/metric.py`, instead of creating a new `Metric("")` that we doesn't matter, we should call `_get_container_host()` on the `metric = Metric("events")` that's created a little bit down below. This has a different order than `telemetry/metric.py` file, but the concept is the same.

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +480,22 @@ def _get_execution_environment(self) -> str:
...
+        Examples:
+            - unix:///var/run/docker.sock -> docker.sock
+            - tcp://localhost:1234 -> localhost:1234
+            - /var/run/docker.sock -> docker.sock
+        """
+
+        try:
+            return PurePath(self._gc.docker_host).name
+        except TypeError:
+            return ""
```
seshubaws: Can we log the error?

vicheey: I updated code to translate container host values directly to engine names. This is outdated.

samcli/lib/telemetry/metric.py
vicheey: Updated to using existing metric object.

samcli/lib/telemetry/event.py
vicheey: Move the assignment to later line after creating metric object and use that instead.

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,48 @@ def _get_execution_environment(self) -> str:
...
+
+        # Socket mappings
+        socket_map = {"finch.sock": "finch", "podman.sock": "podman", "docker.sock": "docker"}
+
+        # Path-specific mappings (checked first for specificity)
+        path_map = {"colima": "colima", "lima": "lima", "orbstack": "orbstack", ".rd": "rancher-desktop"}
+
+        # Check path-specific patterns first
+        for path, engine in path_map.items():
+            if path in docker_host and "docker.sock" in docker_host:
```
valerena: I don't know if it's too important, but we should probably try to make this check `path in docker_host` more specific? I feel like someone can randomly have a path with `climate` for some reason that will be considered as `lima`, or something like that. Can we check for `/lima/` maybe instead of just `lima` (or is it `.lima`?), or something along those lines?

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,48 @@ def _get_execution_environment(self) -> str:
...
+            return "docker-default"
+
+        docker_host = self._gc.docker_host.lower()
+
+        # Socket mappings
+        socket_map = {"finch.sock": "finch", "podman.sock": "podman", "docker.sock": "docker"}
+
+        # Path-specific mappings (checked first for specificity)
+        path_map = {"colima": "colima", "lima": "lima", "orbstack": "orbstack", ".rd": "rancher-desktop"}
+
```
valerena: Very minor thing, and it's okay if you don't want to change it, but I feel like if we're checking this first, then we should also **define** this variable first? It's confusing to define `socket_map` and then `path_map`, but then use `path_map` first.

samcli/lib/telemetry/metric.py
vicheey: Update the search terms to include `.` and `/`. 
```
".colima/": "colima",
 ".lima/": "lima", 
".orbstack/": "orbstack", 
".rd/": "rancher-desktop"
```

samcli/lib/telemetry/metric.py
vicheey: Reordered.

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,64 @@ def _get_execution_environment(self) -> str:
...
+        """
+        Returns a normalized container engine identifier based on DOCKER_HOST.
+        Maps various DOCKER_HOST patterns to standardized container engine names.
+
+        Translation:
+            # Case 1: Not set
+            - (unset)	                                -> ""(empty)	            -> docker-default
+
+            # Case 2: Check Path-specific
+            - unix://~/.colima/default/docker.sock	    -> colima.sock	            -> colina
```
valerena: typo `colina`

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,64 @@ def _get_execution_environment(self) -> str:
...
+            - unix://~/.lima/default/sock/docker.sock	-> lima.sock	            -> lima
+            - unix://~/.rd/docker.sock                  -> 	rancher-desktop.sock	-> rancher-desktop
+            - unix://~/.orbstack/run/docker.sock        -> 	orbstack.sock	        -> orbstack
```
valerena: nit: extra space before `rancher-desktop-.sock` and `orbstack.sock`

samcli/lib/telemetry/metric.py
```diff
@@ -476,6 +479,64 @@ def _get_execution_environment(self) -> str:
...
+            return "docker-default"
+
+        docker_host = self._gc.docker_host.lower()
+
+        # Path-specific mappings (checked first for specificity)
+        path_map = {".colima/": "colima", ".lima/": "lima", ".orbstack/": "orbstack", ".rd/": "rancher-desktop"}
+
+        # Check path-specific patterns first
+        for path, engine in path_map.items():
+            if path in docker_host and "docker.sock" in docker_host:
```
valerena: and just to be even safer, should this be `and docker_host.endswith("docker.sock")` ? Could it be that `docker.sock` is not the end of the value?

vicheey: Should be always at the end.

samcli/lib/telemetry/metric.py
vicheey: Remove the column.

samcli/lib/telemetry/metric.py
vicheey: Updated to use `endWith()`

Issue #7927, opened 2025-03-10
Bug: Error with binary files sent through api


mjkubba:
### Description:
When sending file (postman or frontend) to backend lambda through API Gateway I'm getting an error while doing local testing with sam local start-api    
While the deployed application behave normally with an actual API Gateway and Lambda

### Steps to reproduce:
template:

```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  GAPP-debug

Globals:
  Function:
    Timeout: 360

Resources:
  gappbeapi:
    Type: AWS::Serverless::HttpApi
    Properties:
      CorsConfiguration:
        AllowOrigins:
          - "*"
        AllowMethods:
          - GET
          - POST
          - OPTIONS
        AllowHeaders:
          - Content-Type
          - Accept
          - Access-Control-Allow-Headers
          - Access-Control-Request-Method
          - Access-Control-Request-Headers
          - Authorization
  gappbe:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: fulfillment/ful_deployment_package.zip
      Handler: index.lambda_handler
      Runtime: python3.13
      Architectures:
        - x86_64
      Events:
        HelloWorld:
          Type: HttpApi
          Properties:
            Path: /sendData
            Method: POST
            ApiId: !Ref gappbeapi
      Policies:
      - Statement:
        - Sid: lambda
          Effect: Allow
          Action:
          - lambda:InvokeFunction
          Resource: arn:aws:lambda:us-east-1:*:function:*
```

python code example:

```
# index.py


def lambda_handler(event, context):

    print("do things here")
    try:
        if not event:
            raise ValueError("Empty event object")



        return "All good"

    except Exception as e:
        return "Bad response"
```

curl:
```
curl --location 'http://127.0.0.1:3000/sendData' \
--form 'cxName="postman"' \
--form 'email="emailpost"' \
--form 'vafile=@"/Users/xxxx/Downloads/somefile.pdf"'
```

### Observed result:
```
○ → sam local start-api --debug
2025-03-10 09:20:34,406 | No config file found in this directory.
2025-03-10 09:20:34,408 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/Users/mjkubba/workspace/debug/samconfig.toml'
2025-03-10 09:20:34,409 | Config file location: /Users/mjkubba/workspace/debug/samconfig.toml
[...]

PR #8018, opened 2025-05-12
fix: sam local start api reques to correclty support  binary files in request

vicheey:
https://github.com/aws/aws-sam-cli/issues/7927

From original issue: 
> When sending file (postman or frontend) to backend lambda through API Gateway I'm getting an error while doing local testing with sam local start-api
```
UnicodeDecodeError while processing HTTP request: 'utf-8' codec can't decode byte 0xb5 in position 473: invalid start byte.
```
Update logic in event_constructor.py to decode request for various cases. . 

- This change is mostly additive and should not break existing functionality
- It may change behavior for edge cases, but in ways that align better with API Gateway's actual behavior
- The core logic for common cases remains the same

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/apigw/event_constructor.py
+++ b/samcli/local/apigw/event_constructor.py
@@ -54,12 +54,14 @@ def construct_v1_event(
     if is_base_64:
         LOG.debug("Incoming Request seems to be binary. Base64 encoding the request data before sending to Lambda.")
         request_data = base64.b64encode(request_data)
-
-    if request_data:
-        # Flask does not parse/decode the request data. We should do it ourselves
-        # Note(xinhol): here we change request_data's type from bytes to str and confused mypy
-        # We might want to consider to use a new variable here.
-        request_data = request_data.decode("utf-8")
+        request_data = request_data.decode("utf-8") if request_data else ""
+    elif isinstance(request_data, bytes):
+        try:
+            request_data = request_data.decode("utf-8")
+        except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


vicheey:
Want for https://github.com/aws/aws-sam-cli/pull/8019 to merge. This contains some fix for failing makr pr test.

vicheey:
Waiting for https://github.com/aws/aws-sam-cli/pull/8022 to merge before rerun the workflow.

valerena:
I'm curious how much manual testing you did here, because the unit tests seem nice, but I don't really know if there are cases where the API request actually looks different than the mocked requests data that you added. Did you try this with `curl`/`wget` or another mechanism to actually send a binary file and confirmed that the file was received correctly? And tried with non-binary files, to confirm that we didn't break any of those? 

Ideally you can run the `local start-api` integration tests in `tests/integration/local/start_api/test_start_api.py` to confirm that nothing broke.

But ideally we should have had a test **before** that should have failed because this behavior wasn't working. We do have some integ tests that check with a binary file, so it's weird that no test was catching this problem. https://github.com/aws/aws-sam-cli/blob/8473d598a88ee5e309cb199fa98f7a09560d9491/tests/integration/local/start_api/test_start_api.py#L622

roger-zhangg:
overall looks good, but can we keep the original test while adding new ones?

Issue #7780, opened 2024-12-08
Bug: Inconsistent examples of using the shell commands


miketheman:
### Description:

When running any `--help` command, the examples should reflect what a user would actually copy/paste, and there are some that do not display a space between the `$` representing a shell prompt and the actual command.

Basically, any time there's a command, either remove the preceding `$` or add a space.

### Steps to reproduce:

`sam --help`

or

`sam sync --help`

(This is not a comprehensive audit of all command examples)

### Observed result:

```
...shellsession
Examples:

    Get Started:        $sam init
```

```shellsession
...
  By default, `$sam sync` runs a full AWS Cloudformation stack update.
...

Examples:

    $sam sync --watch --stack-name {stack}

    $sam sync --code --watch --stack-name {stack}

    $sam sync --code --stack-name {stack} --resource-id {ChildStack}/{ResourceId}
...



### Expected result:
<!-- Describe what you expected. -->

```shellsession
Examples:

    Get Started:        $ sam init
```

```shellsession
Examples:

    $ sam sync --watch --stack-name {stack}

    $ sam sync --code --watch --stack-name {stack}

    $ sam sync --code --stack-name {stack} --resource-id {ChildStack}/{ResourceId}

```

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
{
  "version": "1.131.0",
  "system": {
    "python": "3.13.0",
    "os": "macOS-15.1.1-x86_64-i386-64bit-Mach-O"
  },
  "additional_dependencies": {
    "docker_engine": "Not available",
    "aws_cdk": "Not available",
    "terraform": "1.5.7"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}```

vicheey:
@miketheman , Thank you for your feedback. We'll look into this.

vicheey:
We will track the improvement internally to improve our help section for sam cli.

nitintecg:
@vicheey can i contribute to this issue ? if its available

miketheman:
Resolved by #8038

PR #8038, opened 2025-05-17
fix: inconsistent examples of shell commands by adding space after $ symbol

sriram-mv:
#7780

Pretty simple change that improves UX.

Makes commands copy pastable.

N/A

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/commands/sync/command.py
+++ b/samcli/commands/sync/command.py
@@ -77,7 +77,7 @@ HELP_TEXT = """
 """
 
 DESCRIPTION = """
-  By default, `$sam sync` runs a full AWS Cloudformation stack update.
+  By default, `$ sam sync` runs a full AWS Cloudformation stack update.
 
   Running `sam sync --watch` with `--code` will provide a way to run just code
   synchronization, speeding up start time skipping template changes.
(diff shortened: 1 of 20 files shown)


vicheey:
Thank you. The integration test failure for Python 3.11 is due to Click library. We can ignore that for now as team work on the fix.

vicheey (approved):

vicheey:
Waiting for fix.

samcli/commands/sync/command.py
```diff
@@ -77,15 +77,15 @@
 """
 
 DESCRIPTION = """
-  By default, `$sam sync` runs a full AWS Cloudformation stack update.
+  By default, `$ sam sync` runs a full AWS Cloudformation stack update.
```
valerena, line 80: I feel like all of these that are inline on a longer paragraph shouldn't even have the `$` in the first place (like the "`sam sync --watch`" in the following line), but it can be left for a future PR

valerena (approved):

samcli/commands/sync/command.py
sriram-mv: I like that, it makes sense!

Issue #1699, opened 2020-01-02
import existing resources into new stack


gdw2:
### Describe your idea/feature/enhancement

I wish SAM CLI could import existing CF resources as documented here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-new-stack.html#resource-import-new-stack-cli

Specific use case: Destroying a stack leaves some persistent resources (e.g. a database) that could be incorporated into a new stack.

### Proposal

Add a `--resources-to-import` argument to `sam deploy`.

maxmil:
I'm also interested in this. My SAM app is getting large and I would like to break it into several nested apps. I think that I need this feature in order to do so.

aahung:
`create-change-set` (SAM CLI uses this API to create or update stacks) supports "resources-to-import" option, I think it would be easy to implement

icysharp:
@aahung Is this feature already available or being actively worked on?

andreacfm:
@aahung any update?

noisydump:
Hi, is this feature expected in any release?

mericozkayagan:
Any updates ?

softprops:
I think we can do better than asking engineers to construct this file by hand themselves. That’s one of the major dx flaws in the cf interface. I would love it sam could generate on through a guided workflow and then orchestrate the workflow to apply the changes

kitsunde:
I had an unfortunate incident where I was deleting a stack set while working on getting a resource imported, but instead deleted the entire stack in the UI which are just slightly above the button.

I'm not blaming SAM for the incident, but it's unfortunate that the import support with SAM is harder than dealing with a normal CloudFormation import because of the transform and there's no tooling support to make it easier.

cortexcompiler:
There is definitely a need for this.  Example resources that often have `Retain` for the `DeletionPolicy` (especially in test or prod environments) are: LogGroups, DynamoDB tables, RDS..

For LogGroups I believe CDK has baked in a Custom Resource to handle "adopting" the LogGroup if it already exists.  However, there is not solution for database tables and SAM does not even handle LogGroups.  I think it would be great if SAM could navigate the more complex underlying cloudformation APIs for us to do this, in the same way it simplifies serverless resource creation.  TBH, it often feels like the overall developer experience is not considered when the various AWS teams create their APIs.

The following announcement is promising, but to use the "ImportExistingResources" capability you need to use the cloudformation CLI to manage changesets:
[...]

PR #8114, opened 2025-06-18
feat: Add flag for import existing resources

electrofelix:
Pass through option for importing existing resources which allows using
Retain for DeletionPolicy on resources to avoid removal on data
resources if stack needs to be deleted and recreated.

Fixes: #1699


electrofelix:
Main use case I have for this is that sometimes the stack gets into a funk and needs to be deleted and recreated, however if there are data resources in the same application stack specifying `DeletionPolicy: Retain` prevents re-deploying via sam.

The workarounds are to move data resources out to separate stacks, which can work for event buses and DBs, but is a pain when working with queues. End up needing to pass a multitude of resource identifiers in via parameters.

Using this approach, I can mark queues, event buses (+ archives), and any DBs to be retained and be able to teardown and redeploy without dropping messages/events/data, and at the same time avoid needing a complex orchastrated deployment spanning multiple repositories or needing complex CI jobs to handle what stacks to delete vs deploy.


Note: when testing I noticed CloudFormation doesn't handle automatically re-importing `AWS::Logs::LogGroup` resources particular well.

I have a fairly rough script that does the same as this code the AWS CLI, which encountered the same issue suggesting there may be a bug in Cloudformation. Appears it fails to remove some information from resources associated with the original stack meaning they still appear to be part of a stack and it refuses to import. Solving this in Cloudformation would automatically fix it for this feature as well.


It would be nicer for it to be available built in rather than needing to run the script sometimes:
```
TEMP_TEMPLATE=$(mktemp -t template.yaml) || exit 1

stack_name=$(toml get --toml-path ${configName} ${sectionName}.global.parameters.stack_name)
capabilities=$(toml get --toml-path ${configName} ${sectionName}.global.parameters.capabilities)
parameter_overrides=$(toml get --toml-path ${configName} ${sectionName}.global.parameters.parameter_overrides | sed -e 's/^\[//g; s/\]$//g')
params=""
for param in ${parameter_overrides}
do
    param=${param%,}
    param=${param%\'}
    param=${param#\'}
    key=${param%%=*}
    value=${param#*=}
    params="${params}${params:+ }ParameterKey=${key},ParameterValue=${value}"
done

# make sure all aws commands use the requested profile
export AWS_PROFILE=$profile

echo "Packaging stack and then using cloudformation to import existing using config file: '$configName' ..."
command="sam package --config-file $configName --output-template-file ${TEMP_TEMPLATE}"
echo "Executing $command"
$command


command="aws cloudformation describe-stacks --stack-name "${stack_name}""
echo "Checking if stack exists"
echo "Executing $command"
if ! $command >/dev/null 2>&1; then
    change_set_type=CREATE
    stack_complete_command=stack-create-complete
else
    change_set_type=UPDATE
    stack_complete_command=stack-update-complete
fi

change_set_name="change-set-$(date +%s)"
command="aws cloudformation create-change-set --stack-name ${stack_name} --change-set-type ${change_set_type} \
 --change-set-name ${change_set_name} --template-body "file://${TEMP_TEMPLATE}" --capabilities ${capabilities} \
 --parameters ${params} --import-existing-resources"
echo "Executing $command"
$command

sleep 2
echo "Waiting for changeset to be complete"
command="aws cloudformation wait change-set-create-complete --stack-name ${stack_name} --change-set-name ${change_set_name}"
echo "Executing $command"
$command
if [[ $? -ne 0 ]]; then
    failure_reason=$(aws cloudformation describe-change-set --stack-name ${stack_name} --change-set-name ${change_set_name} --query "StatusReason" --output text)
    if [[ "${failure_reason}" == *"The submitted information didn't contain changes."* ]] || [[ "${failure_reason}" == *"No updates are to be performed."* ]]; then
        echo "No changes detected, exiting"
        exit 0
    else
        echo "Changeset failed to create: ${failure_reason}"
        exit 1
    fi
fi
exit 0
```

vicheey:
Thank you for your contribution, @electrofelix. The current code implementation is clear and concise. We just need to verify both forward and backward deployment scenarios that this update would not leave a stack in an unrecoverable state if deployment fail. Would you be able to enhance this further by adding integration tests to cover your specific deployment use case as well?

electrofelix:
@vicheey sure, I presume it's just follow the existing integration tests such as https://github.com/aws/aws-sam-cli/blob/develop/tests/integration/deploy/test_deploy_command.py#L1542-L1593 going with the following steps:

* deploy template with some resources marked retain (eventbus + archive, sqs, & dynamodb)
* delete stack
* re-deploy same template and assert expected resources have import in `deploy_process_execute.stdout.strip()`

Presumably some work around tear will be needed for this scenario? Extend [add_left_over_resources_from_stack](https://github.com/aws/aws-sam-cli/blob/a5282b6b8d7a64c5a9b85cc54223af22b58ea883/tests/integration/deploy/deploy_integ_base.py#L98) to cover the additional resource types added?

Is this better as a separate test file to cover the import functionality under the [deploy tests](https://github.com/aws/aws-sam-cli/tree/a5282b6b8d7a64c5a9b85cc54223af22b58ea883/tests/integration/deploy)?


Might be a few weeks before I will get back to updating

vicheey:
Do you mind help with the `make pr` failure?

electrofelix:
> Do you mind help with the `make pr` failure?

Not at all, just delayed getting back to it

c-ameron:
Thanks so much for this @electrofelix !

@vicheey is this up for consideration? This would help me a lot move from terraform to SAM for some resources.

Thanks!

vicheey:
Hi @electrofelix, 

I've pushed a small follow-up commit on top of your work:
- Fixed a minor typo in the `--import-existing-resources` help text (missing word "resources")
- Added an integration test that creates a pre-existing SSM parameter, deploys with `--import-existing-resources`, and verifies the changeset shows the `Import` action
- Added the `import_existing_resources` parameter to the integration test base helper
CI is running now. Will follow up once results are in.

@c-ameron, 
We're working on getting this merged. I'll need another team member to review and hopefully get this out soon.
Stay tuned!

vicheey:
The integration test: https://github.com/aws/aws-sam-cli/actions/runs/23670588901

tests/unit/commands/samconfig/test_samconfig.py
```diff
@@ -969,6 +969,7 @@ def test_deploy(self, do_cli_mock, template_artifacts_mock1, template_artifacts_
                 "default",
                 False,
                 True,
+                False,
```
seshubaws, line 972: can we add a test for reading this param from the samconfig?

tests/integration/deploy/test_deploy_import_existing_resources.py
```diff
@@ -0,0 +1,62 @@
...
+TEMPLATE = """\
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Template for testing --import-existing-resources flag.
+Resources:
+  TestLogGroup:
+    DeletionPolicy: Retain
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: {log_group_name}
+"""
```
reedham-aws, line 21: I think in other places in the code, we don't have the template inline, but rather in `tests/integration/testdata`

PR #8177, opened 2025-07-18
test(canary): Call Lambda function to reset test account instead of making StepFunctions api call

valerena:
Modify the end of the canary to call Lambda function instead of sending a task action directly to Step Functions

Also update Maven to 3.9.11 (Maven 3.9.10 is not available in that location anymore)

Updated some indentation in some Windows files.

- As part of a canary infrastructure update, the new credentials used to run the canary can no longer make a call to StepFunctions directly, so we added a new Lambda function to wrap the account reset action. This Lambda function calls Step Functions internally so the outcome is the same.

- Changing a direct Step Functions call with a Lambda function call.

- The output of the function is put into a file, and then we print that file (that's how Lambda gives you the result of a function call)

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-ubuntu.yml
+++ b/appveyor-ubuntu.yml
@@ -47,9 +47,9 @@ install:
   - sh: "ls /usr/"
   # install latest maven which is compatible with jdk17
   - sh: "sudo apt-get -y remove maven"
-  - sh: "wget https://dlcdn.apache.org/maven/maven-3/3.9.10/binaries/apache-maven-3.9.10-bin.zip -P /tmp"
+  - sh: "wget https://dlcdn.apache.org/maven/maven-3/3.9.11/binaries/apache-maven-3.9.11-bin.zip -P /tmp"
   - sh: "sudo unzip -d /opt/mvn /tmp/apache-maven-*.zip"
-  - sh: "PATH=/opt/mvn/apache-maven-3.9.10/bin:$PATH"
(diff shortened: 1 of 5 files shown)


appveyor-ubuntu.yml
```diff
@@ -157,7 +157,10 @@ on_finish:
   - sh: 'export AWS_SECRET_ACCESS_KEY=$CI_ACCESS_ROLE_AWS_SECRET_ACCESS_KEY'
   - sh: 'export AWS_SESSION_TOKEN=$CI_ACCESS_ROLE_AWS_SESSION_TOKEN'
 
-  - sh: 'aws stepfunctions send-task-success --task-token "$TASK_TOKEN" --task-output "{}" --region us-west-2'
+    # When we update to AWS CLI v2, we need to add the `--cli-binary-format raw-in-base64-out` option to pass the payload as a JSON string
+  - sh: >
+      aws lambda invoke --function-name "$ACCOUNT_RESET_LAMBDA_ARN" --payload "{\"taskToken\": \"$TASK_TOKEN\", \"output\": \"{}\"}" ./lambda-output.txt --region us-west-2
+  - sh: 'cat ./lambda-output.txt'
```
roger-zhangg, line 163: Could you help to include why this is required?

valerena: I updated the PR description with some more information

vicheey (approved):
LGTM

PR #8196, opened 2025-07-30
feat(local): add --no-watch flag to disable hot reload

roger-zhangg:

#8782

The AWS SAM CLI file watcher caused unstable behavior during local development with `--warm-containers`:

- **Multiple Lambda functions sharing a `CodeUri`** — a single change in any function falsely invalidated *all* of them, killing every running warm container and demoting back to LAZY mode. Reproed in LoopBack apps that auto-create temporary `.sandbox` directories (8+ minute startup times).
- **High-overhead file watching environments** — Microsoft Defender on Windows, large monorepos, etc. make the file observer dominate CPU/IO and effectively block local invocation. See #8782.

Adds a `--no-watch` flag to `sam local start-api` and `sam local start-lambda`. When supplied with `--warm-containers`, SAM CLI skips creating any file observers (template watcher and the per-function `LambdaFunctionObserver`). Local code or template changes will not restart the running container — stop and rerun the command to pick up changes.

The flag is forwarded as a kwarg through `RefreshableSamFunctionProvider` and `WarmLambdaRuntime`. With `no_watch=True`, no observer is constructed, and every `watch` / `unwatch` / `start` / `stop` call becomes a guarded no-op.

Also adds debug logging around `LambdaFunctionObserver` lock acquire / resource processing to help triage future watcher issues.

- `--no-watch` only takes effect with `--warm-containers`. When supplied without warm containers, the flag is ignored and a `WARNING` is logged. The cold-mode `SamFunctionProvider` (which doesn't accept `no_watch`) never receives the flag — addresses the earlier review concern about an uncaught error in that path.
- New kwarg `no_watch: Optional[bool] = False` on `InvokeContext`, `RefreshableSamFunctionProvider`, and `WarmLambdaRuntime`. Defaults preserve existing behavior; no caller is required to update.
- Refreshes a stale `WarmLambdaRuntime.__init__` docstring (`warm_containers` → current parameters) flagged in earlier review.
- New `no_watch` entry in `schema/samcli.json` for both `local_start_api` and `local_start_lambda`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


samcli/lib/utils/file_observer.py
```diff
@@ -415,6 +424,17 @@ class SingletonFileObserver(metaclass=Singleton):
     A Singleton class that will observe some file system paths for any change for multiple purposes.
     """
 
+    # Ignore patterns for file watching to avoid infinite loops and unnecessary events
+    DEFAULT_IGNORE_PATTERNS = [
```
vicheey: Should we provide a way for customer to provide customize ignore pattern? If none is provided, we can use the DEFAULT_IGNORE_PATTERNS.

vicheey (dismissed):

samcli/lib/utils/file_observer.py
roger-zhangg: Yes this should be the idea when we ship it. I will use this version to debug with customer first

reedham-aws (dismissed):

vicheey:
DO NOT MERGE.

roger-zhangg:
This failing test seems unrelated, but will take a look

roger-zhangg:
also passed on my local
```
(.venv) ruojiazh@842f577b9ad2 aws-sam-cli % pytest tests/smoke/test_all_commands.py::TestAllCommands -n 4
======== 510 passed in 111.96s (0:01:51) ========
```

reedham-aws:
What is the reason that we no longer allow customers to provide a list of files/dirs to be ignored? It seems like that was the original intention.

roger-zhangg:
> What is the reason that we no longer allow customers to provide a list of files/dirs to be ignored? It seems like that was the original intention.

The root cause was actually not that. Originally the assumed root cause was few node lib is constantly changing when function is invoking and therefore triggering our file watcher, but then it turns out it's just the file watcher is somehow lagging and it picks up the filechanges way later than it should be. So the lib changes are done in build, which should be a legit change that should be picked up. 

For what here is fixing, is in some sam project setup, it might have multiple lambda function sharing a same code dir (like a same file with different function as entrypoint). When we are doing `sam local start-api`, even only one function changes in code dir, our watcher will think every function has changed. Therefore killing the running container and putting everything into LAZY mode again. So the ignore pattern can't really fix this issue. 

Instead, providing the `--no-watch` flag gives user more control on the scenario above. They can kill and restart the command when a reload is actually required.

samcli/local/lambdafn/runtime.py
```diff
@@ -405,13 +407,23 @@ def __init__(self, container_manager, image_builder, observer=None, mount_symlin
...
         image_builder samcli.local.docker.lambda_image.LambdaImage
             Instance of the LambdaImage class that can create am image
-        warm_containers bool
-            Determines if the warm containers is enabled or not.
+        observer
+            Optional observer for file watching
+        mount_symlinks bool
+            Optional. True if symlinks should be mounted in the container
+        no_mem_limit bool
+            Optional. True if memory limit should be disabled
```
reedham-aws, line 546: These changes are just to update outdated comments?

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1401,7 +1401,13 @@ def test_must_pass_custom_region(self, add_account_id_to_global_mock, get_stacks
...
 
 class TestInvokeContext_add_account_id_to_global(TestCase):
-    def test_must_work_with_no_token(self):
+    @patch("samcli.commands.local.cli_common.invoke_context.get_boto_client_provider_with_config")
+    def test_must_work_with_no_token(self, get_boto_client_provider_with_config_mock):
+        # Mock the STS client to raise an exception (simulating no credentials). Or test might pickup system credential
+        sts_mock = Mock()
+        sts_mock.get_caller_identity.side_effect = Exception("No credentials")
+        get_boto_client_provider_with_config_mock.return_value.return_value = sts_mock
+
```
reedham-aws, line 1572: These changes seem unrelated to the rest of the PR, what are they for?

reedham-aws (dismissed):

samcli/local/lambdafn/runtime.py
roger-zhangg: yes

tests/unit/commands/local/cli_common/test_invoke_context.py
roger-zhangg: This is one test constantly failing locally

roger-zhangg: This test would fail locally when credential could be detected. This change make sure it would ignore outside credentials

samcli/commands/local/start_api/cli.py
```diff
@@ -95,6 +95,12 @@
     required_param_lists=[["ssl_cert_file"]],
     help="Path to SSL key file (default: None)",
 )
+@click.option(
+    "--no-watch",
+    is_flag=True,
+    default=False,
+    help="Disable file watching. Local code changes will not reset running docker container.",
```
valerena: From my understanding, this is only relevant when using `warm_container_initialization_mode`, right? That's probably something we should mention here, so it's easy for customers to understand when this matters. (same for the comment on `start_lambda/cli.py`

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -234,16 +236,18 @@ def __enter__(self) -> "InvokeContext":
...
         }
-
         # don't resolve the code URI immediately if we passed in docker vol by passing True for use_raw_codeuri
         # this way at the end the code URI will get resolved against the basedir option
         if self._docker_volume_basedir:
             _function_providers_args[self._containers_mode].append(True)
+            if self._no_watch:
+                _function_providers_args[self._containers_mode].extend([False, True])
+        elif self._no_watch:
+            _function_providers_args[self._containers_mode].extend([False, False, True])
```
valerena: On the same topic from another comment, if a customer uses `--no-watch` without using `--warm-container-initialization-mode` it could be that this `_function_providers_args[self._containers_mode]` refers to `SamFunctionProvider`, which doesn't have the extra parameter for `no_watch`, and it would fail with an uncaught error. We should probably have a validation to not try to do that when it doesn't make sense.

samcli/commands/local/cli_common/options.py
```diff
@@ -268,6 +268,16 @@ def warm_containers_common_options(f):
             type=click.STRING,
             multiple=False,
         ),
+        click.option(
+            "--no-watch",
```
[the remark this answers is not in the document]
roger-zhangg: Thanks for the careful read — but I don't think this bug reproduces. A couple of clarifications:

**`--no-watch` is registered in `warm_containers_common_options`, not `local_common_options`.**

```python
# samcli/commands/local/cli_common/options.py
def warm_containers_common_options(f):
    """
    Warm containers related CLI options shared by "local start-api" and "local start_lambda" commands
    """
    warm_containers_options = [
        click.option("--warm-containers", ...),
        click.option("--debug-function", ...),
        click.option(
            "--no-watch",
            is_flag=True,
            default=False,
            ...
        ),
    ]
```

That decorator is only applied to `sam local start-api` and `sam local start-lambda`. `sam local invoke` (`samcli/commands/local/invoke/cli.py`) only applies `@invoke_common_options` and `@local_common_options`, so Click never sees `--no-watch` for that command and the `cli()` signature does not need a `no_watch` parameter.

**Verified against the dev install (`samdev`):**

```
$ samdev local invoke --help | grep -i no-watch
(no output — flag is not registered)

$ samdev local invoke --help | tail -3
                                    message generated by AWS SAM CLI and
                                    display timestamps.
    -h, --help                      Show this message and exit.

$ samdev local start-api --help | grep -A1 no-watch
    --no-watch                      Disable file watching for hot reload.
                                    Only applies when --warm-containers is

$ samdev local start-lambda --help | grep -A1 no-watch
    --no-watch                      Disable file watching for hot reload.
                                    Only applies when --warm-containers is
```

No `TypeError`, `local invoke` works as before.

**Should we extend `--no-watch` to `sam local invoke` anyway?**

I'd argue no:

- `sam local invoke` does not support `--warm-containers` — there is no `warm_container_initialization_mode` argument in its `cli()` / `do_cli()`, and it never instantiates `RefreshableSamFunctionProvider` / `WarmLambdaRuntime` / `LambdaFunctionObserver`.
- Each `local invoke` runs in cold mode: spin up a fresh container, invoke once, tear down. There is no file watcher in that lifecycle for the flag to disable.
- Adding the flag there would either be a silent no-op (confusing) or require another "ignored, warning" path identical to the warm/cold validation we already added in `InvokeContext.__enter__`. That's surface-area noise without a real use case behind it.

seshubaws (dismissed):

reedham-aws (dismissed):

roger-zhangg:
Merge after LE are all out

reedham-aws (approved):

valerena (approved):

PR #8211, opened 2025-08-08
fix: remove deprecated license classifiers

mbfreder:

License classifiers are deprecated. When running `python setup.py --version` command is run internally, it also returns a deprecation warning message instead of just the version number, causing a parsing issue. The change on this PR fixes the issue.

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/appveyor-linux-binary.yml
+++ b/appveyor-linux-binary.yml
@@ -52,8 +52,10 @@ install:
   - sh: "mvn --version"
 
   - sh: "source ${HOME}/venv${PYTHON_VERSION}/bin/activate"
-  - sh: "rbenv install 3.3.0"
-  - sh: "rbenv global 3.3.0"
+  - sh: "rbenv install 3.3.7"
+  - sh: "rbenv install 3.4.2"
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -65,7 +65,6 @@ def read_version():
         "Environment :: Other Environment",
         "Intended Audience :: Developers",
         "Intended Audience :: Information Technology",
-        "License :: OSI Approved :: Apache Software License",
```
valerena, line 68: Should we **replace** this with something else? Is there a new alternative to do this instead of the old License classifier?

If this is currently visible on the generated package, we don't want to just suddenly stop adding the license information.

mbfreder: The license is already included [here](https://github.com/aws/aws-sam-cli/blob/ba82b4752d2a0a2bbb807e971a9a5d12ecc79590/setup.py#L53).

valerena (approved):

PR #8270, opened 2025-09-10
feat: Output remote test-event list error to info log so it isn't counted as failure

tobixlea:

Function with no test-events will output an error when `remote test-event list` is executed

-Prevent errors due to no test events being found from being counted as an error in our SAM CLI Ops Dashboard
-Output to log file so that customer does not have to see an error

-Functions with no test-events will not error out with error code

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/build/build_graph.py
+++ b/samcli/lib/build/build_graph.py
@@ -578,7 +578,7 @@ class LayerBuildDefinition(AbstractBuildDefinition):
 
 class FunctionBuildDefinition(AbstractBuildDefinition):
     """
-    LayerBuildDefinition holds information about each unique function build
+    FunctionBuildDefinition holds information about each unique function build
     """
 
     def __init__(


roger-zhangg:
some lint issue

seshubaws (dismissed):

valerena:
I'm thinking that we should probably change it at the source:

https://github.com/aws/aws-sam-cli/blob/94abc0ff6f54f464e3c14e849828e203f3ee35ef/samcli/lib/shared_test_events/lambda_shared_test_event.py#L410-L411

and 

https://github.com/aws/aws-sam-cli/blob/94abc0ff6f54f464e3c14e849828e203f3ee35ef/samcli/lib/shared_test_events/lambda_shared_test_event.py#L417-L418

It probably makes sense for other methods to fail, like `get_event` or `delete_event`, if the registry doesn't exist (because you pass one specific event, so you're assuming that that exists), but for `list_events` I think the original function call shouldn't really fail. For example, the `create_event` method, just creates the registry if it doesn't exist. https://github.com/aws/aws-sam-cli/blob/94abc0ff6f54f464e3c14e849828e203f3ee35ef/samcli/lib/shared_test_events/lambda_shared_test_event.py#L285-L286
so it's really prepared for that case and it doesn't count it as an error, of course.

 So I think we should change this lambda_test_event provider class (`LambdaSharedTestEvent`) instead of catching the error out here.

PR #8285, opened 2025-09-25
feat: Implement Finch container client

vicheey:

Adds Finch container runtime support for AWS SAM CLI local development and testing.

This change enables SAM CLI users to leverage Finch as container engine for local Lambda function development, testing, and debugging without manual configuration of socket path from users. This PR also implement separate client wrappers for Finch and Docker client.

- **Container Runtime Abstraction**: Implements a unified container client system that supports both Docker and Finch through a factory pattern
- **Automatic Detection**: Adds enterprise detection mechanism to automatically identify and configure Finch socket path. 
- **API Compatibility**: Ensures all existing SAM CLI container operations work seamlessly with Finch
- **Test Infrastructure**: Updates integration and unit tests to support both container runtimes
- **Error Handling**: Provides Finch-specific error messages and fallback mechanisms
- **Performance Optimizations**: Implements file locking and build caching to prevent race conditions

- **Positive**: Users can now use Finch as a Docker alternative, potentially improving performance and 
reducing resource usage
- **Minimal Breaking Changes**: Existing Docker workflows remain unchanged; Finch support is additive
- **Test Coverage**: Comprehensive test suite ensures compatibility across both container runtimes
- **Configuration**: New telemetry metrics track container runtime preferences for usage analytics

#### Mandatory Checklist

• [ ] Add input/output [type hints](https://
docs.python.org/3/library/typing.html) to new 
functions/methods
• [ ] Write design document if needed ([Do I need to
write a design document?](https://github.com/aws/
aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design
-document))
• [x] Write/update unit tests
• [x] Write/update integration tests
• [ ] Write/update functional tests if needed
• [x] make pr passes
• [ ] make update-reproducible-reqs if dependencies 
were changed
• [x] Write documentation

By submitting this pull request, I confirm that my 
contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

### DIFF
--- a/tests/unit/lib/telemetry/test_container_engine_metrics.py
+++ b/tests/unit/lib/telemetry/test_container_engine_metrics.py
@@ -75,6 +75,7 @@ class TestContainerEngineMetrics(TestCase):
         metric_attrs = metric_data["metricSpecificAttributes"]
         self.assertEqual(metric_attrs["containerEngine"], "finch")
 
+    @patch("samcli.cli.context.Context.get_current_context")
     @patch("samcli.lib.telemetry.metric.Telemetry")
     @patch("samcli.lib.telemetry.event.EventTracker.send_events")
     @patch("samcli.lib.telemetry.metric.get_all_experimental_statues")
@@ -89,6 +90,7 @@ class TestContainerEngineMetrics(TestCase):
         mock_experimental,
         mock_send_events,
         mock_telemetry,
+        mock_get_context,
     ):
         """Test that container engine is retrieved from context when Docker is used"""
         # Setup mocks
@@ -98,8 +100,11 @@ class TestContainerEngineMetrics(TestCase):
         mock_initial_commit.return_value = "abc123"
 
         # Set actual container runtime in context
-        self.mock_ctx.actual_container_runtime = "docker"
+        self.mock_ctx.actual_container_runtime = "rancher-desktop"
         self.mock_ctx.template_dict = {}
+
+        # Mock the context access for the Metric instance
+        mock_get_context.return_value = self.mock_ctx
 
         # Call the function
         _send_command_run_metrics(self.mock_ctx, 1000, "success", 0)
@@ -117,7 +122,7 @@ class TestContainerEngineMetrics(TestCase):
         # Verify container engine is correctly reported as docker
         self.assertIn("metricSpecificAttributes", metric_data)
         metric_attrs = metric_data["metricSpecificAttributes"]
-        self.assertEqual(metric_attrs["containerEngine"], "docker")
+        self.assertEqual(metric_attrs["containerEngine"], "rancher-desktop")
 
     @patch("samcli.cli.context.Context.get_current_context")
     @patch("samcli.lib.telemetry.metric.Telemetry")


roger-zhangg (dismissed):

tests/integration/package/test_package_command_image.py
```diff
@@ -43,6 +45,14 @@ def setUp(self):
     def tearDown(self):
         super(TestPackageImage, self).tearDown()
 
+    def _assert_finch_docker_success(self, stderr_text, error_message):
```
valerena, line 48: Probably we can change later, but isn't this just "Finch" success?

tests/unit/lib/build_module/test_utils.py
```diff
@@ -49,3 +56,224 @@ def test_make_env_vars_with_both(self):
         }
         result = _make_env_vars(function1, file_env_vars, inline_env_vars)
         self.assertEqual(result, {"ENV_VAR1": "2", "ENV_VAR2": "3"})
+
+
+class TestArchitectureDetection(TestCase):
```
valerena, line 61: To change as a follow-up.. all of these should be one test only, with `parameterized`

tests/unit/lib/telemetry/test_admin_preference_metrics.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Don't set admin_container_preference attribute at all
+        self.mock_ctx.template_dict = {}
+
+        # Call the function
+        _send_command_run_metrics(self.mock_ctx, 1000, "success", 0)
+
+        # Verify telemetry was called
+        mock_telemetry.assert_called_once()
+        telemetry_instance = mock_telemetry.return_value
```
valerena, line 194: I feel like we should be setting the return_value instead of getting it.. But if it works, it works. (I see we do that a lot in other tests too, so I guess it works)

tests/unit/lib/telemetry/test_container_engine_metrics.py
```diff
@@ -0,0 +1,399 @@
...
+        """Test getting container host when context raises RuntimeError"""
+        mock_get_context.side_effect = RuntimeError("No context available")
+
+        with patch.object(self.metric, "_get_container_host_from_env", return_value="docker-default") as mock_env:
+            result = self.metric._get_container_host()
+
+        self.assertEqual(result, "docker-default")
+        mock_env.assert_called_once()
+
+    @patch.dict("os.environ", {}, clear=True)
```
valerena, line 330: This can also be `parameterized` instead of individual tests

tests/unit/local/docker/test_container_client.py
```diff
@@ -0,0 +1,1239 @@
...
+        self.assertTrue(self.client.is_available())
+
+    @patch("samcli.local.docker.container_client.LOG")
+    def test_is_available_failure(self, mock_log):
+        """Test is_available returns False when ping fails and logs debug"""
+        self.client.ping.side_effect = Exception("Connection failed")
+        self.assertFalse(self.client.is_available())
+        mock_log.debug.assert_called_once()
+
+    def test_is_finch_when_runtime_is_finch(self):
```
valerena, line 899: Tests don't seem to relevant, but they can also be parameterized in pairs

tests/unit/lib/telemetry/test_container_engine_metrics.py
```diff
@@ -0,0 +1,404 @@
...
+        self.profile = None
+        self.debug = False
+        self.region = "us-east-1"
+        self.command_path = "sam local start-api"
+        self.experimental = False
+        self.template_dict = {}
+        self.session_id = "test-session-id"
+
+
+class TestContainerEngineMetrics(TestCase):
```
valerena, line 25: Most tests in this class should definitely be `parameterized`, or adding more things to the `setUp`, or with a utility function, so you don't have to put the exact same code multiple times in all tests.

PR #8292, opened 2025-10-03
test: fix validate tests and test-events tests

valerena:
- cfn-lint can send different errors/warning for Lambda deprecated runtimes, so we make an extra check to make sure that the error that we check is the right one.
- `sam remote test-event list` doesn't fail anymore when there are no events.

Tests failures: https://ci.appveyor.com/project/AWSSAMCLI/aws-sam-cli-canary-linux/build/job/fheq10brhbfi6r3w

I also change the order of the commands in `make pr`, so the linting happens before running the tests (so you don't have to wait for all the tests to run to realize `black` doesn't like your files format)

To prevent tests from failing.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ schema:
 	python -m schema.make_schema
 
 # Verifications to run before sending a pull request
-pr: init dev schema black-check
+pr: init schema black-check dev
 
 # lucashuy: Linux and MacOS are on the same Python version,
 # however we should follow up in a different change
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -56,7 +56,7 @@ schema:
 	python -m schema.make_schema
 
 # Verifications to run before sending a pull request
-pr: init dev schema black-check
+pr: init schema black-check dev
```
roger-zhangg, line 59: That's a good catch

roger-zhangg (approved):

seshubaws (approved):

PR #8295, opened 2025-10-03
feat: Support all formats of function name in invoke

### LINKED ISSUE #8289: Feature request: SAM local lambda server should support function ARNs on invoke API.
I wish that when running the `sam local start-lambda` server, I could hit my lambda functions by calling the invoke API with a function ARN instead of explicitly requiring a function name.

### Proposal
Rather than explicitly requiring the lambda runner to take in the function name, it should take in either a function name or ARN, and normalize it somewhere around [here](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/lambda_service/local_lambda_invoke_service.py#L146-L157)

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model) - No

bchampp:
Fixes #8289

https://github.com/aws/aws-sam-cli/issues/8289


The lambda invoke API supports multiple different formats for the function name parameter. However, SAM only supports matching the function name directly as it exists in the provided SAM template.


This change updates SAM so that we support more flexible function name arguments. Now, we normalize the provided function name to be just the function name by extracting it from the full or partial ARN, and/or stripping the version provided.

Now, when invoking a function locally, we first normalize the provided function name field to extract just the function name. 

Testing:
I have a SAM template which has a function "HelloWorld". I tested by invoking different variations that are supported of the function name:
```
# works as it did before
samdev local invoke HelloWorld

# alias, invokes HelloWorld function
samdev local invoke HelloWorld:alias 

# full function arn, invokes HelloWorld function
samdev local invoke arn:aws:lambda:us-east-1:123456789012:function:HelloWorld

# partial function arn, invokes HelloWorld function
samdev local invoke 123456789012:function:HelloWorld
```

I also updated the error handling around invalid function names to match what the lambda service does:
```
samdev local invoke :HelloWorld
                                                                                                                                                                                                                                                                 
Error: 1 validation error detected: Value ':HelloWorld' at 'functionName' failed to satisfy constraint: Member must satisfy regular expression pattern: ^(arn:[^:]+:lambda:[^:]*:\d{12}:function:|\d{12}:function:)?[a-zA-Z0-9-_\.]+(:[\w$-]+)?$
```


None.

#### Mandatory Checklist

- [N/A] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [N/A] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


tests/unit/lib/utils/test_name_utils.py
```diff
@@ -0,0 +1,162 @@
+"""
+Tests for samcli.lib.utils.name_utils module
+"""
+
+import pytest
+
+from samcli.lib.utils.name_utils import normalize_lambda_function_name, InvalidFunctionNameException
+
+
+class TestNormalizeLambdaFunctionName:
```
valerena: Please use format where the test class extends `TestCase` (and assertions are `self.assert*(`) like other files: https://github.com/bchampp/aws-sam-cli/blob/develop/tests/unit/lib/utils/test_architecture.py#L18

tests/unit/lib/utils/test_name_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    """Test cases for normalize_lambda_function_name function"""
+
+    def test_returns_function_name_unchanged_when_not_arn(self):
```
valerena: Most of the tests in this file should be just one test parameterized: https://github.com/bchampp/aws-sam-cli/blob/develop/tests/unit/lib/utils/test_architecture.py#L43-L57

(and you can have another parameterized one with all the exceptions)

valerena (changes requested):

tests/unit/lib/utils/test_name_utils.py
bchampp: Sounds good. Addressed.

tests/unit/lib/utils/test_name_utils.py
bchampp: Updated

tests/unit/local/lambda_service/test_local_lambda_invoke_service.py
```diff
@@ -332,3 +335,115 @@ def test_path_converter_matches_function_full_path(self):
...
+    def test_normalize_function_name_with_malformed_arn(self):
+        """Test that malformed ARNs are returned unchanged"""
+        malformed_arn = "arn:aws:lambda:us-east-1:123456789012"  # Missing function part
+        result = normalize_lambda_function_name(malformed_arn)
+        self.assertEqual(result, malformed_arn)
+
+    def test_normalize_function_name_with_non_lambda_arn(self):
+        """Test that non-Lambda ARNs raise InvalidFunctionNameException"""
+        s3_arn = "arn:aws:s3:::my-bucket"
+        with pytest.raises(InvalidFunctionNameException):
```
valerena:
```suggestion
        with self.assertRaises(InvalidFunctionNameException):
```


and you can remove the `pytest` import

tests/unit/lib/utils/test_name_utils.py
```diff
@@ -0,0 +1,70 @@
...
+            ("my-function:$LATEST", "my-function"),
+            ("not-an-arn", "not-an-arn"),
+            ("123456789012:invalid:my-function", "123456789012:invalid:my-function"),
+            ("my-function_name", "my-function_name"),
+            ("arn:aws:lambda:us-east-1:123456789012:function:my-function_name-test", "my-function_name-test"),
+            ("123456789012:function:my-function_name-test", "my-function_name-test"),
+            ("my-function-with-many-hyphens", "my-function-with-many-hyphens"),
+            ("my_function_with_underscores", "my_function_with_underscores"),
+            ("MyFunctionWithCamelCase", "MyFunctionWithCamelCase"),
+            ("function123WithNumbers", "function123WithNumbers"),
```
valerena: Do we really need all of these? One with mixed chars is enough. Same with `not-an-arn` (basically testing the same as `my-function`) and `my-function_name` above.

I don't think we need to clutter the test files with tests that don't contribute much. I'm open to discuss if you think there's actual value from testing all these cases separately.

bchampp: Yeah I agree. I'll update this to clearly separate the parameters we're passing and avoid unnecessary repetition.

tests/unit/local/lambda_service/test_local_lambda_invoke_service.py
```diff
@@ -332,3 +333,115 @@ def test_path_converter_matches_function_full_path(self):
...
+        self.assertEqual(result, "my-complex-function-name")
+
+    def test_normalize_function_name_with_malformed_arn(self):
+        """Test that malformed ARNs are returned unchanged"""
+        malformed_arn = "arn:aws:lambda:us-east-1:123456789012"  # Missing function part
+        result = normalize_lambda_function_name(malformed_arn)
+        self.assertEqual(result, malformed_arn)
+
+    def test_normalize_function_name_with_non_lambda_arn(self):
+        """Test that non-Lambda ARNs raise InvalidFunctionNameException"""
```
valerena: We don't need to test all of these again. `normalize_lambda_function_name` is already tested in its own file.

bchampp: I removed the specific tests that check different permutations of normalizing the function name from here, but left in the new tests that still check that it normalizes function ARN.

valerena (approved):

PR #8313, opened 2025-10-08
fix(test): Create seperate env for smoke test

roger-zhangg:
https://github.com/aws/aws-sam-cli/actions/runs/17596078472/job/49988516171?pr=8206

Fix the metadata flaky error in smoketest:
```
2025-09-09 21:24:54 Error when loading global config file: /home/runner/.aws-sam/metadata.json
  File "/home/runner/work/aws-sam-cli/aws-sam-cli/samcli/cli/global_config.py", line 308, in _load_config
    json_body = json.loads(body)
                ^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Create a temp dir/env for each test case

No, git action is one time use.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/smoke/test_all_commands.py
+++ b/tests/smoke/test_all_commands.py
@@ -51,8 +51,18 @@ class TestAllCommands(TestCase):
         sam_cmd = "samdev" if os.getenv("SAM_CLI_DEV", 0) else "sam"
         # if a previous smoke test run have been killed, re-running them will fail. so run them in a temp folder
         with tempfile.TemporaryDirectory() as temp:
+            # Create isolated config directory for each test to prevent metadata.json conflicts
+            env = os.environ.copy()
+            # Use a unique subdirectory for SAM CLI config to avoid concurrent access issues
+            config_dir = os.path.join(temp, ".aws-sam")
+            env["__SAM_CLI_APP_DIR"] = config_dir
+
             process = subprocess.Popen(
-                [sam_cmd, cmd_name] + args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=temp
+                [sam_cmd, cmd_name] + args,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                cwd=temp,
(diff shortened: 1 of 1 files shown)


reedham-aws:
I want to make sure I understand the problem and the fix:
### Problem
Sometimes, the smoketest can't load `metadata.json` for an individual test

### Solution
Create a new env with `metadata.json` for every test.

Will this end up being a lot slower?

roger-zhangg:
The smoke test is running in parallel: 4 workers, but the `metadata.json` currently is shared for all the tests. So the race probably comes when a test is modifying the `metadata.json` and another test is trying to read this file. This change ensures every test is using a unique `metadata.json`. 

For 
> end up being a lot slower?

It shouldn't, at most it's like 1 additional file creation per test (total ~500) (and done by 4 threads) and it shouldn't be more than few seconds

roger-zhangg:
In other branch:
image

This branch
image


I guess you just reminded me that we can probably try a much higher parallelism on smoke test given we use separate envs anyway

```
pytest -vv -n 16 tests/functional tests/smoke
943 passed in 121.87s (0:02:01) 
```

Can try to update this in a separate PR

PR #8314, opened 2025-10-08
chore(test): split integ cmd test, increase parallism

roger-zhangg:
Make integ test / smoke test faster

30~45 minute on build cmd test before
~10 minute after fix

build cmd test now distribute on 3 worker
smoke test now runs on 3 worker in windows
removed few deprecated node test

No, total test coverage hasn't changed

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/buildcmd/test_build_cmd_node.py
+++ b/tests/integration/buildcmd/test_build_cmd_node.py
@@ -123,6 +123,8 @@ class TestBuildCommand_NodeFunctions_With_Specified_Architecture(BuildIntegNodeB
         [
             ("nodejs20.x", False, "x86_64"),
             ("nodejs22.x", False, "x86_64"),
+            ("nodejs20.x", "use_container", "x86_64"),
+            ("nodejs22.x", "use_container", "x86_64"),
         ]
     )
     def test_building_default_package_json(self, runtime, use_container, architecture):


roger-zhangg:
uv benefix: 
make init with pip: 
- linux ~30s
- windows ~90 - 180s

make init with uv
- linux ~3 - 5s
- windows ~ 30 - 60 s

tests/integration/buildcmd/test_build_cmd_arm64.py
```diff
@@ -157,12 +129,8 @@ class TestBuildCommand_NodeFunctions_With_Specified_Architecture_arm64(BuildInte
 
     @parameterized.expand(
         [
-            ("nodejs16.x", False),
-            ("nodejs18.x", False),
             ("nodejs20.x", False),
             ("nodejs22.x", False),
```
valerena, line 163: I don't know why these were in the non "_al2023" section, but we can probably merge this with the following test now.

tests/integration/buildcmd/test_build_cmd_node.py
```diff
@@ -166,10 +121,6 @@ class TestBuildCommand_NodeFunctions_With_Specified_Architecture(BuildIntegNodeB
 
     @parameterized.expand(
         [
-            ("nodejs16.x", False, "x86_64"),
-            ("nodejs18.x", False, "x86_64"),
-            ("nodejs16.x", "use_container", "x86_64"),
-            ("nodejs18.x", "use_container", "x86_64"),
             ("nodejs20.x", False, "x86_64"),
             ("nodejs22.x", False, "x86_64"),
```
valerena, line 174: I'm curious why there's no "use_container" test here for these runtimes. Should we add that?

.github/workflows/build.yml
```diff
@@ -114,25 +122,32 @@ jobs:
...
         tests_config:
-          - "-n 2 --reruns 3 tests/integration/buildcmd"
+          - name: "integ-buildcmd-arm64"
+            cmd: "-n 2 --reruns 3  tests/integration/buildcmd/test_build_cmd_arm64.py"
+          - name: "integ-buildcmd-main"
+            cmd: "-n 2 --reruns 3 tests/integration/buildcmd/test_build_cmd_dotnet.py tests/integration/buildcmd/test_build_cmd_provided.py tests/integration/buildcmd/test_build_cmd_rust.py tests/integration/buildcmd/test_build_cmd_python.py tests/integration/buildcmd/test_build_cmd_node.py"
+          - name: "integ-buildcmd-java"
+            cmd: "-n 2 --reruns 5 tests/integration/buildcmd/test_build_cmd_java.py"
+          - name: "integ-buildcmd-other"
+            cmd: "-n 2 --reruns 3 tests/integration/buildcmd --ignore=tests/integration/buildcmd/test_build_cmd_python.py --ignore=tests/integration/buildcmd/test_build_cmd_java.py --ignore=tests/integration/buildcmd/test_build_cmd_dotnet.py --ignore=tests/integration/buildcmd/test_build_cmd_node.py --ignore=tests/integration/buildcmd/test_build_cmd_rust.py --ignore=tests/integration/buildcmd/test_build_cmd_provided.py --ignore=tests/integration/buildcmd/test_build_cmd_arm64.py"
```
valerena: Small, but please keep this list in the same order as when the files are used, so it's easier to see that we're ignoring the same files used in the other commands.

.github/workflows/build.yml
```diff
@@ -114,25 +122,32 @@ jobs:
           - "3.11"
         # folders that is commented below requires credentials, no need to spare time to run them
         tests_config:
-          - "-n 2 --reruns 3 tests/integration/buildcmd"
+          - name: "integ-buildcmd-arm64"
+            cmd: "-n 2 --reruns 3  tests/integration/buildcmd/test_build_cmd_arm64.py"
```
valerena: I feel this should be called "params" instead of "cmd", considering that it actually doesn't have the full command.

valerena:
And I would recommend changing the title of the PR to "**split** integ cmd tests" instead of "break", because it sounds like you're making the tests fail by saying "break"

tests/integration/buildcmd/test_build_cmd_arm64.py
roger-zhangg: I'm not sure on this one, I thought this might has some relation with AppVeyor tests.

tests/integration/buildcmd/test_build_cmd_node.py
roger-zhangg: added

tests/integration/buildcmd/test_build_cmd_arm64.py
valerena: Ah. ok. that makes sense. This is the default, but then the other ones are the ones that are weird, because it's when you're emulating an AL2023 container.. ok.

PR #8348, opened 2025-10-15
fix: support nested stacks when normalizing function name

bchampp:
#8289

Canaries are failing after merging https://github.com/aws/aws-sam-cli/pull/8295, due to SAM supporting nested stacks but the change not handling nested stacks properly. There weren't unit tests covering nested stacks being passed as the function identifier.

I've updated how we normalize the SAM resource to support nested stacks and added unit tests. Now, we parse the last component of the identifier to normalize as the function name. So if you have a nested stack (e.g. `NestedStack/MyFunction`), we'll only try normalizing `MyFunction`.

None.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation

### DIFF
--- a/tests/unit/lib/utils/test_name_utils.py
+++ b/tests/unit/lib/utils/test_name_utils.py
@@ -6,17 +6,23 @@ from unittest import TestCase
 
 from parameterized import parameterized
 
-from samcli.lib.utils.name_utils import normalize_lambda_function_name, InvalidFunctionNameException
+from samcli.lib.utils.name_utils import (
+    normalize_lambda_function_name,
+    normalize_sam_function_identifier,
+    InvalidFunctionNameException,
+)
 
 
-class TestNormalizeLambdaFunctionName(TestCase):
-    """Test cases for normalize_lambda_function_name function"""
+class TestNormalizeSamFunctionIdentifier(TestCase):
+    """Test cases for normalize_sam_function_identifier function"""
 
     @parameterized.expand(
         [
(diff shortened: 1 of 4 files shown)


tests/unit/lib/utils/test_name_utils.py
```diff
@@ -25,36 +31,50 @@ class TestNormalizeLambdaFunctionName(TestCase):
...
     )
     def test_raises_exception_for_invalid_function_names(self, invalid_name):
         """Test that invalid function names raise InvalidFunctionNameException"""
         with self.assertRaises(InvalidFunctionNameException) as exc_info:
-            normalize_lambda_function_name(invalid_name)
+            normalize_sam_function_identifier(invalid_name)
 
         # Check that the error message matches AWS Lambda's format
         self.assertIn("1 validation error detected", str(exc_info.exception))
-        self.assertIn(f"Value '{invalid_name}' at 'functionName'", str(exc_info.exception))
```
bchampp, line 59: I had to remove this assertion because it's using the stripped function name in the error message (e.g. the test would think it's `Value Stack1/Stack2/function-with-invalid-chars! at functionName ...`), but the exception is actually just: `function-with-invalid-chars! at functionName ...`. Let me know if you want me to update the test parameterization to assert this.

bchampp:
See https://github.com/aws/aws-sam-cli/pull/8351 - the original change was reverted, I opened a new PR to add back the change + the fix.

PR #8351, opened 2025-10-15
feat: Support all formats of function name in invoke

bchampp:
#8289

https://github.com/aws/aws-sam-cli/pull/8295 was reverted because it wasn't properly handling nested stacks - we were invalidating cases where the resource identifier had `/` (not covered by unit tests, only the canary that runs after merge). 

I opened an isolated PR for the fix before it was reverted, but this PR has both the initial change and the fix. See just the fix here: https://github.com/aws/aws-sam-cli/pull/8348

I've updated how we normalize the SAM resource to support nested stacks and added unit tests. Now, we parse the last component of the identifier to normalize as the function name. So if you have a nested stack (e.g. `NestedStack/MyFunction`), we'll only try normalizing `MyFunction`.

None.

#### Mandatory Checklist

- [X] Write/update unit tests
- [X] `make pr` passes
- [X] Write documentation


seshubaws (dismissed):

valerena:
Do you think it's possible to add a small basic unit test for the **old code** to check that the nested stack case works? I'm trying to think of a test that would have caught this problem.

bchampp:
> Do you think it's possible to add a small basic unit test for the **old code** to check that the nested stack case works? I'm trying to think of a test that would have caught this problem.

We would have needed a test that exercised the `invoke` call in `local_lambda.py` with a nested stack as the SAM identifier. I'll check what test cases exist for this and add a new one for the nested stack case.

valerena (approved):

Issue #8352, opened 2025-10-16
Bug: sam build - Container runtime not detected in Bitbucket Pipelines with Docker-in-Docker (v1.145.0+)


paibarm:
## Description

SAM CLI version 1.145.0 and 1.145.1 fail to detect Docker when running `sam build` for Container Image Lambda functions in Bitbucket Pipelines with Docker-in-Docker (DinD) configuration.

The error occurs even though:
- Docker daemon is running and accessible
- `docker info` works correctly
- `DOCKER_HOST=tcp://localhost:2375` is properly configured
- Version 1.144.0 works perfectly in the same environment

## Steps to reproduce

**Environment:**
- CI/CD: Bitbucket Pipelines
- Docker service configuration: `docker:dind` (exposes Docker at `tcp://localhost:2375`)
- Lambda type: Container Image (PackageType: Image)
- SAM template: `template.yaml` with Container Image Lambda

**Pipeline configuration:**
```yaml
definitions:
  services:
    docker:
      type: docker
      image: docker:dind

pipelines:
  custom:
    ci-lambda-pre:
      - step:
          image: custom-ci-lambda:latest  # Contains SAM CLI 1.145.1
          services:
            - docker
          script:
            - sam build --template template.yaml
```

**Docker daemon logs confirm it's running:**
```
time="2025-10-16T06:45:04Z" level=info msg="Daemon has completed initialization"
time="2025-10-16T06:45:04Z" level=info msg="API listen on [::]:2375"
```

**Commands executed:**
```bash
# Docker is accessible
$ docker info
# ✅ Works correctly

# Environment variable is set
$ echo $DOCKER_HOST
tcp://localhost:2375

# SAM build fails
$ sam build --template template.yaml
```

## Observed result

```
[2025-10-16T06:45:31Z] 🐳 Docker configured: DOCKER_HOST=tcp://localhost:2375
Starting Build use cache
Building codeuri: /opt/atlassian/pipelines/agent/build/my-lambda runtime: None architecture: x86_64 functions: LambdaFunction
Building image for LambdaFunction function
Setting DockerBuildArgs for LambdaFunction function
Error: Running AWS SAM projects locally requires a container runtime. Do you have Docker or Finch installed and running?
```

## Expected result

SAM CLI should detect Docker via `DOCKER_HOST=tcp://localhost:2375` and successfully build the Container Image Lambda, as it does in version 1.144.0.

## Version comparison

| Version | Status | Notes |
[...]

PR #8358, opened 2025-10-20
fix: Container runtime dind build issue

vicheey:
#8352

- Fix docker client creation failure in dind where DOCKER_HOST is ignored causing the container server availability check to fail
- Fix inaccurate container runtime telemetry due to Click context unavailability

- Update the instantiation logic in Container Client class to account for environment variables and support environment variable override. Implement environment variables merging to pass to docker_py Docker Client constructor.
- Using global variable instead of Click context. 

- Tests need to be updated. 

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/unit/local/docker/test_container_client.py
+++ b/tests/unit/local/docker/test_container_client.py
@@ -591,6 +591,79 @@ class TestFinchContainerClient(BaseContainerClientTestCase):
 
         self.assertEqual(result, expected_result)
 
+    @patch("samcli.local.docker.container_client.LOG")
+    def test_validate_image_count_api_error(self, mock_log):
+        """Test validate_image_count with API error"""
+        self.client.images.list.side_effect = docker.errors.APIError("List failed")
+
+        result = self.client.validate_image_count("test-image", (1, 2))
+
+        self.assertFalse(result)
+        mock_log.warning.assert_called_once()
+
+    def test_get_socket_path_first_call(self):
+        """Test get_socket_path on first call"""
+        with patch("samcli.local.docker.container_client.get_finch_socket_path", return_value=self.finch_socket):
+            result = self.client.get_socket_path()
+
+        self.assertEqual(result, self.finch_socket)
+
+    def test_get_socket_path_caching(self):
+        """Test that get_socket_path caches the result"""
+        with patch("samcli.local.docker.container_client.get_finch_socket_path") as mock_get_finch:
+            mock_get_finch.return_value = self.finch_socket
+
+            # First call
+            result1 = self.client.get_socket_path()
+            # Second call should use cached value
+            result2 = self.client.get_socket_path()
+
+            self.assertEqual(result1, self.finch_socket)
+            self.assertEqual(result2, self.finch_socket)
+            # get_finch_socket_path should only be called once due to caching
+            mock_get_finch.assert_called_once()
+
+    def test_get_socket_path_with_none_result_raises_exception(self):
+        """Test get_socket_path raises exception when get_finch_socket_path returns None"""
+        with patch("samcli.local.docker.container_client.get_finch_socket_path", return_value=None):
+            with self.assertRaises(ContainerInvalidSocketPathException) as context:
+                self.client.get_socket_path()
+
+            self.assertIn("Finch is not supported on current platform!", str(context.exception))
+
+    def test_remove_image_safely_with_running_containers(self):
+        """Test image removal with running container cleanup"""
+        containers = self.create_mock_containers([{"status": "running", "id": "container1"}])
+
+        with patch.object(self.client, "list_containers_by_image", return_value=containers):
+            self.client.remove_image_safely("test-image", force=True)
+
+        # Verify container was stopped and removed
+        containers[0].stop.assert_called_once()
+        containers[0].remove.assert_called_once_with(force=True)
+        self.client.images.remove.assert_called_once_with("test-image", force=True)
+
+    @patch("samcli.local.docker.container_client.LOG")
+    def test_remove_image_safely_image_not_found(self, mock_log):
+        """Test image removal when image is not found"""
+        with patch.object(self.client, "list_containers_by_image", return_value=[]):
+            self.client.images.remove.side_effect = docker.errors.ImageNotFound("Image not found")
+
+            # Should not raise exception
+            self.client.remove_image_safely("test-image", force=True)
+
+            mock_log.debug.assert_called_once()
+
+    @patch("samcli.local.docker.container_client.LOG")
+    def test_list_containers_by_image_api_error(self, mock_log):
+        """Test container listing with API error"""
+        self.client.containers.list.side_effect = docker.errors.APIError("List failed")
+
+        result = self.client.list_containers_by_image("test-image")
+
+        self.assertEqual(result, [])
+        mock_log.warning.assert_called_once()
+
 
 class TestFinchContainerClientInit(BaseContainerClientTestCase):
     """Test the FinchContainerClient __init__ method"""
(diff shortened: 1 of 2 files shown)


samcli/local/docker/container_client.py
```diff
@@ -29,33 +47,52 @@ class ContainerClient(docker.DockerClient, ABC):
...
 
-    def __init__(self, *args, **kwargs):
-        """Initialize the container client with docker.DockerClient compatibility."""
-        super().__init__(*args, **kwargs)
+    # Initialize socket_path
+    socket_path: Optional[str] = None
 
-    @classmethod
-    def from_existing_client(cls, existing_client: docker.DockerClient):
+    def __init__(self, **override_env_params):
```
valerena: What seems a better option to me is to pass the `**kwargs` that want to be overridden, instead of passing the "env vars". 

Meaning, we would be able to call `ContainerClient(base_url="")` instead of `ContainerClient(DOCKER_HOST="")`.

It's seems easier to understand to me considering the DockerPy documentation instead of having to find which env var means what. 

And then the code below you would do the overrides in a different order:

```python
        # Start with system environment variables
        env_params = kwargs_from_env(environment=current_env)
        # Update overrides
        env_params.update(kwargs)
        # Initialize DockerClient with processed parameters
        LOG.debug(f"Creating container client with parameters: {env_params}")
        super().__init__(**env_params)
```

In practice because we don't expose this and it's just an implementation detail, we could change it later. But it seems more logical to me.

vicheey: Implemented the change.

bnusunny (approved):

tests/unit/local/docker/test_container_client.py
```diff
@@ -20,9 +20,185 @@
...
+        finch_socket_path = finch_socket_path or self.finch_socket
+
+        with patch("samcli.local.docker.container_client.get_finch_socket_path", return_value=finch_socket_path), patch(
+            "os.environ", env_vars
+        ), patch("docker.api.client.APIClient", return_value=self.create_mock_api_client(expected_base_url)):
+
+            with patch.object(
+                docker.DockerClient, "__init__", side_effect=docker.DockerClient.__init__, autospec=True
+            ) as spy_init:
+                client = client_class()
```
valerena: Do we understand this piece of code? The nested double `with`s, and returning `spy_init`, which is created inside the `with` all looks questionable.

valerena: I guess "it works", but I would like to see if there's a better way to do this.

tests/unit/local/docker/test_container_client.py
```diff
@@ -20,9 +20,185 @@
...
+            "validate_image_count",
+            "get_archive",
+        ]
+
+        methods_to_bind = methods_to_bind or default_methods
+
+        # Bind methods from the actual class
+        for method_name in methods_to_bind:
+            if hasattr(client_class, method_name):
+                setattr(client, method_name, getattr(client_class, method_name).__get__(client))
```
valerena, line 60: Is this really something we want to do? It seems like very complicated thing to do for the tests. I feel there might be a simpler way to accomplish the same tests in a way that we understand better.

tests/unit/lib/telemetry/test_container_engine_metrics.py
```diff
@@ -26,86 +27,48 @@ class TestContainerEngineMetrics(TestCase):
...
     @patch("samcli.lib.telemetry.metric.get_all_experimental_statues")
     @patch("samcli.lib.telemetry.metric.get_git_remote_origin_url")
     @patch("samcli.lib.telemetry.metric.get_project_name")
     @patch("samcli.lib.telemetry.metric.get_initial_commit_hash")
-    def test_container_engine_from_context_finch(
+    def test_container_engine_from_global_storage_scenarios(
         self,
+        socket_path,
+        expected_engine,
+        description,
```
valerena: We don't really need the `description` here just to show a nicer error message if it fails. The test name and values are enough.

tests/unit/local/docker/test_container_client.py
```diff
@@ -284,60 +500,160 @@ def test_get_archive(self):
...
+        [
+            ("", "", "", "Both DOCKER_HOST and Finch socket are empty strings"),
+            ("unix:///var/run/docker.sock", "", "unix:///var/run/docker.sock", "DOCKER_HOST set, Finch socket empty"),
+            ("", "unix://~/.finch/finch.sock", "", "DOCKER_HOST empty, Finch socket set"),
+            ("tcp://localhost:2375", "unix://~/.finch/finch.sock", "tcp://localhost:2375", "Different socket types"),
+        ]
+    )
+    @patch("samcli.local.docker.container_client.get_finch_socket_path")
+    def test_get_socket_path_finch_comparison_scenarios(
+        self, docker_host, finch_socket, expected_result, description, mock_get_finch_socket_path
```
valerena: Do we really want the description here? It feels like something that's harder to maintain instead of actually providing a lot of value.

tests/unit/local/docker/test_container_client_factory.py
```diff
@@ -510,147 +421,9 @@ def test_admin_preference_returns_none_when_handler_returns_none(self, mock_get_
...
-        mock_try_docker.return_value = mock_client
-
-        result = ContainerClientFactory._create_auto_detected_client()
-
-        self.assertEqual(result, mock_client)
-        mock_log.debug.assert_called_with("Using Docker as Container Engine.")
-        mock_set_context.assert_not_called()
+# TestSetContextRuntimeType and TestCreateClientWithContextStorage classes removed
+# These tests are no longer relevant since _set_context_runtime_type method was removed
+# in favor of the new global telemetry storage approach
```
valerena: This is not a comment we want in the code. It's probably a good comment to have in the PR, but in the code, this immediately gets irrelevant because we don't care anymore about these old `TestSetContextRuntimeType` and `TestCreateClientWithContextStorage` classes.

valerena:
Fix the PR title to start with `fix:`

tests/unit/local/docker/test_container_client_factory.py
vicheey: Address in the [d16a8fe](https://github.com/aws/aws-sam-cli/pull/8358/commits/d16a8fe091d43253d3b9f3cd2fac4d444b16d0b2)

tests/unit/local/docker/test_container_client.py
vicheey: [d16a8fe](https://github.com/aws/aws-sam-cli/pull/8358/commits/d16a8fe091d43253d3b9f3cd2fac4d444b16d0b2)

tests/unit/lib/telemetry/test_container_engine_metrics.py
vicheey: [d16a8fe](https://github.com/aws/aws-sam-cli/pull/8358/commits/d16a8fe091d43253d3b9f3cd2fac4d444b16d0b2)

tests/unit/local/docker/test_container_client.py
vicheey: [d16a8fe](https://github.com/aws/aws-sam-cli/pull/8358/commits/d16a8fe091d43253d3b9f3cd2fac4d444b16d0b2)

tests/unit/local/docker/test_container_client.py
vicheey: [d16a8fe](https://github.com/aws/aws-sam-cli/pull/8358/commits/d16a8fe091d43253d3b9f3cd2fac4d444b16d0b2)

PR #8386, opened 2025-10-31
refactor(test): allow linux integ test to run in github action

roger-zhangg:
Integ Test can not be run in Github Action (GA) before due to multiple reason.
- Few tests (integration/buildcmd especially) takes too much disk size, GA runner only has 18GB of free space by default.
- No good way to sync credential to GA
- GA uses other GA to setup test environments

Provide faster and more stable testing experience
- Uses OIDC to assume test accounts (Valid up to 6 hours)
- Breaks down test even more by separating test that actually doesn't require docker. So for integration/buildcmd there are 3 category: docker/finch/no-container. This avoids test that doesn't require container to be run in container tests and provided better parallelization and efficiency at same time.
- Break down integration/buildcmd test by language so each container test runner only need to download 2-3 types of sam builder image to save space.
- Update few outdated test to use newer runtime
  - Go 1.18 -> 1.25
  - Node 18 -> Node 22
  - Ruby 3.2 -> Ruby 3.4
- Fixed Few issues introduced by 
  - #8373 
  - #8351

Test run: https://github.com/roger-zhangg/aws-sam-cli/actions/runs/18965218688
Newer test run: https://github.com/roger-zhangg/aws-sam-cli/actions/runs/19054600514
Takes only ~30 minute to finish all tests

Job | Run time |  
-- | -- | --
build-integ-arm64-java (docker) | 19m 40s |  
build-integ (docker) | 12m 46s |  
sync (docker) | 25m 4s |  
build-integ-java-python-provided (docker) | 12m 33s |  
other-and-e2e (docker) | 23m 48s |  
package-delete-deploy (docker) | 29m 5s |  
build-integ-dotnet-node-ruby (docker) | 5m 0s |  
local-start2 (docker) | 11m 15s |  
local-start (docker) | 20m 47s |  
build-integ-arm64 (docker) | 10m 56s |  
terraform-build (docker) | 16m 48s |  
local-invoke (docker) | 31m 56s |  
build-integ-arm64 (finch) | 15m 23s |  
build-integ-arm64-java (finch) | 20m 39s |  
terraform-build (finch) | 15m 51s |  
package-delete-deploy (finch) | 29m 58s |  
sync (finch) | 24m 59s |  
local-start (finch) | 22m 41s |  
build-integ-dotnet-node-ruby (finch) | 5m 37s |  
local-invoke (finch) | 28m 42s |  
build-integ (finch) | 16m 14s |  
build-integ (no-container) | 10m 54s |  
build-integ-java-python-provided (no-container) | 17m 51s |  
local-start2 (finch) | 14m 23s |  
build-integ-java-python-provided (finch) | 10m 47s |  
build-integ-arm64-java (no-container) | 2m 30s |  
build-integ-arm64 (no-container) | 9m 1s |  
build-integ-dotnet-node-ruby (no-container) | 6m 12s |  
other-and-e2e (finch) | 25m 47s |  
  | 8h 17m 7s |  

- 1-2 test being flaky in github test 
- We changed the default test account role assume logic. Before the process will delete and re-create the test account role before assuming (this need to wait for 90 second to make sure we are assuming to the new role as the role is using the same name). Now we skipped the role re-creation process so we can finish assume role process in 10 second. But this could have unknown side effect, if that happens, we should revert back to the default behavior by change: tests/get_testing_resources.py (by removing `payload_data = json.dumps({"skip_role_deletion": True})`)

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/tests/integration/remote/test_event/test_remote_test_event.py
+++ b/tests/integration/remote/test_event/test_remote_test_event.py
@@ -1,6 +1,5 @@
 import json
 import os
-import time
 import uuid
 
 import pytest


tests/integration/buildcmd/build_integ_base.py
```diff
@@ -543,6 +565,17 @@ def _test_with_go(self, runtime, code_uri, mode, relative_path, architecture=Non
         newenv["GOPROXY"] = "direct"
         newenv["GOPATH"] = str(self.working_dir)
 
+        # Build with musl target to avoid glibc compatibility issues
+        # This ensures the binary works in the Lambda execution environment
+        if architecture == ARM64:
+            newenv["GOOS"] = "linux"
+            newenv["GOARCH"] = "arm64"
```
vicheey: Curious, why not?

```python
newenv["GOOS"] = "linux"
newenv["GOARCH"] = "arm64" if architecture == ARM64 else "amd64"
newenv["CGO_ENABLED"] = "0"
```

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -543,6 +565,17 @@ def _test_with_go(self, runtime, code_uri, mode, relative_path, architecture=Non
         newenv["GOPROXY"] = "direct"
         newenv["GOPATH"] = str(self.working_dir)
 
+        # Build with musl target to avoid glibc compatibility issues
+        # This ensures the binary works in the Lambda execution environment
+        if architecture == ARM64:
```
vicheey: Nit non blocker: Any reason to rewrite the environment variables instead of :
```
            newenv["GOOS"] = "linux"
            newenv["GOARCH"] = "arm64" of architecture == ARM64 else  "amd64"
            newenv["CGO_ENABLED"] = "0"
```

vicheey (dismissed):

tests/integration/buildcmd/build_integ_base.py
roger-zhangg: CGO_ENABLED will make the binary require the glibc version that's not available in lambda runtime. And therefore fails the test

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -543,6 +565,17 @@ def _test_with_go(self, runtime, code_uri, mode, relative_path, architecture=Non
...
+            newenv["GOOS"] = "linux"
```
seshubaws: nit: this and cgo_enabled can be set outside the if else

tests/integration/buildcmd/test_build_cmd.py
```diff
@@ -1601,7 +1632,7 @@ class TestBuildWithNestedStacks3Level(NestedBuildIntegBase):
 
     template = os.path.join("deep-nested", "template.yaml")
 
-    def test_nested_build(self):
+    def test_nested_build_in_conatiner(self):
```
seshubaws: nit container

tests/integration/buildcmd/build_integ_base.py
seshubaws: i see above comments, disregard

seshubaws:
In PR description says "Integ Test can not be run in github action due to multiple reason."

Could you expand on the reasons for posterity? Otherwise, LGTM

tests/integration/buildcmd/build_integ_base.py
roger-zhangg: We don't need this before I think mainly because we are using Go1.18. I updated it to 1.25

roger-zhangg: Oh wait I misunderstood, I can make this change

tests/integration/buildcmd/test_build_cmd.py
roger-zhangg: Thanks for catching!

seshubaws (approved):

vicheey (approved):

PR #8453, opened 2025-11-20
Fix tenant ID validation error handling in sam local start-api

licjun:
Add proper exception handling for TenantIdValidationError in API Gateway 
service to return HTTP 400 Bad Request with the actual validation message 
instead of generic "Internal server error".

Before this change, if user do sam local start-api and invoke a multi-tenant function without tenant-id header provided, user will see an internal server error, with this change user will see {"message":"The invoked function is enabled with tenancy configuration. Add a valid tenant ID in your request and try again.}


roger-zhangg:
let's add a test to make sure this error is throw as expected

roger-zhangg (approved):

licjun:
> Is this test actually doing anything? Is it like you are throwing an TenantIdValidationError in the test and assert the same error is thrown? Ideally we should just invoke the function with the condition (tenancy function without tenant ID, or non-tenant function with tenant ID) and assert the error. You should not use `invoke_mock.side_effect` directly. Or it's like the test is testing itself

This test is specifically testing error handling, not validation logic.

The actual tenant-id validation logic is already tested in `tests/unit/commands/local/lib/test_local_lambda.py` in the `TestLocalLambda_tenant_id_validation` class, which includes:
• test_multi_tenant_function_without_tenant_id_raises_error
• test_non_multi_tenant_function_with_tenant_id_raises_error
• test_multi_tenant_function_with_tenant_id_succeeds
(https://github.com/aws/aws-sam-cli/blob/develop/tests/unit/commands/local/lib/test_local_lambda.py#L1006)
Those tests properly invoke functions with real configurations and let the actual validation logic run.

For the API Gateway service test, the responsibility is different - it needs to verify that when TenantIdValidationError is raised (from the lambda runner), the API Gateway service correctly:
1. Catches the exception
2. Converts it to an HTTP 400 response
3. Returns the proper error message in JSON format

roger-zhangg (approved):
let's add test back after release

samcli/local/apigw/local_apigw_service.py
```diff
@@ -739,6 +739,9 @@ def _request_handler(self, **kwargs):
 
             # invoke the route's Lambda function
             lambda_response = self._invoke_lambda_function(route.function_name, route_lambda_event, tenant_id)
+        except TenantIdValidationError as e:
+            response_data = jsonify({"message": str(e)})
+            endpoint_service_error = make_response(response_data, 400)  # HTTP 400 Bad Request
```
valerena, line 744: It would be better to be consistent with the other Errors and add a new method inside the [`ServiceErrorResponses`](https://github.com/aws/aws-sam-cli/blob/11f52d8e8b7ec65e5b1a9ee80d8853a58964d45a/samcli/local/apigw/service_error_responses.py#L10) class instead of calling `make_response` right here.

But it's just a code refactor that's not critical.

PR #8454, opened 2025-11-21
feat(docker): use SAM_DOCKER_API_VERSION to control docker api version

roger-zhangg:
if docker API version `1.35` & `1.44` all doest work in certain scenario, the user can't use SAM CLI's docker feature until we release a fix.

this PR provides an escape hatch to use envvar `SAM_DOCKER_API_VERSION` to set a working docker API until we release a new fix.

`SAM_DOCKER_API_VERSION=1.51 sam build -u` to force using 1.51 as docker API version

potentially SAM_DOCKER_API_VERSION could have a conflicting use, but this is very unlikely

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/local/docker/container_client.py
+++ b/samcli/local/docker/container_client.py
@@ -88,7 +88,6 @@ class ContainerClient(docker.DockerClient, ABC):
 
         # Always start with environment variables
         current_env = os.environ.copy()
-        print(current_env)
         self.client_params = kwargs_from_env(environment=current_env)
 
         # Override base_url if explicitly provided


samcli/local/docker/container_client.py
```diff
@@ -87,14 +88,15 @@ def __init__(self, base_url=None):
 
         # Always start with environment variables
         current_env = os.environ.copy()
+        print(current_env)
```
bnusunny: Do we need to keep this line?

bnusunny (changes requested):

PR #8472, opened 2025-12-01
feat: implement lazy loading for CLI groups

bchampp:
The SAM CLI has many CLI groups now, and we see overhead while running a given command due to SAM needing to import all CLI sub-commands (and the modules they need).

Implementing a lazy CLI group base class based on https://click.palletsprojects.com/en/stable/complex/#lazily-loading-subcommands. This defers imports so we only import the given CLI group when it's the command being run.

None - defer imports to be lazy loaded. I've added unit tests to validate the `--help` commands for these CLI groups, but if a requested CLI group doesn't exist (e.g. we're pointing to a non-existing group), the schema generation will fail.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/schema/make_schema.py
+++ b/schema/make_schema.py
@@ -9,7 +9,6 @@ from typing import Any, Dict, List, Optional, Union
 import click
 
 from samcli.cli.command import _SAM_CLI_COMMAND_PACKAGES
-from samcli.cli.lazy_group import LazyGroup
 from samcli.lib.config.samconfig import SamConfig
 from schema.exceptions import SchemaGenerationException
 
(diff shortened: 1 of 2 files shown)


schema/make_schema.py
```diff
@@ -186,18 +187,28 @@ def retrieve_command_structure(package_name: str) -> List[SamCliCommandSchema]:
...
-            command.append(
-                SamCliCommandSchema(
-                    cmd_name,
-                    clean_text(subcommand.help or subcommand.short_help or ""),
-                    get_params_from_command(subcommand),
-                )
-            )
+        # For LazyGroup, load lazy commands first (preserving definition order)
+        if isinstance(module.cli, LazyGroup) and module.cli.lazy_subcommands:
+            for subcommand_name in module.cli.lazy_subcommands.keys():
```
reedham-aws: Could we use `module.cli.list_commands(None)` instead of taking the field directly? This way in theory we can also get non lazy-loaded commands if they exist. It also would allow us to get rid of the second check in the `if`.

reedham-aws (approved):
Approving because I think it's good as is, just had one small suggestion

samcli/commands/local/local.py
```diff
@@ -5,21 +5,19 @@
...
-from .start_api.cli import cli as start_api_cli
-from .start_lambda.cli import cli as start_lambda_cli
+from samcli.cli.lazy_group import LazyGroup
 
 
-@click.group()
+@click.group(
+    cls=LazyGroup,
+    lazy_subcommands={
+        "invoke": "samcli.commands.local.invoke.cli.cli",
```
Vandita2020, line 14: Why does in `list.py` file we use only one `cli` suffix in the module paths, while here and below have `cli` appearing twice in the path

bchampp: The syntax is `{module}.{cli}` -- in this case, it's because the `local invoke` command is `cli` in a file named `cli.py`.

schema/make_schema.py
reedham-aws: Updated in revision, resolving this comment and I've already approved.

valerena: you still need to approve again after any changes though because of our configuration in the repo.

reedham-aws (approved):

roger-zhangg:
Do we have some before/after data comparison?

bchampp:
> Do we have some before/after data comparison?

Yeah - I was able to see about a ~300ms difference calling the `--help` CLI for a subcommand under local. I had to turn off the proxy module thing that gets added in development mode for it to be realistic since that also adds a lot of overhead.

PR #8483, opened 2025-12-03
chore: bind lambda RIE to 0.0.0.0 for durable functions

bchampp:
The Lambda RIE container gets bound to 127.0.0.1 (localhost) by default. The durable executions emulator needs to interact with the RIE endpoint directly - so we're making HTTP calls from within the emulator container across to the RIE endpoint.

In some environments, this isn't accessible by default, and we see errors like `Could not connect to endpoint: host.docker.internal..`. We need to bind the RIE container to 0.0.0.0 instead so it's accessible on any interface. Users can work around this by using `--container-host-interface 0.0.0.0`, but this should by default.

Updating the `DurableLambdaContainer` class to override the lambda container `container_host_interface` parameter to 0.0.0.0.

Lambda container is always bound to 0.0.0.0 for durable executions.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/samcli/lib/clients/lambda_client.py
+++ b/samcli/lib/clients/lambda_client.py
@@ -53,11 +53,15 @@ class DurableFunctionsClient:
             # Create a fresh botocore session
             session = botocore.session.Session()
 
-            # Create the boto3 client with custom service model using the fresh session
+            # Create the boto3 client using the fresh session
             client = session.create_client(
                 "lambda",
                 endpoint_url=endpoint_url,
                 region_name=region,
+                # the emulator doesnt access any AWS resources,
+                # but we need _some_ credentials to create a boto client
+                aws_access_key_id="foo",
+                aws_secret_access_key="bar",
             )
 
             return cls(client)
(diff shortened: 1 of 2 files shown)


samcli/local/docker/durable_lambda_container.py
```diff
@@ -58,6 +58,9 @@ def _update_lambda_environment_with_emulator_endpoint(self, kwargs):
         extra_hosts["host.docker.internal"] = "host-gateway"
         kwargs["extra_hosts"] = extra_hosts
 
+        # Bind to 0.0.0.0 so emulator can reach Lambda via host.docker.internal
```
vicheey, line 61: What would be the effect on existing user who uses `container_host_interface`? With hard coding `0.0.0.0` address, will exsting customers see any issue? Should we update any doc?

bchampp: `--container-host-interface` won't be applied for durable functions (e.g. the lambda container will _always_ get bound to 0.0.0.0 now). I am going to update the public docs to specify this behaviour for local invoke with durable functions.

PR #8486, opened 2025-12-03
feat: add interactive callback CLI

bchampp:
Customers using `sam local invoke` to test durable functions that are using callback operations need a way to know when their functions are suspended and waiting for a callback. Otherwise, the only other way to know is by querying the execution history (`sam local execution history`) and seeing a `CallbackStarted` and `InvocationCompleted` event. This is not the best user experience, and we should be prompting users to respond to pending callbacks immediately.

This PR adds a callback handler which will prompt users to respond to pending callbacks. The prompts include success, fail, heartbeat or stopping the execution entirely. We create a new thread in the polling loop to prompt the user, so we can continue checking the execution state and history. If another process resolves the callback (e.g. you call `sam local callback succeed` from another session), the execution will continue.

None. Improved UX for callbacks in the `local invoke` loop.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/lib/utils/durable_callback_handler.py
+++ b/samcli/lib/utils/durable_callback_handler.py
@@ -1,0 +1,137 @@
+"""
+Interactive callback handler for durable function executions.
+"""
+
+import logging
+from typing import Optional
+
+import click
+
+from samcli.lib.clients.lambda_client import DurableFunctionsClient
+
+LOG = logging.getLogger(__name__)
+
+# Menu choice constants
+CHOICE_SUCCESS = 1
+CHOICE_FAILURE = 2
+CHOICE_HEARTBEAT = 3
+CHOICE_STOP = 4
+
+
+class DurableCallbackHandler:
+    """
+    Handles interactive callback detection and response for durable executions.
+    """
+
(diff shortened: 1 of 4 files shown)


samcli/lib/utils/durable_callback_handler.py
```diff
@@ -0,0 +1,137 @@
...
+            return False
+
+        click.echo(f"\n🔄 Execution is waiting for callback: {callback_id}")
+        click.echo("Choose an action:")
+        click.echo("  1. Send callback success")
+        click.echo("  2. Send callback failure")
+        click.echo("  3. Send callback heartbeat")
+        click.echo("  4. Stop execution")
+
+        choice = click.prompt("Enter choice", type=click.IntRange(1, 4), default=CHOICE_SUCCESS)
```
vicheey, line 95: What happend if customer do unblock callback from somewhere else, do we support this case? would the prompt here be blocking?

bchampp: No -- not blocking, if you respond to the callback from another terminal session, the execution will progress.

PR #8487, opened 2025-12-03
fix(sync): Handle timezone-naive datetimes in sync.toml (#8477)

### LINKED ISSUE #8477: TypeError: can't subtract offset-naive and offset-aware    datetimes during sam sync --watch
Description:

  When running sam sync --watch with a nested stack containing
  an ECR-based Lambda function, the sync fails with a datetime
  comparison error in SAM's internal code.

  SAM CLI Version:
  1.149.0

  OS:
  Linux 6.8.0-87-generic (Ubuntu)

  Steps to Reproduce:

  1. Create a SAM template with a nested stack
  (AWS::CloudFormation::Stack) that references a local template:
  RecruiticsMLStack:
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: ./THEMLTemplate.yaml
      Parameters:
        EnvType: !Ref EnvType

  2. The nested template contains a Lambda function using an ECR
   image:
  generateRecruiticsFlags:
    Properties:
      PackageType: Image
      ImageUri: !Sub "${AWS::AccountId}.dkr.ecr.${AWS::Region}.a
  mazonaws.com/my-repo:latest"

  3. Run sam sync --watch:
  sam sync --watch --stack-name my-stack --region eu-west-1
  --parameter-overrides EnvType=dev --template THETemplate.yaml

  Expected Behavior:

  Infra sync completes successfully.

  Actual Behavior:

  Build succeeds, but infra sync fails with:

  Failed to sync infra. Code sync is paused until template/stack
   is fixed.
    File "samcli/lib/sync/watch_manager.py", line 269, in 
  _execute_infra_sync
    File "samcli/lib/sync/watch_manager.py", line 199, in 
  _execute_infra_context
    File "samcli/lib/sync/infra_sync_executor.py", line 176, in
  execute_infra_sync
  TypeError: can't subtract offset-naive and offset-aware
  datetimes

  Additional Context:

  - The nested stack deploys successfully via sam deploy
  (non-watch mode)
  - SAM shows this warning before the error: The resource 
  AWS::Serverless::Function 'generateRecruiticsFlags' has
  specified ECR registry image for ImageUri. It will not be
  built and SAM CLI does not support invoking it locally.
  - The error appears to be in SAM's datetime handling when
  comparing stack update times, not in the template itself

  Workaround:

  Using sam deploy instead of sam sync --watch works correctly   (this is a terrible work-flow of course with poor feedback loop experience compared to sam sync --watch)

vicheey:
Close #8477, #8679

`sam sync --watch` crashes with `TypeError: can't subtract offset-naive and offset-aware datetimes` when the `.aws-sam/sync.toml` file contains timestamps without timezone information (missing `+00:00` suffix). This breaks the watch functionality for users who have previously run sync commands, preventing them from using the auto-sync feature.

The fix ensures all timestamps loaded from `sync.toml` are timezone-aware UTC by checking if `tzinfo` is `None` after parsing with `datetime.fromisoformat()` and adding UTC timezone if missing. This maintains consistency with how SAM CLI writes timestamps (always UTC timezone-aware) and handles both old sync.toml files (without timezone info) and new ones (with timezone info) correctly. The fix is applied in two locations in `samcli/commands/sync/sync_context.py`:
1. When loading `latest_infra_sync_time` from the sync state
2. When loading `sync_time` for each resource sync state

None. The change is backward compatible and only affects the internal datetime handling. All timestamps are now consistently timezone-aware UTC, which is the expected behavior throughout the sync module.

#### Mandatory Checklist
- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/commands/sync/sync_context.py
```diff
@@ -129,9 +129,14 @@ def _toml_document_to_sync_state(toml_document: Dict) -> Optional[SyncState]:
     if resource_sync_states_toml_table:
         for resource_id in resource_sync_states_toml_table:
             resource_sync_state_toml_table = resource_sync_states_toml_table.get(resource_id)
+            sync_time_str = resource_sync_state_toml_table.get(SYNC_TIME)
```
seshubaws, line 171: it could be that the datetime has a Z at the end right, like 2024-05-08T15:16:43Z, should we do something to handle this

samcli/commands/sync/sync_context.py
```diff
@@ -142,9 +147,12 @@ def _toml_document_to_sync_state(toml_document: Dict) -> Optional[SyncState]:
     latest_infra_sync_time = None
     if sync_state_toml_table:
         dependency_layer = sync_state_toml_table.get(DEPENDENCY_LAYER)
-        latest_infra_sync_time = sync_state_toml_table.get(LATEST_INFRA_SYNC_TIME)
-        if latest_infra_sync_time:
-            latest_infra_sync_time = datetime.fromisoformat(str(latest_infra_sync_time))
+        latest_infra_sync_time_str = sync_state_toml_table.get(LATEST_INFRA_SYNC_TIME)
```
seshubaws, line 190: nit but this is the same logic as above, it can probably be made into a method to use in both places

samcli/commands/sync/sync_context.py
vicheey: TODO: 
[x] Add handler and update tests.

tests/unit/commands/sync/test_sync_context.py
```diff
@@ -403,3 +403,51 @@ def test_datetime_comparison_after_fix(self):
             self.assertIsNotNone(time_diff)
         except TypeError as e:
             self.fail(f"DateTime comparison failed with TypeError: {e}")
+
+
+class TestZSuffixDatetimeHandling(TestCase):
+    """Tests for PR #8487: Handle 'Z' suffix in datetime strings (Python 3.9/3.10 compatibility)"""
```
seshubaws: nit: can remove comment

tests/unit/commands/sync/test_sync_context.py
```diff
@@ -288,3 +288,166 @@ def test_sync_context_has_no_previous_state_if_file_doesnt_exist(self, patched_r
             self.assertIsNone(self.sync_context._previous_state)
             self.assertIsNotNone(self.sync_context._current_state)
             patched_rmtree_if_exists.assert_not_called()
+
+
+class TestTimezoneNaiveDatetimeHandling(TestCase):
+    """Tests for Issue #8477: Timezone-naive datetime handling in sync.toml"""
```
seshubaws: not sure why we're referencing the issue in the test comments

samcli/commands/sync/sync_context.py
```diff
@@ -65,6 +65,40 @@ def update_infra_sync_time(self) -> None:
...
+    Returns
+    -------
+    datetime
+        Timezone-aware datetime in UTC
+    """
+    # Handle 'Z' suffix for UTC timezone (Python 3.9/3.10 compatibility)
+    if datetime_str.endswith("Z"):
+        datetime_str = datetime_str[:-1] + "+00:00"
+
+    parsed_datetime = datetime.fromisoformat(datetime_str)
```
seshubaws: What if there is an invalid datetime string passed here? We should add error handling for that

samcli/commands/sync/sync_context.py
```diff
@@ -142,9 +179,10 @@ def _toml_document_to_sync_state(toml_document: Dict) -> Optional[SyncState]:
     latest_infra_sync_time = None
     if sync_state_toml_table:
         dependency_layer = sync_state_toml_table.get(DEPENDENCY_LAYER)
-        latest_infra_sync_time = sync_state_toml_table.get(LATEST_INFRA_SYNC_TIME)
-        if latest_infra_sync_time:
-            latest_infra_sync_time = datetime.fromisoformat(str(latest_infra_sync_time))
+        latest_infra_sync_time_str = sync_state_toml_table.get(LATEST_INFRA_SYNC_TIME)
```
seshubaws, line 190: are we sure LATEST_INFRA_SYNC_TIME will always exist?

seshubaws (approved):

samcli/commands/sync/sync_context.py
```diff
@@ -65,6 +65,48 @@ def update_infra_sync_time(self) -> None:
         self.latest_infra_sync_time = datetime.now(timezone.utc)
 
 
+def _parse_datetime_from_toml(datetime_str: str) -> datetime:
```
roger-zhangg: One thing I learned from my past experience is never write a cutsom function around date time. There will always be a conner case you didn't consider before. If this is a valid usecase let's try to find a public lib that does it.

vicheey: I am thinking of using epoch time instead of this timezone.

bnusunny (approved):

samcli/commands/sync/sync_context.py
vicheey: Implementation to use epoch completed.

samcli/commands/sync/sync_context.py
```diff
@@ -65,6 +65,45 @@ def update_infra_sync_time(self) -> None:
         self.latest_infra_sync_time = datetime.now(timezone.utc)
 
 
+def _parse_time_from_toml(time_value: Union[str, int, float, datetime]) -> Optional[datetime]:
+    """
+    Parse time from TOML file - supports both epoch and ISO format.
+
+    Handles legacy ISO format strings as sam migrat to using epoch timestamps for backward compatibility.
```
seshubaws: typo: as sam *migrates*

vicheey:
```suggestion
    Handles legacy ISO format strings as sam migrates to using epoch timestamps for backward compatibility.
```

vicheey: Committed change.

PR #8490, opened 2025-12-05
test: fix start-lambda-no-containers test

valerena:
Fix start-lambda-no-containers test and move the creation of the local Lambda client to a separate function

#### Which issue(s) does this change fix? / Why is this change necessary?
`TestFunctionNameFilteringWarmContainersLazy::test_no_containers_before_invoke_with_lazy` test has been failing because it finds one running container even though it shouldn't find any. This happens because another test from the same class does create containers, so then this test finds the containers from that other test. This test-separation should be handled by the variable `mode_env_variable`, which allows the check to only find the containers from this *class*. <- yes, from this class. The hash is created once per class, so the containers in all the tests on the same class will share the hash. 

Split the tests into different classes, so they have different `mode_env_variable` values, and the container count works correctly. 

Nothing. More classes.

I also split the creation of the local Lambda client (used to invoke the local server), which was being done inside every class manually. Now it's a unified method that gets called without any parameter, instead of having the build all the parameters from the client every time. 

#### Mandatory Checklist

- [] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


vicheey (approved):

seshubaws (approved):

vicheey:
The test is failing, pending merge of https://github.com/aws/aws-sam-cli/pull/8491

Issue #7744, opened 2024-11-26
Feature request: Add uv.lock support for python lambdas


JonZeolla:
### Describe your idea/feature/enhancement

Currently you can `sam build` with a `--manifest requirements.txt` but it would be nice to add support for `uv.lock` files as described [here](https://docs.astral.sh/uv/guides/projects/#uvlock)

lucashuy:
Hey, thanks for raising this up with us. Would it be fair to expand this further to supporting `uv` in general with SAM CLI's build processes? Currently, as you might already be aware, we only support `pip` as the tool to install and build Python projects, but we are open to requests.

JonZeolla:
Yes, exactly. In the interim, `uv` does have a [`pip`-compatible interface](https://docs.astral.sh/uv/pip/compatibility/) at `uv pip ...` which could be stepping stone

hellupline:
so far, `pip` is the official tool for python packages,
and the python ecosystem still not solidified on which tooling we should use, there is raw `pip`, `poetry`, `uv`, etc
I think the best aproach is for sam-cli to just support requirements.txt and those tooling export to requirements.txt

in `uv` case:

```uv pip compile -o src/requirements.txt pyproject.toml```
running this just before `sam build` should suffice

trallnag:
There is a pretty extensive documentation on using uv with AWS Lambda: https://docs.astral.sh/uv/guides/integration/aws-lambda/

mbtx2:
Yes, but nothing for SAM.  I think with all the traction uv is getting, this would be a very good addition to the SAM tooling if it could be added.

jinxiao:
Please correct me if I am wrong.
After I have analyzed the code in aws sam cli. There are two things before this can be supported:
1. make lambda-builders to support uv, this command is used to build the lambda code. There is an request_json will be generated then parsed by the lambda-builers, related github repo: https://github.com/aws/aws-lambda-builders 
2. (Optional) Make the build image of sam to support uv, since there is no UV installed by default build image.

reedham-aws:
This feature has been merged (at least in experimental form) in #8549. I will close this issue when it gets released in the next release (1.155.0).

PR #8549, opened 2026-01-07
feat: add experimental uv support

reedham-aws:
#7744

Adds a popular new build system with many improvements over pip: https://docs.astral.sh/uv/.

This PR updates SAM CLI to use the new UV Build workflow create in Lambda Builders: https://github.com/aws/aws-lambda-builders/pull/756. In this repo, we have to add it as a possible option, and it's currently set as an experimental option that's enabled by a beta flag, like `SAM_CLI_BETA_UV_PACKAGE_MANAGER=1 sam build`.

Note: At the time of this PRs creation, the Lambda Builders PR is not merged and released, so this will fail because that backend is not available.

It shouldn't have side effects because this is only enabled if a user uses the `SAM_CLI_BETA_UV_PACKAGE_MANAGER` environment variable.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


schema/samcli.json
reedham-aws, line 1: I'm not sure where these changes come from, they came up after running `make pr` locally.

sriram-mv:
1/ We should add a sample project in https://github.com/aws/aws-sam-cli-app-templates so that you can init a project with which you can use this python uv builder along with `sam build` natively.
2/ we should add integ tests covering the new builder invocation via `SAM CLI`

samcli/lib/build/workflow_config.py
```diff
@@ -335,3 +349,17 @@ def get_config(self, code_dir: str, project_dir: str) -> CONFIG:
     @staticmethod
     def _has_manifest(config: CONFIG, directory: str) -> bool:
         return os.path.exists(os.path.join(directory, config.manifest_name))
+
+
+class ConditionalWorkflowSelector(BasicWorkflowSelector):
```
sriram-mv: can I specify which builder I want to use directly say within the template? this way I can override which builder gets chosen.

reedham-aws: Theoretically yes using the `BuildProperties` section of `Metadata` for a Lambda function. With the way it's set up right now to use an experimental flag, that's tough because that's global, but I wouldn't mind taking another crack at doing it on a per function basis.

reedham-aws: I guess we could still use the experimental flag as a wrapping for this solution, so I can try to update that for this PR.

reedham-aws: Completed this with https://github.com/aws/aws-sam-cli/pull/8549/commits/b7f8673b4ad69391b3edd4646b9100257c83b3df

tests/integration/buildcmd/test_build_cmd_python.py
```diff
@@ -260,14 +262,24 @@ def test_windows_dockerfile_present_sub_dir(self, dockerfile, expected):
...
+        "template",
     ),
     [
-        ("python3.9", "Python"),
-        ("python3.10", "Python"),
-        ("python3.11", "Python"),
-        ("python3.9", "PythonPEP600"),
-        ("python3.10", "PythonPEP600"),
-        ("python3.11", "PythonPEP600"),
+        ("python3.9", "Python",),
```
reedham-aws: I'm open to different ways to set up the parameterization. I did this because I thought it would be best to test existing cases entirely with a `BuildMethod` of uv (lines 274-279) as well as when we're using `pyproject.toml` only.

tests/integration/buildcmd/test_build_cmd_python.py
```diff
@@ -249,6 +249,8 @@ def test_windows_dockerfile_present_sub_dir(self, dockerfile, expected):
 #     ((not RUN_BY_CANARY) or (IS_WINDOWS and RUNNING_ON_CI) and not CI_OVERRIDE),
 #     "Skip build tests on windows when running in CI unless overridden",
 # )
+# FIXME: This second @parameterized_class is not doing anything because these decorators cannot be stacked.
+# We need to integrate this into a single decorator (that will be huge). This is also true of a few classes below.
```
reedham-aws, line 253: I can follow up with this, just didn't want to have a different fix in this PR when this bug is not hurting anything at the moment.

reedham-aws:
Tests are failing because we're not consuming a version of aws-lambda-builders that has the UV Builder. These won't work until we're consuming the so far unreleased 1.62.0 of builders

samcli/lib/build/workflows.py
```diff
@@ -24,6 +24,15 @@
     must_mount_with_write_in_container=False,
 )
 
+PYTHON_UV_CONFIG = CONFIG(
+    language="python",
+    dependency_manager="uv",
+    application_framework=None,
+    manifest_name=None,
```
sriram-mv, line 33: pyproject.toml for manifest name? or does it not matter?

tests/integration/testdata/buildcmd/PythonPyProject/pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[project]
+name = "python-uv-test"
+version = "0.1.0"
+dependencies = [
+  "numpy==2.0.0; python_version <= '3.9'",
```
sriram-mv, line 5: cryptography was another dependency that we can use to test.

tests/integration/buildcmd/test_build_cmd_python.py
```diff
@@ -339,45 +401,61 @@ class TestBuildCommand_PythonFunctions_WithDocker(BuildIntegPythonBase):
...
+            # ("python3.9", "Python", "template_uv.yaml"),
+            # ("python3.10", "Python", "template_uv.yaml"),
+            # ("python3.11", "Python", "template_uv.yaml"),
+            # ("python3.9", "PythonPyProject", "template_uv.yaml"),
+            # ("python3.10", "PythonPyProject", "template_uv.yaml"),
+            # ("python3.11", "PythonPyProject", "template_uv.yaml"),
         ]
     )
-    def test_with_default_requirements_in_container(self, runtime):
+    def test_with_default_requirements_in_container(self, runtime, codeuri, template="template.yaml"):
```
sriram-mv, line 422: I forget, can you help me recollect, do we local invoke after building?

sriram-mv (commented):
Overall looks good, I'd like to see gates green on integ tests after lambda builders is released and tests are uncommented.

samcli/lib/build/workflows.py
reedham-aws: I had to do this because of how SAM CLI passes the manifest name to Lambda Builders. If we set the manifest explicitly as `pyproject.toml`, Builders won't read a `requirements.txt` and successfully autodetect which manifest it needs to read. This is important in the case where a user has both `pyproject.toml` and `requirements.txt`, which is valid as both a pip and uv configuration. I think it's also fine since UV support is explicitly set in the template, so we don't need to figure out based on the manifest.

tests/integration/buildcmd/test_build_cmd_python.py
reedham-aws: Yes, the tests are calling local invoke after building.

tests/integration/testdata/buildcmd/PythonPyProject/pyproject.toml
reedham-aws: Originally, I excluded it because every instance of `cryptography` was commented out in the test data.

However, after I look closer I see that there's a comment that explains why: https://github.com/aws/aws-sam-cli/blob/75bc6c4337dcd4fbf4f8fdb3b73275630f698c40/tests/integration/testdata/buildcmd/Python/requirements.txt#L7-L9

That issue (https://github.com/aws/aws-lambda-builders/issues/29) was actually resolved two days after the comment was added ... in 2018. I think we could add it back in all the directories but I don't really want to mix an unknown change in with all the other ones.

samcli/lib/build/workflows.py
sriram-mv: Ack, can we clarify that with a comment?

tests/integration/testdata/buildcmd/PythonPyProject/pyproject.toml
sriram-mv: thanks for the dive!

bnusunny (approved):

tests/integration/testdata/buildcmd/PythonPyProject/pyproject.toml
seshubaws: can we track adding cryptography back in test data, with a new issue or sim?

seshubaws (approved):

tests/integration/testdata/buildcmd/PythonPyProject/pyproject.toml
reedham-aws: https://github.com/aws/aws-sam-cli/issues/8654

PR #8559, opened 2026-01-09
test: install latest versions of python/java in linux-binary tests

valerena:

In linux-binary integration tests, we were not installing Python 3.14 or java 25. 

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


appveyor-linux-binary.yml
```diff
@@ -82,6 +82,15 @@ install:
   - sh: "./aws_cli/bin/python -m pip install awscli"
   - sh: "PATH=$(echo $PWD'/aws_cli/bin'):$PATH"
 
+  # install python 3.14
```
reedham-aws, line 85: The code itself seems fine, but I personally don't understand what we're adding here. Why was 3.13 not being installed before (or any other version for that matter)?

valerena: That's a good question. I mostly brought the code from the `appveyor-ubuntu` definition. My guess is that the AppVeyor host has some versions of Python installed, so we're just using those without having to install them ourselves, but it doesn't have Python3.14

Issue #6900, opened 2024-04-03
Bug: invocation-type: Event is not supported. RequestResponse is only supported.


ijtarano:
### Description:
It seems like RequestResponse in boto3 invoke method doesn't work with SAM Local even when boto3 accept this parameter (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/invoke.html)

### Steps to reproduce:
```python 
import boto3
import json 

payload = {
}

lambda_client = boto3.client("lambda", endpoint_url="http://localhost:3001", region_name="eu-west-1")
lambda_function_name = "Lambda"

payload_json = json.dumps(payload)

response = lambda_client.invoke(
    InvocationType="Event", # This Event fails
    FunctionName=lambda_function_name,
    Payload=payload_json
)

```

### Observed result:
ClientError: An error occurred (NotImplemented) when calling the Invoke operation: invocation-type: Event is not supported. RequestResponse is only supported.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. Amazon Linux:
2. `sam --version`: SAM CLI, version 1.113.0
3. AWS region: eu-west-1

```
{
  "version": "1.113.0",
  "system": {
    "python": "3.11.3",
    "os": "Linux-4.14.320-243.544.amzn2.x86_64-x86_64-with-glibc2.26"
  },
  "additional_dependencies": {
    "docker_engine": "20.10.23",
    "aws_cdk": "Not available",
    "terraform": "1.7.2"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_RUST_CARGO_LAMBDA"
  ]
}
```

ijtarano:
Yes, of course, It works perfectly with "RequestResponse" parameter ;)

lucashuy:
Thanks for letting us know, this is something that we guard against since async execution isn't something that is currently implemented. I can move this issue to a feature request instead.

As an alternative, it's possible to use our `sam sync --watch` command to deploy to AWS and update code live, similar to how one would use `sam local start-lambda`. This could be a workaround for now.

ericbambach-qwickrate:
Bump - this feature is a use-case for us and we want to test async events locally. As a workaround we use RequestResponse but its not ideal.

[...]

PR #749, opened 2018-11-05
feat(event-invocation-type): Add support for Event invocation type 

ejhayes:
*Issue #, if available:*
This is mentioned in several different issues:
#510
#207 
#12 

*Description of changes:*
This adds support for the event invocation type. #508 made it possible to call local lambda using AWS CLI, but only for `RequestResponse` invocation type. This adds on to the local lambda invocation functionality by allowing `Event` invocation type.

This allows multiple lambda functions to call other lambda functions without having to wait for them to return (i.e. `RequestResponse` invocation type).

*Usage*
Run local lambda service:

```
samdev local start-lambda --debug
```

You'll see that local lambda service is at 127.0.0.1:3001:
```
2018-11-05 15:48:24 Starting the Local Lambda Service. You can now invoke your Lambda 
Functions defined in your template through the endpoint.
2018-11-05 15:48:24 Localhost server is starting up. Multi-threading = True
 * Tip: There are .env files present. Do "pip install python-dotenv" to use them.
2018-11-05 15:48:24  * Running on http://127.0.0.1:3001/ (Press CTRL+C to quit)
```

Test invocation using CLI:
```
aws --endpoint-url http://127.0.0.1:3001 lambda invoke --function-name 'MyFunctionName' --payload '{}' --invocation-type Event function_output.txt
```

This can also be done within a lambda function. Note that the endpoint referenced cannot be 127.0.0.1 since the function code will be running within a container. Docker for Mac has a special hostname for the underlying host machine named `host.docker.internal`. Additional documentation for that can be found [here](https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host):

```
// since this is running in a container, must reference the host machine
// on Mac this is host.docker.internal.
// https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
var ep = new aws.Endpoint('http://host.docker.internal:3001');

// configure lambda to use local endpoint
var lambda = new aws.Lambda({endpoint: ep});

// invoke using Event InvocationType
lambda.invoke({
	FunctionName: 'MyFunctionName',
	InvocationType: 'Event'
}, (err,data) => {
	if (err) {
		console.log('Error: ', err);
		console.log(err);
	}

	console.log('Function Result: ', JSON.stringify(data));		
	
	callback(null, {statusCode: 200, body: 'Done.'});
})
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #510, opened 2018-06-28
Is it possible to test lambda locally which internally call another lambda using SAM LOCAL


suneelthakur1980:
We are in process writing an integration test for lambda.

Here is the situation LambdaA calls  LambdaB internally.
Now I am invoking sam lambda A which internally call lmbdaB...  ON AWS console this is working but not able to work it from SAM local.

However, if lambaB invocation is removed from LambdaA, both lambda works fine in SAM LOCAL.

Is this supported by SAM local, any workaround to move forward?
code snippets

var AWS = require('aws-sdk');
AWS.config.region = 'eu-west-1';
var lambda = new AWS.Lambda();

exports.handler = function(event, context,callback) {
  
  var params = {
   FunctionName: 'lambda-B', // the lambda function we are going to invoke
    InvocationType: 'RequestResponse',
    LogType: 'Tail',
    Payload: JSON.stringify(event,null,null)
  };
  lambda.invoke(params, function(err, data) {
    
const dataPayload=data.Payload;
    console.log("------dataPayload-----"+dataPayload);
    console.log("------Now it should work-----"+JSON.parse(dataPayload).errorType);    
    if (err) {  context.fail(err);
    } else {
      context.succeed('Lambda_SIGNATURE said '+ JSON.stringify(data));      
    }

sanathkr:
Actually, this exact scenario is being enabled by #508. More details on this soon, but the tl;dr version is, we will provide a new command (possibly, `sam local start-lambda`) that will setup a local HTTP endpoint where you can send Lambda Invokes to.

This allows you to perform Lambda calling Lambda locally by just setting a different endpoint something like:

```javascript
var lambda = new AWS.Lambda()
if(process.env.AWS_SAM_LOCAL) {
  // only run inside local lambda runner
   var ep = new AWS.Endpoint('localhost:3000');
   var lambda = new AWS.Lambda({endpoint: ep})
}
```

suneelthakur1980:
Thanks, Sanath for the reply.
That means current SAM Local version 0.4.0 does not support invoking one lambda form another. 
Also if we have complex scenerio where LambdaA -->invoke-->LambdaB --> LambdaC  then  we will have two endpoints for each lambda?  Can we have something like register lambda feature in docker image where we could invoke lambda-like lambda.invoke(params, function(err, data) ..

sanathkr:
[...]

PR #8590, opened 2026-01-25
feat: support for local Event invocation type

monroeclinton:
This PR is a continuation of the discussion from #6900, https://github.com/aws/aws-sam-cli/issues/510#issuecomment-451268559, and #749.

#6900

This is necessary to allow running `Event` invocation types locally to make development more pleasant.

This PR uses `ThreadPoolExecutor` to create an executor that runs `Event` invocation types, allowing for non-blocking invocations. 

- `Event` invocations now return a response with a `202` status code.
- `Event` invocations that have an invalid function return a `404` with `ResourceNotFound`.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/local/lambda_service/local_lambda_http_service.py
+++ b/samcli/local/lambda_service/local_lambda_http_service.py
@@ -8,11 +8,16 @@ from datetime import datetime
 from typing import Dict, Optional, Tuple
 from urllib.parse import unquote
 
-from flask import Flask, Response, request
+from flask import Flask, request
 from werkzeug.routing import BaseConverter
 
 from samcli.commands.local.cli_common.durable_context import DurableContext
-from samcli.commands.local.lib.exceptions import TenantIdValidationError, UnsupportedInlineCodeError
+from samcli.commands.local.lib.exceptions import (
+    InvalidIntermediateImageError,
+    TenantIdValidationError,
+    UnsupportedInlineCodeError,
+)
+from samcli.lib.providers.provider import Function
 from samcli.lib.utils.invocation_type import EVENT
 from samcli.lib.utils.name_utils import InvalidFunctionNameException, normalize_sam_function_identifier
 from samcli.lib.utils.stream_writer import StreamWriter
@@ -244,42 +249,41 @@ class LocalLambdaHttpService(BaseLocalService):
         # Extract durable execution name from headers
         durable_execution_name = flask_request.headers.get("X-Amz-Durable-Execution-Name")
 
+        headers = {"Content-Type": "application/json"}
+
+        try:
+            function = self.lambda_runner.get_function(function_name, tenant_id)
+        except (InvalidFunctionNameException, TenantIdValidationError, InvalidIntermediateImageError) as e:
+            LOG.error("Validation error: %s", str(e))
+            return LambdaErrorResponses.validation_exception(str(e))
+        except FunctionNotFound:
+            normalized_function_name = normalize_sam_function_identifier(function_name)
+            LOG.debug("%s was not found to invoke.", normalized_function_name)
+            return LambdaErrorResponses.resource_not_found(normalized_function_name)
+        except UnsupportedInlineCodeError:
+            return LambdaErrorResponses.not_implemented_locally(
+                "Inline code is not supported for sam local commands. Please write your code in a separate file."
+            )
+
         arguments = {
             "function_name": function_name,
             "request_data": request_data,
             "invocation_type": invocation_type,
             "durable_execution_name": durable_execution_name,
             "tenant_id": tenant_id,
+            "function": function,
         }
 
-        headers = {"Content-Type": "application/json"}
-
         if invocation_type == EVENT:
-            # Validate function exists before submitting async task
-            if validation_error := self._validate_function_for_invocation(function_name):
-                return validation_error
-
-            # LocalLambdaService is blocking, so threads used by ThreadPoolExecutor
-            # will be cleaned up when Python exits by the ThreadPoolExecutor implementation
             self.executor.submit(self._invoke_async_lambda, **arguments)
             return self.service_response("", headers, 202)
-
         try:
             invoke_headers, stdout_stream_string, stdout_stream_bytes = self._invoke_lambda(**arguments)
-        except (InvalidFunctionNameException, TenantIdValidationError) as e:
-            LOG.error("Validation error: %s", str(e))
-            return LambdaErrorResponses.validation_exception(str(e))
         except UnsupportedInvocationType as e:
             LOG.warning(
-                "invocation-type: %s is not supported. Event and RequestResponse are only supported.", invocation_type
+                "invocation-type: %s is not supported. Only Event and RequestResponse are supported.", invocation_type
             )
             return LambdaErrorResponses.not_implemented_locally(str(e))
-        except FunctionNotFound:
-            return self._handle_function_not_found(function_name)
-        except UnsupportedInlineCodeError:
-            return LambdaErrorResponses.not_implemented_locally(
-                "Inline code is not supported for sam local commands. Please write your code in a separate file."
-            )
         except DockerContainerCreationFailedException as ex:
             return LambdaErrorResponses.container_creation_failed(ex.message)
 
@@ -295,47 +299,6 @@ class LocalLambdaHttpService(BaseLocalService):
             headers["x-amz-function-error"] = "Unhandled"
 
         return self.service_response(lambda_response, headers, 200)
-
-    def _validate_function_for_invocation(self, function_name: str) -> Optional[Response]:
-        """
-        Validates that a function exists and can be invoked.
-
-        Parameters
-        ----------
-        function_name : str
-            Name or ARN of the function to validate
-
-        Returns
-        -------
-        Flask.Response or None
-            Error response if validation fails, None if validation succeeds
-        """
-        try:
-            self.lambda_runner.get_function(function_name)
-            return None
-        except FunctionNotFound:
-            return self._handle_function_not_found(function_name)
-        except InvalidFunctionNameException as e:
-            LOG.error("Validation error: %s", str(e))
-            return LambdaErrorResponses.validation_exception(str(e))
-
-    def _handle_function_not_found(self, function_name: str) -> Response:
-        """
-        Handles FunctionNotFound exception by returning appropriate error response.
-
-        Parameters
-        ----------
-        function_name : str
-            Name or ARN of the function that was not found
-
-        Returns
-        -------
-        Flask.Response
-            Error response for function not found
-        """
-        normalized_function_name = normalize_sam_function_identifier(function_name)
-        LOG.debug("%s was not found to invoke.", normalized_function_name)
-        return LambdaErrorResponses.resource_not_found(normalized_function_name)
 
     def _invoke_async_lambda(self, function_name: str, **kwargs) -> None:
         """
@@ -353,6 +316,7 @@ class LocalLambdaHttpService(BaseLocalService):
         invocation_type: str,
         durable_execution_name: Optional[str],
         tenant_id: Optional[str],
+        function: Optional[Function],
     ) -> Tuple[Optional[Dict[str, str]], io.StringIO, io.BytesIO]:
         """
         Invokes a Lambda function and returns the result
@@ -370,6 +334,7 @@ class LocalLambdaHttpService(BaseLocalService):
             invocation_type=invocation_type,
             durable_execution_name=durable_execution_name,
             tenant_id=tenant_id,
+            function=function,
             stdout=stdout_stream_writer,
             stderr=self.stderr,
         )
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -309,7 +309,7 @@ class LambdaRuntime:
                 if invocation_type not in [EVENT, REQUEST_RESPONSE]:
                     raise UnsupportedInvocationType(
                         f"invocation-type: {invocation_type} is not supported. "
-                        "Event and RequestResponse are only supported."
+                        "Only Event and RequestResponse are supported."
                     )
 
                 # The container handles concurrency control internally via its semaphore.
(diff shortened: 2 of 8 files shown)


monroeclinton:
@reedham-aws could you retry rerunning the failed tests? Much appreciated

monroeclinton:
@reedham-aws could I get workflow approval again? thanks

monroeclinton:
@reedham-aws @vicheey thank you, looks like tests are passing. Let me know if there's anything I can do to help get this reviewed, or any feedback I can address

tests/unit/local/lambda_service/test_local_lambda_http_service.py
```diff
@@ -437,7 +443,6 @@ def test_invoke_request_handler_async_invocation_returns_202(
             "",
             {
                 "Content-Type": "application/json",
-                "X-Amz-Durable-Execution-Arn": "arn:aws:lambda:us-west-2:123456789012:function:test-function:$LATEST/durable-execution/test-123",
```
reedham-aws, line 440: I don't quite follow what's happening here; why are we removing the durable execution ARN and changing the `InvocationType` to `DryRun` below?

tests/unit/local/lambda_service/test_local_lambda_http_service.py
```diff
@@ -469,17 +472,159 @@ def test_invoke_request_handler_async_invocation_unsupported_function_returns_er
...
+        # Verify error response is returned
+        self.assertEqual(response, "validation error response")
+        # Verify get_function was called to validate
+        lambda_runner_mock.get_function.assert_called_once_with("Invalid@Function#Name")
+        # Verify executor was NOT called since validation failed
+        executor_mock.submit.assert_not_called()
+        # Verify error was logged
+        log_mock.error.assert_called_once_with("Validation error: %s", "Invalid function name format")
+        # Verify validation exception was called with the error message
+        lambda_error_responses_mock.validation_exception.assert_called_once_with("Invalid function name format")
```
reedham-aws, line 653: These tests feel superfluous considering we test the same things for synchronous events, but I think we can leave them to test if there's somehow an unintended divergence in the path of sync/async.

tests/unit/local/lambdafn/test_runtime.py
```diff
@@ -2255,3 +2312,37 @@ def test_should_reload_when_both_config_and_debug_context_changed(self):
 
         result = _should_reload_container(self.base_config, self.different_config, container, debug_context2)
         self.assertTrue(result)
+
+
+class TestLambdaRuntime_clean_runtime_containers(TestCase):
```
reedham-aws, line 2317: Are these tests related to changes in the PR? If so, could you explain why because I'm not following.

reedham-aws (commented):
The logic looks solid to me. I didn't think about it too hard, but one small concern I have is how we're testing some of this logic. I feel like we'd want to test that the threads are being executed and cleaned properly beyond the original return of the `202`. I think this is difficult to test considering it's a different process, but I'm not too sure. Do you have any thoughts on this @monroeclinton?

Side note: I didn't have time today, but maybe tomorrow I will try it out using `samdev` to confirm it's doing as expected.

tests/unit/local/lambdafn/test_runtime.py
monroeclinton: @reedham-aws I had to get the test coverage to 94%, after adding test coverage to my new lines it was around 93.96% if I recall correctly

tests/unit/local/lambda_service/test_local_lambda_http_service.py
monroeclinton: I believe this test case is asserting that the `Event` (async) invocation returns a `202` status code. This test case mocks the `invoke` function to return that `X-Amz-Durable-Execution-Arn` header. Since this PR now runs the `invoke` function on a thread, we are unable to get the `invoke` response before returning the response with a `202` status code. Prior to this PR, passing `Event` invocation type (like this test does) will cause an exception (as would this test, if not for the mock), so I'm not entirely sure why this test case was included in the durable execution PR. Ideally we could retrieve the ARN and return it, I tried to reverse engineer the `aws-durable-execution-emulator-x86_64` binary to see how the ARN is computed, but since the internal durable execution library uses Python 3.13 I couldn't fully disassemble it with `pycdc`.

For changing the `InvocationType` to `Event` below, since `Event` is now supported we need to pass in a different unsupported invocation type to test the error message when an unsupported invocation type is passed.

monroeclinton:
@reedham-aws I think that's a great idea for additional testing, I'll look more into testing the threads tomorrow

reedham-aws:
Thank you very much for you contributions and responses @monroeclinton!

tests/unit/local/lambda_service/test_local_lambda_http_service.py
reedham-aws: Makes sense. 

I just realized that the below changing from `Event` to `DryRun` is a different method, which definitely cleared up some confusion 😆

monroeclinton:
@reedham-aws I went with using `threading.Event` flags and assertions that the `invoke` function finishes executing (i.e. does not block) after the 202 response being returned. I'm more than happy to explore different ideas for asserting the Event invocation functionality

https://github.com/aws/aws-sam-cli/pull/8590/commits/2664b99b98f317df1971f4d88c269e9af14093ed

I was thinking about how best to assert cleanup of any threads, but since the `LocalLambdaService` is blocking, and only exits with SIGINT/SIGTERM, I think it's pretty much guaranteed that the threads will be cleaned up due to the `ThreadPoolExecutor` implementation

https://github.com/python/cpython/blob/1848ce61f349533ae5892a8c24c2e0e3c364fc8a/Lib/concurrent/futures/thread.py#L23

I had to bump test coverage more with:

https://github.com/aws/aws-sam-cli/pull/8590/commits/80729fb7f4490576b7df97f7a4cd2a0975b8f57e

tests/unit/local/lambda_service/test_local_lambda_http_service.py
```diff
@@ -469,17 +473,191 @@ def test_invoke_request_handler_async_invocation_unsupported_function_returns_er
...
+
+        lambda_runner_mock = Mock()
+        service = LocalLambdaHttpService(lambda_runner=lambda_runner_mock, port=3000, host="localhost")
+        service._invoke_lambda = fake_invoke
+        service.create()
+
+        response = service._invoke_request_handler(function_name="HelloWorld")
+
+        # Assert that first a 202 response is returned
+        assert response.status_code == 202
```
reedham-aws: Test itself looks great, but for style can we use `self.assertEqual(response.status_code, 202)` instead? Same for the other `assert`s below in this function. Sorry to be nitpicky!

reedham-aws (changes requested):
Thank you for the changes, had one really minor nit.

I also tested the feature using `samdev local start-lambda` to see how it handled async events. Everything works how I expected; the server properly returns `202`, but still executes after being enqueued. I tested by sending a stream of requests and got the responses immediately, and saw the invocation results properly stream in server side.

reedham-aws:
> I was thinking about how best to assert cleanup of any threads, but since the `LocalLambdaService` is blocking, and only exits with SIGINT/SIGTERM, I think it's pretty much guaranteed that the threads will be cleaned up due to the `ThreadPoolExecutor` implementation
> 
> https://github.com/python/cpython/blob/1848ce61f349533ae5892a8c24c2e0e3c364fc8a/Lib/concurrent/futures/thread.py#L23

Also, maybe we could add a comment to this effect where we do the `self.executor.submit`?

tests/unit/local/lambda_service/test_local_lambda_http_service.py
monroeclinton: You're good, I love pickiness

monroeclinton:
@reedham-aws Thanks for the feedback, updated the test assertions and added a comment about `ThreadPoolExecutor`

reedham-aws:
Thank you @monroeclinton for the attention to all my feedback! I will try to get another member of the team to review this PR so we can look at merging it.

tests/integration/local/start_lambda/test_start_lambda.py
```diff
@@ -87,14 +87,32 @@ def test_invoke_with_log_type_not_None(self):
...
         )
 
         with self.assertRaises(ClientError) as error:
             self.lambda_client.invoke(FunctionName="EchoEventFunction", InvocationType="DryRun")
 
         self.assertEqual(str(error.exception), expected_error_message)
 
+    @pytest.mark.flaky(reruns=3)
+    @pytest.mark.timeout(timeout=300, method="thread")
+    def test_invoke_with_event_invocation_type(self):
```
bchampp: We don't really test that the function itself _actually_ ran here. It would be nice to capture the stdout from the invocation in case something between where the executor queues the invocation and actually running the function breaks.

samcli/local/lambdafn/runtime.py
```diff
@@ -305,9 +306,10 @@ def invoke(
                 )
             else:
                 # Only RequestResponse supported for regular Lambda functions
-                if invocation_type != "RequestResponse":
+                if invocation_type not in [EVENT, REQUEST_RESPONSE]:
                     raise UnsupportedInvocationType(
-                        f"invocation-type: {invocation_type} is not supported. RequestResponse is only supported."
+                        f"invocation-type: {invocation_type} is not supported. "
+                        "Event and RequestResponse are only supported."
```
bchampp:
```suggestion
                        f"invocation-type: {invocation_type} is not supported. "
                        "Only Event and RequestResponse are supported"
```

tests/integration/local/start_lambda/test_start_lambda_cdk.py
```diff
@@ -112,7 +112,7 @@ def test_invoke_with_log_type_not_None(self):
     def test_invoke_with_invocation_type_not_RequestResponse(self):
```
bchampp, line 112: unrelated to this PR - this test feels like it would be more appropriate in just `test_start_lambda.py`. This file is intended to be CDK specific but has additional validation tests 🤔

tests/unit/local/lambda_service/test_lambda_error_responses.py
```diff
@@ -97,6 +97,19 @@ def test_generic_method_not_allowed(self, service_response_mock):
             405,
         )
 
+    @patch("samcli.local.services.base_local_service.BaseLocalService.service_response")
+    def test_validation_exception(self, service_response_mock):
```
bchampp, line 101: thank you 🫡

samcli/local/lambda_service/local_lambda_http_service.py
```diff
@@ -278,20 +288,94 @@ def _invoke_request_handler(self, function_name):
...
+        return LambdaErrorResponses.resource_not_found(normalized_function_name)
+
+    def _invoke_async_lambda(self, function_name: str, **kwargs) -> None:
+        """
+        Wrapper for _invoke_lambda that runs in an async context (Event invocation type)
+        """
+        try:
+            self._invoke_lambda(function_name=function_name, **kwargs)
+        except Exception as e:
+            LOG.error("Async invocation failed for function %s: %s", function_name, str(e), exc_info=True)
```
bchampp, line 310: It makes sense to sink the exceptions like you're doing here, but how does this appear for the user? Do you have a screenshot of what it looks like when you try async invoking a function that has a syntax error (so the sandbox crashes)?

samcli/local/lambda_service/local_lambda_http_service.py
```diff
@@ -278,20 +288,94 @@ def _invoke_request_handler(self, function_name):
...
+    ) -> Tuple[Optional[Dict[str, str]], io.StringIO, io.BytesIO]:
+        """
+        Invokes a Lambda function and returns the result
+        """
+
+        stdout_stream_string = io.StringIO()
+        stdout_stream_bytes = io.BytesIO()
+        stdout_stream_writer = StreamWriter(stdout_stream_string, stdout_stream_bytes, auto_flush=True)
+
+        normalized_function_name = normalize_sam_function_identifier(function_name)
```
bchampp, line 329: Should we call the `get_function()` utility you added here instead?

samcli/local/lambda_service/local_lambda_http_service.py
monroeclinton: image

There's two cases, one if the invocation fails, with the exception above. Then the case if the user's code raises an exception seen below, where it will log the exception from the container.

image
 
image

bchampp (approved):

monroeclinton:
@bchampp @reedham-aws My https://github.com/aws/aws-sam-cli/pull/8590/commits/57ad2a472cc093f196f586a4463cbabaa83b0f5f commit addresses:

- I realized that `Event` invocation type should probably also throw exceptions for invalid function package type, invalid function architecture, and invalid function tenant id, so I moved all the function validation into `get_function`.
- I added tests for the invalid function exceptions for `Event` invocation type + additional tests for missing `ImageUri` and missing `TenantId`.
- I added an integration test asserting that the `Event` invocation type runs the function.
- I updated `local_lambda_http_service.py` to use the `get_function` utility, and pass `function` as an optional argument to `invoke` so we aren't recomputing it.

Let me know if this looks good

tests/integration/local/start_lambda/test_start_lambda.py
monroeclinton: Added a test for this

bchampp:
> I moved all the function validation into get_function.

I think it might be more clear if we had a separate `validate_tenancy()` function instead of grouping the logic into the `get_function`. It just feels a little out of place to me since it's the only additional field to the function identifier.

monroeclinton:
@bchampp yeah, it's tricky to express the logic, I was thinking on how to break it out, but I think `get_function` should only return a throwable  or a valid `Function`, so raising `TenantIdValidationError` in `get_function` is necessary

monroeclinton:
@reedham-aws @bchampp looks like tests are passing, merge when ready

monroeclinton:
@reedham-aws @roger-zhangg I don't have permission to merge, you'll have to do it

roger-zhangg:
This PR breaks Durable function execution feature (https://github.com/aws/aws-sam-cli/actions/runs/21737966967/job/62707016257). We added fix #8636 without reverting

monroeclinton:
Thanks for finding and fixing, I found the durable function execution confusing because of the compiled binary. It'd be nice to have an open source version of it

PR #8599, opened 2026-01-27
fix(unzip): replace UserException with ValueError exception in `_extract()`

vicheey:
- Fixes Terraform integration test failures with `ModuleNotFoundError: No module named 'samcli'`
- Also, add tmp folder generated during integration test run in `.gitignore`

During Terraform builds, SAM CLI copies `zip.py` from `samcli/local/lambdafn/` to an isolated temporary directory (`.aws-sam-iacs/iacs_metadata/`) where it's executed as part of the build process. The original code imported `UserException` from `samcli.commands.exceptions`, but the `samcli` package is not available in this isolated directory, causing the import to fail with `ModuleNotFoundError`.

This change replaces the custom `UserException` with Python's built-in `ValueError` exception in the `_extract` function within `samcli/local/lambdafn/zip.py`. This eliminates the dependency on SAM CLI's internal module structure, making `zip.py` self-contained and able to run successfully when copied to the isolated metadata directory. The error handling behavior remains functionally equivalent.

- File: `samcli/hook_packages/terraform/hooks/prepare/makefile_generator.py` (lines 96-97)
```python
ZIP_UTILS_MODULE_script_path = os.path.join(samcli_root_path, "local", "lambdafn", ZIP_UTILS_MODULE)
shutil.copy(ZIP_UTILS_MODULE_script_path, output_directory_path)
```

#### Mandatory Checklist

- [ ] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -407,6 +407,7 @@ cov.xml
 coverage.xml
 
 # Temporary scratch directory used by the tests
+tests/integration/*/tmp/*
 tests/integration/buildcmd/scratch
 tests/integration/testdata/buildcmd/Dotnet*/bin
 tests/integration/testdata/buildcmd/Dotnet*/obj
@@ -426,5 +427,3 @@ mise.toml
 
 # Durable executions
 **/.durable-executions-local/
-
-tests/integration/*/tmp/*


.gitignore
```diff
@@ -425,4 +425,6 @@ samcli/lib/init/templates/cookiecutter-aws-sam-hello-java-gradle/**/.gradle/
 mise.toml
 
 # Durable executions
-**/.durable-executions-local/
\ No newline at end of file
+**/.durable-executions-local/
+
+tests/integration/*/tmp/*
```
valerena: You should move this next to the other `tests/integration/` locations: https://github.com/aws/aws-sam-cli/blob/3ea9d6dd8de25431c8441bc08321e1502e594382/.gitignore#L410-L416

PR #7876, opened 2025-02-07
Expanded parameter overrides

sigJoe:
#### Which issue(s) does this change fix?
Issues: #2253 #2380
Discussions: #4591 #6377

#### Why is this change necessary?
It's a roundup of quality of life improvements for users of SAM CLI that need to manage multiple environments with different parameters.

#### How does it address the issue?
- Merges CLI `--parameter-overrides` with samconfig `parameter_overrides`, rather than completely replacing them.
- Adds support for YAML dictionary syntax for key-value pairs (previously only supported lists of strings)
- Flattens nested lists to better support YAML anchors and aliases for list merging
- Adds support for external file references (e.g. `file://MyParams.yaml` or `file://MyParams.toml`)
  - YAML and TOML work with key-value pairs as you'd expect - just don't use tables (sections) in the TOML or you're gonna have a bad time
  - JSON **SHOULD** just work as soon as `FILE_MANAGER_MAPPER` adds it as a supported option. 
- Clear precedence: template -> samconfig -> CLI, with later parameters overwriting previous ones

#### What side effects does this change have?
If someone is **relying** on the previous behaviour where samconfig parameter_overrides get completely discarded when specifying CLI parameter-overrides, this PR could be an issue.

#### Mandatory Checklist
**PRs will only be reviewed after checklist is complete**

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [x] Write/update unit tests
- [ ] Write/update integration tests
- [ ] Write/update functional tests if needed
- [x] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).

PR #8604, opened 2026-01-28
More expanded parameter overrides

sigJoe:
This PR adds two enhancements to the parameter override file functionality introduced in #7876:

1. **Relative path resolution for nested includes** - Parameter files can now reference other files using paths relative to their own location, rather than requiring paths relative to the project root
2. **`$include` key support** - Dict-based parameter files (YAML/TOML) can now include other files using a special `$include` key, providing the same composition capabilities that list-based files already had

1. **Improved ergonomics for nested includes**: When organizing config files in subdirectories (e.g., `./config/a.yaml` and `./config/b.yaml`), it's unintuitive for `a.yaml` to reference `b.yaml` as `file://config/b.yaml` instead of simply `file://b.yaml`. Relative path resolution makes file organization more natural and maintainable.
2. **TOML parity with YAML**: The previous implementation only supported file includes via list syntax, which worked for YAML but did not allow TOML dict-based files to include other files.

1. **Relative path resolution**: Added a `current_file` parameter to track the file being processed. When a `file://` reference is encountered, relative paths are resolved against `current_file.parent` instead of the project root.
2. **`$include` key**: Added support for a special `$include` key in dict-based parameter files that accepts either a single file path or a list of file paths. I picked this key name because it is not a valid cloudformation parameter so it should be fully backwards compatible. Another option was `__include` which didn't need to be quote-escaped in TOML but overall didn't look as pretty.

This changes how `file://` references work for parameter overrides in a way that is not backwards compatible. However, impact is limited since:
- The feature was introduced very recently (#7876)
- It has not yet been documented in official AWS documentation
- The new behavior is more intuitive and aligns with user expectations

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

# Documentation 
## Old and busted
When using `file://` to include external parameter overrides, there were limitations to include further levels of nested files. YAML had to be a list:
```yaml
- StringParam: Pam Param
- NumberParam: 42
- file://other_yaml_params.yaml
- file://toml_params_because_I_can.toml
```
TOML just didn't support it:
```toml
StringParam = "Pam Param"
NumberParam = 42
file://another_toml_file.toml # This will fail because it is not valid TOML
```

## New hotness
The root object can now be a dict if you want using `$include`:
```yaml
# YAML
$include: file://base.yaml
Environment: prod-b
```

And TOML can now join the party too:
```toml
# TOML
'$include' = 'file://default.toml'
Environment = 'prod-b'
```

It's worth nothing that `$include` is a generic parameter inclusion that supports ALL parameter formats. It is not specifically limited to `file://` references. For example, the following snippets are also valid:

```yaml
# YAML
$include: foo=bar
```

```yaml
# YAML
$include:
  - file://default.yaml
  - file://prod.yaml
  - foo: bar
  - x=y
Environment: prod-b
```

```toml
# TOML
"$include" = [
  "file://default.yaml",
  "file://prod.yaml",
  { foo = "bar" },
  "x=y"
]
Environment = 'prod-b'
```


sigJoe:
@bnusunny 

I added a test for `$include` circular references and the type hinting you requested. However, the third one might be worth a small discussion.

> 3. Edge case question: What happens if someone writes `$include: "NotAFileReference" `(without `file://` prefix)? It looks like it would fall through to the legacy parameter matching and fail with a potentially confusing error. Might be worth either:
>    
>    * Adding validation that `$include` values must start with `file://`
>    * Or just documenting the expected format

You are correct it would fall through to legacy parameter matching. It would throw an error `NotAFileReference is not a valid format. It must look something like <list of valid formats>`.

Right now `$include` is just a dummy key that says "process these extra parameters". It was necessary to make `file://` work in nested TOML, but more than that it accepts **_all_** the parameter formats. For example, somebody could technically do this:
```yaml
env: prod
$include:
  - file://morestuff.yaml
  - foo: bar
```
or in TOML:
```toml
env = "prod"
"$include" = [
    "file://morestuff.yaml",
    { foo = "bar" }
]
```
It's very non-opinionated and flexible rather than being limited to file includes, although I could understand if you want this to be a bit more limited.

bnusunny:
Thanks for the follow-up @sigJoe! I think keeping $include generic makes sense here. The flexibility to merge inline parameters alongside file references is actually a nice feature, and restricting it to only file:// would be an artificial limitation.

The current error message when someone passes an invalid value is clear enough - they'll see the list of valid formats. We can always tighten the validation later if users find it confusing, but I'd rather start permissive.

One small suggestion for the docs/PR description: maybe clarify that $include is really "merge these parameters" rather than strictly "include these files" - that way users understand they can mix file refs with inline params if needed.

Otherwise this looks good to merge.

sigJoe:
@bnusunny I updated the documentation in my first post to reflect the generic nature of `$include`. 

Also I'd love to merge this, but as an external contributor I don't have write access to the repository. If you're happy with the current state, could you please handle the merge on your end? Thanks!

PR #8607, opened 2026-01-30
chore: Add action to send PRs and issues to internal slack

seshubaws:

Tested locally with github action library https://github.com/nektos/act

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/notify-slack.yml
+++ b/.github/workflows/notify-slack.yml
@@ -5,16 +5,11 @@ on:
     types: [labeled]
   issues:
     types: [opened]
-  issue_comment:
-    types: [created]
 
 jobs:
   notify-slack:
     runs-on: ubuntu-latest
-    if: |
-      (github.event_name == 'pull_request' && github.event.label.name == 'pr/external') || 
-      github.event_name == 'issues' ||
-      (github.event_name == 'issue_comment' && github.event.issue.pull_request != null)
+    if: (github.event_name == 'pull_request' && github.event.label.name == 'pr/external') || github.event_name == 'issues'
     permissions:
       contents: read
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-slack.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  notify-slack:
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'pull_request' && github.event.label.name == 'pr/external') || github.event_name == 'issues'
+    permissions:
+      contents: read
+    steps:
+      - name: Send External PR Notification
+        if: github.event_name == 'pull_request'
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661 # v2
```
reedham-aws, line 18: Why do we use the commit hash instead of just `rtCamp/action-slack-notify@v2`?

On a separate note, are we confirmed able to use this external GH action?

seshubaws: This is what the owner of the slack bot uses so that's what I used: https://github.com/awslabs/LISA/blob/main/.github/workflows/test-and-lint.yml#L13C1-L36C80. I will ask if AppSec has only approved the hash or if we can use v2.

ETA: When AppSec did pen testing on the bot and repo, they made them pin a hash. It was one of the findings from the open source repo scanner. And to your second question, yes AppSec has approved it.

reedham-aws: Interesting, that hash is the commit with the latest version bump (v2.3.3). IIRC, most GH action versions are floating on the major version, like v2 will implicitly point to latest v2.x.x. That's probably why they make us pin the hash. I wonder if the owner of the slack bot is updating the hash in the workflow steps for every new release. Either way, thank you for the explanation.

bchampp: This was brought up for the durable function SDK repositories too - it's a best practice to pin a specific hash since they're immutable.

seshubaws: > that hash is the commit with the latest version bump (v2.3.3)

Apparently they only recently did the pen testing so that may be why.

PR #8650, opened 2026-02-12
fix: finch linux rootless containerd detection

### LINKED ISSUE #8649: Bug: SAM CLI Finch Detection Fails on Linux with Rootless Containerd
AWS SAM CLI versions 1.151.0 and 1.154.0 fail to detect Finch on Ubuntu Linux (arm64) when using rootless containerd, despite Finch being properly installed and functional. The error message is:

```
Error: Running AWS SAM projects locally requires a container runtime. Do you have Docker or Finch installed and running?
```

## Environment

- **OS**: Ubuntu 24.04.3 LTS
- **Architecture**: aarch64 (ARM64)
- **SAM CLI Version**: 1.151.0 and 1.154.0 (both tested)
- **Finch Version**: v1.14.1
- **Finch Installation Method**: Ubuntu apt repository (`runfinch-finch` package)
- **Containerd**: Rootless mode with nerdctl backend
- **Python**: 3.11.10

## Finch Configuration

Finch is installed via the official Ubuntu apt repository and configured with rootless containerd:

```bash
# Finch installation
curl -fsSL https://artifact.runfinch.com/deb/GPG_KEY.pub | sudo gpg --dearmor -o /usr/share/keyrings/runfinch-finch-archive-keyring.gpg
echo "deb [signed-by=/usr/share/keyrings/runfinch-finch-archive-keyring.gpg arch=$(dpkg --print-architecture)] https://artifact.runfinch.com/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/runfinch-finch.list
apt-get update
apt-get install -y runfinch-finch containerd

# Rootless containerd setup
loginctl enable-linger participant
mkdir -p /run/user/$(id -u participant)
export XDG_RUNTIME_DIR=/run/user/$(id -u)
containerd-rootless-setuptool.sh install
```

## Finch Works Correctly

Finch itself is fully functional:

```bash
$ finch --version
finch version v1.14.1

$ finch info
Client:
Namespace:     finch
Debug Mode:    false
Server:
Server Version: 1.7.28
Storage Driver: overlayfs
Logging Driver: json-file
Cgroup Driver: systemd
Cgroup Version: 2
...

$ finch ps
CONTAINER ID    IMAGE                                     COMMAND                   CREATED          STATUS    PORTS                     NAMES
6fa08d27c593    docker.io/amazon/dynamodb-local:latest    "java -jar DynamoDBL…"    51 seconds ago    Up        0.0.0.0:8000->8000/tcp    hotel-app-dynamodb-local
```

### Debug Output from SAM CLI 1.151.0

```
2026-02-12 20:03:08,098 | ContainerClientFactory.create_client() called
2026-02-12 20:03:08,098 | Admin preference: None
2026-02-12 20:03:08,099 | Using auto-detected client creation
2026-02-12 20:03:08,099 | Trying Docker client creation
2026-02-12 20:03:08,099 | Creating Docker container client from environment variable.
2026-02-12 20:03:08,100 | Creating container client with parameters: {'version': '1.35'}
2026-02-12 20:03:08,100 | DockerContainerClient created successfully
2026-02-12 20:03:08,101 | Fall back docker api version to 1.44: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
2026-02-12 20:03:08,103 | Docker daemon availability check failed with fallback: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
2026-02-12 20:03:08,103 | Docker client not created, trying creating Finch client.
2026-02-12 20:03:08,103 | Creating Finch container client with base_url=unix:///var/run/finch.sock
2026-02-12 20:03:08,104 | Creating container client with parameters: {'base_url': 'unix:///var/run/finch.sock', 'version': '1.35'}
2026-02-12 20:03:08,104 | FinchContainerClient created successfully
2026-02-12 20:03:08,105 | Finch daemon availability check failed: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
2026-02-12 20:03:08,106 | No container runtime available
```

### Analysis

SAM CLI is looking for Finch at `unix:///var/run/finch.sock`, but on Linux with rootless containerd, the socket is at:
- `$XDG_RUNTIME_DIR/containerd/containerd.sock` (typically `/run/user/1001/containerd/containerd.sock`)

Additionally, the containerd socket uses a different API protocol than Docker's socket, so even when pointed to the correct location, SAM CLI's Docker Python library cannot communicate with it.

### 1. Environment Variables
```bash
export XDG_RUNTIME_DIR=/run/user/$(id -u)
export AWS_SAM_CONTAINER_RUNTIME=finch
```
**Result**: No effect on detection

### 2. SAM CLI Metadata File
```bash
mkdir -p ~/.aws-sam
cat > ~/.aws-sam/metadata.json <<EOF
{
  "container_runtime": "finch"
}
EOF
```
**Result**: File is read but ignored; SAM CLI still tries socket detection

### 3. Socket Symlink
```bash
sudo ln -sf /run/user/$(id -u)/containerd/containerd.sock /var/run/finch.sock
```
**Result**: Connection reset by peer (protocol incompatibility)

### 4. DOCKER_HOST Environment Variable
```bash
export DOCKER_HOST="unix:///run/user/$(id -u)/containerd/containerd.sock"
```
**Result**: SAM CLI still cannot communicate with containerd socket

### 5. Docker Command Symlink to Finch
```bash
sudo ln -sf $(which finch) /usr/local/bin/docker
```
**Result**: SAM CLI still tries socket connection before CLI commands

### 6. Docker Command Symlink to nerdctl
```bash
sudo ln -sf $(which nerdctl) /usr/local/bin/docker
```
**Result**: Same issue - socket detection happens first

## Expected Behavior

According to the [AWS SAM CLI documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-finch.html):

> Starting with AWS SAM CLI version 1.145.0, AWS SAM CLI supports Finch as an alternative container runtime to Docker.
> 
> After you install Finch, the AWS SAM CLI automatically detects and uses Finch as a container runtime when Docker does not run.

SAM CLI should:
1. Detect that Finch is installed and available via the `finch` command
2. Use Finch's CLI interface instead of trying to connect to a socket
3. Fall back to using the `finch` command when socket detection fails

## Root Cause

The SAM CLI Finch detection logic appears to be designed primarily for macOS, where Finch uses a VM with a Docker-compatible socket. On Linux:

1. Finch uses nerdctl + rootless containerd directly (no VM)
2. The containerd socket location is user-specific (`$XDG_RUNTIME_DIR/containerd/containerd.sock`)
3. The containerd socket uses a different API than Docker's socket
4. SAM CLI should fall back to using the `finch` CLI command, but this fallback never triggers

## Proposed Solution

SAM CLI should:

1. **Check for `finch` command availability** before attempting socket connections
2. **Use CLI-based container operations** when the `finch` command is available, similar to how it would use the `docker` command
3. **Support `$XDG_RUNTIME_DIR`** for socket detection on Linux
4. **Provide a `--container-runtime` flag** to explicitly specify finch/docker (currently doesn't exist in 1.151.0 or 1.154.0)
5. **Honor the `AWS_SAM_CONTAINER_RUNTIME` environment variable** more reliably

## Workaround for Users

Currently, the only reliable workaround is to install Docker, which defeats the purpose of using Finch.

## Additional Context

- This issue affects workshop and tutorial environments where Docker cannot be installed
- Finch on Linux is officially supported as of October 2024
- The issue exists in both SAM CLI 1.151.0 and 1.154.0
- The `--container-runtime` flag mentioned in some documentation does not exist in these versions

## Related Issues

- #4584 - Original feature request for Finch support
- AWS Documentation: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/install-finch.html

robbrad:
This PR fixes Finch detection on Linux systems using rootless containerd by checking multiple standard socket locations instead of using a hardcoded path.

fixes #8649 

## Problem

SAM CLI fails with "No container runtime available" on Linux with rootless containerd despite Finch being installed and working. The issue is that SAM CLI only checks `/var/run/finch.sock`, but rootless containerd uses `$XDG_RUNTIME_DIR/containerd/containerd.sock`.

## Solution

Updated `LinuxHandler.get_finch_socket_path()` to check multiple locations in priority order:
1. `$XDG_RUNTIME_DIR/containerd/containerd.sock` (rootless containerd)
2. `$XDG_RUNTIME_DIR/finch.sock` (Finch-specific)
3. `~/.finch/finch.sock` (user home directory)
4. `/var/run/finch.sock` (system-wide fallback)
5. Returns `None` when no socket found (enables future CLI fallback)

## Changes

- Updated `samcli/local/docker/platform_config.py`
- Added comprehensive unit tests
- Added debug logging for troubleshooting

## Testing

- All unit tests pass
- Verified with custom test script
- Backward compatible - still checks system socket as fallback

## Risk

LOW - Only affects Linux socket detection, no changes to macOS or Windows

## Related Issues

Fixes issue where SAM CLI fails to detect Finch on Linux with rootless containerd.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/local/docker/platform_config.py
```diff
@@ -123,11 +123,49 @@ def _read_config(self) -> Optional[str]:
...
+            Optional[str]: Socket path if found, None otherwise
+        """
+        
+        # Check XDG_RUNTIME_DIR for rootless containerd
+        xdg_runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
+        if xdg_runtime_dir:
+            # Rootless containerd socket (most common on Linux)
+            containerd_sock = os.path.join(xdg_runtime_dir, "containerd", "containerd.sock")
+            if os.path.exists(containerd_sock):
+                LOG.debug(f"Found Finch socket at XDG_RUNTIME_DIR: {containerd_sock}")
```
reedham-aws: Is the `containerd_sock` exclusively for finch, though? Couldn't other container clients/engines be using this socket?

samcli/local/docker/platform_config.py
```diff
@@ -123,11 +123,49 @@ def _read_config(self) -> Optional[str]:
...
+            return f"unix://{home_finch_sock}"
+        
+        # System-wide socket (fallback)
+        system_sock = "/var/run/finch.sock"
+        if os.path.exists(system_sock):
+            LOG.debug(f"Found Finch socket at system location: {system_sock}")
+            return f"unix://{system_sock}"
+        
+        # No socket found - return None to enable future CLI fallback
+        LOG.debug("No Finch socket found in standard locations")
```
reedham-aws: To me it makes sense that this would be a `LOG.info` or `LOG.warn`.

reedham-aws (commented):
One thing that needs to be modified in this PR is actually where we're collecting metrics: https://github.com/aws/aws-sam-cli/blob/b9aa1ab968b11d64ab24f2bf010ac225534efcf6/samcli/lib/telemetry/metric.py?plain=1#L543-L572
You can see that we determine container engine based on the socket path. Anything that has `finch.sock` will be labeled as finch, but the `containerd.sock` will fall through to unknown. The part I don't quite understand is how we'd categorize containerd here; correct me if I'm wrong, but I don't think this is really a finch specific socket, so it doesn't really fit in that bucket. I left another comment on the PR to that effect.

On another note, I haven't fully reviewed the tests either, I can get to those later. There's also some formatting errors, but those could probably be cleaned up with `make format` or `make lint`.

samcli/local/docker/platform_config.py
```diff
@@ -123,11 +123,49 @@ def _read_config(self) -> Optional[str]:
```
Shubhranshu153, line 123: we should add a basic e2e test with rootless setup to see if it actually works.
Unit test i feel is not sufficient to check system related configs.

samcli/local/docker/platform_config.py
```diff
@@ -123,11 +123,49 @@ def _read_config(self) -> Optional[str]:
 
     def get_finch_socket_path(self) -> Optional[str]:
         """
-        Returns the socket path for Linux.
-        """
-
-        # Default fallback to system socket
-        return "unix:///var/run/finch.sock"
+        Returns the socket path for Linux, checking multiple locations.
```
Shubhranshu153, line 126: needs integ test to check builds work too in rootless mode?

samcli/local/docker/platform_config.py
Shubhranshu153: This will also check the networking works as we probably will need to pull some container.

reedham-aws: My one concern is that an added integration test for this situation will end up adding a lot of complexity to our testing structure, especially because this is a niche situation that requires a non-trivial amount of changes to the environment. I can discuss this further with the team.

samcli/local/docker/platform_config.py
robbrad: I think `LOG.warn` is a good choice here given people omit `--debug`

robbrad: Changed to LOG.warning due to deprecation notice

PR #8655, opened 2026-02-14
fix: update runc to 1.3.4 in Finch CI for symlink bind-mount support

bnusunny:
Finch/containerd fails with 'not a directory' when bind-mounting over a symlink due to a bug in the bundled runc. The Finch team confirmed the fix is to update runc to 1.3.4.

Instead of skipping the affected integration tests, this PR updates runc to 1.3.4 in the GitHub Actions CI pipeline after Finch is installed, which resolves the underlying issue.

### Changes
- Install runc 1.3.4 in the Finch setup step of `integration-tests.yml`, replacing the binaries at `/usr/sbin/runc` and `/usr/bin/runc`
- Revert the `@skipIf` test decorator on `TestInvokeBuildInSourceSymlinkedModules` so symlink tests run on all runtimes

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -250,6 +250,15 @@ jobs:
           echo 'deb [signed-by=/usr/share/keyrings/runfinch-finch-archive-keyring.gpg arch=amd64] https://artifact.runfinch.com/deb noble main' | sudo tee /etc/apt/sources.list.d/runfinch-finch.list
           sudo apt update
           sudo apt install -y runfinch-finch
+          
+          # Update runc to 1.3.4 to fix symlink bind-mount issue
+          RUNC_VERSION="1.3.4"
+          echo "Updating runc to v${RUNC_VERSION}..."
+          wget -q "https://github.com/opencontainers/runc/releases/download/v${RUNC_VERSION}/runc.amd64" -O /tmp/runc
+          sudo install -m 755 /tmp/runc /usr/sbin/runc
+          sudo install -m 755 /tmp/runc /usr/bin/runc
+          echo "runc updated: $(runc --version)"
+          
           sudo systemctl enable --now finch
           sudo systemctl enable --now finch-buildkit
           sleep 3
--- a/tests/integration/local/invoke/test_invoke_build_in_source.py
+++ b/tests/integration/local/invoke/test_invoke_build_in_source.py
@@ -2,11 +2,10 @@ from pathlib import Path
 import shutil
 import tempfile
 import os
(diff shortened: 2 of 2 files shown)


samcli/local/docker/container.py
```diff
@@ -347,13 +348,48 @@ def _create_mapped_symlink_files(self) -> Dict[str, Dict[str, str]]:
...
+                # Replace the symlink with an empty directory so that container runtimes
+                # (e.g. Finch/containerd) can create a valid mountpoint. Without this,
+                # runc fails with "not a directory" when it encounters a symlink at the
+                # mount target path inside the container's rootfs.
+                # We record the original symlink target so it can be restored after
+                # container creation (see _restore_mapped_symlinks).
+                try:
+                    symlink_path = file.path
+                    symlink_target = os.readlink(symlink_path)
+                    os.remove(symlink_path)
```
vicheey: Why do we remove the `symlink_path (file.path)` and recreate it againin the next line?

bnusunny: When SAM CLI mounts symlinks into containers (e.g. during sam local invoke with --build-in-source), Finch/containerd's runc fails with a "not a directory" error. Unlike Docker, which transparently handles symlinks at mount target paths, Finch expects an actual directory at the mount point inside the container's rootfs.

The fix does two things:

In _create_mapped_symlink_files, before mounting a symlink, it temporarily replaces the symlink on the host with an empty directory. This lets the container runtime create a valid bind mount. The original symlink target is recorded in self._replaced_symlinks.

A new _restore_mapped_symlinks method restores the original symlinks after the container is deleted (in the delete method), so the host filesystem is left in its original state for subsequent invocations.

vicheey: Shouldn't the symlink be restored after the container is created? why do we restore the symlink after the container is deleted?

bnusunny: No, we can't restore the symlink right after the container is created. The directory has to be there when the container is running. So we can only restore the symlink after the container is deleted.

seshubaws: Let's reach out to Finch team to see if they can support the same as what Docker does so we don't have to change how we implement this for different containers

roger-zhangg:
If we are just fixing the test, let's add a skip for finch for now. And a TODO to add back when it's fixed

bnusunny:
Since Finch team confirmed this is a bug, we are skipping this test and will add it back when it is fixed.

bnusunny:
The fix is to upgrade runc to 1.3.4

vicheey (approved):

Issue #8656, opened 2026-02-15
Intermediate layer caching on EKS


royassis:
Hey

I have a SAM application with multiple AWS lambda functions, that are image based.

Right now we run our CICD pipeline of a Jenkins server that runs on EC2. 
All the intermediate docker layers are caches on the instance itself.

We are moving our Jenkins stack (workers and controller) from running on EC2 to EKS, so the layers won't be cached.
There is a solution for this with docker buildkit: 
`--cache-to` and `--cache-from`.

What is a sutible solution for this when using AWS SAM?

reedham-aws:
In #8661, I've added the framework for using docker buildkit. I believe once this is merged, it should be possible to use the `--cache-to` and `--cache-from` options with BuildKit via the DockerBuildArgs section in the Metadata of the template:
```yaml
Metadata:
  # ...
  DockerBuildArgs:
    CACHE_TO: !Ref CacheLocation
```
I haven't thought too much about using other methods to achieve this, but I think this will work with your existing flow once merged.

seshubaws:
Hi @royassis, https://github.com/aws/aws-sam-cli/pull/8661 will be available to use in the next release, coming later this week or early next week. Thank you for your patience!

royassis:
> Hi @royassis, https://github.com/aws/aws-sam-cli/pull/8661 will be available to use in the next release, coming later this week or early next week. Thank you for your patience! 

I'm very excited. Thank you.

royassis:
Hey @seshubaws 

I tried to use cache-from and cache-to like so:

```text
Metadata:
  # ...
  DockerBuildArgs:
    CACHE_TO: !Ref CacheLocation
```

When running `sam build --use-buildkit` it indeed finishes the build, but it seems that the layers might not be cached to ECR as I expected, I don't see any new images in my cache repo, after running `sam build`.

When investigating with kiro, it mentioned this, so it might be related:
> But looking at the actual code in CLIBuildClient.build_image, all DockerBuildArgs entries are passed as --build-arg:
> 
> ```python
> if buildargs:
>     for k, v in buildargs.items():
>         cmd.extend(["--build-arg", f"{k}={v}"])
> ```
> 
> So CACHE_TO and CACHE_FROM in DockerBuildArgs become --build-arg CACHE_TO=... and --build-arg CACHE_FROM=... — not --cache-to and --cache-from.

seshubaws:
[...]

Issue #3972, opened 2022-06-14
Feature request: ability to enable Docker Buildkit for `sam build`


ghost:
### Describe your idea/feature/enhancement

I wish SAM CLI would enable me to enable Docker Buildkit when running `sam build` for a function with `PackageType: Image`.

This would enable me to use BuildKit features such as cache mounts for faster and more efficient builds.

### Proposal

A very specific approach would be to add a new option along the lines of `--use-docker-buildkit` or something like that.

That said, something more general might be a good idea.

For example, I tried to enable buildkit by running `export DOCKER_BUILDKIT=1 && sam build` but it didn’t work.

IIRC `docker build` supports various other environment variables, so it might be a good idea to have a general purpose mechanism to set environment variables that’d be exposed to the `docker build` process.

One way to do this might be to simply forward/inherit the environment vars from the user’s shell to/into the `docker build` process.

Another way could be to add a CLI option like `--docker-build-env DOCKER_BUILDKIT=1,FOO=bar`

### Additional Details

This _will not_ require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model).

aviflat:
Huh, I just found docker-py#2230 and it looks like bad news… 😞

qingchm:
I know it's going to be an interesting feature to support but since our builds are based on docker API for python, unless docker supports buildkit on python api I don't see a quick solution to this!

aviflat:
Makes sense. Thanks.

Maybe it’d help if maintainers of downstream libs would comment on that issue? Or otherwise somehow convey to the maintainers of `docker-py` that support for Buildkit would be great?

Thanks!

qingchm:
@aviflat Yep I would assume opening an issue against their repository would be the way to go!

aviflat:
Well, since docker/docker-py#2230 already exists, I was suggesting that one of the maintainers of this project post a comment to that issue, voicing support for the feature.

mildaniel:
Since there isn't much we can do on our side, I will close this for now but please re-open when this is actionable for us.

remidebette:
Hi,
Alternatively, for using the docker cli without reimplementing the wheel, the project https://github.com/gabrieldemarmiesse/python-on-whales readily provides a python wrapper on the CLI.

It works fine with buildx

antonio-carapcea-softwire:
Any news about this?

bnusunny:
[...]

Issue #3852, opened 2022-05-04
Enable support for multiple docker contexts


mew1033:
### Describe your idea/feature/enhancement

Docker just added support for multiple build contexts. This could be especially useful in multi-function lambda deployments where common code needs to be used across multiple functions. For example:
```
root_folder
├── function1
│   ├── function1_code.py
│   └── Dockerfile
├── function2
│   ├── function2_code.py
│   └── Dockerfile
├── common_libraries
│   └── common.py
├── samconfig.toml
└── template.yaml
```
In this scenario, it would be nice to have `sam build` effectively run `docker build --build-context common=../common_libraries function1`. Then all the Dockerfiles could be updated to pull the common libraries instead of having to copy `common.py` into each function's folder.

### Proposal

Add support for multiple docker build contexts. Potentially through the DockerBuildArgs field in the template.

Things to consider:
1. Will this require any updates to the [SAM Spec](https://github.com/awslabs/serverless-application-model)

### Additional Details
https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/

jfuss:
@mew1033 Thanks for the feature request and bringing this Docker release up to us. I think this should be simple and adding it through the `DockerBuildArgs` makes the most sense to me, as that is our current mechanism to allow some customization for the Docker integration.

We may need to update Docker's sdk we use with this.

mew1033:
Not sure if this is a problem, but it looks like docker-py doesn't support buildkit yet: https://github.com/docker/docker-py/issues/2230

PR #8661, opened 2026-02-16
feat: add support for using CLI container clients

reedham-aws:
#3972, #8656, #3852

Currently, the docker-py SDK does not support BuildKit: https://github.com/docker/docker-py/issues/2230. This means that in order for `sam build` to use BuildKit, we must use the Docker (or Finch) CLI.

This change adds a `BuildClient` class that acts as a parent to either the `SDKBuildClient` or the `CLIBuildClient`. If a user passes a `sam build` argument of `--use-buildkit`, the `CLIBuildClient` is used with either `docker buildx build` or `finch build ...` (Finch uses buildkit by default).

For existing builds, there shouldn't be any. Using the CLI is entirely opt in, and the SDK build is doing the same thing as before with only a thin layer in between. There could be a difference between the SDK build and the CLI build for different applications (especially for things like cross platform builds, which buildkit is for). There also isn't any change to the detection of which build engine to use, so it shouldn't change where Docker/Finch is used.

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] Write/update integration tests - **To come in follow up PR on this branch**
- [x] `make pr` passes
- [x] Write documentation


tobixlea (approved):

samcli/lib/build/app_builder.py
```diff
@@ -453,8 +458,12 @@ def _build_lambda_image(self, function_name: str, metadata: Dict, architecture:
             build_args["target"] = cast(str, docker_build_target)
 
         try:
-            (build_image, build_logs) = self._docker_client.images.build(**build_args)
-            LOG.debug("%s image is built for %s function", build_image, function_name)
+            if self._build_client:
+                build_logs = self._build_client.build_image(**build_args)  # type: ignore[arg-type]
+                LOG.debug(f"Image build for {function_name} function using build client")
+            else:
+                (build_image, build_logs) = self._docker_client.images.build(**build_args)
```
vicheey: Since the `self._build_client: BuildClient` can be `SDKBuilderClient` or `CLIBuildClient`, shouldn't we just call `self._build_client.build_image(**build_args)` and get rid of the branching?

samcli/lib/build/app_builder.py
```diff
@@ -161,6 +163,8 @@ def __init__(
             Mount source code directory with write permissions when building inside container.
         mount_symlinks: Optional[bool]
             True if symlinks should be mounted in the container.
+        build_client: Optional[BuildClient]
+            Optional build client for building Image functions with buildkit support.
```
vicheey: 1. I thought build_client is either SDKBuilderClient or SDKBuilderClient. 
2. Shouldn't we always needs a builder client? Any reason that this is optional? If we get rid of the optional part, we can get rid of the if condition that check for build_client.

samcli/lib/build/app_builder.py
reedham-aws: ~~Yes, good point. Not sure why I did this 🤔.~~

Actually, I'm realizing the reason is that `_build_client` could be None. However, this is a bit weird considering what you pointed out considering that the `build_image` method should work either way. I think this requires a bit of refactoring in the `BuildContext` because I think right now we're never actually instantiating an `SDKBuildClient`. That should be fixed.

reedham-aws: And now looking at your second comment I'm seeing you were already on to this.

samcli/lib/build/app_builder.py
reedham-aws: I've reworked the way a lot of this code works, so I'm resolving this comment.

tobixlea (approved):

samcli/local/docker/image_build_client.py
```diff
@@ -0,0 +1,234 @@
...
+    @staticmethod
+    def is_available(engine_type: str) -> Tuple[bool, Optional[str]]:
+        if engine_type == "docker":
+            if not shutil.which("docker"):
+                return (False, "Docker CLI not found")
+
+            result = subprocess.run(
+                ["docker", "buildx", "version"],
+                capture_output=True,
+                check=True,
```
bnusunny: If `docker buildx version` fails, it raises `CalledProcessError` before the `returncode != 0` check is ever reached. This should use `check=False` or wrap in a `try/except` for `CalledProcessError`. 

The Finch path has the same issue.

bnusunny (changes requested):

samcli/local/docker/image_build_client.py
```diff
@@ -0,0 +1,234 @@
...
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True)
+
+        if process.stdout:
+            for line in process.stdout:
+                yield {"stream": line}
+
+        process.wait()
+
+        if process.returncode != 0:
+            yield {"error": f"Build failed with exit code {process.returncode}"}
```
bnusunny: This does not raise an exception. The caller function expects `docker.errors.BuildError` on build failure.

bnusunny (changes requested):

samcli/local/docker/image_build_client.py
reedham-aws: I changed this to be `check=False`. I originally had let it fall through to the `if result.returncode != 0`, but then ruff didn't like the lack of explicit `check`. After that I forgot about the original intention 🙃.

samcli/local/docker/image_build_client.py
reedham-aws: Good callout. I changed this to raise the error instead of stream like this. What I did above was capture the stdout and save it in a list called `build_log`, then pass that as the `build_log` for the `docker.errors.BuildError`, which has the following definition:
```python
class BuildError(DockerException):
    def __init__(self, reason, build_log):
        super().__init__(reason)
        self.msg = reason
        self.build_log = build_log

```

I think this should be fine, although I guess if the logs are long enough memory usage could balloon. Let me know if you think that is an issue.

bnusunny (approved):

PR #8675, opened 2026-02-20
Fix AttributeError when port is already in use

### LINKED ISSUE #7244: `'str' object has no attribute 'decode'` when `PortAlreadyInUse` error is raised.
https://github.com/aws/aws-sam-cli/blob/5570367805bc94e43755e31a493b7cb93677dde3/samcli/local/docker/container.py#L392

Remove `.decode()` to resolve this issue.

dcabib:
## Issue
Fixes #7244

## Problem
When SAM CLI attempts to bind to a port that's already in use, users see a cryptic `AttributeError: 'str' object has no attribute 'decode'` instead of a helpful error message about which port is occupied.

This occurs because:
1. In docker-py >= 7.0, `APIError.explanation` is a `str` (not `bytes`)
2. The code calls `.decode()` on it, causing AttributeError
3. Case-sensitive check ("Ports") misses lowercase variations ("ports")

## Solution
- Added `safe_decode_docker_message()` function that handles both `bytes` (docker-py < 7.0) and `str` (docker-py >= 7.0)
- Updated error detection to use case-insensitive check (`.lower()`)
- Added 8 comprehensive unit tests covering all scenarios

## Testing
- ✅ All 6931 existing tests pass
- ✅ Added 8 new tests (100% passing)
- ✅ Coverage maintained at 94.04%
- ✅ Lint and formatting checks pass
- ✅ Tested with `samdev local invoke` on occupied port

## Result
Users now see the actual port error message:
```
Error: ports are not available: exposing port TCP 127.0.0.1:5005 -> 127.0.0.1:0: listen tcp4 127.0.0.1:5005: bind: address already in use
```

Instead of:
```
AttributeError: 'str' object has no attribute 'decode'
```

## Checklist
- [x] Code builds correctly
- [x] New/updated tests pass
- [x] All tests pass (`make test`)
- [x] Coverage maintained at 94%+
- [x] Lint checks pass (`make lint`)
- [x] Formatting is correct (`make black-check`)
- [x] Issue linked in commit and PR description
- [x] Added comprehensive unit tests
- [x] Backwards compatible (no breaking changes)


tests/unit/local/docker/test_container.py
```diff
@@ -765,6 +765,60 @@ def test_docker_raises_api_error(self):
         with self.assertRaises(APIError):
             self.container.start()
 
+    def test_docker_api_error_port_in_use_with_str_explanation(self):
```
vicheey: Can we shorten the tests here with parameterization?

roger-zhangg (dismissed):
LGTM — the fix is correct and minimal. The `safe_decode_docker_message` utility properly handles both `bytes` (docker-py < 7.0) and `str` (docker-py >= 7.0) return types, and the case-insensitive check is a good defensive improvement.

Two minor observations (non-blocking):
1. If `ex.explanation` is ever `None` (the `APIError.__init__` signature allows `explanation=None`), the utility will return `None`. This is unlikely in practice since the `"ports are not available"` check on `str(ex)` implies the error is well-formed, but you could add a `if message is None: return ""` guard for extra safety.
2. Agree with the existing feedback from @vicheey — the three test cases for `test_container.py` could be consolidated with parameterization (e.g., `@parameterized.expand`) to reduce repetition.

Thanks for the contribution!

roger-zhangg (approved):
All review feedback addressed:
- None guard added to `safe_decode_docker_message`
- Tests parameterized per @vicheey's suggestion

Ready to merge.

PR #8682, opened 2026-02-21
register stack for cleanup in test_sync_infra_s3_bucket_option

licjun:
Cleaning API GW resource after test

The `test_sync_infra_s3_bucket_option` test method creates a CloudFormation stack but doesn't register it for cleanup in `tearDown()`. This causes the stack and its resources (including HTTP API Gateways) to be orphaned when the test completes. 

Adds the missing `self.stacks.append({"name": stack_name})` line after the stack name is generated. This registers the stack for immediate cleanup in `tearDown()`, matching the pattern used in other test methods like `test_sync_infra`. Now the stack will be deleted immediately after the test completes, regardless of pass/fail status.

None. This aligns the cleanup behavior with all other sync integration tests. The stack will be deleted faster (immediately in `tearDown()` instead of waiting for the account reset Lambda), reducing the window for security scanners to detect orphaned resources.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,7 +196,7 @@ jobs:
           fi
           rustup toolchain install stable --profile minimal --no-self-update
           rustup default stable
-          pip install cargo-lambda==$CARGO_LAMBDA_VERSION
+          pip install "cargo-lambda==$CARGO_LAMBDA_VERSION"
           echo "$HOME/.local/bin" >> $GITHUB_PATH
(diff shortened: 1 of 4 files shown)


valerena (approved):

reedham-aws:
Once https://github.com/aws/aws-sam-cli/pull/8687 gets merged we need to merge it here because that's what contains the fix for `chardet` version mismatch that caused the merge to develop to fail.

PR #8666, opened 2026-02-19
refactor(test): restructure integration tests with tiered CI, parallel jobs, and Finch cross-platform smoke tests

roger-zhangg:
#### Which issue(s) does this change fix?

Addresses integration test CI performance and reliability issues. No specific issue number.

#### Why is this change necessary?

The integration test workflow was slow (~2+ hours), flaky due to cross-worker Docker interference, and lacked cross-platform validation. The `container_runtime` matrix (docker/finch/no-container) created too many combinations, and tests that required AWS credentials were mixed with local-only tests.

#### How does it address the issue?

**1. Workflow restructuring (21 parallel matrix jobs)**
- Removed `container_runtime` matrix dimension (docker/finch/no-container)
- Split large jobs into smaller parallel ones: build-x86-1/2, build-arm64, build-x86-container-1/2, build-arm64-container-1/2, terraform-build/start-api/invoke-start-lambda, package, deploy, sync-code/watch, local-invoke/start-api/start-lambda, other-and-e2e, cloud-based-tests, tier1-finch
- Local-only jobs release test accounts early after ECR login

**2. Credential-based test separation**
- `@pytest.mark.requires_credential` marker for tests needing AWS credentials
- Build/local jobs use `-m "not requires_credential"` to exclude cloud tests
- `cloud-based-tests` job collects them via `-m requires_credential`

**3. Docker container cleanup isolation**
- Scoped container cleanup to per-test-class snapshots
- Added `xdist_group` markers: `durable`, `docker_images`, `remote_layers`, `docker_watcher`, `lambda_layers`

**4. Tier 1 cross-platform smoke tests (~60 curated tests)**

A curated subset marked with `@pytest.mark.tier1` runs on every OS/container-runtime combination via the `tier1-finch` job. Each test is a dedicated `test_tier1_*` method calling existing logic with one specific parameter set.

| Category | Test | File |
|----------|------|------|
| **Build: Python** | `test_tier1_python_build` | test_build_cmd_python.py |
| **Build: Python (container)** | `test_tier1_python_build_in_container` | test_build_cmd_python.py |
| **Build: Java** | `test_tier1_java_build` | test_build_cmd_java.py |
| **Build: Java (container)** | `test_tier1_java_build_in_container` | test_build_cmd_java.py |
| **Build: Node.js** | `test_tier1_node_build` | test_build_cmd_node.py |
| **Build: Node.js (container)** | `test_tier1_node_build_in_container` | test_build_cmd_node.py |
[...]

PR #8687, opened 2026-02-23
Restore layer test tearDown image cleanup removed in #8666

roger-zhangg:
Re-add the samcli/lambda-* Docker image cleanup in tearDown for TestLayerVersionBase and TestLayerVersionThatDoNotCreateCache.

This was removed in PR #8666 to avoid cross-test interference in parallel runs, but these test classes are already serialized via xdist_group markers. Without the cleanup, stale cached images cause layer version tests to use outdated layers.

Changes:
- Add `_cleanup_samcli_images` static method to `TestLayerVersionBase` that removes all `samcli/lambda-*` images
- Restore `tearDown` in `TestLayerVersionBase` to call the cleanup + clear layer cache
- Restore `tearDown` in `TestLayerVersionThatDoNotCreateCache` to clean up images
- Add comment noting these tests can't run in parallel

### DIFF
--- a/tests/integration/local/invoke/test_integrations_cli.py
+++ b/tests/integration/local/invoke/test_integrations_cli.py
@@ -799,7 +799,7 @@ class TestUsingConfigFiles(InvokeIntegBase):
         return custom_cred
 
 
-def cleanup_samcli_images(docker_client):
+def cleanup_samcli_images():
     """Remove all samcli/lambda-* images.
 
     Docker's images.list(name="samcli/lambda") does exact repository matching
@@ -807,6 +807,7 @@ def cleanup_samcli_images(docker_client):
     and filter by tag prefix instead.
     """
     try:
+        docker_client = get_validated_container_client()
         all_images = docker_client.images.list()
         for image in all_images:
             for tag in image.tags:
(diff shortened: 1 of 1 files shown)


roger-zhangg:
there was dep issue: https://github.com/psf/requests/issues/7222
Now we need to pin chardet version to <6 for compatibility with request lib

tests/integration/local/invoke/test_integrations_cli.py
```diff
@@ -799,16 +799,35 @@ def _create_cred_file(self, profile):
...
+        """
+        try:
+            all_images = docker_client.images.list()
+            for image in all_images:
+                for tag in image.tags:
+                    if tag.startswith("samcli/lambda-"):
+                        docker_client.remove_image_safely(image.id, force=True)
+                        break
+        except Exception:
+            pass
```
valerena: Does it make sense to put this outside of the class, so we can then reuse it in the other class `TestLayerVersionThatDoNotCreateCache` ?

roger-zhangg: Applied change, thanks

PR #8688, opened 2026-02-23
fix(local invoke): ensure later Lambda layers overwrite earlier layers in Docker image

roger-zhangg:
## Problem
When a function has multiple layers with overlapping files (e.g. both layer1 and layer2 provide \`simple_python_module/simple_python.py\`), the second layer should win per AWS Lambda semantics. However, Docker's \`ADD layer2 /opt\` after \`ADD layer1 /opt\` does not reliably overwrite files in all build backends (BuildKit vs legacy builder).

This caused \`test_download_two_layers\` to fail with \`\"Layer1\" != \"Layer2\"\`.

## Fix
Stage each layer in a unique temp directory (\`/tmp/layer0\`, \`/tmp/layer1\`, etc.) and use \`RUN cp -rf /tmp/layerN/. /opt/ && rm -rf /tmp/layerN\` to explicitly overwrite files, matching real Lambda behavior.

Also adds \`cleanup_samcli_images()\` call to \`tearDownClass\` for proper image cleanup between parameterized test classes.

## Changes
- \`samcli/local/docker/lambda_image.py\`: Changed \`_generate_dockerfile\` to stage layers via cp -rf
- \`tests/unit/local/docker/test_lambda_image.py\`: Updated expected Dockerfile output in unit tests
- \`tests/integration/local/invoke/test_integrations_cli.py\`: Added tearDownClass image cleanup"

### DIFF
--- a/tests/integration/testdata/invoke/layers/some-dir/layer-template-parent.yaml
+++ b/tests/integration/testdata/invoke/layers/some-dir/layer-template-parent.yaml
@@ -19,6 +19,14 @@ Parameters:
     Default: arn:aws:lambda:us-west-2:111111111111:layer:non_existent_layer:1
     Type: String
 
+  LayerAArn:
+    Default: arn:aws:lambda:us-west-2:111111111111:layer:layer_a:1
+    Type: String
+
+  LayerBArn:
+    Default: arn:aws:lambda:us-west-2:111111111111:layer:layer_b:1
+    Type: String
+
 Resources:
   SubApp:
     Type: AWS::Serverless::Application
@@ -29,3 +37,5 @@ Resources:
         LayerTwoArn: !Ref LayerTwoArn
         ChangedLayerArn: !Ref ChangedLayerArn
         NonExistentLayerArn: !Ref NonExistentLayerArn
+        LayerAArn: !Ref LayerAArn
+        LayerBArn: !Ref LayerBArn
--- a/tests/integration/testdata/invoke/nested-templates/layer-template-parent.yaml
+++ b/tests/integration/testdata/invoke/nested-templates/layer-template-parent.yaml
@@ -19,6 +19,14 @@ Parameters:
     Default: arn:aws:lambda:us-west-2:111111111111:layer:non_existent_layer:1
     Type: String
 
+  LayerAArn:
+    Default: arn:aws:lambda:us-west-2:111111111111:layer:layer_a:1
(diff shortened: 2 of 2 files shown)


roger-zhangg:
https://github.com/aws/aws-sam-cli/actions/runs/22322273258/job/64583388205

https://github.com/aws/aws-sam-cli/actions/runs/22324752527

tests/unit/local/docker/test_lambda_image.py
```diff
@@ -435,7 +435,7 @@ def test_generate_dockerfile_with_arm64(self, docker_patch):
         docker_client_mock = Mock()
         docker_patch.from_env.return_value = docker_client_mock
 
-        expected_docker_file = "FROM python\nADD aws-lambda-rie-arm64 /var/rapid/\nRUN mv /var/rapid/aws-lambda-rie-arm64 /var/rapid/aws-lambda-rie && chmod +x /var/rapid/aws-lambda-rie\nADD layer1 /opt\n"
+        expected_docker_file = "FROM python\nADD aws-lambda-rie-arm64 /var/rapid/\nRUN mv /var/rapid/aws-lambda-rie-arm64 /var/rapid/aws-lambda-rie && chmod +x /var/rapid/aws-lambda-rie\nADD layer1 /tmp/layer0\nRUN cp -rf /tmp/layer0/. /opt/ && rm -rf /tmp/layer0\n"
```
valerena: Do we have something in the integration tests that checks that 2 layers are built correctly? (is it the test that you mention in the description?).

It's interesting that we don't have a unit test that checks the behavior with 2 or more layers, but if we have an integration test that checks that, then it should be good enough.

valerena (approved):

reedham-aws (approved):

tests/unit/local/docker/test_lambda_image.py
roger-zhangg: This is a good point, it seems we were checking the layer override (this test). But not really 2 layer working together. I can take a look

roger-zhangg:
Added an addtional test to import 2 layers and test function from both
integ test triggered: https://github.com/aws/aws-sam-cli/actions/runs/22330860970

Also for the first layer, still use Add directly for better performance. 
We only need to handle overwrite issue on the second / or later layers.

roger-zhangg:
Seems new integ test is failing, fixing

PR #8689, opened 2026-02-24
test: add integration tests for CLI client

reedham-aws:
Integration test PR for buildkit/container CLI support.

Added a few parameterized tests for some image based tests, as well as just general tests for using the `--use-buildkit` flag. Worth noting is that when using the CLI, there were a lot of resource warnings that the sockets were being left open. I couldn't figure out a fix to this, but I think they're harmless warnings so I ignored them.

Implementation PR: https://github.com/aws/aws-sam-cli/pull/8661.

### DIFF
--- a/tests/setup_finch.sh
+++ b/tests/setup_finch.sh
@@ -15,10 +15,10 @@ sudo apt-get autoremove -y || true
 
 echo "=== Installing Finch ==="
 for i in {1..3}; do
-    if curl -fsSL https://artifact.runfinch.com/deb/GPG_KEY.pub | sudo gpg --dearmor -o /usr/share/keyrings/runfinch-finch-archive-keyring.gpg; then
-        break
-    fi
-    sleep 10
+  if curl -fsSL https://artifact.runfinch.com/deb/GPG_KEY.pub | sudo gpg --dearmor -o /usr/share/keyrings/runfinch-finch-archive-keyring.gpg; then
+    break
+  fi
+  sleep 10
 done
 
 echo 'deb [signed-by=/usr/share/keyrings/runfinch-finch-archive-keyring.gpg arch=amd64] https://artifact.runfinch.com/deb noble main' | sudo tee /etc/apt/sources.list.d/runfinch-finch.list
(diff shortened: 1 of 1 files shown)


roger-zhangg:
prob need to merge from develop for the chardet thing

reedham-aws:
> would suggest to add a ARM64 test, and then mark one test -> x86/arm64 as https://github.com/aws/aws-sam-cli/blob/develop/CONTRIBUTING.md#tier-1-cross-platform-smoke-tests for running that test on finch / windows (to be added)

Done, I added as tier1, although not sure if this would have fallen more into the tier1_extra category. I think the testing of the CLI is important enough, not sure about the parameters.

reedham-aws:
Integ test run: https://github.com/aws/aws-sam-cli/actions/runs/22508505787

tests/integration/buildcmd/build_integ_base.py
```diff
@@ -140,6 +141,9 @@ def get_command_list(
         if debug:
             command_list += ["--debug"]
 
+        if use_buildkit:
+            command_list += ["--use-buildkit"]
```
vicheey, line 144: Will there be change in VS Code extension for this feature? If so, should we have a negate version of this as well: `--no-use-buildkit`. This should be similar to https://github.com/aws/aws-sam-cli/blob/c95e0ac831e1b400657bb2c99a84dce4b929da4a/samcli/commands/_utils/options.py#L852

reedham-aws: `--no-use-buildkit` exists in the implementation, is it required to be in the test options as well?

reedham-aws:
Latest integration test run: https://github.com/aws/aws-sam-cli/actions/runs/22595806682. I removed the ARM64 test because it wasn't working well in x86 environment. While BuildKit should technically support this, that's more testing a BuildKit feature than our implementation. I think QEMU needs to be set up in some way in the environment for it to fully work.

Edit: All related and non-flaky tests are passing

reedham-aws:
Latest test run: https://github.com/aws/aws-sam-cli/actions/runs/22649580848

Before, the `tier1-finch` tests were failing. This is because the test was incorrectly finding that the finch CLI was unavailable. The problem was that in the CI runner, finch required `sudo` to run, but the code did not supply it. The solution was to make it so that sudo was not required for `finch` or `nerdctl`.
```
sudo chmod +s "$NERDCTL_PATH"
sudo chmod +s /usr/bin/finch
```

PR #8690, opened 2026-02-24
Update sync action to overwrite develop in mirror

seshubaws:

--force-with-lease fails because the action does not fetch the mirror every time. We could do this, but since we want to overwrite the dev branch of the mirror repo every time, we can just use --force since no actual updates should be pushed to the dev branch of the mirror repo, only to other branches.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation

### DIFF
--- a/.github/workflows/sync-to-mirror-repo.yml
+++ b/.github/workflows/sync-to-mirror-repo.yml
@@ -23,4 +23,4 @@ jobs:
       - name: Push to mirror repo
         run: |
           git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}.git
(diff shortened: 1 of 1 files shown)


reedham-aws:
There have been merges to develop in the other repo before, would that conflict with this?

seshubaws:
> There have been merges to develop in the other repo before, would that conflict with this?

Do you mean dependabot updates? That should be ok because it should be the same in both repos. There shouldn't be actual feature updates merged to develop though, that should all be done in branches, and then directly merged to public repo through PR. I've updated the release process to state this too

reedham-aws (approved):

PR #8693, opened 2026-02-25
Add binary integration testing support to GitHub Actions workflow

roger-zhangg:
## Changes

- Add a second schedule trigger (4am PST / 12pm UTC daily) that tests against the SAM CLI nightly binary instead of source
- Add `workflow_dispatch` input `install_mode` with options: `code` (default), `binary` (nightly), `latest-release`
- Add `Makefile` targets: `init-nightly`, `init-latest-release`, `setup-pytest`
- Add `scripts/install-sam-cli-binary.sh` for cross-platform (Linux/macOS) SAM CLI binary installation
- Use `gh release download` instead of raw curl for GitHub release assets
- Unset `SAM_CLI_DEV` in binary flows so tests use `sam` instead of `samdev`
- Create `/usr/local/bin/zig` wrapper for SAM CLI's cargo-lambda Rust builds
- Use pytest venv python for setup/reset scripts when available

## Testing

Tested via `workflow_dispatch` with `install_mode=binary` on the `binary-test` branch.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -6,9 +6,9 @@ on:
     # This translates to Tuesday-Saturday 3am UTC
     - cron: '0 3 * * 2-6'
   release:
-    types: [released, prereleased]
-    # released → latest-release binary test
-    # prereleased → nightly binary test
+    types: [published, edited]
+    # published → latest-release binary test (full releases only)
+    # edited + prerelease → nightly binary test (nightly pre-release gets edited with new artifacts)
   workflow_dispatch:  # Allow manual triggering
     # NOTE: This workflow can only be manually triggered from develop or main branches
     # The other branch will not pass the OIDC permission check
@@ -200,10 +200,10 @@ jobs:
         run: |
           echo "CONTAINER_RUNTIME=${{ matrix.test_suite == 'tier1-finch' && 'finch' || 'docker' }}" >> $GITHUB_ENV
           if [[ "${{ inputs.install_mode }}" == "nightly-release" ]] || \
-             [[ "${{ github.event.action }}" == "prereleased" ]]; then
+             [[ "${{ github.event.action }}" == "edited" && "${{ github.event.release.prerelease }}" == "true" ]]; then
             echo "TEST_TYPE=nightly-release" >> $GITHUB_ENV
           elif [[ "${{ inputs.install_mode }}" == "latest-release" ]] || \
-               [[ "${{ github.event.action }}" == "released" ]]; then
+               [[ "${{ github.event.action }}" == "published" && "${{ github.event.release.prerelease }}" != "true" ]]; then
             echo "TEST_TYPE=latest-release" >> $GITHUB_ENV
           elif [[ "${{ github.ref }}" == "refs/heads/master" ]]; then
             echo "TEST_TYPE=master" >> $GITHUB_ENV


roger-zhangg:
https://github.com/roger-zhangg/aws-sam-cli/actions/runs/22418274086

image

Makefile
```diff
@@ -13,6 +13,23 @@ init:
 		SAM_CLI_DEV=1 pip install -e '.[dev]'; \
 	fi
 
+# Set up a pytest venv with test dependencies
+setup-pytest:
+	python3.11 -m venv $(HOME)/pytest
```
valerena, line 18: Why do we create a virtualenv instead of just install all of this in the standard `python`? (which is what happens with `make init`)

Makefile
```diff
@@ -13,6 +13,23 @@ init:
...
+	$(HOME)/pytest/bin/python3 -m pip install -r requirements/pre-dev.txt
+	$(HOME)/pytest/bin/python3 -m pip install -r requirements/dev.txt
+	$(HOME)/pytest/bin/python3 -m pip install -r requirements/base.txt
```
valerena: do we really need to install the `base` requirements too? shouldn't `dev` be enough to get `pytest`, `parameterized`, etc?

tests/install-sam-cli-binary.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# Install SAM CLI binary from a GitHub release, auto-detecting OS and architecture.
+# Usage: ./scripts/install-sam-cli-binary.sh [<tag>]
```
valerena: To me it would make sense to add it inside `tests/` inside of creating this new `scripts/`, but I'm open if you think it's better there.

.github/workflows/integration-tests.yml
```diff
@@ -194,17 +211,35 @@ jobs:
...
           if [[ "${{ matrix.test_suite }}" == "tier1-finch" ]]; then
             export CONTAINER_RUNTIME=finch
           else
             export CONTAINER_RUNTIME=docker
           fi
-          make init
+          # When triggered by the 4am PST schedule (cron '0 12 * * 2-6'),
+          # or manually with install_mode=binary or latest-release,
+          # install SAM CLI binary instead of running make init
+          if [[ "${{ github.event.schedule }}" == "0 12 * * 2-6" ]] || [[ "${{ inputs.install_mode }}" == "binary" ]]; then
```
valerena: to check for `"${{ github.event.schedule }}" == "0 12 * * 2-6" ` seems a little if we ever change the schedule for example. Is there another way to say "it's coming from a schedule"? Would it be enough to see if `github.event.schedule` is just defined at all?

Makefile
roger-zhangg: This is for binary test, so we don't want to pollute the system python env which might affect python build tests.

Makefile
roger-zhangg: Didn't check too much ,but it's what was done in appveyor. I can take another look

tests/install-sam-cli-binary.sh
roger-zhangg: yes prob better to be in tests

.github/workflows/integration-tests.yml
roger-zhangg: let me find if there's a way, because we have 2 schedule now. one for integ test, one for binary test. So just limit to a schedule will apply to both tests.

Makefile
roger-zhangg: Actually, It's we do use a lot sam-cli functionality in our tests, like 
https://github.com/aws/aws-sam-cli/blob/57d4281bf3bcca6525fa80a95ef3bb6db24f4e3e/tests/integration/validate/test_validate_command.py#L26

So I believe it's actually necessary

roger-zhangg: But that being said, I should still switch to UV

.github/workflows/integration-tests.yml
roger-zhangg: changed to be triggered by tag, instead of crontab.

roger-zhangg: logic:
when https://github.com/aws/aws-sam-cli/releases/tag/v1.154.0.dev202602250901 -> find binary in https://github.com/aws/aws-sam-cli/releases/tag/sam-cli-nightly

when https://github.com/aws/aws-sam-cli/releases/tag/v1.154.0 -> find binary in the tag directly

roger-zhangg: updated to listen on release: release-> trigger latest release flow, prerelease-> trigger nightly flow

Makefile
valerena: Well, in the regular tests we still install the `dev` dependencies, which still pollute the main system python and they're not normally available when users install SAM CLI in python.

I'm not sure if all the weirdness of installing it in a weird location, and then doing a symlink, and also checking the python during the GitHub action is worth it, thinking about how much we "affect" the python build tests.

I don't think that we're breaking any test by installing these in the main python, and if we were "making a test wrongly succeed" by installing these, then we would see the actual error when building inside a container probably. I could understand that if that happens, then we would be confused on why one test is working and the other one is not, but I think that's unlikely and this is just making the testing process more confusing.

Having said that, it's not terrible, so we can keep it, but I leave the comment in case someone else has a similar opinion.

tests/install-sam-cli-binary.sh
valerena: Not super important, but you forgot to update this comment with the location of the file

tests/install-sam-cli-binary.sh
```diff
@@ -0,0 +1,64 @@
...
+    # Nightly installs as sam-nightly; rename to sam
+    if [ -f /usr/local/bin/sam-nightly ]; then
+      sudo mv /usr/local/bin/sam-nightly /usr/local/bin/sam
+    fi
+    ;;
+  *.pkg)
+    sudo installer -pkg "$TMPDIR/$ASSET" -target /
+    # Nightly installs as sam-nightly; rename to sam
+    if [ -f /usr/local/bin/sam-nightly ]; then
+      sudo mv /usr/local/bin/sam-nightly /usr/local/bin/sam
```
valerena: Minor, but you could move this `if` to outside of the `case`

roger-zhangg: For the first comment, I think it make more sense to actually install sam cli in a venv for normal integ tests. So we get a clearer env and consistency accross the tests. I will make that change

roger-zhangg: And yes, the goal is to avoid "making a test wrongly succeed"

roger-zhangg: Seems it's not super straightforward to do, the main blocker as "make pr" if we install sam cli in a venv, we need to use that venv to run all the make PR steps and eventually we need to run pytest in the venv as well. It's basically the same as running in system python directly. Unless we create a new workflow "make init-inte-test" but this makes test confusing. Probably better to stick to just `make init` & install in the system.

roger-zhangg:
Updated a minor case: our nightly release will not publish, it keeps editing an old release so need to listen on edit event.

Issue #8488, opened 2025-12-04
Bug: Linux durable functions emulator binary links to host zlib, causing issues for downstream packagers


chenrui333:
### Description:
The AWS SAM CLI wheel for Linux includes a prebuilt native binary:

`samcli/local/rapid/aws-durable-execution-emulator-<arch>`

which caused some downstream linkage issue 

```
Full linkage --test aws-sam-cli output
  Unwanted system libraries:
    /lib/x86_64-linux-gnu/libz.so.1
```

```
/lib/aarch64-linux-gnu/libz.so.1
  libexec/lib/python3.13/site-packages/samcli/local/rapid/aws-durable-execution-emulator-arm64
```

relates to 
- https://github.com/Homebrew/homebrew-core/pull/256861
- https://github.com/Homebrew/homebrew-core/pull/256993
- https://github.com/aws/aws-sam-cli/pull/8484

vicheey:
Do you mind providing additional information about how to reproduce the issue?

valerena:
What's exactly the problem here? This binary is included in SAM CLI, but it's always used inside a Linux container. it's not supposed to be used directly in a Mac environment. In practice the Mac environment just reads it as a "file" that it's copied inside a container, not meant to be directly executed.

Can those files be kept normally in the brew artifacts as-is?

daeho-ro:
It's not about MacOS. On linux host, the included binaries are linked to the system library `libz.so` that is unwanted in homebrew.

chenrui333:
@valerena also is the emulator open source so that we can build from source code?

roger-zhangg:
Hi @daeho-ro @chenrui333 we are planning to remove this durable function emulator from this package in the future release. It should solve this issue automatically.

chenrui333:
That sounds great, would that mean people need to install the emulator separately?

chenrui333:
@roger-zhangg and folks, any update on this thread?

we (homebrew) found some workaround via https://github.com/Homebrew/homebrew-core/pull/262038

alejandro-cedeno-10:
x2, any update this thread?

valerena:
We're looking into the options to do this. The intention is to have a separate pre-built image with this binary instead of including it in SAM CLI. We don't have a specific timeline in mind yet, so we can't say when this be solved. But we will be communicating the progress here.

reedham-aws:
Closing since emulator image is now being pulled from ECR.

PR #8708, opened 2026-02-26
Feat/use durable functions emulator image

SilanHe:
#8488 

This introduces a fix for the Homebrew linkage issue with the Durable Functions emulator. We are also introducing this so that SAM CLI can receive Durable Functions emulator updates uncoupled with SAM CLI release.

We are automatically vending the emulator via an image published to ECR. See https://github.com/aws/aws-durable-execution-sdk-python-testing/pull/196. Since the arm64 emulator is no longer included as an executable directly in the SAM CLI, this also fixed the issue where Homebrew users of SAM CLI could not emulate durable functions locally without doing some workaround.

As a bonus, this reduces the SAM CLI size.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [n/a] Write documentation

### DIFF
--- a/samcli/lib/build/app_builder.py
+++ b/samcli/lib/build/app_builder.py
@@ -430,7 +430,8 @@ class ApplicationBuilder:
         docker_context = cast(str, metadata.get("DockerContext"))
         # Have a default tag if not present.
         tag = metadata.get("DockerTag", "latest")
-        docker_tag = f"{function_name.lower()}:{tag}"
+        image_name = cast(str, metadata.get("DockerImage", function_name))
+        docker_tag = f"{image_name.lower()}:{tag}"
         docker_build_target = metadata.get("DockerBuildTarget", None)
         docker_build_args = metadata.get("DockerBuildArgs", {})
 
--- a/tests/integration/validate/test_validate_command.py
+++ b/tests/integration/validate/test_validate_command.py
@@ -156,9 +156,10 @@ class TestValidate(TestCase):
     @parameterized.expand(
         [
             ("nodejs",),
-            ("python3.7",),
-            ("nodejs16.x",),
+            ("python3.9",),
             ("nodejs18.x",),
+            ("ruby3.2",),
+            ("dotnet6",),
         ]
     )
     def test_lint_deprecated_runtimes(self, runtime):
@@ -205,19 +206,21 @@ class TestValidate(TestCase):
         supported_runtimes = [
             "dotnet10",
             "dotnet8",
+            "java25",
             "java21",
             "java17",
             "java11",
             "java8.al2",
             "nodejs20.x",
             "nodejs22.x",
+            "nodejs24.x",
             "provided.al2",
             "provided.al2023",
             "python3.10",
             "python3.11",
             "python3.12",
             "python3.13",
-            "ruby3.2",
+            "python3.14",
             "ruby3.3",
             "ruby3.4",
         ]
--- a/tests/unit/lib/build_module/test_app_builder.py
+++ b/tests/unit/lib/build_module/test_app_builder.py
@@ -1749,6 +1749,40 @@ class TestApplicationBuilder_build_lambda_image_function(TestCase):
 
         self.assertEqual(result, "name:Tag")
 
+    def test_can_build_image_with_custom_name(self):
+        metadata = {
+            "Dockerfile": "Dockerfile",
+            "DockerContext": "context",
+            "DockerImage": "custom-Name_With-symbols-v1.2.3",
+            "DockerTag": "Tag",
+            "DockerBuildArgs": {"a": "b"},
+        }
+
+        self.container_client_mock.images.build.return_value = (Mock(), [])
+
+        result = self.builder._build_lambda_image("Name", metadata, X86_64)
+
+        self.assertEqual(result, "custom-name_with-symbols-v1.2.3:Tag")
+
+    def test_build_lambda_image_raises_for_non_dict_docker_build_args(self):
+        metadata = {
+            "Dockerfile": "Dockerfile",
+            "DockerContext": "context",
+            "DockerTag": "Tag",
+            "DockerBuildArgs": "not-a-dict",
+        }
+        with self.assertRaises(DockerBuildFailed):
+            self.builder._build_lambda_image("Name", metadata, X86_64)
+
+    def test_build_lambda_image_uses_latest_tag_when_not_specified(self):
+        metadata = {
+            "Dockerfile": "Dockerfile",
+            "DockerContext": "context",
+        }
+        self.container_client_mock.images.build.return_value = (Mock(), [])
+        result = self.builder._build_lambda_image("Name", metadata, X86_64)
+        self.assertEqual(result, "name:latest")
+
     @patch("samcli.lib.build.app_builder.SDKBuildClient")
     def test_lazy_initialization_creates_sdk_build_client_by_default(self, mock_sdk_build_client_class):
         """Test that _image_build_client is lazily initialized with SDKBuildClient when use_buildkit is False"""
@@ -3156,6 +3190,14 @@ class TestApplicationBuilder_build_function_on_container(TestCase):
 
     def tearDown(self):
         EventTracker.clear_trackers()
+
+    def test_must_raise_when_container_manager_is_none(self):
+        self.builder._container_manager = None
+        config = Mock()
+        with self.assertRaises(RuntimeError):
+            self.builder._build_function_on_container(
+                config, "source_dir", "artifacts_dir", "manifest_path", "runtime", X86_64, None
+            )
 
     @patch(
         "samcli.local.docker.container_client_factory.ContainerClientFactory.get_admin_container_preference",


samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -447,4 +413,12 @@ def _wait_for_ready(self, timeout=30):
         except Exception:
             pass
 
-        raise RuntimeError(f"Durable Functions Emulator container failed to become ready within {timeout} seconds")
+        raise RuntimeError(
+            f"Durable Functions Emulator container failed to become ready within {timeout} seconds. "
+            "You may set the DURABLE_EXECUTIONS_EMULATOR_IMAGE_TAG env variable to a specific image "
+            "to ensure that you are using a compatible version. "
+            "Check https://gallery.ecr.aws/o4w4w0v6/aws-durable-execution-emulator. "
```
valerena: minor, but can you use `self._get_emulator_image()` here too?

maybe
```
f"Check https://${self._get_emulator_image().replace('public.ecr', 'gallery.ecr')}. "
```
?

or have another method that uses the same variables? Basically so if we change the location of the image, we have a consistent messaging too

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -290,13 +242,27 @@ def start(self):
...
+                "--port",
+                str(self.port),
+                "--log-level",
+                "DEBUG",
+                "--lambda-endpoint",
+                "http://host.docker.internal:3001",
+                "--store-type",
+                self._get_emulator_store_type(),
+                "--store-path",
+                "/tmp/.durable-executions-local/durable-executions.db",
```
valerena: I don't see this being used before. Is this a new thing needed?

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -290,13 +242,27 @@ def start(self):
...
+                "--host",
+                "0.0.0.0",
+                "--port",
+                str(self.port),
+                "--log-level",
+                "DEBUG",
+                "--lambda-endpoint",
+                "http://host.docker.internal:3001",
+                "--store-type",
+                self._get_emulator_store_type(),
```
valerena, line 269: In line 187 this is already added to the environment variables, so we probably don't need to add it here again.
```
"EXECUTION_STORE_TYPE": self._get_emulator_store_type(),
```

samcli/local/docker/durable_functions_emulator_container.py
valerena: Similar question about the other properties `--log-level`, `--lambda-endpoint`..

SilanHe: yeah we removed a "layer" to the emulator, these were previously being set by default but not transparently to the SAM CLI code

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: ah, EXECUTION_STORE_TYPE is part of the emulator "layer" that was removed. So it's not being used anymore. I can remove it from the environment variables.

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: ah yeah, I can definitely do that

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: conversely, I can also use the "new" environment variables instead:

https://github.com/aws/aws-durable-execution-sdk-python-testing/blob/5e184cbac9ea9968d71bce1a7995635ac99f8ff0/src/aws_durable_execution_sdk_python_testing/cli.py#L67

valerena: Ah. I see. I think both alternatives work, as long as it's clear. I think it makes sense to keep it here and remove it from the env vars.

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -290,13 +242,27 @@ def start(self):
...
         self.container = self._docker_client.containers.create(
-            image=image_tag,
-            command=[f"/usr/local/bin/{emulator_binary_name}", "--host", "0.0.0.0", "--port", str(self.port)],
+            image=self._get_emulator_image(),
+            command=[
+                "dex-local-runner",
+                "start-server",
+                "--host",
+                "0.0.0.0",
+                "--port",
```
roger-zhangg, line 262: I guess not related to this PR, but do we have plan to assign this port dynamically. Currently it would just use 9014 no matter it's taken or not

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -28,8 +26,7 @@ class DurableFunctionsEmulatorContainer:
     """
 
     _RAPID_SOURCE_PATH = Path(__file__).parent.joinpath("..", "rapid").resolve()
-    _EMULATOR_IMAGE = "public.ecr.aws/ubuntu/ubuntu:24.04"
-    _EMULATOR_IMAGE_PREFIX = "samcli/durable-execution-emulator"
+    _EMULATOR_IMAGE = "public.ecr.aws/o4w4w0v6/aws-durable-execution-emulator:"
```
roger-zhangg: This is not the final URL right

SilanHe: no

bchampp: nit (since it needs a tag - and concatenate the `:` when we combine them): 
```suggestion
    _EMULATOR_IMAGE_PREFIX = "public.ecr.aws/o4w4w0v6/aws-durable-execution-emulator"
```

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -132,6 +134,14 @@ def _get_emulator_port(self):
         """
         return self._get_port(self.ENV_EXTERNAL_EMULATOR_PORT, self.ENV_EMULATOR_PORT, self.EMULATOR_PORT)
 
+    def _get_emulator_image_tag(self):
+        """Get the emulator image tag from environment variable or use default."""
+        return os.environ.get(self.ENV_EMULATOR_IMAGE_TAG, "latest")
+
+    def _get_emulator_image(self):
+        """Get the full emulator image name with tag."""
+        return self._EMULATOR_IMAGE + self._get_emulator_image_tag()
```
bchampp:
```suggestion
        return f'{self._EMULATOR_IMAGE}:{self._get_emulator_image_tag()}'
```

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: oh really, I thought the code already dynamically assigned the port according to the env variable.

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -74,6 +71,11 @@ class DurableFunctionsEmulatorContainer:
     """
     ENV_EMULATOR_PORT = "DURABLE_EXECUTIONS_EMULATOR_PORT"
 
+    """
+    Allow pinning to a specific emulator image tag/version
+    """
+    ENV_EMULATOR_IMAGE_TAG = "DURABLE_EXECUTIONS_EMULATOR_IMAGE_TAG"
```
yaythomas, line 81: `sam local invoke` has a --invoke-image parameter, where customers can pass the location of a specific image to be used as the execution base image instead of the default Lambda base image. 

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html#ref-sam-cli-local-invoke-options-invoke-image

could explore using this established pattern rather than adding  ENV_EMULATOR_IMAGE_TAG?

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: ahh yeah this would be better, thanks

samcli/local/docker/durable_functions_emulator_container.py
yaythomas:
```
"HOST": "0.0.0.0",                           # ❌ UNUSED - hardcoded, passed via --host
"PORT": str(self.port),                      # ❌ UNUSED - passed via --port (command wins)
"LOG_LEVEL": "DEBUG",                        # ❌ UNUSED - hardcoded, passed via --log-level
"EXECUTION_STORE_TYPE": self._get_emulator_store_type(),  # ❌ UNUSED - passed via --store-type (command wins)
"EXECUTION_TIME_SCALE": self._get_emulator_time_scale(),  # ❌ UNUSED - not even a CLI arg!
```

so only

self.port - Can be overridden by user
self._get_emulator_store_type() - Can be overridden by user

so maybe:

```
command=[
    "dex-local-runner", "start-server",
    "--host", "0.0.0.0",
    "--port", str(self.port),
    "--log-level", "DEBUG",
    "--lambda-endpoint", "http://host.docker.internal:3001",
    "--store-type", self._get_emulator_store_type(),
    "--store-path", "/tmp/.durable-executions-local/durable-executions.db",
],
environment={
    "AWS_ACCESS_KEY_ID": "foo",
    "AWS_SECRET_ACCESS_KEY": "bar",
    "AWS_DEFAULT_REGION": "us-east-1",
}
```

samcli/local/docker/durable_functions_emulator_container.py
roger-zhangg: humm, which part set this envvar, I might have missed that

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: I think that is a good idea however due to time constraints I won't have any more capacity to work on this for a while so I'd be happy to have this as a follow up item. I'm guessing I'll have to confer with SAM CLI team for the name of the parameter, updating the docs, etc.

I don't think having it as an environment variable makes it substantially harder to use in the meantime.

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: https://github.com/SilanHe/aws-sam-cli/blob/2a2469c6c113e1ab7f0f93e0cb26a040782828a2/samcli/local/docker/durable_functions_emulator_container.py#L256

https://github.com/SilanHe/aws-sam-cli/blob/2a2469c6c113e1ab7f0f93e0cb26a040782828a2/samcli/local/docker/durable_functions_emulator_container.py#L181

https://github.com/SilanHe/aws-sam-cli/blob/2a2469c6c113e1ab7f0f93e0cb26a040782828a2/samcli/local/docker/durable_functions_emulator_container.py#L85

https://github.com/SilanHe/aws-sam-cli/blob/2a2469c6c113e1ab7f0f93e0cb26a040782828a2/samcli/local/docker/durable_functions_emulator_container.py#L135

I guess technically, the code doesn't dynamically detect if 9014 is currently used since it's the default but we have the ability to set it to something else.

SilanHe:
I ran integration tests locally 

```
SAM_CLI_DEV=1 pytest tests/integration/local/invoke/test_invoke_durable.py::TestInvokeDurable -vv -s && SAM_CLI_DEV=1 pytest tests/integration/local/start_lambda/test_start_lambda_durable.py::TestStartLambdaDurable -vv -s && SAM_CLI_DEV=1 pytest tests/integration/local/start_api/test_start_api_durable.py::TestStartApiDurable -vv -s
```

and they all passed

samcli/local/docker/durable_functions_emulator_container.py
roger-zhangg: Yeah that's what I mean, probably we should consider the dynamic assign. Currently in the tests if I run 2 DAR test in parallel it would fail for port in use.

samcli/local/docker/durable_functions_emulator_container.py
bchampp: Personally I don't think this should be exposed as a new parameter on the CLI. Our expectation is for the latest image to always be "stable", so it should only be overridden in unexpected cases. The way I see this, having this be overridable at all is like a backdoor, and we shouldn't expect customers to need to use this.

Do you agree?

samcli/local/docker/durable_functions_emulator_container.py
bchampp: This port is static intentionally because of the order in which the containers are created. 

When we create an instance of the `DurableLambdaContainer` class, we have to tell the lambda container what port the emulator is listening on (by setting the `AWS_ENDPOINT_URL_LAMBDA` in the container environment to the [emulator port](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/durable_lambda_container.py#L54) first), then [we start (or attach)](https://github.com/SilanHe/aws-sam-cli/blob/develop/samcli/local/docker/durable_lambda_container.py#L87) the emulator. Since the lambda container port itself is set dynamically, we pass it through to the emulator as part of the execution payload and the emulator keeps track of it (this is also how we support concurrent executions).

However, if the emulator port was set dynamically, we'd have to update the lambda runtime container environment _after_ the lambda container is created, which isn't possible.

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -290,13 +238,27 @@ def start(self):
...
+                "dex-local-runner",
+                "start-server",
+                "--host",
+                "0.0.0.0",
+                "--port",
+                str(self.port),
+                "--log-level",
+                "DEBUG",
+                "--lambda-endpoint",
+                "http://host.docker.internal:3001",
```
bchampp, line 267: This needs to be removed. We pass in the lambda endpoint dynamically as part of the payload to starting the execution. I think the emulator is already be falling back on 3001 (this is the `sam local start-lambda` port), but having this here is misleading since it's not actually used for the interactions between the emulator and lambda container.

samcli/local/docker/durable_functions_emulator_container.py
bchampp: Note that this is the path within the container, not the path on the customers machine (but we create a docker "volume" to make it persistent). Maybe you could leave a comment explaining that?

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -447,4 +409,12 @@ def _wait_for_ready(self, timeout=30):
...
 
-        raise RuntimeError(f"Durable Functions Emulator container failed to become ready within {timeout} seconds")
+        raise RuntimeError(
+            f"Durable Functions Emulator container failed to become ready within {timeout} seconds. "
+            "You may set the DURABLE_EXECUTIONS_EMULATOR_IMAGE_TAG env variable to a specific image "
+            "to ensure that you are using a compatible version. "
+            f"Check https://${self._get_emulator_image().replace('public.ecr', 'gallery.ecr')}. "
+            "and https://github.com/aws/aws-durable-execution-sdk-python-testing/releases "
+            "for valid image tags. If the problems persist, you can try updating the SAM CLI version "
+            " in case of incompatibility."
```
bchampp: If the emulator itself doesn't become ready, it likely indicates a problem with the emulator server. We have an environment variable that will capture the emulator logs and persist them: https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/docker/durable_functions_emulator_container.py#L63

I wonder if we should add instructions on how to use this here?

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: I think we need to keep this since we removed the layer where the emulator falls back on http://host.docker.internal:3001.

`http://127.0.0.1:3001` is the default now: https://github.com/aws/aws-durable-execution-sdk-python-testing/blob/0969c71cdaf11000f993bc8a17139c2df3896ce5/src/aws_durable_execution_sdk_python_testing/cli.py#L50

I think even though these are the same ports, these are different.

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: I'll add some.

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -301,13 +238,27 @@ def start(self):
...
+            image=self._get_emulator_image(),
+            command=[
+                "dex-local-runner",
+                "start-server",
+                "--host",
+                "0.0.0.0",
+                "--port",
+                str(self.port),
+                "--log-level",
+                "DEBUG",
```
roger-zhangg, line 265: do we always need debug? Can we only set debug with user run `sam local invoke --debug`

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -301,13 +238,27 @@ def start(self):
...
 
-        # Build image with emulator binary
-        image_tag = self._build_emulator_image()
+        self._pull_image_if_needed()
 
         LOG.debug(f"Creating container with name={self._container_name}, port={self.port}")
         self.container = self._docker_client.containers.create(
-            image=image_tag,
-            command=[f"/usr/local/bin/{emulator_binary_name}", "--host", "0.0.0.0", "--port", str(self.port)],
+            image=self._get_emulator_image(),
```
roger-zhangg, line 256: there's a `--skip-pull-image` flag for local invoke, do you plan to honor that in `_get_emulator_image()` ?

roger-zhangg: With this flag,  when a local image exist, even if it's not the newest one, we will skip the image pull

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -301,13 +238,27 @@ def start(self):
...
+                "--lambda-endpoint",
+                "http://host.docker.internal:3001",
+                "--store-type",
+                self._get_emulator_store_type(),
+                "--store-path",
+                "/tmp/.durable-executions-local/durable-executions.db",  # this is the path within the container
+            ],
             name=self._container_name,
             ports={f"{self.port}/tcp": self.port},
             volumes=volumes,
```
roger-zhangg, line 275: we do have a `--docker-volume-basedir` option in local invoke, does this volume follow that flag?

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -7,7 +7,6 @@
 import time
 from http import HTTPStatus
 from pathlib import Path
-from tempfile import NamedTemporaryFile
```
roger-zhangg, line 10: That's awesome!

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: no, I don't think the volumes mentioned are the same.

`--docker-volume-basedir`: "The location of the base directory where the AWS SAM file exists. If Docker is running on a remote machine, you must mount the path where the AWS SAM file exists on the Docker machine, and modify this value to match the remote machine."

The `volumes` mentioned here refers to a [line](https://github.com/SilanHe/aws-sam-cli/blob/6efae8863aa0f7caa7ff0f6d4089a431428315d9/samcli/local/docker/durable_functions_emulator_container.py#L237) above:

```
volumes = {
            to_posix_path(emulator_data_dir): {"bind": "/tmp/.durable-executions-local", "mode": "rw"},
        }
```

I believe this simply refers to where the durable executions emulator is storing data within the container.

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: That's a good point, I think when we discussed requirements for this work this was missed. The code in this file currently doesn't take that flag into account, I'll have to see if the helper functions to `_pull_image_if_needed` do use this flag.

SilanHe: I was thinking about this again, and I think the `--skip-pull-image` flag refers to a different image than the emulator image. It's actually talking about the lambda runtime image, right? The emulator image is a completely different concept, the durable functions that the customers are trying to invoke aren't directly running on the emulator image but rather the lambda runtime image's dar sdk/the customer's dar sdk would be communicating with the emulator container.

So I think staying with the current design is correct. as it is similar what Brent originally had.

roger-zhangg: Thanks for the explanation, but I think the intention for this flag is to save time to not pull newer images when set. So customer will be confused when this is set but yet we are still pulling images. So I would suggest if this is not a huge effort, let's honor this flag

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -198,87 +199,25 @@ def _get_emulator_binary_name(self):
...
         except docker.errors.ImageNotFound:
-            LOG.debug(f"Pulling emulator image {self._EMULATOR_IMAGE}...")
+            LOG.debug(f"Pulling emulator image {self._get_emulator_image()}...")
 
         try:
-            self._docker_client.images.pull(self._EMULATOR_IMAGE)
-            LOG.info(f"Successfully pulled image {self._EMULATOR_IMAGE}")
+            self._docker_client.images.pull(self._get_emulator_image())
+            LOG.info(f"Successfully pulled image {self._get_emulator_image()}")
         except Exception as e:
```
roger-zhangg, line 224: One side note, If a user has Dar image locally but doesn't have public network access (or throttled by ECR, or any other network issue), the code here will raise for `Failed to pull emulator image` . although it's a niche case, we should still support it as we can just use the existing Dar image even the pull newer image operation failed

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: this is the original setting from before, I don't think the emulator logs get published to the console/terminal. I think these are always in Debug as they're stored separately in case the user wants to debug the emulator

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: Sure I can do that

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: Just tested this as well, I didn't see any logs from the emulator in the console/terminal when running `samdev local invoke HelloWorldFunction --event events/event.json`

samcli/local/docker/durable_functions_emulator_container.py
SilanHe: just supported this in newest commit

PR #8721, opened 2026-03-02
feat: support mount-symlinks option in extract_tarfile

bnusunny:
## Summary

Pass the `mount_symlinks` option through to `extract_tarfile` so that symlinks are preserved during tarball extraction when `--mount-symlinks` is enabled.

## Changes

- Added `mount_symlinks` parameter to `extract_tarfile()` (defaults to `False`, no behavior change without the flag)
- Updated `Container.copy()` to forward `self._mount_symlinks` to `extract_tarfile()`
- Added unit tests for the new parameter

## Testing

- All existing unit tests pass
- Added test for symlink preservation when `mount_symlinks=True`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python:
-          - "3.10.12"
+          - "3.10"
           - "3.11"
     steps:
     # This allows temp file creation on drive D, which won't trigger windows defender scan and leads to faster IO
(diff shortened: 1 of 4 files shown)


samcli/lib/utils/tar.py
```diff
@@ -130,13 +131,30 @@ def extract_tarfile(
 
     unpack_dir Union[str, os.PathLike]
         The directory where the tarfile members will be extracted.
+
+    mount_symlinks bool
```
roger-zhangg, line 135: we need to update cli help document as well right?

valerena: This parameter already exists for sam build. 
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-build.html#ref-sam-cli-build-options-mount-symlinks

roger-zhangg:
waiting for :https://github.com/aws/aws-sam-cli/pull/8723 to be merged to update python3.9 version

.github/workflows/build.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python:
-          - "3.10"
+          - "3.10.12"
```
valerena: 3.10.11 is the last version available 😕https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json 

(probably because 3.10 is old now and newer versions [don't come with official installers](https://www.python.org/downloads/release/python-31020/#:~:text=binary%20installers%20are%20no%20longer%20provided%20for%20it.))

roger-zhangg (dismissed):
LGTM

roger-zhangg (approved):

Issue #1355, opened 2019-08-21
feat(init): load .env file at startup of sam local


rynop:
### Describe your idea/feature/enhancement

It would be great if `sam local` loaded env vars from a `.env` file instead of a non-standard `--env-vars` JSON file.

### Proposal

Create an option `--use-dotenv` that will tell the docker container to load env vars from a `.env` file prior to launching the lambda.  Vars defined in `.env` will apply to ALL lambda functions defined in the sam `template.yml`

This will eliminate the need to dual maintenance a `.env` and a `sam-env.json` file.   A `.env` file is needed for local dev that does not need to run through a lambda simulator (for more rapid development iterations).

`--env-vars` is still useful for variables that are specific to an individual lambda.

sanathkr:
Thanks for the feature request. 

The `.env` files look like a docker-compose thing, but SAM CLI doesn't use docker-compose. How do you envision this `--use-dotenv` flag to work? It would be good if you can describe with a more detailed example

rynop:
Thanks. dotenv (aka `.env`) is not a Docker thing.  It is a pretty common practice across most popular languages: [node](https://github.com/motdotla/dotenv), [go](https://github.com/joho/godotenv), [python](https://pypi.org/project/python-dotenv/) etc.

Flow of how it would work (I propose a shorter `--dotenv` CLI option):
1.  Create a `.env` file (format is standardized, @see any link above)
1.  Run `sam local invoke|start-api|start-lambda --dotenv <path to .env>`
1.  Upon launch of sam docker container, because `--dotenv` was specified, sam local runtime loads the contents of the `.env` file specified into the execution environment. I know I've seen output from the sam local container at startup, state that it saw a `.env` and to install some python module to read it - so I'm guessing sam local is using python under the covers. If so, just leverage https://pypi.org/project/python-dotenv/

Code running inside sam local container, can now access any of the variables defined in the `.env` file as they are now in the execution environment - for ex just as the `PATH` env var is [avail to any lambda](https://docs.aws.amazon.com/lambda/latest/dg/lambda-environment-variables.html) regardless of language/runtime.

bitcod3r:
Hello. Please, do you have any plan or temptative date to implement this new feature? Thanks in advance 🙂

schematis:
This is something I'd like to see as well. I'm coming from the serverless framework and they've implemented this so that you can do `${env:ENV_VAR}` in the template and it will pull from either dotenv or an environment variable.

[...]

PR #8746, opened 2026-03-06
feat: add .env file format support for --env-vars option

licjun:
#1355 

The --env-vars option now auto-detects the file format. When a file is provided:
1. It first attempts to parse the file as JSON (existing behavior, backward compatible)
2. If JSON parsing fails, it falls back to .env format using the python-dotenv library
3. Variables from .env files are treated as global parameters (wrapped in {"Parameters": {...}} structure)
4. If the file is neither valid JSON nor valid .env (no key-value pairs found), an error is raised

Example usage:
bash
# JSON still works as before
sam local invoke MyFunction --env-vars env.json

# .env files now also work
sam local invoke MyFunction --env-vars .env

Note: .env files only support global variables (applied to all functions). For function-specific variables, continue using JSON format.

- Adds python-dotenv~=1.0.0 as a new dependency
- Files that previously failed as invalid JSON may now succeed if they are valid .env format. This is intentional and more permissive behavior.
- No breaking changes — existing JSON files work exactly as before

#### Mandatory Checklist

- [x] Write/update unit tests
- [ ] make pr passes
- [ ] make update-reproducible-reqs if dependencies were changed
- [ ] Write documentation

### DIFF
--- a/tests/integration/local/invoke/test_integrations_cli.py
+++ b/tests/integration/local/invoke/test_integrations_cli.py
@@ -268,24 +268,6 @@ class TestSamPythonHelloWorldIntegration(IntegrationCliIntegBase):
         self.assertEqual(process_stdout.decode("utf-8"), '"GlobalVar"')
 
     @pytest.mark.flaky(reruns=3)
-    def test_invoke_with_dotenv_file(self):
-        command_list = InvokeIntegBase.get_command_list(
-            "EchoGlobalCustomEnvVarFunction",
-            template_path=self.template_path,
-            event_path=self.event_path,
-            env_var_path=self.dotenv_path,
-        )
-
-        process = Popen(command_list, stdout=PIPE)
-        try:
-            stdout, _ = process.communicate(timeout=TIMEOUT)
-        except TimeoutExpired:
-            process.kill()
-            raise
-        process_stdout = stdout.strip()
-        self.assertEqual(process_stdout.decode("utf-8"), '"GlobalVar"')
-
-    @pytest.mark.flaky(reruns=3)
     def test_invoke_with_invoke_image_provided(self):
         command_list = InvokeIntegBase.get_command_list(
             "HelloWorldServerlessFunction",
--- a/tests/unit/commands/local/cli_common/test_invoke_context.py
+++ b/tests/unit/commands/local/cli_common/test_invoke_context.py
@@ -1303,38 +1303,6 @@ class TestInvokeContext_get_env_vars_value(TestCase):
         finally:
             os.unlink(filename)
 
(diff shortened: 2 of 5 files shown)


tests/integration/local/invoke/test_integrations_cli.py
reedham-aws, line 1: I'm not sure I personally think this needs an integration test. I'd be satisfied with unit tests confirming the behavior is the same across file formats. Open to what others think, though. This is not a blocking comment.

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1268,20 +1269,105 @@ def test_must_read_file_and_parse_as_json(self):
...
+
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".env", delete=False) as f:
+            f.write(file_data)
+            filename = f.name
+
+        try:
+            result = InvokeContext._get_env_vars_value(filename)
+            self.assertEqual(expected, result)
+        finally:
+            os.unlink(filename)
```
reedham-aws: Seems like these two probably test the library more than they test our code's functionality.

tests/unit/commands/local/cli_common/test_invoke_context.py
```diff
@@ -1268,20 +1269,105 @@ def test_must_read_file_and_parse_as_json(self):
...
+        try:
+            with self.assertRaises(InvalidEnvironmentVariablesFileException) as ex_ctx:
+                InvokeContext._get_env_vars_value(filename)
+
+            msg = str(ex_ctx.exception)
+            self.assertIn("not in valid JSON or .env format", msg)
+        finally:
+            os.unlink(filename)
+
+    def test_must_parse_dotenv_with_special_characters(self):
```
reedham-aws: Same with this one in terms of testing library functionality.

tests/integration/local/invoke/test_integrations_cli.py
licjun: Makes sense. The integration test exercises the same code path as the global JSON env vars test — the only difference is the file parsing, which is already covered by unit tests.

tests/unit/commands/local/cli_common/test_invoke_context.py
licjun: Fair point.

reedham-aws (approved):

tests/integration/local/invoke/test_integrations_cli.py
reedham-aws: That being said, I could see the value in it. If another reviewer feels strongly, I'd approve either way.

vicheey (approved):

PR #8783, opened 2026-03-11
Fix nightly integration test failures on Windows

roger-zhangg:
## Changes

- **Fix Makefile setup-pytest**: Replace `--only-deps` with compatible install+uninstall approach; split Windows/Linux logic using RUNNER_OS; move SCRIPT_PY/GITHUB_PATH export into Makefile
- **revert watchdog v6 bump**: watchdog v6 introduced few breaking change that breaks `sam sync --watch`, reverting
- **Fix MSYS2 credential mangling**: Add `MSYS2_ARG_CONV_EXCL` and `MSYS_NO_PATHCONV` to prevent Ruby's MSYS2 bash from path-translating AWS credentials containing `/` and `+`

## Testing

Triggered via workflow_dispatch on nightly-integ-fix branch will test directly against this branch, instead of checking out nightly-build branch first. This allows us to debug the nightly binary integ test.


roger-zhangg:
integ test: https://github.com/aws/aws-sam-cli/actions/runs/22970788786
binary test: https://github.com/aws/aws-sam-cli/actions/runs/22970796521

vicheey (approved):

Vandita2020 (approved):

PR #8807, opened 2026-03-16
Fix PermissionError on Windows when building durable functions emulator image

roger-zhangg:
## Problem

Durable function integration tests (`test_tier1_callback`, `test_tier1_execution`, `test_tier1_durable_invoke`) fail on Windows (WSL2) runners with three distinct issues:

1. `PermissionError: [Errno 13] Permission denied: 'D:\\a\\_temp\\...Dockerfile'`
2. `invalid volume specification: 'D:\a\aws-sam-cli\...'`
3. `'charmap' codec can't encode character '\u2705'` and `\r\n` vs `\n` assertion mismatches
All tests pass on Linux.

Failed integ test: https://github.com/aws/aws-sam-cli/actions/runs/23079727144/job/67046559248

### 1. NamedTemporaryFile file locking (PermissionError)
In `_build_emulator_image()`, `NamedTemporaryFile` with default `delete=True` holds an exclusive lock on the file while the `with` block is active. When `tarfile.add()` tries to read the same file, Windows enforces mandatory file locking and raises `PermissionError`. Linux doesn't enforce mandatory locks, so it works there.

### 2. Windows path in Docker volume mount
`_get_emulator_data_dir()` returns a Windows-style path (`D:\a\aws-sam-cli\...`), which is passed directly as a Docker volume mount key. Docker inside WSL2 cannot parse Windows paths. SAM CLI already has `to_posix_path()` for this exact conversion (`D:\a\...` → `/d/a/...`), but it wasn't being used here.

### 3. Emoji encoding + CRLF line endings
CLI output uses Unicode emoji characters (✅❌💓🛑⚠️🔄) that Windows' default `charmap` codec cannot encode, causing `click.echo()` to crash when running the test in github windows runner. Additionally, Windows outputs `\r\n` line endings in stderr, but tests compare against `\n`.

### NamedTemporaryFile
Use `delete=False`, close the file before `tarfile.add()` reads it, and manually clean up with `os.unlink()` in a `finally` block.

### Volume mount path
Apply `to_posix_path()` to the emulator data directory before using it as a Docker volume key. This is a no-op on Linux/macOS (`os.name != 'nt'`), and on Windows it produces the `/d/...` format that both Docker Desktop and WSL2 Docker understand.

### Emoji + CRLF
Set utf-8 as default encoding when running the integration tests.. Add `\r\n` → `\n` normalization in integration test stderr comparisons.

## Testing

- All 3 failing Windows tier1 tests should now pass
- No behavior change on Linux/macOS (to_posix_path is a no-op, emojis were decorative)
- Integ test for this commit https://github.com/aws/aws-sam-cli/actions/runs/23170139649


samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -226,10 +231,14 @@ def _build_emulator_image(self):
         # Generate Dockerfile content
         dockerfile_content = self._generate_emulator_dockerfile(emulator_binary_name)
 
-        # Write Dockerfile to temp location and build image
-        with NamedTemporaryFile(mode="w", suffix="_Dockerfile") as dockerfile:
+        # Write Dockerfile to temp location and build image.
+        # Use delete=False because on Windows, NamedTemporaryFile keeps the file
+        # locked while open, preventing tarfile.add() from reading it.
+        dockerfile = NamedTemporaryFile(mode="w", suffix="_Dockerfile", delete=False)
+        try:
```
valerena, line 238: I don't think you need to change the `with` part that was here before. I think just adding the `dockerfile.close()` should be enough. The `with` basically does the same as a try/finally.

bchampp (approved):

valerena (approved):

samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -287,7 +298,7 @@ def start(self):
         os.makedirs(emulator_data_dir, exist_ok=True)
 
         volumes = {
-            emulator_data_dir: {"bind": "/tmp/.durable-executions-local", "mode": "rw"},
+            to_posix_path(emulator_data_dir): {"bind": "/tmp/.durable-executions-local", "mode": "rw"},
```
valerena, line 301: Not an issue, but something I found and I leave it here in case it helps someone else. 

This `to_posix_path` is done inside our `Container` class https://github.com/aws/aws-sam-cli/blob/7c807bf7e148ee2b4c554413f3f32f156bc1a867/samcli/local/docker/container.py#L244-L245 so it gets applied to any class that extends from it or it uses it.. But this method is using `self._docker_client.containers.create` directly, without passing through the `Container` implementation, so it needs to do this again.

valerena (approved):

samcli/local/docker/durable_functions_emulator_container.py
roger-zhangg: we discussed offline, the `delete=False` is a suggested way to avoid permission issue on windows. So as we are closing the file manually, it's more straightforward to remove the with and use dockerfile var directly

tests/integration/local/callback/test_callback.py
```diff
@@ -41,31 +44,54 @@ def _do_callback_test(self, action, operation_name, callback_type, error_suffix)
...
+            # Also capture process state
+            LOG.error("Process poll() = %s", process.poll())
+        self.assertIsNotNone(callback_id, f"Failed to get callback ID from output. Lines: {output_lines}")
 
         # Send first callback to complete the execution
         succeed_command = self.get_callback_command_list("succeed", callback_id, result="test result")
+        LOG.info("Sending first callback: %s", " ".join(succeed_command))
         result = run_command(succeed_command)
-        self.assertEqual(result.process.returncode, 0)
+        stdout_str = result.stdout.decode("utf-8") if isinstance(result.stdout, bytes) else result.stdout
```
vicheey, line 66: Nit: we should have a util function for this check and conversion. Not a blocker but for clarity.

vicheey (approved):

PR #29, opened 2017-08-13
Environment variable override file should use function logical ID for…

PaulMaddox:
Environment variable override file should use the function logical ID as the map key.

Fixes #28

PR #8832, opened 2026-03-19
fix: preserve original exception context in retry decorator

dreamcreated:
## Problem

The `retry` decorator in `samcli/lib/utils/retry.py` swallows the original exception when all retries are exhausted. This makes it impossible to diagnose the root cause of failures because the original traceback is lost.

**Before this fix:**
```python
while remaining_attempts >= 1:
    try:
        return func(*args, **kwargs)
    except exc:
        ...
raise exc_raise(exc_raise_msg)  # original exception context lost
```

## Fix

Capture the last exception and chain it using `raise ... from last_exception`:

```python
last_exception = None
while remaining_attempts >= 1:
    try:
        return func(*args, **kwargs)
    except exc as e:
        last_exception = e
        ...
raise exc_raise(exc_raise_msg) from last_exception
```

This preserves the original traceback via Python exception chaining (`__cause__`), making it much easier to diagnose the root cause of failures.

## Testing

The change is backwards-compatible — the raised exception type and message remain the same, only the `__cause__` is now populated.

## Related

- Issue tracking this bug: https://github.com/bnusunny/aws-sam-cli/issues/29


seshubaws:
@dreamcreated Thank you for raising this! You said "Fixes https://github.com/aws/aws-sam-cli/pull/29" but I think that may be a typo, what issue does this fix? Just so we can link it properly

dreamcreated:
Hi @seshubaws, thank you for pointing that out! Yes, that was a typo — the original issue report is in the fork at https://github.com/bnusunny/aws-sam-cli/issues/29 (a good-first-issue tracking the same problem). The upstream `aws/aws-sam-cli` does not have a separate tracking issue for this, but the fix addresses the same bug described there.

The core problem: the `retry` decorator loses the original exception traceback by raising `exc_raise(exc_raise_msg)` without chaining it, which makes debugging failures very difficult. The one-line fix `raise exc_raise(exc_raise_msg) from last_exception` preserves the full traceback via Python exception chaining (`__cause__`).

Happy to open a dedicated tracking issue on `aws/aws-sam-cli` if that helps your process!

vicheey:
Thanks for this fix @dreamcreated, 

I traced the git history — the original `retry.py` was written in 2020 by @awood45 for Lambda Extensions support, and the missing exception chaining was a miss.

Code looks good.

bnusunny:
@dreamcreated please also add a unit test in the PR.

PR #8892, opened 2026-04-05
chore(integration): bump cryptography to 46.0.6

EhteshamSid:
A few dependencies in the requirements file have CVEs fixed in newer versions:

- `cryptography` 46.0.5 -> 46.0.6 (CVE-2026-34073)
- `requests` 2.32.5 -> 2.33.0 (CVE-2026-25645)

Bumped each one to the minimum safe version.

There's a mutable default argument in `samcli/lib/sync/infra_sync_executor.py`. The default value is shared across all calls that don't pass that argument, so mutations in one call silently affect the next. Changed the default to `None` with an `if arg is None` guard inside the function.

Changed files: `samcli/lib/sync/infra_sync_executor.py`, `samcli/local/lambdafn/remote_files.py`, `tests/integration/durable_integ_base.py`, `tests/integration/local/start_api/test_start_api.py`, `tests/integration/logs/test_logs_command.py` and 8 more

### DIFF
--- a/.github/workflows/automated-updates-to-sam-cli.yml
+++ b/.github/workflows/automated-updates-to-sam-cli.yml
@@ -97,7 +97,7 @@ jobs:
           python-version: "3.11"
 
       - name: Install uv
-        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
 
       - name: Update aws-sam-translator & commit
         run: |
@@ -166,7 +166,7 @@ jobs:
           python-version: "3.11"
 
       - name: Install uv
-        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
 
       - name: Upgrade aws_lambda_builders & commit
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
         echo "TEMP=D:\\Temp" >> $env:GITHUB_ENV  
       if: ${{ matrix.os == 'windows-latest' }}
     - uses: actions/checkout@v6
-    - uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
+    - uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
       with:
         python-version: ${{ matrix.python }}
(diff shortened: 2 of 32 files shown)


reedham-aws (dismissed):

reedham-aws (commented):
Changes look good, are there any more places that would require an update to the cryptography version? I'm not understanding why it's just `tests/integration/testdata/buildcmd/asset.b998895901bf33127f2c9dce715854f8b35aa73fb7eb5245ba9721580bbe5837/requirements.txt`.

Another example of a place with an old version is here: https://github.com/aws/aws-sam-cli/blob/aaad3ea94fad1775ad6d7b6baff1b52716eeb62f/tests/integration/testdata/buildcmd/Python/requirements.txt?plain=1#L7

EhteshamSid:
Thanks for the review! To answer your question about the other files:

The `cryptography==3.3.2` pins in `tests/integration/testdata/buildcmd/` (PyLayer, Python, PythonImage, etc.) are **intentional test fixtures** - they simulate a sample user Lambda function with a specific pinned dependency to test SAM CLI's build pipeline behavior. Bumping those would change what the tests are validating rather than fix a real exposure.

The `requirements/reproducible-linux.txt`, `reproducible-mac.txt`, and `reproducible-win.txt` files (SAM CLI's own dependency lockfiles) are already at `46.0.6` - that was handled upstream before this PR.

The `asset.b998895.../requirements.txt` file was the outlier - it's a CDK-generated integration test asset that actually gets installed during the test run, so it represented a genuine vulnerable dependency being pulled in. That's the one we updated.

Let me know if you'd like any of the other fixture files updated as well - happy to do so if that's preferred.

reedham-aws:
I don't think the 3.3.2 version is necessarily intentional, just a long time since it was updated 😅. I also don't see the difference between the CDK generated requirements file and the one that I linked before, both should get installed in the buildcmd tests via `sam build`. I think it might just be easier to update all instances.

Also I'd like to note that there are some linting errors in the CI. You can test those locally by running `make pr`.

EhteshamSid:
Fair point - updated all the remaining `3.3.2` pins in the buildcmd test fixtures too. Also fixed the linting issues (unused imports + import ordering in `durable_integ_base.py` and `regression_deploy_base.py`).

reedham-aws (dismissed):

roger-zhangg:
Please don't update `cryptography==3.3.2 ` in test data, there was an attempt of this before but the more recent cryptography requires a more recent glibc which is not available in lambda build images. If we want to update cryptography in test data, we should
-> push to a branch (not develop)
-> run integration test against that branch to verify everything works
-> merge branch to develop

EhteshamSid:
Good call - reverted the `cryptography` bump in all `testdata/buildcmd/` fixtures. Kept the bump in `durable_integ_base.py` and `regression_deploy_base.py` since those run on the CI host rather than inside Lambda build images, so the glibc constraint does not apply there.

Issue #7750, opened 2024-11-27
Bug: sam local start-api authorization log is sometimes misleading


JonZeolla:
### Description:

If you create a serverless function invoked via a httpapi event which uses IAM auth, then run `sam local start-api` you get a log saying that your authorizer is 'None' and it's unsupported/not found. This is a bit misleading and could lead to people believing that their `template.yaml` will deploy with no authorizer, which is not true. I was able to change this log by replacing `ApiId` with some example auth details, but then I had to change from `HttpApi` to `Api` which I don't want to do.

This is the specific log:

> Linking authorizer skipped for route '/example', authorizer 'None' is unsupported or not found

### Steps to reproduce:

Make this `template.yaml`:

```yaml
Resources:
  ExampleFunction:
    Type: AWS::Serverless::Function
    Properties:
      Runtime: python3.12
      CodeUri: src/
      Handler: main.lambda_handler
      Events:
        ExampleHttpApiEvent:
          Type: HttpApi
          Properties:
            Path: /example
            Method: GET
            ApiId: !Ref ExampleHttpApi
  ExampleHttpApi:
    Type: AWS::Serverless::HttpApi
    Properties:
      Auth:
        DefaultAuthorizer: AWS_IAM
        EnableIamAuthorizer: true
```

Then run `sam local start-api` and see:

> Linking authorizer skipped for route '/example', authorizer 'None' is unsupported or not found
> Mounting ExampleFunction at http://127.0.0.1:3000/example [GET]

### Observed result:

```console
➜ sam local start-api --debug
2024-11-27 15:45:07,175 | No config file found in this directory.
2024-11-27 15:45:07,177 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/Users/jonzeolla/src/testing/sam/samconfig.toml'
2024-11-27 15:45:07,177 | Config file location: /Users/jonzeolla/src/testing/sam/samconfig.toml
2024-11-27 15:45:07,178 | Config file '/Users/jonzeolla/src/testing/sam/samconfig.toml' does not exist
2024-11-27 15:45:07,179 | Using SAM Template at /Users/jonzeolla/src/testing/sam/template.yaml
2024-11-27 15:45:07,193 | OSError occurred while reading TOML file: [Errno 2] No such file or directory: '/Users/jonzeolla/src/testing/sam/samconfig.toml'
2024-11-27 15:45:07,193 | Using config file: samconfig.toml, config environment: default
[...]

PR #8912, opened 2026-04-10
fix: improve misleading authorizer log message for non-Lambda authorizers

roger-zhangg:
#7750

When using `sam local start-api` with an `HttpApi` that has `DefaultAuthorizer: AWS_IAM` and `EnableIamAuthorizer: true`, the log message is misleading:

```
Linking authorizer skipped for route '/example', authorizer 'None' is unsupported or not found
```

Two bugs:
1. The log prints `route.authorizer_name` (`None`) instead of `authorizer_name_lookup` (`AWS_IAM`)
2. The message does not distinguish between a known non-Lambda authorizer (like `AWS_IAM`) and a truly missing/invalid authorizer

- Adds a `_NON_LAMBDA_AUTHORIZERS` set (`{"AWS_IAM", "NONE"}`) to `ApiCollector`
- For known non-Lambda authorizers, logs an `INFO` message clarifying it is not supported for local emulation:
  ```
  Authorizer 'AWS_IAM' for route '/example' is not supported for local emulation, requests will not be authorized
  ```
- For truly missing/unknown authorizers, logs a `WARNING`:
  ```
  Authorizer 'NonExistentAuth' for route '/test' was not found, skipping
  ```
- Both cases now correctly log the actual authorizer name instead of `None`

- The log level for missing authorizers changes from `INFO` to `WARNING`, which is more appropriate since it likely indicates a template issue.
- The log message text changes, which could affect users who parse log output.

#### Mandatory Checklist

- [ ] Write/update unit tests
- [ ] `make pr` passes
- [ ] Write documentation


samcli/lib/providers/api_collector.py
```diff
@@ -17,6 +17,9 @@
 
 
 class ApiCollector:
+    # Authorizer types that are valid but not Lambda-based, so cannot be emulated locally
+    _NON_LAMBDA_AUTHORIZERS = {"AWS_IAM", "NONE"}
```
reedham-aws, line 21: Would Cognito authorization need to be included here? I'm not sure if we accept that or not.

roger-zhangg: I think Cognito authorization is ruled out earlier so it won't reach this check

reedham-aws (approved):

seshubaws (approved):

PR #8913, opened 2026-04-11
feat: add ability to pass docker build params in template

reedham-aws:

https://github.com/aws/aws-sam-cli/issues/8656

It had not been previously possible to pass in standard Docker arguments to the CLI build. While this isn't a big issue, it's a nice quality of life feature to add. One big feature this supports is using secrets in image builds.

Adds the ability to read extra parameters from the metadata of a SAM template, similar to how we already read build args that go into the Dockerfile. These parameters are then appended to the Docker/Finch CLI subprocess call.

## Usage Example
With a template like
```
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./src
      DockerTag: latest
      DockerBuildExtraParams:
        - "--secret"
        - "id=demo,src=./secret.txt"
```

and a Dockerfile that has

```
RUN --mount=type=secret,id=demo
```

using `sam build --use-buildkit` will yield a build that uses the secrets defined in the file `./secret.txt`.

What SAM CLI is doing is running a command like `docker buildx build --secret id=demo,src=./secret.txt`.

Should have none because this is opt-in. Integration tests should confirm this.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


reedham-aws:
I tested locally using a dockerfile that consumed a secret and confirmed that this allows users to pass any secrets file and pick it up using `--secret`. It worked as expected, consumed the secret, and did not leave a trace in the image layers.

roger-zhangg (approved):
LGTM

samcli/local/docker/image_build_client.py
```diff
@@ -135,6 +139,12 @@ def build_image(
         if target is not None:
             build_kwargs["target"] = target
 
+        if extra_params:
+            LOG.warning(
+                "DockerBuildExtraParams are not supported with the SDK build client and will be ignored. "
+                "Use --use-buildkit to enable CLI-based builds."
+            )
+
```
valerena, line 147: The only thing I'm not sure is that we don't allow this.. but of course at that point it wouldn't be just command line arguments. I just wondered if the name should be different to account for that and it didn't confuse users.  I think it's okay, but it just let me thinking.

valerena (approved):

samcli/local/docker/image_build_client.py
reedham-aws: You are right, and it's something I had considered. However, because of the way the SDK works and how BuildKit works, it's just not feasible without some rewrites. I will make sure the documentation is clear on this aspect.

PR #8941, opened 2026-04-28
fix: default to empty list for docker build extra params metadata

reedham-aws:
Integration tests are failing after: https://github.com/aws/aws-sam-cli/pull/8913.

https://github.com/aws/aws-sam-cli/actions/runs/25045806939/job/73360436775

Changes the default from `None` to an empty list to avoid trying to convert `None` to TOML.
```
Error: Unable to convert an object of <class 'NoneType'> to a TOML item
```

An empty list would be carried around, but that's not too different from the build args where there's an empty dictionary.

#### Mandatory Checklist

### DIFF
--- a/samcli/lib/samlib/resource_metadata_normalizer.py
+++ b/samcli/lib/samlib/resource_metadata_normalizer.py
@@ -191,7 +191,7 @@ class ResourceMetadataNormalizer:
             SAM_METADATA_DOCKERFILE_KEY: str(dockerfile_path.as_posix()),
             SAM_METADATA_DOCKER_CONTEXT_KEY: str(asset_path),
             SAM_METADATA_DOCKER_BUILD_ARGS_KEY: metadata.get(ASSET_DOCKERFILE_BUILD_ARGS_KEY, {}),
-            SAM_METADATA_DOCKER_BUILD_EXTRA_PARAMS_KEY: metadata.get(ASSET_DOCKER_BUILD_EXTRA_PARAMS_KEY, None),
+            SAM_METADATA_DOCKER_BUILD_EXTRA_PARAMS_KEY: metadata.get(ASSET_DOCKER_BUILD_EXTRA_PARAMS_KEY, []),
         }
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


reedham-aws:
https://github.com/aws/aws-sam-cli/actions/runs/25065781894

Edit: Integration tests that had been failing are now passing. Other failures are flaky.

samcli/lib/samlib/resource_metadata_normalizer.py
```diff
@@ -191,7 +191,7 @@ def _extract_image_asset_metadata(metadata):
             SAM_METADATA_DOCKERFILE_KEY: str(dockerfile_path.as_posix()),
             SAM_METADATA_DOCKER_CONTEXT_KEY: str(asset_path),
             SAM_METADATA_DOCKER_BUILD_ARGS_KEY: metadata.get(ASSET_DOCKERFILE_BUILD_ARGS_KEY, {}),
-            SAM_METADATA_DOCKER_BUILD_EXTRA_PARAMS_KEY: metadata.get(ASSET_DOCKER_BUILD_EXTRA_PARAMS_KEY, None),
+            SAM_METADATA_DOCKER_BUILD_EXTRA_PARAMS_KEY: metadata.get(ASSET_DOCKER_BUILD_EXTRA_PARAMS_KEY, []),
```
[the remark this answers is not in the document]
reedham-aws: Not concerned since this is caught in the integration tests. There are unit tests that test other aspects of this.

valerena, line 194: What's the difference between getting this in the `resource_metadata_normalizer` vs in the `app_builder.py` file?
https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L437

```
docker_build_extra_params = metadata.get("DockerBuildExtraParams", None)
```

should these two be consistent? (both `[]` by default?)

reedham-aws: The `resource_metadata_normalizer` path is for templates from CDK. `app_builder` is downstream and covers all cases. Theoretically, it's fine because the validation in `app_builder` covers this where it's `None`. The reason `None` was a problem is because for CDK it tries to generically convert fields to TOML, while we validate for `app_builder` to avoid this.

All that being said, I will change it just for consistency. It will work either way.

roger-zhangg (approved):

valerena (approved):

PR #9013, opened 2026-05-15
feat: add ECS and AgentCore container build/package/deploy support

### LINKED ISSUE #8933: Feature request: Extend SAM container management to Bedrock AgentCore, ECS, and Fargate
## Description

AWS SAM currently handles container image management (build, package, push) for Lambda functions, but this capability does not extend to other container-based compute targets like Bedrock AgentCore, ECS, or Fargate.

When using Bedrock AgentCore's runtime, developers must manage the entire container lifecycle outside of SAM — building, tagging, and pushing images manually — then deploy the remaining infrastructure through SAM. This breaks the unified development workflow that SAM provides for Lambda.

## Proposed Solution

Extend SAM's container management capabilities (build, package, push) to support:

- **Bedrock AgentCore** runtime containers
- **ECS** task definition containers
- **Fargate** containers

SAM already knows how to manage container images end-to-end for Lambda. Applying the same pattern to other AWS compute services would keep the workflow consistent and significantly reduce friction for developers building multi-service applications.

## Additional Context

This is especially relevant as more developers adopt Bedrock AgentCore for agent-based workloads. Having to drop out of SAM to manage the container, then come back to SAM for the rest of the infrastructure, creates an unnecessary seam in the developer experience.

## SAM CLI Version

N/A — feature request

## Additional Environment Details

N/A

singledigit:

Fixes #8933

SAM CLI provides an excellent developer experience for Lambda Image functions (`sam build && sam deploy`), but users deploying containerized workloads to ECS (Fargate) or Bedrock AgentCore must manage their Docker build/push/deploy pipeline separately — even when these resources live in the same CloudFormation template. This creates a fragmented workflow requiring external tooling for an identical operation: build image → push to ECR → deploy.

Extends the existing Lambda Image build pipeline to recognize `AWS::ECS::TaskDefinition` and `AWS::BedrockAgentCore::Runtime` resources with a `Metadata` block containing `Dockerfile` and `DockerContext`. No new commands — `sam build`, `sam package`, `sam deploy`, and `sam sync` gain awareness of these resource types.

**Template example:**
```yaml
Resources:
  MyAgent:
    Type: AWS::BedrockAgentCore::Runtime
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./agent
      Architecture: arm64
    Properties:
      AgentRuntimeArtifact:
        ContainerConfiguration:
          ContainerUri: placeholder

  MyTask:
    Type: AWS::ECS::TaskDefinition
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./app
      ContainerName: web
    Properties:
      ContainerDefinitions:
        - Name: web
          Image: placeholder
```

**Key implementation details:**
- Reuses `_build_lambda_image()` — same Docker build logic, buildkit support included
- `--resolve-image-repos` auto-creates ECR repos via companion stack
- `ContainerName` metadata targets specific containers in multi-container TaskDefinitions
- `Architecture` metadata sets `--platform` (e.g., `arm64` for AgentCore)
- `ARTIFACT_TYPE = ZIP` to pass the `PackageType` filter (these resources don't have `PackageType`)
- No SAM Transform changes needed — uses native CloudFormation resource types

**Design document:** `designs/container_image_builds_ecs_agentcore.md`

- `sam build` logs "Found N container service resource(s) to build" when applicable resources are present. No behavior change for templates without these resources.
- `--resolve-image-repos` creates ECR repos for ECS/AgentCore in addition to Lambda Image functions.
- `_update_built_resource` adds an optional `metadata` parameter (backward compatible, defaults to `None`).

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [ ] Write documentation


reedham-aws:
Hello, could you review the failures in the tests?

singledigit:
Hi @reedham-aws — the test failures have been resolved:

1. **Lint (ruff PLR2004)** — Extracted magic number to a named constant.
2. **Windows integ test** — Added `@skipIf` for Windows CI (matches existing Lambda image test pattern — Windows runners can't pull Linux images).
3. **PR Reviewer bot findings** — Fixed the substring false-positive in `_force_ecs_deployment` and added unit tests for the `ContainerName` mismatch error paths.

All 51 real checks are green now. The one remaining failure was a pyinstaller-linux infra flake (build artifact not produced), so I've retriggered CI. Ready for review when you get a chance!

bnusunny:
Thanks for this contribution — the design is generally well-shaped and the build/package paths look solid. Before we can merge, there are three correctness issues in the sync path and the multi-container handling that need to be addressed. 
### 1. `sam sync` for ECS pushes a new image but redeploys the old TaskDefinition revision

**File:** `samcli/lib/sync/flows/ecs_container_sync_flow.py:153-156`

`ECRUploader.upload` builds tags via `tag_translation` (`samcli/lib/package/image_utils.py:54`), which produces `{name}-{12-char-image-id}-{tag}`. The 12-char segment is derived from the docker image id, so any source change yields a **new** ECR tag. The flow currently:

1. Pushes the new image under the new tag.
2. Discards the upload return value (compare `ImageFunctionSyncFlow.sync` at `image_function_sync_flow.py:150`, which captures `image_uri = ecr_uploader.upload(...)` and feeds it to `update_function`).
3. Calls `ecs_client.update_service(forceNewDeployment=True)`.

`forceNewDeployment` redeploys the **same** task-definition revision, which still references the previous image tag. The freshly-pushed tag is never consulted, but the logs say "Forced new deployment for service …" so the failure is silent.

To make sync work end-to-end, the flow needs to register a new TaskDefinition revision with the updated `ContainerDefinitions[i].Image` (using the URI returned from `ecr_uploader.upload`) and then update the service to that revision. `RegisterTaskDefinition` is not called anywhere in the codebase today — that step is missing.

### 2. AgentCore sync has no rollout step

**File:** `samcli/lib/sync/flows/ecs_container_sync_flow.py` (whole `sync()` path)

`AWS::BedrockAgentCore::Runtime` is wired to `_create_ecs_container_flow` in `sync_flow_factory.py:374`, so AgentCore resources land in `ECSContainerSyncFlow.sync()`. After the ECR push, the only update step is `_force_ecs_deployment`, which iterates `physical_id_mapping` looking for `"service/"` substrings. AgentCore Runtime physical IDs are ARNs of the form `arn:aws:bedrock-agentcore:…:runtime/…` — no `service/` segment — so the loop body never fires for AgentCore. There is no `bedrock-agentcore` boto client constructed anywhere in `samcli/`, and no `UpdateAgentRuntime` call.

The design doc's success criterion 4 says sync "triggers redeployment for these resources," but the implementation doesn't honor that for AgentCore. Either:
- Implement an AgentCore-specific update step (e.g., `bedrock-agentcore.UpdateAgentRuntime` against `AgentRuntimeArtifact.ContainerConfiguration.ContainerUri`), or
- If AgentCore sync is out of scope for this PR, drop it from `GENERATOR_MAPPING` and document the limitation.

A single `ECSContainerSyncFlow` covering both resource types also feels like the wrong altitude — the runtime APIs are completely different. Consider splitting into `ECSTaskDefinitionSyncFlow` and `AgentCoreRuntimeSyncFlow` so each owns its own update path cleanly.

### 3. Multi-container TaskDefinitions silently mis-deploy when `Metadata.ContainerName` is unset

**Files:** `samcli/lib/build/app_builder.py:420-435`, `samcli/lib/package/packageable_resources.py:693-744`

When a TaskDefinition has more than one entry in `ContainerDefinitions` and no `Metadata.ContainerName`, the build path silently writes `container_defs[0]["Image"] = path` and the package path's `_get_target_index` returns `0`. Sidecars and other containers keep whatever the user wrote (often literal local strings like `"my-app:latest"` that aren't valid image URIs). CloudFormation does not validate `Image` strings, so the deploy succeeds; ECS then rejects the task at launch with `CannotPullContainerError`, far away from the SAM CLI invocation.

The design doc's example template (lines 77–92) explicitly shows `ContainerName: web` for a multi-container TaskDef, so the intent is clear — but the code doesn't enforce it. Two reasonable options:

- **Strict (preferred for multi-container):** if `len(ContainerDefinitions) > 1` and `ContainerName` is missing, raise a clear `DockerBuildFailed`/`InvalidTemplate` error pointing the user at the metadata key.
- **Relaxed:** keep the index-0 fallback but `LOG.warning` when there's more than one container, so the silent footgun is at least visible.

Either way, the docstring at `packageable_resources.py:696` ("The image reference is at `ContainerDefinitions[0].Image`") should be updated — it bakes the silent assumption into the contract.

---

A few smaller things you'll likely want to address in the same revision (non-blocking but worth bundling):

- `_update_built_resource` AgentCore branch uses chained `setdefault` on `AgentRuntimeArtifact`, which corrupts `{"Ref": "Foo"}` intrinsics or raises `AttributeError` on a non-dict value. The Lambda IMAGE branches (`resource_properties["Code"] = {"ImageUri": path}`) overwrite the whole property — please follow that pattern.
- `_force_ecs_deployment`'s `"service/" in physical_id` substring check matches App Runner / AppMesh / VPC Lattice ARNs and triggers spurious `describe_services` calls. Tighten to an ECS service ARN prefix check.
- `SyncCodeResources._accepted_resources` was not extended; `sam sync --code-resource AWS::ECS::TaskDefinition` is rejected by `click.Choice` before any flow is constructed.
- `CONTAINER_BUILD_DEFINITIONS`, `_container_build_definitions`, `get_container_build_definitions`, and `put_container_build_definition` in `build_graph.py` have zero callers and are not (de)serialized in `_read`/`_write`. Either wire them up to provide caching/incremental parity with Lambda, or drop them from this PR.
- `_get_resource_api_calls` returns `ApiCallTypes.UPDATE_FUNCTION_CODE` for ECS/AgentCore. It works today only because of CFN logical-id uniqueness; please add an ECS-appropriate enum value.
- `_has_container_build_metadata` and `_update_built_resource` should `isinstance(metadata, dict)` rather than `if not metadata`, matching the existing convention in `sam_function_provider.py:825`. Today a non-dict `Metadata` crashes with `AttributeError`.

Happy to re-review once these are addressed. Thanks again for the work — this is a feature lots of folks have been asking for.

singledigit:
Hi @bnusunny — thanks for the thorough review. All findings have been addressed and pushed. Here's what changed:

### Blocking issues

**1. `sam sync` deploys old TaskDefinition revision**
`ECRUploader.upload`'s return value is now captured. After the ECR push, the flow calls `ecs.describe_task_definition` to fetch the current definition, swaps the `image` field on the target container, registers a new revision via `ecs.register_task_definition`, then calls `ecs.update_service(taskDefinition=<new_arn>)` — matching the pattern from `ImageFunctionSyncFlow`. Tags are preserved.

**2. AgentCore sync has no rollout step**
`AWS::BedrockAgentCore::Runtime` is now routed to a new `_create_agentcore_flow` that logs a clear warning ("sam sync is not yet supported for AWS::BedrockAgentCore::Runtime — use sam deploy") and returns `None`. I agree the right long-term shape is a separate `AgentCoreRuntimeSyncFlow` with an `UpdateAgentRuntime` call; I've left that as a follow-up rather than a half-finished implementation here.

**3. Multi-container TaskDefinition silent mis-deploy**
`app_builder._update_built_resource` now raises `DockerBuildFailed` when `len(ContainerDefinitions) > 1` and `Metadata.ContainerName` is not set, with a message pointing the user at the metadata key. Went with the strict (preferred) option.

### Non-blocking fixes

- **AgentCore `setdefault` chain** — replaced with direct dict assignment (`resource_properties["AgentRuntimeArtifact"] = {"ContainerConfiguration": {"ContainerUri": path}}`) matching the Lambda IMAGE pattern.
- **`_force_ecs_deployment` ARN check** — tightened to require both `arn:aws:ecs:` prefix and `:service/` segment; App Runner / VPC Lattice / AppMesh ARNs are now correctly skipped.
- **`SyncCodeResources._accepted_resources`** — `AWS::ECS::TaskDefinition` added so `sam sync --code-resource AWS::ECS::TaskDefinition` is accepted by Click.
- **`ApiCallTypes`** — added `UPDATE_CONTAINER_IMAGE = "UpdateContainerImage"`; `ECSContainerSyncFlow._get_resource_api_calls` now uses it instead of `UPDATE_FUNCTION_CODE`.
- **`isinstance(metadata, dict)` guards** — applied in `_update_built_resource`, `_has_container_build_metadata`, and `ECSTaskDefinitionImageResource._get_target_index`.
- **Dead `build_graph.py` code** — `CONTAINER_BUILD_DEFINITIONS`, `get_container_build_definitions`, and `put_container_build_definition` removed (zero callers, not serialized in `_read`/`_write`).

Added 17 unit tests for `ECSContainerSyncFlow` covering the new sync path, ARN filtering, tag propagation, named-container targeting, and error handling.

singledigit:
Good catch on the `_get_target_index` inconsistency — the warning+fallback was the wrong call given that the build path already raises. Fixed to raise `ExportFailedError` with the same message pattern, so `sam package` now fails fast with a clear error on multi-container TaskDefinitions when `Metadata.ContainerName` is unset. Pushed in 99822c6.

singledigit:
@bnusunny All the issues from your review have been addressed and CI is fully green (the 2 remaining failures are infra flakes — network/tar errors on the runners before any tests ran, unrelated to our code). Ready for another look when you have a chance!

Issue #9005, opened 2026-05-14
Sam deploy fails on missing /tmp folder files


SodaDev:
### Title
Bug: `sam deploy` fails on v1.160.0 with nested app template lookup in `/tmp` (`Template file not found at /tmp/lambdaName/template.yaml`)

### Description
After upgrading AWS SAM CLI from `1.159.1` to `1.160.0`, deployment started failing for a project using nested `AWS::Serverless::Application` templates with 

Transform:
  - AWS::LanguageExtensions
  - AWS::Serverless-2016-10-31

The same pipeline and templates work on `1.159.1`.

Failure message:
`Error: Template file not found at /tmp/lambdaName/template.yaml`

### Steps to reproduce

1. Use a parent template (e.g. `deployments/template.yaml`) with nested applications:

```yaml
Resources:
  lambdaName:
    Type: AWS::Serverless::Application
    Properties:
      Location: lambdaName/template.yaml
```

I believe the issue is related to LanguageExtensions logic, but it breaks already existing artifacts.

SodaDev:
Here is some AI investigation on the case

### Title
Bug: `sam deploy` regression in `v1.160.0` with `AWS::LanguageExtensions` + nested `AWS::Serverless::Application` (`Location` not packaged, lookup fails under `/tmp`)

### Description
After upgrading AWS SAM CLI from `1.159.1` to `1.160.0`, deploy started failing for templates that use:

- `Transform: [AWS::LanguageExtensions, AWS::Serverless-2016-10-31]`
- nested apps via `AWS::Serverless::Application` and local `Location` paths (e.g. `./PublisherApi/template.yaml`)

Observed error:
```
Error: Template file not found at /tmp/PublisherApi/template.yaml
```

The same project and pipeline work on `1.159.1`.

---

### Repro (minimal pattern)

Parent template:
```yaml
Transform:
  - AWS::LanguageExtensions
  - AWS::Serverless-2016-10-31

Resources:
  PublisherApi:
    Type: AWS::Serverless::Application
    Properties:
      Location: ./PublisherApi/template.yaml
```

Deploy command:
```bash
sam deploy --template-file deployments/template.yaml --s3-bucket <bucket> ...
```

---

### Expected result
SAM packages/uploads nested app template referenced by `AWS::Serverless::Application.Properties.Location`, and deployment succeeds.

### Observed result
[...]

PR #9017, opened 2026-05-15
refactor(cfn-lang-ext): consolidate copy helpers and relocate prop-name helpers

bnusunny:
## Summary

Follow-up to #9009 (issue #9005). Closes a structural divergence the bot reviews on that PR repeatedly caught: two near-duplicate copy helpers — \`_copy_artifact_uris_for_type\` in \`samcli/lib/package/language_extensions_packaging.py\` and \`_copy_artifact_paths\` in \`samcli/commands/build/build_context.py\` — that differed only in their optional \`dynamic_prop_keys\` skip-set. The first round of bot review on #9009 caught the exact bug class this duplication enables: a flat-\`.get()\` slipping into one helper while the other had been jmespath-aware.

## Changes

- **New module** \`samcli/lib/cfn_language_extensions/property_paths.py\`. Promotes the four prop-name helpers (previously underscore-private in \`language_extensions_packaging.py\` but imported across packages by \`build_context.py\` and \`sam_integration.py\`) into a properly-named, public module with no leading underscores: \`get_prop_value\`, \`set_prop_value\`, \`leaf_prop_name\`, \`resolve_property_paths\`. Adds a fifth helper, \`copy_artifact_properties\`, that consolidates the two duplicated copy implementations into one.
- **New test file** \`tests/unit/lib/cfn_language_extensions/test_property_paths.py\`. Includes a contract test that asserts every entry in \`PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES\` round-trips through \`set_prop_value\` + \`get_prop_value\` and that every leaf is alphanumeric (a precondition for use in CloudFormation Mapping names and the third arg of \`Fn::FindInMap\`). Future additions to the canonical packageable-resource list are now caught at the contract layer rather than at user-template execution time.
- **Delete \`_copy_artifact_uris_for_type\`** from \`samcli/lib/package/language_extensions_packaging.py\`. Both in-file callers (\`_copy_artifact_uris\` and \`_update_foreach_with_s3_uris\`) now call \`copy_artifact_properties\` directly.
- **Delete \`BuildContext._copy_artifact_paths\`** from \`samcli/commands/build/build_context.py\`. The single call site now calls \`copy_artifact_properties\` directly.
- Update inline imports in \`build_context.py\` and \`sam_integration.py\` to point at the new \`property_paths\` module (six inline-import blocks in total).
- Rename \`test_copy_artifact_uris_for_type_*\` → \`test_copy_artifact_properties_*\` and update function-import sites across three test files.
- Delete duplicate \`get_prop_value\` / \`set_prop_value\` tests from \`test_package_context.py\`; the helper-level tests now live in the new \`test_property_paths.py\` file.

## What's NOT in this PR

- **No behavior change.** The consolidated \`copy_artifact_properties\` has the same signature and behavior as the union of the two helpers it replaces (kwargs-only \`foreach_key\` / \`dynamic_prop_keys\` for the language-extensions merge path; both omitted for the simpler build-time call site).
- **No changes to the Mapping-emission pipelines.** The build-side recursive walker (\`_update_foreach_artifact_paths\`) and the package-side detect-then-emit pipeline (\`detect_dynamic_artifact_properties\` → \`_generate_artifact_mappings\` → \`_replace_dynamic_artifact_with_findmap\`) are unchanged. Their consolidation is tracked as a separate piece of follow-up work.

## Test plan

- [x] Baseline (clean develop): \`9191 passed, 25 skipped, 28 subtests passed\`
- [x] After this PR: \`9203 passed, 25 skipped, 56 subtests passed\` — net +12 tests, +28 subtests, 0 regressions
- [x] \`ruff check samcli schema\` clean
- [x] \`mypy --no-incremental setup.py samcli tests schema\` clean
- [x] \`black --check\` clean on all touched files
- [x] End-to-end #9005 repro: \`samdev package --resolve-s3\` against \`language-extensions/tc-026-nested-application\` (parent uses \`AWS::LanguageExtensions\` + nested \`AWS::Serverless::Application\` with local \`Location: ./PublisherApi/template.yaml\`). Output template still rewrites \`Properties.Location\` to an \`https://s3.us-west-2.amazonaws.com/.../...template\` URL; the uploaded nested template still has its \`CodeUri\` rewritten to an \`s3://...\` URI. Behavior unchanged.

## Why now

PR #9009 went through three rounds of bot review, each catching a real silent-corruption bug:

1. flat-\`.get()\` consumer in \`build_context\` (the duplication this PR closes)
2. missing \`SAM*\` Mapping prefixes in \`is_sam_generated_mapping\` after the canonical dict gained new entries
3. collision-keying bug in \`_compute_mapping_name\` after the leaf-vs-dotted introduction

Bugs 2 and 3 were structural: two separate hand-maintained lists that needed to track the canonical packageable-property registry. The fixes for those — deriving the prefix list from the canonical dict, keying collision detection on the leaf — landed in #9009 itself. Bug 1's fix (route \`_copy_artifact_paths\` through the new jmespath helpers) also landed in #9009, but the *duplication* between the two copy helpers remained, leaving the failure mode latent. This PR closes that.

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -504,7 +504,13 @@ class BuildContext:
             elif isinstance(resource_value, dict) and resource_key in modified_resources:
                 # Regular resource - copy updated paths from modified template
                 modified_resource = modified_resources.get(resource_key, {})
-                self._copy_artifact_paths(resource_value, modified_resource)
+                from samcli.lib.cfn_language_extensions.property_paths import copy_artifact_properties
+
+                copy_artifact_properties(
+                    resource_value.get("Properties", {}),
+                    modified_resource.get("Properties", {}),
+                    resource_value.get("Type", ""),
+                )
 
         # Merge generated Mappings into the template
         if all_generated_mappings:
@@ -552,13 +558,13 @@ class BuildContext:
             Generated Mappings section for dynamic artifact properties (empty dict if none)
         """
         from samcli.lib.cfn_language_extensions.models import PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES
-        from samcli.lib.cfn_language_extensions.sam_integration import resolve_collection
-        from samcli.lib.package.language_extensions_packaging import (
-            _get_prop_value,
-            _leaf_prop_name,
-            _resolve_property_paths,
-            _set_prop_value,
+        from samcli.lib.cfn_language_extensions.property_paths import (
+            get_prop_value,
+            leaf_prop_name,
+            resolve_property_paths,
+            set_prop_value,
         )
+        from samcli.lib.cfn_language_extensions.sam_integration import resolve_collection
 
         generated_mappings: Dict[str, Dict[str, Dict[str, str]]] = {}
 
@@ -606,15 +612,15 @@ class BuildContext:
                 continue
 
             candidate_paths = PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES.get(resource_type, [])
-            for prop_name in _resolve_property_paths(candidate_paths, properties):
-                prop_value = _get_prop_value(properties, prop_name)
+            for prop_name in resolve_property_paths(candidate_paths, properties):
+                prop_value = get_prop_value(properties, prop_name)
                 if prop_value is None:
                     continue
 
                 # CloudFormation Mapping names and FindInMap third-args must be alphanumeric,
                 # so dotted property names (e.g. "Command.ScriptLocation") use the leaf segment
                 # for those identifiers while the dotted form is preserved for property addressing.
-                leaf_name = _leaf_prop_name(prop_name)
+                leaf_name = leaf_prop_name(prop_name)
 
                 if contains_loop_variable(prop_value, loop_variable) and collection_values:
                     # Determine which outer loop variables the property references
@@ -653,7 +659,7 @@ class BuildContext:
                         else:
                             lookup_key = {"Ref": loop_variable}
 
-                        _set_prop_value(properties, prop_name, {"Fn::FindInMap": [mapping_name, lookup_key, leaf_name]})
+                        set_prop_value(properties, prop_name, {"Fn::FindInMap": [mapping_name, lookup_key, leaf_name]})
                 else:
                     expanded_key = self._build_expanded_key(
                         resource_template_key,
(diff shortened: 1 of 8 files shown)


samcli/commands/build/build_context.py
```diff
@@ -504,7 +504,13 @@ def _update_original_template_paths(self, original_template: Dict, modified_temp
             elif isinstance(resource_value, dict) and resource_key in modified_resources:
                 # Regular resource - copy updated paths from modified template
                 modified_resource = modified_resources.get(resource_key, {})
-                self._copy_artifact_paths(resource_value, modified_resource)
+                from samcli.lib.cfn_language_extensions.property_paths import copy_artifact_properties
```
reedham-aws, line 507: We should import everything at the top level.

reedham-aws: I guess that's not the pattern that is already being used here, though. We can just keep it like this then, although I think it should be changed.

roger-zhangg (approved):
LGTM. Clean separation of concerns: consolidates `_copy_artifact_uris_for_type` / `_copy_artifact_paths` duplication into a single `copy_artifact_properties`, and promotes the underscore-private jmespath helpers into a properly public `property_paths` module.

Verified:
- **No behavioral changes** — the consolidated helper has the same logic as the union of both predecessors (optional `foreach_key`/`dynamic_prop_keys` kwargs for the merge path, omitted for the simpler build-time path).
- **Tests properly adapted** — existing test coverage is preserved with only the call-site renames (`_copy_artifact_uris_for_type` -> `copy_artifact_properties`), duplicate low-level tests moved to the new `test_property_paths.py`, and new contract tests added (round-trip + alphanumeric-leaf assertions on every entry in `PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES`).
- **Import sites correctly updated** — all 6 inline import blocks in `build_context.py` and the top-level import in `sam_integration.py` point at the new module with the public names.

One note: both this PR and #9018 are in CONFLICTING state with `develop`. After resolving merge conflicts with the base branch, whichever of #9017 / #9018 merges second will also need a trivial rebase against the other (they both modify import blocks in `build_context.py` and `language_extensions_packaging.py`). Despite the PR #9018 description saying merge order doesn't matter, merging #9017 first is slightly cleaner — #9018 still imports `_leaf_prop_name` from `language_extensions_packaging`, which this PR removes.

PR #9018, opened 2026-05-15
refactor(cfn-lang-ext): extract shared Mapping-name primitives for Fn::ForEach

bnusunny:
## Summary

Follow-up to #9009 (issue #9005). Closes the second class of structural divergence the bot reviews on that PR repeatedly caught: the Mapping name, Fn::FindInMap lookup key, and compound Mapping table key for nested Fn::ForEach were each constructed inline in two places — once in the build-time pipeline (\`samcli/commands/build/build_context.py\`) and once in the package-time pipeline (\`samcli/lib/package/language_extensions_packaging.py\`). Bot review #2 caught a missing prefix list; bot review #3 caught a collision counter keyed on the dotted path. Both bugs were duplicate string-construction logic drifting between sites.

This PR extracts the three computations into a single module so neither pipeline can drift from the other again.

## Changes

- **New module** \`samcli/lib/cfn_language_extensions/foreach_mapping_helpers.py\` with three pure functions:
  - \`compute_mapping_name(leaf, nesting_path, *, has_collision, resource_template_key)\` — builds \`SAM<leaf><nesting_path>[<resource-suffix>]\` with collision-aware suffix.
  - \`compute_lookup_key(loop_variable, referenced_outer_vars)\` — builds the second arg of Fn::FindInMap (\`{Ref}\` or \`{Fn::Join}\`).
  - \`compute_compound_mapping_key(outer_values, inner_value)\` — dash-joined Mapping table key for nested Fn::ForEach.
- **New test file** \`tests/unit/lib/cfn_language_extensions/test_foreach_mapping_helpers.py\` with 19 tests, including:
  - Parameterized construction cases for each primitive.
  - **Cross-pipeline equivalence test** — asserts that the build-side collision-counting input and the package-side collision-counting input produce byte-identical Mapping names for the same logical inputs. This is the structural guarantee that future drift can't recur.
- **Replace package-side inline constructions** in \`samcli/lib/package/language_extensions_packaging.py\`:
  - \`_compute_mapping_name\` body becomes a 3-line delegation to \`compute_mapping_name\`.
  - \`_replace_dynamic_artifact_with_findmap\` default \`mapping_name\` and inline \`lookup_key\` construction become helper calls.
  - \`_generate_artifact_mappings\` inline compound-key construction becomes a helper call.
- **Replace build-side inline constructions** in \`samcli/commands/build/build_context.py\`:
  - \`_update_foreach_artifact_paths\` inline \`mapping_name\` and \`lookup_key\` constructions become helper calls; collapses two if/else branches in the process.
  - \`_collect_nested_mapping_entry\` inline compound-key construction becomes a helper call.
  - Removes the now-unused \`sanitize_resource_key_for_mapping\` import.

## What this PR does NOT do

- **Does not unify orchestration.** The build walker stays a recursive walker; the package detect-then-emit pipeline stays a detect-then-emit pipeline. Both still exist; they just compute Mapping identifiers through one shared source of truth.
- **Does not touch auto-dependency-layer Mapping emission** (\`_collect_foreach_layer_mappings\` in \`build_context.py:670\`). That stays in \`build_context.py\` interleaved with the artifact loop. Its Mapping name (\`SAMLayers<nesting_path>\`) is constructed inline at \`build_context.py:666\` and is intentionally not affected.
- **Does not change \`DynamicArtifactProperty\` or \`PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES\`.** No detection changes.
- **Does not consolidate the value extractors** (\`_get_artifact_value\` vs \`_find_artifact_uri_for_resource\`). Their normalization shapes are genuinely different (raw string vs URI dict→string).

## Test plan

- [x] Baseline (clean develop): \`9191 passed, 25 skipped, 28 subtests passed\`
- [x] After this PR: \`9211 passed, 25 skipped, 28 subtests passed\` — net +20 tests, 0 regressions
- [x] Targeted suite (\`cfn_language_extensions\`, \`package\`, \`buildcmd\`): 2052 passed
- [x] \`ruff check samcli schema\` clean
- [x] \`mypy --no-incremental setup.py samcli tests schema\` clean
- [x] \`black --check\` clean on all touched files
- [x] Leaf-collision regression tests from #9009 commit \`5e818202f\` still pass:
  - \`test_leaf_collision_across_resource_types_disambiguates_mapping_name\` (build path)
  - \`test_generate_artifact_mappings_disambiguates_leaf_collision_across_types\` (package path)
- [x] End-to-end #9005 repro: \`samdev package --resolve-s3\` against \`language-extensions/tc-026-nested-application\`. Output template still rewrites \`Properties.Location\` to an \`https://s3.us-west-2.amazonaws.com/.../...template\` URL. Behavior unchanged.

## Relationship to PR #9017

Independent. PR #9017 consolidates the *copy* helpers and relocates the prop-name helpers into a public module. PR B (this one) extracts the *Mapping-name* primitives. They touch different functions in the same files; the merge order doesn't matter.

### DIFF
--- a/samcli/commands/build/build_context.py
+++ b/samcli/commands/build/build_context.py
@@ -38,7 +38,6 @@ from samcli.lib.build.exceptions import (
 from samcli.lib.build.workflow_config import UnsupportedRuntimeException
 from samcli.lib.cfn_language_extensions.sam_integration import (
     contains_loop_variable,
-    sanitize_resource_key_for_mapping,
     substitute_loop_variable,
 )
 from samcli.lib.cfn_language_extensions.utils import is_foreach_key
@@ -551,6 +550,10 @@ class BuildContext:
         Dict[str, Dict[str, Dict[str, str]]]
             Generated Mappings section for dynamic artifact properties (empty dict if none)
         """
+        from samcli.lib.cfn_language_extensions.foreach_mapping_helpers import (
+            compute_lookup_key,
+            compute_mapping_name,
+        )
         from samcli.lib.cfn_language_extensions.models import PACKAGEABLE_RESOURCE_ARTIFACT_PROPERTIES
         from samcli.lib.cfn_language_extensions.sam_integration import resolve_collection
         from samcli.lib.package.language_extensions_packaging import (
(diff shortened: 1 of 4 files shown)


roger-zhangg (approved):
LGTM. This cleanly extracts the three Mapping-identifier computations (mapping name, lookup key, compound table key) into a shared `foreach_mapping_helpers` module, eliminating the structural divergence risk between build-time and package-time pipelines.

Verified:
- **No behavioral changes** — the extracted helpers produce byte-identical outputs to the inline constructions they replace. The `compute_mapping_name` / `compute_lookup_key` / `compute_compound_mapping_key` functions are pure, well-documented, and cover all branches (no-collision vs collision, no-outer-vars vs compound join, etc.).
- **Cross-pipeline equivalence tests** — the `TestCrossPipelineEquivalence` class is a valuable structural safeguard; it asserts both pipelines produce the same Mapping name for equivalent inputs and will fail loudly if a future change drifts one side.
- **Package-side `_compute_mapping_name`** — correctly becomes a thin delegation wrapper (retains its `collision_groups` Counter plumbing, which is pipeline-specific orchestration, while the raw name construction delegates to the shared helper).
- **Build-side simplification** — the two if/else branches for `lookup_key` and the inline compound-key construction in `_collect_nested_mapping_entry` collapse cleanly into single helper calls.

One note: both this PR and #9017 are in CONFLICTING state with `develop`. While the PR description says merge order with #9017 doesn't matter, this PR still imports `_leaf_prop_name` from `language_extensions_packaging` (line ~835 of `build_context.py`). PR #9017 removes that function and replaces it with `leaf_prop_name` in the new `property_paths` module. So merging #9017 first is the cleaner path — after that, the rebase for this PR would just update that one import to `from samcli.lib.cfn_language_extensions.property_paths import leaf_prop_name`.

PR #9031, opened 2026-05-19
fix(cfn-lang-ext): preserve inline-source intrinsics through LE-aware build merge (#9029)

### LINKED ISSUE #9029: Bug: Fn::Sub inside ZipFile inline Lambda code is evaluated at build time with wrong pseudo-parameter values in 1.160.x
### Description:

When a template uses `AWS::LanguageExtensions` as a transform and has an `AWS::Lambda::Function` resource with inline `ZipFile` code containing `!Sub` (to interpolate `${AWS::Region}` or `${AWS::AccountId}`), `sam build` evaluates the `Fn::Sub` **at build time** rather than leaving it as a CloudFormation intrinsic for deployment-time resolution.

The result is that the built template has the substitutions baked in using whatever defaults the build environment has — typically `us-east-1` for region and `123456789012` for account ID — instead of the actual deployment target values. The `Fn::Sub` wrapper is stripped entirely from the built template.

This is a regression introduced in 1.160.0 alongside the in-tree CloudFormation Language Extensions resolver.

### Steps to reproduce:

1. Create `template.yaml`:

```yaml
Transform:
  - AWS::LanguageExtensions
  - AWS::Serverless-2016-10-31

Resources:
  MyTriggerFunction:
    Type: AWS::Lambda::Function
    Properties:
      FunctionName: my-trigger
      Runtime: python3.11
      Handler: index.handler
      Role: !Sub 'arn:aws:iam::${AWS::AccountId}:role/my-role'
      Code:
        ZipFile: !Sub |
          import boto3
          def handler(event, context):
              client = boto3.client('stepfunctions')
              client.start_execution(
                  stateMachineArn='arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:my-state-machine',
              )
```

2. Run `sam build`
3. Inspect `.aws-sam/build/template.yaml`

### Observed result:

The built template has `Fn::Sub` stripped and `${AWS::Region}` / `${AWS::AccountId}` resolved to build-time defaults:

```yaml
Code:
  ZipFile: "import boto3\ndef handler(event, context):\n    client = boto3.client('stepfunctions')\n\
    \    client.start_execution(\n        stateMachineArn='arn:aws:states:us-east-1:123456789012:stateMachine:my-state-machine',\n\
    \    )\n"
```

Note: `Fn::Sub` in other parts of the template (e.g. `Role: !Sub '...'`) is correctly preserved as a CloudFormation intrinsic and not evaluated at build time.

### Expected result:

The built template should preserve `Fn::Sub` as a CloudFormation intrinsic, identical to the behaviour in 1.148.0:

```yaml
Code:
  ZipFile:
    Fn::Sub: "import boto3\ndef handler(event, context):\n    client = boto3.client('stepfunctions')\n\
      \    client.start_execution(\n        stateMachineArn='arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:my-state-machine',\n\
      \    )\n"
```

CloudFormation should then resolve `${AWS::Region}` and `${AWS::AccountId}` to the actual deployment region and account at stack create/update time.

### Additional environment details

**Working:** SAM CLI 1.148.0
**Broken:** SAM CLI 1.160.0, 1.160.1

The regression appears to be tied to the introduction of the in-tree `cfn_language_extensions` resolver in 1.160.0 (see #9010 for a related fix in the same area). The template must have `AWS::LanguageExtensions` in its `Transform` list to trigger the issue — templates without it are unaffected.

1. OS: macOS (arm64)
2. `sam --version`: 1.160.1 (broken), 1.148.0 (working)
3. AWS region: ap-southeast-2

**SAM 1.160.1:**
```json
{
  "version": "1.160.1",
  "system": {
    "python": "3.13.9",
    "os": "macOS-26.5-arm64-arm-64bit-Mach-O"
  }
}
```

**SAM 1.148.0:**
```json
{
  "version": "1.148.0",
  "system": {
    "python": "3.11.10",
    "os": "macOS-26.5-arm64-arm-64bit"
  }
}
```

bnusunny:
## Summary

Fixes #9029. Under `Transform: AWS::LanguageExtensions`, `sam build` was overwriting `Code: {ZipFile: !Sub ...}` with the LE-resolved value, baking default pseudo-parameter substitutions (`us-east-1`, `123456789012`) into the built template.

The merge had three sites that all need the same discipline as the non-LE `ApplicationBuilder.update_template` early-skip — "if no build artifact, don't overwrite":

- **Root-level merge** (`_update_original_template_paths`): skip when `get_full_path(stack_path, key)` is absent from `artifacts`.
- **ForEach static branch** (`_merge_static_artifact_path`): same check per expanded key.
- **ForEach dynamic branch** (`_collect_dynamic_mapping_entries` + nested): same check per expanded key.

## Test plan

- [x] Unit: `pytest tests/unit/commands/buildcmd/test_build_context.py tests/unit/commands/buildcmd/test_build_context_language_extensions.py` — 93 tests pass; new coverage for skip (cases A/B) and refusal (case C).
- [x] Integration testdata + tests added for all three cases:
  - `test_build_zipfile_fnsub_preserves_intrinsic` (case A: root-level)
  - `test_build_foreach_static_zipfile_fnsub_preserves_intrinsic` (case B: ForEach static)
  - `test_build_foreach_dynamic_inline_zipfile_preserved` (case C: ForEach dynamic,)
- [x] `black --check` clean on touched files.


samcli/commands/build/build_context.py
```diff
@@ -481,7 +504,20 @@ def _update_original_template_paths(self, original_template: Dict, modified_temp
...
             The stack being processed
+        artifacts : Optional[Dict[str, str]]
+            Map of resource full paths to built artifact locations. Used to
+            skip merge for resources that produced no build artifact (e.g.
+            inline-source ``ZipFile`` Lambdas). Mirrors the non-LE
+            ``ApplicationBuilder.update_template`` skip. See #9029.
+        stack_output_template_path_by_stack_path : Optional[Dict[str, str]]
+            Map of stack paths to their output template paths (for nested stacks).
         """
+        from samcli.lib.providers.provider import get_full_path
```
valerena: let's not have inline imports, and instead add them at the top

samcli/commands/build/build_context.py
```diff
@@ -711,6 +784,54 @@ def _update_foreach_artifact_paths(
...
+            _set_prop_value(properties, prop_name, artifact_value)
+
+    @staticmethod
+    def _raise_dynamic_no_artifacts_unsupported(resource_template_key: str, loop_variable: str, leaf_name: str) -> None:
+        """Refuse a Fn::ForEach body whose dynamic artifact property produced no
+        build artifacts for any iteration (e.g. an inline-source ``Code: {ZipFile:
+        ...}`` Lambda has no ``CodeUri`` to build). Mappings hold only static
+        strings, so we cannot emit a per-iteration value that preserves deploy-time
+        intrinsics, and we cannot leave the loop variable unresolved. See #9029.
+        """
```
valerena: #9029 doesn't have any `Fn::ForEach` involved, so I'm not sure I understand this comment. 

Also, I would think that we should just skip any transformation if we see something that we shouldn't touch (instead of throwing an error). (assuming it's valid CFN that customers might want to deploy). Maybe just print a warning? I'm thinking if there are two functions and one has this, we don't need to throw a big exception if only one has an issue.

samcli/commands/build/build_context.py
```diff
@@ -787,11 +910,18 @@ def _collect_dynamic_mapping_entries(
...
+        produced a build artifact — same discipline as the static branch and
+        the root-level merge. Iterations whose ``get_full_path(stack_path,
+        expanded_key)`` is absent from ``artifacts`` are skipped. See #9029.
         """
         from samcli.lib.package.language_extensions_packaging import _leaf_prop_name
+        from samcli.lib.providers.provider import get_full_path
 
         mapping_entries: Dict[str, Dict[str, str]] = {}
         leaf_name = _leaf_prop_name(prop_name)
+        artifacts = artifacts if artifacts is not None else {}
```
valerena: We don't want to make this change in every function. If we don't use this variable for anything other than just passing it to the next function, we should just pass it along without any check. Let's only force it to a `dict` if we really need to (for example if it's not Optional in the method we're passing it to)

samcli/commands/build/build_context.py
```diff
@@ -845,6 +986,8 @@ def _collect_nested_mapping_entry(
                 expanded_key = substitute_loop_variable(expanded_key, ovar, oval)
             expanded_key = substitute_loop_variable(expanded_key, loop_variable, coll_value)
 
+            if get_full_path(stack_path, expanded_key) not in artifacts:
```
valerena: Instead of adding the new line above `artifacts = artifacts if artifacts is not None else {}`, we can just change this one here to 
```suggestion
            if not artifacts or get_full_path(stack_path, expanded_key) not in artifacts:
```

samcli/commands/build/build_context.py
bnusunny: This is for the dynamic ForEach case. For example, people could put the loop var in the ZipFile body. Waite, this actually works! Each function has its own inline code with `${ServiceName}` replaced. Yes, we should allow this case. 

```yaml
Resources:
  Fn::ForEach::ServiceLoop:
    - ServiceName
    - ["api", "worker", "scheduler"]
    - ${ServiceName}Function:
        Properties:
          Handler: index.handler
          FunctionName:
            Fn::Sub: ${ServiceName}-handler
          Code: 
               ZipFile: !Sub |
                  def handler(event, context):
                      print("Name:  ${ServiceName}")
```

roger-zhangg (dismissed):

valerena (dismissed):

valerena (approved):

roger-zhangg (approved):

Issue #9027, opened 2026-05-18
Bug: The location parameter is not a valid S3 uri


danjhd:
### Description:
In 1.160.0 and 1.1.60.1:
When using AWS::Include with the Fn::Transform function the path to the file is not correctly set when using `sam deploy`.
The fails to deploy. This is a new bug sice 1.160.0

### Steps to reproduce:
Create  a new template that uses a snippet like this, also create a file called export-events.json in the same folder as the template

```
  Parameter:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      DataType: text
      Value:
        Fn::ToJsonString:
          Fn::Transform:
            Name: AWS::Include
            Parameters:
              Location: export-events.json
```

### Observed result:
```
Waiting for changeset to be created..

2026-05-18 18:40:14,077 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2026-05-18 18:40:14,645 | No container socket path in global storage, falling back to DOCKER_HOST detection:
2026-05-18 18:40:14,647 | Telemetry endpoint configured to be https://aws-serverless-tools-telemetry.us-west-2.amazonaws.com/metrics
2026-05-18 18:40:14,648 | Sending Telemetry: {'metrics': [{'commandRun': {'requestId': 'd342929b-8650-43f2-bdbd-e9abc3b5b18d', 'installationId': 'f58a317c-44f2-41cb-b6c6-bab84e920a15', 'sessionId':
'df1236c0-d92c-437b-9d08-727104116fc4', 'executionEnvironment': 'CLI', 'ci': False, 'pyversion': '3.13.12', 'samcliVersion': '1.160.1', 'osPlatform': 'Windows', 'awsProfileProvided': False, 'debugFlagProvided': True, 
'region': 'eu-west-2', 'commandName': 'sam deploy', 'metricSpecificAttributes': {'projectType': 'CFN', 'gitOrigin': 'b398845365a4018d12e842198ce76bf7364ddf7f9421dafd6aa1cb816b400c3f', 'projectName':
'c30e4251488023561baee9b93d8e6e89e0845010dce7bddb80dcb1d22becc12c', 'initialCommit': 'a87ac7ad8d1009199bbe1f1b1e4bfa0c409753baccc5ef5a26266d8190c8fbf9', 'containerEngine': 'docker-default', 'adminContainerPreference':
[...]

Issue #9004, opened 2026-05-14
Bug: `sam build` fails with "Fn::FindInMap layout is incorrect" when using `AWS::LanguageExtensions` with an unresolved `!Ref` parameter as a map key (regression in 1.160.0)


polyomino24:
### Description:

`sam build` fails with `Fn::FindInMap layout is incorrect` when a template uses `AWS::LanguageExtensions` with `Fn::FindInMap` where a template parameter is used as a map key via `!Ref`, and no value is supplied for that parameter at build time (i.e., without `--parameter-overrides` and no `Default` defined).
This seems to be a regression introduced in 1.160.0 — the same template built successfully with 1.159.1.

### Steps to reproduce:

1. Create a new directory and add the following files:

samconfig.yaml

```yaml
version: 0.1
stg:
  global:
    parameters:
      parameter_overrides:
        - ENV=stg
```

template.yaml

```yaml
AWSTemplateFormatVersion: 2010-09-09
Transform:
  - AWS::Serverless-2016-10-31
  - AWS::LanguageExtensions

Parameters:
  ENV:
    Type: String
    AllowedValues: [stg, prod]

Resources:
  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: .
      Handler: index.handler
      Runtime: nodejs22.x
      Role: !FindInMap [EnvConfig, !Ref ENV, IAM]

Mappings:
  EnvConfig:
    stg:
      IAM: arn:aws:iam::123456789012:role/lambda-role-stg
    prod:
      IAM: arn:aws:iam::123456789012:role/lambda-role-prod
```

index.js

```js
export const handler = async (_event, _context) => {
  return {
    statusCode: 200,
    body: JSON.stringify({
      message: 'Hello World',
    }),
  }
}
```

2. Run `sam build` in that directory with AWS SAM CLI v1.160.0

### Observed result:

<details>
<summary>Show output</summary>

```
% sam build --debug      
2026-05-14 15:42:16,844 | Config file location:                                            
/Users/ta.ishikawa/sam-cli-bug/samconfig.yaml                                              
2026-05-14 15:42:16,849 | Loading configuration values from [default.['build'].parameters] 
(env.command_name.section) in config file at                                               
'/Users/ta.ishikawa/sam-cli-bug/samconfig.yaml'...                                         
2026-05-14 15:42:16,849 | Configuration values successfully loaded.                        
[...]

PR #9033, opened 2026-05-20
feat: make AWS::LanguageExtensions local processing opt-in via --language-extensions flag

bnusunny:
## Summary

- Make local `AWS::LanguageExtensions` processing opt-in (off by default), restoring pre-1.160.0 behavior for templates that declare the transform but don't need local expansion
- Add `--language-extensions / --no-language-extensions` CLI flag to all 8 affected commands: build, package, deploy, sync, validate, local invoke, local start-api, local start-lambda
- Support `SAM_CLI_ENABLE_LANGUAGE_EXTENSIONS=1` env var as fallback (flag wins)
- Support `samconfig.toml` persistence via `language_extensions = true`

## Motivation

SAM CLI 1.160.0 enabled local processing of `AWS::LanguageExtensions` implicitly when the transform was declared. This caused regressions (#9004, #9005, #9027, #9029) for templates that declared the transform for non-`Fn::ForEach` reasons or as a forward-compat placeholder. Making it opt-in lets customers adopt at their own pace.

## Changes

1. `expand_language_extensions()` now requires `enabled: bool` as a keyword-only arg — unwired callers fail with TypeError
2. `resolve_language_extensions_enabled(flag_value: Optional[bool]) -> bool` resolver (flag → env var → False)
3. `@language_extensions_option` Click decorator shared across all commands
4. `SamLocalStackProvider.get_stacks(..., language_extensions_enabled=False)` kwarg threaded through recursion
5. Each command context (Build, Package, Deploy, Sync, Validate, InvokeContext) wired end-to-end
6. `Template` class in artifact_exporter wired for nested-stack recursion
7. Inspection-path callers (sam list, companion stack, image-repo validation) explicitly pass `False`
8. Telemetry regression tests lock in: event fires only when opt-in AND template uses LE
9. Integration tests updated to pass `--language-extensions` for ForEach templates
10. Documentation updated with opt-in instructions

## Test plan

- [x] 9237 unit tests pass (`pytest tests/unit/ -q -W ignore::pytest.PytestUnknownMarkWarning`)
- [x] `--language-extensions` parameter registered on all 8 commands (verified programmatically)
- [x] `make format` clean
- [x] `make lint` shows zero errors
- [x] Integration test `test_build_without_flag_does_not_expand_foreach` verifies opt-out behavior
- [x] Existing ForEach integration tests pass with `--language-extensions` flag


tests/integration/buildcmd/test_build_cmd_language_extensions.py
```diff
@@ -147,3 +147,20 @@ def test_build_nested_foreach_dynamic_codeuri_generates_mappings(self):
...
+
+    def test_build_without_flag_does_not_expand_foreach(self):
+        """Without --language-extensions, sam build does not expand Fn::ForEach."""
+        runtime = self._get_python_version()
+        overrides = {"Runtime": runtime}
+        cmdlist = self.get_command_list(parameter_overrides=overrides)
+
+        command_result = run_command(cmdlist, cwd=self.working_dir)
+
+        if command_result.process.returncode == 0:
```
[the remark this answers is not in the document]
bnusunny: resolved in commit https://github.com/aws/aws-sam-cli/pull/9033/commits/d7e6b4ab1541837a4de38adebdeeab06ac2663a5

samcli/lib/providers/sam_function_provider.py
```diff
@@ -1020,6 +1020,7 @@ def _refresh_loaded_functions(self) -> None:
                     template_file,
                     parameter_overrides=self._parameter_overrides,
                     global_parameter_overrides=self._global_parameter_overrides,
+                    language_extensions_enabled=False,
```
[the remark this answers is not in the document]
bnusunny: Resolved in commit https://github.com/aws/aws-sam-cli/pull/9033/commits/6fbb27b50ccaec7667c628236bf0ab749d3f64da

bnusunny:
**Re: PackageContext not threaded in deploy command (review 2026-05-20T17:51)**

Fixed in ed0d4f965 — `language_extensions=language_extensions` is now passed to the `PackageContext(...)` call inside `do_cli`, matching the sync command's wiring.

bnusunny:
**Re: _detect_foreach_code_changes crash when LE disabled (review 2026-05-20T18:04)**

Fixed in 9f5934be5 — added early return when `language_extensions_enabled` is False, merged with the existing `check_using_language_extension` guard into a single condition.

samcli/lib/cfn_language_extensions/sam_integration.py
```diff
@@ -476,6 +509,8 @@ def detect_dynamic_artifact_properties(
 def expand_language_extensions(
     template: Dict[str, Any],
     parameter_values: Optional[Dict[str, Any]] = None,
+    *,
+    enabled: bool,
```
valerena, line 513: why making it required and always a keyword argument? Wouldn't it be okay to just allow not passing it and defaulting to "False"? (or also allow `, True` to be passed as a positional argument instead of forcing the `, enabled: True`?)

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -125,6 +125,7 @@ def _is_all_image_funcs_provided(template_file, image_repositories, parameters_o
         template_file,
         parameter_overrides=parameters_overrides,
         global_parameter_overrides=global_parameter_overrides,
+        language_extensions_enabled=False,
```
valerena, line 128: No languages_extensions here? Can a customer use ForEach to provide multiple Image functions with different ECR repos or something?

tests/unit/commands/deploy/test_command.py
```diff
@@ -496,9 +502,10 @@ def test_all_args_guided(
                 signing_profiles=self.signing_profiles,
                 use_changeset=self.use_changeset,
                 disable_rollback=True,
-                poll_delay=5,
+                poll_delay=5.0,
```
valerena, line 505: This seems like an unrelated change.. I guess it's not breaking the tests, but it's weird to have it.

samcli/commands/build/build_context.py
```diff
@@ -93,6 +94,7 @@ def __init__(
         mount_with: str = MountMode.READ.value,
         mount_symlinks: Optional[bool] = False,
         use_buildkit: Optional[bool] = False,
+        language_extensions: Optional[bool] = None,
```
vicheey, line 97: `False`?

samcli/commands/_utils/options.py
```diff
@@ -1074,3 +1074,23 @@ def container_env_var_file_click_option(cls):
...
 def container_env_var_file_option(f, cls):
     return container_env_var_file_click_option(cls)(f)
+
+
+def language_extensions_click_option():
+    return click.option(
+        "--language-extensions/--no-language-extensions",
+        "language_extensions",
+        required=False,
+        default=None,
```
vicheey, line 1084: `False`

valerena (dismissed):
Just a couple of questions, but it looks good.

samcli/commands/deploy/deploy_context.py
```diff
@@ -99,7 +101,7 @@ def __init__(
         self.region = region
         self.profile = profile
         self.s3_uploader = None
-        self.deployer = None
+        self.deployer: Optional[Deployer] = None
```
vicheey, line 104: why do we need to change this?

vicheey: Sync offline. For type check

samcli/commands/local/cli_common/invoke_context.py
```diff
@@ -271,6 +275,9 @@ def __enter__(self) -> "InvokeContext":
         if self._function_logical_ids:
             _function_providers_kwargs["function_logical_ids"] = self._function_logical_ids
 
+        if self._containers_mode == ContainersMode.WARM:
+            _function_providers_kwargs["language_extensions_enabled"] = self._language_extensions_enabled
```
vicheey, line 279: Can we have context noting here why we need `language_extensions_enabled` for`WARM` mode only?

samcli/lib/cfn_language_extensions/sam_integration.py
```diff
@@ -497,6 +532,10 @@ def expand_language_extensions(
         The raw template dictionary
     parameter_values : dict, optional
         Template parameter values (may include pseudo-parameters like AWS::Region)
+    enabled : bool
```
vicheey, line 535: The current pattern calls `expand_language_extensions(..., enabled=False)`  at every call site, which internally just returns a passthrough `LanguageExtensionResult` (same template reference, empty dynamic properties, h`ad_language_extensions=False`).

This reads a bit oddly — we're calling a function named "expand" only to immediately short-circuit. Two alternatives worth considering:
  1. Guard at the call site — skip the call entirely when disabled and construct the passthrough result (or just use the template dict directly):
   ```python 
   if language_extensions_enabled:
         lang_ext_result = expand_language_extensions(template_dict, parameter_values=parameter_values, enabled=True)
   else:
        lang_ext_result = LanguageExtensionResult(expanded_template=template_dict, original_template=template_dict)
   ```
  2. Rename to something like `process_language_extensions` so the "do nothing" path doesn't feel semantically wrong.

I see the upside of the current approach — uniform one-liner at each call site, single place that defines "what disabled means." Bringing this up for any follow up change. 

Not blocking, just flagging for discussion.

samcli/commands/build/build_context.py
bnusunny: Intentionally `None` (not `False`). This is a tri-state: `True` = user passed `--language-extensions`, `False` = user passed `--no-language-extensions`, `None` = user didn't pass either (resolver falls back to `SAM_CLI_ENABLE_LANGUAGE_EXTENSIONS` env var). If we defaulted to `False`, the env var could never activate LE.

samcli/commands/_utils/options.py
bnusunny: Same as above — `None` enables the tri-state (`--language-extensions` / `--no-language-extensions` / neither). The resolver treats `None` as "check env var", `False` as "explicitly disabled".

samcli/commands/deploy/deploy_context.py
bnusunny: Added `Optional[Deployer]` type annotation to fix a pre-existing mypy error (`"None" has no attribute "create_and_wait_for_changeset"`). The `make lint` target runs mypy and these errors were blocking CI. The annotation + assert make the type narrowing explicit.

tests/unit/commands/deploy/test_command.py
bnusunny: This adds `poll_delay=5.0` to the mock assertion — needed because the test patches `DeployContext` at the source module (required after restoring cycle-breaking inline imports in `command.py`). The assertion was previously incomplete and happened to pass by coincidence.

samcli/commands/local/cli_common/invoke_context.py
bnusunny: Only WARM mode uses `RefreshableSamFunctionProvider`, which re-reads the template on file changes via `_refresh_loaded_functions()`. That reload calls `get_stacks()` and needs to know the LE setting. COLD mode uses plain `SamFunctionProvider` which is initialized once from the already-expanded stacks and never reloads.

samcli/lib/cli_validation/image_repository_validation.py
bnusunny: Good catch. This is a deliberate decision documented in the plan (Task 13): inspection-path callers like `_is_all_image_funcs_provided` default to `False` because they run before the command context is fully set up (it's a Click decorator). Threading the flag here requires access to `ctx.params` which isn't straightforward in the decorator pattern. Filed as a follow-up — for now, the user gets the same pre-1.160.0 behavior (ForEach-generated Image functions aren't validated locally, CFN catches it at deploy time).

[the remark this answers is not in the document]
bnusunny: Acknowledged — same answer as above. The validator decorator doesn't have easy access to the resolved flag at the point it calls `get_stacks`. This is a known gap that'll be addressed in a follow-up (threading the flag from `ctx.params` in the decorator's `wrapped()` scope). For this PR, the behavior matches pre-1.160.0: ForEach-generated Image functions aren't seen by the validator.

samcli/lib/cfn_language_extensions/sam_integration.py
bnusunny: Agree it reads oddly. The current design prioritizes a single call-site pattern (one-liner, no conditional) and centralizes "what disabled means" in one place. Happy to rename to `process_language_extensions` in a follow-up if the team prefers that. Not blocking.

samcli/lib/cfn_language_extensions/sam_integration.py
bnusunny: Making it required (no default) is intentional — it forces every caller to explicitly decide whether LE is on or off. If we defaulted to `False`, a new call site could silently omit the arg and silently disable LE without the author realizing they need to wire it. With no default, forgetting to pass `enabled=` raises a `TypeError` at test time, which is how we caught all 15+ call sites during this PR.

Keyword-only (`*`) prevents accidental positional misuse — `expand_language_extensions(template, params, True)` would be unclear at the call site about what `True` means. `enabled=True` is self-documenting.

vicheey (approved):

samcli/lib/bootstrap/companion_stack/companion_stack_manager.py
```diff
@@ -307,7 +307,7 @@ def sync_ecr_stack(
     image_repositories = image_repositories.copy() if image_repositories else {}
     manager = CompanionStackManager(stack_name, region, s3_bucket, s3_prefix)
 
-    stacks = SamLocalStackProvider.get_stacks(template_file)[0]
+    stacks = SamLocalStackProvider.get_stacks(template_file, language_extensions_enabled=False)[0]
```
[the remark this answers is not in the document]
bnusunny: Acknowledged. Same pattern as the image_repository_validation gap — this is part of the same follow-up work. The callers do have the flag available, so threading it through sync_ecr_stack() is straightforward. Documenting as a known follow-up item alongside the validator fix.

valerena (approved):

tests/unit/commands/deploy/test_command.py
valerena: This answer didn't make sense to me so I searched it. I don't think this is because of any patching, and it's because 5.0 is the default value: https://github.com/aws/aws-sam-cli/blob/9ffa5ed806fc6931afaf4f0bbbbad19412b845d2/samcli/commands/deploy/command.py#L345-L347

Even though we define `5` as the default value, we convert it into string and then to float. Probably checking that something is `5` and something is `5.0` is the same in Python (because `5 == 5.0`?), so it doesn't really matter which one we use. It's true that it's more correct to do `5.0`, but it's just weird to change it in this PR.

PR #9055, opened 2026-06-02
ci: add PR size labeler with XXL gate

bnusunny:
## Summary

Adds `.github/workflows/pr-size-labeler.yml`, a new GitHub Actions workflow that:

- Labels every PR with one of `size/XS`, `size/S`, `size/M`, `size/L`, `size/XL`, `size/XXL` based on `additions + deletions` across non-excluded files.
- Excludes lockfiles, `schema/**`, snapshots, and image binaries from the count so generated noise doesn't dominate the bucket.
- **Fails the check** when the PR is `size/XXL` (≥ 1,000 changed lines), unless a maintainer applies `size/override`. Removing the override re-arms the gate.
- Posts a single sticky comment on XXL PRs explaining the threshold and the override path. No comments for other sizes.
- Tolerates transient GitHub API errors during sizing/labeling so flakiness never blocks PRs; only the intentional XXL-without-override failure propagates.

Built across six small commits — skeleton → fetch+compute → label management → gate+sticky comment → try/catch → docs.

## Action required from a maintainer before this lands

1. **Create the labels.** The workflow does not auto-create them. Top-of-file comment in the new file lists the exact `gh label create` commands; run them in this repo once.
2. **Add `Size label` to branch protection.** The gate only blocks merges if the check is required.

If the labels do not exist, `addLabels` returns 422; the script's catch swallows it and the gate is silently skipped — so the setup step is load-bearing.

## Behavior on each event

| Event          | What runs                                                              |
|----------------|------------------------------------------------------------------------|
| `opened`       | Compute size, label, post comment + fail if XXL without override.      |
| `synchronize`  | Same, on every push. Concurrency group cancels stale runs.             |
| `reopened`     | Same.                                                                  |
| `labeled`      | Recompute and re-evaluate gate (so adding `size/override` unblocks).   |
| `unlabeled`    | Recompute (so removing `size/override` re-arms the gate).              |

## Test plan

These smokes should be run on a fork before requiring this check in branch protection — `pull_request_target` runs the workflow file from the **base** ref, so the workflow needs to be present on a base branch first.

- [ ] Open a fork PR with > 1,000 non-excluded changed lines. Expect `size/XXL`, sticky comment, failing check. Apply `size/override` → check passes, comment updates (no second comment). Remove `size/override` → check fails again.
- [ ] Open a fork PR with ~50 source-line changes. Expect `size/M`, no comment, passing check. Push another commit that brings the total to ~130. Expect `size/M` removed and `size/L` applied.
- [ ] Open a fork PR that adds 2,000 lines to `package-lock.json` and 5 lines to a `.py` file. Expect `size/XS`.
- [ ] `actionlint .github/workflows/pr-size-labeler.yml` is clean.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -197,7 +197,6 @@ jobs:
         run: pytest -vv ${{ matrix.tests_config.params }}
         env:
           FORCE_RUN_DOCKER_TEST: ${{ matrix.tests_config.name == 'durable-functions' && '1' || '' }}
-          DURABLE_EXECUTIONS_EMULATOR_IMAGE_TAG: ${{ matrix.tests_config.name == 'durable-functions' && 'v1.1.1' || '' }}
 
   smoke-and-functional-tests:
     name: ${{ matrix.tests_config.name }} / ${{ matrix.tests_config.os }} / ${{ matrix.python }}
--- a/.github/workflows/pr-size-labeler.yml
+++ b/.github/workflows/pr-size-labeler.yml
@@ -132,3 +132,55 @@ jobs:
                 if (err.status !== 404) throw err;
               }
             }
+
+            // ---- Sticky comment helper. Identifies our comment by the
+            //      HTML marker on its first line. ----
+            async function upsertStickyComment(body) {
+              const fullBody = `${STICKY_MARKER}\n${body}`;
+              const comments = await github.paginate(github.rest.issues.listComments, {
+                owner,
+                repo,
+                issue_number: pull_number,
+                per_page: 100,
+              });
+              const existing = comments.find(c => c.body && c.body.startsWith(STICKY_MARKER));
+              if (existing) {
+                await github.rest.issues.updateComment({
+                  owner,
+                  repo,
+                  comment_id: existing.id,
+                  body: fullBody,
+                });
+              } else {
+                await github.rest.issues.createComment({
+                  owner,
+                  repo,
+                  issue_number: pull_number,
+                  body: fullBody,
+                });
+              }
+            }
+
+            // ---- XXL gate. ----
+            if (bucket.label === 'size/XXL') {
+              const overrideActive = currentLabels.includes(OVERRIDE_LABEL);
+              const overrideParagraph = overrideActive
+                ? `A maintainer applied \`${OVERRIDE_LABEL}\`; this PR is allowed despite its size. Remove the label to re-arm the gate.`
+                : `A maintainer can apply the \`${OVERRIDE_LABEL}\` label to bypass this check.`;
+              const body = [
+                `**PR size: \`XXL\` (${total} changed lines)**`,
+                ``,
+                `This PR exceeds the 1,000-line threshold. Smaller PRs get reviewed faster and are easier to revert.`,
+                ``,
+                overrideParagraph,
+                ``,
+                `_Excluded from the count: lockfiles, schema, snapshots, images. See \`.github/workflows/pr-size-labeler.yml\` to adjust._`,
+              ].join('\n');
+              await upsertStickyComment(body);
+
+              if (!overrideActive) {
+                core.setFailed(
+                  `PR has ${total} changed lines (>= 1000). Apply the \`${OVERRIDE_LABEL}\` label to bypass.`,
+                );
+              }
+            }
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
roger-zhangg, line 200: why this appears here, maybe rebase

.github/workflows/pr-size-labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+              '**/__snapshots__/**',
+              '**/*.snap',
+              '**/*.svg',
+              '**/*.png',
+              '**/*.jpg',
+              '**/*.jpeg',
+              '**/*.gif',
+              '**/*.ico',
+              '**/*.pdf',
+            ];
```
roger-zhangg, line 75: do we have any of these except `requirements`?

.github/workflows/pr-size-labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+
+            async function upsertStickyComment(body) {
+              const fullBody = `${STICKY_MARKER}\n${body}`;
+              const comments = await github.paginate(github.rest.issues.listComments, {
+                owner,
+                repo,
+                issue_number: pull_number,
+                per_page: 100,
+              });
+              const existing = comments.find(c => c.body && c.body.startsWith(STICKY_MARKER));
```
roger-zhangg, line 96: can we just create a check that success & fail on each commit so we can throw away all these find/update in place logic?

.github/workflows/pr-size-labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+# ONE-TIME SETUP (a maintainer must run these once after this workflow merges):
+#   gh label create size/XS       --color ededed --description "PR size: 0-9 changed lines"
+#   gh label create size/S        --color c2e0c6 --description "PR size: 10-29 changed lines"
+#   gh label create size/M        --color fef2c0 --description "PR size: 30-99 changed lines"
+#   gh label create size/L        --color fbca04 --description "PR size: 100-499 changed lines"
+#   gh label create size/XL       --color e99695 --description "PR size: 500-999 changed lines"
+#   gh label create size/XXL      --color b60205 --description "PR size: 1000+ changed lines"
+#   gh label create size/override --color 0e8a16 --description "Maintainer override for size/XXL gate"
+#
+# Spec: docs/superpowers/specs/2026-06-01-pr-size-labeler-design.md
```
roger-zhangg, line 16: non-relavant comment

.github/workflows/pr-size-labeler.yml
roger-zhangg: A more useful list would be ignore the dep/test/document changes. Could discuss

.github/workflows/pr-size-labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+              await core.summary
+                .addHeading('PR size')
+                .addRaw(`**${bucket.label}** — ${total} changed lines (excluded ${excludedFiles}/${files.length} files)${truncated ? ' _(file list was truncated)_' : ''}`)
+                .write();
+
+              currentLabels = (context.payload.pull_request.labels || []).map(l => l.name);
+              const stale = currentLabels.filter(n => SIZE_LABELS.includes(n) && n !== bucket.label);
+              const alreadyApplied = currentLabels.includes(bucket.label);
+
+              if (!alreadyApplied) {
```
roger-zhangg, line 150: I think the most useful part is when this feature gates huge PRs. But I don't think we care if its a `S` or `XL`. So I'd suggest we just focus on fail/success

.github/workflows/pr-size-labeler.yml
roger-zhangg: wait, this workflow itself is already a check. So we can just fail the workflow if it's more than 1000 useful lines

.github/workflows/pr-size-labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      - name: Compute size, label, and gate
+        uses: actions/github-script@v9
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
```
roger-zhangg, line 41: let's use a standalone script.

roger-zhangg:
Another part is we should update the contribution guild to call this 1000 line limit out & suggestion if they hit this blocker. https://github.com/aws/aws-sam-cli/blob/develop/CONTRIBUTING.md

Issue #9064, opened 2026-06-05
Bug: `AWS::Serverless::Api` gateway responses are not working


rnza0u:
Setting the `GatewayResponses` property does not work when using `local start-api` (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-gatewayresponses).

This issue from 2019 is probably related : https://github.com/aws/aws-sam-cli/issues/1027

Right now every error that originates from the Authorizer or from an unhandled lambda exception cannot have CORS headers (or any sort of customization).

Issue #9087, opened 2026-06-17
Bug: sam local start-api omits HTTP API CORS headers on Lambda authorizer denial


ageorgeh:
### Description:
When using a simple lambda authorizer with a `sam local start-api` CORS headers are only applied to successful authorizations

### Steps to reproduce:
Follow the steps in this repo's README https://github.com/ageorgeh/aws-sam-repro

### Observed result:
Also in the README

### Expected result:
CORS headers from the authorizer as with live AWS http API Gateways

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Linux - NixOS 
2. `sam --version`: 1.160.0
3. AWS region: ap-southeast-2

```
{
  "version": "1.160.0",
  "system": {
    "python": "3.13.13",
    "os": "Linux-6.18.33-x86_64-with-glibc2.42"
  },
  "additional_dependencies": {
    "container_engine": "Docker(v29.5.2)",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
    "SAM_CLI_BETA_PACKAGE_PERFORMANCE",
    "SAM_CLI_BETA_UV_PACKAGE_MANAGER"
  ]
}
```

ageorgeh:
(ai generated) patch im using if anyone needs a quick fix
```patch
--- a/samcli/local/apigw/local_apigw_service.py
+++ b/samcli/local/apigw/local_apigw_service.py
@@ -678,7 +678,9 @@
 
         # check for LambdaAuthorizer since that is the only authorizer we currently support
         if isinstance(lambda_authorizer, LambdaAuthorizer) and not self._valid_identity_sources(request, route):
-            return ServiceErrorResponses.missing_lambda_auth_identity_sources()
+            auth_failure_response = ServiceErrorResponses.missing_lambda_auth_identity_sources()
+            auth_failure_response.headers.update(cors_headers)
+            return auth_failure_response
 
         try:
             route_lambda_event = self._generate_lambda_event(request, route, method, endpoint)
@@ -732,6 +734,7 @@
                 #
                 # Note that returning within a finally block will have the effect of swallowing
                 # any reraised exceptions.
+                auth_service_error.headers.update(cors_headers)
[...]

PR #9114, opened 2026-07-06
fix(local-apigw): apply GatewayResponses headers to authorizer and Lambda-exception error responses

Hardikrepo:
fix(local-apigw): apply GatewayResponses headers to authorizer and Lambda-exception error responses

`sam local start-api` never read the `GatewayResponses` property on `AWS::Serverless::Api`, so responses generated by the local emulator itself (Lambda Authorizer denials, missing identity sources, unhandled Lambda exceptions) never carried any of the headers a customer configured there. Real API Gateway applies `GatewayResponses` (including CORS headers) to exactly these cases, so local behavior diverged from deployed behavior with no way to work around it.

**Root cause:** `SamApiProvider` extracted `Cors`/`Auth`/`StageName` from the `Api` resource but silently dropped `GatewayResponses`, and none of the `ServiceErrorResponses` factory methods accepted extra headers, so there was no path for configured headers to reach the Flask response object.

**Fix:**
- `SamApiProvider` now extracts `GatewayResponses`' `ResponseParameters.Headers` per response type (matching the same `gatewayresponse.header.X` / quoted-string schema `samtranslator` uses), stored on the `Api` model.
- `Api.get_gateway_response_headers(response_type, status_code)` resolves headers for a specific response type, falling back to `DEFAULT_4XX`/`DEFAULT_5XX` the same way API Gateway itself does when a specific type isn't configured.
- `ServiceErrorResponses` methods accept an optional `headers` dict and merge it onto the Flask response.
- `LocalApigwService` wires the `ACCESS_DENIED`, `UNAUTHORIZED`, and `DEFAULT_5XX` response types into the corresponding authorizer/lambda-failure error paths in `_request_handler`.

Only `ResponseParameters.Headers` is wired up locally for now (not `ResponseTemplates` body overrides or `StatusCode` overrides, and not applied to the 404/405 `MISSING_AUTHENTICATION_TOKEN` path) — scoped to the headers/CORS problem reported in the issue. Happy to extend if maintainers want broader coverage in this PR.

Verified with real Flask test-client requests (authorizer deny -> 403 with configured CORS header; unhandled Lambda exception -> 500 with configured header), plus unit tests for extraction, fallback resolution, and header propagation.

#9064

`GatewayResponses` is a documented SAM template property specifically for customizing API Gateway error responses (e.g. adding CORS headers to authorizer/4xx/5xx errors), but `sam local start-api` silently ignored it. This meant local testing could not reproduce a common, documented deployed-API behavior, and there was no local workaround for CORS-blocked error responses.

It parses `GatewayResponses` from the `AWS::Serverless::Api` resource the same way `Cors`/`Auth` are already parsed, resolves the correct response type (with `DEFAULT_4XX`/`DEFAULT_5XX` fallback, matching real API Gateway semantics) for each local error path, and merges the configured headers onto the Flask response returned by the local emulator.

None for existing users: if `GatewayResponses` isn't set on the template, `Api.gateway_responses` extraction returns an empty mapping and `get_gateway_response_headers` returns `{}`, so error responses are unchanged. The change only adds headers when `GatewayResponses` is explicitly configured.

#### Mandatory Checklist

- [ ] Write design document if needed ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document)) — not needed, this is a scoped bugfix to existing, documented behavior
- [x] Write/update unit tests
- [ ] Write/update integration tests — not added; existing local-apigw integration tests didn't cover GatewayResponses and I didn't have a Docker environment available to validate new ones end-to-end. Happy to add if maintainers want.
- [x] `make pr` passes — ran `black`, `ruff` (lint), and the full `tests/unit/local` + `tests/unit/commands/local` suites manually (no regressions; some pre-existing unrelated Windows-path failures exist on `develop` before this change too); did not run the full integration suite (`test-all`) due to no local Docker/AWS environment
- [ ] `make update-reproducible-reqs` if dependencies were changed — no dependency changes
- [ ] Write documentation — no new user-facing property/flag; `GatewayResponses` is already documented at the SAM template level

### DIFF
--- a/samcli/lib/providers/api_collector.py
+++ b/samcli/lib/providers/api_collector.py
@@ -34,6 +34,7 @@ class ApiCollector:
         self.stage_name: Optional[str] = None
         self.stage_variables: Optional[Dict] = None
         self.cors: Optional[Cors] = None
+        self.gateway_responses: Optional[Dict[str, Dict[str, str]]] = None
 
     def __iter__(self) -> Iterator[Tuple[str, List[Route]]]:
         """
@@ -200,6 +201,7 @@ class ApiCollector:
         api.stage_name = self.stage_name
         api.stage_variables = self.stage_variables
         api.cors = self.cors
+        api.gateway_responses = self.gateway_responses
 
         for authorizers in self._authorizers_per_resources.values():
             if len(authorizers):
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -499,6 +499,9 @@ class LayerVersion:
 
 
 class Api:
+    _HTTP_STATUS_CODE_4XX_LOWER_BOUND = 400
+    _HTTP_STATUS_CODE_4XX_UPPER_BOUND = 500
+
     def __init__(self, routes: Optional[Union[List["Route"], Set[str]]] = None) -> None:
         if routes is None:
             routes = []
@@ -517,6 +520,11 @@ class Api:
         self.stage_name: Optional[str] = None
         self.stage_variables: Optional[Dict] = None
 
+        # Headers configured through the API's GatewayResponses property, keyed by response type
+        # (e.g. "UNAUTHORIZED", "ACCESS_DENIED", "DEFAULT_5XX"). Only REST APIs (AWS::Serverless::Api)
+        # support GatewayResponses.
+        self.gateway_responses: Optional[Dict[str, Dict[str, str]]] = None
+
     def __hash__(self) -> int:
         # Other properties are not a part of the hash
         return hash(self.routes) * hash(self.cors) * hash(self.binary_media_types_set)
@@ -524,6 +532,32 @@ class Api:
     @property
     def binary_media_types(self) -> List[str]:
         return list(self.binary_media_types_set)
+
+    def get_gateway_response_headers(self, response_type: str, status_code: int) -> Dict[str, str]:
+        """
+        Returns the headers configured through this API's GatewayResponses property for the given
+        response type. If the specific type isn't configured, falls back to the generic DEFAULT_4XX
+        or DEFAULT_5XX response (matching the fallback behavior of API Gateway itself).
+
+        Parameters
+        ----------
+        response_type : str
+            The API Gateway response type this response corresponds to (e.g. "UNAUTHORIZED")
+        status_code : int
+            The HTTP status code being returned, used to select the DEFAULT_4XX/DEFAULT_5XX fallback
+
+        Returns
+        -------
+        dict
+            Headers configured for this response type, or an empty dict if none are configured
+        """
+        if not self.gateway_responses:
+            return {}
+
+        is_4xx = self._HTTP_STATUS_CODE_4XX_LOWER_BOUND <= status_code < self._HTTP_STATUS_CODE_4XX_UPPER_BOUND
+        default_response_type = "DEFAULT_4XX" if is_4xx else "DEFAULT_5XX"
+        headers = self.gateway_responses.get(response_type) or self.gateway_responses.get(default_response_type)
+        return dict(headers) if headers else {}
 
 
 _CorsTuple = namedtuple(
--- a/samcli/lib/providers/sam_api_provider.py
+++ b/samcli/lib/providers/sam_api_provider.py
@@ -140,6 +140,7 @@ class SamApiProvider(CfnBaseApiProvider):
         cors = self.extract_cors(properties.get("Cors", {}))
         stage_name = properties.get("StageName")
         stage_variables = properties.get("Variables")
+        gateway_responses = self._extract_gateway_responses(logical_id, properties.get("GatewayResponses", {}))
         if not body and not uri:
             # Swagger is not found anywhere.
             LOG.debug(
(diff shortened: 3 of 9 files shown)


roger-zhangg:
Thanks for the detailed write-up, and sorry for the long silence here. Threading `GatewayResponses` through `SamApiProvider` -> `ApiCollector` -> `Api` -> `LocalApigwService` the same way `Cors` already flows is the right shape for this codebase, and splitting extraction / resolution / application into three small pieces made this easy to review.

### What I verified locally (checked out `pr-9114` at bd991d5)

- `pytest tests/unit/local/apigw tests/unit/commands/local/lib/test_provider.py tests/unit/commands/local/lib/test_sam_api_provider.py -q` -> **487 passed**. `ruff check samcli/` and `black --check` are clean. `mypy` on the four changed source files reports only a pre-existing `samcli/lib/utils/graphql_api.py:27` error that is already present at your merge-base (8f25f26) and already fixed on `develop`, so it is not from this PR.
- The header-application layer works against real Flask objects, not just mocks: inside a request context, `ServiceErrorResponses.lambda_authorizer_unauthorized(headers={"Access-Control-Allow-Origin": "*"})` returns 403 with the header present.
- `Globals: Api: GatewayResponses` is picked up correctly (`provider.api.gateway_responses` populated) - worth calling out because that is the shape used in `tests/functional/commands/validate/lib/models/api_with_gateway_responses_implicit.yaml`.
- The quoted-string convention at `samcli/lib/providers/sam_api_provider.py:210-221` matches the existing `extract_cors` handling at `samcli/lib/providers/cfn_base_api_provider.py:145-149`, so that part is consistent with what is already here.

### Required changes

**1. `sam_api_provider.py:213-219` warns on valid templates.** Gateway-response header values are not limited to static quoted strings - mapping expressions such as `Access-Control-Allow-Origin: method.request.header.Origin` are legal. I ran `_extract_gateway_responses` against that input and it logs:

```
WARNING samcli.lib.providers.sam_api_provider: Ignoring GatewayResponses header
'Access-Control-Allow-Origin' for 'DEFAULT_4XX' on resource 'TestApi'. Header values
must be a quoted string (i.e. "'value'" is correct, but "value" is not).
```

That is a false positive on every `sam local start-api` startup for a template that deploys fine. Unresolved intrinsics (`{"Ref": ...}`) hit the same branch. Please drop these to `LOG.debug`, or reword so we are not telling users their working template is wrong.

**2. Only 2 of the 11 call sites pass a real API Gateway response type.** `local_apigw_service.py:682` (`UNAUTHORIZED`, 401) and `:705` (`ACCESS_DENIED`, 403) are correct. The other nine (`:694, :710, :754, :758, :763, :767, :771, :776, :796`) hardcode `DEFAULT_5XX`/`DEFAULT_4XX` as the *response type*. Since `Api.get_gateway_response_headers` only falls back from a specific type to the default and never the other way, a template that configures a more specific type instead of `DEFAULT_5XX` (`INTEGRATION_FAILURE`, `AUTHORIZER_FAILURE`, `AUTHORIZER_CONFIGURATION_ERROR`, ...) gets no headers locally even though the deployed API would apply them. It also makes the `status_code` argument dead at those nine sites, because it just re-derives the type that was already passed in. Passing the real type and letting `status_code` drive the fallback - exactly the pattern you already use at `:705` - fixes both. Worth double-checking which type each local failure path should map to.

**3. Test gap on the third path you wired.** `ACCESS_DENIED` and `DEFAULT_5XX` both have assertions (`test_lambda_auth_unauthorized_response_applies_gateway_response_headers`, `test_request_handler_applies_gateway_response_headers_on_unhandled_lambda_exception`), but the `UNAUTHORIZED` / missing-identity-source wiring at `local_apigw_service.py:681-683` has none: `tests/unit/local/apigw/test_local_apigw_service.py:653` exercises that branch but only asserts the returned object, never the `headers=` kwarg. Please add that assertion. A `DEFAULT_4XX`-fallback assertion at a call site would be good too - the fallback is currently only covered at the `Api` unit level.

**4. Please add an integration test.** `tests/integration/local/start_api/` already has the harness and a `lambda_authorizers/` fixture directory. This change is end-to-end wiring (template -> provider -> collector -> Flask response), which is precisely what unit tests that `@patch("...ServiceErrorResponses")` cannot catch. Happy to help if the Docker setup is the blocker.

### On the linked issues

- **Fixes #9064 for `AWS::Serverless::Api`.** Not for `AWS::ApiGateway::RestApi`: `cfn_api_provider.py` handles `RestApi`/`Method`/`Stage`/`Authorizer` but has no `AWS::ApiGateway::GatewayResponse` branch, and `_extract_gateway_responses` is a private static on `SamApiProvider` rather than living on the shared `CfnBaseApiProvider` next to `extract_cors`. Moving it to the base class now would make the CDK / plain-CFN follow-up cheap. Suggestion, not a blocker.
- **Does not fix #9087**, and I would not list it as such. That issue is `AWS::Serverless::HttpApi` + `CorsConfiguration`; HTTP APIs have no `GatewayResponses` property, `_extract_from_serverless_http` (`sam_api_provider.py:393`) is untouched here, and `cors_headers` are still only merged on the success path at `local_apigw_service.py:800`. #9087 needs a separate change and should stay open.
- The `MISSING_AUTHENTICATION_TOKEN` gap you flagged is fine as a follow-up, but note it is mechanically small: `route_not_found` is registered at `local_apigw_service.py:236,238`, where `self.api` is already in scope.

Overall: the structure does not need reworking, and I would like to get this in. Items 1-4 above are what I need before merge.

PR #9116, opened 2026-07-06
feat: add --express flag to sam delete

### LINKED ISSUE #9115: Feature request: add --express flag to sam delete
## Description

PR #9106 adds `--express/--no-express` support to `sam deploy` and `sam sync` for [CloudFormation Express mode](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-express-mode.html), but `sam delete` does not expose the same flag.

As noted in [PR #9106 comment](https://github.com/aws/aws-sam-cli/pull/9106#issuecomment-4893093479):

```
Usage: sam delete [OPTIONS]
Try 'sam delete -h' for help.

Error: No such option: --express
```

For teams using express mode during development, delete is part of the same iteration loop as deploy/sync. Without express support on `sam delete`, teardown remains on the slower standard path even when stacks were created or updated with `--express`.

## Use Case

During fast development iteration, developers want to tear down stacks quickly without waiting for full resource stabilization and cleanup. CloudFormation Express mode supports this for delete operations:

```bash
aws cloudformation delete-stack --stack-name my-stack \
    --deployment-config '{"Mode": "EXPRESS"}'
```

See the [CloudFormation Express mode documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-express-mode.html#cloudformation-express-mode-delete) and the [delete-stack API reference](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/delete-stack.html).

## Proposed Behavior

Add `--express/--no-express` to `sam delete`, consistent with `sam deploy` and `sam sync`:

- When enabled, pass `DeploymentConfig={"Mode": "EXPRESS"}` to the CloudFormation `DeleteStack` API
- Default: off (for backwards compatibility)
- Persist to `samconfig.toml` via `--save-params` as `express = true/false`
- After delete completes in express mode, surface a message that some resources may still be removing in the background (similar to deploy/sync)

## Current Workaround

Use the AWS CLI directly:

```bash
aws cloudformation delete-stack --stack-name <stack-name> \
    --deployment-config '{"Mode": "EXPRESS"}'
```

Note: this bypasses `sam delete`'s guided cleanup of S3 artifacts, ECR images, and companion stacks.

## Related

- #9106 (adds `--express` to `sam deploy` and `sam sync`)
- #9110 (announcement for express mode support in deploy/sync)

koteshyelamati:
Closes #9115

#9115

#9106 added `--express/--no-express` support to `sam deploy` and `sam sync` for CloudFormation Express mode, but `sam delete` does not expose the same flag. For teams using express mode during development, delete is part of the same iteration loop: without express support on `sam delete`, teardown stays on the slower standard path even when the stack was created with `--express`.

Mirrors the deploy/sync implementation for the delete path:

- `sam delete` gains `--express/--no-express` (default: off, backwards compatible). The flag is persisted to `samconfig.toml` via the existing `--save-params` machinery, consistent with deploy/sync.
- When enabled, `DeleteContext` passes `DeploymentConfig={"Mode": "EXPRESS"}` to every CloudFormation `DeleteStack` call (main stack, ECR companion stack, and the `retain_resources` retry paths).
- `CfnUtils.delete_stack` accepts an optional `deployment_config` and only includes `DeploymentConfig` in the boto call when provided, so non-express behaviour is byte-for-byte unchanged.
- After a successful express delete, a yellow notice is printed that some resources may still be removing in the background — consistent with the deploy/sync messaging from #9106.
- The `express` option is registered in the delete help formatter (`core/options.py`).

The guided cleanup of S3 artifacts, ECR images, and companion stacks is unchanged — express mode only affects the CloudFormation stack deletion calls, which is what the current `aws cloudformation delete-stack --deployment-config` workaround bypasses.

None when the flag is not used: `DeleteStack` is called with exactly the same arguments as before.

#### Checklist

- [x] Write unit tests: flag plumbing from `do_cli` to `DeleteContext`, and `CfnUtils.delete_stack` with/without `DeploymentConfig` (all `tests/unit/commands/delete` and `tests/unit/lib/delete` pass)


roger-zhangg:
Thanks for the contribution @koteshyelamati ! The change LGTM, I added integ test & schema update.

roger-zhangg (approved):

tobixlea (approved):

KyleDoherty1:
Hi folks @roger-zhangg , would it be possible to get a new release of the CLI out with this added?

koteshyelamati:
Thank you @roger-zhangg and @tobixlea for the thorough review and for adding the integration tests and schema update! Really appreciate it.

PR #9127, opened 2026-07-14
feat: detect AI agent invocations in telemetry

madhavdonthula1:
Add an AgentDetector that identifies when SAM CLI is invoked by an AI coding agent by reading environment variables the agent sets. Records it via the existing userAgent telemetry field (emitted as "<agent>/1.0"). This lets us distinguish agent-driven usage from human and CI/CD usage.

Detected agents and their environmental variable signals:
  - Claude Code: CLAUDECODE present
  - Codex: any CODEX_* var present
  - Cursor: CURSOR_AGENT present
  - Gemini CLI: GEMINI_CLI present
  - Kiro: TERM_PROGRAM == kiro, or AWS_EXECUTION_ENV contains amazonq/kiro
  - OpenCode: OPENCODE present
  - GitHub Copilot: COPILOT_AGENT_SESSION_ID present, or GITHUB_COPILOT_CLI_MODE == "true"

Shared variables (AWS_EXECUTION_ENV, TERM_PROGRAM) are substring or exact-value matched, never presence-checked, to avoid misattributing Lambda/CloudShell/CodeBuild or generic VS Code environments as agents. GitHub Copilot detection deliberately avoids any GITHUB_* / GITHUB_ACTION key so it never false-matches ordinary GitHub Actions CI, which the existing CICDDetector already handles.

An existing AWS_TOOLING_USER_AGENT (AWS toolkit) takes precedence over a detected agent. Amazon Q is intentionally deferred; until it is added after Kiro in resolution order, Amazon Q CLI runs resolve to kiro.

Includes unit tests for each agent, the exact-name/too-generic guards (GEMINI_CLI, OPENCODE vs AGENT), the AWS_EXECUTION_ENV and GITHUB_ACTION false-positive guards, and toolkit-takes-precedence behavior.


roger-zhangg (dismissed):

samcli/lib/telemetry/agent_detector.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _is_github_copilot(environ: Mapping) -> bool:
+    """
+    Whether it is running in the standalone GitHub Copilot CLI (not Copilot-in-VS-Code
+    or the cloud agent), via COPILOT_AGENT_SESSION_ID (a per-session UUID from the
+    @github/copilot bundle; undocumented, so re-verify per release) or
+    GITHUB_COPILOT_CLI_MODE=="true". It must NEVER key on a CI variable: Copilot's
+    cloud agent runs inside GitHub Actions, which cicd.py already detects via
+    GITHUB_ACTION, so any GITHUB_* rule would false-match ordinary CI.
```
reedham-aws: nit: I find this comment a little confusing, mostly because it's a little jargony. It's not clear to me what we should "re-verify" and what the importance of the CI variable is.

samcli/lib/telemetry/agent_detector.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    return "COPILOT_AGENT_SESSION_ID" in environ or environ.get("GITHUB_COPILOT_CLI_MODE", "") == "true"
```
reedham-aws: Where does the `GITHUB_COPILOT_CLI_MODE` envvar come from? I couldn't find it in any documentation.

samcli/lib/telemetry/agent_detector.py
```diff
@@ -0,0 +1,116 @@
...
+    # https://docs.anthropic.com/en/docs/claude-code
+    Agent.ClaudeCode: "CLAUDECODE",
+    # https://developers.openai.com/codex/cli
+    Agent.Codex: _is_codex,
+    # match presence, not value: the value is not guaranteed stable
+    Agent.Cursor: "CURSOR_AGENT",
+    # exact name, not a prefix, so GEMINI_CLI_HOME / GEMINI_CLI_NO_RELAUNCH don't match
+    Agent.GeminiCLI: "GEMINI_CLI",
+    Agent.Kiro: _is_kiro,
+    # OpenCode sets OPENCODE=1 in its root CLI middleware
```
reedham-aws, line 60: I'm curious where you found this, I only see it when directly examining the code, but couldn't find it documented anywhere.

samcli/lib/telemetry/user_agent.py
```diff
@@ -16,7 +39,42 @@
...
+         and JetBrains toolkits) and always wins when present, so a toolkit
+         invocation is still attributed to the toolkit even if it happens to run
+         inside an AI agent.
+      2. A detected AI agent (e.g. Claude Code), emitted as "<agent-name>/1.0".
+         This is used only as a fallback when the toolkit user-agent above is not
+         set to a valid value.
+    """
+    toolkit_user_agent = os.environ.get(USER_AGENT_ENV_VAR, "").strip()
+    if toolkit_user_agent and ACCEPTED_USER_AGENT_FORMAT.match(toolkit_user_agent):
+        return toolkit_user_agent
```
reedham-aws: nit: I think we can leave this as `user_agent` for the variable name, just because it's not necessarily guaranteed to to be the toolkit.

tests/unit/lib/telemetry/test_agent_detector.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @patch.dict(
+        "samcli.lib.telemetry.agent_detector.os.environ",
+        {"GITHUB_ACTION": "run1"},
+        clear=True,
+    )
+    def test_detector_does_not_identify_github_actions_as_copilot(self):
+        # Plain GitHub Actions CI (no Copilot session var) must never be attributed to
+        # GitHub Copilot; agent detection returns None so cicd.py can own GITHUB_ACTION.
+        self.assertIsNone(AgentDetector().agent())
```
reedham-aws: Should this also set the `COPILOT_AGENT_SESSION_ID` to confirm that it gets ignored?

reedham-aws (dismissed):
I'm approving because I think it works as is, but I'd like to also hear your answers to the questions I've left in comments. Overall, great work!

samcli/lib/telemetry/agent_detector.py
madhavdonthula1: Good catch. I think it would be worth it to rewrite the doc string so that it does not have a lot of confusing function names and key words. We should also drop the `GITHUB_COPILOT_CLI_MODE` as it is not needed anymore.

samcli/lib/telemetry/agent_detector.py
madhavdonthula1: The `GITHUB_COPILOT_CLI_MODE` does exist, but the CLI does not assign it into a spawned process's env,  so it wouldn't be present when Copilot invokes SAM. Hence, it would not be useful for us so i'll go ahead and delete it and only detect on `COPILOT_AGENT_SESSION_ID` 

If you'd like you can see more about `GITHUB_COPILOT_CLI_MODE` in the published bundle: [github link](https://unpkg.com/@github/copilot-darwin-arm64@1.0.70/app.js).

samcli/lib/telemetry/agent_detector.py
madhavdonthula1: Good catch, it's not documented anywhere official I also found it by directly reading OpenCode's source. It's set in the root CLI middleware here: [github link](https://github.com/sst/opencode/blob/v1.17.20/packages/opencode/src/index.ts )

madhavdonthula1 (commented):
New minor changes that removed:
-  `GITHUB_COPILOT_CLI_MODE == "true"`
- test cases

madhavdonthula1 (commented):
Added antigravity

reedham-aws (approved):

licjun (approved):

samcli/lib/telemetry/agent_detector.py
```diff
@@ -0,0 +1,107 @@
...
+    is shared with AWS Lambda/CloudShell/CodeBuild (e.g. "AWS_Lambda_python3.12"), so
+    it is substring-matched on "amazonq"/"kiro", not checked for presence, to avoid
+    misattributing those environments.
+
+    NOTE: the "amazonq" substring also matches the genuine Amazon Q Developer CLI.
+    Amazon Q is deferred for now, so its CLI runs are attributed to kiro until a
+    dedicated Amazon Q agent is added (which must be ordered AFTER Kiro in the enum).
+    """
+    if environ.get("TERM_PROGRAM", "") == "kiro":
+        return True
```
valerena, line 43: Kiro IDE also adds `TERM_PROGRAM=kiro` to any regular terminal that's opened from the IDE (not only for its agent).

Vercel also checks if it's not an interactive session before saying "it's Kiro", so maybe we can do that.  https://github.com/vercel/detect-agent/blob/48532d9/agents.json#L194-L201

samcli/lib/telemetry/agent_detector.py
```diff
@@ -0,0 +1,107 @@
...
+    aws_execution_env: str = environ.get("AWS_EXECUTION_ENV", "").lower()
+    return "amazonq" in aws_execution_env or "kiro" in aws_execution_env
+
+
+_ENV_VAR_OR_CALLABLE_BY_AGENT: Dict[Agent, Union[str, Callable[[Mapping], bool]]] = {
```
valerena, line 48: This `Union` type kind of hurts my eyes. 

We could define a structure `{ "env_var": <value>, "function":<value>}` where we can put one or the other, or have a function only here, and have a function that returns a function, so you can do 
```
    # https://docs.anthropic.com/en/docs/claude-code
    Agent.ClaudeCode: has_envvar("CLAUDECODE"),
    # https://developers.openai.com/codex/cli
    Agent.Codex: _is_codex,
    # match presence, not value: the value is not guaranteed stable
    Agent.Cursor: has_envvar("CURSOR_AGENT"),
    # exact name, not a prefix, so GEMINI
...


with `has_envvar` something like
```
def has_envvar(name: str):
    return lambda environ: name in environ
```

Or I don't know. something that's nicer than mixing different types and then having to do a `isinstance` down the line.

PR #9136, opened 2026-07-20
feat: add --output json flag (sam build)

madhavdonthula1:
## Before
  ```
  Build Succeeded

  Built Artifacts  : .aws-sam/build
  Built Template   : .aws-sam/build/template.yaml

  Commands you can use next
  =========================
  [*] Validate SAM template: sam validate
  ...
  ```
  Error output is similarly unstructured:
  ```
  Build Failed
  Error: PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: nonexistent-package==1.0.0
  ```

  Note: The error message does not identify which function failed.

  ## After

  **Success output:**
  ```json
  {
    "status": "success",
    "build_dir": ".aws-sam/build",
    "template_file": ".aws-sam/build/template.yaml",
    "resources": [
      {"resource_id": "OrderProcessorFunction", "type": "function", "runtime": "python3.12", "architecture": "x86_64"},
      {"resource_id": "SharedDepsLayer", "type": "layer", "compatible_runtimes": ["python3.12"]}
    ]
  }
  ```

  **Failure output** (note the `resource` field — the text error today does not identify which function failed):
  ```json
  {
    "status": "failure",
    "error": {
      "type": "WorkflowFailedError",
      "message": "PythonPipBuilder:ResolveDependencies - Could not satisfy the requirement: nonexistent-package==1.0.0",
      "resource": "PaymentHandlerFunction"
    }
  }
  ```

  ## File Changes

  ### `samcli/commands/build/command.py`
-   Adds the `--output` Click option accepting `"text"` (default) or `"json"`. Passes the value through `cli()` → `do_cli()` `BuildContext`.

  ### `samcli/commands/build/build_context.py`
  - Store the output format as `self._output`
  - On success: emit a structured JSON object (status, build_dir, template_file, resources list with both functions and layers) instead of the colored "Build Succeeded" banner
  - On failure: emit a JSON error object (type, message, resource) then raise `UserException` to preserve telemetry tracking
  - Text mode follows the exact same code paths as before

  ### `samcli/commands/build/core/options.py`
  Registers `"output"` in `BUILD_STRATEGY_OPTIONS` so it appears in `--help`.

  ### `samcli/lib/build/build_strategy.py`
- Wraps `DefaultBuildStrategy.build_single_function_definition` and `build_single_layer_definition` with try/except to tag 
- `resource_name` on any build exception. Catches `BuildError`, `UnsupportedRuntimeException`, `BuildInsideContainerError`, and `UnsupportedBuilderLibraryVersionError` — covering ZIP, Image, container, and layer build paths uniformly.

  ### `samcli/lib/build/exceptions.py`
 - Adds an optional `resource_name` field to `BuildError`. Defaults to `None`; existing raisers are unaffected.

  ### Benchmark Notes

  Adding `--output json` to `sam build` does not dramatically change agent performance metrics. Tokens, tool calls, and time are roughly equivalent between text and JSON for this command. This is expected: `sam build` is a relatively simple command with clear, short output, and LLM agents parse both formats effectively. We implemented it to keep scope consistent across the project (which adds `--output json` to all major SAM CLI commands), and because the structured error output with the `resource` field provides explicit failure attribution that text mode lacks.

  | Metric | Text Output | JSON Output |
  |--------|-------------|-------------|
  | Tokens (avg per task) | ~28,800 | ~28,400 |
  | Tool calls (error diagnosis) | 5 | 4 |
  | Time (avg) | ~180s | ~165s |
  | Error attribution | Inferred from log line ordering | Explicit `resource` field |
  | Programmatic parsing | Regex/sed required | `json.loads()` directly |

  ## Screenshots
  Screenshot 2026-07-20 at 12 46 40 PM
  Screenshot 2026-07-20 at 12 47 18 PM
  Screenshot 2026-07-20 at 12 47 45 PM


madhavdonthula1:
> ## Code Review Results
> Reviewed: [56510b1](https://github.com/aws/aws-sam-cli/commit/56510b1a4d0a11d3f931103d900dd0fb533ce457)..[54aed55](https://github.com/aws/aws-sam-cli/commit/54aed55e919a013036ffcd2416b711ba683bdd07) Files: 5 Comments: 3
> 
> **Comments on lines outside the diff:**
> 
> **[samcli/commands/build/build_context.py:381]** **[BUG]** In JSON mode both error handlers call `sys.exit(1)` instead of raising `UserException`. `SystemExit` inherits from `BaseException`, not `Exception`, so the `@track_command` decorator on the build CLI (see `samcli/lib/telemetry/metric.py`, which only catches `(UserException, click.Abort, ...)` and `Exception`) will not catch it. That means:
> 
> * `_send_command_run_metrics` is never called for failed `sam build --output json` runs, so `exit_reason` and `exit_code` for these failures are silently dropped from telemetry.
> * The exit path diverges from every other `sam` command, which uniformly funnels errors through `UserException`.
> 
> Consider raising `UserException` after emitting the JSON (the top-level Click framework already turns it into exit code 1), or at minimum route through `click_ctx.exit(1)` after ensuring telemetry has been flushed. A minimal fix:
> 
> ```python
> if self._output == "json":
>    click.echo(json.dumps(error_result, indent=2))
>    # still raise so track_command records the failure;
>    # suppress the default click error output separately if needed
>    raise UserException(str(ex), wrapped_from=wrapped_from) from ex
> ```
> 
> The same issue applies to the `FunctionNotFound` handler around line 381 and the multi-exception handler around line 407.


Changed so that we raise `UserException` now instead of exiting using sys.exit(1)

madhavdonthula1 (commented):
- Replace sys.exit(1) with raise UserException in error handlers to
  preserve telemetry tracking via @track_command decorator
- Wrap both container and in-process build paths in try/except so
  resource_name is tagged on BuildError regardless of build mode
- Include layers in JSON success output alongside functions
- Rename field from logical_id to resource_id (accurate for nested stacks)
- Add type field ("function" or "layer") to each resource entry

madhavdonthula1 (commented):
broaden exception catch to add more build error types

madhavdonthula1 (commented):
- Apply black formatting to command.py and build_strategy.py
- Fix ruff import ordering in build_strategy.py
- Add type: ignore comments for mypy union-attr and index errors
  (dynamic resource_name assignment on non-BuildError exceptions)
- Regenerate schema/samcli.json to include the new --output parameter

samcli/commands/build/core/options.py
```diff
@@ -30,7 +30,7 @@
 
 EXTENSION_OPTIONS: List[str] = ["hook_name", "skip_prepare_infra"]
 
-BUILD_STRATEGY_OPTIONS: List[str] = ["parallel", "exclude", "manifest", "cached", "build_in_source"]
+BUILD_STRATEGY_OPTIONS: List[str] = ["parallel", "exclude", "manifest", "cached", "build_in_source", "output"]
```
reedham-aws: I think it doesn't make sense to have `output` as a build strategy option, these are usually reserved for changing how the build is actually done. Taking a look through the file, it seems like the groups are groupings for display, it might make more sense to have a new group that is like `OUTPUT_OPTIONS`. If you think there's another way that makes sense I'm open to that too.

samcli/commands/build/command.py
```diff
@@ -129,6 +129,12 @@
 @template_option_without_build
 @parameter_override_option
 @docker_common_options
+@click.option(
+    "--output",
+    default="text",
+    help="Output the results from the command in a given output format. " "Supported formats: text (default), json.",
+    type=click.Choice(["text", "json"], case_sensitive=False),
+)
```
reedham-aws: ~~I think it makes more sense to have this as a shared decorator, like `@manifest_option` or `@cached_option` for example. Since we will be doing this for other commands, no need to duplicate code.~~

I believe there already is a `--output` option: https://github.com/aws/aws-sam-cli/blob/fe8f8dbb1160abce8ae13e0a4c5f8b456658afd3/samcli/commands/_utils/options.py?plain=1#L415-L447

samcli/commands/build/build_context.py
```diff
@@ -336,17 +337,51 @@ def run(self) -> None:
...
-                click.secho(msg, fg="yellow")
+            if self._output == "json":
+                resources = [
+                    {
+                        "resource_id": f.full_path,
+                        "type": "function",
+                        "runtime": f.runtime,
+                        "architecture": f.architectures[0] if f.architectures else None,
+                    }
+                    for f in self.get_resources_to_build().functions
```
reedham-aws: `self.get_resources_to_build()` can be called once at the beginning of this block, and then reused here and when we get the layers.

samcli/commands/build/build_context.py
```diff
@@ -336,17 +337,51 @@ def run(self) -> None:
...
+                        "type": "layer",
+                        "compatible_runtimes": layer.compatible_runtimes,
+                    }
+                    for layer in self.get_resources_to_build().layers
+                ]
+                result = {
+                    "status": "success",
+                    "build_dir": build_dir_in_success_message,
+                    "template_file": output_template_path_in_success_message,
+                    "resources": resources,
```
reedham-aws: One thing is that in the old message, we had something like:
```
Commands you can use next
=========================
[*] Validate SAM template: sam validate
[*] Invoke Function: sam local invoke
[*] Deploy: sam deploy --guided
```

Which I think is actually kind of useful for agents. I wonder if we can simplify and put that into the result? Might be doing too much and I don't really think it's to important, so up to you/other reviewers.

samcli/commands/build/build_context.py
reedham-aws, line 389: There are 6 different exceptions here, but looking in `build_strategy.py` shows that we only potentially add resource name to 4 of them. Why is that?

samcli/lib/build/exceptions.py
```diff
@@ -15,8 +17,9 @@ def __init__(self, container_name: str, error_msg: str) -> None:
 
 
 class BuildError(Exception):
-    def __init__(self, wrapped_from: str, msg: str) -> None:
+    def __init__(self, wrapped_from: str, msg: str, resource_name: Optional[str] = None) -> None:
         self.wrapped_from = wrapped_from
+        self.resource_name = resource_name
         Exception.__init__(self, msg)
```
reedham-aws, line 26: Is there a reason we're not adding a resource_name field to other error types that could exist?

samcli/commands/build/build_context.py
```diff
@@ -360,7 +395,22 @@ def run(self) -> None:
         ) as ex:
             caught_exception = ex
 
-            click.secho("\nBuild Failed", fg="red")
+            if self._output == "json":
+                deep_wrap = getattr(ex, "wrapped_from", None)
```
reedham-aws: This line also gets run outside the else, we can probably extract it and put it at the beginning of the except.

samcli/commands/build/build_context.py
```diff
@@ -336,17 +337,51 @@ def run(self) -> None:
...
+                    click.secho(msg, fg="yellow")
         except FunctionNotFound as function_not_found_ex:
             caught_exception = function_not_found_ex
 
+            if self._output == "json":
+                error_result = {
+                    "status": "failure",
+                    "error": {
+                        "type": "FunctionNotFound",
+                        "message": str(function_not_found_ex),
```
reedham-aws: I feel like it's not good to inconsistently apply the `resource` field. Is there a way that you can included the missing resource here even if it's not real?

samcli/commands/build/build_context.py
```diff
@@ -336,17 +337,51 @@ def run(self) -> None:
...
 
-            if self._print_success_message:
-                msg = self._gen_success_msg(
-                    build_dir_in_success_message,
-                    output_template_path_in_success_message,
-                    os.path.abspath(self.build_dir) == os.path.abspath(DEFAULT_BUILD_DIR),
-                )
-
-                click.secho(msg, fg="yellow")
+            if self._output == "json":
```
reedham-aws: I would prefer that this whole block about printing gets moved to a method just to keep the `BuildContext.run` simple.

reedham-aws (commented):
Overall, we need to add new tests for this output behavior (meaning, with `json` enabled). I think at a base level we need unit tests, not sure if we really need integration tests since the underlying behavior is not changing.

samcli/commands/build/core/options.py
madhavdonthula1: This is smart. I went ahead and created a new group called `OUTPUT_OPTIONS` because none of the other groups made sense to mix with `output`. I'll also make sure to make this change in sam deploy and sam init to ensure we have the same structure.

samcli/commands/build/build_context.py
madhavdonthula1: Good catch! After looking into it I realized the `run()` method increased by 50 lines by just adding the json output code blocks. I just made two separate methods `_print_build_success` and `_print_build_failure` that I placed under the '=`_gen_success_msg` so that the output-related code can stay together. I think this should improve readability and will make sure to keep this in mind when developing further.

samcli/commands/build/build_context.py
madhavdonthula1: Yep just changed it so that they all get use the `resources_to_build = self.get_resources_to_build` to make it more simplistic.

samcli/commands/build/build_context.py
madhavdonthula1: Yeah this is a valid point and I actually explored it previously as a possible extension, but for `sam deploy` output. I tested three different types of extensions including one where `next_steps` was the main feature and while checking the benchmarks I noticed it did not improve the agent experience and in fact caused to add more tools called. The agent already has the inputs. It knows which flags it passed, and it knows the standard SAM commands. But to be honest I have not tested this with `sam build` so I'm open to doing that if anyone still finds that as something that would work specifically for `sam build` because theres definitely a lot more next steps for `sam build`.
  | Scenario | With extension | Without |
  |---|---|---|
  | Edge-layer 403 debug (`defined_routes`, Opus) | 13 calls / 363s | 10 calls / 215s |
  | Handler-mismatch debug (`function_readiness`, Opus) | 27 calls / 461s | 13 calls / 240s |
  | Happy-path deploy (Opus) | 6 calls | 7 calls |
  | IAM-permission debug (Opus) | 18 calls | 18 calls |

samcli/commands/build/build_context.py
madhavdonthula1: This is valid. I changed it so that it rather populates the resource or says null so that we dont have to be inconsistent.

samcli/commands/build/build_context.py
madhavdonthula1: One of the exceptions that we do not add resource_name extension to is`InvalidBuildGraphException` which has no resource to find.  It is raised in`build_graph.py:648` when a build definition has zero functions in it, so there is nothing to name. `get_full_path()` reads `self.functions[0]`, which would fail on the same empty list that caused the exception in the first place.

The other one is `ResourceNotFound`. That is raised in `build_context.py:1361` inside `collect_build_resources()`, which runs before the build starts. So `build_strategy.py` is never entered and adding it to that tuple would be dead code. It also means no resource matched the identifier, so there is no build definition to read a name from. 

Honestly I should've listed this information in the PR and will start doing that whenever I come across something that has differing cases.

samcli/commands/build/build_context.py
madhavdonthula1: Yep I changed it for simplicity!

samcli/commands/build/command.py
madhavdonthula1: Good call. I pulled it into a shared `@output_option` in `_utils/options.py`, following the same `*_click_option()` / `*_option()` pattern as`metadata_option`. `build/command.py` is just `@output_option` now, and deploy/init will reuse it.

  On the existing one: it's bundled inside `common_observability_click_options()` with `--start-time` and `--end-time`, so applying it to `build` would drag in two time flags that don't apply. It also has no default (value comes through as `None`) and types on `OutputOption` from the observability module, which felt like unnecessary coupling for a plain text/json switch.

`remote invoke` has its own third `--output` too, so there are three declarations floating around. Unifying them seems worthwhile but would change behavior in `logs`/`traces`/`remote invoke`, so I left it out of this PR. Happy to do it as a follow-up.

samcli/lib/build/exceptions.py
madhavdonthula1: Yep this is valid. I dropped it. The initial thought was that a raise site might want to pass the resource in at construction, so I made it a proper field on`BuildError`. But in practice nothing does: `build_strategy.py` sets `ex.resource_name` at runtime for all four exception types, so the declared param was never actually used.

madhavdonthula1 (commented):
Changes made based on comments

samcli/commands/build/build_context.py
reedham-aws: I'd be curious to know what tools are actually being called after the `sam deploy`. What were you recommending the agent do next? Some of those might have been worthwhile so not sure if raw number of calls being higher is necessarily bad. 

That being said, I think you're right that the agent probably knows enough to do those anyway. We can leave it.

reedham-aws (commented):
Looks good to me now, but we still need tests that actually touch the behavior.

reedham-aws (dismissed):

samcli/commands/_utils/options.py
```diff
@@ -447,6 +447,20 @@ def common_observability_options(f):
     return f
 
 
+def output_click_option():
```
roger-zhangg: These are the same function names as the existing `output_click_option` / `output_option` in `samcli/commands/list/cli_common/options.py`, but with different choices (`text|json` vs `json|table`) and a different default. Two identically-named `output_option` symbols is an import-confusion trap.

We now have four `--output` conventions: `json|table` (list), the `OutputOption` enum (observability), `RemoteInvokeOutputFormat` (remote invoke), and this one.

Since this project is adding `--output json` across all commands, worth settling here since every following PR will copy it. Ideal end state is one shared contract, a full JSON view for agents and text for humans, which is what `OutputOption` (`text`/`json`) already encodes; `list`'s `json|table` is the outlier and can converge later. Minimum ask: don't shadow the existing name.

samcli/commands/build/build_context.py
```diff
@@ -1076,6 +1078,89 @@ def _copy_artifact_paths(self, original_resource: Dict, modified_resource: Dict)
...
+        Parameters
+        ----------
+        artifacts_dir: str
+            A string path representing the folder of built artifacts
+        output_template_path: str
+            A string path representing the final template file
+        resources_to_build: ResourcesToBuildCollector
+            The functions and layers that were built
+        """
+        if self._output == "json":
```
roger-zhangg: `output` is threaded through as a `str` and compared against the `"json"` literal here and again at line 1148, where it's negated (`!= "json"`). A typo in either silently selects the wrong mode and no test would catch it. `click.Choice` only validates the CLI entry point; `BuildContext` is also constructed directly in tests and by other callers.

`samcli/lib/observability/util.py` already defines the enum, and `samcli/commands/logs/command.py:186` shows the conversion pattern:

```python
# command.py -- convert once at the boundary
output=OutputOption(output)

# build_context.py -- typed, and both call sites become positive checks
if self._output is OutputOption.json:
```

That drops the string literals from this file and removes the negated comparison on the failure path.

samcli/commands/build/build_context.py
```diff
@@ -1076,6 +1078,89 @@ def _copy_artifact_paths(self, original_resource: Dict, modified_resource: Dict)
...
+            The error type to report
+        print_text_banner: bool
+            Whether to print the "Build Failed" banner in text mode
+        """
+        if self._output != "json":
+            if print_text_banner:
+                click.secho("\nBuild Failed", fg="red")
+            return
+
+        # The resource key is always present so consumers can rely on the schema. It is
```
roger-zhangg: The "always present" guarantee doesn't quite hold. `InvalidBuildDirException` is a `UserException`, not a `BuildError` subclass, so it's caught by neither handler in `run()` — `sam build --output json` with a bad `--build-dir` emits no JSON at all (empty stdout, exit 1).

Either widen the catch or soften the comment so consumers don't rely on a schema that isn't guaranteed.

samcli/commands/build/build_context.py
```diff
@@ -1076,6 +1078,89 @@ def _copy_artifact_paths(self, original_resource: Dict, modified_resource: Dict)
...
+        # populated from the exception when the failure is attributable to a specific
+        # resource, otherwise from the resource the customer asked to build (which is the
+        # relevant name even when that resource does not exist in the template). It is
+        # None when the failure is not tied to any single resource.
+        resource_name = getattr(ex, "resource_name", None) or self._resource_identifier
+
+        error: Dict[str, Any] = {"type": error_type, "message": str(ex), "resource": resource_name}
+
+        click.echo(json.dumps({"status": "failure", "error": error}, indent=2))
```
roger-zhangg: Minor: `indent=2` pretty-prints. Given the stated motivation is agent consumption, single-line JSON is often preferred for machine parsing. Intentional?

samcli/lib/build/build_strategy.py
```diff
@@ -144,6 +150,22 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
...
+        try:
+            return self._do_build_single_function_definition(build_definition)
+        except (
+            BuildError,
+            UnsupportedRuntimeException,
+            BuildInsideContainerError,
+            UnsupportedBuilderLibraryVersionError,
+        ) as ex:
+            if getattr(ex, "resource_name", None) is None:
+                ex.resource_name = build_definition.get_full_path()  # type: ignore[union-attr]
```
roger-zhangg: The PR description says `samcli/lib/build/exceptions.py` "adds an optional `resource_name` field to `BuildError`", but that file isn't in the diff — so the attribute is being attached dynamically here (hence the `# type: ignore[union-attr]`).

This works today only because none of these exceptions define `__slots__`. Adding one later, a routine perf change, would break failure attribution silently at runtime in an error path.

Worth noting the four caught types share no base but `Exception` (`UnsupportedRuntimeException`, `BuildInsideContainerError` and `UnsupportedBuilderLibraryVersionError` are not `BuildError` subclasses), so the fix as described would only cover one of four. A class-level default on each keeps constructors untouched and lets mypy check it:

```python
class BuildError(Exception):
    resource_name: Optional[str] = None
```

Was `exceptions.py` meant to be included in this PR?

tests/unit/commands/buildcmd/test_build_context.py
```diff
@@ -1411,6 +1412,264 @@ def test_gen_message_with_non_default_build_with_hook_package(self):
         self.assertEqual(msg, expected_msg)
 
 
+class TestBuildContext_print_build_success(TestCase):
```
roger-zhangg, line 1520: These tests call `_print_build_success` / `_print_build_failure` directly, so the user-facing contract isn't covered: that `run()` in JSON mode emits exactly one parseable document on stdout and nothing else. One test doing `json.loads(stdout)` through `run()` would cover more than the ~15 helper tests here — and would have caught the `InvalidBuildDirException` gap noted in `build_context.py`.

schema/samcli.json
```diff
@@ -433,6 +433,16 @@
                   "type": "string",
                   "description": "Name or ID of an existing docker network for AWS Lambda docker containers to connect to, along with the default bridge network. If not specified, the Lambda containers will only connect to the default bridge docker network."
                 },
+                "output": {
```
roger-zhangg, line 436: Is this file generated? I couldn't find a `make schema` target that regenerates it, but if one exists this should be regenerated rather than hand-edited.

samcli/commands/build/build_context.py
```diff
@@ -1076,6 +1078,89 @@ def _copy_artifact_paths(self, original_resource: Dict, modified_resource: Dict)
...
+        resources_to_build: ResourcesToBuildCollector
+            The functions and layers that were built
+        """
+        if self._output == "json":
+            resources: List[Dict[str, Any]] = [
+                {
+                    "resource_id": function.full_path,
+                    "type": "function",
+                    "runtime": function.runtime,
+                    "architecture": function.architectures[0] if function.architectures else None,
```
roger-zhangg: This reports `"architecture": null` for the common case of a template that doesn't set `Architectures`, even though the artifact was built for `x86_64`.

`Function.architecture` (`samcli/lib/providers/provider.py:194`) already resolves this, defaulting to `X86_64` when the property is absent — and it's the value the build actually used, since `ApplicationBuilder._get_build_graph()` passes `function.architecture` into `FunctionBuildDefinition` (`app_builder.py:293`).

So `function.architectures[0] if function.architectures else None` diverges from what was built in two ways: `null` instead of `x86_64` when absent, and silently reporting the first element of a multi-element list that `Function.architecture` would reject as invalid. A consumer using this to select a Lambda architecture or container image gets the wrong answer for any template that omits `Architectures`.

```python
"architecture": function.architecture,
```

Safe to call here — the success path is only reached after `_get_build_graph()` already evaluated `function.architecture` without raising.

Note `test_json_architecture_defaults_to_none_when_absent` and `test_json_multi_architecture_reports_first` currently assert the present behaviour, so they'd need updating too.

samcli/commands/build/build_context.py
```diff
@@ -286,12 +291,12 @@ def run(self) -> None:
                     # if self._mount_with is NOT WRITE
                     # check the need of mounting with write permissions and prompt user to enable it if needed
                     mount_with_write = prompt_user_to_enable_mount_with_write_if_needed(
```
roger-zhangg, line 317: Two interactive prompts in the build path write to **stdout**, so they interleave with the JSON document and make it unparseable:

- `prompt_user_to_enable_mount_with_write_if_needed` → `click.confirm(...)` at `samcli/commands/build/utils.py:106`, no `err=` (defaults to `False`). Triggered by `sam build --use-container` for any workflow with `must_mount_with_write_in_container` — which includes all the dotnet ones.
- `_check_build_method_experimental_flag()` (line 320) → `prompt_experimental` → `click.confirm(Colored().yellow(prompt), default=False)` at `samcli/commands/_utils/experimental.py:275`.

Confirmed `click.confirm()` without `err=True` writes the prompt text to stdout, and it does so before the confirm aborts in a non-interactive environment, so stdout is corrupted either way.

The rest of the CLI already keeps human-facing output off stdout — `LOG` handlers go to stderr, and both the telemetry prompt (`samcli/cli/main.py:161`) and the update notice (`samcli/lib/utils/version_checker.py:94-96`) pass `err=True`. Following that here keeps stdout reserved for the JSON:

```python
if click.confirm(
    f"\nBuilding functions with {config.language} inside containers needs "
    ...
    err=True,
):
```

Worth a test that runs `--use-container --output json` and asserts stdout is still parseable.

samcli/commands/_utils/options.py
madhavdonthula1: Good catch about the name collision. We've renamed our helpers to `structured_output_click_option` / `structured_output_option`. This is the shared contract for all commands adopting `text | json` going forward (deploy, init, etc) will import from the same helper.

We considered reusing the existing `--output option` from `common_observability_click_options()` (line 431), but that one is bundled with `--start-time` and `--end-time` as a group for `sam logs/sam traces`, has no explicit default. It can't be used standalone. The list module's `output_option (json|table)` has different choices and a different default, so it's not compatible either.

samcli/commands/build/build_context.py
madhavdonthula1: Fixed. We now convert the string to `OutputOption(output)` once in `__init__` and use is `OutputOption.json` for all comparisons

samcli/commands/build/build_context.py
madhavdonthula1: Fixed both the gap and the comment. Added a handler for `InvalidBuildDirException` in `do_cli()` that emits the JSON error before re-raising. It's thrown during `BuildContext.__enter__()` before `run()` is even called, so it can't be caught inside `run()`.

Also softened the comment, it no longer claims "always present so consumers can rely on the schema." It now says the resource key is included when available and may be None for failures not tied to a single resource (invalid build dir, template parse errors).

samcli/commands/build/build_context.py
madhavdonthula1: Changed. removed indent=2 from all `json.dumps()` calls. Output is now single-line for easier machine parsing.

samcli/commands/build/build_context.py
madhavdonthula1: Fixed. Changed to function.architecture, which resolves the `x86_64` default and matches the value the build actually used `(via _get_build_graph)`. And updated the two test cases

samcli/commands/build/build_context.py
madhavdonthula1: Fixed using your suggested approach added `err=True` to both `click.confirm()` calls (`prompt_user_to_enable_mount_with_write_if_needed`) in build/utils.py and `prompt_experimental` in `_utils/experimental.py`. This keeps the prompt behavior intact but routes the text to `stderr`, matching the telemetry prompt and update notice. `stdout` stays reserved for JSON. Updated the one test that asserted the old call signature.

samcli/lib/build/build_strategy.py
madhavdonthula1: Yes, `exceptions.py` got lost in a rebase, now included. Applied your suggestion to all four caught types (not just `BuildError`) added `resource_name: Optional[str] = None` as a class-level default to `BuildError`, `BuildInsideContainerError`, `UnsupportedBuilderLibraryVersionError` in `exceptions.py` and `UnsupportedRuntimeException` in `workflow_config.py`. Constructors remain untouched, and both `# type: ignore[union-attr]` comments are gone since mypy can now verify the attribute.

tests/unit/commands/buildcmd/test_build_context.py
madhavdonthula1: Added `test_run_json_mode_emits_single_parseable_document`. It runs `run()` in JSON mode, captures stdout with `redirect_stdout`, and does `json.loads()` on it. Any stray output (prompt, log line, extra echo) makes `json.loads` fail, so it covers the real contract. Kept the helper tests since they cover individual field shapes cheaply.

samcli/lib/build/build_strategy.py
```diff
@@ -210,6 +233,23 @@ def build_single_layer_definition(self, layer_definition: LayerBuildDefinition)
         """
         Build the unique definition and then copy the artifact to the corresponding layer folder
         """
+        try:
```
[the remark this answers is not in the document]
madhavdonthula1: This isn't needed, the `MissingBuildMethodException` raised at`build_strategy.py:256` is imported from `samcli.lib.build.exceptions`, where it's a `BuildErro` subclass (not the `UserException` one). Since `BuildError` is already in the `layer wrapper's` except tuple, that exception is already caught and gets `resource_name = layer_definition.full_path`adding it explicitly would be redundant with the parent catch. The `"resource": null` case you're describing is actually the other `MissingBuildMethodException`. The `UserException` in `samcli.commands.build.exceptions`, raised during `get_resources_to_build() ` before the strategy runs. Verified end-to-end: `sam build MyLayer` `--output json` on a `BuildMethod`-less layer reports `"resource": "MyLayer"`.

samcli/commands/_utils/experimental.py
```diff
@@ -272,7 +272,7 @@ def prompt_experimental(
     if is_experimental_enabled(config_entry):
         update_experimental_context()
         return True
-    confirmed = click.confirm(Colored().yellow(prompt), default=False)
+    confirmed = click.confirm(Colored().yellow(prompt), default=False, err=True)
```
reedham-aws, line 275: I get we're sending this to stderr, but it's not clear to me why and in what scenario that's useful.

samcli/commands/build/command.py
```diff
@@ -230,50 +234,72 @@ def do_cli(  # pylint: disable=too-many-locals, too-many-statements
...
+        # execution error surfaced from BuildContext, including bare-Exception template errors
+        # raised during __enter__/set_up (e.g. InvalidLayerReference, RemoteStackLocationNotSupported)
+        # that are not UserException subclasses. We re-raise unconditionally so @track_command still
+        # records telemetry and wraps non-UserException errors as UnhandledException, and so run()'s
+        # text "Build Failed" banner path is unaffected (no double-emit: run() no longer emits JSON).
+        #
+        # Scope: this covers failures once do_cli runs. Errors raised earlier during click option
+        # processing (invalid flags, or a --hook-name prepare-hook failure re-raised by
+        # track_command) surface as click's standard usage/stderr output, not JSON. Consumers should
+        # treat "exit != 0 with empty stdout" as an invocation error, not an execution failure.
```
reedham-aws, line 299: I find this comment way too hard to read and understand. Cutting it down would help and probably would trigger less AI review.

samcli/commands/build/build_context.py
reedham-aws: Is this not a problem with the `--output text` as well?

samcli/lib/build/build_strategy.py
```diff
@@ -144,6 +150,23 @@ def build_single_function_definition(self, build_definition: FunctionBuildDefini
...
+            return self._do_build_single_function_definition(build_definition)
+        except (
+            BuildError,
+            UnsupportedRuntimeException,
+            UnsupportedBuilderException,
+            BuildInsideContainerError,
+            UnsupportedBuilderLibraryVersionError,
+        ) as ex:
+            if getattr(ex, "resource_name", None) is None:
+                ex.resource_name = build_definition.get_full_path()
```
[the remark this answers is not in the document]
reedham-aws: Can we make `resource_name` a list? I agree with this comment and that preserves the json.

PR #9151, opened 2026-07-31
tests: resolve sam init menu positions dynamically in schemas tests

roger-zhangg:

N/A

`test_init_interactive_with_event_bridge_app_aws_schemas_python` is failing ([run 30657529336](https://github.com/aws/aws-sam-cli/actions/runs/30657529336/job/91277548826)).

The test answered the runtime prompt with `8` and its comment said `# 8: Python 3.9`. But the prompt offers **15** runtimes for the *Infrastructure event management* use case, and position 8 is now `java8.al2023` — `python3.9` has drifted to position **10**:

```
 1. go (provided.al2)   6. java11.al2023   11. python3.14
 2. java25              7. java11          12. python3.13
 3. java21              8. java8.al2023 <- test picked this
 4. java17.al2023       9. java8.al2       13. python3.12
 5. java17             10. python3.9 <- wanted    ...
```

So the test selected a Java runtime and then fed Python-flow answers (`eb-app-python39`, schema paginator choices) into a Java prompt sequence. Every later answer landed on the wrong question, and it finally asserted on `hello_world_function/schema` — a Python layout a Java project never produces.

The positions moved when `java8.al2023` / `java11.al2023` / `java17.al2023` were added (#9125).

Adds `_get_runtime_position` and `_get_use_case_position`, which resolve positions from the same source the prompt uses (`InitTemplates().get_preprocessed_manifest(...)` + `get_sorted_runtimes`) rather than hardcoding an index. This mirrors the existing `_get_registry_position` helper already in the file.

All 8 tests now resolve template/use-case/runtime answers dynamically. **The 4 Java tests and the go test were relying on the same brittle indexing and only passed by luck** — a future runtime addition would have broken them the same way.

Resolved positions are unchanged for Java (4) and go (1), and corrected from 8 → 10 for `python3.9`. Stale comments now name the runtime instead of a number.

##### Changes from review feedback

**Runtimes are now named constants** (`JAVA_RUNTIME_FOR_INIT`, `PYTHON_RUNTIME_FOR_INIT`, `GO_RUNTIME_FOR_INIT`). None of these tests is about a specific runtime version — they exercise the schemas flows — so the version was incidental and repeated 4/3/1 times. `python3.9` is already deprecated, so a one-line change rather than a hunt through the file is not hypothetical. The per-test comments now name the constant rather than the literal version, so changing a constant cannot leave a stale comment behind — the same drift this PR exists to fix.

Scoped to this file deliberately: the runtimes in `test_init_command.py` are passed as explicit `--runtime` CLI arguments, which are position-independent and fail loudly if removed, and each of those tests deliberately pairs a runtime with a dependency manager (`java21`+maven, `nodejs18.x`+npm). Hoisting those into shared constants would obscure what they exercise.

**The manifest is now fetched once per process** (`lru_cache(maxsize=1)`). Both helpers called `get_preprocessed_manifest()` independently, and `InitTemplates._get_manifest` does `requests.get(MANIFEST_URL, timeout=10)` on every call with a clone fallback — two fetches per test. It was also a correctness trap: resolved independently, one call could succeed against `MANIFEST_URL` while the other fell back to the bundled local manifest, giving a use-case position and a runtime position from two different snapshots, which is the exact misalignment these helpers exist to prevent. Measured 1 miss / 3 hits for four resolutions.

**Registry answers.** The go test's comment said `# 4: select aws.events as registries` but it answered `1`, while every sibling resolves it via `_get_registry_position("aws.events")`; it now resolves it too. `..._non_default_profile_selection` stays hardcoded on purpose, and now says why: it drives a non-default profile and an explicit `us-east-1`, so the prompt lists that profile/region's registries, while `_get_registry_position` resolves against the default `Session()` — converting it would query the wrong account and region.

**Removed a stray answer.** That same test carried an extra `3` after the project name — 14 answers against 13 documented prompts. The sibling `..._aws_schemas_python` is identical up to that point and answers `Y` immediately after the name, so the next prompt is the AWS-config confirm, not a numbered choice. Confirmed against click directly:

```
Use default AWS profile? [y/N]: 3
Error: invalid input
Use default AWS profile? [y/N]: N     <- re-prompt consumes the next line
```

The confirm rejected `3`, re-prompted, and ate the `N` intended for it, so the test passed only because click's re-prompt loop swallowed the extra line — the same off-by-one fragility this PR removes, and it made the input contradict its own comment block. Now 13 answers matching 13 documented prompts.

The helpers call `get_preprocessed_manifest`, which these tests already exercise via `sam init` itself, so no new external dependency. If a runtime is ever removed from the manifest the helper raises with the available list, which fails loudly instead of silently selecting the wrong runtime.

Note the same job also hit a separate pytest-xdist crash (`struct.error: 'i' format requires ... <= 2147483647` — a worker tried to send a >2 GiB test report, killing the session and taking `..._non_default_profile_selection` with it). That is very likely a *consequence* of the misaligned answers producing runaway prompt output, so this fix should resolve it, but the 2 GiB report limit is a separate robustness issue not addressed here.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation


tests/integration/init/schemas/test_init_with_schemas_command.py
```diff
@@ -56,8 +90,8 @@ def test_init_interactive_with_event_bridge_app_aws_registry(self):
         aws_registry_pos = _get_registry_position("aws.events")
         user_input = f"""
 1
-8
-4
+{_get_use_case_position(EVENT_BRIDGE_USE_CASE)}
+{_get_runtime_position("java17.al2023")}
```
valerena: This might be a good time to also put this as a constant somewhere... `JAVA_RUNTIME_FOR_INIT` and `PYTHON_RUNTIME_FOR_INIT` (or something like that?)

roger-zhangg: Thanks for the suggestion, fixed in 

https://github.com/aws/aws-sam-cli/pull/9151/changes#diff-54461e8d3ba983d84c58e6ab98cc5e1d2f9963e6dc0262a0f7a3057b06a46dddR31-R33

tests/integration/init/schemas/test_init_with_schemas_command.py
```diff
@@ -34,15 +38,45 @@ def _get_registry_position(registry_name):
...
+def _get_runtime_position(runtime_name):
+    """Return the 1-based menu position of a runtime in the `sam init` runtime prompt.
+
+    The prompt lists the runtimes the app-templates manifest offers for
+    EVENT_BRIDGE_USE_CASE, ordered by `get_sorted_runtimes`. That ordering shifts
+    whenever a runtime is added or removed, so hardcoding a position silently
+    starts selecting a different runtime -- every answer after it then lands on
+    the wrong question. Resolve it the same way `_get_registry_position` does.
+    """
+    runtime_options = InitTemplates().get_preprocessed_manifest(None, None, None, None)[EVENT_BRIDGE_USE_CASE]
```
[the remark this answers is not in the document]
roger-zhangg: fixed

reedham-aws (approved):

valerena (approved):

PR #9172, opened 2026-08-07
feat: output json integration

madhavdonthula1:
# Add `--output json` to sam build and sam deploy

## Overview

- This adds a `--output json` flag to both sam build and sam deploy.
- The flag defaults to text, so every existing command behaves exactly as before.
- JSON output is fully opt in and never turns on by itself.
- CI jobs and agents can now read results as structured data instead of scraping text.
- Both commands share one flag decorator and one `OutputOption` enum.

## Important clarification on output shape

- sam build prints exactly one JSON document at the end of the run.
- sam deploy prints a stream of JSON lines as the deploy progresses.
- Each deploy line is its own object with a `type` field.
- The deploy types are changeset, event, outputs, warning, info, and result.
- Streaming is intentional because a deploy runs for a long time.
- A single document would force a consumer to wait until the very end.

## Shared result contract (both commands)

- Every terminal object has `type: "result"` and a lowercase `status`.
- Failures use `status: "failure"` with `error: {type, message}`.
- So a consumer can handle build and deploy failures with one code path.
- build additionally lists `error.resources` (the affected resources), which is additive.

---

### `samcli/commands/build/command.py`
- The command front door that Click runs first.
- Adds the shared `--output` flag and threads it into do_cli.
- Catches any build failure in one place and prints it as JSON.
- Re-raises the error so telemetry and exit codes are unchanged.

### `samcli/commands/build/build_context.py`
- The worker that orchestrates the build.
- Prints the JSON success document listing every built resource.
- Reports build_dir and template_file as absolute paths in JSON.
- The relpath transform is kept only for the human text banner.
- Reports all affected resources as a list in error.resources.
- Resolves a bare resource id to its full path for a matching namespace.
- Re-raises a filesystem error as InvalidBuildDirException, not a raw OSError.
- Skips the beta python-uv confirm prompt in JSON mode.
- Fails with an actionable error if a JSON build needs a writable mount.

### `samcli/commands/build/utils.py`
- Extracts the write-mount detection out of the interactive prompt.
- Lets the build context detect the .NET write-mount requirement without prompting.
- Routes the mount confirm prompt to stderr in text mode.

### `samcli/lib/build/build_strategy.py`
- Tags build failures with the resources they affect.
- Uses one shared context manager so every strategy attributes uniformly.
- Also covers the cached and incremental build paths.

### `samcli/lib/build/exceptions.py` and `samcli/lib/build/workflow_config.py`
- Add a `resource_names` list attribute to the build exception classes.
- Declaring it on the class also avoids mypy errors.

### `samcli/commands/build/core/options.py`
- Registers the output option so it shows under Output Options in help.

---

### `samcli/commands/deploy/command.py`
- The deploy front door that Click runs.
- Uses the same shared `--output` flag as build.
- Rejects `--guided` and `--confirm-changeset` with JSON up front.
- Wraps package and deploy so any failure ends the stream with a result line.
- Redirects manage_stack progress to devnull under `--resolve-s3` in JSON mode.
- Re-surfaces the resolved bucket as a `{type: info}` line.
- Forces the package progress bar off in JSON mode.

### `samcli/commands/deploy/deploy_context.py`
- Orchestrates the deploy and decides what to emit.
- Parses the flag once into the OutputOption enum.
- Suppresses the human deploy-args table and banners in JSON mode.
- Emits terminal result lines for success, no_changes, and changeset_created.
- The success line carries express, changeset_id, stack_name, and region.
- Emits per-resource auth warnings as `{type: warning}` lines.

### `samcli/lib/deploy/deployer.py`
- The layer that calls CloudFormation directly.
- Stores output mode as the OutputOption enum, raising on a bad value.
- Emits changeset, event, and outputs as JSON lines.
- Event lines carry detailed_status so failure detail is not lost.
- Datetimes are serialized with isoformat.
- Status messages like "Waiting..." are text-only.

### `samcli/commands/deploy/core/options.py`
- Registers the output option for the deploy help screen.

---

### `samcli/commands/_utils/options.py`
- Holds the single shared `--output` flag decorator.
- The accepted values are derived from the OutputOption enum, so they cannot drift.

### `samcli/commands/_utils/table_print.py`
- The shared table decorator skips the table chrome in JSON mode.
- Raises on an unrecognized output mode instead of corrupting the stream.

### `samcli/commands/_utils/cdk_support_decorators.py`
- Routes the CDK-unsupported advisory to stderr so stdout stays clean.

### `samcli/commands/package/package_context.py`
- Routes the preview-runtime warning to stderr in JSON mode.

### `samcli/lib/package/language_extensions_packaging.py`
- Routes the parameter-collection warning to stderr unconditionally.

### `samcli/commands/exceptions.py`
- Adds the shared `resource_names` attribute to the base UserException.

---

## Testing

- Unit tests were added for the new JSON success, failure, and streaming paths.
- Tests cover build attribution, cached builds, and deploy JSON output.
- Key behaviors were mutation checked by breaking the code on purpose.
- `make pr` passes and coverage stays above the required bar.


samcli/commands/deploy/deploy_context.py
```diff
@@ -128,6 +134,8 @@ def run(self):
         """
         Execute deployment based on the argument provided by customers and samconfig.toml.
         """
+        if self._output_mode is OutputOption.json:
```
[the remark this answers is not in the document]
madhavdonthula1: Fixed in cdbf1a4. Removed the logger downgrade; SAM log handlers already write to stderr, so --debug works in JSON mode without touching stdout.

samcli/commands/package/package_context.py
```diff
@@ -285,8 +288,9 @@ def _export_with_language_extensions(self, template_path, original_template_dict
             )
         return output_template
 
-    @staticmethod
-    def _warn_preview_runtime(stacks: List[Stack]) -> None:
+    def _warn_preview_runtime(self, stacks: List[Stack]) -> None:
+        # In JSON mode this warning goes to stderr so it does not corrupt the JSON lines on stdout.
+        to_stderr = self._output_mode is OutputOption.json
```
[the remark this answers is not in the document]
madhavdonthula1: Fixed in f6b48d9. The preview-runtime warning now routes to stderr in JSON  mode so it cannot corrupt the stream.

reedham-aws (commented):
I only reviewed the build part, but that looks pretty good to me from a logic point of view. I do think some of the inline comments are kind of long and hard to read, so maybe that could use some cleanup. Sorry to be nitpicky on the comments, overall, great work! I will review the deploy part later.

samcli/commands/deploy/command.py
```diff
@@ -295,109 +302,151 @@ def do_cli(
...
-            )
+    if guided and output_mode is OutputOption.json:
+        raise click.UsageError("--guided is not compatible with --output json")
+
+    if confirm_changeset and output_mode is OutputOption.json:
+        # Confirming a changeset needs an interactive answer a JSON consumer cannot give. Rejecting
+        # up front avoids a silent no-op deploy that would otherwise emit CONFIRMATION_REQUIRED and
+        # exit 0 without deploying. confirm_changeset is a persisted samconfig value from --guided,
+        # so this combination is easy to hit when a guided project is later deployed in CI.
+        raise click.UsageError("--confirm-changeset is not compatible with --output json")
```
reedham-aws, line 319: I think it's fine and not worth pursuing too hard, but intuitively, interactive flags and `--output json` should not be incompatible.

samcli/commands/deploy/command.py
```diff
@@ -295,109 +302,151 @@ def do_cli(
...
+            if resolve_s3:
+                if bool(s3_bucket):
+                    raise DeployResolveS3AndS3SetError()
+                if output_mode is OutputOption.json:
+                    # manage_stack prints progress to stdout; redirect it so it does not corrupt the
+                    # JSON stream. Then surface the resolved bucket as a JSON info line, since
+                    # print_managed_s3_bucket_info (the only place it appears in text mode) is skipped.
+                    with open(os.devnull, "w") as devnull, contextlib.redirect_stdout(devnull):
+                        s3_bucket = manage_stack(profile=profile, region=region)
+                    click.echo(json.dumps({"type": "info", "managed_s3_bucket": s3_bucket}))
```
reedham-aws, line 356: One thing that concerns me is that the `deploy` command prints multiple lines of `JSON`. I think this could throw off the parser if it's not expecting it. Related to that, I realize we might need to better define the structure of the JSON so that an agent has a chance to read it and understand what it should parse for. I don't think we should try to complete this in this PR, but definitely a follow up.

samcli/commands/deploy/deploy_context.py
```diff
@@ -155,25 +159,31 @@ def run(self):
...
+                self.s3_bucket,
+                self.image_repositories if isinstance(self.image_repositories, dict) else self.image_repository,
+                region,
+                self.capabilities,
+                display_parameter_overrides,
+                self.confirm_changeset,
+                self.signing_profiles,
+                self.use_changeset,
+                self.disable_rollback,
+            )
```
reedham-aws, line 185: Should we be printing this in json?

samcli/commands/deploy/deploy_context.py
```diff
@@ -254,7 +264,13 @@ def deploy(
 
         for resource, authorization_required in auth_required_per_resource:
             if not authorization_required:
-                click.secho(f"{resource} has no authentication.", fg="yellow")
+                if self._output_mode is OutputOption.json:
+                    sys.stdout.write(
+                        json.dumps({"type": "warning", "message": "no authentication", "resource": resource}) + "\n"
+                    )
+                    sys.stdout.flush()
```
reedham-aws: Is there a reason we're using `sys.stdout.write` instead of `click.echo`?

samcli/commands/_utils/template.py
```diff
@@ -453,6 +453,22 @@ def get_template_artifacts_format(template_file):
...
+    parse error raised there would escape click before do_cli's try/except and, in --output json
+    mode, print plain text to stderr with no JSON. Returning [] lets the requirement checks treat the
+    format as undeterminable and the real error resurface in do_cli, where it gets serialized.
+    """
+    try:
+        return get_template_artifacts_format(template_file=template_file)
+    except (TemplateNotFoundException, TemplateFailedParsingException):
+        return []
+
+
```
reedham-aws: I'm worried that adding this changes the default behavior as well. I also am not sure I think it's necessary as if the CLI is failing before `do_cli`, then the non-zero exit code and lack of print to `stdout` should be enough for the agent to know somethings wrong.

Happy to hear an explanation and keep it as well.

samcli/lib/cli_validation/image_repository_validation.py
```diff
@@ -121,12 +125,18 @@ def _is_all_image_funcs_provided(template_file, image_repositories, parameters_o
...
+            template_file,
+            parameter_overrides=parameters_overrides,
+            global_parameter_overrides=global_parameter_overrides,
+            language_extensions_enabled=False,
+        )
+    except (TemplateNotFoundException, TemplateFailedParsingException):
+        # This validator runs before the command body. If the template can't be parsed here, skip
+        # the check (report "all provided") so the parse error resurfaces in do_cli, where --output
+        # json can serialize it, instead of escaping to stderr as plain text.
+        return True
```
reedham-aws: Same question I raised for `get_template_artifacts_format_or_empty`

samcli/commands/deploy/command.py
madhavdonthula1: Agree it feels a bit unintuitive. The reason I reject them instead of silently ignoring them is that both need an interactive answer a JSON consumer can't give, and confirm-changeset especially is there so a human reviews the changeset before it deploys. If we just dropped it in json mode we'd emit CONFIRMATION_REQUIRED and exit 0 without deploying, which is a silent no-op that's easy to hit since confirm-changeset gets persisted into samconfig by guided. Erroring up front felt safer than deploying or no-opping without the user realizing. Happy to revisit if you'd rather have json imply non-interactive.

samcli/commands/deploy/deploy_context.py
madhavdonthula1: Right, this is intentionally skipped in json mode. `print_deploy_args` is a human formatted table, tabbed columns and yellow section headers printed with `click.echo`, so putting it on `stdout` would corrupt the `NDJSON` stream that a consumer parses line by line. The values it shows are inputs the caller already has or are surfaced structurally elsewhere in the stream, like the resolved bucket coming back as a `managed_s3_bucket info line`, so nothing a json consumer needs is actually lost. If you'd want any of those fields as a structured line I can add it, but I don't think the table itself doesn't belong on the `json` stream.

samcli/commands/_utils/template.py
madhavdonthula1: Good points. On changing default behavior, you're right and I fixed it. do_cli now parses the template as its first step, so a bad template raises before any side effect like manage_stack, which makes text mode identical to before, and json mode catches the same error and serializes it. I added tests for both. On whether it's necessary, I'd like to keep it because in our benchmarking a nonzero exit with empty stdout was exactly what agents couldn't read, since they couldn't tell "failed" from "still running." The deferral is what lets the error reach do_cli so stdout always carries a machine-readable failure result, which is the whole point of output json.

samcli/commands/deploy/deploy_context.py
madhavdonthula1: This is for consistency with the rest of the json stream. Every line in the NDJSON output uses `sys.stdout.write(...) + "\n"` followed by an explicit flush, both here in `deploy_context` and across `deployer.py`, so this warning line matches that pattern. Raw write also avoids click's extra terminal handling on what's meant to be a machine readable stream. I kept it uniform on purpose, but if you'd prefer `click.echo` I'd want to switch all of them together rather than just this one.

samcli/commands/deploy/command.py
```diff
@@ -295,109 +302,149 @@ def do_cli(
...
+        if output_mode is OutputOption.json:
+            click.echo(
+                json.dumps(
+                    {
+                        "type": "result",
+                        "status": "failure",
+                        # resources is always present (null here) to match build's failure shape, so a
+                        # consumer can read error.resources uniformly across commands. Deploy failures
+                        # are not attributed to specific resources, so it is null.
+                        "error": {"type": type(ex).__name__, "message": str(ex), "resources": None},
```
reedham-aws: One final concern I have here is that in build, we have a dedicated helper in `build_context.py` that creates the error: https://github.com/madhavdonthula1/aws-sam-cli/blob/74f97c042f4eb1842f5b8d4a60888b36cadcdf72/samcli/commands/build/build_context.py#L78-L97
Here, we do it directly in the function. The result is that they're slightly different as this one does not use `"wrapped_from"`.

Ideally, they would share a function for this as the types are all pretty generic.

reedham-aws (commented):
This is the only comment I have remaining, nice work!

samcli/commands/deploy/command.py
madhavdonthula1: Share a single failure_result_json helper between build and deploy. build had `build_failure_json` while deploy inlined its own dict, and they'd drifted (deploy lacked `wrapped_from` and hardcoded resources: `null`). Extract `failure_result_json` in `observability/util.py`; `build_failure_json` delegates to it and deploy calls it directly, so both emit the same failure shape.

reedham-aws (approved):

roger-zhangg (approved):
LGTM, let's merge and see if integration test picks up anything needs fixing.

reedham-aws:
Merge failed on flaky integ tests.

PR #9184, opened 2026-08-15
fix: sam local invoke leaks Docker container and temp directory on OOM

### LINKED ISSUE #9182: sam local invoke leaks Docker container and temp directory when function is OOM-killed
### Description

When a function invoked via `sam local invoke` (or `sam local start-api`/`start-lambda`) exceeds its `MemorySize` and is OOM-killed, the resulting `ContainerFailureError` is raised *before* the container is stopped/removed and before the per-invocation decompressed-code temp directory is cleaned up. Both are leaked.

### Root cause

```python
# samcli/local/lambdafn/runtime.py:350-362
def _on_invoke_done(self, container):
    if container:
        self._check_exit_state(container)      # <-- raises ContainerFailureError on OOM
        self._container_manager.stop(container) # <-- never reached
    self._clean_decompressed_paths()             # <-- never reached
```

`_on_invoke_done` is invoked unconditionally from `invoke()`'s `finally` block:

```python
# samcli/local/lambdafn/runtime.py:343-346
finally:
    # We will be done with execution, if either the execution completed or an interrupt was fired
    # Any case, cleanup the container.
    self._on_invoke_done(container)
```

`_on_invoke_done` has no try/except of its own, so when `_check_exit_state()` raises `ContainerFailureError` on OOM (`runtime.py:364-382`), that exception propagates immediately — skipping both `self._container_manager.stop(container)` and `self._clean_decompressed_paths()` on the same line and below it.

### Steps to reproduce

1. Create a function with a low `MemorySize` (e.g. 128) that allocates memory past the limit in its handler.
2. Run `sam local invoke` against it.
3. Observe the invocation fails with a memory error as expected, but:
   - `docker ps -a` shows the container still present (not stopped/removed).
   - The temp directory created by `_unzip_file`/`_get_code_dir` for that invocation is not removed from disk.
4. Repeating the invocation accumulates one leaked container and one leaked temp directory per failed run.

### Expected result
Regardless of why `_check_exit_state` raises, the container should still be stopped/removed and the temp directory cleaned up — cleanup should not be skippable by an exception raised partway through `_on_invoke_done`.

### Suggested fix
Wrap the cleanup steps so they always run regardless of `_check_exit_state`'s outcome, e.g.:

```python
def _on_invoke_done(self, container):
    try:
        if container:
            self._check_exit_state(container)
    finally:
        if container:
            self._container_manager.stop(container)
        self._clean_decompressed_paths()
```
(or equivalent try/finally ordering that guarantees `stop()` and `_clean_decompressed_paths()` run before/regardless of re-raising the OOM error.)

### Impact
Repeated local-invoke testing against a memory-constrained function accumulates dead Docker containers and temp directories over a test loop — not catastrophic, but a real, persistent leak that requires manual `docker container prune` / temp-dir cleanup, and gets worse the more a developer iterates on tuning `MemorySize` locally (the exact workflow most likely to repeatedly trigger this).

### Environment
- `aws-sam-cli` version: 1.165.0 (current `develop` branch, commit at time of testing)
- OS: macOS

Adityaj0:
### Which issue(s) does this change fix?

Fixes #9182

`LambdaRuntime._on_invoke_done()` calls `_check_exit_state(container)` before `self._container_manager.stop(container)` and `self._clean_decompressed_paths()`:

```python
def _on_invoke_done(self, container):
    if container:
        self._check_exit_state(container)
        self._container_manager.stop(container)
    self._clean_decompressed_paths()
```

`_on_invoke_done` is invoked unconditionally from `invoke()`'s `finally` block, but has no try/except of its own. When a function is OOM-killed, `_check_exit_state()` raises `ContainerFailureError`, and that exception propagates immediately out of `_on_invoke_done`, skipping both `_container_manager.stop(container)` and `_clean_decompressed_paths()`.

Every OOM'd invocation of `sam local invoke` (or `start-api`/`start-lambda`) therefore leaves a stopped-but-not-removed Docker container and an unzipped-code temp directory on disk. This accumulates over repeated local-invoke testing, which is exactly the workflow most likely to trigger it (iterating on `MemorySize` tuning).

### How does this change work?

Wrap the exit-state check in a `try/finally` so the container is always stopped and the temp directory is always cleaned up, regardless of whether `_check_exit_state` raises:

```python
def _on_invoke_done(self, container):
    try:
        if container:
            self._check_exit_state(container)
    finally:
        if container:
            self._container_manager.stop(container)
        self._clean_decompressed_paths()
```

The `ContainerFailureError` (or any other exception from `_check_exit_state`) still propagates after cleanup runs — this only fixes cleanup, not the error-reporting behavior for OOM.

### What tests ran and what were the results?

Added a regression test, `test_on_invoke_done_stops_container_and_cleans_paths_even_when_check_exit_state_raises`, in `tests/unit/local/lambdafn/test_runtime.py`. It fails against the pre-fix code (`stop()` is never called when `_check_exit_state` raises) and passes after the fix.

Full unit suite (`tests/unit`) passes: 9387 passed, 25 skipped.
`mypy` on the changed file: clean.

### Checklist

- [x] Add/update tests for this change
- [x] `make pr` passes locally (unit tests + mypy on changed files; full local `make pr` target run via `pytest tests/unit -q` and scoped `mypy`)
- [x] Write a clear PR title and description

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/samcli/local/docker/manager.py
+++ b/samcli/local/docker/manager.py
@@ -114,9 +114,14 @@ class ContainerManager:
 
         :param samcli.local.docker.container.Container container: Container to stop
         """
-        if self.do_shutdown_event:
-            container.stop()
-        container.delete()
+        # container.delete() is what actually removes the container; it must run even if
+        # container.stop() raises (e.g. a docker.errors.APIError), otherwise a failed stop()
+        # leaves the container running/orphaned with no further cleanup attempt.
+        try:
+            if self.do_shutdown_event:
+                container.stop()
+        finally:
+            container.delete()
 
     def pull_image(self, image_name, tag=None, stream=None):
         """
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -356,10 +356,13 @@ class LambdaRuntime:
         container: Container
            The current running container
         """
-        if container:
-            self._check_exit_state(container)
-            self._container_manager.stop(container)
-        self._clean_decompressed_paths()
+        try:
+            if container:
+                self._check_exit_state(container)
+        finally:
+            if container:
+                self._container_manager.stop(container)
+            self._clean_decompressed_paths()
 
     def _check_exit_state(self, container: Container):
         """
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -360,9 +360,11 @@ class LambdaRuntime:
             if container:
                 self._check_exit_state(container)
         finally:
-            if container:
-                self._container_manager.stop(container)
-            self._clean_decompressed_paths()
+            try:
+                if container:
+                    self._container_manager.stop(container)
+            finally:
+                self._clean_decompressed_paths()
 
     def _check_exit_state(self, container: Container):
         """
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -494,10 +494,19 @@ class LambdaRuntime:
         Clean the temporary decompressed code dirs
         """
         LOG.debug("Cleaning all decompressed code dirs")
+        # Snapshot and clear the list up front, under the lock, so that a failure removing one
+        # directory can't abort the loop and leave every entry (including ones already removed
+        # or added after this call started) stuck in the list forever -- since this list is only
+        # ever appended to, a stuck entry would otherwise block cleanup of every directory added
+        # in subsequent invokes for the lifetime of this LambdaRuntime instance.
         with self._lock:
-            for decompressed_dir in self._temp_uncompressed_paths_to_be_cleaned:
-                shutil.rmtree(decompressed_dir)
+            paths_to_clean = self._temp_uncompressed_paths_to_be_cleaned
             self._temp_uncompressed_paths_to_be_cleaned = []
+        for decompressed_dir in paths_to_clean:
+            try:
+                shutil.rmtree(decompressed_dir)
+            except OSError:
+                LOG.warning("Failed to remove temporary directory %s", decompressed_dir, exc_info=True)
 
     def get_or_create_emulator_container(self):
         """
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -360,11 +360,18 @@ class LambdaRuntime:
             if container:
                 self._check_exit_state(container)
         finally:
-            try:
-                if container:
+            # Best-effort cleanup: a failure here should not replace an in-flight exception
+            # from _check_exit_state() above (e.g. ContainerFailureError on OOM) with a raw
+            # Docker/OS error, and each step must run independently of the other's success.
+            if container:
+                try:
                     self._container_manager.stop(container)
-            finally:
+                except Exception:
+                    LOG.warning("Failed to stop/remove container during cleanup", exc_info=True)
+            try:
                 self._clean_decompressed_paths()
+            except Exception:
+                LOG.warning("Failed to clean decompressed code directories during cleanup", exc_info=True)
 
     def _check_exit_state(self, container: Container):
         """
--- a/samcli/local/lambdafn/runtime.py
+++ b/samcli/local/lambdafn/runtime.py
@@ -462,7 +462,12 @@ class LambdaRuntime:
 
         if code_path and os.path.isfile(code_path) and code_path.endswith(self.SUPPORTED_ARCHIVE_EXTENSIONS):
             decompressed_dir: str = _unzip_file(code_path, mount_symlinks=self._mount_symlinks)
-            self._temp_uncompressed_paths_to_be_cleaned += [decompressed_dir]
+            # Must hold the same lock _clean_decompressed_paths() uses to swap this list out --
+            # `start-api`/`start-lambda` run threaded, so a request thread appending here can
+            # race with a concurrent cleanup's snapshot-and-clear, either losing this entry
+            # entirely or (worse) having it removed by rmtree while still in use.
+            with self._lock:
+                self._temp_uncompressed_paths_to_be_cleaned.append(decompressed_dir)
             return decompressed_dir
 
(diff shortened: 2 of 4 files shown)


Adityaj0:
Verified this independently and it's correct.

Traced `ContainerManager.stop()` (`samcli/local/docker/manager.py:111`) → `Container.stop()`/`Container.delete()` (`samcli/local/docker/container.py:363,392`): both catch `docker.errors.NotFound` (fine) but re-raise `docker.errors.APIError` for anything other than the "removal of container ... is already in progress" message. So with the single flat `finally` block from the previous commit, a stop-time API error (unrelated to OOM — e.g. a transient daemon hiccup while removing the container) would still propagate out of `_on_invoke_done` before `_clean_decompressed_paths()` runs, leaving the temp dir behind — the exact leak class this PR is fixing, just moved one step later. It would also mask the original `ContainerFailureError` from the OOM check, since a new exception raised while another is propagating replaces it as what surfaces to the caller.

Pushed `3f23ceb`: nested the `stop()` call and `_clean_decompressed_paths()` into their own `try/finally`, matching what was suggested, so a failure in either cleanup step can't suppress the other, and the original exception from `_check_exit_state` still propagates normally when both cleanup steps succeed.

Added two regression tests in `tests/unit/local/lambdafn/test_runtime.py`:
- `test_on_invoke_done_cleans_paths_even_when_container_manager_stop_raises` — `stop()` raises, `_clean_decompressed_paths` must still be called
- `test_on_invoke_done_cleans_paths_when_both_check_exit_state_and_stop_raise` — OOM (`ContainerFailureError`) *and* a subsequent stop failure, cleanup still runs

Verified: `tests/unit/local/lambdafn/` full suite passes (134 passed), including the existing regression test from this PR and both new ones. `black --check` clean on both changed files.

Adityaj0:
Good catch — you're right that nesting the try/finally only fixed the "both steps always run" half of this; it didn't stop a cleanup failure from replacing whatever exception was already in flight.

Pushed a fix (67676eb) that makes both cleanup steps best-effort: `stop()` and `_clean_decompressed_paths()` are now each wrapped in their own `try/except`, logged via `LOG.warning(..., exc_info=True)` on failure instead of raising. This means:

- Both cleanup steps still always run, independently of each other (unchanged from the previous fix).
- A failure in either cleanup step no longer replaces an in-flight exception from `_check_exit_state()` — an OOM'd invoke now reliably surfaces `ContainerFailureError` ("Container invocation failed due to maximum memory usage") even if `stop()` or the temp-dir cleanup also fails.
- When there's no in-flight exception (the normal case), a cleanup failure is now swallowed with a log warning rather than surfacing a bare `docker.errors.APIError`/`OSError` to the user, since these are best-effort steps that shouldn't determine the invoke's overall success/failure.

Updated the existing `test_on_invoke_done_cleans_paths_even_when_container_manager_stop_raises` and `test_on_invoke_done_cleans_paths_when_both_check_exit_state_and_stop_raise` to match (the latter now asserts `ContainerFailureError` propagates, not the cleanup-only `RuntimeError`), and added two new tests covering `_clean_decompressed_paths()` itself raising, both alone and alongside the original `ContainerFailureError`. Full `tests/unit/local/lambdafn/` suite (136 tests) passes, ruff format clean.

Adityaj0:
Both confirmed real, and you're right that catching-and-logging at the call site wasn't enough — the leaks were still happening one level down, just silently now instead of loudly. Fixed both at the source (41f7ea8):

**`ContainerManager.stop()`** — wrapped `container.stop()` in `try/finally` so `container.delete()` (the call that actually removes the container) always runs, even if `stop()` raises. Added `test_must_call_delete_even_when_container_stop_raises`, confirmed it fails pre-fix (`delete` never called) and passes post-fix.

**`_clean_decompressed_paths()`** — you're right this is worse than a one-off leak: since `self._temp_uncompressed_paths_to_be_cleaned` is append-only and was only cleared after the loop finished, a single failing `rmtree()` call left *every* entry stuck in the list permanently, and every subsequent invoke on the same `LambdaRuntime` (reused for the life of `start-api`/`start-lambda`) would re-hit that same first bad entry and abort again — silently blocking cleanup of every dir added afterward. Snapshot-and-clear the list under the lock up front, then clean each path independently with its own try/except so one failure can't block the rest or get stuck. Went with your suggested shape essentially as-is. Added two tests: one confirming all paths are cleaned and the list is cleared on the happy path, and one with a failing entry in the middle of the list confirming the other two still get cleaned and the list still ends up empty (not stuck) despite the failure.

Full `tests/unit/local` (1303 tests) passes, ruff format clean.

Adityaj0:
Both real, and both go deeper than the previous fixes reached:

**Race between producer and consumer** — confirmed: `_get_code_dir()` (unchanged by this PR, line 465) does `self._temp_uncompressed_paths_to_be_cleaned += [decompressed_dir]` with no lock, while `_clean_decompressed_paths()` snapshots-and-clears that same list under `self._lock`. Since `+=` on a list attribute is a load/extend/store-back sequence, not an atomic in-place mutation, it can interleave with the lock-protected swap exactly as you described. Fixed by wrapping the append in `with self._lock:` too, so the snapshot is actually meaningful.

**Failed removals dropped permanently** — also confirmed as a regression from the previous fix: the snapshot-and-clear approach fixed "one bad entry blocks/stalls everything" but introduced "one bad entry silently vanishes forever," trading one leak class for another, quieter one. Requeued failed paths back into the list (under lock) instead of discarding them, so a transient failure gets retried on the next cleanup pass rather than leaking for the life of the process.

Pushed c298d1b. Added `test_appending_to_cleanup_list_holds_the_same_lock_cleanup_uses` (asserts `_get_code_dir` enters/exits `self._lock` and the entry lands in the list), and `test_failed_path_is_requeued_for_retry_not_dropped_permanently` (asserts a failed path stays queued and is actually retried — successfully — on the next call). Updated the earlier "list is cleared" assertion, which is no longer correct now that failures are requeued rather than dropped.

Full `tests/unit/local` (1305 tests) passes, ruff format clean.

Issue #8275, opened 2025-09-17
Bug: start-api does not use correct function when duplicate resource / construct ids in CDK


jrbarnard:
### Description:
When running sam local start-api with a CDK synthed template, it finds the wrong Lambda function to run, this is due to a `Serverless::Function` with a logical ID of `Lambda` and a `Lambda::Function` generated by CDK whose resource id (in the cdk path) is `Lambda`.
The resource metadata normalizer uses the resource id from the cdk path if it's there meaning there is conflict.

The logic doing the cdk path extraction is within `get_resource_id`, that is then set to the `SamResourceId` which is then used in place of the logical ID when doing function extraction.

There are a couple of mitigations we can look at that I'm aware of:
* Explicitly set a `SamResourceId` on the Lambda function - this seems unecessary just to get it running locally
* Change the resource ids / logical ids - this will cause redeployment, again seems unecessary to get it running locally

### Steps to reproduce:
* Create a CDK stack with two lambda functions, both Resource IDs should be `Lambda`, one should be nested in a sub construct
* Set up a rest api to point to the first lambda

E.g.
```
new Function(this, 'Lambda', {
    code: Code.fromInline(`
        exports.handler = (event, context, callback) => {
            callback(null, "Hello World!");
        };
    `),
    description: `test`,
    runtime: Runtime.NODEJS_22_X,
    handler: 'handler.handle',
  });
```

### Observed result:
* It runs the second function instead of the one linked to the API
* I've run it with debug and gone through the code and I can see during the function detection stage it pulls out the resource ID from the cdk path and uses that as the SamResourceId that is then used when identifying functions for invoke

### Expected result:
* It should run the API function code even if they share a resource ID
* I'd expect this normalisation function to take into account nested resources so we don't end up with duplicate ids - is there any reason we can't always just use logical IDs?

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

```
{
  "version": "1.139.0",
  "system": {
    "python": "3.13.3",
    "os": "macOS-15.6.1-arm64-arm-64bit-Mach-O"
  },
  "additional_dependencies": {
    "docker_engine": "27.4.0",
    "aws_cdk": "Not available",
    "terraform": "Not available"
  },
  "available_beta_feature_env_vars": [
    "SAM_CLI_BETA_FEATURES",
    "SAM_CLI_BETA_BUILD_PERFORMANCE",
    "SAM_CLI_BETA_TERRAFORM_SUPPORT",
[...]

PR #9186, opened 2026-08-15
fix: raise on ambiguous resource ID collision across sibling nested stacks

### LINKED ISSUE #9185: sam sync/local operations silently target the wrong physical resource when --resource-id collides across sibling nested stacks
### Description

`get_resource_by_id()` in `samcli/lib/providers/provider.py` is the core lookup used by `sam sync --resource-id <ID>` (and the underlying file-watch/code-trigger machinery used by `sam sync --watch`) to resolve a bare, unqualified resource logical ID to the deployed resource it should operate on.

When a bare ID (no `Stack/` path prefix) is given and it matches a resource with the same logical ID in **more than one nested stack**, with no match in the root stack to prefer, the function silently returns whichever stack happens to come first in its internal stack list — with no error, warning, or disambiguation prompt.

```python
def get_resource_by_id(
    stacks: List[Stack], identifier: ResourceIdentifier, explicit_nested: bool = False
) -> Optional[Dict[str, Any]]:
    ...
    search_all_stacks = not identifier.stack_path and not explicit_nested
    for stack in stacks:
        if stack.stack_path == identifier.stack_path or search_all_stacks:
            found_resource = None
            for logical_id, resource in stack.resources.items():
                ...
                if resource_id == identifier.resource_iac_id or (...):
                    found_resource = resource
                    break
            if found_resource:
                return cast(Dict[str, Any], found_resource)   # <-- first match wins, silently
    return None
```

### Reproduction

A SAM template using nested stacks where two different nested `AWS::CloudFormation::Stack` resources each contain a resource with the same logical ID (e.g. a shared/reusable nested-stack module deployed twice, or two independently-authored nested stacks that happen to reuse a common name like `Function1`) — no root-level resource with that ID.

```
sam sync --watch --resource-id Function1
```

### Expected behavior

Since the ID is genuinely ambiguous (matches deployed resources in two different nested stacks), the CLI should either ask the user to disambiguate or fail with a clear error telling them to qualify the ID with the full stack path (e.g. `NestedStackA/Function1`), which `ResourceIdentifier` already supports as valid input syntax.

### Actual behavior

The lookup silently returns whichever nested stack's resource happens to be first in the internal (construction-order-dependent) stack list. Since `sam sync` bypasses full CloudFormation deployment and pushes code directly to the resolved physical resource (e.g. via `UpdateFunctionCode`), this means the wrong live, deployed Lambda function (or other resource) can silently receive the local code changes intended for a different one — with no error printed and no indication anything went wrong.

Note: this is distinct from, and does not affect, the existing (and intentional/tested) precedence where a root-stack resource always wins over a same-named nested-stack resource for a bare ID — that behavior is correct and unaffected by this report. The bug is specifically the nested-vs-nested collision case, which has no precedence rule and no test coverage today.

### Fix

PR incoming — raises a clear, actionable error when a bare ID matches more than one nested stack with no root-stack match to prefer, instead of silently picking the first one.

Adityaj0:
Fixes #9185.

## Which issue(s) does this change fix?

#9185

## Scope note (added after review)

`get_resource_by_id()` and `get_resource_full_path_by_id()` are also used outside `sam sync`
--- `package_context.py`, `guided_context.py`, and `image_repository_validation.py` all call
`get_resource_full_path_by_id()` to map a user-supplied `--image-repositories` function ID to a
repository URI. The new `AmbiguousResourceIdentifier` therefore now also reaches
`sam package`/`sam deploy --guided`/image-repository validation for a template with two sibling
nested stacks sharing a logical ID that a user references there, not just the `sam sync
--resource-id` path this PR was originally scoped around. This is intentional (see the extended
docstring on `get_resource_full_path_by_id`): mapping an image repository to the wrong function is
the same "operates on the wrong resource" hazard motivating this whole change, and unlike `sam
sync`'s interactive usage there is no further confirmation step before it's baked into a deploy.
Flagging explicitly since it widens this PR's user-facing blast radius beyond the issue it was
filed against.

## Why is this change necessary?

`get_resource_by_id()` (`samcli/lib/providers/provider.py`) resolves a bare, unqualified resource logical ID (e.g. what \`sam sync --resource-id Function1\` and the underlying file-watch/code-trigger machinery for \`sam sync --watch\` use) by searching every stack and returning the first match, with no check for ambiguity:

```python
for stack in stacks:
    if stack.stack_path == identifier.stack_path or search_all_stacks:
        ...
        if found_resource:
            return cast(Dict[str, Any], found_resource)   # first match wins, silently
```

If two different **nested** stacks both contain a resource with the same logical ID, and there's no root-stack resource with that ID to prefer, this silently returns whichever stack happens to come first in the internal stack list — no error, no warning. Since \`sam sync\` bypasses full CloudFormation deployment and pushes local code changes directly to the resolved physical resource, an ambiguous \`--resource-id\` can silently push code to the wrong live, deployed Lambda function (or other resource).

## What does this change do?

Changes \`get_resource_by_id\` to detect this specific case and raise a new \`AmbiguousResourceIdentifier\` (a \`UserException\`, matching the existing \`ResourceNotFound\` etc. in \`samcli/commands/local/cli_common/user_exceptions.py\`) with a message telling the user how to disambiguate using the existing \`Stack/ResourceId\` identifier syntax that \`ResourceIdentifier\` already parses.

Importantly, this **preserves** the existing, intentional, and already-tested precedence rule that a root-stack resource always wins over a same-named nested-stack resource for a bare ID (see \`TestGetResourceByID::test_get_resource_by_id_implicit_root\`, which deliberately constructs a root-vs-nested collision and asserts root wins — that test still passes unchanged). The new error only fires for the previously-unhandled case: no root match, and two or more *nested* stacks collide.

## Description of how you validated changes

Added \`TestGetResourceByIDAmbiguousNestedStacks\` to \`tests/unit/commands/local/lib/test_provider.py\`:
- raises \`AmbiguousResourceIdentifier\` when two sibling nested stacks both have a matching resource and there's no root match
- does not raise when only one nested stack matches
- does not raise when the caller explicitly qualifies the ID with a stack path

All 118 tests in \`tests/unit/commands/local/lib/test_provider.py\` pass (115 existing + 3 new), including all pre-existing root/nested precedence tests, unchanged. Also ran the full \`tests/unit/lib/sync/\`, \`tests/unit/lib/utils/test_resource_trigger.py\`, \`tests/unit/lib/utils/test_code_trigger_factory.py\`, \`tests/unit/lib/utils/test_resource_type_based_factory.py\`, and \`tests/unit/lib/providers/\` suites (374 passed, 4 skipped, all pre-existing skips) — no regressions in any of \`get_resource_by_id\`'s call sites. \`black --check\` clean.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sam-cli/blob/develop/CONTRIBUTING.md) doc
- [x] Local run of \`brazil-build test\`/\`pytest\` passes for the affected tests (ran via a local venv + pytest directly)
- [x] Added unit tests covering the new behavior
- [ ] Added integration tests (not applicable — internal resolver bug, fully covered by unit tests; the failure mode requires two nested stacks with a colliding logical ID, which is exercised by the new unit test fixtures)
- [x] Did not modify or add to generated files

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*

### DIFF
--- a/samcli/commands/local/cli_common/user_exceptions.py
+++ b/samcli/commands/local/cli_common/user_exceptions.py
@@ -47,6 +47,12 @@ class ResourceNotFound(UserException):
     """
 
 
+class AmbiguousResourceIdentifier(UserException):
+    """
+    A bare resource logical ID (without a stack path) matched resources in more than one stack
+    """
+
+
 class InvalidLayerVersionArn(UserException):
     """
     The LayerVersion Arn given in the template is Invalid
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -896,35 +896,40 @@ def get_resource_by_id(
         `sam sync --resource-id`), so that specific case is now a clear, actionable error instead.
     """
     search_all_stacks = not identifier.stack_path and not explicit_nested
-    matches: List[Tuple[Stack, Dict[str, Any]]] = []
+    matches: List[Tuple[Stack, str, Dict[str, Any]]] = []
     for stack in stacks:
         if stack.stack_path == identifier.stack_path or search_all_stacks:
             found_resource = None
+            found_resource_id = None
             for logical_id, resource in stack.resources.items():
                 resource_id = ResourceMetadataNormalizer.get_resource_id(resource, logical_id)
                 if resource_id == identifier.resource_iac_id or (
                     not identifier.stack_path and logical_id == identifier.resource_iac_id
                 ):
                     found_resource = resource
+                    found_resource_id = resource_id
                     break
 
             if found_resource:
                 if not stack.stack_path:
                     # The root stack always takes priority over nested stacks.
                     return cast(Dict[str, Any], found_resource)
-                matches.append((stack, found_resource))
+                matches.append((stack, cast(str, found_resource_id), found_resource))
                 if not search_all_stacks:
                     break
 
     if len(matches) > 1:
-        colliding_paths = [get_full_path(stack.stack_path, identifier.resource_iac_id) for stack, _ in matches]
+        # Build the remediation paths from the *matched* (normalized) resource ID, not the raw
+        # identifier the user typed: for CDK/SamResourceId resources the two can differ, and a
+        # path built from the raw logical ID would never resolve on retry.
+        colliding_paths = [get_full_path(stack.stack_path, resource_id) for stack, resource_id, _ in matches]
         raise AmbiguousResourceIdentifier(
             f"Resource ID '{identifier.resource_iac_id}' is ambiguous: it matches resources in more than one "
             f"nested stack ({', '.join(colliding_paths)}). Qualify it with the full stack path, "
             f"e.g. '{colliding_paths[0]}'."
         )
 
-    return cast(Dict[str, Any], matches[0][1]) if matches else None
+    return cast(Dict[str, Any], matches[0][2]) if matches else None
 
 
 def get_resource_full_path_by_id(stacks: List[Stack], identifier: ResourceIdentifier) -> Optional[str]:
@@ -1091,8 +1096,7 @@ def get_function_build_info(
         loadable = imageuri and check_path_valid_type(imageuri) and Path(imageuri).is_file()
         if not buildable and not loadable:
             LOG.debug(
-                "Skip Building %s function, as it is missing either Dockerfile or DockerContext "
-                "metadata properties.",
+                "Skip Building %s function, as it is missing either Dockerfile or DockerContext metadata properties.",
                 full_path,
             )
             return FunctionBuildInfo.NonBuildableImage
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -11,9 +11,10 @@ from collections import namedtuple
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, NamedTuple, Optional, Set, Union, cast
+from typing import Any, Dict, Iterator, List, NamedTuple, Optional, Set, Tuple, Union, cast
 
 from samcli.commands.local.cli_common.user_exceptions import (
+    AmbiguousResourceIdentifier,
     InvalidFunctionPropertyType,
     InvalidLayerVersionArn,
     UnsupportedIntrinsic,
@@ -883,8 +884,19 @@ def get_resource_by_id(
     -------
     Dict
         Resource dict
+
+    Raises
+    ------
+    AmbiguousResourceIdentifier
+        If a bare logical ID (no stack path given) matches resources in more than one *nested*
+        stack, with no match in the root stack to prefer. The root stack, if it has a match, always
+        takes priority (this is relied upon by existing behavior); but silently picking the first
+        nested-stack match in stack-list order when there is no root match - and more than one
+        nested stack collides - could operate on the wrong physical resource (e.g.
+        `sam sync --resource-id`), so that specific case is now a clear, actionable error instead.
     """
     search_all_stacks = not identifier.stack_path and not explicit_nested
+    matches: List[Tuple[Stack, Dict[str, Any]]] = []
     for stack in stacks:
         if stack.stack_path == identifier.stack_path or search_all_stacks:
             found_resource = None
@@ -897,8 +909,22 @@ def get_resource_by_id(
                     break
 
             if found_resource:
-                return cast(Dict[str, Any], found_resource)
-    return None
+                if not stack.stack_path:
+                    # The root stack always takes priority over nested stacks.
+                    return cast(Dict[str, Any], found_resource)
+                matches.append((stack, found_resource))
+                if not search_all_stacks:
+                    break
+
+    if len(matches) > 1:
+        colliding_paths = [get_full_path(stack.stack_path, identifier.resource_iac_id) for stack, _ in matches]
+        raise AmbiguousResourceIdentifier(
+            f"Resource ID '{identifier.resource_iac_id}' is ambiguous: it matches resources in more than one "
+            f"nested stack ({', '.join(colliding_paths)}). Qualify it with the full stack path, "
+            f"e.g. '{colliding_paths[0]}'."
+        )
+
+    return cast(Dict[str, Any], matches[0][1]) if matches else None
 
 
 def get_resource_full_path_by_id(stacks: List[Stack], identifier: ResourceIdentifier) -> Optional[str]:
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -941,7 +941,16 @@ def get_resource_full_path_by_id(stacks: List[Stack], identifier: ResourceIdenti
     -------
     str
         return resource full path
+
+    Raises
+    ------
+    AmbiguousResourceIdentifier
+        If a bare logical ID (no stack path given) matches resources in more than one *nested*
+        stack, with no match in the root stack to prefer. Mirrors the precedence/ambiguity rule
+        enforced by get_resource_by_id, so the two entry points cannot disagree on what a bare ID
+        resolves to.
     """
+    matches: List[str] = []
     for stack in stacks:
         if identifier.stack_path and identifier.stack_path != stack.stack_path:
             continue
@@ -950,8 +959,19 @@ def get_resource_full_path_by_id(stacks: List[Stack], identifier: ResourceIdenti
             if resource_id == identifier.resource_iac_id or (
                 not identifier.stack_path and logical_id == identifier.resource_iac_id
             ):
-                return get_full_path(stack.stack_path, resource_id)
-    return None
+                if not stack.stack_path:
+                    # The root stack always takes priority over nested stacks.
+                    return get_full_path(stack.stack_path, resource_id)
+                matches.append(get_full_path(stack.stack_path, resource_id))
+                break
+
+    if len(matches) > 1:
+        raise AmbiguousResourceIdentifier(
+            f"Resource ID '{identifier.resource_iac_id}' is ambiguous: it matches resources in more than one "
+            f"nested stack ({', '.join(matches)}). Qualify it with the full stack path, e.g. '{matches[0]}'."
+        )
+
+    return matches[0] if matches else None
 
 
 def get_resource_ids_by_type(stacks: List[Stack], resource_type: str) -> List[ResourceIdentifier]:
--- a/samcli/lib/providers/provider.py
+++ b/samcli/lib/providers/provider.py
@@ -954,6 +954,18 @@ def get_resource_full_path_by_id(stacks: List[Stack], identifier: ResourceIdenti
         stack, with no match in the root stack to prefer. Mirrors the precedence/ambiguity rule
         enforced by get_resource_by_id, so the two entry points cannot disagree on what a bare ID
         resolves to.
+
+        Note this function is also used outside `sam sync --resource-id` (package_context,
+        guided_context, image_repository_validation, all mapping a user-supplied image function
+        ID to an image repository URI): an ambiguous match there means the wrong function could
+        silently receive the wrong image repository, which is the same "operates on the wrong
+        physical resource" hazard this function exists to prevent for sync, so raising here is
+        intentional rather than sync-specific. This is a deliberately *stricter* policy than
+        `SamFunctionProvider.get()` (used by e.g. `sam local invoke`), which instead warns and
+        picks the alphabetically-first match; that entry point resolves a *running* local
+        function for interactive use, where a warning is recoverable, whereas the callers of this
+        function feed the result into a deploy/package/validation decision without a further
+        confirmation step.
     """
     matches: List[str] = []
     for stack in stacks:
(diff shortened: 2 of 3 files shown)


Adityaj0:
Thanks for the review — verified this independently and it's correct.

I traced `get_resource_full_path_by_id` (`samcli/lib/providers/provider.py:930`, pre-fix at `develop`) line by line: it iterates `stacks` and returns on the *first* match (`return get_full_path(stack.stack_path, resource_id)` inside the inner loop), with no root-stack preference and no ambiguity check — exactly the same bug class this PR fixes in `get_resource_by_id`, just unaddressed here. And you're right about the blast radius: `package_context.py:140`, `guided_context.py:358`, and `image_repository_validation.py:138` all resolve user-supplied `--image-repositories Function1=uri` keys through this function, so a colliding bare ID could silently attach the wrong nested resource's full path (wrong ECR mapping, or a spurious "not all image functions provided" failure) on the packaging path.

Pushed `1ac9443`: applied the same root-stack-priority + nested-stack-collision rule to `get_resource_full_path_by_id` (raises the existing `AmbiguousResourceIdentifier` on nested/nested collision, root still wins unconditionally), and added `TestGetResourceFullPathByIDAmbiguousNestedStacks` in `tests/unit/commands/local/lib/test_provider.py` mirroring the existing `TestGetResourceByIDAmbiguousNestedStacks` (collision raises, single nested match doesn't, explicit stack-path qualification doesn't, root-priority-over-nested holds even when root appears last in the stack list).

Verified: full `tests/unit/commands/local/lib/test_provider.py` (122 passed), plus `tests/unit/lib/cli_validation/`, `tests/unit/commands/deploy/test_guided_context.py`, `tests/unit/commands/package/`, `tests/unit/lib/providers/`, `tests/unit/lib/sync/`, and the resource-trigger/code-trigger-factory/resource-type-based-factory suites (670 passed, 4 pre-existing skips, no regressions). `black --check` clean on both changed files.

Adityaj0:
Good catch, and confirmed real — for a bare logical ID that only matches via the logical-ID fallback (a CDK resource with `SamResourceId`/`aws:cdk:path` metadata, where the normalized ID differs from the logical ID), the suggested qualified retry path was built from the raw `identifier.resource_iac_id` and could never actually resolve, since a stack-qualified lookup only matches on the normalized ID.

Pushed a fix (4dce905) that tracks the matched (normalized) `resource_id` alongside each candidate in `matches`, and builds `colliding_paths` from that instead — matching `get_resource_full_path_by_id`'s existing (correct) approach, so the two entry points now agree on identical input.

Added `test_ambiguity_error_suggests_a_path_that_actually_resolves_for_cdk_resources`, using the exact scenario you described: two sibling nested stacks each with a `Function1` logical ID carrying `SamResourceId: Function1-x` metadata. It asserts the error message names `NestedStackA/Function1-x` (not `NestedStackA/Function1`), and — going one step further — that feeding the suggested qualified path back into `get_resource_by_id` actually resolves to the right resource, so the regression can't silently reappear if the message format changes again. Confirmed it fails on the pre-fix code (suggests `NestedStackA/Function1`, a dead end) and passes with the fix.

Full `tests/unit/commands/local/lib/test_provider.py` (123 tests) plus the broader dependent suites (517 tests) pass, ruff format clean.

Adityaj0:
Both fair, and I confirmed the caller behavior claim directly — \`package_context.py\`, \`guided_context.py\`, and \`image_repository_validation.py\` all treat a falsy/graceful return as the norm, so yes, this raise now reaches package/deploy/validate-image-repositories, not just \`sam sync --resource-id\`.

On reflection I don't think this should be scoped back or made non-fatal for those callers, though — an ambiguous match in \`--image-repositories Function1=<uri>\` means the wrong function could silently get the wrong image repository mapped to it, which is exactly the same "operates on the wrong physical resource" class of hazard the whole PR exists to close for sync. Silently picking one is arguably worse there than for sync, since there's no interactive confirmation step before it's baked into the deploy. So: erroring is the right call, but you're right it needed to be said explicitly rather than left implicit.

On the \`SamFunctionProvider.get()\` divergence: agreed that's a real inconsistency worth documenting rather than leaving as an surprise for the next person touching either function. I'm inclined to leave the two resolution strategies as-is rather than unify them in this PR — \`sam local invoke\`/\`SamFunctionProvider.get()\` resolves a function for one-off interactive/dev use where a warning-and-continue is recoverable, while this function's callers all feed the result into a deploy/package/validate decision with no further human-in-the-loop step. Unifying them would be a separate, larger behavioral change (and risks its own regressions) outside what this PR set out to fix.

Pushed 383d8f0: expanded the \`AmbiguousResourceIdentifier\` docstring section to state both of these explicitly — the broadened blast radius beyond sync, and why it's intentionally stricter than \`SamFunctionProvider.get()\` — so this reads as a documented decision rather than something the next reader has to rediscover. No behavior change, `test_provider.py` (123 tests) still green.

I'll also flag this in the PR description for anyone reviewing the merge, since "package/deploy can now hard-fail on ambiguous image-repository IDs" is worth surfacing at that level too, not just in a docstring.

roger-zhangg:
Maintainer review — apologies for the silence on this one.

**Short version:** the diagnosis is correct, the change is well targeted, and I verified the interesting claims rather than taking them on faith. I have one refactor I'd like before merge, one thing I want stated in the description, and a handful of nits. No design objection.

### Answering your open points from [this comment](https://github.com/aws/aws-sam-cli/pull/9186#issuecomment-5304740223)

**1. Raising (not warning) on the package/deploy path — agreed.** All three callers of `get_resource_full_path_by_id` iterate user-supplied `--image-repositories` keys and treat a falsy return as "skip this mapping": `samcli/commands/package/package_context.py:140`, `samcli/commands/deploy/guided_context.py:358`, `samcli/lib/cli_validation/image_repository_validation.py:138`. So the old behaviour genuinely was "attach the URI to whichever sibling won the race". `AmbiguousResourceIdentifier` is a `UserException`, i.e. a `click.ClickException` with `exit_code = 1` (`samcli/commands/exceptions.py:18-35`), so this surfaces as a clean `Error: ...` and not a traceback — and the remediation it prints is valid `--image-repositories` input, since those keys are compared against `function.full_path`. Fine by me.

One caveat, not a blocker: in `image_repository_validation.py` the call sits inside `_is_all_image_funcs_provided`, whose whole job is to return a bool that a validator converts into a `click.BadOptionUsage` carrying `option_name="--image-repositories"`. Raising through it replaces that option-scoped error with a generic one. If it's cheap, catching there and folding the text into `image_repos_error_msg` would preserve the option context.

**2. Not unifying with `SamFunctionProvider.get()` — agreed, out of scope.** I confirmed the divergence you documented is real: `samcli/lib/providers/sam_function_provider.py:147-160` sorts by `full_path.lower()`, logs a yellow warning, and returns `found_fs[0]`. Leaving it alone here is right.

### Please state in the description that this does not fix #8275

Both new checks are **per-stack**: the inner loop `break`s on the first match within a stack (`samcli/lib/providers/provider.py:911` and `:983`), so at most one candidate per stack ever reaches `matches`. An *intra*-stack normalized-ID collision is therefore still resolved silently by dict order. Checked against your branch with real `Stack` objects:

```
one root stack:
  logical_id 'Lambda'            (AWS::Serverless::Function, no metadata) -> resource_id 'Lambda'
  logical_id 'LambdaCDK8B889E1D' (aws:cdk:path MyStack/Lambda/Resource)   -> resource_id 'Lambda'
  get_resource_by_id('Lambda')           -> NO RAISE, returns the Serverless one
  get_resource_full_path_by_id('Lambda') -> NO RAISE, returns 'Lambda'
```

That is exactly #8275's reported setup, and it also resolves through `SamFunctionProvider.get()` rather than these two functions — so #8275 is untouched on both counts. Worth saying explicitly so nobody closes it off this PR.

### CDK breaking-change risk: low, but for a non-obvious reason

I went looking for this specifically, since "SAM CLI now hard-errors on duplicate resource IDs" sounds alarming for CDK users. Duplicate construct IDs across sibling *nested* stacks do **not** collide today, because `ResourceMetadataNormalizer.get_resource_id` refuses to parse an `aws:cdk:path` longer than 3 segments (`samcli/lib/samlib/resource_metadata_normalizer.py:260-271`) and falls back to the hash-suffixed CFN logical ID. A function inside a CDK nested stack has >=4 segments, so it keeps its unique logical ID. Verified:

```
sibling nested stacks, both constructs named 'Handler', no SamResourceId:
  NestedA: HandlerABC123 (aws:cdk:path App/MyStack/NestedA/Handler/Resource) -> 'HandlerABC123'
  NestedB: HandlerXYZ789 (aws:cdk:path App/MyStack/NestedB/Handler/Resource) -> 'HandlerXYZ789'
  get_resource_by_id('Handler') -> NO RAISE, None
```

So the new raise only fires where normalized IDs genuinely collide: hand-authored nested stacks reusing a logical ID (the #9185 case), or an explicitly duplicated `SamResourceId`. Both are truly ambiguous and a hard error is the right answer. Please put that reasoning in the description too — it's the difference between a scary-looking change and a narrow one.

Also verified positively:
- The 4dce905 fix works end to end: with `SamResourceId: Handler` duplicated in two nested stacks, the message suggests `NestedA/Handler`, and feeding that back into `get_resource_by_id` resolves to the right resource.
- Root priority holds with the root stack listed last, for both functions.
- `sam sync --resource-id` hits this synchronously on the main thread (`samcli/commands/sync/command.py:548`, before `executor.execute()`), so it fails fast at startup rather than mid-flow in a worker thread. Good.
- `--watch` cannot trip on it: `samcli/lib/sync/watch_manager.py:142` and `samcli/commands/sync/command.py:543` build identifiers via `get_all_resource_ids`/`get_resource_ids_by_type`, which emit full paths, so `stack_path` is always set. And `samcli/lib/sync/flows/layer_sync_flow.py:234` passes `explicit_nested=True`, which makes `search_all_stacks` false and confines the search to the root stack, so `matches` can never exceed one there.

### The one change I'd ask for

Please do take the bot's original suggestion and factor the shared rule (root priority + collision detection) into a single helper used by both functions. The rule is now written twice, and 4dce905 exists precisely because the two copies had already drifted on normalized-vs-raw ID. Today `get_resource_by_id` builds its candidate list at `provider.py:925` and `get_resource_full_path_by_id` builds the equivalent at `provider.py:982`; one `_resolve_bare_id(...)` returning the candidates would make the next drift impossible and would let you collapse the two long duplicated docstring sections into one.

### Nits, take or leave

- `provider.py:914` and `provider.py:979` use `not stack.stack_path`; there's already a `Stack.is_root_stack` property that is literally `return not self.stack_path` (`provider.py:685-689`).
- `matches` in `get_resource_by_id` (`provider.py:899`) stores the whole `Stack` but only ever reads `stack.stack_path`; `List[Tuple[str, str, Dict[str, Any]]]` would do and drops the `Tuple`/`Stack` coupling.
- The `LOG.debug` string reflow in `get_function_build_info` (`provider.py:1111`) is unrelated to this fix — `develop` still has the two-line form. Looks like a newer formatter ran; please drop it to keep the diff on-topic.
- Test asymmetry: `TestGetResourceFullPathByIDAmbiguousNestedStacks` has `test_root_stack_takes_priority_over_nested_match` with root listed last, but `TestGetResourceByIDAmbiguousNestedStacks` has no equivalent. Worth adding, because "root always wins regardless of list order" is a **new** guarantee for `get_resource_by_id` — the old code was purely first-match-wins, and root-first only held in practice because `SamLocalStackProvider.get_stacks` builds `stacks = [root] + children` (`samcli/lib/providers/sam_stack_provider.py:290`, `:325`). So the PR description's "already-tested precedence rule" is a little generous: the old test happened to list root first.

### Test results

- `python -m pytest tests/unit/lib/providers -q` -> **7 passed**. Heads up that this directory contains only `test_sam_stack_provider_language_extensions.py` and exercises none of the changed code; the relevant suite is `tests/unit/commands/local/lib/test_provider.py`.
- Your two new classes -> **8 passed**.
- `tests/unit/commands/local/lib/test_provider.py` + `tests/unit/lib/sync` + `tests/unit/lib/cli_validation` + `tests/unit/commands/package` + `tests/unit/commands/deploy/test_guided_context.py` + the resource-trigger / code-trigger-factory / resource-type-based-factory suites -> **786 passed, 4 skipped** at `383d8f0`.
- I also rebased `383d8f0` onto current `develop` locally: the rebase applies cleanly, no conflicts, and `test_provider.py` + `tests/unit/lib/providers` + `tests/unit/lib/sync` + `tests/unit/lib/cli_validation` -> **514 passed, 4 skipped**.

### On the BLOCKED merge state, in case you were wondering

It is **not** a rebase problem. The API reports `mergeable: true`, `develop`'s protection has `required_status_checks.strict = false` so being behind doesn't block, and the rebase is clean as noted above. It's blocked on review: `required_approving_review_count: 2` with `require_code_owner_reviews: true`, and there are currently zero human approvals — the three existing reviews are all `aws-sam-tooling-bot`, state `COMMENTED`. The required `PR Workflow` check also hasn't reported for this fork PR (`gh pr checks` shows only `lambda-tooling pr reviewer`), so a maintainer needs to authorize the workflow run first. Nothing for you to do there beyond the items above.

Leaving this as a comment rather than a formal review so the code owners can weigh in on the shared-helper refactor.

PR #9269, opened 2026-09-10
refactor(durable): Remove time scale configuration from emulator container

nvasiu:

Related to https://github.com/aws/aws-durable-execution-sdk-python/pull/523

The time scale feature has been removed from the durable execution emulator (see linked SDK PR). The `DURABLE_EXECUTIONS_TIME_SCALE` environment variable and the `DURABLE_EXECUTION_TIME_SCALE` container environment variable are no longer recognized by the emulator, so SAM CLI should stop setting them.

- Removes the `ENV_TIME_SCALE` class variable and its docstring
- Removes the `_get_emulator_time_scale()` method
- Updates `_get_emulator_environment()` to return an empty dict
- Updates unit tests to remove time scale assertions and the `custom_time_scale` test case

Users who previously set `DURABLE_EXECUTIONS_TIME_SCALE` will no longer see it forwarded to the emulator container. Since the emulator itself no longer supports this variable, there is no functional impact.

#### Mandatory Checklist

- [x] Write/update unit tests
- [x] `make pr` passes
- [x] Write documentation

### DIFF
--- a/samcli/local/docker/durable_functions_emulator_container.py
+++ b/samcli/local/docker/durable_functions_emulator_container.py
@@ -51,13 +51,6 @@ class DurableFunctionsEmulatorContainer:
     ENV_STORE_TYPE = "DURABLE_EXECUTIONS_STORE_TYPE"
 
     """
-    Allow overriding the timescale used by the emulator. For example, if you have
-    a context.wait(3 months), you probably don't want to actually wait 3 months in
-    a local development loop. This lets you override that!
-    """
-    ENV_TIME_SCALE = "DURABLE_EXECUTIONS_TIME_SCALE"
-
-    """
     Capture the logs from the emulator on cleanup - this can be useful for debugging
     what happened, since once the container is gone, the logs are too.
     """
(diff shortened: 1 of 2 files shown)


samcli/local/docker/durable_functions_emulator_container.py
```diff
@@ -50,13 +50,6 @@ class DurableFunctionsEmulatorContainer:
     """
     ENV_STORE_TYPE = "DURABLE_EXECUTIONS_STORE_TYPE"
 
-    """
-    Allow overriding the timescale used by the emulator. For example, if you have
-    a context.wait(3 months), you probably don't want to actually wait 3 months in
-    a local development loop. This lets you override that!
-    """
-    ENV_TIME_SCALE = "DURABLE_EXECUTIONS_TIME_SCALE"
```
bchampp, line 58: Rather than removing it entirely, is there a way we can still provide a feature for customers to use the new `skip_time` option added in the emulator instead?

bchampp (changes requested):

PR #9280, opened 2026-09-20
fix(local): give each start-api request its own requestId and time

### LINKED ISSUE #9279: Bug: every `sam local start-api` request gets the same requestContext.requestId, requestTimeEpoch and requestTime
### Description:

Under `sam local start-api` every request the Lambda sees carries the same `requestContext.requestId`. `requestTimeEpoch` and `requestTime` are frozen the same way. The values get fixed when the process imports the module and they stay put until you restart the dev server. Deployed API Gateway hands out fresh ones per request, so it's a local versus deployed split.

It hits anything that keys off those fields. Log correlation collapses, because every line from every request carries one id. An idempotency or dedupe check sees the second request as a replay of the first. A handler that works out request age from `requestTimeEpoch` watches that age grow without bound while the server stays up, so past its threshold it rejects everything.

The values are default arguments, in `RequestContext.__init__` and `RequestContextV2.__init__` (`samcli/local/events/api_event.py`) and in `construct_v2_event_http` (`samcli/local/apigw/event_constructor.py`). Python evaluates a default once, at import time.

### Steps to reproduce:

On `develop` at 5b20d0ec, with the package installed from source. This drives the same service `sam local start-api` serves, so no Docker or AWS credentials are needed:

```python
import json
import time

from samcli.lib.providers.provider import Api
from samcli.local.apigw.local_apigw_service import LocalApigwService
from samcli.local.apigw.route import Route

events = []


class Runner:
    def is_debugging(self):
        return False

    def invoke(self, name, event_str, stdout=None, stderr=None, tenant_id=None):
        events.append(json.loads(event_str))
        stdout.write_str(json.dumps({"statusCode": 200, "body": "ok"}))


route = Route(function_name="Fn", path="/hello", methods=["GET"], event_type=Route.API)
service = LocalApigwService(api=Api(routes=[route]), lambda_runner=Runner())
service.create()
client = service._app.test_client()

client.get("/hello")
time.sleep(1.1)
client.get("/hello")

for event in events:
    context = event["requestContext"]
    print(context["requestId"], context["requestTimeEpoch"], context["requestTime"])
```

```
python repro.py
```

### Observed result:

```
feb46947-11df-4541-9ead-6a908432ad85 1789873940239 20/Sep/2026:03:12:20 +0000
feb46947-11df-4541-9ead-6a908432ad85 1789873940239 20/Sep/2026:03:12:20 +0000
```

Same id, same epoch, same time string, for two requests sent 1.1 seconds apart. An HTTP API route with payload format 2.0 does the same, with `timeEpoch` in place of `requestTimeEpoch` and `time` in place of `requestTime`.

### Expected result:

Two different `requestId` values, and a `requestTimeEpoch` that tracks the clock, the way deployed API Gateway behaves.

1. OS: macOS 26.6.2 (Darwin 25.6.0)
2. `sam --version`: source checkout of `develop` at 5b20d0ec, where `samcli.__version__` is 1.166.2
3. AWS region: not used, the repro makes no AWS calls

Python 3.13.15.

Om-singhaI:

Fixes #9279

`RequestContext.__init__` builds its `requestId`, `requestTimeEpoch` and `requestTime` in default arguments, `construct_v2_event_http` does the same for `requestTimeEpoch` and `requestTime`, and `RequestContextV2.__init__` for `requestId`. Python evaluates a default once, when the module is imported, so `sam local start-api` hands every invocation the same id and the same timestamps for the whole life of the process. Deployed API Gateway gives each request its own.

`construct_v1_event` and `_build_v1_context` both build a `RequestContext` without passing `request_id`, `request_time_epoch` or `request_time`, so the frozen id reaches the Lambda event and the REQUEST authorizer event alike.

One request also has to agree with itself. `_request_handler` builds the function event, the authorizer event and the identity source context through separate calls, so an id resolved inside each builder would come out different in each of them. Deployed API Gateway passes one request id to the authorizer and to the function it authorizes, and an identity source of `$context.requestId` has to resolve to an id that is in the emitted event.

Separately, payload format 2.0 reports `requestContext.timeEpoch` in milliseconds. `construct_v2_event_http` and `_build_v2_context` used `int(time())`, which is seconds. The `HttpApi.json` sample event in this repo shows the unit: `"timeEpoch": 1583348638390` next to `"time": "12/Mar/2020:19:03:58 +0000"`.

* `RequestContext.__init__`: `request_id`, `request_time_epoch` and `request_time` now default to `None` and get resolved in the body.
* `RequestContextV2.__init__`: same for `request_id`.
* `construct_v2_event_http`: same for `request_time_epoch` and `request_time`, plus the missing `:param` lines.
* `_request_handler` resolves one request id and one timestamp pair per incoming request, and threads them into `_generate_lambda_event`, `_generate_lambda_authorizer_event` and `_valid_identity_sources`, which pass them down to `_build_v1_context`, `_build_v2_context`, `construct_v1_event` and `construct_v2_event_http`. Every parameter added along that path is a keyword argument defaulting to `None`.
* `construct_v2_event_http` and `_build_v2_context` now compute `timeEpoch` with `time_ns() // 1_000_000`, the expression the payload 1.0 path already used.
* Unit tests for each of those.

Callers that pass a value other than `None` get exactly what they got before, since the bodies only fill in `None`. Passing `None` explicitly now gets a generated value instead of `None`, and nothing in the tree does that. Two requests to the same local endpoint now differ in `requestContext.requestId`, `requestTimeEpoch` and `requestTime`, which is what the deployed service does. Within one request the REQUEST authorizer event, the function event and a `$context.requestId` identity source all agree on one id and one timestamp pair.

`requestContext.timeEpoch` in a payload 2.0 event is now about a thousand times larger, because it is milliseconds. A local function that read the old value as seconds sees a different number. Milliseconds is what the deployed service sends and what this repo's own sample event carries.

Existing parameter names and order are unchanged, the new ones are appended, and there are no new dependencies.

#### Testing

Python 3.13.15, macOS, on top of `develop` at 5b20d0ec.

* `make pr`: the `test-all` step is 9444 passed, 25 skipped, 31 subtests passed in 58.94s, with total coverage 94.16% against the 94% floor. `black --check setup.py samcli tests schema` leaves 1393 files unchanged, `ruff check samcli schema` passes, and `mypy` finds no issues in 1381 source files.
* `pytest tests/unit/local`: 1308 passed, 2 subtests passed in 21.27s. Base is 1299 plus the 9 new tests.
* `pytest tests/unit/local/apigw tests/unit/local/events`: 319 passed in 0.53s, base 310.
* Restoring only the three source files from `develop` and keeping the tests: 8 of the 9 new tests fail. The ninth asserts that one request uses one id, and on `develop` every id is the same frozen value, so it passes for the wrong reason. Restoring the source files from the first version of this commit instead, where each builder resolved its own id, makes that test fail along with the millisecond assertions.

Four existing assertions in `test_local_apigw_service.py` grew `request_id`, `request_time_epoch` and `request_time` entries, since `_request_handler` now passes those to the event constructors.

#### Mandatory Checklist

- [x] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods (no new production functions or methods; the new parameters on the typed service methods are annotated)
- [x] Write/update unit tests
- [ ] Write/update integration tests (the unit tests cover both payload versions)
- [x] Write/update functional tests if needed (not needed)
- [x] `make pr` passes
- [x] `make update-reproducible-reqs` if dependencies were changed (no dependency changes)
- [x] Write documentation (no user facing docs change)


Om-singhaI:
Both points are fixed. The branch still holds one commit, amended.

**One request id and one time per incoming request**

Resolving the values per instance stopped the process wide reuse, but it split a single request across several ids. `_request_handler` builds the function event, the authorizer event and the identity source context through separate calls, and each call drew its own uuid.

`_request_handler` now resolves one request id and one timestamp pair when the request arrives, then threads them into `_generate_lambda_event`, `_generate_lambda_authorizer_event` and `_valid_identity_sources`, which pass them down to `_build_v1_context`, `_build_v2_context`, `construct_v1_event` and `construct_v2_event_http`. The new parameters are keyword arguments defaulting to `None`, so a direct caller of any of those builders still generates its own values.

So a REQUEST authorizer now sees the same `requestContext.requestId` as the function it authorizes, and an identity source of `$context.requestId` resolves to the id that is in the emitted event.

**`timeEpoch` in milliseconds**

`construct_v2_event_http` and `_build_v2_context` both used `int(time())`, which is seconds. Payload format 2.0 reports `requestContext.timeEpoch` in milliseconds, and the `HttpApi.json` sample event in this repo agrees: `"timeEpoch": 1583348638390` sits next to `"time": "12/Mar/2020:19:03:58 +0000"`. Both places now use `time_ns() // 1_000_000`, the same expression the payload 1.0 path already used.

The assertion that pinned the seconds value is updated to milliseconds. I checked the rest of the tree for anything else on the seconds default. Two other fixtures build their own epoch value and pass it explicitly, so neither reads the default. The one feeding `construct_v2_event_http` moved to `time_ns() // 1_000_000`. The one in `test_api_event.py` stays on `int(time())`, since it only feeds `RequestContextV2`, which never computes a time.

**Tests**

There are 9 new tests. One drives `_request_handler` against an HTTP route with a REQUEST authorizer whose identity source is `$context.requestId`, then asserts the authorizer event, the function event and the resolved identity value all carry the same id, and the same `requestTimeEpoch` and `requestTime`. Another runs two requests and asserts the ids differ. The 2.0 constructor test patches `time_ns` and asserts millisecond values. Two more assert that an explicitly passed id and time reach the event untouched.

Four existing assertions in `test_local_apigw_service.py` grew `request_id`, `request_time_epoch` and `request_time` entries, because `_request_handler` now passes them to the event constructors.

Negative control: with the three source files restored from `develop` and the tests kept, 8 of the 9 new tests fail. The ninth is the one asserting a single request uses one id, and on `develop` every id is the same frozen value, so it passes for the wrong reason. Restoring the source files from the previous version of this commit instead makes that test fail, together with the millisecond assertions.

`make pr` passes on Python 3.13.15, macOS: 9444 passed, 25 skipped, 31 subtests passed in 58.94s, total coverage 94.16% against the 94% floor. `black --check` leaves 1393 files unchanged, `ruff check samcli schema` passes, and `mypy` reports no issues in 1381 source files.

